PR #269, opened 2014-06-24
Added one period back variables, for processing data in time periods past

rjurney:
Widely needed feature. One always processes data from last hour, day, week, etc.


suvodeep-pyne:

PR #277, opened 2014-07-08
Adding the ability to get yesterday's date

Kurios:
Working with Hadoop is fun. The problem with hadoop is, you never know when something is going to get done today. Especially when its not from your team. Our solution? Work off of yesterday's data. Now if only we knew when yesterday was...


xingfei:
I've got the same problem as you. Azkaban supports jexl expressions, you can get yesterday date using jexl.
Sample code is:
`day=$(new("org.joda.time.DateTime").minusDays(1).toString("yyyy-MM-dd"))`

suvodeep-pyne:

PR #308, opened 2014-08-25
Error Logs as Mail Attachments

tanvi09:
Included the functionality of adding " Inline message Errors" in the mail as attachments from the _job files. For fast tracking of errors, this saves time to go and check the logs. 

Properties are configurable via the config properties files and are defined as

``` properties
mail.attach.name=.*job.*.log
mail.inline.error=true
```

**_mail.attach.name**_ supports regex like  ._.log  or  ._.job.\* . This would attach files matching this regex as attachments from the job folder.
**_mail.inline.error**_ would also have the error statements as inline text in the mail body.

Hope this helps!


kaneda:
@kingster wouldn't just the log lines leading up to and include the *failure* be what you're looking for? @erwa

suvodeep-pyne:

PR #359, opened 2014-11-22
Increasing HeaderBufferSize on AzkabanExecutorServer

appanasatya:
No of total flows in running and preparing state increase leads to increase in headerbuffer size(by get api call of Line 703 of ExecutorManager.java) 

and there by leading to FULL HEAD (httpclient. Response Exception) as it exceeding 4k which is default value 
and there by throwing IOException from handleAjaxUpdateRequest of ExecutorServlet and there by all flows in preparing state are getting Evicted. (Evicting flow " + flow.getExecutionId() + ". The executor is unresponsive. Line 854 of ExecutorManager.java).

### DIFF
--- a/azkaban-execserver/src/package/conf/azkaban.properties
+++ b/azkaban-execserver/src/package/conf/azkaban.properties
@@ -25,6 +25,12 @@ executor.flow.threads=30
 jetty.connector.stats=true
 executor.connector.stats=true
 
+
 # uncomment to enable inmemory stats for azkaban
 #executor.metric.reports=true
 #executor.metric.milisecinterval.default=60000
+
+#ConnectorBufferSizes
+headerBufferSize=4096
+requestBufferSize=8192
+responseBufferSize=24576
--- a/azkaban-webserver/src/package/conf/azkaban.properties
+++ b/azkaban-webserver/src/package/conf/azkaban.properties
@@ -51,3 +51,8 @@ cache.directory=cache
 # JMX stats
 jetty.connector.stats=true
 executor.connector.stats=true
+
+#ConnectorBufferSizes
+headerBufferSize=4096
+requestBufferSize=8192
+responseBufferSize=24576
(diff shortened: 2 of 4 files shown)


appanasatya:
Corrected kb related spellchecks and made few corrections..

azkaban-execserver/src/package/conf/azkaban.properties
```diff
@@ -25,6 +25,12 @@ executor.flow.threads=30
...
 
+
 # uncomment to enable inmemory stats for azkaban
 #executor.metric.reports=true
-#executor.metric.milisecinterval.default=60000
\ No newline at end of file
+#executor.metric.milisecinterval.default=60000
+
+#ConnectorBufferSizes
+headerBufferSize=4096
```
johnyu0520: Could we make this name to be 
"jetty.connector.*"

This way it is more clear the setting is regarding jetty.connectors,
and also fits with the "jetty.connector.stats" specification above

azkaban-webserver/src/package/conf/azkaban.properties
```diff
@@ -51,3 +51,8 @@ cache.directory=cache
 # JMX stats
 jetty.connector.stats=true
 executor.connector.stats=true
+
+#ConnectorBufferSizes
+headerBufferSize=4096
```
johnyu0520: Could we make this name to be 
"jetty.connector.*"

This way it is more clear the setting is regarding jetty.connectors,
and also fits with the "jetty.connector.stats" specification above

johnyu0520:
Hey Satya Harish,
The change looks good, except for a minor name change.

There's a recent merge request for setHeaderBufferSize for exec server.  Expect some minor merge conflicts.  However, since your change is more comprehensive, I think we should use the code you have in place.

Thanks!
John

suvodeep-pyne:

PR #390, opened 2015-01-23
Validate the configurations of the validator plugins when the Azkaban we...

Victsm:
...b server starts.


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -80,6 +80,7 @@ public ProjectManager(ProjectLoader loader, Props props) {
     // itself.
     Props prop = new Props(props);
     prop.put(ValidatorConfigs.PROJECT_ARCHIVE_FILE_PATH, "initialize");
+    new XmlValidatorManager(prop);
```
erwa: Why do we need to create an XmlValidatorManager here even though we don't store it in any variable? It would be good to add a comment explaining this.

PR #393, opened 2015-01-29
Separate the warning and error messages generated by validator plugin.

Victsm:


azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1582,34 +1582,50 @@ private void ajaxHandleUpload(HttpServletRequest req,
...
+            for (String msg : report.getInfoMsgs()) {
+              switch (ValidationReport.getInfoMsgLevel(msg)) {
+                case ERROR:
+                  errorMsgs.append(ValidationReport.getInfoMsg(msg) + "<br/>");
+                  break;
+                case WARN:
+                  warnMsgs.append(ValidationReport.getInfoMsg(msg) + "<br/>");
+                  break;
+                default:
+                    break;
```
erwa: Indentation

erwa:
LGTM. BTW, did you figure out why Azkaban doesn't show any error message at all (instead of truncating them) when there are too many error messages/too many flows?

Victsm:
The problem is related to the size of the message that can fit in the Cookie. Azkaban puts these error/warning messages inside a Cookie and when the length of the message goes over 4000, the entire message gets discarded from the cookie. I added code to truncate the message at the length of 4000 to avoid missing messages again in the future.

Victsm:
Updated the pull request.

Issue #394, opened 2015-01-29
Property to override smtp port in azkaban.properties


sathish316:
While sending emails using smtp.gmail.com, the port number should be 587, but Azkaban uses default port of 25 which doesn't work. Can you please provide a property mail.port which can be passed as mail.smtp.port in EmailMessage#sendEmail?

hyurtseven81:
+1

ismailsimsek:
+1

sathish316:
There is a Pull request open for this - https://github.com/azkaban/azkaban/pull/395

HappyRay:
#872

PR #395, opened 2015-01-30
Fix to override smtp port in azkaban.properties

sathish316:
This commit fixes Issue #394


azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
```diff
@@ -26,6 +26,7 @@
   private boolean usesAuth;
 
   private String mailHost;
+  private String mailPort;
```
logiclord: Why not have mailPort as int since your are already parsing while calling t.connect ?

azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
```diff
@@ -38,6 +39,7 @@
   public AbstractMailer(Props props) {
     this.azkabanName = props.getString("azkaban.name", "azkaban");
     this.mailHost = props.getString("mail.host", "localhost");
+    this.mailPort = props.getString("mail.port", Emailer.DEFAULT_SMTP_PORT);
```
logiclord: there is a getInteger method, if you have int mailPoint

sathish316:
Makes sense. I've changed mailPort data type to int.

logiclord:
LGTM @hluu  any comments ?

azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
```diff
@@ -38,6 +39,7 @@
   public AbstractMailer(Props props) {
     this.azkabanName = props.getString("azkaban.name", "azkaban");
     this.mailHost = props.getString("mail.host", "localhost");
+    this.mailPort = props.getInt("mail.port", Emailer.DEFAULT_SMTP_PORT);
```
hluu: It seems a bit weird for AbstractMailer class to depend on one of its subclasses (Emailer).  How about defining the constant DEFAULT_SMTP_PORT inside AbstractMailer class?

sathish316:
This was done with the assumption that Emailer is for SMTP and there could be other emailers in the future (for example, using Mutt) where DEFAULT_SMTP_PORT in AbstractMailer will not make sense. Anyway, i've moved constant to AbstractMailer to avoid depending on subclass.

sathish316:
Is there any update on merging this fix to master?

jfgreen:
I just ran into this issue myself and was about to open my own PR, but then I saw this. Anything stopping a merge? I am happy to contribute if there are any issues that need resolving.

kddeisz:
Bumping this because I just ran into this as well. We just locked down our network ACLs and it'd be great if this could go through 587.

kddeisz:
@logiclord Any chance this could get merged?

logiclord:
@HappyRay Can you please have a look ? changes looks good to me.

HappyRay:
The branch is out of date. Could someone bring it up to date?

How was it tested? Any way to test it using an automated test?

sathish316:
This branch is synced with master.
There was no automation test. It was tested by deploying a patched version of Azkaban using gmail smtp config in azkaban.properties with mail.port=587.

kddeisz:
@HappyRay Since it's been updated - any way this could be merged? Short of implementing something with a docker compose I'm not sure how you would test this with automated tests.

HappyRay:
@sathish316 have you considered using mocking objects to verify at least that the mail class gets the correct configuration?

We don't use the non default port currently and we don't have resources to test every change manually. If you include some automated tests, you can be more confident that your feature won't break because of a later change.

wangqiaoshi:
@HappyRay @kddeisz , i  also need mailPort too. because of automatic  test for PR and get stuck. so i want to do a automatic test .
@HappyRay  do you mean add a mock object of mailServer?

HappyRay:
What I am looking for is some kind of automated tests that will ensure that the default case and the new feature continues to work.

HappyRay:
@jamiesjc if you refactor the email code, you may consider making the port configurable.

inramana:
Looks like this has been submitted as part of #872

PR #397, opened 2015-02-03
Added Pig jobtype documentation about "use.user.pig.jar" and clarified t...

erwa:
...oken cancellation

Here's what it looks like:
piguseuserpigjar


logiclord:
LGTM

PR #411, opened 2015-03-18
Changed 'Succeed' to 'Success' in the filter. Maps to what the table doe...

Antall:
...s display


erwa:
I think it should be "Succeeded", to match "Failed" and "Killed".

Antall:
A flow that has completed successfully shows the status of 'Success' in the table, I thought the filter should represent what the table displays. Are you suggesting that the table should also show "Succeeded" in the status column?

erwa:
Yeah, I think "Succeeded" makes more grammatical sense.

Antall:
How do I run tests across the entire project?

erwa:
Does `gradle test` in the root of the project work? I also think tests are automatically run when you do a `gradle distZip`. If you want to skip tests, you can do `gradle distZip -x test`.

suvodeep-pyne:

PR #416, opened 2015-03-24
[dev-email-errors] Email errors inline or attached

kaneda:
- Updated https://github.com/azkaban/azkaban/pull/308 to take in a max lines, so that inline emails don't exceed 5 errors by default
- Ensure that these settings default to off


kaneda:
@erwa This should be a lot safer in all cases. The features default to off and there's a default max amount of errors to give in the inline message.

I also took the liberty of reformatting this to style best practices and removed the import .* (I personally don't like this at all, not for efficiency but for readability reasons)

kaneda:
@erwa could take a look at this when you get a sec?

kaneda:
@erwa any chance you could take a peak at this when you get a moment?

kaneda:
@davidzchen could you take a peak when you get a sec?

kaneda:
@davidzchen / @erwa any chance you could look at this?

Issue #418, opened 2015-04-21
azkaban memory awareness


weikang2002:
azkaban box is hit by out of memory quite often because azkaban executor server doesn't check system memory availability when spawning new processes. We are fixing this issue by the following.

1. We are adding memory check when executor server spawns new processes. System free memory info is obtained from /proc/meminfo. (One additional Azkaban property "executor.memCheck.interval" can be used to determine how often to check meminfo by the dedicated thread. Two additional jobtype plugin properties "memCheck.enabled", "memCheck.freeMemDecrAmt" can be used to determine if we should do the memcheck or not at runtime)

2. We log top memory consumers/processes in the JVM shutdown hook (it will be called when OOM happens, and system decides to kill web server or executor server)

3. We add job checks when user uploads their zip files to web server, to make sure Xms Xmx is not too big. Default values for max Xms and max Xmx is 1G and 2G respectively. (two additional azkaban properties "job.max.Xms", "job.max.Xmx" can be used to override default vaules)

PR #422, opened 2015-04-22
add system memory awareness to Azkaban

weikang2002:
refer to issue #418

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
@@ -152,8 +152,8 @@ public class JavaProcessJob extends ProcessJob {
   protected Pair<Long, Long> getProcMemoryRequirement() {
     String strInitMemSize = getInitialMemorySize();
     String strMaxMemSize = getMaxMemorySize();
-    long initMemSize = azkaban.utils.Utils.parseMemString2KB(strInitMemSize);
-    long maxMemSize = azkaban.utils.Utils.parseMemString2KB(strMaxMemSize);
+    long initMemSize = azkaban.utils.Utils.parseMemString(strInitMemSize);
+    long maxMemSize = azkaban.utils.Utils.parseMemString(strMaxMemSize);
 
     return new Pair<Long, Long>(initMemSize, maxMemSize);
   }
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import org.apache.log4j.Logger;
 
+import azkaban.utils.Pair;
 import azkaban.utils.Props;
 
 public class JavaProcessJob extends ProcessJob {
@@ -146,5 +147,14 @@ public class JavaProcessJob extends ProcessJob {
     }
 
     return "";
+  }
+
+  protected Pair<Long, Long> getProcMemoryRequirement() {
+    String strInitMemSize = getInitialMemorySize();
+    String strMaxMemSize = getMaxMemorySize();
+    long initMemSize = azkaban.utils.Utils.parseMemString2KB(strInitMemSize);
+    long maxMemSize = azkaban.utils.Utils.parseMemString2KB(strMaxMemSize);
+
+    return new Pair<Long, Long>(initMemSize, maxMemSize);
   }
 }
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -53,8 +53,8 @@ public class ProcessJob extends AbstractProcessJob {
     }
 
     Pair<Long, Long> memPair = getProcMemoryRequirement();
-    boolean memGranted = SystemMemoryInfo.requestMemory(memPair.getFirst(), memPair.getSecond());
-    if (!memGranted) {
+    boolean isMemGranted = SystemMemoryInfo.canSystemGrantMemory(memPair.getFirst(), memPair.getSecond());
+    if (!isMemGranted) {
       throw new Exception(String.format("Cannot request memory (Xms %d kb, Xmx %d kb) from system for job %s",
               memPair.getFirst(), memPair.getSecond(), getId()));
     }
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -38,6 +38,8 @@ public class ProcessJob extends AbstractProcessJob {
   public static final String COMMAND = "command";
   private static final long KILL_TIME_MS = 5000;
   private volatile AzkabanProcess process;
+  private static final String MEMCHECK_ENABLED = "memCheck.enabled";
+  private static final String MEMCHECK_FREEMEMDECRAMT = "memCheck.freeMemDecrAmt";
 
   public ProcessJob(final String jobId, final Props sysProps,
       final Props jobProps, final Logger log) {
@@ -52,11 +54,14 @@ public class ProcessJob extends AbstractProcessJob {
       handleError("Bad property definition! " + e.getMessage(), e);
     }
 
-    Pair<Long, Long> memPair = getProcMemoryRequirement();
-    boolean isMemGranted = SystemMemoryInfo.canSystemGrantMemory(memPair.getFirst(), memPair.getSecond());
-    if (!isMemGranted) {
-      throw new Exception(String.format("Cannot request memory (Xms %d kb, Xmx %d kb) from system for job %s",
-              memPair.getFirst(), memPair.getSecond(), getId()));
+    if (sysProps.getBoolean(MEMCHECK_ENABLED, true)) {
+      long freeMemDecrAmt = sysProps.getLong(MEMCHECK_FREEMEMDECRAMT, 0);
+      Pair<Long, Long> memPair = getProcMemoryRequirement();
+      boolean isMemGranted = SystemMemoryInfo.canSystemGrantMemory(memPair.getFirst(), memPair.getSecond(), freeMemDecrAmt);
+      if (!isMemGranted) {
+        throw new Exception(String.format("Cannot request memory (Xms %d kb, Xmx %d kb) from system for job %s",
+                memPair.getFirst(), memPair.getSecond(), getId()));
+      }
     }
 
     List<String> commands = null;
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -26,7 +26,9 @@ import org.apache.log4j.Logger;
 
 import azkaban.jobExecutor.utils.process.AzkabanProcess;
 import azkaban.jobExecutor.utils.process.AzkabanProcessBuilder;
+import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import azkaban.utils.SystemMemoryInfo;
 
 /**
  * A job that runs a simple unix command
@@ -48,6 +50,13 @@ public class ProcessJob extends AbstractProcessJob {
       resolveProps();
     } catch (Exception e) {
       handleError("Bad property definition! " + e.getMessage(), e);
+    }
+
+    Pair<Long, Long> memPair = getProcMemoryRequirement();
+    boolean memGranted = SystemMemoryInfo.requestMemory(memPair.getFirst(), memPair.getSecond());
+    if (!memGranted) {
+      throw new Exception(String.format("Cannot request memory (Xms %d kb, Xmx %d kb) from system for job %s",
+              memPair.getFirst(), memPair.getSecond(), getId()));
     }
 
     List<String> commands = null;
@@ -99,6 +108,10 @@ public class ProcessJob extends AbstractProcessJob {
 
     // Get the output properties from this job.
     generateProperties(propFiles[1]);
+  }
+
+  protected Pair<Long, Long> getProcMemoryRequirement() {
+    return new Pair<Long, Long>(0L, 0L);
   }
 
   protected void handleError(String errorMsg, Exception e) throws Exception {
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -115,6 +115,13 @@ public class ProcessJob extends AbstractProcessJob {
     generateProperties(propFiles[1]);
   }
 
+  /**
+   * This is used to get the min/max memory size requirement by processes.
+   * SystemMemoryInfo can use the info to determine if the memory request
+   * can be fulfilled. For Java process, this should be Xms/Xmx setting.
+   *  
+   * @return pair of min/max memory size
+   */
   protected Pair<Long, Long> getProcMemoryRequirement() {
     return new Pair<Long, Long>(0L, 0L);
   }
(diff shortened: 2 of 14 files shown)


azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -409,4 +408,27 @@ public static String createPeriodString(ReadablePeriod period) {
...
     return periodStr;
   }
+
+  public static long parseMemString2KB(String strMemSize) {
+    if (strMemSize == null) {
+      return 0L;
+    }
+
+    long size = 0L;
+    if (strMemSize.endsWith("g") || strMemSize.endsWith("G")
```
logiclord: It's not necessary that we will have an alphabet in the end for xmx. Things like -Xmx83886080 is also valid in bytes, which will make your code return 0 KB size.

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
```diff
@@ -366,6 +376,27 @@ private void constructFlow(Flow flow, Node node, Set<String> visited) {
...
+  private void jobPropertiesCheck() {
+    String maxXms = props.getString(JOB_MAX_XMS, MAX_XMS_DEFAULT);
+    String maxXmx = props.getString(JOB_MAX_XMX, MAX_XMX_DEFAULT);
+    long sizeMaxXms = azkaban.utils.Utils.parseMemString2KB(maxXms);
+    long sizeMaxXmx = azkaban.utils.Utils.parseMemString2KB(maxXmx);
+
+    for (String jobName : jobPropsMap.keySet()) {
+      Props jobProps = jobPropsMap.get(jobName);
+      String xms = jobProps.getString(XMS, null);
+      if (xms != null && azkaban.utils.Utils.parseMemString2KB(xms) > sizeMaxXms) {
```
logiclord: Do we have properties resolved at this point ? Someone may have a 
XMS={FROM_FLOW} which will not be a valid value, if unresolved ?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -50,6 +52,13 @@ public void run() throws Exception {
       handleError("Bad property definition! " + e.getMessage(), e);
     }
 
+    Pair<Long, Long> memPair = getProcMemoryRequirement();
```
logiclord: It will be a good I idea to have a config to enable disable this feature from azkaban.properties or even curl requests

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -0,0 +1,165 @@
...
+ * This class is used to maintain system memory information. Processes utilizing
+ * large amount of memory should consult this class to see if the system has enough
+ * memory to proceed the operation.
+ * 
+ * Memory information is obtained from /proc/meminfo, so only Unix/Linux like system
+ * will support this class.
+ * 
+ * All the memory size used in this function is in KB
+ */
+public class SystemMemoryInfo {
```
logiclord: It will be a good idea to abstract SystemMemoryInfo implementation from the caller. Just wondering.. should we make it an interface SystemMemoryInterface and have SystemMemoryInfo implement it so later on we can change implementation without affecting client code or even use reflection to instantiate class reading conf ?

logiclord:
Looks Good. Suggesting some minor changes.

azkaban-common/src/main/java/azkaban/utils/Utils.java
hluu: I wonder if there is already existing code for parsing xms and xms settings?

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
```diff
@@ -60,8 +67,9 @@
   private Set<String> errors;
   private Set<String> duplicateJobs;
 
-  public DirectoryFlowLoader(Logger logger) {
+  public DirectoryFlowLoader(Props props, Logger logger) {
```
hluu: This may break existing code because a change in the constructor signature.  Can we overload the constructor?  Whether we use the props, needs to guard against props being null.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -0,0 +1,165 @@
...
+      freeMemDecrAmt = memDecrAmt;
+
+      //schedule a thread to read it
+      logger.info("Scheduled thread to read /proc/meminfo every 30 seconds");
+      scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+      scheduledExecutorService.scheduleAtFixedRate(new MemoryInfoReader(), 0, 30, TimeUnit.SECONDS);
+    }
+  }
+
+  public synchronized static boolean requestMemory(long xms, long xmx) {
```
hluu: Please document the behavior and use conventional method name when a method return a boolean

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -0,0 +1,165 @@
...
+    memCheckEnabled = memChkEnabled && memInfoExists;
+    if (memCheckEnabled) {
+      //initial reading of the mem info
+      readMemoryInfoFile();
+      freeMemDecrAmt = memDecrAmt;
+
+      //schedule a thread to read it
+      logger.info("Scheduled thread to read /proc/meminfo every 30 seconds");
+      scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+      scheduledExecutorService.scheduleAtFixedRate(new MemoryInfoReader(), 0, 30, TimeUnit.SECONDS);
```
hluu: can we make 30 seconds configurable?  the default can be 30 seconds.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -0,0 +1,165 @@
...
+      } catch (IOException e) {
+        logger.error("Exception in closing the buffered reader", e);
+      }
+    }
+  }
+
+  static class MemoryInfoReader implements Runnable {
+    @Override
+    public void run() {
+      readMemoryInfoFile();
```
hluu: It would be good to add a catch all clause and log out the exception to make it easier to debug.

azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -409,4 +408,27 @@ public static String createPeriodString(ReadablePeriod period) {
 
     return periodStr;
   }
+
+  public static long parseMemString2KB(String strMemSize) {
```
hluu: What is the significance of 2KB at the end of the method name?

azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -802,6 +804,12 @@ public static void main(String[] args) throws Exception {
     Runtime.getRuntime().addShutdownHook(new Thread() {
 
       public void run() {
+        try {
+          logTopMemoryConsumers();
+        } catch (Exception e) {
```
hluu: Should catch Throwable to catch unexpected exception so it doesn't bring down the web server

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -50,6 +52,13 @@ public void run() throws Exception {
...
+    boolean memGranted = SystemMemoryInfo.requestMemory(memPair.getFirst(), memPair.getSecond());
```
johnyu0520: isMemoryAvailable (or something similar) might be a better variable name for the boolean, because we are not actually asking for memory like how malloc is doing it.
Otherwise, at least name it in the boolean variable name convention, e.g. isMemGranted

azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
```diff
@@ -147,4 +148,13 @@ protected String createArguments(List<String> arguments, String separator) {
 
     return "";
   }
-}
+
+  protected Pair<Long, Long> getProcMemoryRequirement() {
```
johnyu0520: Maybe add quick a comment on what this method is doing and how is it computing the ProcMemory?   It doesn't seem like a trivial method.

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
johnyu0520: +1
you might need to do
loadAllProperties();
prepareAllJobProps(flow, this);

to have the variables resolved?

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
johnyu0520: Given this is a static method, users/developers may not realize that an init needs to be called first.
In this case do you want to help the users/developers out by doing init for them, print a warning, or something of that sort?

azkaban-common/src/main/java/azkaban/utils/Utils.java
weikang2002: Fixed to take care of the bytes situation.

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
weikang2002: Fixed to resolve the properties.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
weikang2002: Mem check logic can be disabled by putting "executor.memCheck.enabled=false" in azkaban.properties.

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
weikang2002: All existing code are taken care of by adding additional param.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
weikang2002: Fixed by changing method name.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
weikang2002: I am afraid this feature introduced too many configurable properties. Let's settle with 30 seconds for now. We can see if we have this needs or not later

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
weikang2002: Catch Throwable for the entire thread logic.

azkaban-common/src/main/java/azkaban/utils/Utils.java
weikang2002: Removed 2KB

azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
weikang2002: We are already in the shutting down process, so no need for that.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
weikang2002: Change the method, varable name.

weikang2002:
made some code changes based on comments. Will close this one, and make another pull request with newest changes.

azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -409,4 +408,35 @@ public static String createPeriodString(ReadablePeriod period) {
...
+        || strMemSize.endsWith("m") || strMemSize.endsWith("M")
+        || strMemSize.endsWith("k") || strMemSize.endsWith("K")) {
+      String strSize = strMemSize.substring(0, strMemSize.length() - 1);
+      size = Long.parseLong(strSize);
+    } else {
+      size = Long.parseLong(strMemSize);
+    }
+
+    long sizeInKb = 0L;
+    if (strMemSize.endsWith("g") || strMemSize.endsWith("G")) {
```
johnyu0520: this part need to handle strMemSize.endsWith("k") as well.

weikang2002: Good catch. Fixed.

weikang2002:
I put in the newest code, please take a look. As discussed, "memCheck.enabled", "memCheck.freeMemDecrAmt" properties will be at the jobType plugins level. Because of this, javaprocess won't be covered by these new properties. However, javaprocess is rarely used anyway.

azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -409,4 +408,37 @@ public static String createPeriodString(ReadablePeriod period) {
 
     return periodStr;
   }
+
+  /**
+   * @param strMemSize : memory string in the format such as 1G, 500M, 3000K, 5000 
+   * @return : long value of memory amount in kb
+   */
+  public static long parseMemString(String strMemSize) {
```
hluu: Are there tests for this method?

azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
hluu: Ah, got it.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
hluu: We can make it configurable with a default value.  That will give us an opportunity to adjust if needed.

johnyu0520:
LGTM

weikang2002:
Another revisioin: make free-mem-check-thread interval configurable. Added some unit test for pasring mem string.

azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
hluu: Any comments on John's suggestion?

azkaban-common/src/main/java/azkaban/utils/Utils.java
weikang2002: test cases added

azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
weikang2002: comments added for this func

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -0,0 +1,174 @@
...
+    File f = new File(MEMINFO_FILE);
+    memCheckEnabled = f.exists() && !f.isDirectory();
+    if (memCheckEnabled) {
+      //initial reading of the mem info
+      readMemoryInfoFile();
+
+      //schedule a thread to read it
+      logger.info(String.format("Scheduled thread to read /proc/meminfo every %d seconds", memCheckInterval));
+      scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+      scheduledExecutorService.scheduleAtFixedRate(new MemoryInfoReader(), 0, memCheckInterval, TimeUnit.SECONDS);
```
hluu: Add else condition for non-Linux system and print out the memory reader will not enabled

azkaban-common/src/main/java/azkaban/utils/DirectoryFlowLoader.java
logiclord: Actually these sort of changes used to break reportal deployment, which involved replacing jars e.g. replacing azkaban-common jar but now azkaban-executor jar. But now things should be fixed as we have gradle build process for reportal clusters.

logiclord:
LGTM

weikang2002:
Test is done. These multiple commits will resolve #418

hluu:
LGTM

Issue #430, opened 2015-05-26
Project whitelist support


weikang2002:
We need a project whitelist config file, so that some projects can be excluded from some actions in Azkaban.

One use case is the job/process memory checking we recently added (issue #418), some projects may have valid rationale to skip the memory check. In this case, project whitelist can be used to add these projects in.

The project whitelist will be loaded when azkaban starts up. We can also dynamically reload it without restarting azkaban like the following. Here the session must be the admin's session.

curl -k -b azkaban.browser.session.id={session_id} "http://hostname:port/manager?reloadProjectWhitelist=true"

PR #431, opened 2015-05-26
project whitelist support

weikang2002:
Refer to issue #430

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -55,7 +55,7 @@ public class ProcessJob extends AbstractProcessJob {
       handleError("Bad property definition! " + e.getMessage(), e);
     }
 
-    if (sysProps.getBoolean(MEMCHECK_ENABLED, true) && jobProps.getBoolean(AZKABAN_MEMORY_CHECK, false)) {
+    if (sysProps.getBoolean(MEMCHECK_ENABLED, true) && jobProps.getBoolean(AZKABAN_MEMORY_CHECK, true)) {
       long freeMemDecrAmt = sysProps.getLong(MEMCHECK_FREEMEMDECRAMT, 0);
       Pair<Long, Long> memPair = getProcMemoryRequirement();
       boolean isMemGranted = SystemMemoryInfo.canSystemGrantMemory(memPair.getFirst(), memPair.getSecond(), freeMemDecrAmt);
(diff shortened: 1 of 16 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectWhitelist.java
```diff
@@ -0,0 +1,143 @@
...
+ * This class manages project whitelist defined in xml config file.
+ * An single xml config file contains different types of whitelisted
+ * projects. For additional type of whitelist, modify WhitelistType enum.
+ * 
+ * The xml config file should in the following format. Please note
+ * the tag <MemoryCheck> is same as the defined enum MemoryCheck
+ * 
+ * <ProjectWhitelist>
+ *  <MemoryCheck>
+ *      <project projectname="project1" />
```
logiclord: How about using project Id ?
Project names can be re-assigned if original owner deletes existing project with a given name and another user create another project with same name, which will give special privileges to new project.

logiclord:
One suggestion. It will be good to test on a cluster, interface changes might break some dependency of azkaban-plugins on azkaban.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -54,7 +55,7 @@ public void run() throws Exception {
       handleError("Bad property definition! " + e.getMessage(), e);
     }
 
-    if (sysProps.getBoolean(MEMCHECK_ENABLED, true)) {
+    if (sysProps.getBoolean(MEMCHECK_ENABLED, true) && jobProps.getBoolean(AZKABAN_MEMORY_CHECK, false)) {
```
johnyu0520: jobProps.getBoolean(AZKABAN_MEMORY_CHECK, false)
should this be defaulted to true? so that we will by default check for memory?

azkaban-execserver/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -651,6 +652,11 @@ private void prepareJobProperties(ExecutableNode node) throws IOException {
     node.setInputProps(props);
   }
 
+  private void customizeJobProperties(Props props) {
```
johnyu0520: do we want to name the function a bit better so that we know what exactly is being "customized"?
maybe insertMemoryCheckToJobProperties (or something)

otherwies commenting the method would also help

azkaban-common/src/main/java/azkaban/project/ProjectWhitelist.java
weikang2002: Project id is just hard to read.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
weikang2002: This is to be consistent with ExecutionOptions.memoryCheck, whose default value is false. But according to the logic, AZKABAN_MEMORY_CHECK should always be in job properties.

azkaban-execserver/src/main/java/azkaban/execapp/FlowRunner.java
weikang2002: I will add some comments.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
weikang2002: OK, change both places to use 'true' as default value.

logiclord:
LGTM

PR #445, opened 2015-07-13
Catch exceptions in individual triggers

logiclord:
To fix an Issue that leads to complete break down of Azkaban schedule triggering mechanism.  We observed this issue when a scheduled job has SLA dependency on a job which does not exist in flow that was resulting in a null pointer exception.


logiclord:
Issue #446

johnyu0520:
LGTM

one minor suggestion is in the catch block, we should put either logs or comments to the effect of "skipping this trigger, moving on to the next one" (in addition to the logging the error, as you have already done)

logiclord:
Thanks John. I have added the comment.

PR #449, opened 2015-07-28
Fix log level/message in two places

elad:


hluu:
LGTM.  Thanks for making this fix.

Issue #450, opened 2015-07-28
Reduce ajax refresh interval for flow status and log update page


logiclord:
We need to reduce ajax refresh interval for flow status and log update page. Currently we have a 5 sec and 30 sec interval respectively. Such short interval creates a lot of unnecessary jetty connections on webserver and cause delays for real users.

PR #451, opened 2015-07-28
Reducing ajax refresh interval for flow status and log update page #450

logiclord:
We need to reduce ajax refresh interval for flow status and log update page. Currently we have a 5 sec and 30 sec interval respectively. Such short interval creates a lot of unnecessary jetty connections on webserver and cause delays for real users.


hluu:
LGTM

johnyu0520:
Looks good, and we can observe the feedback from user on this timing changing

PR #453, opened 2015-08-12
Add feature to write out job properties in job log when executing.

evlstyle:


evlstyle:
An Api used in the change doesn't belong to Java 1.6, which broke the build, will fix tomorrow.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -43,6 +43,9 @@
   public static final String JOB_PROP_ENV = "JOB_PROP_FILE";
   public static final String JOB_NAME_ENV = "JOB_NAME";
   public static final String JOB_OUTPUT_PROP_FILE = "JOB_OUTPUT_PROP_FILE";
+  private static final String SENSITIVE_JOB_PROP_NAME_SURFFIX = "_X";
```
logiclord: SUFFIX ?

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -81,6 +84,31 @@ protected void resolveProps() {
...
+        for(Map.Entry<String, String> entry : flattenedProps.entrySet()){
+          String key = entry.getKey();
+          String value = key.endsWith(SENSITIVE_JOB_PROP_NAME_SURFFIX)?
+                                      SENSITIVE_JOB_PROP_VALUE_PLACEHOLDER :
+                                      entry.getValue();
+          this.info(String.format("%s=%s",key,value));
+        }
+        this.info("****** End Job properties  ******");
+      } catch (Exception ex){
+        log.error("failed to log job properties ", ex);
```
hluu: Does this go through Azkaban log file or go to job log?

azkaban-common/src/main/java/azkaban/utils/Props.java
```diff
@@ -811,6 +812,30 @@ public void storeFlattened(OutputStream out) throws IOException {
   }
 
   /**
+   * Returns a map of all the flattened properties
```
hluu: Should we mentioned that the properties are sorted?

azkaban-common/src/main/java/azkaban/utils/Props.java
```diff
@@ -811,6 +812,30 @@ public void storeFlattened(OutputStream out) throws IOException {
...
+
+    // creating a treeMap here so that the flattened list can be sorted
+    // alphabetically by the key value.
+    Map<String,String> returnVal = new TreeMap<String,String>();
+
+    // to keep the logic in sync with the rest piece of code,
+    // when there is a conflict, value from the child takes the priority.
+    for (Props curr = this; curr != null; curr = curr.getParent()) {
+      for (String key : curr.localKeySet()) {
+          if (!returnVal.containsKey(key)){
```
hluu: Is it necessary to check?

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
evlstyle: it goes to the job log in case there are no exceptions during the process, if for reason the process throws exception, the exception will go to Azkaban log (as I figured it doesn't make sense to try again to log to Job log when "logging to job log" is throwing exception, when that happens we just log the incident in the azkaban log and move on)

azkaban-common/src/main/java/azkaban/utils/Props.java
```diff
@@ -811,6 +812,30 @@ public void storeFlattened(OutputStream out) throws IOException {
   }
 
   /**
+   * Returns a map of all the flattened properties
+   *
+   * @Return
+   */
+  public Map<String,String> getFlattened(){
```
logiclord: Actually we don't need this method. It gives same function as props.getMapByPrefix("") i.e. prefix with empty string.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -81,6 +84,31 @@ protected void resolveProps() {
     jobProps = PropsUtils.resolveProps(jobProps);
   }
 
+  /**
+   * prints the current Job props to the Job log.
+   */
+  protected void logJobProperties() {
```
logiclord: Should we make it generic to accept props as argument so we can leverage in case we want to extend to say sysProps ?

azkaban-common/src/main/java/azkaban/utils/Props.java
evlstyle: fixed in comment.

azkaban-common/src/main/java/azkaban/utils/Props.java
evlstyle: I think it is, once the iterator moves to the parent (which means the curr is now pointing to the parent) anything that previously has been defined by the child shouldn't be overwritten, the check in place is to make sure the rule is followed. Please let me know if I understand the concern correctly.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
evlstyle: Good catch,  fixed in code.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
evlstyle: thought it as well at the first place, we now depend on a job prop value to determine if the content should be print or not and for other props we may want to have different indicators to make this judge, so I would recommend ship the code as for now and  it is fairly an easy change if we have request to support more in a latter time. plz let me know if  this makes sense.

azkaban-common/src/main/java/azkaban/utils/Props.java
evlstyle: The new one returns a set that is sorted by the key value, which the old one doesn't , plus the old function seems need some reflecting work as the logic seems not trivial therefore I created this new one.  Second thought I think it is a better idea to fix the old one rather than creating a new one . I'll fix that .

azkaban-common/src/main/java/azkaban/utils/Props.java
```diff
@@ -811,20 +812,28 @@ public void storeFlattened(OutputStream out) throws IOException {
...
   /**
+   * Returns a map of all the flattened properties, the item in the returned map is sorted alphabetically
+   * by the key value.
+   *
+   *
+   * @Return
+   */
+  public Map<String,String> getFlattened(){
+    TreeMap<String,String> returnVal = new TreeMap<String,String>(); 
+    returnVal.putAll(getMapByPrefix(""));
```
johnyu0520: instead of using getMapByPrefix(), maybe just use the Prop class's _current variable?

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -81,6 +84,31 @@ protected void resolveProps() {
...
+    if (this.jobProps != null &&
+        this.jobProps.getBoolean(JOB_DUMP_PROPERTIES_IN_LOG, false)){
+      try {
+        Map<String,String> flattenedProps = this.jobProps.getFlattened();
```
johnyu0520: I was just thinking if there's a way to implement this feature (and sort it) without touching the Props.

I think it can be done with:
TreeSet<String> keySet = jobProps.getKeySet();
for( String key : keySet ){
// do the magic
}

A bit less code and a bit less code change?

evlstyle: yep, it is a great idea to pull the keySet here as the implamentation does go recursively and gets all the keys from the hierarchy chain, however the concern for this implementation comes from the code to pull the value from the key, it seems like the props object only exposes  the function get(object) to fetch the value from the key, which furtunately does have the logic to go all the way up the chain when finding the value, however as the way it implements it need to go thru the chain each time to find a value for the key, it will be expensive if a value from the top parent is pulled for many times with this implementation. I agree the keySet route does have the benefit to avoid touching the props class, however I feel it might be nicer to provide this extra method, which caches the flatted key value pairs,  to fulfill the functionality demand of the feature and avoid potential performance draw backs.  Please let me know your idea. (btw, seems like jobProps.getkeySet returns only a HashSet, so it seems like we need to explictly initialize a TreeSet from the return of the getKeySet() call rather than assigning directly, let me know if I got it wrong :)  )

johnyu0520: I think there's merits going both ways, so providing a flatten method is probably also fine

if we are going the treeset route then yes, we'll still need to new a TreeSet...

hluu:
LGTM

Issue #454, opened 2015-08-12
Purge all data for already deleted project


logiclord:
Idea is to add support for hard deleting APIs so that we can clean up outdated projects on a regular basis and save db space.

davidzchen:
Is this issue fixed by the commits listed here?

PR #455, opened 2015-08-12
Purge all data for already deleted project #454

logiclord:
Azkaban already has method cleanOlderProjectVersion in place to hard delete a project version (https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java#L1038), which is used by Azkaban to do house keeping when a new project version is uploaded (https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/ProjectManager.java#L496). I have leveraged this exisitng functionality.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -169,6 +169,67 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
     return project;
   }
 
+    /**
+     * Fetch first project with a given name {@inheritDoc}
+     *
+     * @see azkaban.project.ProjectLoader#fetchProjectByName(java.lang.String)
+     */
+    @Override
+    public Project fetchProjectByName(String name)
+        throws ProjectManagerException {
+        Connection connection = getConnection();
+
+        Project project = null;
+        try {
+            project = fetchProjectByName(connection, name);
+        } finally {
+            DbUtils.closeQuietly(connection);
+        }
+
+        return project;
+    }
+
+    private Project fetchProjectByName(Connection connection, String name)
+        throws ProjectManagerException {
+        QueryRunner runner = new QueryRunner();
+        // Fetch the project
+        Project project = null;
+        ProjectResultHandler handler = new ProjectResultHandler();
+        try {
+            List<Project> projects =
+                runner.query(connection,
+                    ProjectResultHandler.SELECT_PROJECT_BY_NAME, handler, name);
+            if (projects.isEmpty()) {
+                throw new ProjectManagerException(
+                    "No active project with name " + name + " exists in db.");
+            }
+
+            project = projects.get(0);
+        } catch (SQLException e) {
+            logger.error(ProjectResultHandler.SELECT_PROJECT_BY_NAME
+                + " failed.");
+            throw new ProjectManagerException(
+                "Query for existing project failed. Project " + name, e);
+        }
+
+        // Fetch the user permissions
+        List<Triple<String, Boolean, Permission>> permissions =
+            fetchPermissionsForProject(connection, project);
+
+        for (Triple<String, Boolean, Permission> perm : permissions) {
+            if (perm.getThird().toFlags() != 0) {
+                if (perm.getSecond()) {
+                    project
+                        .setGroupPermission(perm.getFirst(), perm.getThird());
+                } else {
+                    project.setUserPermission(perm.getFirst(), perm.getThird());
+                }
+            }
+        }
+
+        return project;
+    }
+
   private List<Triple<String, Boolean, Permission>> fetchPermissionsForProject(
       Connection connection, Project project) throws ProjectManagerException {
     ProjectPermissionsResultHandler permHander =
@@ -1136,6 +1197,9 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
 
   private static class ProjectResultHandler implements
       ResultSetHandler<List<Project>> {
+    private static String SELECT_PROJECT_BY_NAME =
+        "SELECT id, name, active, modified_time, create_time, version, last_modified_by, description, enc_type, settings_blob FROM projects WHERE name=?";
+
     private static String SELECT_PROJECT_BY_ID =
         "SELECT id, name, active, modified_time, create_time, version, last_modified_by, description, enc_type, settings_blob FROM projects WHERE id=?";
 
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -141,7 +141,7 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
           runner.query(connection, ProjectResultHandler.SELECT_PROJECT_BY_ID,
               handler, id);
       if (projects.isEmpty()) {
-        throw new ProjectManagerException("No active project with id " + id
+        throw new ProjectManagerException("No project with id " + id
             + " exists in db.");
       }
 
@@ -201,7 +201,7 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
                     ProjectResultHandler.SELECT_PROJECT_BY_NAME, handler, name);
             if (projects.isEmpty()) {
                 throw new ProjectManagerException(
-                    "No active project with name " + name + " exists in db.");
+                    "No project with name " + name + " exists in db.");
             }
 
             project = projects.get(0);
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -229,8 +229,7 @@ public class ProjectManager {
             try {
                 fetchedProject = projectLoader.fetchProjectByName(name);
             } catch (ProjectManagerException e) {
-                throw new RuntimeException(
-                    "Could not load project from store.", e);
+                logger.error("Could not load project from store.", e);
             }
         }
         return fetchedProject;
@@ -251,8 +250,7 @@ public class ProjectManager {
             try {
                 fetchedProject = projectLoader.fetchProjectById(id);
             } catch (ProjectManagerException e) {
-                throw new RuntimeException(
-                    "Could not load project from store.", e);
+                logger.error("Could not load project from store.", e);
             }
         }
         return fetchedProject;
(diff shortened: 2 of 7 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -204,6 +204,34 @@ public Project getProject(int id) {
     return projectsById.get(id);
   }
 
+    /**
+     * fetch inactive project from db by project_id
+     *
+     * @param name
+     * @return
+     */
+    public Project getInactiveProject(String name) {
```
evlstyle: function name is misleading, actually the function here will get a project with name match disregard the status, so it is not necessary to be Inactive.  Recommend to rename this to GetProject or GetProjectByName, state it in the meta that the returning result disregards the project status.  

Same comment for the function below, which gets the project by Id.

evlstyle: function name is misleading, actually the function here will fetch a project with a successful name match, no matter it is active or not .  Simple fix here is to check the project and make sure it is inactive before returning, therefore the function works as it's name indicates. 
Same comment for the function below, which gets the project by Id.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -169,6 +169,67 @@ private Project fetchProjectById(Connection connection, int id)
...
+        try {
+            project = fetchProjectByName(connection, name);
+        } finally {
+            DbUtils.closeQuietly(connection);
+        }
+
+        return project;
+    }
+
+    private Project fetchProjectByName(Connection connection, String name)
```
hluu: I am surprised there isn't a method that does this already.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -249,6 +277,25 @@ public Project createProject(String projectName, String description,
...
+     * of a project and log event in project_events table
+     *
+     * @param project
+     * @param deleter
+     * @return
+     * @throws ProjectManagerException
+     */
+    public synchronized Project purgeProject(Project project, User deleter)
+        throws ProjectManagerException {
+        projectLoader.cleanOlderProjectVersion(project.getId(),
```
hluu: Should we check for null project and deleter parameters?

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
```diff
@@ -214,6 +213,96 @@ private static void clearDB() {
     DbUtils.closeQuietly(connection);
   }
 
+    /** Test case to validated permissions for fetchProjectByName **/
+    @Test
+    public void testPermissionRetrivalByFetchProjectByName()
```
hluu: Great to see tests

hluu: seems like the method can be more clear in terms of what it is testing.

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
```diff
@@ -214,6 +213,96 @@ private static void clearDB() {
...
+
+        Project fetchedProject = loader.fetchProjectByName(project.getName());
+        assertProjectMemberEquals(project, fetchedProject);
+        Assert.assertEquals(permOverride,
+            fetchedProject.getUserPermission(user.getUserId()));
+    }
+
+    /** Default Test case for fetchProjectByName **/
+    @Test
+    public void testProjectRetrivalByFetchProjectByName()
```
hluu: Retrival?  Do you mean Retrieval?

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
```diff
@@ -214,6 +213,96 @@ private static void clearDB() {
...
+        if (!isTestSetup()) {
+            return;
+        }
+        ProjectLoader loader = createLoader();
+        try {
+            loader.fetchProjectByName("NonExistantProject");
+        } catch (ProjectManagerException ex) {
+            System.out.println("Test true");
+        }
+        Assert.fail("No exception");
```
hluu: Maybe make this message more clear i.e "Expecting exception, but didn't get one"

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -518,6 +520,50 @@ private void handleDownloadProject(HttpServletRequest req,
...
+
+        // project is already deleted
+        if (valid) {
+            project = projectManager.getInactiveProject(projectName);
+            // only eligible users can purge a project
+            if (!hasPermission(project, user, Type.ADMIN)) {
+                ret.put("error", "Cannot purge. User '" + user.getUserId()
+                    + "' is not an ADMIN.");
+                valid = false;
+                ;
```
hluu: extra semicolon

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -518,6 +520,50 @@ private void handleDownloadProject(HttpServletRequest req,
...
+        if (valid) {
+            try {
+                projectManager.purgeProject(project, user);
+            } catch (ProjectManagerException e) {
+                ret.put("error", e.getMessage());
+                valid = false;
+            }
+        }
+
+        ret.put("success", valid);
```
hluu: So the return object can have "success" as key and valid value as false.  Is that OK?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
logiclord: Yes, I named it InActive as there was already a method getProject to fetch active projects from Cache and to minimize code refractoring.
I have merged two methods and added another method to test if a project is Active.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
logiclord: I expect to throw exception in that scenario. For purge API, I have handled that case while invoking this method.

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
logiclord: fixed

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
logiclord: fixed

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
logiclord: changed

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
logiclord: removed

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
logiclord: yes

logiclord:
Incorporated feedback comments and updated the review.

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
johnyu0520: nit picking, retrieval is mis-spelt :)

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
```diff
@@ -214,6 +213,96 @@ private static void clearDB() {
     DbUtils.closeQuietly(connection);
   }
 
+    /** Test case to validated permissions for fetchProjectByName **/
+    @Test
+    public void testPermissionRetrivalByFetchProjectByName()
+        throws ProjectManagerException {
+        if (!isTestSetup()) {
+            return;
```
johnyu0520: just curious, if the test is not setup, should we setting up the test, or should we be returning success with the test?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -518,6 +520,51 @@ private void handleDownloadProject(HttpServletRequest req,
...
+     * validate readiness of a project and user permission and use
+     * projectManager to purge the project if things looks good
+     **/
+    private void handlePurgeProject(HttpServletRequest req,
+        HttpServletResponse resp, Session session) throws ServletException,
+        IOException {
+        User user = session.getUser();
+        String projectName = getParam(req, "project");
+        HashMap<String, Object> ret = new HashMap<String, Object>();
+        boolean valid = true;
```
johnyu0520: Just thinking, should the variable "valid" be renamed to something better?
Does it stand for "shouldPurge", or does it stand for "purgeOperationSuccessful"?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -518,6 +520,51 @@ private void handleDownloadProject(HttpServletRequest req,
...
+
+        try {
+            Project project = null;
+
+            // project is already deleted
+            if (valid) {
+                project = projectManager.getProject(projectName);
+                // only eligible users can purge a project
+                if (project == null
+                    || !hasPermission(project, user, Type.ADMIN)) {
```
johnyu0520: Looks like this if statement contains 2 possible scenarios.  One is project already deleted.  The other is user does not have admin permission.  Should they be sending out different error messages?

azkaban-common/src/test/java/azkaban/project/JdbcProjectLoaderTest.java
logiclord: makes sense. incorporated the feedback

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
logiclord: updated

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
logiclord: updated

logiclord:
incorporated feedback

johnyu0520:
LGTM

PR #458, opened 2015-08-14
increase default char limit for execution log names

samtalasila:
Increasing the default char limit for execution log names to avoid cases where we cannot access the logs.

example:
```Caused by: java.sql.SQLException: Value too long for column "NAME VARCHAR(128) NOT NULL SELECTIVITY 1": "'<redacted>... (136)"; SQL statement:```

@orenmazor


samtalasila:
Hey @orenmazor, not sure if this is the proper venue to do it, but I think something like this is needed.

PR #465, opened 2015-08-21
Added integration with the IntelliJ IDEA IDE.

FelixGV:
This change makes it possible to run  `./gradlew idea` in order to create proper .ipr files for the main project and subprojects in order to import easily into IntelliJ IDEA.


suvodeep-pyne:
This has already been added.

PR #466, opened 2015-08-21
HTML escaping in job logs.

FelixGV:
Previously, only flow logs were escaped.


hluu:
Any information on testing Felix?

FelixGV:
I have not had time to test, unfortunately... Was pulled in too many meetings last Friday ):

onurkaraman:
I think I was hit by this issue yesterday. I had a line that does something like:
`logger.info("abc: <xyz>.")`

The resulting page skipped the `"<xyz>"` chunk and just displayed:
`abc: .`

However, looking at the raw data from the ajax that fills in the job log window shows:
`abc: xyz.`

FelixGV:
Good to know I'm not the only one twitching at that @onurkaraman ;) ...

@hluu, this is does not seem like a high risk change, does it? Do you think we could merge it in?

-F

hluu:
LGTM

PR #467, opened 2015-08-22
Database layer changes to add executors and executor_events

logiclord:


logiclord:
Updated exception messages

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -0,0 +1,104 @@
...
+  private boolean active = true;
+
+  // TODO: ExecutorStats to be added
+
+  public Executor(int id) {
+    this.id = id;
+  }
+
+  public Executor(int id, String host, int port) {
+    this.id = id;
```
hluu: Should we do basic validations on these constructor parameters?
port can't < 0
host can't be null
id can't < 0

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Class to represent an AzkabanExecutorServer details for ExecutorManager
+ *
+ * @author gaggarwa
+ */
+public class Executor {
+  private final int id;
+  private String host;
+  private int port;
+  private boolean active = true;
```
hluu: why default for active to true?

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+package azkaban.executor;
+
+/**
+ * Class to represent an AzkabanExecutorServer details for ExecutorManager
+ *
+ * @author gaggarwa
+ */
+public class Executor {
```
hluu: Does it make sense to make this class immutable?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
       String flowContains, String userNameContains, int status, long startData,
       long endData, int skip, int num) throws ExecutorManagerException;
 
+  /**
+   * Fetch all executors from executors table with active = true
+   *
+   * @return List<Executor>
+   * @throws ExecutorManagerException
```
hluu: Under what conditions this exception is thrown?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+   */
+  public List<Executor> fetchActiveExecutors() throws ExecutorManagerException;
+
+  /**
+   * Fetch executor from executors with a given (host, port)
+   *
+   * @return Executor
+   * @throws ExecutorManagerException
+   */
+  public Executor fetchExecutor(String host, int port)
```
hluu: Does this method return null if it can't find one that matches the give host and port?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+    throws ExecutorManagerException;
+
+  /**
+   * Fetch executor from executors with a given executorId
+   *
+   * @return Executor
+   * @throws ExecutorManagerException
+   */
+  public Executor fetchExecutor(int executorId) throws ExecutorManagerException;
```
hluu: Does this method return null if it can't find one that matches the given executorId?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+  public Executor addExecutor(String host, int port)
+    throws ExecutorManagerException;
+
+  /**
+   * inactivate an executor in executors table
+   *
+   * @param executorId
+   * @throws ExecutorManagerException
+   */
+  public void inactivateExecutor(int executorId)
```
hluu: Does it make sense to combine these two methods into updateExecutor(Executor executor) method?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+  /**
+   * Log an event in executor_event audit table
+   *
+   * @param executor
+   * @param type
+   * @param user
+   * @param message
+   * @return isSuccess
+   */
+  boolean postEvent(Executor executor, EventType type, String user,
```
hluu: Should this throw an exception instead so caller can get more info. on the nature of the failure?  boolean is not very expressive.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+   * @param executor
+   * @param num
+   * @param skip
+   * @return List<ExecutorLogEvent>
+   * @throws ExecutorManagerException
+   */
+  List<ExecutorLogEvent> getExecutorEvents(Executor executor, int num, int skip)
+    throws ExecutorManagerException;
+
   public void addActiveExecutableReference(ExecutionReference ref)
```
hluu: I am a bit confused about what this method does based on its name.  More doc. please

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -105,5 +182,4 @@ public Props fetchExecutionJobOutputProps(int execId, String jobId)
 
   public int removeExecutionLogsByTime(long millis)
```
hluu: exact match? or greater than or less then?  more doc. please.

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @author gaggarwa
+ */
+public class ExecutorLogEvent {
+  /**
+   * Log event type messages. Do not change the numeric representation of each
+   * enum. Only represent from 0 to 255 different codes.
+   */
+  public static enum EventType {
+    ERROR(128), HOST_UPDATE(1), PORT_UPDATE(2), ACTIVATION(3), INACTIVATION(4),
```
hluu: just curious, why 128 for error?

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
```diff
@@ -0,0 +1,99 @@
...
+      case 3:
+        return ACTIVATION;
+      case 4:
+        return INACTIVATION;
+      case 5:
+        return CREATED;
+      case 128:
+        return ERROR;
+      default:
+        return ERROR;
```
hluu: Not sure about default to ERROR.  Makes more sense to thrown an IllegalArgumentException.

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private final int executorId;
+  private final String user;
+  private final Date time;
+  private final EventType type;
+  private final String message;
+
+  public ExecutorLogEvent(int executorId, String user, Date time,
+    EventType type, String message) {
+    this.executorId = executorId;
```
hluu: Should we do any kind of validations?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+  /**
+   * Fetch num events associated with a given executor, starting from skip
+   *
+   * @param executor
+   * @param num
+   * @param skip
+   * @return List<ExecutorLogEvent>
+   * @throws ExecutorManagerException
+   */
+  List<ExecutorLogEvent> getExecutorEvents(Executor executor, int num, int skip)
```
hluu: Does this method return null or empty list when there are not matches?  Please add documentation about this behavior.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -773,6 +776,205 @@ private Connection getConnection() throws ExecutorManagerException {
...
 
+
+  /**
+   *
+   * {@inheritDoc}
+   *
+   * @see azkaban.executor.ExecutorLoader#fetchActiveExecutors()
+   */
+  @Override
+  public List<Executor> fetchActiveExecutors() throws ExecutorManagerException {
```
hluu: Does this method return null or empty list when there are not matches?  Please add documentation about this behavior.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -773,6 +776,205 @@ private Connection getConnection() throws ExecutorManagerException {
...
+
+  private void addExecutorHelper(String host, int port)
+    throws ExecutorManagerException {
+    final String INSERT =
+      "INSERT INTO executors " + "(host, port) values (?,?)";
+    QueryRunner runner = createQueryRunner();
+    String errString = String.format("Error adding %s:%d ", host, port);
+    try {
+      int rows = runner.update(INSERT, host, port);
+      if (rows == 0) {
```
hluu: I wonder under what circumstance that update method returns 0 rows?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -773,6 +776,205 @@ private Connection getConnection() throws ExecutorManagerException {
...
+    QueryRunner runner = createQueryRunner();
+
+    final String INSERT_PROJECT_EVENTS =
+      "INSERT INTO executor_events (executor_id, event_type, event_time, username, message) values (?,?,?,?,?)";
+    Date updateDate = new Date();
+    try {
+      runner.update(INSERT_PROJECT_EVENTS, executor.getId(), type.getNumVal(),
+        updateDate, user, message);
+    } catch (SQLException e) {
+      e.printStackTrace();
```
hluu: better to throw an exception

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -773,6 +776,205 @@ private Connection getConnection() throws ExecutorManagerException {
...
+    QueryRunner runner = createQueryRunner();
+
+    ExecutorLogsResultHandler logHandler = new ExecutorLogsResultHandler();
+    List<ExecutorLogEvent> events = null;
+    try {
+      events =
+        runner.query(ExecutorLogsResultHandler.SELECT_EXECUTOR_EVENTS_ORDER,
+          logHandler, executor.getId(), num, skip);
+    } catch (SQLException e) {
+      logger.error(e);
```
hluu: Looks like this exception is not being thrown up.  Silent killer :)

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -293,6 +334,209 @@ public void testUploadExecutableNode() throws Exception {
 
   }
 
+  /* Test active executors fetch from empty executors */
+  @Test
+  public void testFetchEmptyExecutors() throws Exception {
+    if (!isTestSetup()) {
+      return;
```
hluu: better to throw an exception, otherwise this test will silently pass every time the setup step is not properly done.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -293,6 +334,209 @@ public void testUploadExecutableNode() throws Exception {
...
+
+    for (int index = 0; index < eventLogs.size(); ++index) {
+      ExecutorLogEvent eventLog = eventLogs.get(index);
+      Assert.assertEquals(eventLog.getExecutorId(), executor.getId());
+      Assert.assertEquals(eventLog.getUser(), user.getUserId());
+      Assert.assertEquals(eventLog.getType(), events[index + skip]);
+      Assert.assertEquals(eventLog.getMessage(),
+        message + events[index + skip].getNumVal());
+    }
+    clearDB();
```
hluu: What about performing clearDB() in a general test cleanup method so each test doesn't have to remember to do this?

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
evlstyle: +1  

in C# usually we assign the integers so that the enum supports "|" operation. Say item1(1) item2(2)  and when passing we do passedVal = item1(1)|item2(2)   when checking we do if (  item1  & passedVal  ) == item1 then ....      not sure if it works the same in Java  :)

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
johnyu0520: I think "skip" means "offset"?

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -0,0 +1,104 @@
...
+  private final int id;
+  private String host;
```
johnyu0520: I think host and port should also be final? Given that they are all driven from the DB and should not be modified

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -47,6 +48,82 @@ public ExecutableFlow fetchExecutableFlow(int execId)
...
+  public Executor fetchExecutor(int executorId) throws ExecutorManagerException;
+
+  /**
+   * create an executor and insert in executors table. Fails, if a executor with
+   * (host, port) already exist
+   *
+   * @return Executor
+   * @throws ExecutorManagerException
+   */
+  public Executor addExecutor(String host, int port)
```
johnyu0520: When this fails, does this return null or does this throw an exception?

azkaban-common/src/main/java/azkaban/executor/Executor.java
logiclord: makes sense, I can't remember why I though there will be a use case to have just id and not host:port.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: good catch. forgot to clean.

azkaban-common/src/main/java/azkaban/executor/Executor.java
logiclord: good idea. Added validation

azkaban-common/src/main/java/azkaban/executor/Executor.java
logiclord: we are mostly dealing with active executor and any object created will be for an active executor. But I think we can remove the default value to give more flexibility.

azkaban-common/src/main/java/azkaban/executor/Executor.java
logiclord: No, we need to change active status of the Executor. We will need to keep inactive executor for some time when an admin will mark an executor invalid but we will need to maintain status of the flows previously scheduled on inactivated executor.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: added

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: yes

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: yes

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: Good idea. Feedback incorporated.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: updated. I wanted to avoid failures due to insertion in executor audit table.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: This method is to fetch events recorded in executor audit table, inserted by postExecutorEvents;

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: this method is not part of review, it already exist

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
logiclord: These statuses are inserted in database, so we want to easily identify a status while debugging some issues . Error is assign a higher code to anticipate any new status which may come in between the missing range.

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorLogEvent.java
logiclord: I think we can keep things light and skip the validation here.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: it returns empty list, added in doc

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: changed.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
logiclord: These tests are sort of integration tests and not turned on by default as their need db.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
logiclord: We are doing clearDB() after all the test but we want each test to clean up all the insertions it make so there is no dependencies between the tests which can give incorrect test results.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
logiclord: yes, added a note in doc

logiclord:
Incorporated feedback.

johnyu0520:
lgtm

evlstyle:
LGTM

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
hluu: That was my point, then it is reasonable to declare a method which will be called by junit to call after each test method is invoked.  This provides an opportunity to perform any kind of clean up between each test.

hluu:
LGTM

PR #469, opened 2015-08-25
DSP- 7976/7977 implement dispatcher / interface of statistics calculation module.

evlstyle:


azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+import azkaban.utils.JSONUtils;
+import azkaban.utils.Pair;
+import azkaban.utils.RestfulApiClient;
+
+/** Client class that will be used to handle all Restful API calls between Executor and the host application.
+ * */
+public class ExecutorApiClient extends RestfulApiClient<Map<String, Object>> {
+
+  // hide the constructor. we what this class to be in singleton.
```
logiclord: we want this class to be in singleton. ?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+  private ExecutorApiClient(){}
+
+  // international cache for the object instance.
```
logiclord: internal cache ?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+  private static ExecutorApiClient instance = null;
+
+  /**Singleton creator of the class.
+   * */
+  public static ExecutorApiClient getInstance() {
```
logiclord: nit: A bit verbose, commenting.. just mentioning a singleton class should be fine ?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+  /**function to get executor status .
+   * @param executorHost    Host name of the executor.
+   * @param executorPort    Host port.
+   * @param action          query action.
+   * @param param           extra query parameters
+   * @return  the de-serialized JSON object in Map<String, Object> format.
+   * */
+  @SuppressWarnings("unchecked")
+  public Map<String, Object> callExecutorStats(String executorHost, int executorPort,
```
logiclord: I though API class will not have functionality specific method as we discussed. I think we should have this in executor manager only and executor manage should use rest API exposed from this class ?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+      logger.info("[Differentiator] differentiator chose " + (result?  object1.toString(): object2.toString()));
+      if (result) result1++; else result2++;
+    }
+
+    logger.info(String.format("Result : %s vs %s ",result1,result2));
+    return new Pair<Integer,Integer>(result1,result2);
+  }
+
+  @Override
+  public int compare(T o1, T o2) {
```
logiclord: May be I am missing something, but should ExecutableFlow be available to comparator (ranker) ?
This will enable us to have more signals to chose an Executor A over Executor B in context of Executable flow C ? 
Just to make it more extensible and avoiding interfaces changes later on.

azkaban-common/src/main/java/azkaban/executor/dispatcher/Dispatcher.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+
+/** Definition of the dispatcher interface.
+ *  an implementation of the dispatcher interface provides the functionality
+ *  to return a candidate from the candidateList that suits best for the dispatchingObject.
+ *  @param K : type of the candidate.
+ *  @param V : type of the dispatching object.
+ */
+public interface Dispatcher <K,V> {
```
logiclord: use apache java formatting template to auto-format all files.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: yes as I see no point we need more than one instance of it . As this class need to extend the abstract class therefore we need to create instance of it, otherwise it is just a static helper class which handles the restful API communications.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: I'll make a better wording for it,  it is just the instance that previously created.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: sure I can clean up a little bit.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: Either way works for me ( the talk we had haven't closed to an conclusion) I thought it might be better to keep the logic on how to deal with the Restful API apart from the ExecutorManager.  below are the two reasons - 
a.  as the majority of logic on how to talk with Restful API now resides in this new class I think it doesn't make much sense to separate the logic and leave a portion in separate classes. 
b. as we have already made the decision to abstract out the API communication logic out of the Manager I think the talk to API should be transparent to the Manager (say for a specific request, the Manager should not need to worry about if a POST call instead of a GET call should be made to the RESTful api). 

With that said I prefer to reside all the Restful API logic in this new class and expose only generic methods to the Manager (this callExecutorStats as an example).  Let me know if this makes sense to you, again,  I'm cool either ways.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: In case you missed, please take a look at the dispatcher class and pay attention to the filter logic.  the Flow info will be used there to filter out the "invalid" executors, let me know if that addresses (or partially  addresses :) ) your concern. 

As per our design talk the Flows should be generic and should be treated the same and therefore not counted in the comparison  between the executors (the filter was also not talked during the design and I added it for good).  And per the current design I think it makes sense to keep the flow info apart when doing the comparison between two executors,  at the tome two executors enter the comparison logic, both of them have already passed the filtering logic, which means both are ideally suitable for executing the flow as per the definition,  so the comparator will try to finger out , based on a rule set defined by the system, when comparing executor A to B, which one is better ? and this is how the logic works.      For your concern, I think what you want (let me know if I get it wrong)  is  "for this specific Flow tell me  whether executor A or executor B suits better ?" I think this is different feature ask and was not what we planed, plus I think if we go that route it gives more chance for the user who defined the flow the flexibility to "pick" executors, which may not be what we want .

azkaban-common/src/main/java/azkaban/executor/dispatcher/Dispatcher.java
evlstyle: will do.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+  // hide the constructor. we what this class to be in singleton.
+  private ExecutorApiClient(){}
+
+  // international cache for the object instance.
+  private static ExecutorApiClient instance = null;
+
+  /**Singleton creator of the class.
+   * */
+  public static ExecutorApiClient getInstance() {
+    if (null == instance) {
```
hluu: Since the constructor doesn't do much, another way to initialize an instance of this class w/o doing any synchronization is:

private static ExecutorApiClient instance = new ExecutorApiClient();

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    return instance;
+  }
+
+  /**Implementing the parseResponse function to return de-serialized Json object.
+   * @param response  the returned response from the HttpClient.
+   * */
+  @SuppressWarnings("unchecked")
+  @Override
+  protected Map<String, Object> parseResponse(HttpResponse response)
```
hluu: It would be good to describe the behavior where this method can return either null or empty map.  Would it be simpler to always return empty map?  Should the caller behavior differently based on nulll or empty map?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: Smart.  Will fix .

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
hluu: Seems like this method is getting a map of executor stats.  Should we name it that way?  getExecutorStats(...)

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -0,0 +1,101 @@
...
+   * @param executorHost    Host name of the executor.
+   * @param executorPort    Host port.
+   * @param action          query action.
+   * @param param           extra query parameters
+   * @return  the de-serialized JSON object in Map<String, Object> format.
+   * */
+  @SuppressWarnings("unchecked")
+  public Map<String, Object> callExecutorStats(String executorHost, int executorPort,
+      String action, Pair<String, String>... params) throws IOException {
+
```
hluu: Should we do any validation on method argument?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
hluu: I feel a bit uneasy about the Object part of the Map<String,Object>.  How does the client deal with Object?  Should there be a base class to restrict what can be as the value so caller can have deal it value in a generic fashion w/o the need for casting.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+  /** gets the name of the current implementation of the candidate comparator.
+   * @returns : name of the comparator.
+   * */
+  public abstract String getName();
+
+  /** differentiate method which will kick in when the comparator list generated an equality result for
+   *  both sides. the differentiate method will try best to make sure a stable result is returned.
+   * */
+  protected boolean differentiate(T object1, T object2){
+    if (object2 == null) return true;
```
hluu: It is not clear to me why return true when object2 is null.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /** function to register a factorComparator to the internal Map for future reference.
+   * @param factorComparator : the comparator object to be registered.
+   * */
+  protected void registerFactorComparator(FactorComparator<T> comparator){
+      if (null == comparator ||
+          Integer.MAX_VALUE - this.getTotalWeight() < comparator.getWeight() ) {
+        logger.info(
+            "skipping registerFactorComparator as the comaractor is null or has an invalid weight value.");
```
hluu: Does it make more sense to throw an Exception rather than silently accept the given comparator when it is null or invalid weight?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+  /** function update the weight of a specific registered factorCompartor.
+   * @param factorName : the name of the registered factorComparator to adjust.
+   * @param weight:      the new weight value to be adjusted to.
+   * @return -1 if the factor doesn't exist or the weight value specified is invalid,
+   *          the original value before update otherwise.
+   * */
+  public int adjustFactorWeight(String factorName, int weight){
+    // shortcut if the input is invalid.
+    if (factorName == null ||
+        factorName == "" ||
```
hluu: Better to use factorName.length() == 0.  String comparison in Java is usually done using String.equals(String str) method to compare string content, otherwise == operator will compare object instance references.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+        weight < 0 ||
+        Integer.MAX_VALUE - this.getTotalWeight() < weight){
+      logger.info(
+          "skipping adjustFactorWeight as one or more of the input parameters are invalid");
+      return -1;
```
hluu: should we throw an exception?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    FactorComparator<T> value = this.factorComparatorList.get(factorName);
+
+    // shortcut if the key doesn't exist.
+    if (null == value){
+      logger.info(String.format(
+          "unable to udpate weight as the specified factorName %s doesn't exist",
+          factorName));
+      return -1;
```
hluu: should we throw an exception?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+    int returnVal = value.getWeight();
+    value.updateWeight(weight);
+    return returnVal;
+  }
+
+  /** function returns the total weight of the registered comparators.
+   * @return the value of total weight.
+   * */
+  public int getTotalWeight(){
+    int totalWeight = 0 ;
```
hluu: Just a thought, does it make sense to cache the totalWeight?  I guess we will not have that many FactorComparators, so maybe it is ok to compute the total weight each time this method is called.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+   *  the comparison follows the following logic -
+   *  1. if both objects are equal return 0 score for both.
+   *  2. if one side is null, the other side gets all the score.
+   *  3. if both sides are non-null value, both values will be passed to all the registered FactorComparators
+   *     each factor comparator will generate a result based off it sole logic the weight of the comparator will be
+   *     added to the wining side, if equal, no value will be added to either side.
+   *  4. final result will be returned in a Pair container.
+   *
+   * */
+  public Pair<Integer,Integer> getReult(T object1, T object2){
```
hluu: Method is name is too generic and doesn't explain what this method does.  Better name please.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+    logger.info(String.format("start comparing '%s' with '%s',  total weight = %s ",
+        object1 == null ? "(null)" : object1.toString(),
+        object2 == null ? "(null)" : object2.toString(),
+        this.getTotalWeight()));
+
+    // short cut if object equals.
+    if (object1 ==  object2){
```
hluu: This is comparing object instance references, is this the intention?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+      logger.info(String.format("Result : %s vs 0 (right is null)",this.getTotalWeight()));
+      return new Pair<Integer, Integer>(this.getTotalWeight(),0);
+    }
+
+    // both side is not null,put them thru the full loop
+    int result1 = 0 ;
+    int result2 = 0 ;
+    Collection<FactorComparator<T>> comparatorList = this.factorComparatorList.values();
+    Iterator<FactorComparator<T>> mapItr = comparatorList.iterator();
+    while (mapItr.hasNext()){
```
hluu: Using for loop is more clean and don't have to worry about advancing the iterator.

for (FactorComparator<T> comparator : comparatorList) {
   ....
}

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateFilter.java
```diff
@@ -0,0 +1,81 @@
...
+   * @returns : name of the filter.
+   * */
+  public abstract String getName();
+
+  /** function to register a factorFilter to the internal Map for future reference.
+   * @param factorfilter : the Filter object to be registered.
+   * */
+  protected void registerFactorFilter(FactorFilter<T,V> filter){
+      if (null == filter ) {
+        logger.info("skipping registerFactorFilter as the comaractor is null or has an invalid weight value.");
```
hluu: I think it is better to throw exception

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+      // add or replace the filter.
+      this.factorFilterList.put(filter.getFactorName(),filter);
+      logger.info(String.format("Factor filter added for '%s'.",
+          filter.getFactorName()));
+  }
+
+  /** function to get the filtering result.
+   * */
+  public boolean check(T item, V object){
```
hluu: check what?  please come up with a better name?

Returning true means what? false means what?  Please add documentation to explain the behavior

azkaban-common/src/main/java/azkaban/executor/dispatcher/Dispatcher.java
hluu: It seems like this interface doesn't do any dispatching, is that true?

azkaban-common/src/main/java/azkaban/executor/dispatcher/ExecutorDispatcher.java
```diff
@@ -0,0 +1,113 @@
...
+    return returnVal;
+  }
+
+  /**constructor of the class.
+   * @param filter CandidateFilter object to be used to perform the candidate filtering.
+   * @param comparator CandidateComparator object to be used to find the best suit candidate from the filtered list.
+   * */
+  public ExecutorDispatcher(CandidateFilter<K,V> filter,
+      CandidateComparator<K> comparator){
+    this.filter = filter;
```
hluu: Should we add some validations  here?

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
```diff
@@ -0,0 +1,78 @@
...
+
+package azkaban.executor.dispatcher;
+
+import java.util.Comparator;
+import org.apache.log4j.Logger;
+
+/** wrapper class for a factor comparator .
+ *@param T: the type of the objects to be compared.
+ */
+public final class FactorComparator<T>{
```
hluu: Factor is a fairly generic name, can we have something more descriptive?

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /** private constructor of the class. User will create the instance of the class by calling the static
+   *  method provided below.
+   * @param factorName : the factor name .
+   * @param weight : the weight of the comparator.
+   * @ comparator : function to be provided by user on how the comparison should be made.
+   * */
+  private FactorComparator(String factorName, int weight, Comparator<T> comparator){
+    this.factorName = factorName;
+    this.weight = weight;
```
hluu: Can weight be negative?

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /** static function to generate an instance of the class.
+   *  refer to the constructor for the param definitions.
+   * */
+  public static <T> FactorComparator<T> create(String factorName, int weight, Comparator<T> comparator){
+
+    if (null == factorName || factorName == "" || weight < 0 || null == comparator){
+      logger.error("failed to create instance of FactorComparator, at least one of the input paramters are invalid");
+      return null;
```
hluu: Let's throw an exception

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
```diff
@@ -0,0 +1,78 @@
...
+    this.weight = weight;
+    this.comparator = comparator;
+  }
+
+  /** static function to generate an instance of the class.
+   *  refer to the constructor for the param definitions.
+   * */
+  public static <T> FactorComparator<T> create(String factorName, int weight, Comparator<T> comparator){
+
+    if (null == factorName || factorName == "" || weight < 0 || null == comparator){
```
hluu: Please use factorName.equals("")

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorFilter.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /** static function to generate an instance of the class.
+   *  refer to the constructor for the param definitions.
+   * */
+  public static <T,V> FactorFilter<T,V> create(String factorName, Filter<T,V> filter){
+
+    if (null == factorName || factorName == "" || null == filter){
+      logger.error("failed to create instance of FactorFilter, at least one of the input paramters are invalid");
+      return null;
```
hluu: throw exception

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorFilter.java
```diff
@@ -0,0 +1,77 @@
...
+    this.factorName = factorName;
+    this.filter = filter;
+  }
+
+  /** static function to generate an instance of the class.
+   *  refer to the constructor for the param definitions.
+   * */
+  public static <T,V> FactorFilter<T,V> create(String factorName, Filter<T,V> filter){
+
+    if (null == factorName || factorName == "" || null == filter){
```
hluu: factorName.equals("")

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorFilter.java
```diff
@@ -0,0 +1,77 @@
...
+  // interface of the filter.
+  public interface Filter<K,V>{
+
+    /**check function for the filter.
+     * @param itemToCheck:  object to be checked.
+     * @param sourceObject: object which contains statistics based on which a decision is made whether
+     *                      the object being checked need to be filtered or not.
+     * @return true if the check passed, false if check failed, which means the item need to be filtered.
+     * */
+    public boolean check(K itemToCheck, V sourceObject);
```
hluu: check -> shouldFilter?

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
```diff
@@ -0,0 +1,239 @@
...
+  /** helper function to build a valid URI.
+   *  @param host   host name.
+   *  @param port   host port.
+   *  @param path   extra path after host.
+   *  @param isHttp indicates if whether Http or HTTPS should be used.
+   *  @param params extra query parameters.
+   *  @return the URI built from the inputs.
+   *  @throws IOException
+   * */
+  public static URI BuildUri(String host, int port, String path,
```
hluu: method name should start with lower case letter

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
```diff
@@ -0,0 +1,239 @@
...
+    } catch (URISyntaxException e) {
+      throw new IOException(e);
+    }
+
+    return returningUri;
+  }
+
+  /** helper function to fill  the request with header entries .
+   * */
+  private static HttpMessage completeRequest(HttpMessage request,
```
hluu: completeRequest -> addHeaders?

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
```diff
@@ -0,0 +1,239 @@
...
+      for (NameValuePair pair : headerEntries){
+        request.addHeader(pair.getName(), pair.getValue());
+      }
+    }
+    return request;
+  }
+
+  /** helper function to fill  the request with header entries and posting body .
+   * */
+  private static HttpEntityEnclosingRequestBase completeRequest(HttpEntityEnclosingRequestBase request,
```
hluu: same comment as above

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
```diff
@@ -0,0 +1,635 @@
...
+    }
+    return returnVal;
+  }
+
+  @After
+  public void tearDown() throws Exception {
+  }
+
+  @Test
+  public void testExecutorFilter() throws Exception {
```
hluu: It will be easier to understand and maintain if these tests are broken into self-contained separate test methods.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
```diff
@@ -0,0 +1,635 @@
...
+
+    // expect the first item to be selected, memory wise it is the max.
+    Assert.assertEquals(this.getExecutorByName("Executor11"),nextExecutor);
+
+    // add the priority factor.
+    // expect again the #9 item to be selected.
+    comparator.registerComparerForPriority(6);
+    nextExecutor = Collections.max(this.executorList, comparator);
+    /*
+      2 [main] INFO azkaban.executor.dispatcher.CandidateComparator  - start comparing 'Executor8' with 'Executor2',  total weight = 5
```
hluu: I wonder if it makes sense to include these comments in the test because if they change tomorrow, then they are out of dated.

azkaban-common/src/main/java/azkaban/executor/dispatcher/Dispatcher.java
evlstyle: true,  the interface only selects a candidate to handle the work but does not actually put the task on to it . 

please see a design pattern discussion thread below, in which the dispatcher classes does the similar @ http://www.javaworld.com/article/2075719/design-patterns/dispatcher-eases-workflow-implementation.html

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: sure, actually  this method was taken from the Manager class which was named as such. As per the discussion we will remove the method out from the class for now.

johnyu0520: @hluu  I think this is only getting stats for one particular Executor

Also, I am thinking will this method take the Executor object going forward?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+
+  /** gets the name of the current implementation of the candidate comparator.
+   * @returns : name of the comparator.
+   * */
+  public abstract String getName();
+
+  /** differentiate method which will kick in when the comparator list generated an equality result for
+   *  both sides. the differentiate method will try best to make sure a stable result is returned.
+   * */
+  protected boolean differentiate(T object1, T object2){
```
johnyu0520: I don't know if tie-break (or something similar) is more intuitive?

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateFilter.java
johnyu0520: I know the variable naming is meant to be generic, but in our case, how does ExecutableFlow and Executor map to "item" and "object"? which one maps to which? is it possible to make the naming a bit more specific?

azkaban-common/src/main/java/azkaban/executor/dispatcher/ExecutorDispatcher.java
```diff
@@ -0,0 +1,113 @@
...
+    this.comparator = comparator;
+  }
+
+  @Override
+  public K getNext(List<K> candidateList, V dispatchingObject) {
```
johnyu0520: should this method be called getMax, getBest, getTop, or something like that?
getNext gives me the impression of an iterator, so I would expect two calls to getNext() return 2 different values.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
johnyu0520: +1
also makes the test a bit hard to read

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
```diff
@@ -0,0 +1,635 @@
...
+
+  // test samples.
+  protected ArrayList<MockExecutorObject> executorList = new ArrayList<MockExecutorObject>();
+
+  @Before
+  public void setUp() throws Exception {
+    BasicConfigurator.configure();
+
+    executorList.clear();
+    executorList.add(new MockExecutorObject("Executor1",8080,50.0,2048,5,new Date(), 20, 6400));
```
johnyu0520: Just thinking
Do we always need that much (12) executors to unit test our dispatching features?
Wondering if it's possible to prove certain aspects of our dispatch with less executors

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
logiclord: Yes, I am looking for "for this specific Flow tell me whether executor A or executor B suits better ?" part. I think it makes sense to add this in interface (even if we are not use it), to give flexibility to extend and use other features derived out of ExecutableFlow and Executor combination. 

For your other concern about end user manipulating properties to prefer some executor. If user is able to directly manipulate then there is a flaw in new admin level implementation which derived new features, on the other hand some properties will obviously lead to prefer some executors and that can be desirable in certain situation e.g. we may prefer to run spark jobs on current high memory instance but will also work with other instance due to other issues like memory, In an multi-cluster executors we may allow users to show affinity to certain cluster and help them avoid distcp, etc These types of cases cannot be accommodated by filters and need to have ExecutableFlow or a derived feature vector.

Yes, template solves that thing in this scenario.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: fixed the first comment by returning the empty list, added comment in document as well. 
for the second comment as it is how the de-serializer returns result I would suggest leave as is.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
evlstyle: as we discussed I'm commenting out this method from this class.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: changed to tieBreak :)

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: in situation when comparator is called on two null value.   Good catch that I should call this as well in the shortcut logic for the both null scenario .

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: throwing IllegalArguementException instead.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: throws exception instead.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: hum.... the code actually doesn't cache it and does compute every time the method is called. And the reason not to cache is that we have the function to adjust the weight so it will be a little bit messy when the weight is changed and we need to update the cache.  please let me know if I get it right .

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: renamed to "getComparisonScore"

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: it is. the comparator will only return <0,0> result in the case two passed object are reference equals, as it is reference equals, there is no way to tell which one is "better" than the other, so it doesn't matter which side to return as they are the same thing. If the objects are same in value, they will flow down to the logic and will finally one side will be picked as an winner, and the logic is ideally to yield stable result.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateFilter.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateFilter.java
evlstyle: function name changed to "analyzeTarget"  param names changed to "filteringTarget" and "referencingObject". documentation updated accordingly.

azkaban-common/src/main/java/azkaban/executor/dispatcher/ExecutorDispatcher.java
evlstyle: we allow null value here, please refer to the logic in "public K getNext(List<K> candidateList, V dispatchingObject) "

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
evlstyle: actually I don't allow this intentionally. please let me know if there is a justification that we should allow negative values (negative values will be filtered out in the static creator function. )

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorFilter.java
evlstyle: fixed

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
evlstyle: fixed

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorFilter.java
evlstyle: fixed

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
evlstyle: actually the intention of this function is to "make the request complete", if you noticed there are two function has the same name and does different things , but the goal is the same, to "make the request complete". Also this is a private function,  so I think the name might be ok.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
evlstyle: reason to include the log at the first place is to give an idea how the log will look like and why the specific item is expected to be selected, as long as people got the idea I'm cool to remove the lines :)

azkaban-common/src/main/java/azkaban/executor/dispatcher/ExecutorDispatcher.java
hluu: Good point.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
```diff
@@ -0,0 +1,419 @@
...
+import java.util.Date;
+import org.apache.log4j.BasicConfigurator;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import azkaban.executor.dispatcher.*;
+
+public class DispatcherTest {
```
hluu: Do we have any negative test cases? what about when there is only 1 executor?

evlstyle: added more test cases covering edge and negative.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
evlstyle: Added more test to cover corner and negative test scenarios .

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
evlstyle: for this specific one I think it is fine coz the test just try out some cases against a single API, I will add boundary and negative tests in separate methods.

azkaban-common/src/main/java/azkaban/executor/dispatcher/ExecutorDispatcher.java
evlstyle: renamed to getBest.

johnyu0520:
looks good

logiclord:
LTGM

azkaban-common/src/main/java/azkaban/executor/dispatcher/Dispatcher.java
hluu: When I see the name Dispatcher, I was expecting it to do some kind of dispatching.  This class is more about candidate selection.

azkaban-common/src/main/java/azkaban/executor/dispatcher/CandidateComparator.java
hluu: OK

azkaban-common/src/main/java/azkaban/executor/dispatcher/FactorComparator.java
hluu: if it can't be negative, then should there be some sort of validation?

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
hluu: OK.  However the name "completeRequest" doesn't tell the reader much.  Since it is a private method, then it doesn't matter too much.

azkaban-common/src/test/java/azkaban/executor/DispatcherTest.java
hluu: thank you.

hluu:
Good to go.

PR #471, opened 2015-08-28
change the class names from dispatcher to selector, disable couple of the getters and setters in the selector class.

evlstyle:
mostly name change, should be safe.


logiclord:
LGTM

Issue #475, opened 2015-08-28
Rotate web and exec server log


hluu:
Currently Azkaban doesn't leverage log4j to rotate log files and therefore those files become pretty large which makes it a bit difficult during the debugging process.

davidzchen:
Is this issue fixed by the commits listed here?

Codefor:
seems not

PR #476, opened 2015-08-28
rotate web and exec server log issue #475

hluu:


azkaban-execserver/src/main/resources/log4j.properties
```diff
@@ -1,12 +1,12 @@
-log4j.rootLogger=INFO, Console
+log4j.rootLogger=INFO, ExecServer
 log4j.logger.azkaban.execapp=INFO, ExecServer
 
-log4j.appender.ExecServer=org.apache.log4j.RollingFileAppender
+log4j.appender.ExecServer=org.apache.log4j.DailyRollingFileAppender
```
evlstyle: seems like people have concerns about using  DailyRollingFileAppender and recommend using RollingFileAppender instead ? - 

@ https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/DailyRollingFileAppender.html 

DailyRollingFileAppender extends FileAppender so that the underlying file is rolled over at a user chosen frequency. DailyRollingFileAppender has been observed to exhibit synchronization issues and data loss. The log4j extras companion includes alternatives which should be considered for new deployments and which are discussed in the documentation for org.apache.log4j.rolling.RollingFileAppender.

azkaban-common/src/main/java/azkaban/utils/StringUtils.java
```diff
@@ -88,4 +89,19 @@ public static String join2(Collection<String> list, String delimiter) {
 
     return buffer.toString();
   }
+
+  private static final Pattern BROWSWER_PATTERN = Pattern
+      .compile(".*Chrome.*|.*Firefox.*|.*Safari.*");
```
johnyu0520: Just wondering, are we going to try match these patterns case specific, or non case specific?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -58,6 +60,9 @@
   private static final String SESSION_ID_NAME = "azkaban.browser.session.id";
   private static final int DEFAULT_UPLOAD_DISK_SPOOL_SIZE = 20 * 1024 * 1024;
 
+  private static final Pattern BROWSWER_PATTERN = Pattern
```
johnyu0520: consolidate this browser pattern with the StringUtils.BROWSER_PATTERN?

azkaban-execserver/src/main/resources/log4j.properties
hluu: I believe that is case only when multiple JVMs are sending logs to same file.  In our case, it should be OK.

azkaban-common/src/main/java/azkaban/utils/StringUtils.java
logiclord: Its better to use browser engine instead of browser name https://developer.mozilla.org/en-US/docs/Browser_detection_using_the_user_agent

azkaban-webserver/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
hluu: This line is removed.  Not needed

azkaban-common/src/main/java/azkaban/utils/StringUtils.java
hluu: Good idea.  Done

logiclord:
LGTM

johnyu0520:
Looks good

Issue #439, opened 2015-06-19
Improve scalability through supporting multiple exec servers


hluu:
## Current Challenge
Azkaban consists of two servers - web server and executor server.  The current design is the web server assigns flow executions to a single executor server.  Web server mainly provides the UI and the scheduling features, and executor server is responsible for running the jobs in the flow.  Azkaban is running into a scalability issue due to the increased usage as well as user jobs are taking longer and longer to finish because of our large ever increasing data sets. Given the increased number of jobs the executor server has to execute concurrently, it is running into system memory capacity issue and therefore the server hosting Azkaban web server and executor is busy with performing swapping and the web server response is dramatically slow.

## Proposed Design
At the high level, Azkaban server scalability can be improvied by deploying the web server on its own box and deploying multiple executor servers on separate boxes .  With this enhancement, the job executions load can spread across multiple machines and therefore Azkaban server can support more load and can tolerate server failure.
Since Azkaban web server controls the scheduler and there is only one web server, it can still guarantee that a single job is executed only once, even though there are multiple executor servers.
The implementation of this design should make sure it is still easy and possible to run Azkaban in single-server mode, which is very useful for development and debugging purposes.

## Enhancements to Web Server
Accept a list of one or more executor servers - i.e <server name>:<port>.  This information should be backed by a database table so web server doesn't have to be bounced when this list is modified.
When executing a flow, web server needs to decide which executor to send a job execution request to. The logic for the load balancing algorithm should be pluggable.  We will provide a default implementation out of the box.
The JMX page needs to be enhanced to show which flows are being executed on which executor server as well as their current states.
The web server may need to adjust its internal thread pool based # of running executors.  This way, as we scale down from 5 to 1 executor servers, the web server doesn't overwhelm the one executor with flows to execute.
Need a new property to specify the implementation of the pluggable load balancing logic.
When it is time to figure out which executor to execute a flow, web server will delegate that responsibility to the load balancer.

## Enhancements to Executor Server
Executor server should support various states - INITIALIZING, ACCEPTING, NOT_ACCEPTING, SHUTTING_DOWN.  May need to come up with better name for these states.
INITIALIZING - while server is coming up.  It doesn't accept requests
ACCEPTING - up and running, accepting requests
NOT_ACCEPTING - up and running, but not accepting requests.  Good for draining out the currently running flows as we are doing rolling upgrade.
SHUTTING_DOWN - coming down, not accepting requests. (not sure if we need this)
[...]

PR #477, opened 2015-08-29
modify Executor side to support Multiple executor feature  

evlstyle:
The pull request contains the following features. 

1. RestfulAPIClient 
Abstract the logic from the ExectorMgr that handles the communication to Exectors via RestfulAPI into a separate class . 
Code includes  - 
a. A base class with all the base logic for RESTful API call inplace. 
b. A wrapper class for ExecutorApiClient which parses the result from API call in JSON string. 
c. unit tests. 

2. ExecutorInfo Servlet 
New servlet from the executor side that exposes the ExecutorInfo (memory, cpu, flow capacity, last assigned time etc.)
Code includes  - 
a. implementation of all the getter methods that fetch the KPI value upon request. 
b. unit tests. 
 
3. ExecutorSelector 
Mechanism to select the most suitable executor based from a list of the candidates based on the weighted statistics. 
Code includes  - 
 a.  Framework to register filter and weighted comparator that will be used in candidate selection. 
 b.  Easy to use wrapper class that takes in filter name List and comparator name/wieght KVPs to initialize the selector.  
 c.  Implementation of system level filters for flow capacity and memory. 
 d.  Implementation of  comparators for the supported statistics KPIs . 
 e.  unit tests.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -24,14 +24,14 @@ import azkaban.utils.Utils;
  *
  * @author gaggarwa
  */
-public class Executor {
+public class Executor implements Comparable<Executor> {
   private final int id;
   private final String host;
   private final int port;
   private boolean isActive;
   // cached copy of the latest statistics from  the executor.
-  private Statistics cachedExecutorStats;
-  private Date lastUpdated;
+  private ServerStatistics cachedExecutorStats;
+  private Date lastStatsUpdatedTime;
 
   /**
    * <pre>
@@ -113,13 +113,13 @@ public class Executor {
     return id;
   }
 
-  public Statistics getExecutorStats() {
+  public ServerStatistics getExecutorStats() {
     return this.cachedExecutorStats;
   }
 
-  public void setExecutorStats(Statistics stats) {
+  public void setExecutorStats(ServerStatistics stats) {
     this.cachedExecutorStats = stats;
-    this.lastUpdated = new Date();
+    this.lastStatsUpdatedTime = new Date();
   }
 
   /**
@@ -128,10 +128,15 @@ public class Executor {
    *         specific object is never refreshed.
    * */
   public Date getLastStatsUpdatedTime(){
-    return this.lastUpdated;
+    return this.lastStatsUpdatedTime;
   }
 
   public void setActive(boolean isActive) {
     this.isActive = isActive;
+  }
+
+  @Override
+  public int compareTo(Executor o) {
+    return null == o ? 1 : this.hashCode() - o.hashCode();
   }
 }
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -30,7 +30,7 @@ public class Executor implements Comparable<Executor> {
   private final int port;
   private boolean isActive;
   // cached copy of the latest statistics from  the executor.
-  private ServerStatistics cachedExecutorStats;
+  private ExecutorInfo cachedExecutorStats;
   private Date lastStatsUpdatedTime;
 
   /**
@@ -113,11 +113,11 @@ public class Executor implements Comparable<Executor> {
     return id;
   }
 
-  public ServerStatistics getExecutorStats() {
+  public ExecutorInfo getExecutorStats() {
     return this.cachedExecutorStats;
   }
 
-  public void setExecutorStats(ServerStatistics stats) {
+  public void setExecutorStats(ExecutorInfo stats) {
     this.cachedExecutorStats = stats;
     this.lastStatsUpdatedTime = new Date();
   }
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -16,6 +16,10 @@
 
 package azkaban.executor;
 
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
 import azkaban.utils.Utils;
 
 /**
@@ -28,8 +32,9 @@ public class Executor {
   private final String host;
   private final int port;
   private boolean isActive;
-
-  // TODO: ExecutorStats to be added
+  // cached copy of the latest statistics from  the executor.
+  private Statistics cachedExecutorStats;
+  private Date lastUpdated;
 
   /**
    * <pre>
@@ -88,6 +93,13 @@ public class Executor {
     return true;
   }
 
+  @Override
+  public String toString(){
+    return String.format("%s (id: %s)",
+        null == this.host || this.host.length() == 0 ? "(empty)" : this.host,
+        this.id);
+  }
+
   public String getHost() {
     return host;
   }
@@ -102,6 +114,24 @@ public class Executor {
 
   public int getId() {
     return id;
+  }
+
+  public Statistics getExecutorStats() {
+    return this.cachedExecutorStats;
+  }
+
+  public void setExecutorStats(Statistics stats) {
+    this.cachedExecutorStats = stats;
+    this.lastUpdated = new Date();
+  }
+
+  /**
+   * Gets the timestamp when the stats are last updated.
+   * @return date object represents the timestamp, null if the stats of this
+   *         specific object is never refreshed.
+   * */
+  public Date getLastStatsUpdatedTime(){
+    return this.lastUpdated;
   }
 
   public void setActive(boolean isActive) {
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -17,9 +17,6 @@
 package azkaban.executor;
 
 import java.util.Date;
-import java.util.HashMap;
-import java.util.Map;
-
 import azkaban.utils.Utils;
 
 /**
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2014 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.executor;
+
+import azkaban.utils.Utils;
+
+/**
+ * Class to represent an AzkabanExecutorServer details for ExecutorManager
+ *
+ * @author gaggarwa
+ */
+public class Executor {
+  private final int id;
+  private final String host;
+  private final int port;
+  private boolean isActive;
+
+  // TODO: ExecutorStats to be added
+
+  /**
+   * <pre>
+   * Construct an Executor Object
+   * Note: port should be a within unsigned 2 byte
+   * integer range
+   * </pre>
+   *
+   * @param executor_id
+   * @param executor_host
+   * @param executor_port
+   */
+  public Executor(int id, String host, int port, boolean isActive) {
+    if (!Utils.isValidPort(port)) {
+      throw new IllegalArgumentException(String.format(
+        "Invalid port number %d for host %s, executor_id %d", port, host, id));
+    }
+
+    this.id = id;
+    this.host = host;
+    this.port = port;
+    this.isActive = isActive;
+  }
+
+  @Override
+  public int hashCode() {
+    final int prime = 31;
+    int result = 1;
+    result = prime * result + (isActive ? 1231 : 1237);
+    result = prime * result + ((host == null) ? 0 : host.hashCode());
+    result = prime * result + id;
+    result = prime * result + port;
+    return result;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj)
+      return true;
+    if (obj == null)
+      return false;
+    if (!(obj instanceof Executor))
+      return false;
+    Executor other = (Executor) obj;
+    if (isActive != other.isActive)
+      return false;
+    if (host == null) {
+      if (other.host != null)
+        return false;
+    } else if (!host.equals(other.host))
+      return false;
+    if (id != other.id)
+      return false;
+    if (port != other.port)
+      return false;
+    return true;
+  }
+
+  public String getHost() {
+    return host;
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  public boolean isActive() {
+    return isActive;
+  }
+
+  public int getId() {
+    return id;
+  }
+
+  public void setActive(boolean isActive) {
+    this.isActive = isActive;
+  }
+}
--- a/azkaban-common/src/main/java/azkaban/executor/Executor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Executor.java
@@ -113,19 +113,19 @@ public class Executor implements Comparable<Executor> {
     return id;
   }
 
-  public ExecutorInfo getExecutorStats() {
+  public ExecutorInfo getExecutorInfo() {
     return this.cachedExecutorStats;
   }
 
-  public void setExecutorStats(ExecutorInfo stats) {
-    this.cachedExecutorStats = stats;
+  public void setExecutorInfo(ExecutorInfo info) {
+    this.cachedExecutorStats = info;
     this.lastStatsUpdatedTime = new Date();
   }
 
   /**
-   * Gets the timestamp when the stats are last updated.
-   * @return date object represents the timestamp, null if the stats of this
-   *         specific object is never refreshed.
+   * Gets the timestamp when the executor info is last updated.
+   * @return date object represents the timestamp, null if the executor info of this
+   *         specific executor is never refreshed.
    * */
   public Date getLastStatsUpdatedTime(){
     return this.lastStatsUpdatedTime;
--- a/azkaban-common/src/main/java/azkaban/executor/ServerStatistics.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ServerStatistics.java
@@ -19,15 +19,24 @@ package azkaban.executor;
 import java.util.Date;
 import java.util.Map;
 
-  public class Statistics {
+
+ /** Class that exposes the statistics from the executor server.
+  *  List of the statistics -
+  *  remainingMemoryPercent;
+  *  remainingMemory;
+  *  remainingFlowCapacity;
+  *  numberOfAssignedFlows;
+  *  lastDispatchedTime;
+  *  cpuUsage;
+  *
+  * */
+  public class ServerStatistics {
     private double remainingMemoryPercent;
     private long   remainingMemory;
     private int    remainingFlowCapacity;
     private int    numberOfAssignedFlows;
     private Date   lastDispatchedTime;
-    private long   remainingStorage;
     private double cpuUsage;
-    private int    priority;
 
     public double getCpuUsage() {
       return this.cpuUsage;
@@ -69,22 +78,6 @@ import java.util.Map;
       this.lastDispatchedTime = value;
     }
 
-    public long getRemainingStorage() {
-      return this.remainingStorage;
-    }
-
-    public void setRemainingStorage(long value){
-      this.remainingStorage = value;
-    }
-
-    public int getPriority () {
-      return this.priority;
-    }
-
-    public void setPriority (int value) {
-      this.priority = value;
-    }
-
     public int getNumberOfAssignedFlows () {
       return this.numberOfAssignedFlows;
     }
@@ -93,22 +86,18 @@ import java.util.Map;
       this.numberOfAssignedFlows = value;
     }
 
-    public Statistics(){}
+    public ServerStatistics(){}
 
-    public Statistics (double remainingMemoryPercent,
+    public ServerStatistics (double remainingMemoryPercent,
         long remainingMemory,
         int remainingFlowCapacity,
         Date lastDispatched,
-        long remainingStorage,
         double cpuUsage,
-        int  priority,
         int numberOfAssignedFlows){
       this.remainingMemory = remainingMemory;
       this.cpuUsage = cpuUsage;
       this.remainingFlowCapacity = remainingFlowCapacity;
-      this.priority = priority;
       this.remainingMemoryPercent = remainingMemoryPercent;
-      this.remainingStorage = remainingStorage;
       this.lastDispatchedTime = lastDispatched;
       this.numberOfAssignedFlows = numberOfAssignedFlows;
     }
@@ -116,17 +105,15 @@ import java.util.Map;
     @Override
     public boolean equals(Object obj)
     {
-        if (obj instanceof Statistics)
+        if (obj instanceof ServerStatistics)
         {
           boolean result = true;
-          Statistics stat = (Statistics) obj;
+          ServerStatistics stat = (ServerStatistics) obj;
 
           result &=this.remainingMemory == stat.remainingMemory;
           result &=this.cpuUsage == stat.cpuUsage;
           result &=this.remainingFlowCapacity == stat.remainingFlowCapacity;
-          result &=this.priority == stat.priority;
           result &=this.remainingMemoryPercent == stat.remainingMemoryPercent;
-          result &=this.remainingStorage == stat.remainingStorage;
           result &=this.numberOfAssignedFlows == stat.numberOfAssignedFlows;
           result &= null == this.lastDispatchedTime ? stat.lastDispatchedTime == null :
                             this.lastDispatchedTime.equals(stat.lastDispatchedTime);
@@ -138,9 +125,9 @@ import java.util.Map;
 
     // really ugly to have it home-made here for object-binding as base on the
     // current code base there is no any better ways to do that.
-    public static Statistics fromJsonObject(Map<String,Object> mapObj){
+    public static ServerStatistics fromJsonObject(Map<String,Object> mapObj){
       if (null == mapObj) return null ;
-      Statistics stats = new Statistics ();
+      ServerStatistics stats = new ServerStatistics ();
 
       final String remainingMemory = "remainingMemory";
       if (mapObj.containsKey(remainingMemory) && null != mapObj.get(remainingMemory)){
@@ -157,11 +144,6 @@ import java.util.Map;
         stats.setRemainingFlowCapacity(Integer.parseInt(mapObj.get(remainingFlowCapacity).toString()));
       }
 
-      final String priority = "priority";
-      if (mapObj.containsKey(priority) && null != mapObj.get(priority)){
-        stats.setPriority(Integer.parseInt(mapObj.get(priority).toString()));
-      }
-
       final String numberOfAssignedFlows = "numberOfAssignedFlows";
       if (mapObj.containsKey(numberOfAssignedFlows) && null != mapObj.get(numberOfAssignedFlows)){
         stats.setNumberOfAssignedFlows(Integer.parseInt(mapObj.get(numberOfAssignedFlows).toString()));
@@ -170,11 +152,6 @@ import java.util.Map;
       final String remainingMemoryPercent = "remainingMemoryPercent";
       if (mapObj.containsKey(remainingMemoryPercent) && null != mapObj.get(remainingMemoryPercent)){
         stats.setRemainingMemoryPercent(Double.parseDouble(mapObj.get(remainingMemoryPercent).toString()));
-      }
-
-      final String remainingStorage = "remainingStorage";
-      if (mapObj.containsKey(remainingStorage) && null != mapObj.get(remainingStorage)){
-        stats.setRemainingStorage(Long.parseLong(mapObj.get(remainingStorage).toString()));
       }
 
       final String lastDispatched = "lastDispatchedTime";
--- a/azkaban-common/src/main/java/azkaban/executor/Statistics.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Statistics.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.executor;
+
+import java.util.Date;
+
+  public class Statistics {
+    private double remainingMemoryPercent;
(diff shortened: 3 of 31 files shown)


azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,99 @@
...
+
+package azkaban.executor;
+
+import java.util.Date;
+
+  public class Statistics {
+    private double remainingMemoryPercent;
+    private long   remainingMemory;
+    private int    remainingFlowCapacity;
+    private Date   lastDispatched;
```
johnyu0520: I don't have a very strong preference between using Date or long (millis since epoch) to represent time

However, if we are using Date before java 8, I think it's better to use JodaTime.  I think it is a lot more intuitive than the java built in Date functionality.

azkaban-common/src/main/java/azkaban/executor/selector/CandidateSelector.java
```diff
@@ -74,6 +74,10 @@ public K getBest(List<K> candidateList, V dispatchingObject) {
        return null;
```
johnyu0520: nit, typo on line 73

azkaban-common/src/main/java/azkaban/executor/selector/CandidateSelector.java
```diff
@@ -74,6 +74,10 @@ public K getBest(List<K> candidateList, V dispatchingObject) {
...
      }
 
+     if (null == comparator){
+       logger.info("candidate comparator is not specified, default comparator from 'Collections' class will be used.");
```
johnyu0520: wondering what does this mean:
"default comparator from Collections class"

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+  @Override
+  public String getName() {
+    return "ExecutorComparator";
+  }
+
+  private interface ComparatorCreator{
+    FactorComparator<Executor> create(int weight);
+  }
+
+  /**
```
johnyu0520: using <pre> </pre> in java doc will help with preserving formatting

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+   * helper function that does the object  on two statistics, comparator can leverage this function to provide
+   * shortcuts if   the statistics object is missing from one or both sides of the executors.
+   * @param stat1   the first statistics  object to be checked .
+   * @param stat2   the second statistics object to be checked.
+   * @param caller  the name of the calling function, for logging purpose.
+   * @param result  result Integer to pass out the result in case the statistics are not both valid.
+   * @return true if the passed statistics are NOT both valid, a shortcut can be made (caller can consume the result),
+   *         false otherwise.
+   * */
+  private static boolean statisticsObjectCheck(Statistics statisticsObj1, Statistics statisticsObj2, String caller, Integer result){
```
johnyu0520: I think the comment is not up to date?

also given that the return type is boolean, I am assuming the method name would something like isXXX, or hasYYY

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -28,8 +29,9 @@
   private final String host;
   private final int port;
   private boolean isActive;
-
-  // TODO: ExecutorStats to be added
+  // cached copy of the latest statistics from  the executor.
```
hluu: I wonder if it makes more sense to call this class ExecutorInfo.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,169 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.executor;
+
+import java.util.Date;
+import java.util.Map;
+
+  public class Statistics {
```
hluu: What kind of statistics does this class contain?  Maybe add that to the class name.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+    // register the creator for memory comparator.
+    comparatorCreatorRepository.put(MEMORY_COMPARATOR_NAME, new ComparatorCreator(){
+      @Override public FactorComparator<Executor> create(int weight) { return getMemoryComparator(weight); }});
+
+    // register the creator for priority comparator.
+    comparatorCreatorRepository.put(PRIORITY_COMPARATOR_NAME, new ComparatorCreator(){
+      @Override public FactorComparator<Executor> create(int weight) { return getPriorityComparator(weight); }});
+
+    // register the creator for remaining TMP size comparator.
+    comparatorCreatorRepository.put(PRIORITY_COMPARATOR_NAME, new ComparatorCreator(){
```
hluu: Should this have a different name than PRIORITY_COMPARATOR_NAME?

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+
+  /**
+   * constructor of the ExecutorComparator.
+   * @param comparatorList   the list of comparator, plus its weight information to be registered,
+   *  the parameter must be a not-empty and valid list object.
+   * */
+  public ExecutorComparator(Map<String,Integer> comparatorList) {
+    if (null == comparatorList|| comparatorList.size() == 0){
+      logger.error("failed to initialize executor comparator as the passed comparator list is invalid or empty.");
+      throw new IllegalArgumentException("filterList");
```
hluu: I think you can move the message into the IllegalArgumentException exception.  The log statement is optional.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+    // register the comparators, we will now throw here if the weight is invalid, it is handled in the super.
+    for (Entry<String,Integer> entry : comparatorList.entrySet()){
+      if (comparatorCreatorRepository.containsKey(entry.getKey())){
+        this.registerFactorComparator(comparatorCreatorRepository.
+            get(entry.getKey()).
+            create(entry.getValue()));
+      } else {
+        logger.error(String.format("failed to initialize executor comparator as the comparator implementation for requested factor '%s' doesn't exist.",
+            entry.getKey()));
+        throw new IllegalArgumentException("comparatorList");
```
hluu: same comment as previous one regarding exception message

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    
+    // really ugly to have it home-made here for object-binding as base on the
+    // current code base there is no any better ways to do that.
+    public static Statistics fromJsonObject(Map<String,Object> mapObj){
+      if (null == mapObj) return null ;
+      Statistics stats = new Statistics ();
+
+      final String remainingMemory = "remainingMemory";
```
hluu: Should these strings be constants at class member level?

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
hluu: boolean makes more sense.

Method name doesn't quite match what it does.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+      @Override
+      public int compare(Executor o1, Executor o2) {
+        Statistics stat1 = o1.getExecutorStats();
+        Statistics stat2 = o2.getExecutorStats();
+
+        Integer result = 0;
+        if (statisticsObjectCheck(stat1,stat2,REMAININGFLOWSIZE_COMPARATOR_NAME,result)){
+          return result;
+        }
+        return ((Integer)stat1.getRemainingFlowCapacity()).compareTo(stat2.getRemainingFlowCapacity());
```
hluu: is the casting necessary?

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+   * */
+  private static FactorComparator<Executor> getLstDispatchedTimeComparator(int weight){
+    return FactorComparator.create(LSTDISPATCHED_COMPARATOR_NAME, weight, new Comparator<Executor>(){
+
+      @Override
+      public int compare(Executor o1, Executor o2) {
+        Statistics stat1 = o1.getExecutorStats();
+        Statistics stat2 = o2.getExecutorStats();
+
+        Integer result = 0;
```
hluu: this can just be int to avoid auto-boxing.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,307 @@
...
+   * */
+  private static FactorComparator<Executor> getMemoryComparator(int weight){
+    return FactorComparator.create(MEMORY_COMPARATOR_NAME, weight, new Comparator<Executor>(){
+
+      @Override
+      public int compare(Executor o1, Executor o2) {
+       Statistics stat1 = o1.getExecutorStats();
+       Statistics stat2 = o2.getExecutorStats();
+
+       Integer result = 0;
```
hluu: same here

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorFilter.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import azkaban.executor.ExecutableFlow;
+import azkaban.executor.Executor;
+import azkaban.executor.Statistics;
+
+
+public final class ExecutorFilter extends CandidateFilter<Executor, ExecutableFlow> {
```
hluu: Javadoc please

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorSelector.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Contractor of the class.
+   * @param filterList      name list of the filters to be registered,
+   *                        filter feature will be disabled if a null value is passed.
+   * @param comparatorList  name/weight pair list of the comparators to be registered ,
+   *                        again comparator feature is disabled if a null value is passed.
+   * */
+  public ExecutorSelector(List<String> filterList, Map<String,Integer> comparatorList) {
+    super(null == filterList || filterList.size() == 0 ?         null : new ExecutorFilter(filterList),
```
hluu: small thing - filterList.isEmpty() is shorter

azkaban-common/src/main/java/azkaban/executor/selector/FactorComparator.java
```diff
@@ -39,6 +39,7 @@ private FactorComparator(String factorName, int weight, Comparator<T> comparator
     this.factorName = factorName;
     this.weight = weight;
     this.comparator = comparator;
+    logger.info("comparator created for " + this.factorName);
```
hluu: is this necessary?

azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
```diff
@@ -602,4 +605,101 @@ public void testSelectorListWithNullValue() throws Exception {
...
+  public void testCreatingExectorfilterObject() throws Exception{
+    List<String> validList = new ArrayList<String>(ExecutorFilter.getAvailableFilterNames());
+    try {
+      new ExecutorFilter(validList);
+    }catch (Exception ex){
+      Assert.fail("creating ExecutorFilter with valid list throws exception . ex -" + ex.getMessage());
+    }
+  }
+
+  @Test
```
hluu: Looks like the junit Expected exception is useful here - http://www.mkyong.com/unittest/junit-4-tutorial-2-expected-exception-test/

azkaban-execserver/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -142,6 +143,9 @@
 
   private Object executionDirDeletionSync = new Object();
 
+  // date time of the the last flow submitted.
+  private Date lastFlowSubmitted = null;
```
hluu: lastFlowSubmittedDate?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,284 @@
...
+   * Handle all get request to Statistics Servlet {@inheritDoc}
+   *
+   * @see javax.servlet.http.HttpServlet#doGet(javax.servlet.http.HttpServletRequest,
+   *      javax.servlet.http.HttpServletResponse)
+   */
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+      throws ServletException, IOException {
+
+    if (null == lastRefreshedTime ||
+        new Date().getTime() - lastRefreshedTime.getTime() > cacheTimeInMilliseconds){
```
hluu: Can we use System.currentTimeMillis() to achieve the same goal but w/o creating new object?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,284 @@
...
+            totalMemory = totalMemory / sampleCount;
+            logger.info(String.format("total memory - %s , free memory - %s", totalMemory, freeMemory));
+            stats.setRemainingMemory(freeMemory);
+            stats.setRemainingMemoryPercent(totalMemory == 0? 0 :
+              ((double)freeMemory/(double)totalMemory));
+          }
+        }
+      }
+      catch (Exception ex){
+        logger.error("failed fetch system memory info " +
```
hluu: Please also log out the ex

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  /**
+   * call the data providers to fill the returning data container for statistics data.
+   * This function refreshes the static cached copy of data in case if necessary.
+   * */
+  protected synchronized void populateStatistics(){
+    //check again before starting the work.
+    if (null == lastRefreshedTime ||
+        new Date().getTime() - lastRefreshedTime.getTime() > cacheTimeInMilliseconds){
```
hluu: System.currentTimeMillis()?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,284 @@
...
+      final Statistics stats = new Statistics();
+
+      List<Thread> workerPool = new ArrayList<Thread>();
```
hluu: It seems like this is more complicated than it needs to be.  How long does each one of these methods take to complete?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,284 @@
...
+          // set the value.
+          if (sampleCount > 0){
+            double finalResult = (us + sy + wi)/sampleCount;
+            logger.info("Cpu usage result  - " + finalResult );
+            stats.setCpuUpsage(finalResult);
+          }
+        }
+      }
+      catch (Exception ex){
+        logger.error("failed fetch system memory info " +
```
hluu: log out exception please

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -28,8 +29,9 @@
...
+  private Statistics cachedExecutorStats;
+  private Date lastUpdated;
```
logiclord: I think will be better named as lastStatsUpdatedTime ?

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.Date;
+import java.util.Map;
+
+  public class Statistics {
+    private double remainingMemoryPercent;
+    private long   remainingMemory;
+    private int    remainingFlowCapacity;
+    private int    numberOfAssignedFlows;
+    private Date   lastDispatchedTime;
+    private long   remainingStorage;
```
logiclord: is it /tmp disk space ?

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,186 @@
...
+
+    public void setRemainingStorage(long value){
+      this.remainingStorage = value;
+    }
+
+    public int getPriority () {
+      return this.priority;
+    }
+
+    public void setPriority (int value) {
```
logiclord: I thought we removed priority as per yesterday's conversation ?

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,186 @@
...
+
+  public class Statistics {
+    private double remainingMemoryPercent;
+    private long   remainingMemory;
+    private int    remainingFlowCapacity;
+    private int    numberOfAssignedFlows;
+    private Date   lastDispatchedTime;
+    private long   remainingStorage;
+    private double cpuUsage;
+    private int    priority;
```
logiclord: percentage of flow capacity used for comparator as per yesterdays conversation ?

azkaban-common/src/main/java/azkaban/executor/Statistics.java
```diff
@@ -0,0 +1,186 @@
...
+                            this.lastDispatchedTime.equals(stat.lastDispatchedTime);
+          return result;
+        }
+        return false;
+    }
+
+
+    // really ugly to have it home-made here for object-binding as base on the
+    // current code base there is no any better ways to do that.
+    public static Statistics fromJsonObject(Map<String,Object> mapObj){
```
logiclord: Try Jackson ObjectMapper/ObjectWriter based utils from https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/utils/JSONUtils.java#L49 ?

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
```diff
@@ -0,0 +1,286 @@
...
+      public FactorComparator<Executor> create(int weight) { return getCpuUsageComparator(weight); }});
+  }
+
+
+  /**
+   * constructor of the ExecutorComparator.
+   * @param comparatorList   the list of comparator, plus its weight information to be registered,
+   *  the parameter must be a not-empty and valid list object.
+   * */
+  public ExecutorComparator(Map<String,Integer> comparatorList) {
```
logiclord: Should weights be double ?

azkaban-execserver/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -131,6 +131,7 @@ public AzkabanExecutorServer(Props props) throws Exception {
     root.addServlet(new ServletHolder(new ExecutorServlet()), "/executor");
     root.addServlet(new ServletHolder(new JMXHttpServlet()), "/jmx");
     root.addServlet(new ServletHolder(new StatsServlet()), "/stats");
+    root.addServlet(new ServletHolder(new StatisticsServlet()), "/stastics");
```
logiclord: /DispatcherStatistics ? /statistics is a bit confusing with /stats

azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
logiclord: we may need to upgrade junit version, I tried it some time back but there were some issues

azkaban-common/src/main/java/azkaban/executor/Executor.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: now I expose both the remainingFlowCapacity and the numberOfAssignedFlows which represents the percentage of flow capacity I think we are fine here,  I've already modified the comparator to use the numberOfAssignedFlows to judge the weight.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: yes I haven't got a chance to clean it up and will do today.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,287 @@
...
+   * NOTE:
+   * a double value will be used to present the remaining memory,
+   *         a returning value of '55.6' means 55.6%
+   */
+  protected void fillRemainingMemoryPercent(Statistics stats){
+    if (new File("/bin/bash").exists() &&  new File("/usr/bin/free").exists()) {
+      java.lang.ProcessBuilder processBuilder =
+          new java.lang.ProcessBuilder("/bin/bash", "-c", "/usr/bin/free -m -s 0.1 -c 5 | grep Mem:");
+      try {
+        ArrayList<String> output = new ArrayList<String>();
```
logiclord: List = ArrayList ?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,287 @@
...
+  /**
+   * fill the result set with the percent of the remaining system memory on the server.
+   * @param stats reference to the result container which contains all the results, this specific method
+   *              will only work work on the property "remainingMemory" and "remainingMemoryPercent".
+   *
+   * NOTE:
+   * a double value will be used to present the remaining memory,
+   *         a returning value of '55.6' means 55.6%
+   */
+  protected void fillRemainingMemoryPercent(Statistics stats){
```
logiclord: Should we move code which really determine  totalMemory and FreeMemory as it can be used in other places ? In this method we can can that new util method and populate executor stats.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  /**
+   * call the data providers to fill the returning data container for statistics data.
+   * This function refreshes the static cached copy of data in case if necessary.
+   * */
+  protected synchronized void populateStatistics(boolean noCache){
+    //check again before starting the work.
+    if (noCache || null == lastRefreshedTime ||
+        new Date().getTime() - lastRefreshedTime.getTime() > cacheTimeInMilliseconds){
```
logiclord: new Date().getTime() =>  System.currentTimeMillis() ?

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
```diff
@@ -0,0 +1,287 @@
...
+      for (Thread thread : workerPool){thread.start();}
+
+      // wait for all the threads to finish their work.
+      // NOTE: the result container itself is not thread safe, we are good as for now no
+      //       working thread will modify the same property, nor have any of them
+      //       need to compute values based on value(s) of other properties.
+      for (Thread thread : workerPool){
+        try {
+          // we gave maxim 5 seconds to let the thread finish work.
+          thread.join(5000);;
```
logiclord: We should guard this populateStatistics method to make sure that we don't make 2 attempts in parallel e.g. suppose we receive an http request to fetch statistics and that created 5 threads and wait for 5 sec. Within that 5 sec we receive another http call (very likely as from webserver side we have a refresh interval of 1 sec), for this second call we don't want to create 5 new threads to fetch stats ?

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
evlstyle: total weight now doesn't have a cap (say you can have 100 or 1000 as weight) so I figured it doesn't make much sense to use double, int will be good enough.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/CandidateSelector.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/CandidateSelector.java
evlstyle: actually there is a  bug here, when the comparator passed over to the Collections.max function is null and the type of the object to select from doesn't implement Comparable the function will throw exception instead of using any default comparators to get the job done, very nice finding,  fixed by restricting the object type to extend from Comparable.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
evlstyle: fixed

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorComparator.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorFilter.java
evlstyle: added..

azkaban-common/src/main/java/azkaban/executor/selector/ExecutorSelector.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/selector/FactorComparator.java
evlstyle: removed, as it will be logged when registered.

azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
evlstyle: then I'll take the safe route :) plus seems like the Expected Exception will not work here coz here the case is that I don't expect exception.

azkaban-execserver/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
evlstyle: changed to serverstatistics

azkaban-execserver/src/main/java/azkaban/execapp/FlowRunnerManager.java
evlstyle: fixed.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
evlstyle: fixed.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
evlstyle: I would recommend to follow with the design doc we have in hand, coz modify this will trigger a chained modification of quite a lot of places (comparator etc. )

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
evlstyle: fixed.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
evlstyle: function is synchronized so it is guarded, and please see the logic in "doGet", the whole mechanism  is implemented in the classic "double checked locking" manner to ensure thread safety.

azkaban-execserver/src/main/java/azkaban/execapp/StatisticsServlet.java
evlstyle: fixed.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: actually the strings are only used within the method so I think it may not create a lot benefit to declare it at class level.

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: this is removed as we are not planning to populate during this release.

azkaban-common/src/main/java/azkaban/executor/Executor.java
evlstyle: renamed to serverStatistics to align with the name of the RESTful API . Please let me know if  this looks better.

johnyu0520: I was actually thinking of having the Statistics class below named ExecutorInfo or ExecutorStats..

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: fixed, file name changed , please see the new code in ServerStatistics.java

azkaban-common/src/main/java/azkaban/executor/Statistics.java
evlstyle: fixed by using the Mapper directly rather than the JsonUtils,  also fixed the RestfulAPiClient to adopt the change.

azkaban-common/src/main/java/azkaban/executor/ServerStatistics.java
```diff
@@ -0,0 +1,163 @@
...
+      this.remainingFlowCapacity = remainingFlowCapacity;
+      this.remainingMemoryPercent = remainingMemoryPercent;
+      this.lastDispatchedTime = lastDispatched;
+      this.numberOfAssignedFlows = numberOfAssignedFlows;
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (obj instanceof ServerStatistics)
```
hluu: What if obj is null?

azkaban-common/src/main/java/azkaban/executor/Executor.java
evlstyle: object renamed to ExecutorInfo

azkaban-common/src/main/java/azkaban/executor/ServerStatistics.java
evlstyle: if obj is null instanceof will return false, which is expected and we will return false here for the equals check. 

Please refer to http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.20.2 for detauls,

hluu:
LGTM

johnyu0520:
LGTM

davidzchen:
Tracking bug: #439

PR #478, opened 2015-09-01
AzkabanWebServer Modifications to support MultiExecutors

logiclord:
* Added QueueProcessorThread to do the real dispatching and refreshing Executor stats
  * JMX attributes added to /jmx servlet
  * Pause and resume features
  * Refreshing and basic queue processing to execute flows
* Changes in executor manager to correct all existing methods and new methods to deal with deal with multiple executors
* New methods in JDBCExecutorLoader to support the notion of queued flows.
* Added functionality to make queued flows recoverable, now we can bounce webserver without loosing any flow queued at webserver side.
* Modified Stats servlet and coresponding APIs to add executor option
* Display executor_id on executions page
* Cancel a flow which is not yet dispatched to any executor
* DB modification scripts
* Unit tests

State representation in DB:
Queued flow: This will have an entry in execution_flows table with NULL value for executor_id (i.e. not yet assigned) and will have have an entry in active_executing_flows
Running flow: This will have an entry in execution_flows table with the responsible executor its exxecuting and will have have also have an entry in active_executing_flows


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -58,38 +60,53 @@
...
+    new ConcurrentHashMap<Integer, Pair<ExecutionReference, ExecutableFlow>>();
+  final private ConcurrentHashMap<Integer, ExecutableFlow> recentlyFinished =
+    new ConcurrentHashMap<Integer, ExecutableFlow>();
 
-  private ExecutingManagerUpdaterThread executingManager;
+  /* map to easily access queued flows */
+  final private ConcurrentHashMap<Integer, Pair<ExecutionReference, ExecutableFlow>> queuedFlowMap =
+    new ConcurrentHashMap<Integer, Pair<ExecutionReference, ExecutableFlow>>();
+  /* web server side queue */
+  final private ConcurrentLinkedQueue<Pair<ExecutionReference, ExecutableFlow>> queuedFlows =
```
hluu: maybe just need better naming to distinguish this member variable and the one above it.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -58,38 +60,53 @@
...
+    new ConcurrentLinkedQueue<Pair<ExecutionReference, ExecutableFlow>>();
+
+  final private Set<Executor> activeExecutors = new HashSet<Executor>();
+
+  final private ExecutingManagerUpdaterThread executingManager;
+  final private QueueProcessorThread queueProcessor;
 
   private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
```
hluu: please add a comment about the duration so readers wouldn't have to do the math.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -98,12 +115,85 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+      } else if (!executor.isActive()) {
+        executor.setActive(true);
+        executorLoader.updateExecutor(executor);
+      }
+      activeExecutors.add(new Executor(executor.getId(), executorHost,
+        executorPort, true));
+    }
+
+    if (azkProps.getBoolean(AZKABAN_USE_MULTIPLE_EXECUTORS, false)) {
+      activeExecutors.addAll(executorLoader.fetchActiveExecutors());
```
hluu: If we switched from running local executor to multiple executors, would this also local the local executor?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -98,12 +115,85 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+        executor = executorLoader.addExecutor(executorHost, executorPort);
+      } else if (!executor.isActive()) {
+        executor.setActive(true);
+        executorLoader.updateExecutor(executor);
+      }
+      activeExecutors.add(new Executor(executor.getId(), executorHost,
+        executorPort, true));
+    }
+
+    if (azkProps.getBoolean(AZKABAN_USE_MULTIPLE_EXECUTORS, false)) {
```
hluu: Would be good to log out whether web server is running in local executor or multiple executors modes, if it is the later, also print out list of servers (host/port/status)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -130,10 +220,33 @@ public long getLastCleanerThreadCheckTime() {
...
+   * @see azkaban.executor.ExecutorManagerAdapter#fetchExecutor(int)
+   */
+  @Override
+  public Executor fetchExecutor(int executorId) throws ExecutorManagerException {
+    for (Executor executor : activeExecutors) {
+      if (executor.getId() == executorId) {
+        return executor;
+      }
+    }
+    return executorLoader.fetchExecutor(executorId);
```
hluu: There is no need to update activeExecutors after fetching from DB?  How come?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -130,10 +220,33 @@ public long getLastCleanerThreadCheckTime() {
...
+  }
+
+  @Override
   public Set<String> getPrimaryServerHosts() {
```
hluu: What does it mean primary server?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -141,11 +254,8 @@ public long getLastCleanerThreadCheckTime() {
   public Set<String> getAllActiveExecutorServerHosts() {
     // Includes non primary server/hosts
     HashSet<String> ports = new HashSet<String>();
-    ports.add(executorHost + ":" + executorPort);
-    for (Pair<ExecutionReference, ExecutableFlow> running : runningFlows
-        .values()) {
-      ExecutionReference ref = running.getFirst();
-      ports.add(ref.getHost() + ":" + ref.getPort());
+    for (Executor executor : activeExecutors) {
```
hluu: Are we saying activeExecutors can contain duplicate server hosts and that is why ports is of type HashSet?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+
+  /**
+   * Gets a list of all the active (running, non-dispatched) executions for a
+   * given project and flow {@inheritDoc}
+   *
+   * @see azkaban.executor.ExecutorManagerAdapter#getRunningFlows(int,
+   *      java.lang.String)
+   */
   @Override
   public List<Integer> getRunningFlows(int projectId, String flowId) {
```
hluu: This method is returning a list of execution ids, but method name is "getRunningFlows".  I am a bit confused.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
   public boolean isFlowRunning(int projectId, String flowId) {
-    for (Pair<ExecutionReference, ExecutableFlow> ref : runningFlows.values()) {
+    boolean isRunning = false;
+    isRunning =
+      isRunning
+        || isFlowRunningHelper(projectId, flowId, queuedFlowMap.values());
+    isRunning =
+      isRunning
+        || isFlowRunningHelper(projectId, flowId, runningFlows.values());
+    return false;
```
hluu: seems like this method always return false.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+   * database {@inheritDoc}
+   *
+   * @see azkaban.executor.ExecutorManagerAdapter#getExecutableFlow(int)
+   */
   @Override
   public ExecutableFlow getExecutableFlow(int execId)
-      throws ExecutorManagerException {
-    Pair<ExecutionReference, ExecutableFlow> active = runningFlows.get(execId);
-    if (active == null) {
+    throws ExecutorManagerException {
```
hluu: Looks like formatting is off

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+    if (runningFlows.containsKey(execId)) {
+      return runningFlows.get(execId).getSecond();
+    } else if (queuedFlowMap.containsKey(execId)) {
+      return queuedFlowMap.get(execId).getSecond();
```
hluu: what if queuedFlowMap doesn't contain an entry with execId, will there be NPE?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+   * @see azkaban.executor.ExecutorManagerAdapter#getRunningFlows()
+   */
   public String getRunningFlowIds() {
     List<Integer> allIds = new ArrayList<Integer>();
-    for (Pair<ExecutionReference, ExecutableFlow> ref : runningFlows.values()) {
-      allIds.add(ref.getSecond().getExecutionId());
-    }
+    getRunningFlowsIdsHelper(allIds, queuedFlowMap.values());
+    getRunningFlowsIdsHelper(allIds, runningFlows.values());
+    Collections.sort(allIds);
```
hluu: Is sorting necessary?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -976,6 +996,65 @@ public void postExecutorEvent(Executor executor, EventType type, String user,
...
+  }
+
+  /**
+   *
+   * {@inheritDoc}
+   *
+   * @see azkaban.executor.ExecutorLoader#fetchExecutorByExecution(int)
+   */
+  @Override
+  public Executor fetchExecutorByExecution(int executionId)
```
hluu: fetchExecutorByExecutionId?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1178,13 +1257,80 @@ public String handle(ResultSet rs) throws SQLException {
...
   }
 
+  /**
+   * JDBC ResultSetHandler to fetch queued executions
+   */
+  private static class FetchQueuedExecutableFlows implements
+    ResultSetHandler<List<Pair<ExecutionReference, ExecutableFlow>>> {
+    // Select queued unassigned flows
+    private static String FETCH_QUEUED_EXECUTABLE_FLOW =
+      "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, "
```
hluu: do we have index on join column(s)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: maybe queuedFlowMap and queuedFlowList, to show that they are closely related? (just an idea)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -130,10 +220,33 @@ public long getLastCleanerThreadCheckTime() {
...
+      }
+    }
+    return executorLoader.fetchExecutor(executorId);
+  }
+
+  @Override
   public Set<String> getPrimaryServerHosts() {
-    // Only one for now. More probably later.
+    // TODO: do we want to have a primary
     HashSet<String> ports = new HashSet<String>();
```
johnyu0520: is variable name "ports" a misnomer? As we seem to be putting server:port pair into the HashSet

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+   */
+  private void loadQueuedFlows() throws ExecutorManagerException {
+    queuedFlows.addAll(executorLoader.fetchQueuedFlows());
+    for (Pair<ExecutionReference, ExecutableFlow> ref : queuedFlows) {
+      queuedFlowMap.put(ref.getSecond().getExecutionId(), ref);
+    }
+  }
+
+  /**
+   * Gets a list of all the active (running, non-dispatched) executions for a
```
johnyu0520: maybe change the wording to "running flows and non-dispatched flows"
I misread the comment to (running && non-dispatched)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
+   * given project and flow {@inheritDoc}
+   *
+   * @see azkaban.executor.ExecutorManagerAdapter#getRunningFlows(int,
+   *      java.lang.String)
+   */
   @Override
   public List<Integer> getRunningFlows(int projectId, String flowId) {
     ArrayList<Integer> executionIds = new ArrayList<Integer>();
```
johnyu0520: List<Integer> = new ArrayList<Integer>

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -153,189 +263,328 @@ public long getLastCleanerThreadCheckTime() {
...
-    for (Pair<ExecutionReference, ExecutableFlow> ref : runningFlows.values()) {
+    getRunningFlowsHelper(projectId, flowId, executionIds,
+      queuedFlowMap.values());
+    getRunningFlowsHelper(projectId, flowId, executionIds,
+      runningFlows.values());
+    return executionIds;
+  }
+
+  /* Helper method for getRunningFlows */
+  private void getRunningFlowsHelper(int projectId, String flowId,
```
johnyu0520: I was thinking that it might be better for readability if we do
1. this method returns List<Integer>
2. the calling method calls  executionIds.addAll(retval)

Wondering if people have thoughts/preferences to this

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -597,22 +864,16 @@ public String submitExecutableFlow(ExecutableFlow exflow, String userId)
...
-            "Execution submitted successfully with exec id "
-                + exflow.getExecutionId();
-      } catch (ExecutorManagerException e) {
-        executorLoader.removeActiveExecutableReference(reference.getExecId());
-        throw e;
-      }
-
+      Pair<ExecutionReference, ExecutableFlow> pair =
+        new Pair<ExecutionReference, ExecutableFlow>(reference, exflow);
+      queuedFlowMap.put(exflow.getExecutionId(), pair);
```
johnyu0520: want to create a queue and dequeue method for accessing the Map and the List? this way we can make sure people will not forget to do one or the other operation

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -98,12 +115,85 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+   *    this method add local executor as active in DB
+   * 2. If a local executor is specified and it is marked inactive in db,
+   *    this method will convert local executor as active in DB
+   * 3. If azkaban.use.multiple.executors is set true, this method will
+   *    load all active projects
+   * </pre>
+   *
+   * @throws ExecutorManagerException
+   */
+  public void setupExecutors() throws ExecutorManagerException {
```
evlstyle: logic registering self as a new executor (in DB) I think is better  to be resided to the exeServer side  .  Here we should only load all the settings from the DB to keep the logic clean.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
evlstyle: seems like the activeExecutors list is build when the webserver is initialized, and after that seems no way to add a new executor to the list without bounce the webserver. I guess it will become a common user scenario to add a new executor to a specific environment and I think to avoid restarting the web box in such event will be really nice.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -130,10 +220,33 @@ public long getLastCleanerThreadCheckTime() {
   }
 
   @Override
+  public Set<Executor> getAllActiveExecutors() {
+    return activeExecutors;
```
evlstyle: guess returning a copy of the list might be a better idea as the in memory list is your only  "data of truth",  you may not want the list to be messed up by the caller.

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
```diff
@@ -18,16 +18,22 @@
 
 public class ExecutionReference {
```
evlstyle: not sure if it is true but it seems like we no longer need this ExecutionReference any more, the new executor IS in the concept of this executionReference so wrapping up again the executor is a little bit over.

azkaban-common/src/main/java/azkaban/executor/ExecutorManagerAdapter.java
```diff
@@ -88,6 +88,18 @@ public ExecutableFlow getExecutableFlow(int execId)
...
+  /**
+   * <pre>
+   * Returns All running with executors and queued flows
+   * Note, returns empty list if there isn't any running or queued flows
+   * </pre>
+   *
+   * @return
+   * @throws IOException
+   */
+  public List<Pair<ExecutableFlow, Executor>> getActiveFlowsWithExecutor()
```
evlstyle: I really think we should create a new Class, or recycle some rarely used Class (say the executionReference) to represent this data structure so that we can replace the first and second call with some meaningful getter names.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: +1 on Evan's logic.  Might be more straightforward to reason

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1256,76 +1361,287 @@ private void fillUpdateTimeAndExecId(List<ExecutableFlow> flows,
...
+            queuedFlows.poll();
+            queuedFlowMap.remove(exflow.getExecutionId());
+
+            try {
+              reference.setExecutor(choosenExecutor);
+              executorLoader.assignExecutor(choosenExecutor.getId(),
+                exflow.getExecutionId());
+              // TODO: ADD rest call to do an actual dispatch
+              callExecutorServer(reference, ConnectorParams.EXECUTE_ACTION);
+              runningFlows
```
evlstyle: where is the logic to remove the flow from the queuedFlow list after a successful dispatch ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1256,76 +1361,287 @@ private void fillUpdateTimeAndExecId(List<ExecutableFlow> flows,
...
-      int from, int length, Status status) throws ExecutorManagerException {
-    return executorLoader.fetchFlowHistory(projectId, flowId, from, length,
-        status);
+  private void processQueuedFlows(long maxContinousSubmission) {
+    try {
+      Pair<ExecutionReference, ExecutableFlow> runningCandidate;
+      int submissionNum = 0;
+      while (submissionNum < maxContinousSubmission
+        && (runningCandidate = queuedFlows.peek()) != null) {
+        synchronized (runningCandidate) {
```
evlstyle: I think we will still have racing condition if only synchronize on the runningCandidate.  

think the scenario - 
two threads hit this simultaneously, both passed the peek and got the same object reference,  then one thread passes into the synchronized block and one waiting,  thread #1 processes the flow, assign executor , put it in the running queue then exit the block, then thread #2 kicks, do all the work again (which may assign a different executor) and put again the flow in the running flow list, which is dup , which is not what we want.  

Let me know if I got it right ,  I think we should be locking on the queuedFlows object instead of  the runningcandidate.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1256,76 +1361,287 @@ private void fillUpdateTimeAndExecId(List<ExecutableFlow> flows,
...
+                  new Pair<ExecutionReference, ExecutableFlow>(reference,
+                    exflow));
+            } catch (ExecutorManagerException e) {
+              logger.error("Failed to process queued flow", e);
+              // TODO: allow N errors and re-try
+              finalizeFlows(exflow);
+            }
+          }
+        }
+        submissionNum++;
```
evlstyle: nicer place to increase the submissionNum might be in the block where a successful dispatch happened.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
evlstyle: I think it is better to sync on the queuedFlows, it is actually a required change assuming you want to remove the item from the queue after a successful dispatching,  the logic of which seems to be missing right now in the code,  as Pair is immutable, so locking on the queuedFlows should provide no less security compared to locking on the returned object from queuedFlows.peek().

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
logiclord: We want to avoid interface changes in ExecutorManager, due to which we need to keep this class.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: good idea. name changed to queuedFlowList

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: simplified use case to assume that we will not have local as well as multi-executor simultaneously. Though we need to make a db entry due to the way we are now defining a flow running i.e. it has an assigned executor. So in local mode, we need to have an executor to assign to. This db requirement is for the use case where we use the ability to bounce webserver without affecting running flows.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: added duration. Legacy code :(

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: @hluu 
as discussed offline, now no local and multiexecutor mode is mutually exclusive. Switching between from local<->multiexecutor needs a change in flag in azkaban.properties hence we need to bounce webserver for that. 

@evlstyle 
Yup that is why this method is public. I plan to have a curl request for the time being. In case of multi executor mode we will be able to refresh executors without restarting.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Added log statement for the mode. Executor lists is available in /jmx page. I have changed behaviour of /jmx page from showing only executors which are running a flow to all active executors (executorManager is managing).

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Legacy code. I have made all executors as primary for the time being.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Again legacy code. ActiveExecutors are now set so we don't need a set here.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Yup I was talking about this method when I mentioned about confusing methods. This is an existing public method which now returns ids of running as well as non-dispatched flows.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: executor fetched may not be an active executor.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: nope is returns true when isFlowRunningHelper returns true i.e. when project-flow ids are present in either queuedFlowMap or runningFlows.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: fixed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: no that that condition is guarded with queuedFlowMap.containsKey(execId)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: To be backward compatible. Older interface returns a sorted version.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: changed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: legacy code. Changed to executors.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
logiclord: on all except execution_flows(executor_id). Added for executor_id

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Again legacy code :(. Fixed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: I like this idea. Updated unless someone has some concerns.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Even I though of this and created these methods. It will be part of next iteration of code change.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Very good point. Fixed to return a readonly version.

azkaban-common/src/main/java/azkaban/executor/ExecutorManagerAdapter.java
logiclord: Even I want a separate class, Ideally I will prefer to recycle executionReference but We need to be backward compatible and Pair<ExecutableFlow, Executor> is used through out ExecutorManager. So we cannot make this change without an interface change.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: As discussed offline, queue is now a BoundedQueue so we remove from queue using take() method each time we process a flow.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: good idea. added

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Using BoundedQueue has obviated this change.

logiclord:
Incorporated feedback and added queue processing as decided in design discussion today.

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -33,6 +33,8 @@
   public static final String CONCURRENT_OPTION_SKIP = "skip";
   public static final String CONCURRENT_OPTION_PIPELINE = "pipeline";
   public static final String CONCURRENT_OPTION_IGNORE = "ignore";
+  public static final String FLOW_PRIORITY = "flowPriority";
+  public static final int DEFAULT_FLOW_PRIORITY = 5;
```
johnyu0520: Do we want to document somewhere saying whether bigger number is higher/lower priority?

azkaban-common/src/main/java/azkaban/executor/Executor.java
```diff
@@ -107,4 +107,8 @@ public int getId() {
   public void setActive(boolean isActive) {
     this.isActive = isActive;
   }
+
+  public String toString() {
+    return String.format("%s:%d", host, port);
```
johnyu0520: might want to include id and isActive? (should be helpful for debugging)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -71,8 +80,20 @@
   private ConcurrentHashMap<Integer, ExecutableFlow> recentlyFinished =
       new ConcurrentHashMap<Integer, ExecutableFlow>();
 
-  private ExecutingManagerUpdaterThread executingManager;
+  /* map to easily access queued flows */
+  final private ConcurrentHashMap<Integer, Pair<ExecutionReference, ExecutableFlow>> queuedFlowMap =
```
johnyu0520: I am actually thinking since these code are mostly newly added, is this actually a good time to create a separate class and put them outside of this already big ExecutorManager?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: @logiclord  I think you accidentally hard coded the return value?

azkaban-common/src/main/java/azkaban/executor/Executor.java
hluu: +1

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -0,0 +1,303 @@
...
+      ExecutorManager manager = createMultiExecutorManagerInstance();
+      ExecutableFlow flow1 = createExecutableFlow("exec1");
+      flow1.getExecutionOptions().setConcurrentOption(
+        ExecutionOptions.CONCURRENT_OPTION_SKIP);
+
+      User testUser = getTestUser();
+      manager.submitExecutableFlow(flow1, testUser.getUserId());
+      manager.submitExecutableFlow(flow1, testUser.getUserId());
+      manager.enableQueueProcessorThread();
+      manager.submitExecutableFlow(flow1, testUser.getUserId());
```
hluu: Is this third submit necessary?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -0,0 +1,303 @@
...
+  /* Test renabling queue process thread to pause restart dispatching */
+  @Test
+  public void testEnablingQueueProcessThread() throws ExecutorManagerException {
+    ExecutorManager manager = createMultiExecutorManagerInstance();
+    Assert.assertEquals(manager.isQueueProcessorThreadActive(), false);
+    manager.enableQueueProcessorThread();
+    Assert.assertEquals(manager.isQueueProcessorThreadActive(), true);
+  }
+
+  /* Test submit a non-dispatched flow */
```
hluu: what is a non-dispatched flow?

azkaban-common/src/main/java/azkaban/jmx/JmxExecutorManagerMBean.java
```diff
@@ -39,4 +39,14 @@
 
   @DisplayName("OPERATION: getPrimaryExecutorHostPorts")
   public List<String> getPrimaryExecutorHostPorts();
+
+  @DisplayName("OPERATION: isQueueProcessorActive")
+  public boolean isQueueProcessorActive();
+
+  @DisplayName("OPERATION: getUndispatchedFlows")
```
hluu: Is undispatched same as queued?

getQueuedFlows?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -371,18 +672,34 @@ public JobMetaData getExecutionJobMetaData(ExecutableFlow exFlow,
...
+    throws ExecutorManagerException {
     synchronized (exFlow) {
-      Pair<ExecutionReference, ExecutableFlow> pair =
+      if (runningFlows.containsKey(exFlow.getExecutionId())) {
+        Pair<ExecutionReference, ExecutableFlow> pair =
           runningFlows.get(exFlow.getExecutionId());
-      if (pair == null) {
+        callExecutorServer(pair.getFirst(), ConnectorParams.CANCEL_ACTION,
+          userId);
+      } else if (queuedFlowMap.containsKey(exFlow.getExecutionId())) {
```
johnyu0520: just thinking, do you think we should check in the queuedFlowMap first, then check in the runningFlows?

Otherwise there is the very very very slim chance of when the check is being performed on RunningFlows, it was being taken out of the queue, and when we check the queue, it is already in the RunningFlows

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1709,156 @@ private void cleanExecutionLogs() {
...
+
+      while (isActive() && (runningCandidate = waitAndFetchQueueHead()) != null) {
+        ExecutionReference reference = runningCandidate.getFirst();
+        ExecutableFlow exflow = runningCandidate.getSecond();
+
+        long currentTime = System.currentTimeMillis();
+        if (currentTime - lastProcessingTime > activeExecutorsRefreshWindow
+              || currentContinuousFlowProcessed >= maxContinuousFlowProcessed) {
+          refreshExecutors(); // Refresh executor stats to be used by selector
+          lastProcessingTime = currentTime;
```
johnyu0520: should lastProcessingTime something like [last|previous]Executor[Refresh|Update]Time?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -0,0 +1,303 @@
...
+    props.put(ExecutorManager.AZKABAN_USE_MULTIPLE_EXECUTORS, "true");
+    props.put(ExecutorManager.AZKABAN_QUEUEPROCESSING_ENABLED, "false");
+
+    loader.addExecutor("localhost", 12345);
+    loader.addExecutor("localhost", 12346);
+    return new ExecutorManager(props, loader, new HashMap<String, Alerter>());
+  }
+
+  /* Helper method to create an ExecutableFlow from serialized description */
+  private ExecutableFlow createExecutableFlow(String flowName)
```
johnyu0520: A method that can be put in a TestUtils class and reused?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -0,0 +1,303 @@
...
+    Assert.assertTrue(managerActiveFlows.containsAll(testFlows)
+      && testFlows.containsAll(managerActiveFlows));
+
+    // Verify getQueuedFlowIds method
+    Assert.assertEquals("[1, 2]", manager.getQueuedFlowIds());
+  }
+
+  /* Test submit duplicate flow when previous instance is not dispatched */
+  @Test
+  public void testDuplicateQueuedFlows() throws ExecutorManagerException,
```
johnyu0520: expecting an exception?
http://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -334,6 +335,149 @@ public void testUploadExecutableNode() throws Exception {
 
   }
 
+  /* Test exception when assigning a non-existent executor to a flow */
+  @Test
```
johnyu0520: change to test for exception

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -334,6 +335,149 @@ public void testUploadExecutableNode() throws Exception {
...
+    try {
+      loader.assignExecutor(flow.getExecutionId(), 1);
+      Assert.fail("Expecting exception, but didn't get one");
+    } catch (ExecutorManagerException ex) {
+      System.out.println("Test true");
+    }
+  }
+
+  /* Test exception when assigning an executor to a non-existent flow execution */
+  @Test
```
johnyu0520: change to test for exception

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -821,6 +829,15 @@ private void ajaxExecuteFlow(HttpServletRequest req,
     ret.put("execid", exflow.getExecutionId());
   }
 
+  /* Reset flow priority if submitting user is not a Azkaban admin */
+  private void fixFlowPriorityByPermission(ExecutionOptions options, User user) {
```
johnyu0520: just wondering, is it better if we default to DEFAULT_FLOW_PRIORITY and only bump it up to ADMIN_PRIORITY if the submitter is an admin?

Though logically it will behave the same

azkaban-common/src/main/java/azkaban/executor/Executor.java
logiclord: changed

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
logiclord: From user's perspective it will be covered in documentation and from code perspective examples and details are defined in ExecutableFlowPriorityComparator and ExecutableFlowPriorityComparatorTest

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Nice suggestion. Refactored to defined a separate data structure class and corresponding tests.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
logiclord: nope.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
logiclord: non-dispatched  = Webserver queued flow. Initially i was planning to use non-dispatched nomenclature.

azkaban-common/src/main/java/azkaban/jmx/JmxExecutorManagerMBean.java
logiclord: yes. Changed

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
logiclord: Changed in some tests. Its not recommended to be used when multiple lines can through same expected exception but we want to validate exception from a specific line of code.

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
logiclord: ExecutorManager manager has no visibility into the permission level for the submit user. So we need to fix up set/unset properties in ExecutorServlet. Moreover we don't want all admin submitted flow to have high priority

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
logiclord: same reply as above.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
logiclord: same reply as above.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: changed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: We are synchronizing on exflow so if a flow is taken out then cancel flow call will be blocked until we finish processing the flow.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
logiclord: incorporated

logiclord:
Incorporated feedback comments.

johnyu0520:
LGTM

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
```diff
@@ -0,0 +1,81 @@
...
+
+import azkaban.utils.Pair;
+
+public final class ExecutableFlowPriorityComparator implements
+  Comparator<Pair<ExecutionReference, ExecutableFlow>> {
+  private static Logger logger = Logger
+    .getLogger(ExecutableFlowPriorityComparator.class);
+
+  @Override
+  public int compare(Pair<ExecutionReference, ExecutableFlow> pair1,
```
hluu: Please add javadoc for the comparison logic - descending order by priority,  ascending order by update time and ascending order by execution id

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
```diff
@@ -0,0 +1,81 @@
...
+
+package azkaban.executor;
+
+import java.util.Comparator;
+
+import org.apache.log4j.Logger;
+
+import azkaban.utils.Pair;
+
+public final class ExecutableFlowPriorityComparator implements
```
hluu: Add javadoc, that this will be used in the priority queue

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
```diff
@@ -0,0 +1,81 @@
...
+      && options.getFlowParameters() != null
+      && options.getFlowParameters()
+        .containsKey(ExecutionOptions.FLOW_PRIORITY)) {
+      try {
+        priority =
+          Integer.valueOf(options.getFlowParameters().get(
+            ExecutionOptions.FLOW_PRIORITY));
+      } catch (NumberFormatException ex) {
+        priority = ExecutionOptions.DEFAULT_FLOW_PRIORITY;
+        logger.error("Failed to parse flow priority for exec_id = "
```
hluu: should add add ex to the log.error()?

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
```diff
@@ -18,16 +18,19 @@
...
+
+  public ExecutionReference(int execId) {
+    this.execId = execId;
+  }
+
+  public ExecutionReference(int execId, Executor executor) {
     this.execId = execId;
-    this.host = host;
-    this.port = port;
+    this.executor = executor;
```
hluu: Should we check for null because we will reference it later - executor.getHost(), executor.getPort()?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -59,10 +63,19 @@
  */
 public class ExecutorManager extends EventHandler implements
     ExecutorManagerAdapter {
+  static final String AZKABAN_QUEUEPROCESSING_ENABLED =
```
hluu: small comment - is AZKABAN prefix necessary?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -98,14 +120,139 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+   * 3. In local mode, If a local executor is specified and it is marked inactive in db,
+   *    this method will convert local executor as active in DB
+   * </pre>
+   *
+   * @throws ExecutorManagerException
+   */
+  public void setupExecutors() throws ExecutorManagerException {
+    Set<Executor> newExecutors = new HashSet<Executor>();
+
+    if (isMultiExecutorMode()) {
```
hluu: cool

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -539,80 +853,97 @@ private void applyDisabledJobs(List<Object> disabledJobs,
...
+            .format(
+              "Failed to submit %s for project %s. Azkaban has overrun its webserver queue capacity",
+              flowId, exflow.getProjectName());
+        logger.error(message);
+      } else {
+        int projectId = exflow.getProjectId();
+        exflow.setSubmitUser(userId);
+        exflow.setSubmitTime(System.currentTimeMillis());
+
+        List<Integer> running = getRunningFlows(projectId, flowId);
```
hluu: runningFlowIds?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -539,80 +853,97 @@ private void applyDisabledJobs(List<Object> disabledJobs,
...
-        options = new ExecutionOptions();
-      }
+      logger.info("Submitting execution flow " + flowId + " by " + userId);
 
       String message = "";
-      if (options.getDisabledJobs() != null) {
-        applyDisabledJobs(options.getDisabledJobs(), exflow);
-      }
+      if (queuedFlows.isFull()) {
+        message =
```
hluu: If it is a manual submit, how can we report this error back to user?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -539,80 +853,97 @@ private void applyDisabledJobs(List<Object> disabledJobs,
...
+          String
+            .format(
+              "Failed to submit %s for project %s. Azkaban has overrun its webserver queue capacity",
+              flowId, exflow.getProjectName());
+        logger.error(message);
```
hluu: if it is a scheduled flow, can we update the flow log with this error?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -539,80 +853,97 @@ private void applyDisabledJobs(List<Object> disabledJobs,
...
+            throw e;
+          }
+        }
         message +=
-            "Execution submitted successfully with exec id "
-                + exflow.getExecutionId();
-      } catch (ExecutorManagerException e) {
-        executorLoader.removeActiveExecutableReference(reference.getExecId());
-        throw e;
+          "Execution submitted successfully with exec id "
```
hluu: This message seems weird - successfully in catch clause?  how come?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
           - executionLogsRetentionMs);
     }
   }
-}
+
+  /*
+   * This thread is responsible for processing queued flows.
```
hluu: Anything special that should be called out to give readers a sense of the responsibility of this class?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+        synchronized (this) {
+          try {
+            // start processing queue if active, other wait for sometime
+            if (isActive) {
+              processQueuedFlows(activeExecutorRefreshWindowInMilisec,
+                activeExecutorRefreshWindowInFlows);
+            }
+            wait(QUEUE_PROCESSOR_WAIT_IN_MS);
+          } catch (Exception e) {
+            logger.info(
```
hluu: How come not logger.error()?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+   * This thread is responsible for processing queued flows.
+   */
+  private class QueueProcessorThread extends Thread {
+    private static final long QUEUE_PROCESSOR_WAIT_IN_MS = 1000;
+    private static final int MAX_DISPATCHING_ERRORS_PERMITTED = 5;
+    private final long activeExecutorRefreshWindowInMilisec;
+    private final int activeExecutorRefreshWindowInFlows;
+
+    private boolean shutdown = false;
+    private boolean isActive = true;
```
hluu: I believe this needs to be volatile, because it will be updated and read by different threads.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+  /*
+   * This thread is responsible for processing queued flows.
+   */
+  private class QueueProcessorThread extends Thread {
+    private static final long QUEUE_PROCESSOR_WAIT_IN_MS = 1000;
+    private static final int MAX_DISPATCHING_ERRORS_PERMITTED = 5;
+    private final long activeExecutorRefreshWindowInMilisec;
+    private final int activeExecutorRefreshWindowInFlows;
+
+    private boolean shutdown = false;
```
hluu: I believe this needs to be volatile, because it will be updated and read by different threads.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+      setActive(isActive);
+      this.activeExecutorRefreshWindowInFlows =
+        activeExecutorRefreshWindowInFlows;
+      this.activeExecutorRefreshWindowInMilisec =
+        activeExecutorRefreshWindowInTime;
+      this.setName("AzkabanWebServer-QueueProcessor-Thread");
+    }
+
+    public void setActive(boolean isActive) {
+      this.isActive = isActive;
```
hluu: Would it be helpful for debugging purpose to log out a statement when this method is called.  This way, we can easily tell from the log.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+      long lastExecutorRefreshTime = System.currentTimeMillis();
+      Pair<ExecutionReference, ExecutableFlow> runningCandidate;
+      int currentContinuousFlowProcessed = 0;
+
+      while (isActive() && (runningCandidate = queuedFlows.fetchHead()) != null) {
+        ExecutionReference reference = runningCandidate.getFirst();
+        ExecutableFlow exflow = runningCandidate.getSecond();
+
+        long currentTime = System.currentTimeMillis();
+        if (currentTime - lastExecutorRefreshTime > activeExecutorsRefreshWindow
```
hluu: It would be more clear to refactor this piece of logic into a method and name a method clearly to represent the logic

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+          || currentContinuousFlowProcessed >= maxContinuousFlowProcessed) {
+          refreshExecutors(); // Refresh executor stats to be used by selector
+          lastExecutorRefreshTime = currentTime;
+          currentContinuousFlowProcessed = 0;
+        }
+
+        // process flow with current snapshot of activeExecutors
+        processFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
```
hluu: Seems expensive to create an instance of HashSet in a loop.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+        currentContinuousFlowProcessed++;
+      }
+    }
+
+    /* process flow with a snapshot of available Executors */
+    private void processFlow(ExecutionReference reference,
+      ExecutableFlow exflow, Set<Executor> availableExecutors)
+      throws ExecutorManagerException {
+      synchronized (exflow) {
```
hluu: why synchronized on exflow?  multiple threads are manipulating a specific instance of exflow?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+        Executor selectedExecutor = selectExecutor(exflow, availableExecutors);
+        if (selectedExecutor != null) {
+          try {
+            dispatch(reference, exflow, selectedExecutor);
+          } catch (ExecutorManagerException e) {
+            logger.debug(String.format(
```
hluu: I wonder if this should be logger.warn() instead of logger.debug(), which we will not see in normal running mode.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+              "Executor %s responded with exception for exec: %d",
+              selectedExecutor, exflow.getExecutionId()), e);
+            handleDispatchExceptionCase(reference, exflow, selectedExecutor,
+              availableExecutors);
+          }
+        } else {
+          handleNoExecutorSelectedCase(reference, exflow);
```
hluu: Good to see helper method so the logic flow doesn't get cluttered.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+      } else {
+        remainingExecutors.remove(lastSelectedExecutor);
+        // try other executors except chosenExecutor
+        processFlow(reference, exflow, remainingExecutors);
+      }
+    }
+
+    private void handleNoExecutorSelectedCase(ExecutionReference reference,
+      ExecutableFlow exflow) throws ExecutorManagerException {
+      reference.setNumErrors(reference.getNumErrors() + 1);
```
hluu: Should we add logging to indicate this condition?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+      Executor choosenExecutor;
+      // TODO: use dispatcher
+      choosenExecutor = availableExecutors.iterator().next();
+      return choosenExecutor;
+    }
+
+    private void handleDispatchExceptionCase(ExecutionReference reference,
+      ExecutableFlow exflow, Executor lastSelectedExecutor,
+      Set<Executor> remainingExecutors) throws ExecutorManagerException {
+      reference.setNumErrors(reference.getNumErrors() + 1);
```
hluu: Should we add logging to indicate this condition?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1384,4 +1665,161 @@ private void cleanExecutionLogs() {
...
+      // to be moved after db update once we integrate rest api changes
+      reference.setExecutor(choosenExecutor);
+      // TODO: ADD rest call to do an actual dispatch
+      callExecutorServer(reference, ConnectorParams.EXECUTE_ACTION);
+      executorLoader.assignExecutor(exflow.getExecutionId(),
+        choosenExecutor.getId());
+
+      // move from flow to running flows
+      runningFlows.put(exflow.getExecutionId(),
+        new Pair<ExecutionReference, ExecutableFlow>(reference, exflow));
```
hluu: Should we add logging to indicate successful dispatch for flow to specific executor?

azkaban-common/src/main/java/azkaban/executor/QueuedExecutions.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.concurrent.PriorityBlockingQueue;
+
+import org.apache.log4j.Logger;
+
+import azkaban.utils.Pair;
+
+/**
+ * <pre>
+ * Composite data structure to represent non-dispatched flows in webserver.
+ * This data structure wraps a blocking queue and a concurrent hashmap.
```
hluu: Would be good to understand what the blocking queue and concurrent hashmap are used for and why they are needed.

azkaban-common/src/test/java/azkaban/executor/ExecutableFlowPriorityComparatorTest.java
```diff
@@ -0,0 +1,128 @@
...
+import azkaban.utils.JSONUtils;
+import azkaban.utils.Pair;
+import azkaban.utils.Props;
+import azkaban.utils.TestUtils;
+
+/**
+ * Test class for ExecutableFlowPriorityComparator
+ * */
+
+public class ExecutableFlowPriorityComparatorTest {
```
hluu: Are there any tests based on execution ids?

azkaban-sql/src/sql/update.execution_logs.2.1.sql
```diff
@@ -1,6 +0,0 @@
-ALTER TABLE execution_logs ADD COLUMN attempt INT DEFAULT 0;
```
hluu: Any reasons why we are removing these files?

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
logiclord: moved embedded comments to java doc

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
logiclord: added

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowPriorityComparator.java
logiclord: added

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: just out of convention since AZKABAN_QUEUEPROCESSING_ENABLED =
    "azkaban.queueprocessing.enabled" property is exposed via azkaban.properties.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Legacy code. This is a public method, changing this method will change interface of ExecutorManager.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: it is outside the catch but inside the synchronized block. Catch completes at line 942

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Great catch

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: fixed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: I was planning to have it logged in servlet which will take care of curl request to enable/disable it. changed to log in this method.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: to synchronize for other existing operations like cancel.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Actually flow has not failed at this point so I didn't intend to log this step under normal running mode to avoid over logging.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: added log to track stage transition for a flow to handleNoExecutorSelectedCase and handleDispatchExceptionCase.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: added log to track stage transition for a flow to handleNoExecutorSelectedCase and handleDispatchExceptionCase.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: added

azkaban-common/src/main/java/azkaban/executor/QueuedExecutions.java
logiclord: as commented BlockingQueue is the actual queue for undispatched flows and map to easily access queued flows (to support existing operations)

azkaban-sql/src/sql/update.execution_logs.2.1.sql
logiclord: These changes are already part of existing create* scripts

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
logiclord: added a ExecutorManagerException for invalid arguments.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: expanded documentation

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: As discussed offline,  I have added comment for this if statement. All the variables in this if statements are loop variables and it will be better to have it here instead of having a method with 4 arguments.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: As discussed offline, we are going ahead with cloning activeExecutor set as we will have just one copy of executors at a time.

azkaban-common/src/test/java/azkaban/executor/ExecutableFlowPriorityComparatorTest.java
logiclord: added

logiclord:
Incorporated new feedback comments

hluu:
LGTM

davidzchen:
Tracking bug: #439

PR #484, opened 2015-09-09
switch the process to collect statistics data back into single thread mode

evlstyle:
as the atomic procedure is now fast enough, switching back to the single thread mode.


hluu:
LGTM

PR #485, opened 2015-09-09
Adjust data type for some of the interfaces for easy consumption.

evlstyle:
change some of the interface from asking for List to Collection object so that the caller doesn't need to do conversion if it maintains an object of type Set.


logiclord:
LGTM

PR #486, opened 2015-09-10
fix exception while property value is null

flowbehappy:
The value of properties may be null when using hadoop 2. We should ignore them to avoid exceptions when passing to `Matcher#matcher` in `resolveVariableReplacement()`


hluu:
Thanks for the patch.

davidzchen:
Can we merge this?

suvodeep-pyne:

PR #487, opened 2015-09-11
Integrating ExecutorSelector and exposing curl requests

logiclord:
This pull requests include :- 
* Integration changes to use ExecutorSelector form #469
* Using ExecutorRESTApiClient instead of custom built REST calls from executor manager
* New Curl requests to enable/disable queueprocessing and to reload active executors
* Ability to override dispatcher
* Unit Tests


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1002,57 +1050,63 @@ private void cleanOldExecutionLogs(long millis) {
...
+   * Helper method used by ExecutorManager to call executor and return raw json
+   * string
+   */
+  private String callExecutorForJsonString(String host, int port, String path,
+    List<Pair<String, String>> paramList) throws IOException {
+    if (paramList == null) {
+      paramList = new ArrayList<Pair<String, String>>();
+    }
+
+    ExecutorApiClient apiclient = ExecutorApiClient.getInstance();
```
evlstyle: seems no cache a copy of the reference here, you may just call ExecutorApiClient.getInstance().httpGet()

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1776,12 +1776,49 @@ private void processFlow(ExecutionReference reference,
...
-      Executor choosenExecutor;
-      // TODO: use dispatcher
-      choosenExecutor = availableExecutors.iterator().next();
+      Executor choosenExecutor = getUserSpecifiedExecutor(exflow);
+
+      // If no executor was specified by admin
+      if (choosenExecutor == null) {
+        logger.info("Using dispatcher for execution id :"
+          + exflow.getExecutionId());
+        ExecutorSelector selector = new ExecutorSelector(filterList, comparatorWeightsMap);
```
evlstyle: do we need a new instance of the selector one select is called ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -201,15 +215,44 @@ private boolean isMultiExecutorMode() {
...
+          public void run() {
+            try {
+              String jsonResponse =
+                callExecutorForJsonString(executor.getHost(),
+                  executor.getPort(), "/serverstastics", null);
+              executor.setExecutorInfo(ExecutorInfo
+                .fromJSONString(jsonResponse));
+            } catch (Exception e) {
+              logger.error("Failed to update ExecutorInfo executorId :"
+                + executor, e);
```
evlstyle: nit - executor.toString() writes out the executor hostname  but not the id .

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -201,15 +215,44 @@ private boolean isMultiExecutorMode() {
    */
   private void refreshExecutors() {
     synchronized (activeExecutors) {
-      // TODO: rest api call to refresh executor stats
+      ExecutorService taskExecutors =
+        Executors.newFixedThreadPool(activeExecutors.size());
```
evlstyle: nit - newFixedThreadPool() throws illigalArguementException if the input is <=0,  better to short cut if the activeExecutors is empty beforehead.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -201,15 +215,44 @@ private boolean isMultiExecutorMode() {
...
+          }
+        });
+      }
+
+      taskExecutors.shutdown();
+      try {
+        // wait 5 seconds for all executors to be refreshed
+        taskExecutors.awaitTermination(5, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        logger.error("Timed out while waiting for executorInfo refresh", e);
```
evlstyle: InterruptedException is raised when the current thread (the thread hosts the Executor) is terminated, it does not indicate that the threads in pool incurred timeout while awaited for finishing.  the function call  awaitTermination() actually returns a Boolean value which indicates if the wait time is expired. You may want to read that value and write log accordingly. also it might be nicer to tell which Executor timed out as it might be a key piece of data for debugging.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1776,12 +1776,49 @@ private void processFlow(ExecutionReference reference,
...
+              ExecutionOptions.USE_EXECUTOR));
+          executor = fetchExecutor(executorId);
+
+          if (executor == null) {
+            logger
+              .warn(String
+                .format(
+                  "User specified executor id: %d for execution id: %d is not active, Looking up db.",
+                  executorId, exflow.getExecutionId()));
+            executor = executorLoader.fetchExecutor(executorId);
```
evlstyle: better to log if DB attempt also fails to find the executor. I see this condition is not logged in the upper stream.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -121,41 +125,51 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+    String filters = azkProps.getString(AZKABAN_EXECUTOR_SELECTOR_FILTERS, "");
+    if (filters != null) {
+      filterList = Arrays.asList(StringUtils.split(filters, ","));
+    }
+
+    // initliatize comparator feature weights for executor selector from
+    // azkaban.properties
+    Map<String, String> compListStrings =
+      azkProps.getMapByPrefix(AZKABAN_EXECUTOR_SELECTOR_COMPARATOR_PREFIX);
+    if (compListStrings != null) {
```
hluu: If there are no selector comparator prefixes then everything else still works?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -121,41 +125,51 @@ public ExecutorManager(Props props, ExecutorLoader loader,
...
+        comparatorWeightsMap.put(entry.getKey(), Integer.valueOf(entry.getValue()));
+      }
+    }
+
+    // configure queue processor
+    queueProcessor =
+      new QueueProcessorThread(azkProps.getBoolean(
+        AZKABAN_QUEUEPROCESSING_ENABLED, true), azkProps.getLong(
+        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_MS, 1000), azkProps.getInt(
+        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_FLOW, 1000));
```
hluu: should we have MS in the key name as well so it is consistent with AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_MS?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -201,15 +215,44 @@ private boolean isMultiExecutorMode() {
    */
   private void refreshExecutors() {
     synchronized (activeExecutors) {
-      // TODO: rest api call to refresh executor stats
+      ExecutorService taskExecutors =
```
hluu: Thread pool is a bit expensive, is it possible to reuse the taskExecutors thread pool?  any reason why you don't want to reuse?

This method can just submit refresh tasks.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
hluu: What are the impacts when tasks don't come back in 5 seconds?  stale statistics?

It would be nice to have visibility into this refresh tasks  - i.e last refresh timestamp in the JMX page.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1776,12 +1776,49 @@ private void processFlow(ExecutionReference reference,
...
+            logger
+              .warn(String
+                .format(
+                  "User specified executor id: %d for execution id: %d is not active, Looking up db.",
+                  executorId, exflow.getExecutionId()));
+            executor = executorLoader.fetchExecutor(executorId);
+          }
+
+        } catch (Exception ex) {
+          logger.error("Failed to fetch user specified executor for exec_id = "
```
hluu: What are the scenarios where an exception can happen?  Does it make sense to bubble up the exception rather than logging it?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -135,6 +136,12 @@ private void handleAJAXAction(HttpServletRequest req,
           ajaxFetchExecutableFlowInfo(req, resp, ret, session.getUser(), exFlow);
         }
       }
+    } else if (ajaxName.equals("reloadExecutors")) {
```
hluu: Do we have a jmx bean to indicate the status of the queue processor?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: what does this AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_FLOW control?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1776,12 +1776,49 @@ private void processFlow(ExecutionReference reference,
       }
     }
 
+    /* Helper method to fetch  overriding Executor, if a valid user has specifed otherwise return null */
+    private Executor getUserSpecifiedExecutor(ExecutableFlow exflow) {
```
johnyu0520: an aesthetic choice:  Do you think we should just pass in the ExecutionOptions object? This method doesn't seem to require other things in the exflow.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1776,12 +1776,49 @@ private void processFlow(ExecutionReference reference,
       }
     }
 
+    /* Helper method to fetch  overriding Executor, if a valid user has specifed otherwise return null */
```
johnyu0520: update the javadoc to be more understandable?

Also my understanding on the logic is that if anything happens, it will return a null.
Is it possible that you want to return with different types of exceptions (or different exception messages) so that we know what went wrong in this method? (db connection error vs not being able to find specified user specified executor)

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: +1 on reusing.
I was also thinking whether a threadpool's necessary, or just an arraylist would do?  Because you are not using the threadpool to limit parallel execution or anything?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1812,27 +1848,20 @@ private void handleNoExecutorSelectedCase(ExecutionReference reference,
       .info(String
         .format(
           "Reached handleNoExecutorSelectedCase stage for exec %d with error count %d",
```
johnyu0520: do we want to print a message that is more user/oncall actionable?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -158,6 +165,61 @@ private void handleAJAXAction(HttpServletRequest req,
     }
   }
 
+  /**
+   * Enables queueProcessor if @param status is true and disables queueProcessor
+   * is @param status is false.
```
johnyu0520: nit: formatting is off, and second "is" should be "if"?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -158,6 +165,61 @@ private void handleAJAXAction(HttpServletRequest req,
...
+   */
+  private void ajaxUpdateQueueProcessor(HttpServletRequest req,
+    HttpServletResponse resp, HashMap<String, Object> returnMap, User user,
+    boolean status) {
```
johnyu0520: should status be renamed to enableQueue or queueEnabled?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: This is to keep things clean for multiple calls

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: This in line with your request to instantiate selector every time with choose an executor, probably to avoid synchronization issues with changing weights. For the time being, No we can have a class level instance.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Good catch, I think this change got dropped as it was conflicting for #478. Modified toString method.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: We should never have a activeExecutors.size() =  0 as we fail AzkabanWebServer, if there is no activeExecutors and we do not update activeExecutors if new executor set is empty.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: added

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: ExecutorManager will pass null comparatorMap to ExecutorSelector which is handled by selector.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: @johnyu0520  maximum number of flows that can execute before we will have to refresh executorInfo
@hluu Do mean add in_ms in azkaban.activeexecutor.refresh.milisecinterval ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: We can have executorManagerException and SQLException.

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
logiclord: yes

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: We return null as we want to default to normal behaviour of asking selector to pick an executor. Though I have added more log to identify different exception scenarios.

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
logiclord: updated

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
logiclord: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
johnyu0520: I guess my question is, 
might the caller function/code want to know which exception scenario it ran into?

If you have it in logs humans will understand why but it would be manual intervention

johnyu0520:
LGTM

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: changed to _IN_NUM_FLOW

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: As discussed, moved executor service as member of executor manager and used futures to handle different scenarios

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Exposed last successful refresh time in jmx and used futures to handle different scenarios

logiclord:
Incorporated feedback

hluu:
LGTM

PR #490, opened 2015-09-15
Azkaban to run user jobs using user credentials instead of using Azkaban

johnyu0520:
This will close a very big security hole where everybody who is running commands on the local Azkaban box is running as the user Azkaban.  In the not so bad case, different users might set Kerberos tickets and affect one another.  In the worst case, if Azkaban user has special privileges, people will be able to exploit that and wreak havoc


azkaban-execserver/src/main/resources/execute-as-user.c
```diff
@@ -0,0 +1,111 @@
...
+#include <sys/stat.h>
+#include <sys/mount.h>
+#include <sys/types.h>
+#include <pwd.h>
+
+  FILE *LOGFILE = NULL;
+  FILE *ERRORFILE = NULL;
+  int SETUID_OPER_FAILED=10;
+
+/**
```
hluu: Would be good to mention that this code is borrowed from Hadoop

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -81,8 +83,23 @@ public void run() throws Exception {
...
     File[] propFiles = initPropsFiles();
     Map<String, String> envVars = getEnvironmentVariables();
-
+  
+    String nativeLibFolder = null;
+    String executeAsUserBinary = null;
+    String userToProxy = null;
+    boolean isExecuteAsUser = sysProps.getBoolean(EXECUTE_AS_USER, false);
+    
+    if(isExecuteAsUser){
```
hluu: Not exactly sure how we would go about doing unit testing on this feature, but something to think about.

azkaban-execserver/src/main/resources/execute-as-user.c
```diff
@@ -0,0 +1,111 @@
...
+        ERRORFILE=stderr;
+
+
+
+    char *uid = argv[1];
+
+    // for loop to calculate the length to malloc
+    int i;
+    int total_len = 0;
+    for(i=2;i<argc;i++){
```
logiclord: indenting is off for c file

azkaban-execserver/src/main/resources/execute-as-user.c
```diff
@@ -0,0 +1,111 @@
...
+        total_len += strlen(argv[i])+1;
+    }
+    printf("total_len: %d\n", total_len);
+
+    // allocate memory and clear memory
+    char *cmd = malloc(total_len+2);
+    memset(cmd, ' ', total_len+2);
```
logiclord: cmd can be NULL if system failed to allocate memory

logiclord: writing space will write int "32" in each byte for the memory allocated to char pointer. You can use cmd memset(cmd, '\0', total_len+2) or memset(cmd, 0, total_len+2)

azkaban-execserver/src/main/resources/execute-as-user.c
```diff
@@ -0,0 +1,111 @@
+#include <dirent.h>
```
logiclord: license ?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
johnyu0520: I've confirmed the flag to work, which will enable the worst case scenario of rolling back.
However, how to warn users beforehand and the impact analysis will be something harder to do

azkaban-execserver/src/main/resources/execute-as-user.c
johnyu0520: done

azkaban-execserver/src/main/resources/execute-as-user.c
johnyu0520: done

azkaban-execserver/src/main/resources/execute-as-user.c
johnyu0520: sorry, what should this be changed to?

azkaban-execserver/src/main/resources/execute-as-user.c
johnyu0520: fixed both malloc and memset

johnyu0520:
closing the pull request for now as more modification is in progress

PR #491, opened 2015-09-15
Auto upgrade database up to database.properties version and track draining flows

logiclord:
At present, Azkaban neglect update* scripts while automatically creating database when we run Azkaban in Solo Server mode.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/database/AzkabanDatabaseSetup.java
+++ b/azkaban-common/src/main/java/azkaban/database/AzkabanDatabaseSetup.java
@@ -242,9 +242,15 @@ public class AzkabanDatabaseSetup {
         upgradeList.put(key, upgradeVersions);
       }
     }
+    for (String key : missingTables) {
+      List<String> upgradeVersions = findOutOfDateTable(key, "");
+      if (upgradeVersions != null && !upgradeVersions.isEmpty()) {
+        upgradeList.put(key, upgradeVersions);
+      }
+    }
   }
 
-  private List<String> findOutOfDateTable(String table, String version) {
+  private List<String> findOutOfDateTable(String table, String currentVersion) {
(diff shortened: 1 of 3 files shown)


hluu:
LGTM

johnyu0520:
LGTM

azkaban-common/src/main/java/azkaban/database/AzkabanDatabaseSetup.java
```diff
@@ -255,30 +261,29 @@ private void findOutOfDateTables() {
...
-
-          if (versionNum.compareTo(version) == 0) {
-            versions.add(versionNum);
+        if (updateScriptVersion.endsWith(".")) {
+          updateScriptVersion = updateScriptVersion.substring(0, updateScriptVersion.length() - 1);
+
+          // add to update list if updateScript will update above current
+          // version and upto targetVersion in database.properties
+          if (updateScriptVersion.compareTo(currentVersion) > 0
+            && updateScriptVersion.compareTo(this.version) <= 0) {
```
evlstyle: You may want to state somewhere that padding must be enforced in the major/minor versions reach multi digits. otherwise this string comparison may fail.

evlstyle: alternatively,  as you have already tried to parse the integers, maybe it will be nicer to use the parsed integer(s) to do the comparison instead of using string.

evlstyle: Other than that that change looks good to me.

logiclord: We already have a standard format to specify azkaban version i.e. digit.digit.digit. Moreover, we will control future versions which automatically takes care of padding.

PR #494, opened 2015-09-16
add time param replacement

flowbehappy:
In schedule job, parameters based on current time are very common, like a daily job consumes a file named `/tmp/from/${time:-1day,yyyyMMdd}.txt`, and `${time:-1day,yyyyMMdd}` will be replaced into `20150915` if the day it execute is September 16, 2015, i.e. the real file name passed to the job will be `/tmp/from/20150915.txt`. Although technically a job can do this itself, it  would be a great convenient.

This patch enable time parameters in job configures, the grammar is `${time:<delta period>,<time format>}`. The format is following ISO8601,  and the unit of delta period could be `year`, `month`, `day`, `hour`, `minute` and `second`.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/ParamReplacement.java
+++ b/azkaban-common/src/main/java/azkaban/utils/ParamReplacement.java
@@ -1,9 +1,11 @@
 package azkaban.utils;
 
 import com.google.common.collect.Lists;
+import org.joda.time.DateTime;
+import org.joda.time.Period;
+import org.joda.time.format.PeriodFormatter;
+import org.joda.time.format.PeriodFormatterBuilder;
 
-import java.text.SimpleDateFormat;
-import java.util.Date;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/utils/ParamReplacement.java
```diff
@@ -0,0 +1,88 @@
...
+            .appendSuffix("minute").appendSeconds().appendSuffix("second")
+            .toFormatter();
+
+        public String name() {
+          return "time";
+        }
+
+        public String replacement(String param) {
+          if (param == null) {
+            param = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX";
```
juhoautio: Could make that configurable globally:
```java
resolvedProps.getString("time.replacement.default.format", "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
```

So that I could for example set `time.replacement.default.format=yyyyMMdd` on the highest level and then just use `${time:-1day}` in some jobs to yield `20160203` etc.

flowbehappy: ok

suvodeep-pyne:

Issue #496, opened 2015-09-22
Update Azkaban documentation for other_namenodes property


hluu:

logiclord:
Fixed in #496

PR #497, opened 2015-09-22
Update Azkaban documentation for other_namenodes property #496

hluu:


johnyu0520:
+1 for changing second example to hdfs://
otherwise LGTM

johnyu0520:
LGTM

PR #500, opened 2015-09-22
Fixing single executor to multi-executor transition and work around work around for global busy state

logiclord:
* Fixing single executor to multi-executor transition
* Work around work around for global busy state
* Fixing cold start to refreshExecutors first time we start processing the queue.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -173,7 +173,7 @@ public class ExecutorManager extends EventHandler implements
       new QueueProcessorThread(azkProps.getBoolean(
         AZKABAN_QUEUEPROCESSING_ENABLED, true), azkProps.getLong(
         AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_MS, 1000), azkProps.getInt(
-        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_NUM_FLOW, 1000));
+        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_NUM_FLOW, 10));
 
     queueProcessor.start();
   }
@@ -255,8 +255,9 @@ public class ExecutorManager extends EventHandler implements
           // max 5 secs
           String jsonString = refreshPair.getSecond().get(5, TimeUnit.SECONDS);
           executor.setExecutorInfo(ExecutorInfo.fromJSONString(jsonString));
-          logger.info("Successfully refreshed ExecutorInfo for executor: "
-            + executor);
+          logger.info(String.format(
+            "Successfully refreshed executor: %s with executor info : %s",
+            executor, jsonString));
         } catch (TimeoutException e) {
           wasSuccess = false;
           logger.error("Timed out while waiting for ExecutorInfo refresh"
@@ -1809,7 +1810,7 @@ public class ExecutorManager extends EventHandler implements
     private void processQueuedFlows(long activeExecutorsRefreshWindow,
       int maxContinuousFlowProcessed) throws InterruptedException,
       ExecutorManagerException {
-      long lastExecutorRefreshTime = System.currentTimeMillis();
+      long lastExecutorRefreshTime = 0;
       Pair<ExecutionReference, ExecutableFlow> runningCandidate;
       int currentContinuousFlowProcessed = 0;
 
@@ -1830,9 +1831,32 @@ public class ExecutorManager extends EventHandler implements
           currentContinuousFlowProcessed = 0;
         }
 
-        exflow.setUpdateTime(currentTime);
-        // process flow with current snapshot of activeExecutors
-        processFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
+        /**
+         * <pre>
+         *  TODO: Work around till we improve Filters to have a notion of GlobalSystemState.
+         *        Currently we try each queued flow once to infer a global busy state
+         * Possible improvements:-
+         *   1. Move system level filters in refreshExecutors and sleep if we have all executors busy after refresh
+         *   2. Implement GlobalSystemState in selector or in a third place to manage system filters. Basically
+         *      taking out all the filters which do not depend on the flow but are still being part of Selector.
+         * Assumptions:-
+         *   1. no one else except QueueProcessor is updating ExecutableFlow update time
+         *   2. re-attempting a flow (which has been tried before) is considered as all executors are busy
+         * </pre>
+         */
+        if(exflow.getUpdateTime() > lastExecutorRefreshTime) {
+          // put back in the queue
+          queuedFlows.enqueue(exflow, reference);
+          long sleepInterval =
+            activeExecutorsRefreshWindow
+              - (currentTime - lastExecutorRefreshTime);
+          // wait till next executor refresh
+          sleep(sleepInterval);
+        } else {
+          exflow.setUpdateTime(currentTime);
+          // process flow with current snapshot of activeExecutors
+          processFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
+        }
         currentContinuousFlowProcessed++;
       }
     }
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -117,8 +117,10 @@
 
   public ExecutorManager(Props props, ExecutorLoader loader,
       Map<String, Alerter> alters) throws ExecutorManagerException {
+    alerters = alters;
```
hluu: why named alters?  how about this.alerts = alerts?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -173,7 +173,7 @@ private void setupMultiExecutorMode() {
       new QueueProcessorThread(azkProps.getBoolean(
         AZKABAN_QUEUEPROCESSING_ENABLED, true), azkProps.getLong(
         AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_MS, 1000), azkProps.getInt(
-        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_NUM_FLOW, 1000));
+        AZKABAN_ACTIVE_EXECUTOR_REFRESHINTERVAL_IN_NUM_FLOW, 10));
```
hluu: I think the INTERVAL part is causing the confusion.

How about AZKABAN_ACTIVE_EXECUTOR_REFRESH_BY_NUM_FLOW?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -255,8 +255,9 @@ public String call() throws Exception {
           // max 5 secs
           String jsonString = refreshPair.getSecond().get(5, TimeUnit.SECONDS);
           executor.setExecutorInfo(ExecutorInfo.fromJSONString(jsonString));
-          logger.info("Successfully refreshed ExecutorInfo for executor: "
-            + executor);
+          logger.info(String.format(
```
hluu: How big is the jsonString?  Do we really need to see the jsonString? or may be log out jsonString in a separate statement with debug level.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1729,6 +1727,28 @@ private void cleanExecutionLogs() {
...
+  /**
+   * Calls executor to dispatch the flow, update db to assign the executor and
+   * in-memory state of executableFlow
+   */
+  private void dispatch(ExecutionReference reference, ExecutableFlow exflow,
+    Executor choosenExecutor) throws ExecutorManagerException {
+    exflow.setUpdateTime(System.currentTimeMillis());
+    callExecutorServer(exflow, choosenExecutor,
+      ConnectorParams.EXECUTE_ACTION);
+    executorLoader.assignExecutor(choosenExecutor.getId(),
```
johnyu0520: I am just wondering: what happens if this sql command does not complete and throws an exception?
The system will be in an inconsistent state of having submitted the flow, but not having it registered in the DB (and not be in the RunningFlows map, etc).  Should we handle this scenario?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1729,6 +1727,28 @@ private void cleanExecutionLogs() {
     }
   }
 
+  /**
+   * Calls executor to dispatch the flow, update db to assign the executor and
+   * in-memory state of executableFlow
+   */
+  private void dispatch(ExecutionReference reference, ExecutableFlow exflow,
+    Executor choosenExecutor) throws ExecutorManagerException {
+    exflow.setUpdateTime(System.currentTimeMillis());
```
hluu: should we update the time after successfully calling executor?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1811,9 +1831,32 @@ private void processQueuedFlows(long activeExecutorsRefreshWindow,
...
+         * Assumptions:-
+         *   1. no one else except QueueProcessor is updating ExecutableFlow update time
+         *   2. re-attempting a flow (which has been tried before) is considered as all executors are busy
+         * </pre>
+         */
+        if(exflow.getUpdateTime() > lastExecutorRefreshTime) {
+          // put back in the queue
+          queuedFlows.enqueue(exflow, reference);
+          long sleepInterval =
+            activeExecutorsRefreshWindow
```
hluu: can this be negative? under what scenario that it can be negative?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1811,9 +1831,32 @@ private void processQueuedFlows(long activeExecutorsRefreshWindow,
...
+              - (currentTime - lastExecutorRefreshTime);
+          // wait till next executor refresh
+          sleep(sleepInterval);
+        } else {
+          exflow.setUpdateTime(currentTime);
+          // process flow with current snapshot of activeExecutors
+          processFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
```
johnyu0520: nit-picking on names again :)
Can we name it something along the lines of SelectAndDispatchFlow? and also spell it out in the javadoc for it?
"process" can potentially mean lots of different things even within the ExecutorManager.java context

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: changed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: changed

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: As discussed, it will have about 4 lines with 25-30 chars each. We will retaining it as info for the time being .

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: Fixed as discussed.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: no we may fail to call executor and our priority queue logic depends on update time. Moreover, executor server will update updatetime once it is successfully dispatched.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: it can't be negative.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
logiclord: changed

logiclord:
Incorporated feedback. Please have a re-look.

hluu:
For some reason the CI failed weird error.

logiclord:
Seems like a transient issue, It should pass as I seems to work on my box. I have re-triggered the build.

hluu:
LGTM

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1729,6 +1726,36 @@ private void cleanExecutionLogs() {
...
+  /**
+   * Calls executor to dispatch the flow, update db to assign the executor and
+   * in-memory state of executableFlow
+   */
+  private void dispatch(ExecutionReference reference, ExecutableFlow exflow,
+    Executor choosenExecutor) throws ExecutorManagerException {
+    exflow.setUpdateTime(System.currentTimeMillis());
+
+    try {
+      executorLoader.assignExecutor(choosenExecutor.getId(),
```
evlstyle: the call to assignExecutor should be placed outside the try catch block, as if that thing fails the filed in DB will not be updated with the executor id, therefore there is no need to reset it.

logiclord: makes sense. Though unassign will be a no-op but It will be cleaner. Updated the change.

PR #506, opened 2015-09-28
set refresh interval to 10 sec and using remaining executors

logiclord:
increasing executors refresh interval to 10 sec and using remaining executors instead of all executors


johnyu0520:
LGTM

PR #512, opened 2015-09-29
Removing cached ExecutorInfos and changing defaults

logiclord:
* Removing cached ExecutorInfos
* changing default params for maxDispatchingErrors


johnyu0520:
LGTM

PR #514, opened 2015-10-06
Fix to use appropriate flow watcher for pipelined execution

logiclord:
Pipelined flow watcher depends on execution ids to be a sorted list.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -453,14 +453,12 @@ public class ExecutorManager extends EventHandler implements
   }
 
   /**
-   * <pre>
    * Gets a list of all the active (running flows and non-dispatched flows)
    * executions for a given project and flow {@inheritDoc}. Results should
    * be sorted as we assume this while setting up pipelined execution Id.
    *
    * @see azkaban.executor.ExecutorManagerAdapter#getRunningFlows(int,
    *      java.lang.String)
-   * </pre>
    */
   @Override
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -453,11 +453,14 @@ private void loadQueuedFlows() throws ExecutorManagerException {
   }
 
   /**
+   * <pre>
```
davidzchen: Why the <pre> tags?

johnyu0520: I believe <pre> will help to preserve the format once it goes into HTML

davidzchen: I don't think that is necessary here because it is just a paragraph and not code. `<pre>` is used displaying code blocks within in your documentation: http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html

Also, AFAIU, using `<pre>` here will not render the Javadoc tags within this documentation, such as `{@inheritdoc}` because they are treated as code and will be displayed as such.

johnyu0520: I see.  That's a good point

logiclord: removed <pre>. It's a good learning, we have been using pre for a while.

davidzchen:
LGTM

johnyu0520:
LGTM

PR #524, opened 2015-10-08
Re-enable DirectoryFlowLoaderTest and ExecutableFlowTest.

### LINKED ISSUE #518: Fix azkaban.util.DirectoryFlowLoaderTest
Re-enable azkaban.util.DirectoryFlowLoaderTest

### LINKED ISSUE #519: Fix azkaban.executor.ExecutableFlowTest
Re-enable azkaban.executor.ExecutableFlowTest

davidzchen:
* Copy resource files needed by test into correct resources directory.
* Update test code to correctly load resource files.
* Re-enable tests

Fixes #518
Fixes #519


azkaban-common/src/test/resources/azkaban/executor/embedded/innerFlow.job
```diff
@@ -0,0 +1,5 @@
+type=javaprocess
+java.class=azkaban.test.executor.SleepJavaJob
```
logiclord: why don't we reuse flows from https://github.com/azkaban/azkaban/tree/master/azkaban-test/src/test/resources/executions ?

azkaban-common/src/test/java/azkaban/utils/DirectoryFlowLoaderTest.java
```diff
@@ -17,42 +17,56 @@
...
+    Assert.assertNotNull(url);
+    File resourceFile = new File(url.toURI());
+    File parentDir = resourceFile.getParentFile();
+    return parentDir;
+  }
 
-  @Ignore @Test
-  public void testDirectoryLoad() {
+  @Test
+  public void testDirectoryLoad() throws Exception {
```
logiclord: should we throw specific exceptions ?

azkaban-common/src/test/resources/azkaban/executor/embedded/innerFlow.job
davidzchen: Unfortunately, there does not seem to be a standard way to have this kind of cross-subproject dependency for resources in Gradle. The resources for a test must also be under the same package directory in the resource source set as the test. For example, tests under `test/azkaban/executor` need their resources to be under `resources/azkaban/executor`.

Do you know of a non-hackish way to use resource files from a separate subproject in Gradle? There are some other build systems such as [Bazel](http://bazel.io/docs/build-encyclopedia.html#java_test.resources) that have an easy way to specify this kind of dependency.

Otherwise, I think having a bit of duplication for test files should be fine, especially since these test files are relatively small.

azkaban-common/src/test/java/azkaban/utils/DirectoryFlowLoaderTest.java
davidzchen: Done.

davidzchen:
Friendly ping.

azkaban-common/src/test/resources/azkaban/executor/embedded/innerFlow.job
logiclord: I am not sure if you will consider it as hack. Can you have a look at https://github.com/azkaban/azkaban/blob/multipleexecutors_canary/azkaban-common/src/test/java/azkaban/utils/TestUtils.java approach ?

davidzchen: Thanks! Actually, something like that is not a bad idea. I have removed the duplicated test files and added a utility class to `azkaban-test` (`TestExecutions`) with a method that other tests can use to get a `File` object for the directory of one of the flows in `azkaban-test`.

This way, we build a JAR from `azkaban-test` that contains this class and all the test files, and other tests can use the `TestExecutions` class, which will load the resource files in the `azkaban-test` jar using its own class loader.

I have also moved all the test flows in `azkaban-test` into the correct resource directory so that they can be loaded from the jar, though this would mean that we would need to update the paths in the `TestUtils` class that you wrote.

davidzchen:
@logiclord - Thanks! Actually, something like that is not a bad idea. I have removed the duplicated test files and added a utility class to azkaban-test (TestExecutions) with a method that other tests can use to get a File object for the directory of one of the flows in azkaban-test.

This way, we build a JAR from azkaban-test that contains this class and all the test files, and other tests can use the TestExecutions class, which will load the resource files in the azkaban-test jar using its own class loader.

I have also moved all the test flows in azkaban-test into the correct resource directory so that they can be loaded from the jar, though this would mean that we would need to update the paths in the TestUtils class that you wrote.

logiclord:
LGTM

PR #545, opened 2015-10-13
number of processed based executor refresh invalidates inferred global state 

logiclord:
a number of processed based executor refresh invalidates logic to identify global busy state

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1872,7 +1872,11 @@ public class ExecutorManager extends EventHandler implements
           // process flow with current snapshot of activeExecutors
           selectExecutorAndDispatchFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
         }
-        currentContinuousFlowProcessed++;
+
+        // do not count failed flow processsing (flows still in queue)
+        if(queuedFlows.getFlow(exflow.getExecutionId()) == null) {
+          currentContinuousFlowProcessed++;
+        }
       }
     }
 


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1872,7 +1872,11 @@ private void processQueuedFlows(long activeExecutorsRefreshWindow,
           // process flow with current snapshot of activeExecutors
           selectExecutorAndDispatchFlow(reference, exflow, new HashSet<Executor>(activeExecutors));
         }
-        currentContinuousFlowProcessed++;
+
+        // do not count failed flow processsing (flows still in queue)
+        if(queuedFlows.getFlow(exflow.getExecutionId()) == null) {
+          currentContinuousFlowProcessed++;
```
johnyu0520: I am curious, shouldn't 
"currentContinuousFlowProcessed++;"
be added right after line 1873?

otherwise we will count 1 extra when a exflow wakes up from the sleep section,.even though it hasn't been assigned

logiclord: If a flow wakes up from sleep then its still in queue so we will not increment the count.

johnyu0520: ok I guess the logic will perform correctly.  

Just curious why not put the currentContinuousFlowProcessed++;
command inside the else statement?

johnyu0520: after sync up with Gaurav, looks good

PR #548, opened 2015-10-14
Run Travis CI against both JDK 7 and 8.

### LINKED ISSUE #547: Run CI against both JDK 7 and 8
Travis has support for testing against multiple JDK versions: http://blog.travis-ci.com/support_for_multiple_jdks/

We should update our CI config to test against both JDK 7 and 8 -- the two versions we currently support.

davidzchen:
Also, upgrade to Travis's container-based infrastructure.

Fixes #547

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,8 @@
 languages: java
+# TODO(dzc): Add openjdk8 once it is available in Travis.
+jdk:
+  - openjdk7
+  - oraclejdk7
+  - oraclejdk8
+sudo: false
 script: ./gradlew distTar


davidzchen:
Ping.

.travis.yml
```diff
@@ -1,2 +1,7 @@
 languages: java
+jdk:
+  - openjdk7
```
logiclord: why not open openjdk8 ?

davidzchen: Unfortunately, Travis only supports oraclejdk8 for now and has not added support for openjdk8 yet. Adding openjdk8 to this config results in an error.

logiclord:
LGTM

.travis.yml
davidzchen: I have added a TODO comment to add openjdk8 once it becomes available.

PR #549, opened 2015-10-21
Bug fix - assign permission to a group while a user with same name exists 

evlstyle:
Fix a bug when assigning project permission to a group the name of which equals to a user account which already exists in the permission list of the same project.


logiclord:
LGTM

azkaban-sql/src/sql/update.project_permissions.3.0.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE project_permissions drop PRIMARY KEY, add PRIMARY KEY(project_id, name,isGroup);
```
johnyu0520: format nit: maybe a space behind the comma?

otherwise LGTM

Issue #564, opened 2015-11-11
Improve error message of scheduleFlow API


wyukawa:
It is difficult to understand the error message of scheduleFlow API.

example

```
Flow test_flow cannot be found in project azkaban.project.Project@13846b77
```

https://github.com/azkaban/azkaban/blob/master/azkaban-webserver/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L630

In this case, I don't know the Azkaban project name.

I expect the following

```
Flow test_flow cannot be found in project test
```

I always change schedule by using Azkaban API, so I will be happy if it is easy to understand the error message.

PR #566, opened 2015-11-13
Issue #564 - Improve error message of scheduleFlow API

wyukawa:
https://github.com/azkaban/azkaban/issues/564

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
@@ -637,7 +637,7 @@ public class ScheduleServlet extends LoginAbstractAzkabanServlet {
     if (flow == null) {
       ret.put("status", "error");
       ret.put("message", "Flow " + flowName + " cannot be found in project "
-          + project);
+          + projectName);
       return;
     }
 


juhoautio:
Hey that's really useful 👍

wyukawa:
Rebase for master changes.

ping.

HappyRay:
Thanks. 

Merged since the API design issue/question is an existing one.

wyukawa:
Thank you for merging.

My opinion is as follows.

Hmm, in my current usage, I use /manager API with project name, as result I get project ID.
https://github.com/wyukawa/eboshi/blob/master/eboshi/project.py

And then I use /schedule API with project name and project ID.
https://github.com/wyukawa/eboshi/blob/master/eboshi/add_schedule.py

Although current API desigin requires more cost because project ID is necessary, I don't think about introducing more opportunities for clients to make mistakes.

And compatibility is important.

Thanks

HappyRay:
Isn't there an opportunity for clients to specify an inconsistent project name and project ID to this API? In this case, the cost of such mistake seem to only affect logging. Would it be better to ignore the project name parameter and get the name from the project object directly?

fsi206914:
@HappyRay It makes full senses to me. I got confused when I was reading this part of code.

wyukawa:
+1

PR #532, opened 2015-10-09
adding execute-as-user feature.  refer to documentation on usage details

johnyu0520:
A lot cleaner commit messages.

Thanks to David Chen for all the tips and tricks!

PR #567, opened 2015-11-13
Merging execute-as-user into Azkaban 3

johnyu0520:
Most of the code has been reviewed in this previous pull request (which was closed and replaced by this one: https://github.com/azkaban/azkaban/pull/532)

I have squashed the recent commits, but the commits from Oct and Sep are beyond repair.  I'll remember to squash them in the future.

Thanks,
John


azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -0,0 +1,125 @@
...
+                getuid(), getgid(), geteuid(), getegid());
+        return SETUID_OPER_FAILED;
+    }
+
+    return 0;
+}
+
+int main(int argc, char **argv){
+
+// set up the logging stream
```
logiclord: nit indentation

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -0,0 +1,125 @@
...
+    fflush(LOGFILE);
+    retval = execvp(*user_argv, user_argv);
+    fprintf(LOGFILE, "system call return value: %d\n", retval);
+
+    // sometimes system(cmd) returns 256, which is interpreted to 0, making a failed job a successful job
+    // hence this goofy piece of if statement.
+    if (retval != 0){
+        return 1;
+    }
+    else{
```
logiclord: nit indentation. I think you can reformat this file.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -354,7 +355,7 @@ public Job buildJobExecutor(String jobId, Props jobProps, Logger logger)
       if (pluginLoadProps != null) {
         pluginLoadProps = PropsUtils.resolveProps(pluginLoadProps);
       } else {
-        pluginLoadProps = new Props();
+        pluginLoadProps = pluginSet.getCommonPluginLoadProps();
```
logiclord: This will fail command job if we do not have plugins directory. We should check whether pluginLoadProps = Null after this else and set pluginLoadProps = new Props(); in that scenario.

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
```diff
@@ -155,7 +170,14 @@ public boolean softKill(final long time, final TimeUnit unit)
     checkStarted();
     if (processId != 0 && isStarted()) {
       try {
-        Runtime.getRuntime().exec("kill " + processId);
+        if (isExecuteAsUser) {
+          String cmd =
+              String.format("%s %s kill %d", executeAsUserBinary,
```
logiclord: shall we make "kill" a constant like KILL_COMMAND

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
```diff
@@ -101,13 +114,15 @@ public void run() throws IOException {
...
-      if (exitCode != 0) {
-        throw new ProcessFailureException(exitCode, errorGobbler.getRecentLog());
-      }
 
       // try to wait for everything to get logged out before exiting
       outputGobbler.awaitCompletion(5000);
       errorGobbler.awaitCompletion(5000);
+
+      if (exitCode != 0) {
+        throw new ProcessFailureException(exitCode, errorGobbler.getRecentLog());
```
logiclord: Should we append stdout as well ? 
While fixing Java 1.8 test cases, I observed that some failures where logged in stdout and not stderr.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -143,6 +175,15 @@ public void run() throws Exception {
     generateProperties(propFiles[1]);
   }
 
+  private boolean determineExecuteAsUser(Props sysProps, Props jobProps) {
+    boolean isExecuteAsUser = sysProps.getBoolean(EXECUTE_AS_USER, false);
+    // putting an override in case user needs to override. A temporary opening
+    if (jobProps.containsKey(EXECUTE_AS_USER_OVERRIDE))
+      isExecuteAsUser = jobProps.getBoolean(EXECUTE_AS_USER_OVERRIDE, false);
```
logiclord: 1 You don't need default value of "false" as you have already verified that EXECUTE_AS_USER_OVERRIDE is present in jobProps
2 nit parenthesis ?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -101,16 +104,45 @@ public void run() throws Exception {
...
+    Map<String, String> envVars = getEnvironmentVariables();
     envVars.put(KRB5CCNAME, getKrb5ccname(jobProps));
 
+    // determine whether to run as Azkaban or run as effectiveUser
+    String executeAsUserBinaryPath = null;
+    String effectiveUser = null;
+    boolean isExecuteAsUser = determineExecuteAsUser(sysProps, jobProps);
+
+    if (isExecuteAsUser) {
+      String nativeLibFolder = sysProps.getString(NATIVE_LIB_FOLDER);
```
logiclord: If nativeLibFolder don't exist, should we validate and throw an informed error ?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -101,16 +104,45 @@ public void run() throws Exception {
...
+      executeAsUserBinaryPath =
+          String.format("%s/%s", nativeLibFolder, "execute-as-user");
+      effectiveUser = getEffectiveUser(jobProps);
+      if ("root".equals(effectiveUser)) {
```
logiclord: what if root is really a user in azkaban-users.xml and hence a valid azkaban user as well as unix user, who wants to execute a job ?

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcessBuilder.java
```diff
@@ -116,4 +124,19 @@ public String toString() {
     return "ProcessBuilder(cmd = " + Joiner.on(" ").join(cmd) + ", env = "
         + env + ", cwd = " + workingDir + ")";
   }
+
+  public AzkabanProcessBuilder setExecuteAsUser() {
```
logiclord: may be rename to enableExecuteAsUser

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -0,0 +1,125 @@
+/*
```
davidzchen: Is this being built with the build system? I don't see any build.gradle changes here.

johnyu0520: looks like it disappeared by accident.  Added back in

azkaban-common/src/main/c/execute-as-user.c
johnyu0520: done

azkaban-common/src/main/c/execute-as-user.c
johnyu0520: (won't fix for now)

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
johnyu0520: If the concern is to make sure that the specified file exists, I have verified during testing that it will throw a FileNotFoundException further down the line, which is the expected behavior

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
johnyu0520: Agreed we are sacrificing this option.
However, since execute-as-user can pose as root and potentially take over the Azkaban box, I would rather be safe than sorry, and make this trade off

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
johnyu0520: Good point.

I did this because the 2 lines for EXECUTE_AS_USER_OVERRIDE (there are actually 2 different parameters in this method) is meant to be a short term loophole.  I would imaging those 2 lines be removed when the time comes to close this loophole.

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
johnyu0520: done

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
johnyu0520: done

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcessBuilder.java
johnyu0520: done

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
johnyu0520: done

logiclord:
LGTM

erwa:
What is the internal JIRA for this feature?

PR #572, opened 2015-11-19
Fix to retain flows executing on inactive executors

logiclord:
This change is needed to support rolling upgrade.


johnyu0520:
LGTM

PR #577, opened 2015-11-23
Added plugin directory to the property before resolving as needed by new TDCH jobs

jinhyukchang:
Added plugin directory to the property before resolving as needed by new TDCH jobs


johnyu0520:
LGTM

logiclord:
LGTM

PR #578, opened 2015-11-24
Add context properties 

rapidoo:
In aim to facilite apache proxy on Azkaban, add a propertie web.context=/azkaban

### DIFF
--- a/azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
+++ b/azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
@@ -760,10 +760,10 @@ public class AzkabanWebServer extends AzkabanServer {
         azkabanSettings.getString("web.resource.dir", DEFAULT_STATIC_DIR);
     logger.info("Setting up web resource dir " + staticDir);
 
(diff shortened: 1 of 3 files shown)


azkaban-webserver/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -759,6 +759,10 @@ public static void main(String[] args) throws Exception {
     String staticDir =
         azkabanSettings.getString("web.resource.dir", DEFAULT_STATIC_DIR);
     logger.info("Setting up web resource dir " + staticDir);
+
+    String contextDir =
+        azkabanSettings.getString("web.context", "/");
```
davidzchen: How about `web.site_root` instead? That might be more descriptive.

rapidoo: Sure, please replace the var and the properties 

Fred

rapidoo:
Sorry I miss to use the var in Context ...

rapidoo:
replace by https://github.com/azkaban/azkaban/pull/579

PR #584, opened 2015-12-07
Fix soloserver start script

juhoautio:
Gets rid of this kind of error:

```
log4j:ERROR setFile(null,true) call failed.
java.io.FileNotFoundException: /azkaban-execserver.log (Permission denied)
```


hluu:
LGTM.

Thanks for submitting this patch.

PR #601, opened 2016-01-25
Extending purge API to support purge with project id

logiclord:


azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -520,56 +520,58 @@ private void handleDownloadProject(HttpServletRequest req,
...
-            // project is already deleted
-            if (isOperationSuccessful) {
-                project = projectManager.getProject(projectName);
-                if (project == null) {
-                    ret.put("error", "no project with name : " + projectName);
-                    isOperationSuccessful = false;
-                }
-            }
+      if (hasParam(req, "id")) {
+        project = projectManager.getProject(getParam(req, "id"));
```
evlstyle: you need to cast the getParam result to int here , otherwise getProject will regard it as name but id .

logiclord: Good catch. Fixed as discussed offline.

Issue #605, opened 2016-02-04
Azkaban HTTP callback may skip notification


juhoautio:
Sometimes Azkaban fails to execute the configured HTTP Job callback.

In this case the following error message is printed into the job log:

```
INFO - !!!! WE ARE NOT SUPPORTING JOB CALLBACKS FOR STATUS: READY
```

or

```
INFO - !!!! WE ARE NOT SUPPORTING JOB CALLBACKS FOR STATUS: QUEUED
```

There's a timing/concurrency issue causing this.
- An event is fired about the state change
- When the event is handled, the job status is fetched again from the `ExecutableNode`
- At this point the status may have already been changed, for example from `FAILED` to `QUEUED`, in case of a retry.
- Fix would be to include the actual status into the event at the time of firing it

References
- PR #443 added this feature
- The original spec for job callbacks in issue #442

juhoautio:
Fixed by https://github.com/azkaban/azkaban/pull/606

PR #606, opened 2016-02-04
Fix issue #605: Azkaban HTTP callback may skip notification

juhoautio:
Fix issue #605: Azkaban HTTP callback may skip notification.

- Instead of getting the mutable node status in callbacks, passing the job status - as at the time of firing the event - in additional event data.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/event/Event.java
+++ b/azkaban-common/src/main/java/azkaban/event/Event.java
@@ -16,6 +16,8 @@
 
 package azkaban.event;
 
+import com.google.common.base.Preconditions;
+
 public class Event {
   public enum Type {
     FLOW_STARTED,
@@ -29,16 +31,14 @@ public class Event {
 
   private final Object runner;
   private final Type type;
-  private final Object eventData;
+  private final EventData eventData;
   private final long time;
-  private final boolean shouldUpdate;
 
-  private Event(Object runner, Type type, Object eventData, boolean shouldUpdate) {
+  private Event(Object runner, Type type, EventData eventData) {
     this.runner = runner;
     this.type = type;
     this.eventData = eventData;
     this.time = System.currentTimeMillis();
-    this.shouldUpdate = shouldUpdate;
   }
 
   public Object getRunner() {
@@ -53,24 +53,22 @@ public class Event {
     return time;
   }
 
-  public Object getData() {
+  public EventData getData() {
     return eventData;
   }
 
-  public static Event create(Object runner, Type type) {
-    return new Event(runner, type, null, true);
+  /**
+   * Creates a new event.
+   *
+   * @param runner runner.
+   * @param type type.
+   * @param eventData EventData, null is not allowed.
+   * @return New Event instance.
+   * @throws NullPointerException if EventData is null.
+   */
+  public static Event create(Object runner, Type type, EventData eventData) throws NullPointerException {
+    Preconditions.checkNotNull(eventData, "EventData was null");
+    return new Event(runner, type, eventData);
   }
 
-  public static Event create(Object runner, Type type, Object eventData) {
-    return new Event(runner, type, eventData, true);
-  }
-
-  public static Event create(Object runner, Type type, Object eventData,
-      boolean shouldUpdate) {
-    return new Event(runner, type, eventData, shouldUpdate);
-  }
-
-  public boolean isShouldUpdate() {
-    return shouldUpdate;
-  }
 }
--- a/azkaban-common/src/main/java/azkaban/event/EventData.java
+++ b/azkaban-common/src/main/java/azkaban/event/EventData.java
@@ -1,0 +1,42 @@
+package azkaban.event;
+
+import azkaban.executor.ExecutableNode;
+import azkaban.executor.Status;
+
+/**
+ * Carries an immutable snapshot of the status data, suitable for asynchronous message passing.
+ */
+public class EventData {
+
+  private final Status status;
+  private final String nestedId;
+
+  /**
+   * Creates a new EventData instance.
+   *
+   * @param status node status.
+   */
+  public EventData(Status status) {
+    this(status, null);
+  }
+
+  /**
+   * Creates a new EventData instance.
+   *
+   * @param status node status.
+   * @param nestedId node id, corresponds to {@link ExecutableNode#getNestedId()}.
+   */
+  public EventData(Status status, String nestedId) {
+    this.status = status;
+    this.nestedId = nestedId;
+  }
+
+  public Status getStatus() {
+    return status;
+  }
+
+  public String getNestedId() {
+    return nestedId;
+  }
+
+}
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -33,6 +33,7 @@ import org.apache.log4j.RollingFileAppender;
 
 import azkaban.event.Event;
 import azkaban.event.Event.Type;
+import azkaban.event.EventData;
 import azkaban.event.EventHandler;
 import azkaban.execapp.event.BlockingStatus;
 import azkaban.execapp.event.FlowWatcher;
@@ -268,18 +269,18 @@ public class JobRunner extends EventHandler implements Runnable {
     if (Status.isStatusFinished(nodeStatus)) {
       quickFinish = true;
     } else if (nodeStatus == Status.DISABLED) {
-      changeStatus(Status.SKIPPED, time);
+      nodeStatus = changeStatus(Status.SKIPPED, time);
       quickFinish = true;
     } else if (this.isKilled()) {
-      changeStatus(Status.KILLED, time);
+      nodeStatus = changeStatus(Status.KILLED, time);
       quickFinish = true;
     }
 
     if (quickFinish) {
       node.setStartTime(time);
-      fireEvent(Event.create(this, Type.JOB_STARTED, null, false));
+      fireEvent(Event.create(this, Type.JOB_STARTED, new EventData(nodeStatus)));
       node.setEndTime(time);
-      fireEvent(Event.create(this, Type.JOB_FINISHED));
+      fireEvent(Event.create(this, Type.JOB_FINISHED, new EventData(nodeStatus)));
       return true;
     }
 
@@ -428,21 +429,23 @@ public class JobRunner extends EventHandler implements Runnable {
 
     // Start the node.
     node.setStartTime(System.currentTimeMillis());
+    Status finalStatus = node.getStatus();
     if (!errorFound && !isKilled()) {
-      fireEvent(Event.create(this, Type.JOB_STARTED, null, false));
+      fireEvent(Event.create(this, Type.JOB_STARTED, new EventData(finalStatus)));
       try {
         loader.uploadExecutableNode(node, props);
       } catch (ExecutorManagerException e1) {
         logger.error("Error writing initial node properties");
       }
 
-      if (prepareJob()) {
+      Status prepareStatus = prepareJob();
+      if (prepareStatus != null) {
         // Writes status to the db
         writeStatus();
-        fireEvent(Event.create(this, Type.JOB_STATUS_CHANGED), false);
-        runJob();
+        fireEvent(Event.create(this, Type.JOB_STATUS_CHANGED, new EventData(prepareStatus)));
+        finalStatus = runJob();
       } else {
-        changeStatus(Status.FAILED);
+        finalStatus = changeStatus(Status.FAILED);
         logError("Job run failed preparing the job.");
       }
     }
@@ -454,29 +457,30 @@ public class JobRunner extends EventHandler implements Runnable {
       // So we set it to KILLED to make sure we know that we forced kill it
       // rather than
       // it being a legitimate failure.
-      changeStatus(Status.KILLED);
+      finalStatus = changeStatus(Status.KILLED);
     }
 
     int attemptNo = node.getAttempt();
     logInfo("Finishing job " + this.jobId + " attempt: " + attemptNo + " at "
         + node.getEndTime() + " with status " + node.getStatus());
 
-    fireEvent(Event.create(this, Type.JOB_FINISHED), false);
+    fireEvent(Event.create(this, Type.JOB_FINISHED, new EventData(finalStatus)), false);
     finalizeLogFile(attemptNo);
     finalizeAttachmentFile();
     writeStatus();
   }
 
-  private boolean prepareJob() throws RuntimeException {
+  private Status prepareJob() throws RuntimeException {
     // Check pre conditions
     if (props == null || this.isKilled()) {
       logError("Failing job. The job properties don't exist");
-      return false;
+      return null;
     }
 
+    Status finalStatus;
     synchronized (syncObject) {
       if (node.getStatus() == Status.FAILED || this.isKilled()) {
-        return false;
+        return null;
       }
 
       if (node.getAttempt() > 0) {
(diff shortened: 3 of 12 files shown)


juhoautio:
@logiclord @johnyu0520 @evlstyle – sorry to spam, but is there anything preventing to merge this? It's clearly a bug that's making the notification mechanism quite unstable.

juhoautio:
@HappyRay, wondering if you would have time to check this? We've been running in production with this change for half a year without problems.

juhoautio:
@HappyRay Could you check this? I just rebased & resolved conflicts that had appeared in the meanwhile.

azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Carries a snapshot of the status data, suitable for asynchronous message passing.
+ */
+public class EventData {
+
+  private Status status;
+  private String nestedId;
+
+  public static EventData create(Status status) {
```
HappyRay: What's the motivation for having this static method and making the constructor private?

azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -0,0 +1,34 @@
...
+
+import azkaban.executor.Status;
+
+/**
+ * Carries a snapshot of the status data, suitable for asynchronous message passing.
+ */
+public class EventData {
+
+  private Status status;
+  private String nestedId;
```
HappyRay: What's the purpose of the nestedId? Add a comment?

azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public Status getStatus() {
+    return status;
+  }
+
+  public String getNestedId() {
+    return nestedId;
+  }
+
+  public EventData withNestedId(String nestedId) {
```
HappyRay: What's the use case for making this field mutable after the instance is created?

azkaban-common/src/main/java/azkaban/event/Event.java
```diff
@@ -53,24 +53,13 @@ public long getTime() {
...
   }
 
-  public static Event create(Object runner, Type type) {
-    return new Event(runner, type, null, true);
-  }
-
-  public static Event create(Object runner, Type type, Object eventData) {
-    return new Event(runner, type, eventData, true);
+  public static Event create(Object runner, Type type, EventData eventData) {
+    Preconditions.checkNotNull(eventData, "EventData was null");
```
HappyRay: This will throw an exception if the condition is not met. How did you make sure that all the callers don't pass in null? 

It's a risk that the exception may cause a thread to die. Before this change, such a bug may have been hidden but may not cause a production issue.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -268,18 +269,18 @@ private boolean handleNonReadyStatus() {
...
       quickFinish = true;
     } else if (this.isKilled()) {
-      changeStatus(Status.KILLED, time);
+      nodeStatus = changeStatus(Status.KILLED, time);
       quickFinish = true;
     }
 
     if (quickFinish) {
       node.setStartTime(time);
-      fireEvent(Event.create(this, Type.JOB_STARTED, null, false));
```
HappyRay: Why this change? The old code sets shouldUpdate to false. You removed it. Could you explain your design?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -454,29 +457,31 @@ public void run() {
...
-      changeStatus(Status.KILLED);
+      finalStatus = changeStatus(Status.KILLED);
     }
 
     int attemptNo = node.getAttempt();
     logInfo("Finishing job " + this.jobId + " attempt: " + attemptNo + " at "
         + node.getEndTime() + " with status " + node.getStatus());
 
-    fireEvent(Event.create(this, Type.JOB_FINISHED), false);
+    // TODO THIS IS THE PLACE FOR SURE
```
HappyRay: What do you mean?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -454,29 +457,31 @@ public void run() {
...
     writeStatus();
   }
 
-  private boolean prepareJob() throws RuntimeException {
+  private Status prepareJob() throws RuntimeException {
     // Check pre conditions
     if (props == null || this.isKilled()) {
       logError("Failing job. The job properties don't exist");
-      return false;
+      return null;
```
HappyRay: Can this return value cause public static Event create(Object runner, Type type, EventData eventData) {
 -    return new Event(runner, type, null, true);		 +    Preconditions.checkNotNull(eventData, "EventData was null"); to throw an exception potentially?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -454,29 +457,31 @@ public void run() {
...
     }
 
+    Status finalStatus;
     synchronized (syncObject) {
       if (node.getStatus() == Status.FAILED || this.isKilled()) {
-        return false;
+        return null;
```
HappyRay: Same as above. How do you test this error condition?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -606,18 +612,21 @@ private void runJob() {
...
   }
 
-  private void changeStatus(Status status) {
+  private Status changeStatus(Status status) {
     changeStatus(status, System.currentTimeMillis());
+    return status;
   }
 
-  private void changeStatus(Status status, long time) {
+  private Status changeStatus(Status status, long time) {
```
HappyRay: The caller has the status value already, what's the advantage of returning the same value?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: Sorry, should remove this. Added comment before the fix.

azkaban-common/src/main/java/azkaban/event/EventData.java
juhoautio: I mimicked the style of existing code like `Event.create`. As far as I'm concerned, it could be a public constructor, too.

azkaban-common/src/main/java/azkaban/event/EventData.java
juhoautio: No use case other than making it easy to see that "it's the nestedId that is being set here". I can change to have another `create` method and make both fields `final`.

juhoautio:
> We haven't seen this problem from our internal users.
> 
> I am not familiar with this part of the code yet. Could you elaborate why you got the error you described when the status is different from the value at the time of event firing?

Thanks for jumping in! The error is explained in detail in the issue,  https://github.com/azkaban/azkaban/issues/605.

Trying to explain in brief why we got the error:
- We have configured `job.notification.success` and `job.notification.failure` in our common properties, so these are pretty much enabled for all of our jobs
- Whenever a job ends, Azkaban fires an event (`Type.JOB_FINISHED`) about the state change
- Azkaban handles the notification callbacks asynchronously. The event contains a node with mutable state field, which is sometimes modified by other threads before the notification is handled.

azkaban-common/src/main/java/azkaban/event/EventData.java
juhoautio: I don't really know, does it require a comment? Previously `Node.getNestedId()` was used in some event handlers, so I'm carrying it here, as the `Node` is replaced with `EventData` in `Event`.

azkaban-common/src/main/java/azkaban/event/Event.java
juhoautio: I just modified this method signature. Code wouldn't compile if I hand't modified _all_ its callers. So I did, and made sure that it isn't called with null. If someone adds a new call to `Event.create`, they can test that it works?

I can remove the null-check here if you want, but it will be then much harder to debug where the null came from if NPE happens later in an asynchronous callback. At that point there won't be a useful stacktrace to point who created the event with a null value. What do you prefer?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: I removed `boolean shouldUpdate` from `Event.java` because it wasn't used anywhere anymore.

It seems to be a historical leftover, for example see this file from 2013 where it was originally used:
https://github.com/azkaban/azkaban/blob/0fe79f570a0bb789cbe33c20d464f3816e1b55d6/src/java/azkaban/execapp/FlowRunner.java#L522 :)

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: No. First of all, the null check is for `EventData`, not for `Status`.

More importantly, `prepareJob()` is a private method and this is the only occasion where it's used:

```java
      Status prepareStatus = prepareJob();
      if (prepareStatus != null) {
        // Writes status to the db
        writeStatus();
        fireEvent(Event.create(this, Type.JOB_STATUS_CHANGED, EventData.create(prepareStatus)));
        finalStatus = runJob();
      } else {
        finalStatus = changeStatus(Status.FAILED);
        logError("Job run failed preparing the job.");
      }
```

As you can see, if `prepareStatus` is null, it isn't even passed to `EventData`.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: Same private method (`prepareJob()`). No need for specific testing in my opinion.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: It's used like this:

```java
    if (Status.isStatusFinished(nodeStatus)) {
      quickFinish = true;
    } else if (nodeStatus == Status.DISABLED) {
      nodeStatus = changeStatus(Status.SKIPPED, time);
      quickFinish = true;
    } else if (this.isKilled()) {
      nodeStatus = changeStatus(Status.KILLED, time);
      quickFinish = true;
    }
```

So it's just to do with less lines inside the if-else blocks. Alternative would look like:

```java
    if (Status.isStatusFinished(nodeStatus)) {
      quickFinish = true;
    } else if (nodeStatus == Status.DISABLED) {
      nodeStatus = Status.SKIPPED;
      changeStatus(nodeStatus, time);
      quickFinish = true;
    } else if (this.isKilled()) {
      nodeStatus = Status.KILLED;
      changeStatus(nodeStatus, time);
      quickFinish = true;
    }
```

I find the first one easier to read.

azkaban-common/src/main/java/azkaban/event/EventData.java
fsi206914: I often see **nestedId** from fetchExecutions API. 

 "nodes" : [ {
    "nestedId" : "***",
    "in" : [ "***_a" ],
    "startTime" : 1474698890218,
    "updateTime" : 1474698890246,
    "id" : "***_b",
    "endTime" : 1474698890242,
    "type" : "command",
    "attempt" : 0,
    "status" : "SUCCEEDED"
  }, {
    "nestedId" : "****_a",
    "startTime" : 1474698890151,
    "updateTime" : 1474698890219,
    "id" : "****_a",
    "endTime" : 1474698890188,
    "type" : "command",
    "attempt" : 0,
    "status" : "SUCCEEDED"
}]

HappyRay:
Yes, I have read your issue description. 

To answer my own question earlier:

"INFO - !!!! WE ARE NOT SUPPORTING JOB CALLBACKS FOR STATUS: READY"
It's because the node status changes back to the ready state after the event has fired and no callback should be issued for ready state. Am I right?

HappyRay:
The CI build failure is a known issue. We are working on it. I restarted the build for you.

azkaban-common/src/main/java/azkaban/event/EventData.java
HappyRay: nestedid seems to be a human readable id for the node. Maybe add a comment that this value corresponds to the return value of azkaban.executor.ExecutableNode#getNestedId?

juhoautio: Added javadocs with comment about nestedId.

azkaban-common/src/main/java/azkaban/event/Event.java
juhoautio: Added javadocs & stuff to `Event.create`. I don't see how compiler could enforce it, but I put there `@NotNull` annotation, which might help some tools / IDEs detect possible violations of that constraint.

azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -0,0 +1,34 @@
...
+public class EventData {
+
+  private final Status status;
+  private final String nestedId;
+
+  public static EventData create(Status status) {
+    return create(status, null);
+  }
+
+  public static EventData create(Status status, String nestedId) {
```
juhoautio: Yeah, indeed I only did that because `Event` uses that pattern. I now modified EventData to have just public constructors.

juhoautio:
> It's because the node status changes back to the ready state after the event has fired and no callback should be issued for ready state.

Yes, it's a concurrency issue. I would approach it from a bit different angle though, because callbacks themselves are not at fault here. No event should be fired with `Type.JOB_FINISHED` that has a "not done" status attached to it. Callback handling code doesn't expect to see for example that "a job finished with status READY".

juhoautio:
I tried to fix all the issues you pointed out + synced with master. Good to go?

azkaban-common/src/main/java/azkaban/event/Event.java
```diff
@@ -53,24 +54,22 @@ public long getTime() {
...
+  /**
+   * Creates a new event.
+   *
+   * @param runner runner.
+   * @param type type.
+   * @param eventData EventData, null is not allowed.
+   * @return New Event instance.
+   * @throws NullPointerException if EventData is null.
+   */
+  public static Event create(Object runner, Type type, @NotNull EventData eventData) throws NullPointerException {
```
HappyRay: Nit: would making it a checked exception be too restrictive? I assume it doesn't for the existing callers since they have to catch exceptions already. It's a small concern though.

HappyRay:
Please fix the build error first before merging.

juhoautio:
> Please fix the build error first before merging.

Sorry.

Build fixed.

I removed that `@NotNull` annotation because I would've had to add another library dependency to keep it.

azkaban-common/src/main/java/azkaban/event/Event.java
juhoautio: Oh, but there is no checked exception. NullPointerException is a RuntimeException so doesn't require a catch block. I declared it with `throws` for clarity. Am I missing something?

HappyRay: My bad. What you have looks good.
I have approved this change. I am not sure how the new code review approval works. Could you try merging yourself?

If it doesn't work, please let me know and I can merge it for you.
Thanks!

juhoautio:
I would try merging myself, but the approval had expired because I modified the PR after that. Now I clicked "Update branch", which added merge commit from master (is that ok?). Please merge this if you still accept :) Thanks for your time and insight on this.

PR #608, opened 2016-02-11
Update README: Azkaban2 -> Azkaban 3

juhoautio:
Just to avoid confusion. The current master is version 3.0.0, so making the README reflect that.


logiclord:
Thanks

PR #609, opened 2016-02-11
Fixed errors in AJAX API documentation of schedule related actions

juhoautio:
- Easy to check that `action=scheduleFlow` and `action=removeSched` are only handled with `POST`, here: [ScheduleServlet.java#L384-L387](https://github.com/azkaban/azkaban/blob/master/azkaban-webserver/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L384-L387)
- Neither of them can be found in [ExecutorServlet](https://github.com/azkaban/azkaban/blob/master/azkaban-execserver/src/main/java/azkaban/execapp/ExecutorServlet.java) – so definitely the URL path should be `/schedule` for both.

### DIFF
--- a/_includes/docs/latest/ajax-api.html
+++ b/_includes/docs/latest/ajax-api.html
@@ -1004,10 +1004,10 @@ This API call schedules a flow.
 </p>
 <ul class="list-unstyled">
   <li>
-    <strong>Method:</strong> GET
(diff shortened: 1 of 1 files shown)


logiclord:
Thanks LGTM

PR #612, opened 2016-02-24
1. include confirmation of projectId and version after successful dow…

johnyu0520:
…nloading of file

2. adding "getGroupPermissions" and "getProxyUsers" API
3. other changes are due to reformatting


logiclord:
LGTM

PR #615, opened 2016-03-09
Enhancing Azkaban /schedule endpoint ajax API

logiclord:


johnyu0520:
Verified with Gaurav that changing from Map<String, String> to Map<String, Object> should not have downstream impact.  The reasoning is because the ultimate return format from the API is json, and also serializing the ExecutionOptions object is not an issue for the ObjectMapper

LGTM

PR #625, opened 2016-03-16
enhancing ProjectServlet: adding ajax api fetchallprojects, and fetchuserprojects

johnyu0520:
enhancing ProjectServlet: adding ajax api fetchallprojects, and fetchuserprojects


logiclord:
LGTM

PR #629, opened 2016-03-28
Add support for multiple user managers

BrianGallew:
It turns out that using the LDAP user manager for real users is ideal, but it would be nice if you could use local users (i.e. xml auth) for application accounts (e.g. automated uploads).  This patch enables specifying multiple authenticator classes (comma separated) which will all be used.  Order matters, however, in that if a user exists in more than one, the first one wins.


BrianGallew:
Can we look at merging this?

devangorder:
+1 on merging this!

BrianGallew:
I've rebased so there shouldn't be any merge conflicts now.

BrianGallew:
Is there any point in submitting PRs if one is not actively part of the Azkaban project?  This was a functional patch when I submitted it a year ago, but if you're not interested in community contributions, it's not clear that I am interested in helping out at all.

ameyamk:
Hey Brian,

Thanks for following up. Few questions:
1. Can we not use existing XML file based authentication?
2. We use headless accounts and add them to XML file for further API authentication

Finally in near-medium term we are adding a feature for all users to get secret key which can be used for API authentication. Would that solve your problem?

Since there are these features in existing system and are on road map we are hesitant to merge this in right now.

BrianGallew:
The Azkaban config file names a class which is used for authentication.  A
class.  That being the case, the only way to have multiple authentication
handlers would be to write an authentication class that encapsulates all of
the auth methods you want to use.

In my specific case, I want to use LDAP auth, but be able to fall back to
XMLAuth when LDAP isn't available (i.e. define a "root" account that's
always accessible, or an account used for publishing new flows from
Jenkins).  The patch I provided allows specification of a comma-separated
list of authentication classes, each of which is tried, in order, until
either a successful authentication occurs, OR all of the classes have
failed.  This precisely meets the use case described.  I'm sure Azkaban is
very, very special, but even a casual perusal of web-based technologies
shows that it is quite common to allow multiple authentication methods, so
the reluctance to allow this seems more than a little bizarre.

I'm not sure how this interferes with any planned features, but I can
promise you that I can't be troubled to rewrite the patch *again*.  If you
want community participation, then *you* have to, you know, "participate".

BrianGallew:
Apparently this feature is uninteresting to anyone but me and devangorder.

mvineza:
> Hey Brian,
> 
> Thanks for following up. Few questions:
> 
> 1. Can we not use existing XML file based authentication?
> 2. We use headless accounts and add them to XML file for further API authentication
> 
> Finally in near-medium term we are adding a feature for all users to get secret key which can be used for API authentication. Would that solve your problem?
> 
> Since there are these features in existing system and are on road map we are hesitant to merge this in right now.

Hi @ameyamk , I know this issue has been closed. But I'm interested on the secret key feature. Is it still in the roadmap? Seems I didn't see any information about it on recent releases. Please let me know. Thanks.

PR #635, opened 2016-04-07
Fixes "Auto Pan Zoom" option in the flow graph view.

### LINKED ISSUE #634: Auto Pan Zoom button does not work in Graph UI
When using the Job Filter in the left hand side of the flow graph, the "Auto Pan Zoom" option does not work.

Pull request to follow.

bmsq:
Fixes #634 

Selecting the "Auto Pan Zoom" button while using Job Filters in the Flow Graph View causes a javascript error.

This PR removes the "transform" variable which is the source of the issue but is never used.


fsi206914:
@bmsq May we know why removing transform assignment is able to fix this Auto Pan Zoon bug?

bmsq:
No worries.

getTransformToElement() is not a valid function for an SVG node.  When the JavaScript hits this line, it raises an error and doesn't execute the remainder of the function which prevents auto pan and zoom from working as expected.  

I was going to guess what the function was expected to do and implement a standards compliant equivalent. However, this function is called and assigned to a local variable which is redundant and never used.  The simple fix is to remove the variable and the broken call to getTransformToElement().

We auto generate a large flow containing +400 jobs with more than a 1,000 dependencies, so the search feature with auto pan and zoom is pretty critical for navigating the graph.

fsi206914:
LGTM. Thanks @bmsq for the explanation.

HappyRay:
Any reason why this one was not merged? @kunkun-tang

kunkun-tang:
This issue is already fixed in #1338 . Closing

HappyRay:
Thank @kunkun-tang

PR #647, opened 2016-04-28
	Add data pipeline and UI to show trigger status in azkaban

evlstyle:
Add data pipeline and UI to show trigger status in azkaban


evlstyle:
silent merging as we will have a massive review from master branch before the thing goes into the azkaban master.

PR #649, opened 2016-05-03
V3

inoviaazkaban:
The changes include

1) An Active Directory Connector named LdapUserManager, which allows azkaban users to login using their corporate LDAP username.
2) Some changes to azkaban-users.xml to map groups to their corresponding corporate LDAP groups
3) A file editor on the web-UI, which allows accesses the sub-filesystem on the executor server area so that an active job can can be instantly modified over the web-ui, and the results quickly seen, without the need to upload yet another zip file and re-schedule it.
4) Some fixes to failing test cases


devangorder:
any update on when this pull request can be merged? we'd love to see these features incorporated, especially the LDAP integration

devangorder:
Has anyone gotten a chance to review this yet? I built it from the fork, but haven't been able to successfully log into my LDAP yet and was curious on anyone else's experience.

@inoviaazkaban @inoviavenkat @davidzchen @hluu @logiclord

inoviavenkat:
Please mail me the Exception you see upon attempting a login.

/Venkatesh

On 21 June 2016 at 18:14, devangorder <notifications@github.com> wrote:

> Has anyone gotten a chance to review this yet? I built it from the fork,
> but haven't been able to successfully log into my LDAP yet and was curious
> on anyone else's experience.
>
> @inoviaazkaban <https://github.com/inoviaazkaban> @inoviavenkat
> <https://github.com/inoviavenkat> @davidzchen
> <https://github.com/davidzchen> @hluu <https://github.com/hluu> @logiclord
> <https://github.com/logiclord>
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/azkaban/azkaban/pull/649#issuecomment-227491127>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AQVHRfeQ6ZOu48xtB5r6SxVa25EQIrhlks5qOA3kgaJpZM4IWUdB>
> .
>



-- 
Venkatesh
Software Developer, Inovia AB
Mobile: +46 707188865

inoviavenkat:
You should provide the connection details to your ActiveDirectory/LDAP 
and some more mapping details in the file azkaban.properites, by editing the 
properties below.

user.manager.ldap.provider.url=ldap://hostname:port
user.manager.ldap.security.principal={0}
user.manager.ldap.attribute.group=group
user.manager.ldap.groups.mapping.file=conf/groups.xml
user.manager.ldap.security.authentication=simple
user.manager.ldap.initial.context.factory=com.sun.jndi.ldap.LdapCtxFactory
user.manager.class=azkaban.user.LdapUserManager

Where

user.manager.ldap.provider.url - is the protocol, hostname and port number to your LDAP

user.manager.ldap.security.principal - can be {0}, if the complete LDAP DN(Distinguished Name) is used as the login-username on the Azkaban login page. Otherwise, it can be partial LDAP DN that has {0} as its substring at an appropriate position, which will subsequently be substituted by the user-id entered on the Azkaban login page. For instance, 'CN={0},OU=Development,OU=Technology,DC=XYZ,DC=com", and {0} formatted to hold the user's login-id entered on the login-page

user.manager.ldap.attribute.group - is the LDAP attribute identifying all the LDAP group names a user belongs to

user.manager.ldap.groups.mapping.file - conf/groups.xml, the azkaban group configuration file, available at the given location in this git-commit

user.manager.ldap.security.authentication - simple, in most common configurations

user.manager.ldap.initial.context.factory  - com.sun.jndi.ldap.LdapCtxFactory, the default value in most common situations

user.manager.class - azkaban.user.LdapUserManager, the class implementing UserManager.java

@devangorder @davidzchen @hluu @logiclord

devangorder:
Thanks for posting further detail on what each of these variables should be set to. I am now able to connect to my LDAP after adding the full string to the principal value, but I am still not seeing it map LDAP groups to the Azkaban roles I have specified in my groups.xml. I think this is due to the way my LDAP is set up, and the user.manager.ldap.attribute.group value. It seems my LDAP is set up with the opposite association as what you expect; we create a base group of Users, then create Groups and within the groups note what users belong to them. If I understand your approach right, it looks like you expect the Users to be mapped to Groups via attributes of the users themselves. Below is an example of how our OpenLDAP LDIF looks to give you better context:

> ou=Users
> 	cn=person1
> 	cn=person2
> 	cn=person3
> 
> ou=Azkaban_Groups
> 	cn = azkaban_admin
> 		- uniqueMember: cn=person1,ou=Users
> 		- uniqueMember: cn=person2,ou=Users
> 	cn = azkaban_readonly
> 		- uniqueMember: cn=person3,ou=Users
> 		
> ou=AnotherApp_Groups
> 	cn = another_app_admin
> 		- uniqueMember: cn=person1,ou=Users
> 		- uniqueMember: cn=person2,ou=Users
> 	cn = another_app_readonly

Would you suggest adding 'group' attributes to each user that should have access to azkaban_admin or azkaban_readonly? Right now, anyone in the LDAP can log in with ADMIN permissions, so there is no role restriction.

inoviavenkat:
Yes, what you assumed about the LDAP here is correct; Each user in the LDAP here has attributes named 'memberOf' and the value of this attribute is the DN, the user belongs to.

And my azkaban.properties property looks like
user.manager.ldap.attribute.group=memberOf

Below is the example LDIF snippet

ou=Sales
cn = Customer1
cn = Customer2
cn = Customer3

ou=Marketting
cn = SocialMedia
cn = Promotions

ou=Users
cn=person1
- memberOf: cn=Customer1,ou=Sales
- memberOf: cn=Customer2,ou=Sales
cn=person2
- memberOf: cn=Customer1,ou=Sales
cn=person3
- memberOf: cn=SocialMedia,ou=Marketting
- memberOf: cn=Customer2,ou=Promotions


To, cater to your case, some implementation of the method
  public User azkaban.user.UserManager.getUser(String username, String password)
should be changed.

The returned object azkaban.user.User, should be constructed such that 
azkaban.user.User.getGroups(), returns List<String>, where each String represents a group. 
I populate this List<String> from the LDAP attribute memberOf, and in your case, you should
loop through all the groups the user logged in belongs to, and populate the List<String> returned
by azkaban.user.User.getGroups() with those groups.

Rest of the code can remain as is. The configuration file groups.xml could simply continue to map
LDAP groups, to corresponding azkaban permissions.

devangorder:
Thanks for the further suggestions. I reworked my LDAP based on your suggestions above, and now see the LDAP group/Azkaban role association I was going for! Now, I just have one more problem. Anyone who is in the LDAP is able to login even if they're not in the specified Azkaban groups I created. Granted, they don't have any permissions to do anything within the Azkaban web interface, but they can still view the Scheduling, Executing, and History tabs that contain job names and such. Ideally if a User is not in one of the Azkaban_Groups, they wouldn't even be able to login at all. See below for my updated ldif:

> ou=Users
> cn=person1
> \- member: cn=azkaban_admin,ou=Azkaban_Groups,dc=\<ldap name\>,dc=com
> \- member: cn=another_app_admin,ou=AnotherApp_Groups,dc=\<ldap name\>,dc=com
> cn=person2
> \- member: cn=azkaban_admin,ou=Azkaban_Groups,dc=\<ldap name\>,dc=com
> \- member: cn=another_app_admin,ou=AnotherApp_Groups,dc=\<ldap name\>,dc=com
> cn=person3
> \- member: cn=azkaban_readonly,ou=Azkaban_Groups,dc=\<ldap name\>,dc=com
> cn=person4
> 
> ou=Azkaban_Groups
> cn = azkaban_admin
> \- uniqueMember: cn=person1,ou=Users
> \- uniqueMember: cn=person2,ou=Users
> cn = azkaban_readonly
> \- uniqueMember: cn=person3,ou=Users
> 
> ou=AnotherApp_Groups
> cn = another_app_admin
> \- uniqueMember: cn=person1,ou=Users
> \- uniqueMember: cn=person2,ou=Users
> cn = another_app_readonly

So in this example, person1 and person2 can login with admin privileges as expected, as well as person3 with read only access. However, person4 can also login and navigate around the site, granted they don't have admin or full read access. Ideally person4 would not even be able to login at all. I tried setting my principal value to limit to only ou=Azkaban_Groups, but no one was able to login then. Below is what I had to set it to:

> user.manager.ldap.security.principal=cn={0},ou=Users,dc=\<ldap name\>,dc=com

Do you have any further suggestions of how I could limit login access to only those within the Azkaban_Groups (azkaban_admin, azkaban_readonly), and not everyone within Users?

inoviavenkat:
You can add the following snippet of code as a last line in the try block of the method
azkaban.user.LdapUserManager.getUser(...), and then a user who does not belong to
any valid groups, will not be allowed to login.

public User getUser(String username, String password) throws UserManagerException {
.
.
.
try {
.
.
.
//Last line of code in the try block
if(user.getGroups().size() == 0) {
	throw new AuthenticationException(username + " does not belong to any valid LDAP groups");
} // close if

} // close try
// rest of the code follows as is
...

/Venkatesh

devangorder:
I tried your suggestion above, but since we currently manage access to multiple applications through the same User base in our LDAP, users who were members of other groups were still able to login. For example, consider if person4 was actually this in the ldif above:

> cn=person4
> \- member: cn=another_app_admin,ou=AnotherApp_Groups,dc=\<ldap name\>,dc=com

Even though he's not in an Azkaban group, his group size is still greater than 0. So, I set it to the following:

> if ( ! (user.isInGroup("azkaban_admin") || user.isInGroup("azkaban_readonly")) ) {
>                               throw new AuthenticationException(username + " does not belong to any valid LDAP groups");
>                         }

and it restricted access as desired. We may rework our user bases soon to match more of the style you expect above, and have one unique user base per application (in which case user.getGroups().size() would work fine). But I shared what I did above in case any one else has their LDAP set up in a similar structure. Not sure what is the better generic solution to commit.

inoviavenkat:
Yes! I just see that the fix I provided above to restrict users like 'person4' could not have worked, since
user.getGroups(), according the current implementation returns the count of groups, the user belongs to, as in LDIF.  There are two possible ways to fix this as below, either of them should work.

**FIX 1)**
With a tiny check, we can create the user object to contain only those groups which s/he belongs to and also have some role mapping in the conf/groups.xml 

CODE_CHANGE in the private method as below

	private void addGroupsAndPermissionsToUser(DirContext dirContext, User user) throws Exception {
//..
//.. a few lines of code before you notice the while loop below
           
			while (neInner.hasMore()) {
				Object neObject = neInner.next();
				if (neObject != null) {
					String neValue = neObject.toString();
					groupName = fetchGroupCNfromDN(neValue);
					if (isValid(groupName)) {
                                                **// HERE IS THE CHECK THAT THE GROUP A USER BELONGS TO ALSO HAS SOME ROLE MAPPING**
						if(groupToPermissions.containsKey(groupName)) {
							user.addGroup(groupName);
							roleName = groupToPermissions.get(groupName);
							if (isValid(roleName)) {
								user.addRole(roleName);
							}
						}
					}
				}
			}
		}
	}

**FIX 2)**
This fix is very similar to the one I provided previously and which did not work, with a modification that
we check for user.getRoles().size() instead of user.getGroups().size(). So that we let the user login if he
has at least one azkaban role disregarding which group actually granted that role to him/her.

CODE_CHANGE as below

public User getUser(String username, String password) throws UserManagerException {
.
.
.
try {
.
.
.
//Last line of code in the try block
if(user.getRoles().size() == 0) {
throw new AuthenticationException(username + " does not belong to any valid LDAP groups");
} // close if

} // close try
// rest of the code follows as is
...

devangorder:
If I understand your two fix options above correctly, Fix 1 will only add relevant LDAP groups to a User's definition if they correspond to an Azkaban Role within the groups.xml, whereas Fix 2 returns the Azkaban Roles of a User rather than their Groups. Either seems like it would work. Fix 1 seems like a better root cause fix to me, as we only pay attention to relevant Azkaban groups rather than other groups a user may belong to in the LDAP that we don't need to know about. Would you agree?

Also, when I made my change above I noticed the " does not belong to any valid LDAP groups" message was not displayed on the webserver when I tried to login with an account in the LDAP who was not in the Azkaban groups. It was still showing "LDAP Authentication : Invalid Credentials" which is not the proper error message for this scenario, so I updated the catch below as such:

> catch (AuthenticationException e) {
>                         logger.debug(e.getLocalizedMessage());
>                         throw new UserManagerException("LDAP Authentication : " + e.getMessage(), e);
>                 }

Not sure if you have a better idea for how to display that.

inoviavenkat:
Yes, I do agree that Fix 1 is more like a root cause fix. However I cannot imagine any new feature in future which would break Fix 2 alone, or some other requirement that might break Fix 1 alone. Well, both of them work fine and I cannot pick one upon the other since I have no good reason. Lets just go with Fix 1!

The ideal way to handle the Exception and error message display would be to add a new Exception named AuthorizationException, throw this when user.getRoles().size() is 0, and handle it in a third catch(AuthorizationException ae) block. On the other hand, what you have done is also good enough. Simpler code.

devangorder:
Ah, I like the idea of throwing a separate AuthorizationException as that seems more appropriate. It would also clean up the error message a little, as with my code above it will show "Incorrect Login. LDAP Authentication : [LDAP: error code 49 - Invalid Credentials]" instead of '"LDAP Authentication : Invalid Credentials"' when you try signing in with an incorrect password. We could format these two separate error messages distinct from each other if we threw and caught different exceptions.

devangorder:
Hey @inoviavenkat, have you been able to commit the revisions we discussed above to your fork? They look good to me and work when i edit them locally in my build.

@inoviaazkaban @inoviavenkat @davidzchen @hluu @logiclord - can someone take another look at this pull request? I've been running of it for a few weeks and have not encountered any issues. I think it's ready to merge into master once someone with write access gives it a final look over. Thanks!

inoviavenkat:
No, I have not performed any of these code changes we discussed above in my local development branch, since these changes are not required for the LDAP/ldif configuration we use here. 

Also, I am not currently working on azkaban-customizations and into completely unrelated projects.
But, given that the changed have worked in your local environment, you could probably mail me your local copy of LdapUserManager.java to venkatesh.nie@gmail.com, and I can just push the file as is to this branch which I have submitted for 'Pull Request'. However, I do not know if git would automatically update this new file into this same 'Pull Request' which has already been submitted.

inoviaazkaban:
Hi,

Currently, for an azkaban-job, the user can specify a start-date and the
period-of-interval
for recurrent executions.

In addition to these, it would be a very helpful to specify an stop-date
too, past which the job would
no more be triggered/executed. The current job definition does not support
an stop-date and I am
trying to implement it (picture attached)

However, I am badly stuck with the implementation due to lack of
documentation in the code and below
are my specific questions.  Any ideas to move further from here or
connecting to me somebody who
has knowledge of these areas of code, would be much appreciated.

1) The job details entered on the web-UI (attached page) are persisted as a
blob
in the column 'data' in the database table named 'triggers'. Which java
class
reads this data further and translates them to equivalent
threads/processes. Is this
blob further used by the ExecutorServer to create some other entries in a
different table?

2) Which existing class/method is the best choice to modify with a tiny
check to ensure that
current execution time is not past the stop-date.
I am thinking it should be
*azkaban.trigger.builtin.BasicTimeChecker.eval()  *
Any better ideas?

3) I did notice that
*azkaban.scheduler.TriggerBasedScheduleLoader.scheduleToTrigger(Schedule)*
is adding translating the schedule into a Trigger, which consists of a
TriggerCondition
and an ExpiryCondition, both of them composed of
*azkaban.trigger.builtin.BasicTimeChecker.*
However, they are almost identical, differing only with the id
BasicTimeChecker_1 vs BasicTimeChecker_2.
Where are these two Condition actually called to perform the check?

Thanks in advance.
Venkatesh

+46707188865

PR #662, opened 2016-06-01
Fix #661 : Log download timeout is too short

### LINKED ISSUE #661: Log download timeout is too short
There is a hard coded timeout of 10 seconds from the start of the download of execution logs to the end. This is too short for longer logs and for connections that have longer latencies.

The suggested action item to refresh the page won't help either. 

"
The log is taking a long time to finish loading. Azkaban has stopped loading them. Please click Refresh to restart the load.
"

HappyRay:
Issue:

There is a hard coded timeout of 10 seconds from the start of the download of execution logs to the end. This is too short for longer logs and for connections that have longer latencies.

The suggested action item to refresh the page won't help either.

"
The log is taking a long time to finish loading. Azkaban has stopped loading them. Please click Refresh to restart the load.
"

Fix:
Removed the limit which is set in the client side Javascript.

Test:
Manually shortened the timeout value to 10 milliseconds to reproduce the behavior. Applied the fix and verified that the log displayed correctly.

### DIFF
--- a/azkaban-webserver/src/web/js/azkaban/model/job-log.js
+++ b/azkaban-webserver/src/web/js/azkaban/model/job-log.js
@@ -26,9 +26,6 @@ azkaban.JobLogModel = Backbone.Model.extend({
     var requestURL = contextURL + "/executor";
     var finished = false;
 
-    var date = new Date();
-    var startTime = date.getTime();
-
     while (!finished) {
       var requestData = {
         "execid": execId,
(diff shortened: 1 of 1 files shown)


georgezhlw:
looks good.

logiclord:
LGTM

HappyRay:
@wagnermarkd Thanks for the comment.

I think the refresh button can still be useful when a user wants to refresh the log display e.g. when the job is still running, or when the download was disrupted somehow.

johnyu0520:
LGTM

PR #663, opened 2016-06-01
Add getProjectId AJAX endpoint

erwa:
Also removed a `return` statement that was causing no JSON to be returned.

Internal JIRA: DSP-9239


azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -223,90 +223,89 @@ private void handleAJAXAction(HttpServletRequest req,
...
+        }
+      } else {
+        ret.put("error", "Cannot execute command " + ajaxName);
+      }
     }
 
-    if (ret != null) {
-      this.writeJSON(resp, ret);
-    }
+    this.writeJSON(resp, ret);
```
HappyRay: Is it possible that ret is null here? What happens if it is null?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -223,90 +223,89 @@ private void handleAJAXAction(HttpServletRequest req,
...
-      if (handleAjaxPermission(project, user, Type.WRITE, ret)) {
-        ajaxSetJobOverrideProperty(project, ret, req);
-      }
     } else {
-      ret.put("error", "Cannot execute command " + ajaxName);
+      ret.put("projectId", project.getId());
+      String ajaxName = getParam(req, "ajax");
+      if (ajaxName.equals("getProjectId")) {
+        // Do nothing, since projectId is added to all AJAX requests.
+      } else if (ajaxName.equals("fetchProjectLogs")) {
```
HappyRay: Would using a dispatch table be easier to read and maintain?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
erwa: `ret` won't be null because it is initialized to an empty HashMap on line 220. The null check was unnecessary.

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
erwa: Java doesn't make using dispatch tables easy because functions are not first-class objects. I could use reflection, but I think that actually makes the code harder to read and maintain.

erwa:
Removed some unused and unneeded parameters in revision 2.

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
HappyRay: Would it work using a table that maps strings to objects that implements an interface?

In any case, since this is the existing code you just moved around, your change looks fine to me.

HappyRay:
LGTM

Issue #664, opened 2016-06-03
Problem with create project when project name or description contain chinese


EdwardsBean:
Azkaban create table sql should add utf8, otherwise cannot create project when user input chinese.

PR #665, opened 2016-06-03
Fixes create project error with chinese #664

EdwardsBean:
Fixes [https://github.com/azkaban/azkaban/issues/664](url)

Add error output when create project failed.

Add create table with utf8

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -310,7 +310,7 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
       connection.commit();
 
     } catch (SQLException e) {
-      logger.error(INSERT_PROJECT + " failed.");
+      logger.error(INSERT_PROJECT + " failed.", e);
       try {
         connection.rollback();
       } catch (SQLException e1) {
--- a/azkaban-common/src/main/java/azkaban/utils/Props.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Props.java
@@ -16,16 +16,10 @@
 
 package azkaban.utils;
 
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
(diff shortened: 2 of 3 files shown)


HappyRay:
@EdwardsBean 

What tests have you done?

What happens if the table was created using the original default character set ( Is it Latin1?) and they run this code to read from that table?

Do other tables suffer from the same issue? 
How about project descriptions?

Could you include unit tests?

EdwardsBean:
@HappyRay 

Unit tests can reuse azkaban-common/../JdbcProjectLoaderTest

If the table was created using the original default character set (Latin1) and they can run this code to read from that table

When use input content contains none latin support character, mysql  will not allow to insert(SQLException).

Right now only project name and project descriptions suffer from the issue.

HappyRay:
@EdwardsBean are you referring to 
azkaban.project.JdbcProjectLoaderTest#testProjectRetrievalByFetchProjectByName ? It doesn't test non-ascII characters, does it?

The test is currently skipped for me because it requires a test DB to be setup. But the instruction is not clear how it should be set up. As the result, I don't think it runs on the CI system automatically.

===
According to 
http://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1
Only the ASCII characters are compatible between these two encodings. 
If so, there can be problems reading tables created with latin1. 
A DB update may be needed.

EdwardsBean:
@HappyRay

You are right.
When project name or project description contains non ASCII characters(project table with latin1), in azkaban project dashboard show messy code of non ASCII characters.
Changing project table to UTF8 is really necessary for those languages none western world.

McDulliBiBi:
Hi ,
I want to use Chinese words in the job files , but I found that after inserting into the mysql db , the content of the job become messy, in table project_files and project_properties. How can I solve it ? 

Thanks!

HappyRay:
Sorry about the long delay.

@EdwardsBean  Are you still interested in working on this PR? 
If you can merge in the latest code, I can consider merging it now that we have functional DB unit tests using in memory H2 DB.

@kunkun-tang @li-afaris  what do you think?

inramana:
Following up from @HappyRay's comment- @EdwardsBean do you still want to merge this? Please rebase on latest code.

PR #666, opened 2016-06-08
Fix flow alert message

juhoautio:
The message for all flow alerts used to be this:


logiclord:
LGTM

georgezhlw:
Juho, I think we could add flow status FAILED/KILLED to the error message. And it's not "seems", it's "really" :)

juhoautio:
@georgezhlw, I added another commit:

1. Add flow status to list of reasons in flow failure alert message
  - Usually the status is FAILED or KILLED
2. A bit of better message about executor _really_ not running the flow anymore

juhoautio:
@logiclord @georgezhlw how about pulling this in?

logiclord:
@HappyRay can you please take a look ?

HappyRay:
Thanks @juhoautio for the pull request!

I agree that this message is confusing. 

I assume you included the flow status so that users don't have to click the link to see if a flow has been killed or failed. 

The code is duplicated in two places. What do you think of changing the azkaban.executor.mail.DefaultMailCreator#createErrorEmail method to always include the flow status so that this extra information is included in both cases and also in other code paths that would call into this method?  

And remove the extra reasons from these two places. The last extra reason seems redundant. If a flow has failed or been killed, doesn't it imply it is no longer running?

===

Also could you make the change description a bit more descriptive like what you have in your comment? 

===
Could you include in your change description how you tested this change? Ideally a screenshot with the test result would be great.

juhoautio:
Added commit here as well, but rather have a squashed single commit -> new PR: https://github.com/azkaban/azkaban/pull/680

Issue #668, opened 2016-06-16
SlaAlertAction Error


vvii:
While trigglered Sla on a scheduled flow, I got an error message in azkaban-webserver.log
![qq20160616-1](https://cloud.githubusercontent.com/assets/1711944/16107284/060de946-33ce-11e6-8d81-34cdc2798913.png)

and I found the public method **sendSlaAlertEmail** in file azkaban-common/src/main/java/azkaban/utils/Emailer.java
doesn't call 

> message.setTLS(tls);

this would lead to above exception.

juhoautio:
Candidate fix: https://github.com/azkaban/azkaban/pull/673

juhoautio:
Fix was merged, this issue could be closed/resolved.

fsi206914:
#673 fixed. Thank you guys.

PR #669, opened 2016-06-16
Fixes issue #668: SlaAlertAction Error

vvii:


azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -93,7 +93,7 @@ private void sendSlaAlertEmail(SlaOption slaOption, String slaMessage) {
           super.createEmailMessage(subject, "text/html", emailList);
 
       message.setBody(body);
-
+      message.setTLS(tls);
```
juhoautio: I would say this is not the best fix because it still leaves other places like `SendEmailAction` broken.

See this alternative fix for the same issue: https://github.com/azkaban/azkaban/pull/673/files

vvii: Right, you did a better job if considered other bugs like this.

But, it seems no necessary to call the statement

> message.setTLS(tls)

at the public method **sendFirstErrorMessage** and **sendErrorEmail** in azkaban-common/src/main/java/azkaban/utils/Emailer.java based on your fixes.

I don't know why the previously submit did this things. maybe you can remove the statement from the method sendFirstErrorMessage and the other likely method in azkaban-common/src/main/java/azkaban/utils/Emailer.java in your fixes? 
Or just use the fixes in this issue to keep consistency with previously submit code ?

juhoautio: I think this fix should be good enough – keep it simple. Let's see if repo owners have any comments.

fsi206914:
@vvii @juhoautio   Guys, is that correct that **class SendEmailAction** is never called?

juhoautio:
I believe this can be closed, as https://github.com/azkaban/azkaban/pull/673 was merged, fixing the same problem.

fsi206914:
#673 was merged. Closing this.

PR #673, opened 2016-06-23
Fix issue #668: SlaAlertAction Error

juhoautio:
Fix broken e-mail sending (SLA and SendEmailAction)
- There was an NPE because lately added field `EmailMessage._tls` wasn't set
- This broke SLA alert e-mail sending entirely

Issue: https://github.com/azkaban/azkaban/issues/668

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
@@ -30,6 +30,7 @@ public class AbstractMailer {
   private String mailPassword;
   private String mailSender;
   private String azkabanName;
+  private String tls;
 
   private String referenceURL;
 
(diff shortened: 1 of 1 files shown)


juhoautio:
I'm going to test this, didn't get to do it yet.

juhoautio:
@HappyRay any feedback? This is quite important as SLA e-mails don't work at all in the current version.

logiclord:
LGTM

fsi206914:
Important fix!! BTW, Is that right that **class SendEmailAction** never used?

juhoautio:
@fsi206914, "fixing also `SendEmailAction`" is mostly hypothetical.

This PR is primarily to fix the SLA e-mails, but so that any other usages of `AbstractMailer` benefit as well.

PR #674, opened 2016-06-25
HadoopShell jobtype and property override documentation

logiclord:
Documenting user-facing aspects of hadoopShell jobtype and uncovering mysteries of property overrides in azkaban.

### DIFF
--- a/_includes/docs/latest/hadoopshell-type.html
+++ b/_includes/docs/latest/hadoopshell-type.html
@@ -29,7 +29,7 @@
 
 <p>For multiple commands, do it like <code>command.1, command.2</code>, etc.</p>
 
-<p>Here are some common configurations that make a <code>hadoopJava</code> job for a user:</p>
+<p>Here are some common configurations that make a <code>hadoopShell</code> job for a user:</p>
 
 <table class="table table-striped table-condensed table-bordered">
   <thead>
@@ -41,7 +41,7 @@
   <tbody>
     <tr>
       <td>type</td>
-      <td>The type name as set by the admin, e.g. <code>hadoopJava</code></td>
+      <td>The type name as set by the admin, e.g. <code>hadoopShell</code></td>
     </tr>
     <tr>
       <td>dependencies</td>
@@ -90,10 +90,3 @@
     </tr>
   </tbody>
 </table>
-
-<div class="bs-callout bs-callout-info">
-  <h4>Sample Job Package</h4>
-
-  <p>Here is a sample job package, just to show how it works:</p>
-  <a class="btn btn-primary" href="https://s3.amazonaws.com/azkaban2/azkaban2/samplejobs/command.zip" target="_blank"><span class="glyphicon glyphicon-download-alt"></span> Download command.zip</a> (Uploaded May 13, 2013)
-</div>
--- a/_includes/docs/latest/hadoopshell-type.html
+++ b/_includes/docs/latest/hadoopshell-type.html
@@ -1,0 +1,99 @@
+<h2 id="hadoopshell-type">HadoopShell Job Type</h2>
+
+<p>In large part, this is the same <code>Command</code> type. The difference is its ability to talk to a Hadoop cluster securely, via Hadoop tokens.</p>
+
+<p>The HadoopShell job type is one of the basic built-in types. It runs multiple UNIX commands using java processbuilder.  Upon execution, Azkaban spawns off a process to run the command.</p>
+
+<h3>How To Use</h3>
+
+<p>The <code>HadoopShell</code> job type talks to a secure cluster via Hadoop tokens. The admin should specify <code>obtain.binary.token=true</code> if the Hadoop cluster security is turned on. Before executing a job, Azkaban will obtain name node token and job tracker tokens for this job. These tokens will be written to a token file, to be picked up by user job process during its execution. After the job finishes, Azkaban takes care of canceling these tokens from name node and job tracker. </p>
+
+<p>Since Azkaban only obtains the tokens at the beginning of the job run, and does not requesting new tokens or renew old tokens during the execution, it is important that the job does not run longer than configured token life.</p>
+
+<p>One can run one or multiple commands within one command job. Here is what is needed:</p>
+
+<table class="table table-bordered table-striped table-condensed">
+  <thead>
+    <tr>
+      <th>Type</th>
+      <th>Command</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>command</td>
+      <td>The full command to run</td>
+    </tr>
+  </tbody>
+</table>
+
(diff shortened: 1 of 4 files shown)


_includes/docs/latest/property-override.html
```diff
@@ -0,0 +1,77 @@
...
+		<tr>
+			<td><code>{job-name}.job</code> job specification</td>
+			<td>These are user specified property in actual job file</td>
+			<td>Highest (5)</td>
+		</tr>
+	</tbody>
+</table>
+
+<p>
+	Following properties are not visible to the users. Depending on jobtype implemenation these properties are used for constraining user jobs and properties. These are the same properties which are merged to form <code>sysProps</code> in <code>AbstractProcessJob.java</code>
```
juhoautio: implemenation -> implementation

_includes/docs/latest/property-override.html
```diff
@@ -0,0 +1,77 @@
...
+		<tr>
+			<td><code>private.properties</code> in <code>jobtype/{jobtype-name}</code> directory</td>
+			<td>These are admin configured properties during Azkaban setup. Restricted to a specific jobtype.</td>
+			<td>Highest (1)</td>
+		</tr>
+	</tbody>
+</table>
+
+<p> 
+	<code>azkaban.properties</code> is another type of properties which are only used for controlling Azkaban webserver and execserver configuration. Please note that <code>jobProps</code>, <code>sysProps</code> and <code>azkaban.properties</code> are 3 different types of properties and are not merged in general (depends on jobtype implemenation).
```
juhoautio: implemenation -> implementation

_includes/docs/latest/property-override.html
```diff
@@ -0,0 +1,77 @@
...
+		<tr>
+			<th>PropertySource</th>
+			<th>Description</th>
+			<th>Priority</th>
+		</tr>
+	</thead>
+	<tbody>
+		<tr>
+			<td><code>commonprivate.properties</code> in <code>jobtype</code> directory</td>
+			<td>These are admin configured properties during Azkaban setup. Global to all jobtypes.</td>
```
juhoautio: The difference between common & commonprivate (in jobtype directory) is still a bit unclear to me.

Correct me if I'm wrong, but I've understood that in practice `common.properties` is only exposed when running an actual job. `commonprivate.properties` is also loaded when registering the jobtype class initially.

One specific case was that to allow Azkaban initialize the `command` job type with less available memory, we had to set `memCheck.enabled=false` in `commonprivate.properties`. Setting it in `common.properties` didn't help, because job type manager doesn't load it at the crucial phase. This feels pretty much like trial and error to me, but maybe there is some logical explanation to this :)

_includes/docs/latest/hadoopshell-type.html
```diff
@@ -0,0 +1,99 @@
...
+    <tr>
+      <td>command</td>
+      <td>The full command to run</td>
+    </tr>
+  </tbody>
+</table>
+
+<p>For multiple commands, do it like <code>command.1, command.2</code>, etc.</p>
+
+<p>Here are some common configurations that make a <code>hadoopJava</code> job for a user:</p>
```
georgezhlw: Should be HadoopShell

_includes/docs/latest/hadoopshell-type.html
```diff
@@ -0,0 +1,99 @@
...
+<table class="table table-striped table-bordered table-condensed">
+  <thead>
+    <tr>
+      <th>Parameter</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>hadoop.security.manager.class</td>
```
georgezhlw: Add default value for these properties.

_includes/docs/latest/hadoopshell-type.html
```diff
@@ -0,0 +1,99 @@
...
+      <td>Whether Azkaban should request tokens. Set this to true for secure clusters.</td>
+    </tr>
+  </tbody>
+</table>
+
+<div class="bs-callout bs-callout-info">
+  <h4>Sample Job Package</h4>
+
+  <p>Here is a sample job package, just to show how it works:</p>
+  <a class="btn btn-primary" href="https://s3.amazonaws.com/azkaban2/azkaban2/samplejobs/command.zip" target="_blank"><span class="glyphicon glyphicon-download-alt"></span> Download command.zip</a> (Uploaded May 13, 2013)
```
georgezhlw: This zip file is wrong. It's a normal "command" job, not a "HadoopShell" job.

HappyRay:
@logiclord looks great to me. 

Do you mind responding to other reviewers' comments so that we can merge this change soon.

Since the HadoopShell job type is similar to the Hadoop Java job type, what do you think about just referencing the Hadoop java job type documentation and only highlight the differences. This way the documentation is not duplicated and it makes future updates to the common portion easier. However there are benefits of reading one job types's configuration in one place.

Thanks!

_includes/docs/latest/hadoopshell-type.html
logiclord: I don't think that there is any default param for this. It is not a required param.

_includes/docs/latest/hadoopshell-type.html
logiclord: I have removed it for the time being to avoid confusion. I kept it as it should be similar to command and I not have access to AWS storage link.

_includes/docs/latest/hadoopshell-type.html
logiclord: Fixed. Thanks.

_includes/docs/latest/property-override.html
logiclord: Fixed. Thanks.

_includes/docs/latest/property-override.html
logiclord: Fixed. Thanks.

_includes/docs/latest/property-override.html
logiclord: Hi @juhoautio,

Yes, it is some what confusing and depends on implementation. common.properties consist of properties visible to user job (from environment variable and many other ways) while commonprivate.properties are only visible to jobtype class. For example:- HadoopPig for pig jobtype, HadoopJava for java jobtype, etc. 

To answer your other question, where to set a property is actually an implementation specific question that can only be answered by looking at code. A simple rule of thumb is to categorize props as jobProps and sysProps. For example: memcheck.enable=true required setting sysProps https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java#L76

-Gaurav

logiclord:
Hi folks.

Thanks for the feedback. I have incorporated all the feedbacks. I have kept a separate reading section of hadoopShell for easy understanding. Otherwise, it will be hard for many users to understand that HadoopShell is a wrapper on top of HadoopJava.

 Can you please have a look ?

PR #680, opened 2016-07-17
Fixes to flow notification e-mails

juhoautio:
- Removed misleading mention about "executor bounce"[1]
- Fixed e-mail subject of first error message: "failed" => "encountered a failure"
- All flow notification e-mails to include flow status (as asked by @georgezhlw and @HappyRay)
- Created a unit test for DefaultMailCreator

--

[1] One of the "Reason" messages for all flow failure alerts used to be this:

--

Screenshots:

azkaban-email-screenshot-2
azkaban-email-screenshot-1
azkaban-email-screenshot-0


juhoautio:
@HappyRay

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1447,10 +1444,7 @@ private void finalizeFlows(ExecutableFlow flow) {
...
         if (alerter != null) {
           try {
-            alerter
-                .alertOnError(
-                    flow,
-                    "Executor no longer seems to be running this execution. Most likely due to executor bounce.");
+            alerter.alertOnError(flow);
           } catch (Exception e) {
             // TODO Auto-generated catch block
             e.printStackTrace();
```
juhoautio: Could change that to `logger.error(e);` if you say so.

HappyRay: It would be great if you do. Thanks!

HappyRay:
Very nice description! Thanks for the screen shots and tests.

azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
```diff
@@ -287,4 +287,13 @@ public EmailMessage println(Object str) {
 
     return this;
   }
+
+  public String getBody() {
```
HappyRay: This method is for unit testing only, right?

How about making it package private? This way the readers would know that this is not a public API outside this package and would reduce the search space.

azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
```diff
@@ -287,4 +287,13 @@ public EmailMessage println(Object str) {
...
+    return _body.toString();
+  }
+
+  public String getSubject() {
```
HappyRay: Same comment as getBody method above.

azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -325,6 +327,18 @@ public static String formatDuration(long startTime, long endTime) {
     return days + "d " + hours + "h " + minutes + "m";
   }
 
+  public static void setCurrentTimeMillisFixed(long millis) {
```
HappyRay: This is only used for unit testing, right?
How about changing the name to something like setCurrentTimeMillisForTesting? 
Or alternatively add a comment. 

Ideally, I would prefer using a mocking framework to mock out the call to System.currentTimeMillis. However the change required would be much more significant. What you have done is an improvement. Having a test is much better than not having one!

azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -325,6 +327,18 @@ public static String formatDuration(long startTime, long endTime) {
...
+    currentTimeMillisFixed = millis;
+  }
+
+  public static void setCurrentTimeMillisSystem() {
```
HappyRay: How about a name like useSystemTime? Or add a comment explaining what this is for?

azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java
```diff
@@ -0,0 +1,121 @@
+package azkaban.executor.mail;
```
HappyRay: Thanks for adding unit tests!

azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java
```diff
@@ -0,0 +1,121 @@
...
+  private String clientPortNumber;
+  private TimeZone defaultTz;
+
+  @Before
+  public void setUp() throws Exception {
+    defaultTz = TimeZone.getDefault();
+    assertNotNull(defaultTz);
+    // EEST
+    TimeZone.setDefault(TimeZone.getTimeZone("Europe/Helsinki"));
+    Utils.setCurrentTimeMillisFixed(1468745661608L);
```
HappyRay: Does it make sense to use a constant and add a comment to explain what it represents?

azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
juhoautio: > This method is for unit testing only, right?

Yes. But no harm done if someone calls them in actual application code.

> How about making it package private? This way the readers would know that this is not a public API outside this package and would reduce the search space.

Can't use empty/default package because different packages:
`azkaban.utils.EmailMessage`
`azkaban.executor.mail.DefaultMailCreator`

In this case, I hope it's acceptable to add these getters.

azkaban-common/src/main/java/azkaban/utils/Utils.java
juhoautio: I realized that joda-time is being used in the project already so switched to use `DateTimeUtils` and got rid of the getters/setters.

juhoautio:
@HappyRay: added 3 commits according to your comments. Good to go?

HappyRay:
Looks good to me. 

Thanks for creating a new pull request to keep the change history cleaner.

Do you mind doing the same for the new changes?

It would be great if you can separate the logging change into a separate pull request since that seems to be an independent change.

juhoautio:
Here, in single commit: https://github.com/azkaban/azkaban/pull/687

I'll create a PR about ExecutorManager error logging after that's been merged, because it may conflict.

Issue #306, opened 2014-08-15
After upgrading from 2.1 I no longer see anything in the "Recently Finished" tab


jsharley:
I upgraded from 2.1 to the distributed 2.5 version and the recently finished items stopped populating.  Thinking it might be a bug in that build I pulled down the master branch and built 2.6.2.  Still broken.  Is anyone else seeing this behavior?

fsi206914:
#683  fixed.

PR #683, opened 2016-07-18
After upgrading from 2.1 I no longer see anything in the Recently Fin…

arumugarani:
…ished tab #306

### DIFF
--- a/azkaban-webserver/src/main/resources/azkaban/webapp/servlet/velocity/executionspage.vm
+++ b/azkaban-webserver/src/main/resources/azkaban/webapp/servlet/velocity/executionspage.vm
@@ -140,7 +140,7 @@
               </tr>
             </thead>
             <tbody>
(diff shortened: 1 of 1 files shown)


fsi206914:
It works! We are not familiar with velocity. @arumugarani , May I know why **recentlyFinished.isEmpty()** doesn work? and what does **!$null.isNull** mean?

Thank you!

arumugarani:
it should also work, but it should be !recentlyfinished.isEmpty :)
And more over in the place where recentlyfinished has been filled up has a
possible value of null if nothing is there to dispaly and hence I made the
same check as if "CurrentlyRunning" tab

fsi206914:
LSGM. Will merge your request soon.

arumugarani:
Please mark the bug #306 https://github.com/azkaban/azkaban/issues/306 as
closed

PR #685, opened 2016-07-19
fix create project bug

fsi206914:
This pull request revert the create project permission logic.  If lockdownCreateProjects is true, only admin is able to create project; otherwise, users (including admins) are always able to create project without being checked.


HappyRay:
Could you reference the commit that introduced the regression? 

When the setting is set, users with createproject permission can also create projects. You may want to reference the documentation about this setting in your change description.

Please include what tests you have done.

The code change LGTM.

fsi206914:
Introduction: Our team deployed the latest azkaban web server last week. we met a serious issue: every user could not create a project. After investigation, this is caused by the logic we are modifying now. we are reverting the logic back before this [pull request](https://github.com/azkaban/azkaban/pull/648/commits/d246e167e52dacd993ea1df6b7f39456f8dc9aef).

In the documentation, we already mentioned that when lockdown.create.project is true, only admins are able to create project. The latest commit ignored this, so we need to change this.

Experiments are conducted in our test cluster. New code works very good.

PR #687, opened 2016-07-20
Fixes to flow notification e-mails

juhoautio:
(Original PR: https://github.com/azkaban/azkaban/pull/680, here squashed into a single commit)

- Removed misleading mention about "executor bounce"[1]
- Fixed e-mail subject of first error message: "failed" => "encountered a failure"
- All flow notification e-mails to include flow status (as asked by @georgezhlw and @HappyRay)
- Created a unit test for DefaultMailCreator

--

[1] One of the "Reason" messages for all flow failure alerts used to be this:


HappyRay:
Thanks!

PR #598, opened 2016-01-08
Add jetty.hostname property to properly generate email links within Azkaban email notifications (even on tomcat)

sengelbert:
Simple change to add jetty.hostname property for correct email url links.

PR #688, opened 2016-07-22
enables to ovveride email links

ismailsimsek:
gives ability to override host-name and port-number in the email links. its useful especially there is nginx redirection on port number or on host-name. 

if following parameters are not used then it will use jetty parameters as before. 
mail.jetty.hostname
mail.jetty.ssl.port
mail.jetty.port


ismailsimsek:
similar to pull request 598. but with current jetty parameters i cant remap the port number when there is  nginx port redirection in between.

sjakthol:
This looks good and would definitely be useful for us (and I would guess to other users of Azkaban too). Would it be possible to get this moving forward now that it has waited for a bit over year to get attention?

HappyRay:
Sorry about the long delay. 
This seems to be useful.

The new config keys need to be defined in azkaban.Constants now. And see my comments in #598

@jamiesjc could you take a look when you get a chance?

azkaban-soloserver/src/package/conf/azkaban.properties
```diff
@@ -37,6 +37,9 @@ executor.flow.threads=30
 # mail settings
 mail.sender=
 mail.host=
+# override mail url parametters!
+# mail.jetty.hostname=dev-bi.ui.techops.eu
+# mail.jetty.ssl.port=443
```
jamiesjc: Would you mind also adding the example for "mail.jetty.port"?

ismailsimsek:
must be ok now.

following two files are removed which were conflicting, right ? 
 azkaban-soloserver/src/package/conf/azkaban.properties
 azkaban-webserver/src/package/conf/azkaban.properties

jamiesjc:
Yes, the two properties files can be removed.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -78,12 +78,12 @@ public Emailer(final Props props, final CommonMetrics commonMetrics) {
 
     if (props.getBoolean("jetty.use.ssl", true)) {
       this.scheme = HTTPS;
-      this.clientPortNumber = Integer.toString(props.getInt("jetty.ssl.port",
-          Constants.DEFAULT_SSL_PORT_NUMBER));
+      this.clientPortNumber = Integer.toString(props.getInt("mail.jetty.ssl.port", props.getInt("jetty.ssl.port",
```
HappyRay: Please define the new key names in the Constant class. This will make them easier to discover.

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -86,6 +86,11 @@
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
 
+    // mail hostname and port parametters, when this parametters set in properties file this parametters are used generate email links.
```
HappyRay: Could you describe when these settings should be used?

HappyRay: "Parameters to generate links to Azkaban pages in emails?"

HappyRay:
Are these configs for mails specifically? 
I would imagine there are other use cases where the user facing AZ url generation are needed. 
If so, can you name these parameters more generically?
I can't think of good names myself e.g. azkaban.server.hostname.external?

---
Could you help me understand the relationship between these parameters and jetty.hostname?

What's your set up? How is jetty.hostname used in your setup?

ismailsimsek:
Hi 
i needed it just for email messages, because url in the email and the url of the azkaban in-front of the reverseproxy were different that was causing broken urls in the email, in my case reverseproxy was changing url and port-number. dont know where else it can be used. agree that's better and more generic naming, i will change names to

```
jetty.hostname.external
jetty.ssl.port.external
jetty.port.external
```

https://en.wikipedia.org/wiki/Reverse_proxy#/media/File:Reverse_proxy_h2g2bob.svg

```
# mail hostname and port parametters, when this parametters set then these parametters are used to generate email links.
# if these parametters are not set then jetty.hostname, and jetty.port(jetty.ssl.port if ssl configured) are used.
# its used when there is reverse proxy between azkaban server and user, this parametters set to the values that enduser sees
# localhost:8081 -> proxy -> myazkaban:443 -> enduser
```

HappyRay:
Typo: parameters

HappyRay:
Your proposal LGTM.
---
Does it make more sense to describe it in the opposite direction?
localhost:8081 -> proxy -> myazkaban:443 -> enduser
 user -> ...
---
> # mail hostname and port parametters, when this parametters set then these parametters are used to generate email links.

Move it to the end of the description as an example of how this information may be used?

HappyRay:
The .external doesn't seem right as a subcomponent.

How about

azkaban.webserver.external_hostname
azkaban.webserver.external_ssl_port
azkaban.webserver.external_port
?

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -86,6 +86,14 @@
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
 
+    // following parameters are used when there is reverse proxy between azkaban server and user, set this parameters to the valies that enduser sees
```
HappyRay: "User facing web server configurations used to construct the user facing server URLs. They are useful when there is a reverse proxy between Azkaban web servers and users."?

typo: values

I use https://www.grammarly.com/ and intellij to check my spelling.

azkaban-solo-server/src/main/resources/conf/azkaban.properties
```diff
@@ -25,6 +25,13 @@ executor.port=12321
 # mail settings
 mail.sender=
 mail.host=
+# following parameters are used when there is reverse proxy between azkaban server and user, set this parameters to the valies that enduser sees
```
HappyRay: Thanks for adding comments here as well. 
To reduce duplication, maybe we should add a comment in the beginning of this file to refer to the Constant class for documentation?

ismailsimsek: agree its duplicated and maybe too much comment, i can remove last lines, regarding location it makes sense keeping comment or some part of it in properties file, then its will be readable after build, to the user that seting up the configuration?

ismailsimsek:
ok it should be up to date now

HappyRay:
Thanks!

PR #689, opened 2016-07-24
support postgresql as backend datastore

huangfushun:
Changes list below:
1.  force AzkabanDatabaseUpdater terminate when task is done.
2.  enable postgresql datasource when  azkaban.properties set database.type=postgresql 
3.  add unit test for postgresql related loader
4.  update sql script for postgresql table creation.
5.  upgrade dbutils to 1.6 to support insert statement return auto_generated_key
6.  fix execution_jobs query


SurjitDas:
Hi, I also need support postgresql as backend datastore for my project in the organisation, when this is going to be released? What I can see is, the stated code changes has so far, not been pushed into the lastest 3.1 tag (or am I checking in the wrong tag?). Kindly suggest.

HappyRay:
Thanks for the PR and sorry about the long delay.

Maintaining a copy of the SQL DDL statement is error prone. Since we only use mysql DB. We can't easily provide good test coverage for other DB backends. And the provided automated tests seem to require some additional set up in order to be effective. 

And the DB code has been refactored recently.

We can consider adding this support if the code is structured in such a way that maintaining it wouldn't be too hard.

huangfushun:
thanks.
it will be great if we can support this feature .
we may think about making db as a plugin that can be easily replaced. so any project especially for an exiting production system that won't or hard to setup a new mysql db.
close this pr for now.

HappyRay:
@huangfushun 
Thanks.
Another way is to try to use standard SQL features.
Do you remember the causes for treating the new DB differently other than creating the DB source diferently?

huangfushun:
hi, the most different are sql statement and db ddl.  eg. merge/limit.   if we all use standard sql then it will be easy to change to different db source.

HappyRay:
Thanks. If you can summarize these differences, it will help us move in that direction.

toby-allsopp:
@huangfushun Have you done any further work on adding PostgreSQL support? Having the Azkaban data stored in PostgreSQL would be useful for our purposes also.

@HappyRay What is your current thinking on the best way to go about adding support for another backend database type? One option is to use a library that can generate the database-specific DDL and queries - another is to use only queries that work across all database types and maintain separate DDL.

HappyRay:
> One option is to use a library that can generate the database-specific DDL and queries - another is to use only queries that work across all database types and maintain separate DDL.

We haven't evaluated the options. If the latter can work, it may be simpler. Do you have experience using such a library?

toby-allsopp:
> We haven't evaluated the options. If the latter can work, it may be simpler. Do you have experience using such a library?

No, I don't, but it sounds appealing in theory.

The approach we're most likely to attempt is to make the SQL work on both MySQL and PostgreSQL as much as possible and branch the code for cases when it's not possible. Do you think such an approach would be something you could potentially merge?

toby-allsopp:
I've created #2197 with an approach similar to that in this PR.

PR #692, opened 2016-07-25
documentation should explain the scheduletime  time zone info

fsi206914:
documentation should explain the scheduletime  time zone info

### DIFF
--- a/_includes/docs/latest/ajax-api.html
+++ b/_includes/docs/latest/ajax-api.html
@@ -1045,8 +1045,8 @@ This API call schedules a flow.
       <td>The name of the flow.</td>
     </tr>
     <tr>
-      <td>scheduleTime</td>
-      <td>The time to schedule the flow. Example: 12,00,pm,PDT</td>
+      <td>scheduleTime(with timezone)</td>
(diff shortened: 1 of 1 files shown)


_includes/docs/latest/ajax-api.html
```diff
@@ -1045,8 +1045,8 @@ <h3 id="api-schedule-a-flow">Schedule a Flow</h3>
       <td>The name of the flow.</td>
     </tr>
     <tr>
-      <td>scheduleTime</td>
-      <td>The time to schedule the flow. Example: 12,00,pm,PDT</td>
+      <td>scheduleTime(with timezone)</td>
+      <td>The time to schedule the flow. Example: 12,00,pm,PDT (Unless UTC is specified, Azkaban will take current server's default timezone instead)</td>
```
HappyRay: Do you mean the specified time zone at the end will be ignored unless it is UTC?

HappyRay:
Thanks for updating the doc!

_includes/docs/latest/ajax-api.html
fsi206914: @HappyRay 
Yes. The code logic [here](https://github.com/azkaban/azkaban/blob/master/azkaban-webserver/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L694) might be not correct.

logiclord:
LGTM

PR #694, opened 2016-07-27
Build versioning and refactor of build scripts

suvodeep-pyne:
Added plugin group: 'com.cinnober.gradle', name: 'semver-git', version: '2.2.0' to enable semantic versioning of azkaban artifacts.
Refactor: Moved all subproject build code to respective projects.


suvodeep-pyne:
The semver plugin requires Java 8 and therefore Java 7 builds are failing.

HappyRay:
@suvodeep-pyne 
Thanks for your help with the project!

Could you create an issue and associate it with this pull request?

@juhoautio do you see a problem with dropping Java 7 support?

suvodeep-pyne:
closes #695

juhoautio:
@HappyRay we use Java 8. Can't comment on behalf of other users, but personally I think it's ok to require Java 8 from now on.

.gitignore
```diff
@@ -1,11 +1,17 @@
...
 *.swp
 .classpath
 .gradle/
-.idea/
 .project
 .settings/
 /bin
 bin/
-build/
+build
```
HappyRay: Why this change? We want to ignore the build directory but not the file build if any under the project root, right?

.gitignore
```diff
@@ -1,11 +1,17 @@
...
+
+
+# Intellij files
```
HappyRay: Nice change and comment.

build.gradle
```diff
@@ -6,12 +6,15 @@ buildscript {
     }
   }
   dependencies {
+    classpath group: 'com.cinnober.gradle', name: 'semver-git', version: '2.2.0'
```
HappyRay: Nit: could you keep the style consistent? I don't have a strong opinion either way.

build.gradle
```diff
@@ -6,12 +6,15 @@ buildscript {
...
     classpath 'com.linkedin:gradle-dustjs-plugin:1.0.0'
     classpath 'de.obqo.gradle:gradle-lesscss-plugin:1.0-1.3.3'
     classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.8'
   }
 }
 
+apply plugin: 'com.cinnober.gradle.semver-git'
+apply plugin: 'idea'
```
HappyRay: Nice.

HappyRay:
@juhoautio Thanks for your comment.

HappyRay:
@suvodeep-pyne I like your comments in your change.

HappyRay:
LGTM. Let's give the community members a few more days to comment on dropping Java 7 support. I will remind myself to merge this change by next Wed. What do you think?

Could you prepare a pull request to remove java 7 test in travis-ci?

juhoautio:
> Let's give the community members a few more days to comment on dropping Java 7 support

How about mentioning that plan on the mailing list?

build.gradle
juhoautio: @suvodeep-pyne When I `Open...` the azkaban folder in IntelliJ IDEA, IDEA imports it automatically from the gradle build files. IDEA sets source & test folders and imports dependencies.

Is there some benefit from letting gradle generate the project files instead? I don't mind having the `idea` plugin here, but I hope it isn't run by default when building distTar.

juhoautio:
> The semver plugin requires Java 8 and therefore Java 7 builds are failing.

That was discussed in https://github.com/cinnober/semver-git/issues/1 and suggested:

> Please use the version 2.2.1 released in April 13, which was actually rebuilt with Java 7.

@suvodeep-pyne did you try that? Although it would be nice to drop Java 7 just to simplify the build infra and get Java 8 syntax goodies :)

.gitignore
suvodeep-pyne: Yes. This will ignore all files / directories / sub directories named "build" as opposed to just directories / sub directories named "build". Personally I haven't seen code / files added with the name "build". They are mostly auto-generated by build tools. If you recommend reverting this, i'll do it.

build.gradle
suvodeep-pyne: @juhoautio yes. IntelliJ can understand gradle projects but the support is not that great. Gradle idea plugin is much more robust in building an intelliJ project. Also, you can configure version control, etc so it becomes very easy for a new contributor to download, run "gradle idea", launch the project and start working.

Like the eclipse plugin, idea task is run only when invoked. distTar doesn't have any reason to declare idea as a dependent task. In gradle, the list of tasks to be run is constructed using a dependency task graph. Running **distTar** will invoke **build** since, **distTar** dependsOn **build** and **build** dependsOn **compilaJava** and so on and so forth. Unless a task A is dependent on task B in the task graph, B will not be invoked.

suvodeep-pyne:
Thanks @juhoautio for pointing that out! 2.2.1 works like a charm. I have updated the pull request with the changes.

@HappyRay I think we can pull this change **without** moving to Java 8. Although, I think that would be a good move. 

Please note that it would be good to create a new release tag (example 3.1.0) with this commit.
The version plugin is showing 2.7.0 as the base version which is incorrect since we have already released 3.0.0. This may be a bug which is happening due to the complex git history that azkaban has.

suvodeep-pyne:
@HappyRay 
I created issue #696 for discussing dropping Java 7 support. Let's discuss this issue there.

HappyRay:
@juhoautio Thanks for the review and suggestion.

@suvodeep-pyne 
Let's create a new release tag as you suggested.
Could you also start a discussion about dropping Java 7 support in the mailing list and reference the issue you created?

HappyRay:
I used the squash option to merge this pull request as described at
https://github.com/blog/2141-squash-your-commits

It seems to work well. I can still see the individual commits Suvodeep made and its associated comments.

The only issue I noticed was that in the conversation view, the link to the full diff is not working. But if I go to the file view, all the contexts are there. So this doesn't seem to be a big deal. This may not be an issue if I commented on the commit view instead of the file changed view.

PR #699, opened 2016-08-01
fix wrong tag name when building tar.gz

fsi206914:
fix wrong tag name when building tar.gz

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,14 +6,12 @@ buildscript {
     }
   }
   dependencies {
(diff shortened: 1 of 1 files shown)


suvodeep-pyne:
Hi @fsi206914 

I think the incorrect versioning was happening since the 3.1.0 tag wasn't an annotated tag. I fixed that. It should build with the correct version now.

Please let us know if that is not the case.

Regards
Suvodeep

fsi206914:
Closing this as already fixed.

PR #700, opened 2016-08-04
Add executor_id column to execution_flows table's creation script

binhnv:
The column executor_id is missing from execution_flows table and it caused any brand new install does not work (java.sql.SQLException: Unknown column 'ex.executor_id'). This is the change to fix the issue.
Note: The issue does not happen if it is upgrading install since the upgrade SQL script has statement to add the column.

### DIFF
--- a/azkaban-sql/src/sql/create.execution_flows.sql
+++ b/azkaban-sql/src/sql/create.execution_flows.sql
@@ -11,6 +11,7 @@ CREATE TABLE execution_flows (
 	end_time BIGINT,
 	enc_type TINYINT,
 	flow_data LONGBLOB,
+	executor_id INT DEFAULT NULL,
 	PRIMARY KEY (exec_id)
 );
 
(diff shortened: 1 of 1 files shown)


HappyRay:
@fsi206914 Did you run into this problem the last time you set up a new cluster?

fsi206914:
@binhnv Thanks for letting me know. you are correct. I believe you are setting up multi executor mode. We will look into how to adjust this.

fsi206914:
@binhnv Please use the last two files () in azkaban-sql folder. https://github.com/azkaban/azkaban/blob/master/azkaban-sql/src/sql/update.execution_flows.3.0.sql

Currently we have to keep the original sql clauses to protect backward compatibility. some people might be using older version of azksban

binhnv:
@fsi206914 How can this change break backward compatibility? Do you use this file to upgrade or install old version?

fsi206914:
A new pull request #824 was made to resolve this problem. Closing.

PR #702, opened 2016-08-04
Adding Active purge API (delete all project files except the current …

kunkun-tang:
This pull request is to help delete active projects with more than one version files. All version files except the current (max) version file will be deleted.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -302,15 +302,15 @@ public class ProjectManager {
   }
 
     /**
-     * Permanently delete all project files and properties data for all versions
-     * of a project and log event in project_events table
+     * Permanently delete all inactive project files (already deleted by the user) and properties data
+     * for all version of a project and log event in project_events table
      *
      * @param project
      * @param deleter
      * @return
      * @throws ProjectManagerException
      */
-    public synchronized Project purgeProject(Project project, User deleter)
+    public synchronized Project purgeInactiveProject(Project project, User deleter)
         throws ProjectManagerException {
         projectLoader.cleanOlderProjectVersion(project.getId(),
             project.getVersion() + 1);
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -320,6 +320,38 @@ public synchronized Project purgeProject(Project project, User deleter)
...
+   */
+  public synchronized Project purgeActiveProject(Project project, User deleter)
+      throws ProjectManagerException {
+
+    /*
+     * Check if max version matches. If the project version is equal to max of
+     * version in table project_versions, we do the purging; Otherwise, exception
+     * is thrown.
+     */
+    if (projectLoader.getLatestProjectVersion(project) == project.getVersion()) {
```
HappyRay: Maybe explain the reason for this check? This is there because of the assumptions the cleanOlderProjectVersion made, right?

azkaban-webserver/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -563,29 +566,38 @@ private void handlePurgeProject(HttpServletRequest req,
...
+        }
 
-      // only eligible users can purge a project
-      if (isOperationSuccessful && !hasPermission(project, user, Type.ADMIN)) {
-        ret.put("error", "Cannot purge. User '" + user.getUserId()
-            + "' is not an ADMIN.");
-        isOperationSuccessful = false;
-      }
+        // only eligible users can purge a project
+        if (isOperationSuccessful && !hasPermission(project, user, Type.ADMIN)) {
```
HappyRay: This logic seems to be common in both cases. Right? If so, do it in a common place?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
fsi206914: @HappyRay Did you indicate that I should state that cleanOlderProjectVersion will remove all version older than a specific version number.

HappyRay:
Is this still needed?

kunkun-tang:
Closing this PR.
I would guess #1346 made a change to the internal project file code. We would rebuild the API if we propose thie `purge active projects` API.

Issue #701, opened 2016-08-04
azkaban3.0 schedule flow bug


hitwangyu:
When I use azkaban3.0 and shedule a flow, I find a bug. The bug is as follow.
First shedule a flow with recurring period  1d.  Then I reSchedule the flow with recurring period 1h.
Then I restart webServer and I find the flow's period recover to 1d.
So I read azkaban source code and I find the cause is that updating table triggers is at the time when the flow is triggered to run. So if you change the schedule config and restart webserver before the flow's nextExecutionTime, the schedule config will recover to the old.

fsi206914:
Hi @hitwangyu , thanks for reporting this bug!! You are correct. The azkaban server seems not updating new Trigger to DB immediately when the schedule changes. I'll look into this later.

fsi206914:
Can be closed due to #708

PR #708, opened 2016-08-17
adding update trigger to db function

kunkun-tang:
This pull request is for fixing #701. After one schedule is created, if users want to modify the existing schedule, the current workaround has to recreate a new schedule request. Then updateTrigger is called, but no writes to database occurred during this process, until nextTriggerTime. I added the updating database step to fixe this bug.

Test/Experiment is conducted in our test cluster, and this fix worked.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
@@ -60,7 +60,7 @@ public class TriggerManager extends EventHandler implements
   private final Object syncObj = new Object();
 
   private String scannerStage = "";
-  
+
   public TriggerManager(Props props, TriggerLoader triggerLoader,
       ExecutorManager executorManager) throws TriggerManagerException {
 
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -156,6 +156,12 @@ public void updateTrigger(Trigger t) throws TriggerManagerException {
     synchronized (syncObj) {
       runnerThread.deleteTrigger(triggerIdMap.get(t.getTriggerId()));
       runnerThread.addTrigger(t);
+
+      try {
+        triggerLoader.updateTrigger(t);
+      } catch (TriggerLoaderException e) {
+        throw new TriggerManagerException(e);
```
HappyRay: It may be useful for debugging purposes to save the original exception as an inner exception.

HappyRay:
It would be nice to add automated tests. How hard would it be?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
fsi206914: Hi @HappyRay , thanks for your suggestion. It's still not clear to me how to implement your thoughts.

fsi206914:
Let me investigate the automated test on Scheduling service.

HappyRay:
@kunkun-tang what's your plan for this PR?

kunkun-tang:
This issue should be already fixed by #1328 . Closing now.

PR #710, opened 2016-08-18
this test missed the type classes, and fixed

fsi206914:
Was working on this JDBC Test, and found this bug.

Since we always ignored tests by inserting @ignore, no failing message show up. Actually, without this fix, the following individual tests will fail

fixed by this change.


HappyRay:
LGTM

PR #713, opened 2016-08-23
Make SVG zoom cross browser compatible

### LINKED ISSUE #712: Scrollwheel zoom on graph does not work in IE
The azkaban UI uses the non-standard "DOMMouseScroll" event which doesn't work for Internet Explorer.  The HTML standard "wheel" event should be used instead for maximum browser compatibility.  

This event has been available from Firefox 17, Chrome 31, IE 9.0 and Safari 7.

Pull request to follow.

bmsq:
Fixes #712 

Updates SVG zoom navigation to use cross compatible "wheel" event listeners rather than "DOMMouseScroll".

Verified on Firefox 47, Chrome 52 and Internet Explorer 11.


fsi206914:
@bmsq 
Thanks for reporting this bug. Do you know more context why event.wheelDelta does not work for IE? Is it due to our outdated jquery version?

bmsq:
Hi

event.wheelDelta is specific to the DOMMouseScroll event that you are using.  This is not a standard browser event.

See MDN for details:
https://developer.mozilla.org/en-US/docs/Web/Events/DOMMouseScroll

The standards compliant equivalent to this event is wheel.  This event has a slightly different structure which doesn't include event.wheelDelta but event.deltaY provides the same functionality.  

Once again, refer to MDN for details:
https://developer.mozilla.org/en-US/docs/Web/Events/wheel

As for your question regarding jQuery, the code which registers for scroll events is registering directly with the browser rather than jQuery. A quick Google suggests jQuery doesn't provide any support for mouse wheel events without a 3rd party plugin.

I hope that helps :)

HappyRay:
LGTM. Please resolve conflicts.

@fsi206914 did you test this change?

fsi206914:
@HappyRay I believe I already tested before. @bmsq , you might want to resolve the conflicts before merging.

Thanks.

bmsq:
I don't understand why conflicts were being reported, I was able to perform a git pull from upstream without any intervention.  Maybe I just don't understand how github/travis works?

Anyway, all sorted now.  @HappyRay, over to you.

HappyRay:
I tested the feature without your change on Safari, Firefox and Chrome and they all worked. Does this fix only help in IE?

HappyRay:
How do you plan to deal with the js being cached? 
@fsi206914 chose to add a date based version number to the js include statement in one of his changes to force js files to be refreshed. 

I am not too worried about this one though since using the older version seems fine for most users.

bmsq:
The older version does not work for IE. The change uses standard html mouse wheel events ensuring compatibility across all browsers that support the html standard.

I work at a number of sites that force users to use IE (why do corporates do this? Especially for supposedly "security" reasons). Without this fix, these users can't easily view large flows because they are unable to zoom in.

Does azkaban not rely on a small(ish) cache max-age and the ETag validation token? I would have thought letting the client cache expire before picking up the new version would have been fine.

PR #714, opened 2016-08-23
Track hits to Dr. Elephant which are coming from azkaban. 

rajagopr:
This change enables us to track the links originating from azkaban in piwik dashboard.

Notes:
Azkaban is https and the outgoing link to Dr. Elephant is http and hence by policy "referrer" header won't be set by the browser. An alternate way will be to set the meta tag in the relevant page. Going by the above approach since it doesn't require any code change.


akshayrai:
Great! LGTM

fsi206914:
Since this pull request is outdated, may we close this pull request?

rajagopr:
Closing...

PR #718, opened 2016-09-01
adding bower Install plugin and install moment.js, moment-timezone.js…

fsi206914:
added bower-Installer plugin to our repo, and use this plugin to update three js libraries:

bootstrap-datetimepicker.min.js
moment-timezone-with-data-2010-2020.min.js
moment.min.js


HappyRay:
LGTM. Thanks for making the code better!

What testings have you done?

Clarify in the description the benefits of this change i.e. no longer need to check in the js files?

fsi206914:
Found a glitch if bootstrap-datetimepicker.min.js is updated to new version. 

**pickDate** and **pickTime** are retired in recent bootstrap-datetimepicker.js. So I made a replace to all affected variables in order to accord with current API.

Verified  all datetimepicker locations in Repo. Should be good now. Also experimented in my local environment. Didn't find bugs.

PR #722, opened 2016-09-07
Fix bower js libraries not ready when copying js files

fsi206914:
After the previous commit #718 , bowerInstall and jsLibCopy are being executed parallel. In this pull request, we enforce jsLibCopy has to depend on bowerInstall, in order to let bower js libraries are all set before coping them.


fsi206914:
Close this due to the new pull request #723

PR #723, opened 2016-09-07
fix copyJSLibs must dependon bowerInstall (bowerInstall is not completed before copyJSLibs)

fsi206914:
Yesterday, we faced an issue that build folder (under web-server) doesn't have js files downloaded by bower plugin.

In the previous pull request commit #718 , bowerInstall and **copyJSLibs** are at the same dependency level. As a result, bowerInstall always takes longer time than **copyJSLibs**. JS lib files are often not ready before the copy. In this pull request, we enforce copyJSLibs has to depend on bowerInstall, in order to let bower js libraries are all set before coping them.

### DIFF
--- a/azkaban-webserver/build.gradle
+++ b/azkaban-webserver/build.gradle
@@ -1,3 +1,24 @@
+plugins {
+    id 'com.craigburke.bower-installer' version '2.5.1'
+}
+
+bower {
+    installBase = 'bower' // <1>
+
+    'moment'('2.14.1'){
+      source 'min/moment.min.js' >> '/'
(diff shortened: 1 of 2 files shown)


HappyRay:
LGTM
I can't merge because of the test failure though.

fsi206914:
Resolved the test failure problem in #724 . Will merge of #724 automatically make this pull request pass?

HappyRay:
You will need to rebase anyway. That should trigger another build/test.

fsi206914:
Merged.

Issue #730, opened 2016-09-12
Refactor of gradle distribution code


suvodeep-pyne:
Currently, the gradle code for building distributions of azkaban modules is hosted in the root project build script. Although this works fine in terms of building artifacts, it is not modular / easier to maintain since the expected way of using the gradle distribution plugin is to have the build code inside the submodule which it is trying to build.

For example:
The distribution code for azkaban-webserver should reside in ./azkaban-webserver/build.gradle instead of build.gradle.

Also the copying of files etc are being handled manually which is nicely supported by the DSL. This can be refactored.

PR #731, opened 2016-09-13
Refactor of gradle distribution code

suvodeep-pyne:
Currently, the gradle code for building distributions of azkaban modules is hosted in the root project build script. Although this works fine in terms of building artifacts, it is not modular / easier to maintain since the expected way of using the gradle distribution plugin is to have the build code inside the submodule which it is trying to build.

For example:
The distribution code for azkaban-webserver should reside in ./azkaban-webserver/build.gradle instead of build.gradle.

The distribution code logic has been moved into each submodule. Each sub-project is more self contained.

Also the copying of files etc are being handled manually which is nicely supported by the DSL. This has been refactored so that the distribution plugin configuration is able to take care of the copy logic

Removal of azkaban.version file. The azkaban.version file used to store the version number of the tar ball according to the git tag. Since we store the version in the file name itself as well as in manifest attributes of jar files, the version file provides a second source of truth. Thus, the code to create this file has been removed.

All bash scripts have now been moved to src/bash instead of package/bin.
All conf scripts have been moved to src/main/resources instead of package/conf. This change is motivated by the general java project structure where properties files and configuration reside in resource directories.

Transitive dependencies has been enabled since Azkaban runs as an independent system and there doesn't make much sense managing transitive dependencies manually.

Changes in build location:
Previously tar balls (*.tar.gz) and zip files were generated in build/distribution. With this change, the tar balls and zips will be generated in <subproject>/build/distribution/

Testing:
All new tar balls were diffed against previously created tar balls.
 - There are changes in jar files since some unused dependencies were dropped. Also, transitive dependencies were turned on
 - The project structure is retained. However, empty directories like extlib, plugins, etc were removed.
 - azkaban.version file was removed.


azkaban-common/build.gradle
```diff
@@ -1,10 +1,5 @@
 configurations {
-  all {
-    transitive = false
-  }
-  errorprone {
-    transitive = true
-  }
+  errorprone
```
HappyRay: What happens if this line is removed? Will this plugin use the same default config?

azkaban-common/build.gradle
```diff
@@ -25,24 +20,18 @@ model {
 
 dependencies {
   compile('com.google.guava:guava:13.0.1')
-  compile('com.h2database:h2:1.3.170')
```
HappyRay: How did you determine that these dependencies are not needed?

azkaban-common/build.gradle
```diff
@@ -25,24 +20,18 @@ model {
...
-  compile('commons-configuration:commons-configuration:1.8')
   compile('commons-dbcp:commons-dbcp:1.4')
   compile('commons-dbutils:commons-dbutils:1.5')
   compile('commons-fileupload:commons-fileupload:1.2.1')
   compile('commons-io:commons-io:2.4')
   compile('commons-lang:commons-lang:2.6')
-  compile('commons-logging:commons-logging:1.1.1')
-  compile('commons-pool:commons-pool:1.6')
   compile('javax.mail:mail:1.4.5')
   compile('javax.servlet:servlet-api:2.5')
```
HappyRay: Any suggestion on how to keep the dependencies' versions consistent across submodules?

azkaban-execserver/src/package/conf/azkaban.private.properties
```diff
@@ -1 +0,0 @@
-# Optional Properties that are hidden to the executions
```
HappyRay: Why is this line removed?

azkaban-sql/build.gradle
```diff
@@ -1,3 +1,10 @@
+apply plugin: 'distribution'
+
+/**
```
HappyRay: Thanks for adding comments.

HappyRay:
Nice change description!

azkaban-common/build.gradle
suvodeep-pyne: I commented out all the dependencies and enabled them 1 by 1. removed the rest which didn't cause compilation errors.

azkaban-common/build.gradle
suvodeep-pyne: They need not be right? If 2 submodules are independent of each, there may not be relation between the version numbers that each of them use. And also we can upgrade each individually for that matter.

If however, the submodules are dependent, then the dependent submodule will automatically import the dependencies of the parent submodule since I turned transitive dependencies on. It was disabled previously.

azkaban-common/build.gradle
suvodeep-pyne: Actually, this is no longer required. removing it.

Since the errorprone plugin is applied to all subprojects, we do not need any custom configurations for plugin integration. This was required for previous versions of the plugin / gradle.

azkaban-common/build.gradle
HappyRay: Nice. Thanks for doing this!

azkaban-webserver/build.gradle
```diff
@@ -119,56 +108,39 @@ dustjs {
...
+      from (lesscss.dest) {
+        into 'web/css'
+      }
+      from (dustjs) {
+        into 'web/js'
+      }
+      from ('bower') {
+        into 'web/js'
+      }
+      from (bowerInstall) {
```
fsi206914: How can you guarantee that bowerInstall is finished before copying?

suvodeep-pyne: see line 141. `from (bowerInstall) {`

creates an implicit dependence on bowerInstall from the distribution tasks.

fsi206914:
It looks like squash-merge in github works good. So we may have multiple commits in one pull request when updating code, right? Squash could help us rebase.

suvodeep-pyne:
Hey Liang

There are different concepts here.

Squashing is basically merging commits together. See
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Rebasing and Merging are different concepts. Although you can use the
rebase feature in git to squash commits. See
https://www.atlassian.com/git/tutorials/merging-vs-rebasing/the-golden-rule-of-rebasing

Regards
Suvodeep

fsi206914:
@suvodeep-pyne Thanks a lot your detailed explanation.

suvodeep-pyne:
@fsi206914 you are most welcome!

suvodeep-pyne:
closes #730

Issue #725, opened 2016-09-08
Deploy action through restli does not report project upload errors


KyleFung:
When requesting the server to do a deploy action through restli, failure to upload a project zip (possibly by having blacklisted .jar files) returns no errors. Currently, Azkaban simply returns the last version number of the project.

Pull request to follow.

KyleFung:
Fixed via commit [5d6d7e4](https://github.com/azkaban/azkaban/commit/5d6d7e4a36fe30e9e63b6cea027861552cde5858)

PR #733, opened 2016-09-14
Throw exception on deploy action fail. Added unit tests to make sure this behaviour happens (#725)

KyleFung:
Fixed issues mentioned in comments on last pull request to (#725). Also added unit tests.

Exception type in this action needed to be changed to RestLiServiceException because RestLi code ultimately catches them and sends a response back to the user. Before, we were throwing exception types RestLi cannot handle, so it always returned a status 500 response to the user, even if it wasn't an internal server error. If we throw a RestLiServiceException, we can specify what status code is sent to the user, in this case that is status 400.


HappyRay:
Have you checked why the ci build failed?

HappyRay:
Explain in your comments why you changed the exceptions?

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -97,22 +101,56 @@ public String deploy(@ActionParam("sessionId") String sessionId,
...
 
       logger.info("Downloaded to " + archiveFile.toString());
-      projectManager.uploadProject(project, archiveFile, "zip", user, props);
     } catch (IOException e) {
       String errorMsg =
           "Download of URL " + packageUrl + " to " + archiveFile.toString()
               + " failed";
       logger.error(errorMsg, e);
-      throw new ProjectManagerException(errorMsg, e);
+      if (tempDir.exists()) {
```
HappyRay: Is deleting the temp dir new logic?

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -97,22 +101,56 @@ public String deploy(@ActionParam("sessionId") String sessionId,
...
+    }
+
+    try {
+      // Check if project upload runs into any errors, such as the file
+      // having blacklisted jars
+      Props props = new Props();
+      Map<String, ValidationReport> reports = projectManager.uploadProject(project, archiveFile, "zip", user, props);
+      checkReport(reports);
+      return Integer.toString(project.getVersion());
+    } catch (RestLiServiceException e) {
```
HappyRay: Why catching this exception?

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -97,22 +101,56 @@ public String deploy(@ActionParam("sessionId") String sessionId,
...
+
+  public void checkReport(Map<String, ValidationReport> reports) throws RestLiServiceException {
+    StringBuffer errorMsgs = new StringBuffer();
+    for (Map.Entry<String, ValidationReport> reportEntry : reports.entrySet()) {
+      ValidationReport report = reportEntry.getValue();
+      if (!report.getErrorMsgs().isEmpty()) {
+        errorMsgs.append("Validator " + reportEntry.getKey()
+            + " reports errors: ");
+        for (String msg : report.getErrorMsgs()) {
+          errorMsgs.append(msg);
```
HappyRay: Consider making each error on its own line?

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -97,22 +101,56 @@ public String deploy(@ActionParam("sessionId") String sessionId,
...
+      throw e;
     } finally {
       if (tempDir.exists()) {
         FileUtils.deleteDirectory(tempDir);
       }
     }
-    return Integer.toString(project.getVersion());
+  }
+
+  public void checkReport(Map<String, ValidationReport> reports) throws RestLiServiceException {
```
HappyRay: Does it need to be public method?

azkaban-webserver/src/test/java/azkaban/webapp/servlet/DeployProjectTest.java
```diff
@@ -0,0 +1,126 @@
...
+    // report with warnings
+    boolean pass = true;
+    try {
+      resource.checkReport(reports);
+    } catch (RestLiServiceException e) {
+      pass = false;
+    } catch (Exception e) {
+      pass = false;
+    }
+    assertEquals(true, pass);
```
HappyRay: Consider asserting an exception is thrown directly?
http://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests

azkaban-webserver/src/test/java/azkaban/webapp/servlet/DeployProjectTest.java
```diff
@@ -0,0 +1,126 @@
...
+    ProjectManagerResource resource = new ProjectManagerResource();
+    Map<String, ValidationReport> reports = new LinkedHashMap<String, ValidationReport>();
+    for (int i = 0; i < 3; i++) {
+      Set<String> warnMsgs = new HashSet<String>();
+      ValidationReport warnRpt = new ValidationReport();
+      warnMsgs.add("test warn level info message.");
+      warnRpt.addWarningMsgs(warnMsgs);
+      reports.put("warn " + i, warnRpt);
+    }
+    // We expect that no exceptions are thrown given a
```
HappyRay: Nice comment.

azkaban-webserver/src/test/java/azkaban/webapp/servlet/DeployProjectTest.java
```diff
@@ -0,0 +1,126 @@
...
+    } catch (RestLiServiceException e) {
+      if (e.getStatus() != HttpStatus.S_400_BAD_REQUEST) {
+        pass = false;
+      } else {
+        pass = true;
+      }
+    } catch (Exception e) {
+      pass = false;
+    }
+    assertEquals(true, pass);
```
HappyRay: Consider asserting that the exception contains the expected error msgs?

azkaban-webserver/src/test/java/azkaban/webapp/servlet/DeployProjectTest.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  @Test
+  public void testWarnErrorDeploy() throws Exception {
```
HappyRay: Consider adding a test for multiple errors?

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
KyleFung: It is not new logic. Previously, if the download or validation failed for some exception, we would delete the directory if it existed.
Now, I separated the downloading part and the validation part of the project file. If there is some kind of exception during the download, we delete the file (as done before). If the download succeeds, we try to validate it. Regardless of validation failing or not, the directory will need to be deleted, so there is another directory delete under the try clause trying to validate the file.

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
KyleFung: This exception is caught so that I could be sure that an error message would be printed in the logs. However, upon further inspection, it seems that another part of the code will make a similar log entry, so I will remove this catch.

azkaban-webserver/src/restli/java/azkaban/restli/ProjectManagerResource.java
KyleFung: I agree that it should be private. If that's the case, should we keep these tests? Someone may change the implementation of this class later down the road and remove this method and these tests wouldn't work anymore.

HappyRay:
You have conflicts.

HappyRay:
Pls resolve the conflict and ensure the check passes.

KyleFung:
I messed up the merge. Please don't pull this code yet.

KyleFung:
Had a problem with some files being renamed. It is resolved now.

azkaban-web-server/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -97,22 +101,51 @@ public String deploy(@ActionParam("sessionId") String sessionId,
...
+      throw e;
     } finally {
       if (tempDir.exists()) {
         FileUtils.deleteDirectory(tempDir);
       }
     }
-    return Integer.toString(project.getVersion());
+  }
+
+  void checkReport(Map<String, ValidationReport> reports) throws RestLiServiceException {
```
suvodeep-pyne: nit: checkReports seems more apt. You are taking a collection as an argument as opposed to a single Report object

azkaban-web-server/src/test/java/azkaban/restli/DeployProjectTest.java
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      resource.checkReport(reports);
+    } catch (RestLiServiceException e) {
+      //Ensure we have the right status code and exit
+      assertEquals(e.getStatus(), HttpStatus.S_400_BAD_REQUEST);
+      assertEquals(e.getMessage(), "Validator error  reports errors: This should show up"
+          + System.getProperty("line.separator"));
+    }
+  }
+}
```
suvodeep-pyne: nit: newline missing

azkaban-web-server/src/test/java/azkaban/restli/DeployProjectTest.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+
+public class DeployProjectTest {
+  @Test
+  public void testWarnedDeploy() {
+    ProjectManagerResource resource = new ProjectManagerResource();
+    Map<String, ValidationReport> reports = new LinkedHashMap<String, ValidationReport>();
+    for (int i = 0; i < 3; i++) {
+      Set<String> warnMsgs = new HashSet<String>();
+      ValidationReport warnRpt = new ValidationReport();
```
suvodeep-pyne: Nice job with the tests!

While this works. Consider having a simple method for created the kind of mock objects that you need for all your testing.

All your tests require creating ValidationReport objects with warn / error messages so why not create a simple method that gives you that with ease. 
Example: 
`
ValidationReport validationWarn(String msg) {
  // return new created ValidationReport object with the msg. 
}
  ValidationReport validationError(String msg);
`

This results in lesser code duplication plus makes it more maintainable in case you need to add some other field later on.

Make your tests as simple / small as possible and extract such constructs out so that adding new cases is a very minimal effort.

PR #734, opened 2016-09-15
Moving Azkaban Hadoop Security code from azkaban-plugins to this repo

suvodeep-pyne:
The current changes consolidates azkaban-plugins/hadoopsecuritymanager-yarn and azkaban-plugins/hadoopsecuritymanager-common into a single submodule azkaban-hadoop-security-plugin in this repo. Following this commit, the Azkaban hadoop security code in the azkaban-plugins will be deprecated but will remain as is. However, future deployments of Azkaban will use this module as opposed to the one in azkaban-plugins.

Motivation: Moving the plugin code into the main repo helps in consolidation. The azkaban-plugins repo is not maintained well. It currently uses hardcoded and uploaded dependencies in the repo which is not a good model for managing artifacts. The new module works with gradle and pulls dependencies using ivys. This allows easier management of the hadoop security code and also makes making changes much easier by removing cross project dependencies.

azkaban-hadoop-security-plugin: The plugins hadoopsecuritymanager-commmon and hadoopsecuritymanager-yarn code has been merged into this single project. The Hadoop 1 management code has not been imported. We will need to rethink the structure a bit if we do have a requirement to maintain multiple hadoop branches. However, since that isn't the case now. I'm suggesting a simplified flat structure for the project.

Code Changes: Very minimal code changes have been made to make the code compile. No changes in code logic. The only changes are wrapping toString() calls on byte arrays to Arrays.toString(). This is flagged as an error by the errorprone library.

Testing: The build succeeds with the required artifacts. There are almost zero code changes which makes the risk of this plugin failing to be extremely low.

Downside of this change: Since the hadoop-security-plugin depends on hadoop and hive. (Yes ! it depends on hive too!) Every time we change hadoop versions we may need new artifacts. I still don't consider these to be a big problem because Hadoop interfaces rarely change. And even if they do, firing a new version of Azkaban is not that big a deal. However, if we want to maintain compatibility with different versions of hadoop at the same time, then it could be tricky.

### DIFF
--- a/azkaban-hadoop-security-plugin/build.gradle
+++ b/azkaban-hadoop-security-plugin/build.gradle
@@ -1,0 +1,27 @@
+apply plugin: 'distribution'
+
+ext.hadoopVersion = "2.6.1"
+
+dependencies {
+  compile project(":azkaban-common")
+
+  compile "org.apache.hadoop:hadoop-common:$hadoopVersion"
+  compile "org.apache.hadoop:hadoop-mapreduce-client-common:$hadoopVersion"
+  compile "org.apache.hadoop:hadoop-mapreduce-client-core:$hadoopVersion"
+  compile "org.apache.hive:hive-metastore:0.13.1"
+}
+
+/**
+ * Just package the jar.
+ * Since, rest of the dependencies are just hadoop and hive. They are not packaged inside the plugin.
+ * It is assumed that classpaths of hadoop, hive, pig, etc will be externally fed into the application.
+ */
+distributions {
+  main {
+    contents {
+      from(jar) {
+        into 'lib'
+      }
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


HappyRay:
"the hadoop security code to managed" do you mean " to be managed?"

HappyRay:
I wonder if we should remove the old code in the plugin repo instead to avoid confusion. Let's discuss off line.

HappyRay:
Consider opening an issue that describes your overall plan for the plugin repo and link to this commit?

HappyRay:
Explain why you combined these two projects and why you left out the hadoopsecuritymanager project?

HappyRay:
Thanks for taking on this long overdue tech debt clean up effort!

HappyRay:
Other than the comments, LGTM.

HappyRay:
How did you test this change?

suvodeep-pyne:
@HappyRay 
I created #736 to capture the entire effort. I also updated the PR description. I haven't pulled Hadoop 1 code from the repo. Do you think discarding the 3 projects makes sense?

I just wanted to point out there are no tests for this code. It has been copied as is. It builds correctly and generates the necessary artifacts. We will need to add tests as we work on this.

PR #742, opened 2016-09-19
refactor js code in schedule-panel js

fsi206914:
I'm working on the Azkaban Javascript Test initiative. By going through other standard front end projects and javascript lesson make ups, I know that 
1) .this part of code should be placed in **util/model**, for ease of test. 
2). schedule-panel.js is a **view** js, and it should only include DOM operations.
3). moving this cron-util code to date.js, since date.js is a util library file in Azkaban. schedulepanel.vm already imports date.js

Tested locally, and it should all work

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/schedule-panel.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/schedule-panel.js
@@ -186,28 +186,6 @@ var cron_output_id  = "#cron-output";
 var cron_translate_id  = "#cronTranslate";
 var cron_translate_warning_id  = "#translationWarning";
 
-// Uni Cron use 0-6 as Sun--Sat, but Quartz use 1-7. Due to later.js only supporting Unix Cron, we have to make this transition.
-// The detailed Unix Cron Syntax: https://en.wikipedia.org/wiki/Cron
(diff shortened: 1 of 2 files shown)


HappyRay:
Please clarify what "this" means in your commit description.

azkaban-web-server/src/web/js/azkaban/view/schedule-panel.js
```diff
@@ -217,6 +195,9 @@ function updateOutput() {
 
 function updateExpression() {
   $('#nextRecurId').html("");
+
+  //Unix Cron use 0-6 as Sun--Sat, but Quartz use 1-7. Due to later.js only supporting Unix Cron, we have to make this transition.
+  //transformFromQuartzToUnixCron is defined in util/date.js
```
HappyRay: The same comment is in the date.js already. What do you think about removing this to make the comment easier to maintain?

HappyRay:
LGTM

azkaban-web-server/src/web/js/azkaban/view/schedule-panel.js
fsi206914: I intended to place it here, in order to let readers know why we need this transition. What do you think?

PR #746, opened 2016-09-21
Hadoop security hole fix

suvodeep-pyne:
Hadoop Tokens retrieved by Azkaban for running user processes were left world readable in the temp directory. This poses a major security threat since anyone can fish tokens from the tmp directory and use them to authenticate with hadoop using Azkaban.

This has been fixed by ensuring that the tokens created by Azkaban are readable only to the specific user. They are left readable/writable by the azkaban group which is helpful for performing cleanup tasks later on.

Code Changes: The Hadoop 2 Security Manager after fetching the token now, restricts permissions on the file and changes the file ownership to the specific user executing the job.

Testing: Build succeeds. This has been manually tested on a Kerberos enabled Hadoop Cluster. Most Acceptance tests pass. Those which fail are due to time outs for unavailable services.

Note: This has been only fixed in Hadoop 2. For deployments, azkaban-hadoopsecuritymanageryarn jar has been deprecated and should not be used. please use the azkaban-hadoop-security-plugin jar instead.


HappyRay:
Could you please explain why automated test is not included?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.apache.log4j.Logger;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+
+/**
+ *
+ * Created by spyne on 9/20/16.
```
HappyRay: We have source control, do we still need the author name and date here?

suvodeep-pyne: removed it. I have it set up in my IDE and it adds by default. will fix that.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return exitCode;
+  }
+
+  private String constructExecuteAsCommand(String user, String command) {
+    return String.format("%s %s %s", path, user, command);
+  }
+
+  /**
+   * TODO Refactor: function copied from azkaban.jobExecutor.ProcessJob#partitionCommandLine(java.lang.String)
```
HappyRay: Nice. It would be even better to link to a ticket.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * Need to be refactored to use just one method. Copy was created to avoid introducing dependencies to
+   * azkaban-common
+   *
+   * Splits the command into a unix like command line structure. Quotes and
+   * single quotes are treated as nested strings.
```
HappyRay: Add an example?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param command
+   * @return
+   */
+  public static String[] partitionCommandLine(final String command) {
```
HappyRay: Is there a library method that can do the same?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -125,11 +134,13 @@
...
   private static URLClassLoader ucl;
 
-  private final RecordFactory recordFactory = RecordFactoryProvider
-      .getRecordFactory(null);
+  private final RecordFactory recordFactory = RecordFactoryProvider.getRecordFactory(null);
+  private final ExecuteAsUser executeAsUser;
 
   private HadoopSecurityManager_H_2_0(Props props)
       throws HadoopSecurityManagerException, IOException {
+    String executeAsUserBinaryFilePath = String.format("%s/%s", props.getString(NATIVE_LIB_FOLDER), "execute-as-user");
```
HappyRay: Is it worth definging execute-as-user as a constant?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -804,6 +771,77 @@ private void getToken(String userToProxy) throws InterruptedException,
...
+   *  Writes credentials to a token file and sets permissions such that only the user being proxied has read access
+   *  to the file
+   *
+   * @param user user to be proxied
+   * @param credentials Credentials to be written to file
+   * @param tokenFile file to be written
+   * @param logger logger to use
+   * @throws IOException
+   */
+  private void prepareTokenFile(final String user,
```
HappyRay: Nice use of the final in many places.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -804,6 +771,77 @@ private void getToken(String userToProxy) throws InterruptedException,
...
+                                final Credentials credentials,
+                                final File tokenFile,
+                                final Logger logger) throws IOException {
+    writeCredentialsToFile(credentials, tokenFile, logger);
+    try {
+      assignPermissions(user, tokenFile, logger);
+    } catch (IOException e) {
+      // On any error managing token file. delete the file
```
HappyRay: Nice defensive programming.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -804,6 +771,77 @@ private void getToken(String userToProxy) throws InterruptedException,
 
   }
 
+  /**
+   * Prepare token file.
+   *  Writes credentials to a token file and sets permissions such that only the user being proxied has read access
```
HappyRay: Technically the azkaban user has read access too. 

If you would prefer avoiding duplicating comments, may be point to the comment of the other method?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -804,6 +771,77 @@ private void getToken(String userToProxy) throws InterruptedException,
...
+   * Step 1. Set file permissions to 460. Readable to self and readable / writable azkaban group
+   * Step 2. Set user as owner of file.
+   *
+   * @param user user to be proxied
+   * @param tokenFile file to be written
+   * @param logger logger to use
+   */
+  private void assignPermissions(String user, File tokenFile, Logger logger) throws IOException {
+    final String changePermissionsCommand = String.format("chmod 460 %s", tokenFile.getAbsolutePath());
+    int result = executeAsUser.execute(System.getProperty("user.name"), changePermissionsCommand);
```
HappyRay: Instead of partitioning the argument string, have you considered passing an array or something that is easier to parse? Python subprocess module uses that approach.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -804,6 +771,77 @@ private void getToken(String userToProxy) throws InterruptedException,
...
+    if (result != 0) {
+      throw new IOException("Unable to modify permissions. User: " + user);
+    }
+
+    final String changeOwnershipCommand = String.format("chown %s:azkaban %s", user, tokenFile.getAbsolutePath());
```
HappyRay: This assumes that the user running this process is "azkaban". It's probably a safe assumption. It would be better if this is documented somewhere. 

Should 'azkaban' be a constant?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ *
+ * Created by spyne on 9/20/16.
+ */
+public class ExecuteAsUser {
+  private final static Logger log = Logger.getLogger(ExecuteAsUser.class);
+
+  private final String path;
+
+  public ExecuteAsUser(final String path) {
```
suvodeep-pyne: I actually updated the API itself.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
```diff
@@ -0,0 +1,116 @@
...
+    Process process = new ProcessBuilder()
+        .command(partitionCommandLine(constructExecuteAsCommand(user, command)))
+        .inheritIO()
+        .start();
+
+    int exitCode;
+    try {
+      exitCode = process.waitFor();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
suvodeep-pyne: yes. thanks for catching that. fixed it.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
suvodeep-pyne: Do you think creating a ticket for every TODO is a good idea? I am thinking it to be an overhead as opposed to creating simple TODOs.

It is a plus since, it's a quick recommendation to anybody who is modifying the code as opposed to creating a separate ticket for modifying the code.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
suvodeep-pyne: Updated it. I didn't modify any piece of that code. Hence I didn't bother to add tests. Moreover, I think this piece of code should be handled by a library as opposed to doing this by hand. I don't find it to be maintainable.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
suvodeep-pyne: exactly my point. I think library utilities should exist to do the job. That's why I added a todo to fix that. The same logic is used in 2 places.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: Actually I updated the API to have this logic inside it. My plan is to have ExecuteAsUser as the default way of leveraging the execute-as-user binary.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: Thanks for pointing it out. updated it.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: I actually like this idea. This simplifies things a lot. I'll implement this and update the pr.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: Actually the user is not assumed and is derived. However, this "azkaban" is the group. I introduced constants for the purpose but I think this should a configurable parameter.

build.gradle
```diff
@@ -44,6 +44,10 @@ subprojects {
         )
       }
     }
+
+    dependencies {
+      testCompile('junit:junit:4.12')
+    }
```
fsi206914: This is good improvement!

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
HappyRay: I recognize that creating a ticket adds overhead. So use your judgement.

The benefit of creating a ticket is to remind / force us to follow through. 

If the plan is to address the todo items shortly, a ticket would be required anyway.  If we don't intend to address them soon, I agree the overhead of a ticket may be relatively high.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/ExecuteAsUser.java
HappyRay: I must have missed your comment that this is code copied from the existing code.

build.gradle
HappyRay: Is it needed for this pull request?

In the future, what do you think of keeping different changes in separate commits?

HappyRay: This change seems to be related to the unit test you added. Right? If the change is small, maybe mention it in the change description if you want to avoid the overhead of making another change?

HappyRay:
LGTM.

Maybe add to the description as to why no automated tests are included in this pull request?

Issue #696, opened 2016-07-29
Stop support for Java 7


suvodeep-pyne:
Creating this issue to receive feedback from community regarding discontinuing Java 7 support for azkaban. Java 8 has quite a few syntax goodies and it's been a while since it has been launched. If people feel they have strong reasons to continue to support Java 7, let us discuss that here.

suvodeep-pyne:
This has already been fixed.

PR #747, opened 2016-09-21
Stopping support for Java 7. Only supported Java version is Java 8

suvodeep-pyne:
This patch will disable travis build tests for Java 7 and retain only java 8 tests. The motivation is to reduce support load and also incorporate a bunch of syntax goodies and features supported by Java 8.

Since build already succeeds / is deployed for Java 8. No further testing is required.

Also, updated Gradle IntelliJ configuration to use Java 8 language level for the project.


suvodeep-pyne:
closes #696

fsi206914:
LGTM!

HappyRay:
I noticed that you made some other improvements as part of this pull request as well.

It would be great if you can mention these changes in the change description.

Maybe consider mentioning that the proposal to retire java 7 support has been published in the associated issue for over a month now and we haven't received any objection. 

LGTM

PR #750, opened 2016-09-23
Removing non functional distribution tasks for azkaban-test

suvodeep-pyne:
The distribution tasks for azkaban-test were pointing to an invalid directory and were not resulting in any artifacts. Removing this dead code. The zip / tar archives are not used anywhere else in the project.


suvodeep-pyne:
@fsi206914 Hey Liang, travis build fails a bunch of times for node js. Is this a known issue?
On restart, it works fine.

fsi206914:
I even don't realize we have this folder! We should remove it. LGTM!

HappyRay:
LGTM
It appears that this project is not complete. The test resources may still be useful?

suvodeep-pyne:
The test resources are being used. It's just the distribution tasks which
are not. Also, if you notice the distribution job configuration in the
gradle script, the file locations in the script are incorrect. That's the
reason it doesn't generate those artifacts. And following that our internal
systems complain about missing artifacts.

That's why I want to delete them.

PR #756, opened 2016-09-28
assmble js Test initiative

fsi206914:
Travis works for this change.

During this change:
1). replace bower plugin to general nodejs plugin to support both JS download and Test Evaluation.
2). wrap up mocha component, and add one unit Test to evaluate Transfromation from Unix Cron to Quartz.
3). Enforce gradle to download nodejs every time, in order to resolve _npm can not be found_ issue.
4). Bring up rewire.js in order to be able to get local function of javascript file.
5). enforce clean temporary files during **gradle clean**


fsi206914:
@suvodeep-pyne Tools guy taught me how to debug gradle in IDEA. The root cause of npm_install failling in IDEA is that $PATH is not right in IDEA. I enforce gradle always downloadin the specific nodejs version, and resolved.

suvodeep-pyne:
@fsi206914 Thanks for sharing that.

fsi206914:
@suvodeep-pyne Do you have any comment for this pull request? I noticed that our pull request sometimes failed for some reason. This change enforced downloading a specific node js version, might solve that issue.
@HappyRay

HappyRay:
"Enforce gradle to download nodejs every time, in order to resolve npm can not be found issue." Is this going to slow down the build process? Is there no other way? How did others solve the same problem?

HappyRay:
@suvodeep-pyne could you take a look?

fsi206914:
@HappyRay I didn't observe slowness in my local laptop.
Also, it should be one-off process. Unless you do gradle clean, you don't need to download after the first build.

azkaban-web-server/build.gradle
```diff
@@ -12,7 +12,28 @@ buildscript {
 }
 
 plugins {
-    id 'com.craigburke.bower-installer' version '2.5.1'
+  id "com.moowork.node" version "0.13"
+}
+
+node {
```
suvodeep-pyne: Hey @fsi206914 
Could you please add some comments about the role of this plugin and how's it being used.

azkaban-web-server/build.gradle
```diff
@@ -50,27 +71,55 @@ sourceSets {
...
-bower {
-  installBase = 'bower' // <1>
+task movingJsTojsToPackage(dependsOn: ['npm_install']) {
 
-  'moment'('2.14.1'){
-    source 'min/moment.min.js' >> '/'
+  doLast {
+    copy {
+      from "node_modules/later/later.min.js"
+      into 'build/jsToPackage'
```
suvodeep-pyne: use $buildDir instead of hardcoding build.
You can either use buildDir + '/foo' or "$buildDir/foo"

Same for the ones below.

azkaban-web-server/build.gradle
```diff
@@ -50,27 +71,55 @@ sourceSets {
...
+    }
   }
+}
 
-  'moment-timezone'('0.5.5') {
-    source 'builds/moment-timezone-with-data-2010-2020.min.js'  >> '/'
-  }
+task cleanAll {
+  delete '.gradle'
```
suvodeep-pyne: You don't need to delete hidden directories like .gradle. They are used by gradle to manage different gradle versions and dependencies.

azkaban-web-server/build.gradle
```diff
@@ -50,27 +71,55 @@ sourceSets {
...
+task movingJsTojsToPackage(dependsOn: ['npm_install']) {
 
-  'moment'('2.14.1'){
-    source 'min/moment.min.js' >> '/'
+  doLast {
+    copy {
+      from "node_modules/later/later.min.js"
+      into 'build/jsToPackage'
+    }
+    copy {
```
suvodeep-pyne: Is this supposed to be shipped in the tar ball. Then this should be part of the distribution configuration. You may not need a 2 step copy

azkaban-web-server/package.json
```diff
@@ -0,0 +1,24 @@
+{
```
suvodeep-pyne: What is this json used for ?

azkaban-web-server/build.gradle
fsi206914: fixed.

azkaban-web-server/package.json
fsi206914: This json is called by npm_install. Will add a comment inside the json.

azkaban-web-server/build.gradle
fsi206914: Fixed. I noticed $buidDir is only recognized by double quote. Single quote can not identify $buildDir, do you know why?

azkaban-web-server/build.gradle
fsi206914: Added.

azkaban-web-server/build.gradle
fsi206914: Yes. I considered moving copyJS to Distribution Task. I was worried that would make Distribution Task looks Ugly.

azkaban-web-server/build.gradle
suvodeep-pyne: yes. you are correct. That's the reason I pointed it out.
Symbols inside double quoted strings are resolved by gradle while single quoted string is not. This comes from Groovy and you can read more about it: http://groovy-lang.org/syntax.html#_double_quoted_string

azkaban-web-server/build.gradle
suvodeep-pyne: I understand but it totally makes sense to me since the distribution task is the one place holder to gather all that is required in the package. This would avoid having copy tasks in different places.

This may increase the size of the distribution config but that's ok :)

azkaban-web-server/package.json
suvodeep-pyne: I am afraid you won't be able to create comments inside json files. However, I just wanted to ask if this needs to be in the root directory of the subproject? 

I mean, we generally keep json / resource files in resources etc, but this seems to be a package configuration. I am not that aware about nodejs so I'll let you decide whether this needs to reside in the root level of the project. What are the best practices for this ?

azkaban-web-server/build.gradle
```diff
@@ -11,8 +11,32 @@ buildscript {
...
+
+  // Version of npm to use.
+  npmVersion = '3.9.3'
+
+  // Base URL for fetching node distributions (change if you have a mirror).
+  distBaseUrl = 'https://nodejs.org/dist'
+
+  // If true, it will download node using above parameters.
+  // If false, it will try to use globally installed node.
+  download = true
```
suvodeep-pyne: nice comments ! really helps ! 👍

HappyRay:
LGTM. Thanks for adding the first JS unit test. With your work, adding more JS unit tests will be easier.

Thanks @suvodeep-pyne for the through review.

fsi206914:
Thanks a lot for @suvodeep-pyne 's reviews. Learned a lot.

PR #758, opened 2016-09-29
Add support for upstream gateways or proxies in front of Azkaban web server

mradamlacey:
The following is a pull request to add support in Azkaban's web server for upstream proxies or gateways that mangle the 'real' client IP address.

The scenario where:

client > gateway > Azkaban web server

Can cause issues if the gateway is load balanced (and therefore has different IP addresses) or uses a connection pool to the Azkaban web server as this will cause subsequent requests to have a different Client IP/port.

The pull request uses the `X-Forwarded-For` header if provided, and strips off the port from the client IP address used for session validation.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/WebUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/WebUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012 LinkedIn Corp.
+ * Copyright 2016 LinkedIn Corp.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -17,6 +17,7 @@
 package azkaban.utils;
 
 import java.text.NumberFormat;
+import java.util.Map;
 
 import org.joda.time.DateTime;
 import org.joda.time.DurationFieldType;
@@ -32,6 +33,8 @@ public class WebUtils {
   private static final long ONE_MB = 1024 * ONE_KB;
   private static final long ONE_GB = 1024 * ONE_MB;
   private static final long ONE_TB = 1024 * ONE_GB;
+
+  public static final String X_FORWARDED_FOR_HEADER = "X-Forwarded-For";
 
   public String formatDate(long timeMS) {
     if (timeMS == -1) {
@@ -163,5 +166,42 @@ public class WebUtils {
       return nf.format(sizeBytes / (double) ONE_KB) + " kb";
     else
       return sizeBytes + " B";
+  }
+
+  /**
+   * Gets the actual client IP address inspecting the X-Forwarded-For
+   * HTTP header or using the provided 'remote IP address' from the
+   * low level TCP connection from the client.
+   *
+   * If multiple IP addresses are provided in the X-Forwarded-For header
+   * then the first one (first hop) is used
+   *
+   * @param httpHeaders List of HTTP headers for the current request
+   * @param remoteAddr The client IP address and port from the current request's TCP connection
+   * @return The actual client IP address
+   */
+  public String getRealClientIpAddr(Map<String, String> httpHeaders, String remoteAddr){
+
+    // If some upstream device added an X-Forwarded-For header
+    // use it for the client ip
+    // This will support scenarios where load balancers or gateways
+    // front the Azkaban web server and a changing Ip address invalidates
+    // the session
+
+    String clientIp = httpHeaders.getOrDefault(X_FORWARDED_FOR_HEADER, null);
+    if(clientIp == null){
+      clientIp = remoteAddr;
+    }
+    else{
+      // header can contain comma separated list of upstream servers - get the first one
+      String ips[] = clientIp.split(",");
+      clientIp = ips[0];
+    }
+
+    // Strip off port and only get IP address
+    String parts[] = clientIp.split(":");
+    clientIp = parts[0];
+
+    return clientIp;
   }
 }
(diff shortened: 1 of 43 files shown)


azkaban-web-server/src/restli/java/azkaban/restli/ResourceUtils.java
```diff
@@ -56,4 +59,30 @@ public static User getUserFromSessionId(String sessionId, String ip)
...
+    // This will support scenarios where load balancers or gateways
+    // front the Azkaban web server and a changing Ip address invalidates
+    // the session
+    Map<String, String> headers = context.getRequestHeaders();
+
+    String clientIp = headers.getOrDefault("X-Forwarded-For", null);
+    if(clientIp == null){
+      clientIp = (String) context.getRawRequestContext().getLocalAttr("REMOTE_ADDR");
+    }
+    else{
```
HappyRay: Could you think of a way to reduce the code logic duplication in this method and the other one?

mradamlacey:
How do you execute unit tests for the project?

Not that familiar with gradle (sorry).  

`./gradlew test`

Does that do the trick?

Also, for unit tests I'm thinking of just testing a utility class that implements the logic to parse out the client ip...  I don't think it's worth the effort to spin up the test infrastructure for a servlet which would be more of an integration test.  Let me know if you suggest another route.

suvodeep-pyne:
yes. You can execute tests with `./gradew test`. `./gradlew build` will run tests anyway and build the entire project.

Unit tests are good as long as they capture your changes. We expect the tests to implicitly document the expectation from your change. Of course additional context helps people understand the code / logic better.

mradamlacey:
Updated the pull request with the following:

- Unit tests for ResourceContextUtils class (low level helper used to parse out the client IP)
- Unit tests around AbstractLogin servlet - with necessary mocking to get a request to be successfully processed

I didn't refactor the 'parse out the client IP and compare to session IP' bit, as the servlet vs. the REST service implementation was very different - I didn't see a ton of value having a common library that all it does compare two strings - would still need the calling context to decide to where to pull the strings from.

Let me know your thoughts or what else you feel is needed for this PR.

azkaban-web-server/src/restli/java/azkaban/restli/ResourceUtils.java
HappyRay: What about passing the client IP ( the last hop remote address ) into a common method? One call will be a value from req.getRemoteAddr() and another will be from context.getRawRequestContext().getLocalAttr("REMOTE_ADDR");

azkaban-web-server/src/test/java/azkaban/fixture/MockLoginAzkabanServlet.java
```diff
@@ -0,0 +1,127 @@
+package azkaban.fixture;
```
HappyRay: Pls add copyright notice.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTest.java
```diff
@@ -0,0 +1,193 @@
...
+        HttpServletResponse resp = getResponse(writer);
+
+
+        MockLoginAzkabanServlet servlet = MockLoginAzkabanServlet.getServletWithSession(sessionId,
+                "user", "127.0.0.1");
+
+
+        servlet.doPost(req, resp);
+
+        // Assert that our response was written (we have a valid session)
```
HappyRay: Is the comment valid?

mradamlacey:
Hi updated the pull request to refactor out a common method in `WebUtils` class to parse the real client ip.

Updated the spots in the Login servlet and the ResourceUtils class.

Updated unit tests.

Addressed your other comment about adding license file.

thanks!

azkaban-common/src/main/java/azkaban/utils/WebUtils.java
```diff
@@ -164,4 +165,29 @@ else if (sizeBytes >= ONE_KB)
     else
       return sizeBytes + " B";
   }
+
+  public String getRealClientIpAddr(Map<String, String> httpHeaders, String remoteAddr){
```
HappyRay: Could you add JavaDoc?

azkaban-common/src/main/java/azkaban/utils/WebUtils.java
```diff
@@ -164,4 +165,29 @@ else if (sizeBytes >= ONE_KB)
...
+
+    // If some upstream device added an X-Forwarded-For header
+    // use it for the client ip
+    // This will support scenarios where load balancers or gateways
+    // front the Azkaban web server and a changing Ip address invalidates
+    // the session
+
+    String clientIp = httpHeaders.getOrDefault("X-Forwarded-For", null);
```
HappyRay: Consider making "X-Forwarded-For" a constant?

azkaban-web-server/src/test/java/azkaban/fixture/MockLoginAzkabanServlet.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright 2014 LinkedIn Corp.
```
HappyRay: Make it 2016

mradamlacey:
Just updated to address your latest changes.  Let me know if anything else is needed.

HappyRay:
LGTM. Thanks!
One copyright notice date is still 2014. But I can merge it as it is.

PR #760, opened 2016-10-05
Fix broken Azkaban Solo Server due to removed h2database dependency

pranayhasan:
Deploying Azkaban Solo Server throws Class Not Found error : org.h2.Driver.
Azkaban-Solo-Server uses in-memory h2 DB (runtime dependency), when deployed locally and has been removed recently while refactoring.

### DIFF
--- a/azkaban-common/build.gradle
+++ b/azkaban-common/build.gradle
@@ -39,6 +39,8 @@ dependencies {
   compile('org.mortbay.jetty:jetty-util:6.1.26')
   compile('org.quartz-scheduler:quartz:2.2.1')
 
+  runtime('com.h2database:h2:1.4.192')
+
   testCompile(project(':azkaban-test').sourceSets.test.output)
   testCompile('org.hamcrest:hamcrest-all:1.3')
 }


azkaban-common/build.gradle
```diff
@@ -16,6 +16,7 @@ model {
 
 dependencies {
   compile('com.google.guava:guava:13.0.1')
+  compile('com.h2database:h2:1.4.184')
```
nntnag17: This seems to be a very old version. Can't we use the latest version?

nntnag17: Latest is 1.4.192 in mavenCentral.
https://mvnrepository.com/artifact/com.h2database/h2/1.4.192

pranayhasan: Done

suvodeep-pyne:
Where do we use this ? It would be great if you can add something in the description detailing the issue / motivation for the PR.

Azkaban builds successfully without any problems today. Are you sure this is a compile time dependency?

suvodeep-pyne:
@pranayhasan 
Another question is that you mentioned the solo-server is failing. However, you made changes to azkaban-common?

pranayhasan:
@suvodeep-pyne **azkaban-solo-server** has a transitive dependency on **azkaban-common** sub-project as can be seen in _azkaban-solo-server/build.gradle_.

fsi206914:
I met this issue when I run solo server in Idea IntelliJ. I resolved by discarding h2 conf and use real Mysql Instead.

suvodeep-pyne:
@pranayhasan 
Thanks for helping us with this. runtime dependency looks good. I am aware of the transitive dependency.
I have 1 question though: whether this should be a runtime dependency of azkaban-common or azkaban-solo-server ? Because, I am not sure whether azkaban-common always uses h2 DB. However, this seems more reasonable when a user wants to churn up a solo instance of Azkaban.

suvodeep-pyne:
Fixed in #814

Issue #761, opened 2016-10-05
Azkaban can potentially print sensitive Hadoop token information


KyleFung:
Azkaban currently prints out the toString() of Hadoop tokens into logs. Some tokens may expose sensitive information through its toString(). If logs were to be publicly accessible between all users, users will gain access to tokens of other users.

KyleFung:
Closed via #762

PR #762, opened 2016-10-05
Removed logging of toString() and getIdentifier() of tokens in HadoopSecurityManager_H_2_0 (#761)

KyleFung:
Token toString() and getIdentifier() may reveal sensitive information that should not be visible in logs, if the logs are to be publicly visible. Hence, we are removing logging statements that include this information.


suvodeep-pyne:
don't print the id as well. remember ? HIVE_DELEGATION_TOKEN dumps sensitive info in the identifier.
Also check if we are printing this elsewhere in the class ?

suvodeep-pyne:
@KyleFung Can you update the PR title and description with a little more detail. Then the title can be a little more concise.

HappyRay:
LGTM, please update the branch again.

PR #765, opened 2016-10-07
update Scheduling Doc

fsi206914:
Update Scheduling Documentation.

literally checked in local site.


HappyRay:
What do you think about adding a link to the wiki page which has more details?

What do you think of moving our docs to the wiki which is easier to update?

fsi206914:
Link was added. If we will make a big change in future, like AZ 4.0, we might seriously consider rewrite and organize docs in Wiki only.

Issue #697, opened 2016-07-31
github page ( for docs)  build warning


HappyRay:
The page build completed successfully, but returned the following warning:

You are attempting to use the 'pygments' highlighter, which is currently unsupported on GitHub Pages. Your site will use 'rouge' for highlighting instead. To suppress this warning, change the 'highlighter' value to 'rouge' in your '_config.yml' and ensure the 'pygments' key is unset. For more information, see https://help.github.com/articles/page-build-failed-config-file-error/#fixing-highlighting-errors.

For information on troubleshooting Jekyll see:

  https://help.github.com/articles/troubleshooting-jekyll-builds

If you have any questions you can contact us by replying to this email.

HappyRay:
@fsi206914 do you think you can fix this? This is a simple fix.

fsi206914:
#766 should fix this.

PR #766, opened 2016-10-11
fix github page warning by replacing highlighter value

fsi206914:
Followed the instructions in #697 . Fixed. 

Has to be varied by merging this pull request.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,8 @@
 name: Azkaban
 safe: true
 lsi: false
-pygments: true
+# pygments: true
+highlighter: rouge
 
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -1,7 +1,8 @@
 name: Azkaban
 safe: true
 lsi: false
-pygments: true
+# pygments: true
```
HappyRay: Why not just remove it?
Having commented out code is not a good practice in general.

fsi206914: Just removed.

fsi206914:
Comments were fixed, and Merging.

Issue #768, opened 2016-10-14
Startup shell scripts do not set a log4j properties file


KyleFung:
During startup of webserver, execserver, or single server, log4j is configured using a file that is taken from inside of a jar. If a user puts a log4j.properties file into the conf directory to configure the logger, this file is ignored.

Instead we should check if such a file exists in the conf directory and use it if it's there. If it doesn't exist, I think using the file already existing in the jar suffices as a default.

KyleFung:
Done via commit [62fc290](https://github.com/azkaban/azkaban/commit/62fc2901fe53bf95b1ab95879f1148c8b5f5c8a6)

PR #769, opened 2016-10-14
Enable using log4j.properties file in conf directory (#768)

KyleFung:
Right now, Azkaban searches for a log4j.properties file baked inside of
a jar file. This change makes the Azkaban startup script search for this
file inside of the conf directory and use it if it exists. If it does
not exist, then we will revert to using the properties file found in the
jar by default.

### DIFF
--- a/azkaban-exec-server/src/main/bash/azkaban-executor-start.sh
+++ b/azkaban-exec-server/src/main/bash/azkaban-executor-start.sh
@@ -2,6 +2,10 @@
 
 azkaban_dir=$(dirname $0)/..
 
+# Specifies location of azkaban.properties, log4j.properties files
+# Change if necessary
+conf=$azkaban_dir/conf
+
 if [[ -z "$tmpdir" ]]; then
 tmpdir=/tmp
 fi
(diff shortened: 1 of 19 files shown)


HappyRay:
Can the AZ code itself look for the log4j config in the config directory first?

suvodeep-pyne:
do we need to do the same for exec server ?

KyleFung:
Yes, we can. This patch updates the startup script of solo server, exec server and web server.

azkaban-exec-server/src/main/bash/azkaban-executor-start.sh
```diff
@@ -44,6 +44,10 @@ serverpath=`pwd`
 if [ -z $AZKABAN_OPTS ]; then
   AZKABAN_OPTS="-Xmx3G"
 fi
+# Set the log4j configuration file
```
HappyRay: Doesn't this code assume that the conf dir will be under the binary dir? The location can be overwritten by using the -conf parameter eg. -conf temp/conf 

What do you think about adding the conf directory to the search path? Or use PropertyConfigurator see http://stackoverflow.com/questions/7390521/change-location-of-log4j-properties  What's the implication for log4j2?

HappyRay:
Another advantage of having the code search for the conf dir first is that it will remove the duplicate logic for the three scripts.

HappyRay:
Discussed with Kyle off line. It's tricky to add the conf directory to the search path before log4j is initialized.

So we will go with Kyle's original design.

Thanks

Issue #770, opened 2016-10-17
Add Kafka runtime dependencies to gradle build


KyleFung:
To use KafkaAppender in log4j, some Kafka runtime dependencies are required. These need to be added to the list of run time dependencies in the gradle build script.

KyleFung:
If it's a runtime dependency, people who want to use the Kafka Appender should download the jar into their class path by themselves.
Closing this issue as unnecessary.

PR #771, opened 2016-10-17
Added runtime dependencies for KafkaAppender to gradle build (#770)

KyleFung:
Some jar dependencies are needed to use the KafkaAppender in Log4j. These were added to the list of runtime dependencies.

### DIFF
--- a/azkaban-common/build.gradle
+++ b/azkaban-common/build.gradle
@@ -41,6 +41,9 @@ dependencies {
 
   testCompile(project(':azkaban-test').sourceSets.test.output)
   testCompile('org.hamcrest:hamcrest-all:1.3')
+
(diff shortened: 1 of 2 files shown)


HappyRay:
I believe we have to use the internal version of Kafka. Have you check with the Kafka team?

KyleFung:
The internal version of Kafka should be able to communicate with the external version of Kafka. I've also tested this using the external Kafka on our machines and they work fine.

suvodeep-pyne:
Just recording what was discussed offline.

This code introduces compile time dependencies on Kafka. I'm just wondering whether that's the best way to go.

KyleFung:
Closing this as if people want to use Kafka at run time, they should include the runtime dependency in themselves.

PR #772, opened 2016-10-18
Bug Fix: Hardcoded build directory causes some missing files in the distribution

suvodeep-pyne:
Since our internal build system uses a custom build directory, the distribution created was missing some files. Changing to $buildDir removes such issues.

### DIFF
--- a/azkaban-web-server/build.gradle
+++ b/azkaban-web-server/build.gradle
@@ -174,7 +174,7 @@ distributions {
       from (dustjs) {
         into 'web/js'
       }
-      from ('build/jsToPackage') {
+      from ("${buildDir}/jsToPackage") {
         into 'web/js'
       }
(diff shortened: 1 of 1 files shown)


suvodeep-pyne:
@fsi206914 please take a look when you get a chance

fsi206914:
Yes. The original hardcode gradle statement makes our system choose a wrong folder to copy. This change works finally.

HappyRay:
Seems to make sense to me.
For my education, could you give an example of how the build directory variable is set and how it may be different from "build"?

suvodeep-pyne:
@HappyRay : you can take a look at this : http://stackoverflow.com/questions/23340946/gradle-global-build-directory

HappyRay:
Thanks

PR #777, opened 2016-10-25
Fix execute as user override issue

fsi206914:
Previous execute-as-user committers left a security hole that users could make use of. In user's job conf, users are able to define EXECUTE_AS_USER_OVERRIDE's boolean value to enable or disable execute_as_user feature.

During this change, we canceled user's capability, and removed all related code.

Verified in our Experimental AZ cluster, and no issues were found.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -176,12 +176,7 @@ public class ProcessJob extends AbstractProcessJob {
   }
 
   private boolean determineExecuteAsUser(Props sysProps, Props jobProps) {
-    boolean isExecuteAsUser = sysProps.getBoolean(EXECUTE_AS_USER, false);
-    // putting an override in case user needs to override. A temporary opening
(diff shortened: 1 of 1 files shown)


suvodeep-pyne:
@fsi206914 quick question: shouldn't we always execute as user ?

fsi206914:
@suvodeep-pyne I believe jobs should execute as user. If we make execute-as-user to false, most jobs also work. @georgezhlw should have more context about this.

HappyRay:
I have the same question. Why shouldn't we always execute as user? If so, should we remove all references to it? e.g. in the run method.

fsi206914:
You both have good question. Since that will be a separate security question, how about re-evaluating it after merging this.

evlstyle:
execute-as-user was a BRS that ws left over at development time by the author of the feature to A/B testing the feature and make sure in the case the new feature doesn't work we can quickly revert back without redeploying the code, he planned to remove this before he moved out from the team so removing/disabling this is the right move, and yes jobs should always run as user .

HappyRay:
Thanks @evlstyle !

PR #780, opened 2016-10-26
disable cache fetch when it is looking for job state.

fsi206914:
Welcome comments to this pull request.

One issue was found in our production cluster: Even though one job is actually running, but it shows preparation state in job graph page. It affects our users checking logs. 

I did an investigation on this issue, and found that it should be something wrong with the cache implementation, since database has the right job state info. 

I just disabled cache fetching logic in getExecutableFlow method. Checked all callers of this method, and they all come from web server logics. Then I believe this should be a safe change.

Verified in our experimental cluster, and no issues were found.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -536,21 +536,14 @@ public class ExecutorManager extends EventHandler implements
   }
 
   /**
-   * Fetch ExecutableFlow from an active (running, non-dispatched) or from
-   * database {@inheritDoc}
+   * Fetch ExecutableFlow from database {@inheritDoc}
    *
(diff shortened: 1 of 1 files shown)


HappyRay:
LGTM

suvodeep-pyne:
Where are these caches being used? If these caches aren't being maintained correctly then shouldn't we remove all logic that depends on them?

I think the preparing step is showing for a reasonable amount of time which means that the caches aren't being updated for long periods of time. Is this the case ?

fsi206914:
@suvodeep-pyne 
Yes, I agree. I'd suggest putting this log in cluster and see what happened. If it works very well, we may remove all cache logic, since Ray and I didn't figure why we needed a cache for fetching state. It looks redundant.

Your though is right. However, the period of time can not be predicted. Sometimes a couple of minutes; Sometimes, half an hour.

PR #782, opened 2016-10-27
Add current Uploader userId field to Project class and set it in ProjectManager

pranayhasan:
Useful for Validating Projects based on current Uploader ID. 

Context:
Currently ProjectValidator interface passes Project object to validate an uploaded project. The Project object has no infomation on user who is trying to upload the project Zip. So, adding a field currentUploaderId makes it possible for classes extending ProjectValidator to validate on basis of current uploader.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/Project.java
+++ b/azkaban-common/src/main/java/azkaban/project/Project.java
@@ -39,6 +39,7 @@ public class Project {
   private int version = -1;
   private long createTimestamp;
   private long lastModifiedTimestamp;
+  private String currentUploaderID;
   private String lastModifiedUser;
   private String source;
   private LinkedHashMap<String, Permission> userPermissionMap =
@@ -331,6 +332,24 @@ public class Project {
 
   public void setLastModifiedUser(String lastModifiedUser) {
     this.lastModifiedUser = lastModifiedUser;
+  }
+
+  /**
+   * Getter method for current uploader userId
+   *
+   * @return currentUploaderId
+   */
+  public String getCurrentUploaderID() {
+    return currentUploaderID;
+  }
+
+  /**
+   * Setter method for current Uploader userId
+   *
+   * @param currentUploaderID
+   */
+  public void setCurrentUploaderID(String currentUploaderID) {
+    this.currentUploaderID = currentUploaderID;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -39,6 +39,7 @@
   private int version = -1;
   private long createTimestamp;
   private long lastModifiedTimestamp;
+  private String currentUploaderID;
```
suvodeep-pyne: By current, do you mean the latest uploader of the project ?

How is this information being persisted? Or is this a transient field?
Also I am not sure whether this data is set on all code paths. Since, every project will be uploaded by some user, this should not be an optional data.

pranayhasan: Yes, it is the ID of the latest uploader who tried uploading to azkaban be it successful/not. It is transient as there is no DB update. There exists a field lastModifiedUser which gets persisted after the upload succeeds and is fetched from the DB.

What do you mean by all code paths? Why do you think it is optional?

pranayhasan: Basically we want to have a list of projects and users who are allowed to upload to a cluster. 
Seems that your concern is that this authorization is cluster specific and not project specific. And so why do we want to include the uploader information in Project? Yes, you are right. 

For this we need a Cluster Admin who can have a whitelist/blacklist of projects and users who are allowed to upload. Azkaban has users with Admin access and no cluster Admin, so how do we do it?

Project Validation:
We already have a ProjectValidator interface which makes azkaban-plugins to implement project specific validation. So, we can use it to whitelist projects. No changes need in Azkaban for doing this.

User Validation:
We can have a new GenericValidator interface, which takes some specific metadata and validates on it. We pass the current user uploading zip to this and validate it. Also, if we want to validate on specific use-case in the future, if would be easy to add to this metadata.
Does this sound good?

HappyRay: I share with @suvodeep-pyne's concern that this information doesn't seem to fit well in the logic definition of a project. 

Could you describe your new GenericValidator interface better? Maybe in the issue?

HappyRay:
@pranayhasan since the issue is closed, should this PR be closed too?

pranayhasan:
@HappyRay This was done as a part of blocking zip uploads by maintaining a whitelist of users and projects where some custom validation was required. If we are **NOT** looking to add any validation in Azkaban, we are good to close this.

PR #789, opened 2016-11-01
Fix cron expression validation

fsi206914:
Currently The Verifying capability of CronExpression is not sufficient. During this pull request, I further check validity of a cron expression by obtaining the next available time of cron.

Unite Test is added.

Verified through local environment as well.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/Utils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Utils.java
@@ -490,6 +490,12 @@ public class Utils {
     } else return null;
   }
 
+  /**
+   *
+   * @param cronExpression
+   * @param timezone
+   * @return if the cronExpression is valid or not.
+   */
   public static boolean isCronExpressionValid(String cronExpression, DateTimeZone timezone) {
     if (!CronExpression.isValidExpression(cronExpression)) {
       return false;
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/utils/Utils.java
```diff
@@ -489,7 +490,22 @@ public static CronExpression parseCronExpression(String cronExpression, DateTime
...
+   * @param cronExpression
+   * @param timezone
+   * @return if the cronExpression is valid or not.
+   */
+  public static boolean isCronExpressionValid(String cronExpression, DateTimeZone timezone) {
+    if (!CronExpression.isValidExpression(cronExpression)) {
+      return false;
+    } else {
+      CronExpression cronExecutionTime = parseCronExpression(cronExpression, timezone);
+      if (cronExecutionTime == null || cronExecutionTime.getNextValidTimeAfter(new Date()) == null) {
```
suvodeep-pyne: so we are trying to avoid scheduling flows in the past. right ? Can you please add a small note.

nit: else statement is not required.

fsi206914: Good suggestion. Incorporated.

fsi206914:
I think we can push this pull request. @HappyRay

Issue #792, opened 2016-11-03
Failed upload has different response in Web UI vs Ajax interface


KyleFung:
Uploading a broken zip through the Web UI has a visual message telling users that the upload failed. However, doing it through the Ajax interface still tells you the same thing, but is accompanied with a HTTP 200 response (success). This causes scripts to break as they should be expecting 400 (bad request).

KyleFung:
There are many other problems similar to this that need to be addressed. The issue to fix that is filed at #795

PR #793, opened 2016-11-03
Made failed uploads return a HTTP 400 response (#792)

KyleFung:
Failed uploads through Ajax interface reported HTTP code 200 when it
should be 400. There are other similar errors in this module, but this
commit only fixes failed uploads.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -91,7 +91,7 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       "lockdown.create.projects";
   private static final String LOCKDOWN_UPLOAD_PROJECTS_KEY =
       "lockdown.upload.projects";
-  
+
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1749,7 +1754,7 @@ private void ajaxHandleUpload(HttpServletRequest req,
           error =
               error.substring(0, 512) + "<br>Too many errors to display.<br>";
         }
-        ret.put("error", "Installation Failed.<br>" + error);
```
HappyRay: Is it necessarily a client error?

KyleFung: It is impossible from here to tell if it is 400 or 500. I suppose we should err on the side of caution and return 500.

HappyRay:
As discussed, pls create a ticket to track future enhancement including our discussion of the alternative approach of using an exception to return the errors.

KyleFung:
Yeah I made it here: #795. I forgot to reference it in this pull request.

PR #797, opened 2016-11-04
Option to create Kafka Appender for user job logs (#778)

KyleFung:
JobRunner now checks for properties in jobs to check if Kafka logging is
enabled. If it is, then it will attempt to create a Kafka appender. If
that fails, it will report the failure but not crash the application.

User side:
A basic example usage of this would be to drop a properties file (named maybe like kafka.properties) with the following lines into the root directory of your project:
azkaban.job.logging.kafka.enable=true

This property should cause all the jobs in your project to log to Kafka, using a producer provided by the executor server. Note that the executor server can override this property and refuse to provide this producer.

The property azkaban.job.logging.kafka.enable is by default false, so if left alone, the default behavior is to not log to Kafka.

Server admin side:
There are two properties to configure the Kafka producer that would be used as a logger for user jobs and they can be set in azkaban.properties. They are:
azkaban.server.logging.kafka.brokerList
azkaban.server.logging.kafka.topic

If any of these two properties are missing, then the executor server will refuse to attempt to make the producer. This can doubly act as a killswitch if the server admins decide they don't want anyone logging to Kafka, or don't have a Kafka cluster set up.

If during the creation of the Kafka logger, something wrong happens, then we will fail silently (not crash the app), and report the failure in the flow log.


HappyRay:
Would it be better to make 
"azkaban.job.kafkalogging.brokerlist=localhost:9092
azkaban.job.kafkalogging.topic=azkaban-web-server"
configurable by the Azkaban server administrators. 

Users don't necessarily know or care about this details. And it would be more reliable to have a centralized place to control this information.

KyleFung:
I was thinking we could centralize the information into one place, but I suppose users wouldn't really know those details and it'd waste their time to find them.

We can instead put them in the azkaban.properties of exec server.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,28 +213,64 @@ private void createLogger() {
...
-            new RollingFileAppender(loggerLayout, absolutePath, true);
-        fileAppender.setMaxBackupIndex(jobLogBackupIndex);
-        fileAppender.setMaxFileSize(jobLogChunkSize);
-        jobAppender = fileAppender;
-        logger.addAppender(jobAppender);
-        logger.setAdditivity(false);
-      } catch (IOException e) {
-        flowLogger.error("Could not open log file in " + workingDir
-            + " for job " + this.jobId, e);
+      if (props.getBoolean("azkaban.job.kafkalogging.enabled", false)) {
```
HappyRay: Consider making azkaban.job.kafkalogging.enabled a constant?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,28 +213,64 @@ private void createLogger() {
...
+          + " for job " + this.jobId, e);
+    }
+  }
+
+  private void attachKafkaAppender() {
+    kafkaAppender = null;
+
+    try {
+      KafkaLog4jAppender kafkaProducer = new KafkaLog4jAppender();
+      kafkaProducer.setBrokerList(props.getString("azkaban.job.kafkalogging.brokerlist"));
```
HappyRay: What's the impact to the future migration to log4j2?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,28 +213,64 @@ private void createLogger() {
...
+              + "\"submituser\": \"" + props.getString("azkaban.flow.submituser") + "\","
+              + "\"execid\": \"" + props.getString("azkaban.flow.execid") + "\","
+              + "\"projectversion\": \"" + props.getString("azkaban.flow.projectversion") + "\""
+              + "}"));
+      kafkaProducer.activateOptions();
+      kafkaAppender = kafkaProducer;
+      logger.addAppender(kafkaAppender);
+      logger.setAdditivity(false);
+      flowLogger.info("Created kafka appender for " + this.jobId);
+    } catch (Exception e) {
```
HappyRay: Is it possible to catch more specific exceptions?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,28 +213,64 @@ private void createLogger() {
...
+      flowLogger.error("Could not create Kafka Appender for job " + this.jobId, e);
```
HappyRay: What will happen if kafaAppender is not null in this case? What's the effect of the removeAppender call later?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: It could be possible to add more exception types.
However we need to keep using Exception as a catch-all since we want any and all failures to silently happen (we shouldn't crash the application due to the logger failing).

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: It's pretty hard to tell. We could try to close and set null the appender in this catch clause to prevent invalid operations in the future.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -67,6 +69,7 @@
   private Logger flowLogger = null;
 
   private Appender jobAppender;
+  private Appender kafkaAppender;
```
suvodeep-pyne: Let's use Optional here since this is a configurable parameter which may or may not be present.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: Looking at the source code for removeAppender, it shouldn't be an issue if the appender is corrupt or non existent as it is just a remove operation from a Vector (won't blow up if it's not there).

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: How would you envision using log4j2's ability to dynamically adjust logging here? What's the use case?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: Many of the literal strings in this file are not stored as constants somewhere. Would it look right to add this in?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,26 +215,55 @@ private void createLogger() {
...
-        logger.setAdditivity(false);
+        attachFileAppender(createFileAppender());
       } catch (IOException e) {
+        removeAppender(jobAppender);
         flowLogger.error("Could not open log file in " + workingDir
             + " for job " + this.jobId, e);
       }
+
+      AzkabanExecutorServer exec = AzkabanExecutorServer.getApp();
+      if (exec != null && !exec.getAzkabanProps().getBoolean("kafkaLogging.globalDisable", false)
```
suvodeep-pyne: There is a props already in the JobRunner. Any reason why we didn't use that ?

KyleFung: That props is the properties object of the job itself. We want to configure the globalDisable variable at the server level (from azkaban.properties), which is accessed as above.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -238,10 +272,55 @@ private void createAttachmentFile() {
...
+  }
+
+  private KafkaLog4jAppender createKafkaAppender() {
+    KafkaLog4jAppender kafkaProducer = new KafkaLog4jAppender();
+    Props azkProps = AzkabanExecutorServer.getApp().getAzkabanProps();
+    kafkaProducer.setSyncSend(false);
+    kafkaProducer.setBrokerList(azkProps.getString("kafkaLogging.brokerList"));
+    kafkaProducer.setTopic(azkProps.getString("kafkaLogging.topic"));
+    kafkaProducer.setLayout(new EnhancedPatternLayout(
+        "{\"category\": \"%c{1}\","
```
suvodeep-pyne: I would prefer something like schema to be a resource file since managing this giant string in java will be a pain. What do you think ?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
-        logger.setAdditivity(false);
+        attachFileAppender(createFileAppender());
       } catch (IOException e) {
+        removeAppender(jobAppender);
         flowLogger.error("Could not open log file in " + workingDir
             + " for job " + this.jobId, e);
       }
+
+      AzkabanExecutorServer exec = AzkabanExecutorServer.getApp();
+      if (exec != null && !exec.getAzkabanProps().getBoolean("kafkaLogging.globalDisable", false)
```
HappyRay: Consider a more consistent naming convention for the config property names?

e.g.
azkaban.server.logging.enableKafka
azkaban.job.logging.enableKafka

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+          && props.getBoolean("azkaban.job.kafkaLogging.enabled", false)) {
```
HappyRay: What happens if the kafka server config is not configured and user enables this feature at the job level?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+
   private void createAttachmentFile() {
     String fileName = createAttachmentFileName(node);
     File file = new File(workingDir, fileName);
     attachmentFileName = file.getAbsolutePath();
   }
 
+  private void attachKafkaAppender(KafkaLog4jAppender appender) {
+    // If present, remove the existing kafka appender
+    if (kafkaAppender.isPresent()) {
```
HappyRay: If you are sure and expect this should not happen, may be change it to an assert?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+    }
+
+    kafkaAppender = Optional.of(appender);
+    logger.addAppender(kafkaAppender.get());
+    logger.setAdditivity(false);
+    flowLogger.info("Attached new Kafka appender for job " + this.jobId);
+  }
+
+  private KafkaLog4jAppender createKafkaAppender() {
+    KafkaLog4jAppender kafkaProducer = new KafkaLog4jAppender();
```
HappyRay: Consider adding a unit test for this method?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+    Props azkProps = AzkabanExecutorServer.getApp().getAzkabanProps();
+    kafkaProducer.setSyncSend(false);
+    kafkaProducer.setBrokerList(azkProps.getString("kafkaLogging.brokerList"));
```
HappyRay: Use constants for config names and follow a consistent naming convention as suggested above?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+    kafkaProducer.setTopic(azkProps.getString("kafkaLogging.topic"));
+
+    JSONObject layout = new JSONObject();
```
HappyRay: Nice usage of the class to make the code easier to read than using raw strings.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +217,118 @@ private void createLogger() {
...
+    kafkaProducer.setLayout(new EnhancedPatternLayout(layout.toString()));
+    kafkaProducer.activateOptions();
+
+    flowLogger.info("Created kafka appender for " + this.jobId);
+    return kafkaProducer;
+  }
+
+  private void removeAppender(Optional<Appender> appender) {
+    removeAppender(appender.get());
+    appender = Optional.empty();
```
HappyRay: Is this step necessary? I thought it would only affect the reference in this method's scope.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: It will throw an UndefinedPropertyException and fail. This exception is caught and handled so it will not crash the server.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: Yes, I think it would be useful to have a killswitch like this that users cannot override.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
KyleFung: Done. A bit of refactoring of the job and flow runner classes had to be done.

KyleFung: Added to Constants.java

azkaban-common/src/main/java/azkaban/server/Constants.java
```diff
@@ -22,4 +22,16 @@
 	public static final String AZKABAN_PRIVATE_PROPERTIES_FILE = "azkaban.private.properties";
 	public static final String DEFAULT_CONF_PATH = "conf";
 	public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
+
+  public static final String AZKABAN_SERVER_LOGGING_KAFKA_GLOBAL_DISABLE = "azkaban.logging.kafka.globalDisable";
+  public static final String AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST = "azkaban.logging.kafka.brokerList";
```
HappyRay: What do you think of a name like azkaban.server.logging.kafka.brokerList? This way the name signals that this is a server config.

azkaban-common/src/main/java/azkaban/server/Constants.java
```diff
@@ -22,4 +22,16 @@
 	public static final String AZKABAN_PRIVATE_PROPERTIES_FILE = "azkaban.private.properties";
 	public static final String DEFAULT_CONF_PATH = "conf";
 	public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
+
+  public static final String AZKABAN_SERVER_LOGGING_KAFKA_GLOBAL_DISABLE = "azkaban.logging.kafka.globalDisable";
```
HappyRay: What do you think about removing this config and just use presence of the brokerlist and topic config as an indication if admins would like to enable this feature?

azkaban-common/src/main/java/azkaban/server/Constants.java
```diff
@@ -22,4 +22,16 @@
...
+  public static final String AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST = "azkaban.logging.kafka.brokerList";
+  public static final String AZKABAN_SERVER_LOGGING_KAFKA_TOPIC = "azkaban.logging.kafka.topic";
+
+  public static final String AZKABAN_FLOW_PROJECT_NAME = "azkaban.flow.projectname";
```
HappyRay: Are these configs injected by the AZ itself ( vs. created by users )?

azkaban-common/src/main/java/azkaban/server/Constants.java
```diff
@@ -22,4 +22,16 @@
...
+  public static final String AZKABAN_FLOW_FLOW_ID = "azkaban.flow.flowid";
+  public static final String AZKABAN_FLOW_SUBMIT_USER = "azkaban.flow.submituser";
+  public static final String AZKABAN_FLOW_EXEC_ID = "azkaban.flow.execid";
+  public static final String AZKABAN_FLOW_PROJECT_VERSION = "azkaban.flow.projectversion";
+
+  public static final String AZKABAN_JOB_LOGGING_KAFKA_ENABLE = "azkaban.job.logging.kafka.enable";
```
HappyRay: What do you think about creating separate constant classes? One for server config, one for flow config, one for job config? 
This file seems to be define constants that are not user visible.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +221,112 @@ private void createLogger() {
...
-        logger.addAppender(jobAppender);
-        logger.setAdditivity(false);
+        attachFileAppender(createFileAppender());
       } catch (IOException e) {
+        removeAppender(jobAppender);
         flowLogger.error("Could not open log file in " + workingDir
             + " for job " + this.jobId, e);
       }
+
+      if (!azkabanProps.getBoolean(Constants.AZKABAN_SERVER_LOGGING_KAFKA_GLOBAL_DISABLE, false) &&
```
HappyRay: See my earlier comments about using other configs to determine if admin has enabled the kafka logging globally. Maybe log an info message if users enables this feature but admin disables it?

HappyRay:
Without unit tests, how do you protect against regressions?

HappyRay:
Could you update the change description to reflect your new changes?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: This has an added benefit of dealing with the situation when the admins forgot to config one of the two server configs more explicitly and make it easier to debug.

azkaban-common/src/main/java/azkaban/server/Constants.java
KyleFung: Yes, that was actually a typo on my part.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +221,112 @@ private void createLogger() {
...
+    flowLogger.info("Attached new Kafka appender for job " + this.jobId);
+  }
+
+  private KafkaLog4jAppender createKafkaAppender() throws UndefinedPropertyException {
+    KafkaLog4jAppender kafkaProducer = new KafkaLog4jAppender();
+    kafkaProducer.setSyncSend(false);
+    kafkaProducer.setBrokerList(azkabanProps.getString(Constants.AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST));
+    kafkaProducer.setTopic(azkabanProps.getString(Constants.AZKABAN_SERVER_LOGGING_KAFKA_TOPIC));
+
+    JSONObject layout = new JSONObject();
```
suvodeep-pyne: nice !

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -137,9 +138,9 @@
    * @throws ExecutorManagerException
    */
   public FlowRunner(ExecutableFlow flow, ExecutorLoader executorLoader,
-      ProjectLoader projectLoader, JobTypeManager jobtypeManager)
+      ProjectLoader projectLoader, JobTypeManager jobtypeManager, Props serverProps)
```
suvodeep-pyne: Quick question: What is the differences between server props and azkaban props? I see different usages from the code. Does it make sense to combine the two ?

KyleFung: They are the same. I changed my mind on what to call it and switched all of them to azkabanProps but I guess I missed this one.

azkaban-common/src/main/java/azkaban/constants/JobProperties.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012 LinkedIn Corp.
+ * Copyright 2016 LinkedIn Corp.
```
HappyRay: Thanks for updating the date

azkaban-common/src/main/java/azkaban/constants/PropertyFilePaths.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.constants;
+
+public class PropertyFilePaths {
```
HappyRay: The context key doesn't seem to be path related.

What do you think about naming it something more generic like Server and put all server internal constants here?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -210,38 +223,116 @@ private void createLogger() {
...
+        if (azkabanProps.containsKey(ServerProperties.AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST)
+            && azkabanProps.containsKey(ServerProperties.AZKABAN_SERVER_LOGGING_KAFKA_TOPIC)) {
+          try {
+            attachKafkaAppender(createKafkaAppender());
+          } catch (Exception e) {
+            removeAppender(kafkaAppender);
+            flowLogger.error("Failed to create Kafka appender for job " + this.jobId, e);
+          }
+        } else {
+          flowLogger.error("Kafka appender not created as brokerlist or topic not provided by executor server");
```
HappyRay: Maybe info? This doesn't indicate an error in most cases, right?

PR #801, opened 2016-11-09
Remove execute.as.user in conf

jamiesjc:
The execute-as-user feature is already applied for around 1 year. Every production cluster make it to true in configurations. We should disable this conf now, and let it be always true.

# configs for execute-as-user
execute.as.user=true

I removed the code to read execute-as-user from configuration so that jobs will always be executed by user instead of Azkaban.

Can you review the changes before I test on the clusters?

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -108,11 +108,14 @@ public class ProcessJob extends AbstractProcessJob {
     Map<String, String> envVars = getEnvironmentVariables();
     envVars.put(KRB5CCNAME, getKrb5ccname(jobProps));
 
-    // determine whether to run as Azkaban or run as effectiveUser
+    // determine whether to run as Azkaban or run as effectiveUser,
+    // by default, run as effectiveUser
     String executeAsUserBinaryPath = null;
     String effectiveUser = null;
     boolean isExecuteAsUser = determineExecuteAsUser(sysProps, jobProps);
 
+    // nativeLibFolder specifies the path for execute-as-user file,
+    // which will change user from Azkaban to effectiveUser
     if (isExecuteAsUser) {
       String nativeLibFolder = sysProps.getString(NATIVE_LIB_FOLDER);
       executeAsUserBinaryPath =
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -175,7 +175,7 @@ public class ProcessJob extends AbstractProcessJob {
   }
 
   private boolean determineExecuteAsUser(Props sysProps, Props jobProps) {
-    return sysProps.getBoolean(EXECUTE_AS_USER, false);
+    return sysProps.getBoolean(EXECUTE_AS_USER, true);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -111,37 +110,28 @@ public void run() throws Exception {
     // determine whether to run as Azkaban or run as effectiveUser
```
fsi206914: This comment needs to be removed or changed.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -111,37 +110,28 @@ public void run() throws Exception {
...
-      String nativeLibFolder = sysProps.getString(NATIVE_LIB_FOLDER);
-      executeAsUserBinaryPath =
-          String.format("%s/%s", nativeLibFolder, "execute-as-user");
-      effectiveUser = getEffectiveUser(jobProps);
-      if ("root".equals(effectiveUser)) {
-        throw new RuntimeException(
-            "Not permitted to proxy as root through Azkaban");
-      }
+
+    String nativeLibFolder = sysProps.getString(NATIVE_LIB_FOLDER);
```
fsi206914: Consider add a comment to specify what is nativeLibFolder for.

fsi206914:
Looks good. Please validate this change in our AZ test instance.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -174,10 +164,6 @@ public void run() throws Exception {
     generateProperties(propFiles[1]);
   }
 
-  private boolean determineExecuteAsUser(Props sysProps, Props jobProps) {
-    return sysProps.getBoolean(EXECUTE_AS_USER, false);
```
logiclord: I think it is always good to have an option. I believe it will be good to keep the config and just change the default to "true" so you can remove default value from all .properties file.

HappyRay: @logiclord Thanks for your comment. 

IMHO having option in general is typically a tradeoff between complexity and flexibility. I will comment on the conversation.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -175,7 +178,7 @@ public void run() throws Exception {
   }
 
   private boolean determineExecuteAsUser(Props sysProps, Props jobProps) {
```
HappyRay: Is this method still necessary? The second param is no longer used.

azkaban-common/src/test/java/azkaban/jobExecutor/JavaProcessJobTest.java
```diff
@@ -108,13 +108,13 @@ public void setUp() throws IOException {
...
     props.put("fullPath", ".");
-    
+
     props.put(CommonJobProperties.PROJECT_NAME, "test_project");
     props.put(CommonJobProperties.FLOW_ID, "test_flow");
     props.put(CommonJobProperties.JOB_ID, "test_job");
     props.put(CommonJobProperties.EXEC_ID, "123");
     props.put(CommonJobProperties.SUBMIT_USER, "test_user");
-    
+    props.put("execute.as.user", "false");
```
HappyRay: Explain why this property is necessary?

HappyRay:
You may want to reference the old pull request you abandoned.

HappyRay:
Please consider cleaning up the commit messages before merging.

PR #803, opened 2016-11-10
removing jmx reporter related code since Deprecated

kunkun-tang:
Originally AZ used an JMX Emitter plugin to send JMX to remote metrics collector. However, it will be deprecated soon since we will bring up dropwizard metrics to AZ due to its powerfulness.

As we are approaching completing new metrics architecture, we decided to remove jmx reporter inside the AZ code.

verified in my local environment.


HappyRay:
The concern is that some users may still rely on it. 

Maybe we can leave it there for now?

Issue #805, opened 2016-11-12
Delete all config files except for those used in solo server


KyleFung:
Currently we keep configuration files like azkaban.properties checked in to our repo. As a result we have these configuration files baked into the jars of Azkaban. This doesn't make sense as our configurations do not belong inside our binaries.

We should delete them all, and have users specify a directory for their configurations.

KyleFung:
Closing as change was merged (#805).

PR #806, opened 2016-11-12
Deleted Azkaban config files from repo (#805)

KyleFung:
Currently, the deleted files are being written into Azkaban jars. These
do not belong here, and can cause hidden configurations from being used.


HappyRay:
Can we keep a copy of the sample configs somewhere so that people can use them as a reference?

azkaban-web-server/src/main/resources/conf/azkaban-users.xml
```diff
@@ -1,7 +0,0 @@
-<azkaban-users>
-	<user username="azkaban" password="azkaban" roles="admin" groups="azkaban" />
-	<user username="metrics" password="metrics" roles="metrics"/>
```
fsi206914: When Intellij runs, will this file be enabled? Have you evaluated about this?

HappyRay:
What will be logged if the log4j.properties files are removed and there is no log4j configs provided via the command line? I thought these files provide default logging settings for the application.

HappyRay:
Maybe we can keep the config in the solo server as an example and make solo server easier to setup? see #800

azkaban-web-server/src/main/resources/conf/azkaban-users.xml
suvodeep-pyne: @fsi206914 
I didn't quite understand your question. Can you please clarify your concern?

suvodeep-pyne:
@HappyRay I think the solo server config has been left intact for the exact same reason

suvodeep-pyne:
@HappyRay , the default log4j configuration was leading to exceptions while running tests. The problem is that the default configuration will also conflict with external configuration which is difficult to debug because the log4j.properties gets shipped inside the jar.

The solo server ships with a default log4j config

azkaban-web-server/src/main/resources/conf/azkaban-users.xml
fsi206914: @suvodeep-pyne Never mind. I thought some azkaban-user.xml file might be loaded when Intellij runs the main method of az-solo-server.

HappyRay:
@suvodeep-pyne I thought the default log4j config will be replaced by an external config if provided. What exceptions did you run into?

KyleFung:
@HappyRay The external one overrides the default through the start shell script, or through the "conf" option in the command line. I believe that running tests does not invoke either of these, so no overrides will occur.

KyleFung:
Actually it seems that solo-server relies on the config files that we want to delete. We should add a conf directory solely for the use of solo-server.

KyleFung:
Due to some weird mess up with git, I am continuing this work at #814.

PR #807, opened 2016-11-12
Fix comments for execute-as-user changes

jamiesjc:
Made some changes to simplify the code.
Added some comments for test cases.

PR #808, opened 2016-11-12
Fixed comments for execute.as.user changes (#807)

jamiesjc:
Removed a private function to make the code easier to read.
Added comments to the changes in test cases.
Abandoned previous pull request (#807).

### DIFF
--- a/azkaban-common/src/test/java/azkaban/jobExecutor/JavaProcessJobTest.java
+++ b/azkaban-common/src/test/java/azkaban/jobExecutor/JavaProcessJobTest.java
@@ -114,7 +114,11 @@ public class JavaProcessJobTest {
     props.put(CommonJobProperties.JOB_ID, "test_job");
     props.put(CommonJobProperties.EXEC_ID, "123");
     props.put(CommonJobProperties.SUBMIT_USER, "test_user");
-    props.put("execute.as.user", "false");
+
+    //The execute-as-user binary requires special permission. It's not convenient to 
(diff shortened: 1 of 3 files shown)


azkaban-common/src/test/java/azkaban/jobExecutor/JavaProcessJobTest.java
```diff
@@ -114,7 +114,11 @@ public void setUp() throws IOException {
     props.put(CommonJobProperties.JOB_ID, "test_job");
     props.put(CommonJobProperties.EXEC_ID, "123");
     props.put(CommonJobProperties.SUBMIT_USER, "test_user");
-    props.put("execute.as.user", "false");
+
```
HappyRay: Consider consolidating the common code in a common setup method in tests?

jamiesjc:
The configuration for props doesn't seem to be common for most test cases. Even for these two similar test cases, they have different props like "type". It might not be necessary to create a new common setup method?

fsi206914:
@HappyRay @jamiesjc  Program readability is also important to readers. I don't have preferences given this case.

HappyRay:
@jamiesjc please update the documentation as needed to reflect the new default. 
When we release a new version, we need to highlight this change. A user who upgrades will need to know the new behavior and adjust their configs accordingly..

fsi206914:
Confused if this change is merged or not. @jamiesjc

PR #810, opened 2016-11-14
skip emailing users when internal error happens in finalizing flow

chengren311:
skip emailing users when internal error happens when finalizing flow. This is to prevent users receive unnecessary alerting emails when db updates fail.


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1421,58 +1422,56 @@ private void finalizeFlows(ExecutableFlow flow) {
...
+        }
+        if (options.getFlowParameters().containsKey("alert.type")) {
+          String alertType = options.getFlowParameters().get("alert.type");
+          Alerter alerter = alerters.get(alertType);
+          if (alerter != null) {
+            try {
+              alerter.alertOnSuccess(flow);
+            } catch (Exception e) {
+              // TODO Auto-generated catch block
+              e.printStackTrace();
```
HappyRay: Could you make a separate pull request later to fix the logging here and others in this file?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1421,58 +1422,56 @@ private void finalizeFlows(ExecutableFlow flow) {
...
+      if (flow.getStatus() == Status.FAILED || flow.getStatus() == Status.KILLED) {
+        if (options.getFailureEmails() != null && !options.getFailureEmails().isEmpty()) {
           try {
-            alerter.alertOnError(flow);
+            mailAlerter.alertOnError(flow);
           } catch (Exception e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
-            logger.error("Failed to alert by " + alertType);
+            logger.error(e);
```
HappyRay: Have you considered raise the same exception here?

HappyRay:
How did you test this change?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: this is just exception in sending emails. logging the exception should be ok. If we raise the same exception,  the appropriate way to handle it by downstream handler is log it anyway.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: sure, will do

chengren311:
@HappyRay i tested it locally as sole mode. Manually throws an exception in the db deletion method and verified it won't send any alerting emails.

PR #812, opened 2016-11-14
Removing deprecated schedule panel conf

fsi206914:
The Flexible Scheduling feature is already rolled out to our production cluster for around 2 months. It's time to disable deprecated scheduling feature by removing related files and configurations.


HappyRay:
Thanks for cleaning up the code.

PR #813, opened 2016-11-14
Executor to use any port if not configured. Added REST APIs to activate / deactivate executor

suvodeep-pyne:
When an executor server is launched, it uses the configured port or falls back to a default port. This behavior has been modified such that when the executor is not configured to a specific port it defaults to any available port that it can find on the machine. The advantage is that the user doesn't need to worry about port collisions when he/she doesn't need a specific port.

  Also, The executor now automatically adds an entry about itself to the DB post launch. Also, it exposes the following REST APIs to allow users to activate / deactivate itself
  example:
  http://localhost:1234/executor?action=activate
  http://localhost:1234/executor?action=deactivate

On a successful start, the executor writes the port to a file called executor.port by default.

Contains small refactors like consolidating constants. simplifying function signature.

No automated tests were added due to the nature of the change. No additional functionality was added by existing pieces of code was used. However, the changes were tested manually.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
+++ b/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
@@ -129,11 +129,13 @@ public class DataSourceUtils {
 
     private static MonitorThread monitorThread = null;
 
+    private final String url;
+
     private MySQLBasicDataSource(String host, int port, String dbName,
         String user, String password, int numConnections) {
       super();
 
-      String url = "jdbc:mysql://" + (host + ":" + port + "/" + dbName);
+      url = "jdbc:mysql://" + (host + ":" + port + "/" + dbName);
       addConnectionProperty("useUnicode", "yes");
       addConnectionProperty("characterEncoding", "UTF-8");
       setDriverClassName("com.mysql.jdbc.Driver");
@@ -196,10 +198,7 @@ public class DataSourceUtils {
           PreparedStatement query = connection.prepareStatement("SELECT 1");
           query.execute();
         } catch (SQLException e) {
-          // TODO Auto-generated catch block
-          e.printStackTrace();
-          logger
-              .error("MySQL connection test failed. Please check MySQL connection health!");
+          logger.error("Unable to reach MySQL server on " + url);
         } finally {
           DbUtils.closeQuietly(connection);
         }
--- a/azkaban-common/src/main/java/azkaban/server/Constants.java
+++ b/azkaban-common/src/main/java/azkaban/server/Constants.java
@@ -21,6 +21,5 @@ public class Constants {
 	public static final String AZKABAN_PROPERTIES_FILE = "azkaban.properties";
 	public static final String AZKABAN_PRIVATE_PROPERTIES_FILE = "azkaban.private.properties";
 	public static final String DEFAULT_CONF_PATH = "conf";
-
-	public static final String AZKABAN_EXECUTOR_PORT_FILE = "executor.port.filename";
+	public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
 }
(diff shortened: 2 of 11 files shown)


HappyRay:
"No additional functionality was added by existing pieces of code was used." Do you mean " no other additional functionality was added? "

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -129,11 +129,13 @@ private DataSourceUtils() {
 
     private static MonitorThread monitorThread = null;
 
+    private final String url;
```
HappyRay: Why this change?

azkaban-common/src/main/java/azkaban/server/Constants.java
```diff
@@ -16,6 +16,11 @@
 
 package azkaban.server;
 
-public class ServerConstants {
+public class Constants {
 	public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
+	public static final String AZKABAN_PROPERTIES_FILE = "azkaban.properties";
```
HappyRay: Is this class shared between the web server and executor?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -134,34 +171,35 @@ public AzkabanExecutorServer(Props props) throws Exception {
...
+    } catch (ExecutorManagerException e) {
+      logger.error("Error inserting executor entry into DB", e);
+      Throwables.propagate(e);
     }
+  }
 
-    logger.info("Azkaban Executor Server started on port " + portNumber);
+  private void dumpPortToFile() {
+    // By default this should write to the working directory
+    String filename = props.getString(AZKABAN_EXECUTOR_PORT_FILE, "executor.port");
```
HappyRay: Is it necessary to make the port file name configurable?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -134,34 +171,35 @@ public AzkabanExecutorServer(Props props) throws Exception {
...
+    try (BufferedWriter writer = new BufferedWriter(new FileWriter(filename))) {
+      writer.write(String.valueOf(getPort()));
+      writer.write("\n");
+    } catch (IOException e) {
+      logger.error(e);
```
HappyRay: Our automation may rely on the port file being there. Should we allow the exception to be bubbled up?

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
suvodeep-pyne: I wanted to log the url in case of DB connection failures. 
Apparently there is no easy API to access the url from that class.

azkaban-common/src/main/java/azkaban/server/Constants.java
suvodeep-pyne: yes. everything in azkaban-common is.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: No. In fact I don't see any advantage in doing it.
I just did it since I didn't see any disadvantage either. It allows people to put whatever name they like.
This can be changed to a constant.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: yes. I updated the code

HappyRay:
LGTM.

In the future, the API change seems to be independent of the port change. I would suggest making the independent changes separate pull requests.

PR #814, opened 2016-11-15
Deleted Azkaban config files from repo

KyleFung:
Currently, the deleted files are being written into Azkaban jars. These
do not belong here, and can cause hidden configurations from being used.

Consequently, azkaban-solo-server will not work as it relies on these
files, so solo-server was given its own conf directory.

Also, h2 was added as a runtime dependency for solo-server, which
happened to be missing for the past little while.


KyleFung:
Due to some weird mess up with git, I am continuing the work from #806 here.

suvodeep-pyne:
@KyleFung
I think you are accidentally including your older PR# in your new PR. Example 805. This might confuse people. You should be able to edit / remove older PR numbers from your requests.

PR #815, opened 2016-11-15
Fix db upload issue

fsi206914:
When too many projects are being uploaded simultaneously, we find an issue that the mysql server can not work properly such that memory starvation happens.  We enforce az committing to db when uploading every single chunk, in order to resolve this problem.

Already verified in our testing cluster.


HappyRay:
In 
public void uploadProjectFile(Project project, int version, String filetype,
      String filename, File localFile, String uploader)
    logger.info("Uploading to " + project.getName() + " version:" + version
        + " file:" + filename);
    Connection connection = getConnection();

    try {
      uploadProjectFile(connection, project, version, filetype, filename,
          localFile, uploader);
      connection.commit();
      logger.info("Commiting upload " + localFile.getName());
    } catch (SQLException e) {
      logger.error(e);
      throw new ProjectManagerException("Error getting DB connection.", e);
    } finally {
    }
  
Does it make sense to do
connection.rollback();
in the catch block?
See 
http://stackoverflow.com/questions/15761791/transaction-rollback-on-sqlexception-using-new-try-with-resources-block
This may be important when we implement connection pooling.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -382,6 +382,25 @@ private void uploadProjectFile(Connection connection, Project project,
     }
 
     logger.info("Md5 hash created");
+
+    /**
+     * Insert a version row to table project_versions.
+     * If something goes wrong during uploading project files to DB, the latest version number
+     * in project_versions refers to the destructive project_file. Since AZ always let the number+1
```
HappyRay: Do you mean "incomplete project files?"

HappyRay: Do you mean " this is desirable to ensure that the next version number will be unique"?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -416,16 +438,20 @@ private void uploadProjectFile(Connection connection, Project project,
       IOUtils.closeQuietly(bufferedStream);
     }
 
-    final String INSERT_PROJECT_VERSION =
-        "INSERT INTO project_versions (project_id, version, upload_time, uploader, file_type, file_name, md5, num_chunks) values (?,?,?,?,?,?,?,?)";
+    /**
+     * Since we initialized num_chunks to 0 before uploading files, we set it
```
HappyRay: Not sure I follow the comment.

HappyRay:
Consider explaining how this change would solve the problem and how it handles error cases.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -416,16 +438,20 @@ private void uploadProjectFile(Connection connection, Project project,
...
-          version, updateTime, uploader, filetype, filename, md5, chunk);
+      runner.update(connection, UPDATE_PROJECT_NUM_CHUNKS, chunk, project.getId(), version);
+      connection.commit();
     } catch (SQLException e) {
-      logger.error(e);
-      throw new ProjectManagerException("Error updating project version "
-          + project.getName(), e);
+      throw new ProjectManagerException(
+          "Error updating project file chunk_num " + project.getId() + ":"
+              + chunk, e);
```
HappyRay: Nice improvement to the logging. Thanks.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -382,6 +382,25 @@ private void uploadProjectFile(Connection connection, Project project,
     }
 
     logger.info("Md5 hash created");
+
+    /**
+     * Insert a version row to table project_versions.
+     * If something goes wrong during uploading project files to DB, the latest version number
```
HappyRay: Do you mean " something goes wrong after this entry is inserted"?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
HappyRay: Explain why this update goes first before inserting the data chunks?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -402,9 +421,12 @@ private void uploadProjectFile(Connection connection, Project project,
           logger.info("Running update for " + filename + " chunk " + chunk);
           runner.update(connection, INSERT_PROJECT_FILES, project.getId(),
               version, chunk, size, buf);
+
+          // We enforce az committing to db when uploading every a single chunk.
```
HappyRay: Explain the reason. 
How about " to reduce the transaction duration and conserve sql server resources"?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
fsi206914: Fixed.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
fsi206914: Fixed.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
fsi206914: Fixed.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
fsi206914: Fixed.

PR #819, opened 2016-11-17
Cleanup process ID and executor.port files on shutdown

suvodeep-pyne:
The executor server dumps the current server port to executor.port file. On shutdown clean it up.
Also remove process ID files which are created upon launch.
+refactor


azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
```diff
@@ -1,11 +1,19 @@
...
 
-kill $proc
+kill ${pid}
+if [ $? -ne 0 ]; then
+    echo "Error: Shutdown failed"
+    exit 1;
+fi
 
-cat /dev/null > $azkaban_dir/currentpid
+rm  ${installdir}/currentpid
```
fsi206914: One silly question: Why do we need to remove them? Does rpm uninstall being able to remove them?

suvodeep-pyne: @fsi206914 
This commit actually has nothing to do with rpm build. It is just that I thought no pid file would make more sense than empty file.
Currently, when server runs, the pid file is populated. When it's shut down, the pid file is either empty or not present. Instead of that, the state is being simplified to not be present on shutdown.

azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
```diff
@@ -1,11 +1,19 @@
...
+rm  ${installdir}/executor.port
```
HappyRay: Is it still useful to keep the port file so that we can match the number in our logs?

I don't have a strong opinion either way. If we find such a need later, we can easily remove this line.

HappyRay:
Do you plan to make the same change to our internal version of shutdown scripts too? Or do you plan to start using this version?

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
```diff
@@ -1,8 +1,15 @@
...
-#!/bin/bash
-proc=`cat $azkaban_dir/currentpid`
-echo "killing AzkabanWebServer"
-kill $proc
+pid=`cat ${installdir}/currentpid`
+echo "Killing Web Server. [pid: $pid]"
 
-cat /dev/null > $azkaban_dir/currentpid
+kill ${pid}
+if [ $? -ne 0 ]; then
```
suvodeep-pyne: I was noticing instances where shutdown wasn't shutting down the process in solo server.

suvodeep-pyne:
Unless we have custom stuff to do in our internal scripts, we can directly use these.

azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
suvodeep-pyne: The port number will be in the logs. 
Do you want to have an easy access to the last port number? Just want to understand the use case. My motivation is to clean up the temp files on shutdown.

fsi206914:
Is the bash folder newly created recently? Will the bash scripts directly go to the production cluster from now on? I believe azkaban4 also has some bash scripts, like shutdown? Should we remove them in future?

suvodeep-pyne:
nope. it's exists. git will show if files are created / modified.
These scripts won't be in effect immediately since we replace those bash scripts from the ones in our config repo. However, I would like to get rid of all bash scripts in config repo but before that I would like to clean up the existing scripts.

PR #820, opened 2016-11-18
Show version number in Azkaban logo in Web server UI

suvodeep-pyne:
Currently there is no way of finding the Azkaban version of the web UI. With this patch the version will show up under the Azkaban logo and will persist on every page.

I have attached a preview below on how it looks

screenshot 2016-11-17 19 58 18


HappyRay:
LGTM.

I noticed that you reformatted the code in this change as well. It's good to clean up the code. However it makes it a bit harder to review and trace the change history later IMHO.

suvodeep-pyne:
It gets difficult to read nested logic if the code is not indented properly.

Issue #821, opened 2016-11-18
Create PatternLayout that escapes troublesome characters


KyleFung:
When we use log4j to send a JSON message the official PatternLayout will send corrupt JSON if certain characters are present. We use this PatternLayoutEscaped class as a thin interface around PatternLayout to escape all these characters.

An example use case of this is if we are sending JSON with a KafkaAppender. Multi lined messages result in corrupt JSON, requiring that the new lines be escaped before being appended.

So far we know we need to escape backslashes, tabs, newlines and quotes.

KyleFung:
Closed via #822

PR #822, opened 2016-11-18
Create PatternLayout that escapes troublesome characters (#821)

KyleFung:
When we use log4j to send a JSON message the official PatternLayout will
send corrupt JSON if certain characters are present. We use this
PatternLayoutEscaped class as a thin interface around PatternLayout to
escape all these characters.

An example use case of this is if we are sending JSON with a
KafkaAppender. Multi lined messages result in corrupt JSON, requiring
that the new lines be escaped before being appended.

So far we know we need to escape backslashes, tabs, newlines and quotes.


azkaban-common/src/main/java/azkaban/utils/PatternLayoutEscaped.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.apache.log4j.PatternLayout;
+import org.apache.log4j.spi.LoggingEvent;
+
+/**
+ * When we use log4j to send a JSON message the official PatternLayout will send corrupt JSON if certain characters are
+ * present. We use this PatternLayoutEscaped class as a thin interface around PatternLayout to escape all these
+ * characters.
+ */
+public class PatternLayoutEscaped extends PatternLayout {
```
HappyRay: Is this mechanism compatible with log4j2 and slf4j?

HappyRay:
How did you test?

azkaban-common/src/main/java/azkaban/utils/PatternLayoutEscaped.java
KyleFung: Why slf4j? Looking at the source for log4j2's PatternLayout class these will not be compatible. Also I believe that with log4j2, a scheme like this is not necessary.

KyleFung:
@HappyRay testing was done by logging messages containing these characters, confirming that they come out corrupted using the default PatternLayout, and then confirming that they work using the one created in this commit.

azkaban-common/src/main/java/azkaban/utils/PatternLayoutEscaped.java
HappyRay: I was thinking how a component that uses slf4j would be affected. I guess this change doesn't affect the producer of logging messages. 

wrt log4j2, could you share more details how the migration plan would be? We may eventually move to log4j2.

HappyRay:
Could you add unit tests?

KyleFung:
@HappyRay Unit tests added.

PR #824, opened 2016-11-18
update sql clauses to 3.10.0

kunkun-tang:
Multi Executor mode is already applied in production for more than 1 years. It's time to update the sql clauses to match the latest multi executor Mode.


HappyRay:
How did you test?

HappyRay:
@jamiesjc could you review this change?

fsi206914:
I already released the new sql changes to one of our new AZ DB. Will evaluate how it behave after AZ is set up there. @HappyRay

azkaban-sql/src/sql/database.properties
```diff
@@ -1 +1 @@
-version=3.0
+version=3.10.0
```
fsi206914: I've no much idea how to modify this version. Just give it a recent number.

suvodeep-pyne: I don't think this file is used anywhere. I'm ok with deleting this file.

suvodeep-pyne:
Hey @fsi206914, How are the sql files being used / read in the code ?

fsi206914:
@suvodeep-pyne I believe these sql files were not read/used in the internal AZ code. I just copy tables one by one every time I set up an AZ db.

fsi206914:
removed database.properties file.

HappyRay:
The file database.properties is referenced in  azkaban.database.AzkabanDatabaseSetup#findOutOfDateTable.

kunkun-tang:
@HappyRay I checked the logic. There is a boolean configuration database.check.version. If it is true, we will do the setup table automatically when web server start. I believe we should keep safe and get the file back to repo.

kunkun-tang:
Evaluated in our another az cluster, and it is working well.

Issue #823, opened 2016-11-18
azzkaban api call for fetch running execution is not working


tk0485:
trying to run :
curl -k -data "session.id=34ba08fd-5cfa-4b65-94c4-9117aee48dda&ajax=getRunning&project=azkaban-test-project&flow=test"' https://localhost:8443/executor

but its not working, its giving me error: couldn't resolve host 
is this endpoint has been depreciated? I checked it in azkaban source code and it seems to be there but for some reason its not working. is it a bug? or am I doing something wrong?

fsi206914:
@tk0485 are testing Azkaban using your solo server? the default port should be 8081, right?

tk0485:
yes, I'm using a solo server and the port is configured. Other api calls work fine and I can get the session id fine too except for this call.

KyleFung:
Maybe it's a copy paste error into github, but it seems your quotes don't match. Can you confirm this?
Just before your host name you have a double quote followed by a single quote.

tk0485:
@KyleFung as you have already guessed .. its a copy paste from azkaban github page, I deleted the single quote when I used the curl command in my tests.

tk0485:
I found the reason for my issue. There is a missing dash from the --data in the curl command, the curl command should like:
curl -k --data "session.id=34ba08fd-5cfa-4b65-94c4-9117aee48dda&ajax=getRunning&project=azkaban-test-project&flow=test" https://localhost:8443/executor

instead of having a single dash leading "data", I also deleted the extra single quote before hostname. Could Azkaban team please update their documentation to reflect the correct command.

KyleFung:
Fixed. Thanks for reporting.

KyleFung:
Closed via #826

PR #826, opened 2016-11-21
Fixed some typos in example curl commands (#823)

KyleFung:
Some instances of "--data" are written as "-data". Also there was a
random single quote floating in one of the commands. Removed them so as
not to mislead new users.


fsi206914:
LGTM.

Issue #818, opened 2016-11-17
Schedule job failed due to "azkaban.utils.UndefinedPropertyException: Missing required property 'azkaban.native.lib'"


ldzhjn:
17-11-2016 01:13:13 PST foo INFO - 1 commands to execute.
17-11-2016 01:13:13 PST foo INFO - effective user is: azkaban
17-11-2016 01:13:13 PST foo ERROR - Job run failed!
azkaban.utils.UndefinedPropertyException: Missing required property 'azkaban.native.lib'
at azkaban.utils.Props.getString(Props.java:478)
at azkaban.jobExecutor.ProcessJob.run(ProcessJob.java:120)
at azkaban.execapp.JobRunner.runJob(JobRunner.java:595)
at azkaban.execapp.JobRunner.run(JobRunner.java:446)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
at java.lang.Thread.run(Thread.java:745)
17-11-2016 01:13:13 PST foo ERROR - Missing required property 'azkaban.native.lib' cause: null
17-11-2016 01:13:13 PST foo INFO - Finishing job foo attempt: 0 at 1479373993806 with status FAILED

This is the error log when I try to schedule a simple command job on solo server.
I tried adding azkaban.native.lib in the job file and azkaban.properties file, the error is the same.

ldzhjn:
I was following this guide to setup the solo server and I didn't install any plugins
http://azkaban.github.io/azkaban/docs/latest/#solo-setup

hderms:
I have the same issue

hderms:
I was trying to evaluate whether to use Azkaban versus another workflow management tool but the fact that the example doesn't work is disconcerting

fsi206914:
@hderms This issue is fixed by #827, and is under review now.

rajivchodisetti:
Am not using Solo server mode but trying to install stand alone executor and webserver , but am also running into the same issue , I haven't installed any plugins 


29-11-2016 17:02:49 IST dim_date ERROR - Job run failed!
azkaban.utils.UndefinedPropertyException: Missing required property 'azkaban.native.lib'
[...]

PR #809, opened 2016-11-14
Consolidate common code808

jamiesjc:
Combined common code to configure job properties in one method.
Changes based on comments in Pull request #807.

PR #827, opened 2016-11-22
Fix issues on solo server with Execute-As-User changes(#818)

jamiesjc:
Users are seeing issues with the execute.as.user change(#809) on solo server without installing plugins.
Actually there is no need to change user on solo server and jobs can be executed by Azkaban itself.
So I added the new plugin folder to put configuration file for solo server and set execute.as.user to false there. Verified locally on solo server and it should fix the issue (#818).


fsi206914:
LGTM

PR #828, opened 2016-11-22
bring dropwizard Metrics to AZ to get a better monitoring feature

kunkun-tang:
This pull request is aimed at bringing dropwizard metrics to AZ, per the following causes:

1). Metrics Library has integrated powerful JVM interpretation tools, and is quite easy to use.

2). Metrics Reporter is pluggable, which help us extend it in our plugin easily

This change is to have a new monitoring architecture for AZ, so I only propose a few WEB and EXEC metrics for example. We will be able to add more metrics, which is registered at WEB and EXEC registers.


kunkun-tang:
Found a imperfection in code. The reporter plugin constructor should at least have a name field, in order to distinguish webserver or executor who launches the reporter thread.

azkaban-common/src/main/java/azkaban/constants/JobProperties.java
```diff
@@ -17,6 +17,6 @@
 package azkaban.constants;
 
 public class JobProperties {
-  // Job property that enables/disables using Kafka logging of user job logs
-  public static final String AZKABAN_JOB_LOGGING_KAFKA_ENABLE = "azkaban.job.logging.kafka.enable";
-}
+	// Job property that enables/disables using Kafka logging of user job logs
+	public static final String AZKABAN_JOB_LOGGING_KAFKA_ENABLE = "azkaban.job.logging.kafka.enable";
+}
```
suvodeep-pyne: newline missing

kunkun-tang: Added.

kunkun-tang:
Already evaluated in our staging server. Pushing.

PR #831, opened 2016-11-23
make project directory as shared among multiple deployments

chengren311:
Since we are moving projects dir to a shared dir among multiple deployments, we have to deal with race condition where multiple deployments are trying to modifying the same project file. 
The commit consists of following changes:

1. make rename operation atomic so that it's safer for multiple deployments to perform rename.
2. For simplicity, only active deployment is allowed to perform deletion of projects.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.lang.Thread.State;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -46,6 +45,7 @@ import azkaban.execapp.event.RemoteFlowWatcher;
 import azkaban.execapp.metric.NumFailedFlowMetric;
 import azkaban.executor.ExecutableFlow;
 import azkaban.executor.ExecutionOptions;
+import azkaban.executor.Executor;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.ExecutorManagerException;
 import azkaban.jobtype.JobTypeManager;
@@ -148,6 +148,9 @@ public class FlowRunnerManager implements EventListener,
   // date time of the the last flow submitted.
   private long lastFlowSubmittedDate = 0;
 
+  // whether the current executor is active
+  private volatile boolean isActive = false;
+
   public FlowRunnerManager(Props props, ExecutorLoader executorLoader,
       ProjectLoader projectLoader, ClassLoader parentClassLoader)
       throws IOException {
@@ -188,6 +191,8 @@ public class FlowRunnerManager implements EventListener,
 
     this.validateProxyUser =
         azkabanProps.getBoolean("proxy.user.lock.down", false);
+
+    setActive(true);
 
     cleanerThread = new CleanerThread();
     cleanerThread.start();
@@ -260,6 +265,10 @@ public class FlowRunnerManager implements EventListener,
     return allProjects;
   }
 
+  public void setActive(boolean isActive) {
+    this.isActive = isActive;
+  }
+
   public long getLastFlowSubmittedTime(){
     // Note: this is not thread safe and may result in providing dirty data.
     //       we will provide this data as is for now and will revisit if there
@@ -312,10 +321,9 @@ public class FlowRunnerManager implements EventListener,
               lastRecentlyFinishedCleanTime = currentTime;
             }
 
-            if (currentTime - OLD_PROJECT_DIR_INTERVAL_MS > lastOldProjectCleanTime) {
+            if (currentTime - OLD_PROJECT_DIR_INTERVAL_MS > lastOldProjectCleanTime && isActive) {
               logger.info("Cleaning old projects");
               cleanOlderProjects();
-
               lastOldProjectCleanTime = currentTime;
             }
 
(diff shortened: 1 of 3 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -156,6 +156,7 @@ public FlowRunnerManager(Props props, ExecutorLoader executorLoader,
     projectDirectory =
         new File(props.getString("azkaban.project.dir", "projects"));
 
+    logger.info("projectDirectory:"+projectDirectory.getAbsolutePath());
```
fsi206914: Delete it pls.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -16,14 +16,14 @@
 
 package azkaban.execapp;
 
+import azkaban.executor.Executor;
```
fsi206914: put it to the right location.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -260,6 +261,13 @@ public boolean accept(File dir, String name) {
     return allProjects;
   }
 
+  private boolean isActive() throws ExecutorManagerException {
+    String executorHost = AzkabanExecutorServer.getHost();
+    int executorPort = azkabanProps.getInt("executor.port");
+    Executor executor = executorLoader.fetchExecutor(executorHost, executorPort);
+    return executor != null ? executor.isActive() : false;
```
fsi206914: follow intellij's suggestion.

chengren311:
addressed the comments

chengren311:
@HappyRay, @fsi206914 
Looks good to you guys?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -191,6 +192,12 @@ private void insertExecutorEntryIntoDB() {
     }
   }
 
+  private void updatePortInProps() {
+    if(!props.containsKey("executor.port")) {
```
HappyRay: When will the props contain this key?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -260,6 +260,13 @@ public boolean accept(File dir, String name) {
     return allProjects;
   }
 
+  private boolean isActive() throws ExecutorManagerException {
+    String executorHost = AzkabanExecutorServer.getHost();
+    int executorPort = azkabanProps.getInt("executor.port");
+    Executor executor = executorLoader.fetchExecutor(executorHost, executorPort);
+    return executor != null && executor.isActive();
```
HappyRay: Is it simpler to rely on the active and deactivate API calls?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -577,7 +583,7 @@ private void setupFlow(ExecutableFlow flow) throws ExecutorManagerException {
       projectVersion.setupProjectFiles(projectLoader, projectDirectory, logger);
       projectVersion.copyCreateSymlinkDirectory(execPath);
     } catch (Exception e) {
-      e.printStackTrace();
+      logger.error("Error in setting up project directory "+projectDirectory+", "+e);
```
HappyRay: Nice.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -260,6 +260,13 @@ public boolean accept(File dir, String name) {
     return allProjects;
   }
 
+  private boolean isActive() throws ExecutorManagerException {
```
HappyRay: Where is this method used?

chengren311: it will be used in FlowRunnerManager's cleanup thread. Cleanup thread won't perform deletion when isActive returns false

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: can you point out relevant code?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: So this method is not used yet. Right? Would it be better to include that change in the next pull request then so that they are together?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: See #813

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -260,6 +265,10 @@ public boolean accept(File dir, String name) {
     return allProjects;
   }
 
+  public void setActive(boolean isActive) {
```
shuzhang1989: should we have a javabean?

PR #833, opened 2016-11-24
Change the Azkaban executor and web start script to redirect error ou…

HappyRay:
…tput to a log file correctly

Previously the error output was still sent to the stdout.

The change has been tested in our production clusters.


fsi206914:
LGTM.

PR #834, opened 2016-11-24
Add more error logging to the code that fetches project properties fr…

HappyRay:
…om DB

Let the exception from JSON decode method PropsUtils.fromJSONString to throw and catch it in the caller where more information is available to be logged.
 This way it is easier to identify which project is having this issue from logs.

PropsUtils.fromJSONString is called only in one place and the exception is caught.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -1112,6 +1112,8 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
 
       return properties.get(0).getSecond();
     } catch (SQLException e) {
+      logger.error("Error fetching property " + propsName
+          + " Project " + projectId + " version " + projectVer, e);
       throw new ProjectManagerException("Error fetching property " + propsName,
           e);
     }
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -1120,6 +1122,7 @@ public Props fetchProjectProperty(int projectId, int projectVer,
   @Override
   public Props fetchProjectProperty(Project project, String propsName)
       throws ProjectManagerException {
+    // TODO: 11/23/16 call the other overloaded method fetchProjectProperty internally.
```
fsi206914: Confused what does the TODO message mean?

azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
```diff
@@ -330,17 +330,10 @@ public static String toJSONString(Props props, boolean localOnly) {
...
-          (Map<String, String>) JSONUtils.parseJSONFromString(json);
-      Props props = new Props(null, obj);
-      return props;
-    } catch (IOException e) {
-      logger.error("Encountered error during parsing project properties json. ", e);
-      return null;
-    }
+  public static Props fromJSONString(String json) throws IOException {
+    Map<String, String> obj = (Map<String, String>) JSONUtils.parseJSONFromString(json);
+    Props props = new Props(null, obj);
```
fsi206914: we can also directly return new Props ?

HappyRay: Yes.

I don't have a strong opinion either way. 

I wanted to minimize my changes.

PR #839, opened 2016-11-30
shared project directory

chengren311:
1. make rename operation atomic (#831)
2. only active deployment is allowed to perform deletion of projects
This is the prequisite of making project directory as shared among multiple deployments


suvodeep-pyne:
from offline chat: This is a cherrypicked commit from master to hotfix branch.

Issue #840, opened 2016-12-01
Redirect System out and err to log4j


KyleFung:
For debugging purposes we will want to redirect messages printed through System.out and System.err into log4j. Currently, these are being put into the console but being missed by our logs, making it hard to find things like exceptions and the like.

KyleFung:
Closed via #841

PR #841, opened 2016-12-01
Redirect System out and err to log4j (#840)

KyleFung:
Created a class under utils that will tie out and err to log4j. This
class does this in the main functions of AzkabanWebServer and
AzkabanExecServer, so that all out going messages will be caught by
log4j.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
+++ b/azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2016 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.utils;
+
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+
+/**
+ * A class to encapsulate the redirection of stdout and stderr to log4j
+ * This allows us to catch messages written to the console (although we should
+ * not be using System.out to write out).
+ */
+
+public class StdOutErrRedirect {
+  private static final Logger logger = Logger.getLogger(StdOutErrRedirect.class);
+  private static final PrintStream infoStream = createStream(System.out, Level.INFO);
+  private static final PrintStream errorStream = createStream(System.out, Level.ERROR);
+
+  public static void redirectOutAndErrToLog() {
+    System.setOut(infoStream);
+    System.setErr(errorStream);
+  }
+
+  private static PrintStream createStream(PrintStream stream, Level level) {
+    return new LogStream(stream, level);
+  };
+
+  private static class LogStream extends PrintStream {
+    private final Level level;
+
+    public LogStream(OutputStream out, Level level) {
+      super(out);
+      this.level = level;
+    }
+
+    // Underlying mechanism to log to log4j - all print methods will use this
+    private void write(String string) {
+      logger.log(level, string);
+    }
+
+    // String
+    @Override
+    public void println(final String string) {
+      print(string);
+    }
+
+    @Override
+    public void print(final String string) {
+      write(string);
+    }
+
+    // Boolean
+    @Override
+    public void println(final boolean bool) {
+        print(bool);
+    }
+
+    @Override
+    public void print(final boolean bool) {
+        write(String.valueOf(bool));
+    }
+
+    // Int
+    @Override
+    public void println(final int i) {
+        print(i);
+    }
+
+    @Override
+    public void print(final int i) {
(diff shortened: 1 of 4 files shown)


kunkun-tang:
I'm confused that Ihow do you define "out going messages" ?

KyleFung:
I meant any messages that the server tries to relay to the outside world will be caught by log4j. This includes exceptions and stack traces.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,42 @@
...
+  private static final Logger logger = Logger.getLogger(StdOutErrRedirect.class);
+
+  public static void redirectOutAndErrToLog() {
+    System.setOut(createStream(System.out, Level.INFO));
+    System.setErr(createStream(System.err, Level.ERROR));
+  }
+
+  private static PrintStream createStream(PrintStream stream, Level level) {
+    return new PrintStream(stream) {
+      public void println(final String string) {
```
suvodeep-pyne: Curious: Will this work if I print an integer ?

Best practices: if you are overriding/implementing a function, use annotations to indicate them. Like in this case using @Override.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.PrintStream;
+
+import org.apache.log4j.Logger;
+import org.apache.log4j.Level;
+
+public class StdOutErrRedirect {
+  private static final Logger logger = Logger.getLogger(StdOutErrRedirect.class);
+
+  public static void redirectOutAndErrToLog() {
+    System.setOut(createStream(System.out, Level.INFO));
```
suvodeep-pyne: In the solo server case, there will be 2 PrintStream objects created. What do you think about maintaining a single instance of each in the class itself?

Do you see any advantages of creating a fresh stream every time a redirect is called ?

azkaban-solo-server/src/main/resources/log4j.properties
```diff
@@ -1,7 +1,5 @@
 log4j.rootLogger=INFO, Console
-log4j.logger.azkaban.webapp=INFO, server
-log4j.logger.azkaban.execapp=INFO, server
-log4j.logger.azkaban.soloserver=INFO, server
+log4j.logger.azkaban=INFO, server
```
suvodeep-pyne: Thanks for tidying this up

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
KyleFung: It seems you're right about the integer.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
KyleFung: Latest version creates static instances of them.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
KyleFung: Fixed by overriding a bunch of print/println functions. Unfortunately this was the only way, apart from overriding write(), but that option seemed very error prone and not as simple as overriding print/println.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * A class to encapsulate the redirection of stdout and stderr to log4j
+ * This allows us to catch messages written to the console (although we should
+ * not be using System.out to write out).
+ */
+
+public class StdOutErrRedirect {
+  private static final Logger logger = Logger.getLogger(StdOutErrRedirect.class);
+  private static PrintStream infoStream = createStream(System.out, Level.INFO);
```
suvodeep-pyne: preferably final ?

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,134 @@
...
+    System.setOut(infoStream);
+    System.setErr(errorStream);
+  }
+
+  private static PrintStream createStream(PrintStream stream, Level level) {
+    return new LogStream(stream, level);
+  };
+
+  private static class LogStream extends PrintStream {
+    private Level level;
```
suvodeep-pyne: nit: Let's keep variables final whenever they can be

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,134 @@
...
+      this.level = level;
+    }
+    // String - all other overridden methods will call this
+    @Override
+    public void println(final String string) {
+      print(string);
+    }
+    @Override
+    public void print(final String string) {
+      logger.log(level, string);
```
suvodeep-pyne: Can we have a separate differently named private method which does this, similar to the OutputStream in java. That makes the code much more readable than calling an already heavily overloaded and overridden method.

azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  private static class LogStream extends PrintStream {
+    private Level level;
+
+    public LogStream(OutputStream out, Level level) {
+      super(out);
+      this.level = level;
+    }
+    // String - all other overridden methods will call this
```
suvodeep-pyne: nit: formatting: single line gap between methods

PR #846, opened 2016-12-01
highlight property override change

kunkun-tang:
The AZ production users complained that 

> We've had a number of production issues where config properties had been
overridden in the Azkaban UI as a hot fix but then those changes didn't get
reflected in git. Subsequent deployments overrode those changes which
caused the issues.

To resolve this issue, the first step discussed is to highlight the property override change in project log.

Added Unit Test and Evaluated in my lcoalhost environment.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -32,9 +32,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.log4j.Logger;
 
 import azkaban.flow.Flow;
-import azkaban.project.DirectoryFlowLoader;
 import azkaban.project.ProjectLogEvent.EventType;
-import azkaban.project.ProjectWhitelist.WhitelistType;
 import azkaban.project.validator.ValidationReport;
 import azkaban.project.validator.ValidationStatus;
 import azkaban.project.validator.ValidatorConfigs;
@@ -44,6 +42,7 @@ import azkaban.user.Permission;
 import azkaban.user.Permission.Type;
 import azkaban.user.User;
 import azkaban.utils.Props;
+import azkaban.utils.PropsUtils;
 import azkaban.utils.Utils;
 
 public class ProjectManager {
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -354,16 +353,22 @@ public Props getJobOverrideProperty(Project project, String jobName)
     return projectLoader.fetchProjectProperty(project, jobName + ".jor");
   }
 
-  public void setJobOverrideProperty(Project project, Props prop, String jobName)
+  public void setJobOverrideProperty(Project project, Props prop, String jobName, User modifier)
       throws ProjectManagerException {
     prop.setSource(jobName + ".jor");
```
chengren311: can we put this ".jor" to constants files btw ?

kunkun-tang: What's the advantage of this change? I can't find many usages of ".jor".

azkaban-common/src/test/java/azkaban/utils/PropsUtilsTest.java
```diff
@@ -208,6 +208,30 @@ public void testCyclesResolveProps() throws IOException {
...
+
+    newProps2.put("a", "a_value1");
+    newProps2.put("b", "b_value1");
+    newProps2.put("c", "c_value1");
+
+    String message2= PropsUtils.getPropertyDiff(oldProps, newProps2);
+
+    Assert.assertEquals(message2, "Newly created Properties: [ c, c_value1], \n");
+  }
+
```
jamiesjc: Do we need to test diffProperties?

kunkun-tang: Thanks @jamiesjc for reminding me adding an extreme test case.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
shuzhang1989: @chengren311 +1, don't see the advantage

PR #848, opened 2016-12-06
get self executor status by writing a simple executor API

kunkun-tang:
The acceptance tests need to specify the target executor ID, which has to be fetched from DB directly. This pull request writes a new API to have it.

Verified in localhost.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ConnectorParams.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ConnectorParams.java
@@ -36,6 +36,7 @@ public interface ConnectorParams {
   public static final String RELOAD_JOBTYPE_PLUGINS_ACTION = "reloadJobTypePlugins";
   public static final String ACTIVATE = "activate";
   public static final String DEACTIVATE = "deactivate";
+  public static final String GET_STATUS = "getStatus";
 
   public static final String MODIFY_EXECUTION_ACTION = "modifyExecution";
   public static final String MODIFY_EXECUTION_ACTION_TYPE = "modifyType";
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
@@ -45,6 +45,9 @@ import azkaban.utils.FileIOUtils.JobMetaData;
 import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.JSONUtils;
 
+import static java.util.Objects.requireNonNull;
+
+
 public class ExecutorServlet extends HttpServlet implements ConnectorParams {
   private static final long serialVersionUID = 1L;
   private static final Logger logger = Logger.getLogger(ExecutorServlet.class
@@ -102,6 +105,9 @@ public class ExecutorServlet extends HttpServlet implements ConnectorParams {
         } else if (action.equals(ACTIVATE)) {
           logger.warn("Setting ACTIVE flag to true");
           setActive(true, respMap);
+        } else if (action.equals(GET_STATUS)) {
+          logger.debug("Get Executor Status: ");
+          getStatus(respMap);
         } else if (action.equals(DEACTIVATE)) {
           logger.warn("Setting ACTIVE flag to false");
           setActive(false, respMap);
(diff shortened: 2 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -368,6 +371,20 @@ private void setActive(boolean value, Map<String, Object> respMap)
     }
   }
 
+  private void getExecutorId(Map<String, Object> respMap)
+      throws ServletException {
+    try {
+      ExecutorLoader executorLoader = application.getExecutorLoader();
+      Executor executor = executorLoader.fetchExecutor(application.getHost(), application.getPort());
+      Preconditions.checkState(executor != null, "Unable to obtain self entry in DB");
+      respMap.put("executor_id", Integer.toString(executor.getId()));
```
chengren311: can we put this "executor_id" to the constants files?

kunkun-tang:
Found an issue to block this PR to merge. Since we don't know the port in advance, this API might be difficult to use.

azkaban-common/src/main/java/azkaban/executor/ConnectorParams.java
```diff
@@ -36,6 +36,7 @@
   public static final String RELOAD_JOBTYPE_PLUGINS_ACTION = "reloadJobTypePlugins";
   public static final String ACTIVATE = "activate";
   public static final String DEACTIVATE = "deactivate";
+  public static final String GET_EXECUTOR_ID = "getExecutorId";
```
suvodeep-pyne: nit: call it getId ? you are calling on the executor itself ?

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -368,6 +371,20 @@ private void setActive(boolean value, Map<String, Object> respMap)
     }
   }
 
+  private void getExecutorId(Map<String, Object> respMap)
+      throws ServletException {
+    try {
+      ExecutorLoader executorLoader = application.getExecutorLoader();
+      Executor executor = executorLoader.fetchExecutor(application.getHost(), application.getPort());
```
suvodeep-pyne: There is a requireNonNull API in java Objects class itself that you can directly use.

suvodeep-pyne: Also this code is now common for uses, in the Executor server class and in this class, set active method. Should we refactor that into a single method to fetch current executor object ?

kunkun-tang:
Your idea sounds tons make sense to me! Will work on this when having bandwidth.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
kunkun-tang: As long as we documented this parameter, it should be fine I believe.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
kunkun-tang: This is a cool API! just replaced.

kunkun-tang:
I renamed the API and added isAlive parameter to here.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -368,6 +374,22 @@ private void setActive(boolean value, Map<String, Object> respMap)
     }
   }
 
+  private void getExecutorStatus(Map<String, Object> respMap)
+      throws ServletException {
+    try {
+      ExecutorLoader executorLoader = application.getExecutorLoader();
+      final Executor executor = requireNonNull(executorLoader.fetchExecutor(application.getHost(), application.getPort()),
+          "The executor could not be null");
```
suvodeep-pyne: nit: grammar. "The current executor cannot be null"

azkaban-common/src/main/java/azkaban/executor/ConnectorParams.java
```diff
@@ -36,6 +36,7 @@
   public static final String RELOAD_JOBTYPE_PLUGINS_ACTION = "reloadJobTypePlugins";
   public static final String ACTIVATE = "activate";
   public static final String DEACTIVATE = "deactivate";
+  public static final String GET_EXECUTOR_STATUS = "getExecutorStatus";
```
suvodeep-pyne: nit: same comment again. Why not `GET_STATUS` and `getStatus` ? The API is being called on the executor itself. Personally, I prefer succinct names which not include redundant information.

kunkun-tang: Changed

PR #849, opened 2016-12-08
Executor REST API for safe shutdown

suvodeep-pyne:
This patch allows Azkaban Executor to expose a shutdown REST API for safe shutdown. 
API: `<hostname>:<port>/executor?action=shutdown`

Safe shutdown implies the following:
1. Mark itself inactive in the DB
2. Signal shutdown on the execution pool. No further jobs are accepted.
3. Awaits termination of the pool.
4. Shuts down the Jetty server post shutdown of execution pool

In odd cases the executor will still fail to completely kill itself. Those issues will be addressed in a upcoming different patch.


azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -169,6 +169,7 @@ public String getDBType() {
 
       public MonitorThread(MySQLBasicDataSource mysqlSource) {
         this.setName("MySQL-DB-Monitor-Thread");
+        setDaemon(true);
```
kunkun-tang: Does this thread hung when we do shutdown?

suvodeep-pyne: Don't think so. Note that this change will not fix the shutdown hang problem.

However, the monitor thread isn't something that we need to hold up the executor for. That's the reason I marked it as a daemon thread.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -890,4 +865,31 @@ public void afterExecute(Runnable r) {
...
+   * This shuts down the flow runner. The call is blocking and awaits execution of all jobs.
+   */
+  public void shutdown() {
+    logger.warn("Shutting down FlowRunnerManager...");
+    executorService.shutdown();
+    boolean result = false;
+    while (!result) {
+      logger.info("Awaiting Shutdown. # of executing flows: " + getNumRunningFlows());
+      try {
+        result = executorService.awaitTermination(1, TimeUnit.MINUTES);
```
kunkun-tang: I read the code again, and found that if result == false, the logic stays in while loop. Is that what we want?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -588,4 +580,51 @@ public int getPort() {
...
+   * Note: This should be run in a separate thread.
+   *
+   * Shutdown the server. (blocking call)
+   *  - waits for jobs to finish
+   *  - doesn't accept any new jobs
+   */
+  private void shutdownInternal() {
+    getFlowRunnerManager().shutdown();
+    try {
+      shutdownNow();
```
kunkun-tang: interested to learn why we need to shutdown, then shutdownnow (sounds redundant)

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -588,4 +580,51 @@ public int getPort() {
...
+      logger.warn("Shutdown AzkabanExecutorServer complete");
+    } catch (Exception e) {
+      logger.error(e);
```
kunkun-tang: Should we write some messages inside the log?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -890,4 +865,31 @@ public void afterExecute(Runnable r) {
...
+      } catch (InterruptedException e) {
+        logger.error(e);
+      }
+    }
+    logger.warn("Shutdown FlowRunnerManager complete.");
+  }
+
```
kunkun-tang: I'm new to ExecutorService, but the logic in my mind should be like:
if(executorService.awaitTermination(1, TimeUnit.MINUTES)) 
   print "shut down completes"
else
   shutdownnow();

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -357,16 +359,42 @@ private void handleReloadJobTypePlugins(Map<String, Object> respMap)
...
-        executor.setActive(value);
-        executorLoader.updateExecutor(executor);
-        flowRunnerManager.setActive(value);
-      } else {
-        logger.warn("Set active action ignored. Executor is already " + (value? "active" : "inactive"));
-      }
+      setActiveInternal(value);
+      respMap.put(STATUS_PARAM, RESPONSE_SUCCESS);
+    } catch (Exception e) {
+      logger.error(e);
```
kunkun-tang: May we have some messages inside the error log?

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -357,16 +359,42 @@ private void handleReloadJobTypePlugins(Map<String, Object> respMap)
...
+   * @param respMap json response object
+   * @throws ServletException
+   */
+  private void shutdown(Map<String, Object> respMap)
+      throws ServletException {
+    try {
+      logger.warn("Shutting down executor...");
+
+      // Set the executor to inactive. Will receive no new flows.
+      setActiveInternal(false);
```
kunkun-tang: What if DB is dead, are we able to shutdown executor as usual?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Yes. The logic is to wait for the flows to terminate. Timeout is a minute. You can see the awaitTermination documentation.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: well. shutdown refers to a safe shutdown, while shutdownNow refers to an immediate (may be unsafe) shutdown. This is a default convention for lot of libraries. You can see `java.util.concurrent.ExecutorService#shutdownNow` for example.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: The logging happens before this call. The main API call logs the shutdown call before calling the internal APIs

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Nope. Why would you want to kill flows ? That would be a problem for running flows. The idea of shutdown is to wait for executions to complete before shutdown.

So shutdown ≠ kill. shutdownNow is similar to kill

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
suvodeep-pyne: All these events are logged if you go down the trace. The db component logs the db change events.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
suvodeep-pyne: No. The shutdown would fail. It will be unsafe to shutdown the executor while it might still receive active traffic. right ? The goal is to set the executor to inactive first before attempting shutdown.

If you want to shutdown now, you can just use the shutdown bash script or kill the process.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
kunkun-tang: We have a concern here. we have some jobs running over 10 days and never ends. May we shutdown it by calling this function.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
kunkun-tang: I see. Thanks.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
kunkun-tang: Then I'm confused in what cases would we use rest call to shut down an executor?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Not through this API. This API can be enhanced in the future to have a timeout functionality.
Killing executors still needs to be done from the shell using the bash script

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
suvodeep-pyne: Currently this is intended for deployment use - when we update executors, we shutdown the older ones. This is intended to be the base commit. There are other issues which needs to be fixed and I will send out patches for those once this goes in.

PR #850, opened 2016-12-12
Adding new Jobtypes to Documentation.  

dwhu:
Adding sections for:
- Script Jobs (https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/jobExecutor/ScriptJob.java)
- Python Jobs - Script (https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/jobExecutor/PythonJob.java)
- Ruby Jobs (https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/jobExecutor/RubyJob.java)

### DIFF
--- a/_includes/docs/latest/builtin-jobtypes.html
+++ b/_includes/docs/latest/builtin-jobtypes.html
@@ -74,6 +74,74 @@
 	</tbody>
 </table>
 
+<h3>Script</h3>
+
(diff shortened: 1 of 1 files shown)


dmeijboom:
This looks very helpfull to me.. I didn't know there was a Python or Ruby job-type. Can this be merged?

dmeijboom:
Also maybe it would be nice to also state that if you want to use python3 on debian systems you can simply set the python variable in jobProps to python3 which then will be used as executable for python scripts.

HappyRay:
These job types are not documented. I am curious if these job types over significant value over simply using the command job type.

HappyRay:
These job types have been removed. #1431

PR #854, opened 2016-12-14
Made a Log4j pattern layout that appends stack traces to messages

KyleFung:
When we use the log4j Kafka appender, it seems that the appender simply does not log the stack trace anywhere. Seeing as the stack trace is a very important piece of information, we create our own PatternLayout class that appends the stack trace to the log message that reported it, so that all the information regarding that error can be found one in place.


azkaban-common/src/test/java/azkaban/utils/PatternLayoutEscapedTest.java
```diff
@@ -41,6 +41,17 @@ public void beforeTest() {
   }
 
   @Test
+  public void testWithException() {
```
suvodeep-pyne: Nice !

PR #855, opened 2016-12-15
Add docs for files generated by ./gradlew distTar

schatten:
Mention the location of tar files and how they can be un-archived.


kunkun-tang:
LGTM.

PR #857, opened 2016-12-16
adding Application metrics

kunkun-tang:
Before this change, we only have web and executor metrics registers, which are only able to collect metrics in exec or web packages. In order to collect more application metrics, I propose a CommonMetrics class in common repo, which is able to access code in azkaban-common more conveniently, then able to gather metrics there.


kunkun-tang:
Closing this PR, and  creating #904

PR #861, opened 2016-12-29
Fix bugs in prepare execution

juhoautio:
The major problems that this fixes:
- Wasn't enabling jobs inside sub-flows in case of certain flow statuses
- Wasn't enabling jobs if they were left in running or queued state

So the following changes:
- Always handle statuses of jobs inside all sub-flows
- Also enable for rerun if job was left in running or queued state
- Disable if job wasn't present in the flow during the original execution (see node.missing in the code)
- Simplify code

I don't know why sometimes jobs get left in running / queued state even after the execution has been killed. That's some other bug. This fix makes the UI behave nicely also with that kind of executions.

This fix was validated against some existing killed & failed executions. Some jobs/sub-flows that were not present in the flow during the original execution were enabled - after this fix they are not. Some jobs/subflows that were left in running or queued (although the flow itself has killed status) were disabled - with these fixes they are enabled as desired.


kunkun-tang:
@juhoautio Thanks for fixing bugs! We also encountered sometimes jobs get left in running / queued state even after the execution has been killed. It should be some bugs in the backend service, since I noticed that DB's state still kept running.

azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
```diff
@@ -496,18 +498,14 @@ var disableFinishedJobs = function(data) {
       node.status = "READY";
       node.disabled = true;
     }
-    else if (node.status == "SUCCEEDED" || node.status=="RUNNING") {
+    else if (node.status == "SUCCEEDED" || node.missing) {
```
kunkun-tang: I don't quite understand this statement. Now that node is SUCCEEDED, why we have to disable it?

azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
```diff
@@ -496,18 +498,14 @@ var disableFinishedJobs = function(data) {
...
       node.disabled = true;
     }
-    else if (node.status == "CANCELLED") {
```
kunkun-tang: Why this if clause could be removed ? will it be replaced by node.missing?

azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
juhoautio: Because if a job succeeded, we don't want to run that job again when preparing a new execution. Note that this function is only for the use case when you prepare a new execution based on a previous execution. Typically that only makes sense when some of the jobs were failed/cancelled/killed. If you prepare an execution based on a succeeded execution, then all jobs are initially disabled by this code.

azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
juhoautio: Not by `node.missing`. It's handled by the else branch below. If status is anything else than disabled, skipped or succeeded, the node is enabled (those two first status aren't shown in this diff because there was no change, but you can see them if you look a the whole file).

I'm removing `node.status="READY"` which would change the color of cancelled jobs from peach to normal grey. I think it's better that all job states (and thus colors) are kept from the original execution when preparing a new one. The original status/color is also kept for green and blue jobs (although succeeded jobs are disabled so their color is light green, not normal green).

kunkun-tang: Thanks a lot for education. Actually I thought **cancelled** is equal to FAILED_FINISHING.

kunkun-tang: The cancelled nodes will be still enabled after this unsuccessful execution. It makes sense to me that we should keep the peach color when preparing a new execution.

azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
```diff
@@ -300,6 +300,8 @@ azkaban.FlowExecuteDialogView = Backbone.View.extend({
         if (node.type == "flow" && statusNode.type == "flow") {
           this.assignInitialStatus(node, statusNode);
         }
+      } else {
```
kunkun-tang: Does node.missing refer to some  job being left in running or queued state incorrectly? Could you please add some comment in this else block?

juhoautio: This function (`assignInitialStatus`) traverses the jobs in the flow and looks up the same job in original flow to assign an initial status. If that lookup fails, it means that a new job has been added to the flow since the original execution. In that case there is no initial status to copy from the original execution, so we set `node.missing=true` to mark that.

I don't know if adding a comment in the code is the best way to improve. How about I rename this flag? For example, `node.noInitialStatus` or `node.notPresentInOriginalExecution`. What would you prefer?

kunkun-tang: Thanks. node.noInitialStatus looks good to me.

juhoautio:
Review changes:

- Renamed variable `missing -> noInitialStatus`
- Added comment _"job wasn't present in this flow during the original execution"_

PR #866, opened 2017-01-04
Fix Noop job failure by loading common.properties file to all jobtypes.

jamiesjc:
Currently default jobtypes do not load jobProps from common.properties file. Noop job, which is one of the default jobtypes, fails to resolve some parameters configured in common.properties. 
According to the document, common.properties should be global to all jobtypes including the default jobtypes. And it should be safe to do so because these are admin configured properties.
With the fix, default jobtypes will also load jobProps from common.properties file and it's verified on the testing cluster.


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -340,8 +340,13 @@ public Job buildJobExecutor(String jobId, Props jobProps, Logger logger)
...
 
-      // TODO: should the logic below mirror the logic for PluginLoadProps?
       Props pluginJobProps = pluginSet.getPluginJobProps(jobType);
+      // For default jobtypes, even though they don't have pluginJobProps configured,
+      // they still need to load properties from common.properties file if it's present
+      // because common.properties file is global to all jobtypes.
+      if(pluginJobProps == null) {
+        pluginJobProps = pluginSet.getCommonPluginJobProps();
+      }
       if (pluginJobProps != null) {
```
kunkun-tang: The best practice for me should be written using if else structure.

If (props == null) {
// For default job type, like javaprocess, noop, python.
} else {
// For custom plugin type, like hadoop, hive
}

kunkun-tang: Another thing is that custom jobtype will still read jobprops defined in azkaban-plugin, right? You might want to emphasize it here in comments.

jamiesjc: The if clause starting in line 350 actually applies to both default and custom plugin jobtypes. If changed to if else structure, then for default jobtypes, this part will be skipped and jobProps will not be populated for default jobtypes.

Issue #862, opened 2017-01-02
Inconsistent behavior: Downstream jobs are cancelled for a upstream job which is still in retry due to failure.


hkiran:
Explanation:
SAMPLE FLOW : TEST_FLOW_A
	
	JOB_A
	  |
	JOB_B
	  |
	JOB_C
	
JOB_B and JOB_C are cancelled due to JOB_A failure. But the problem is JOB_A still retry. Ideally the downstream jobs should only be cancelled after all the retries are exhausted. This issue does not happen every time/run.

LOGS:
INFO - Running flow 'TEST_FLOW_A'.
INFO - Submitting job 'TEST_FLOW_A:JOB_A' to run.
INFO - Cancelling 'TEST_FLOW_A:JOB_B' due to prior errors.
INFO - Cancelling 'TEST_FLOW_A:JOB_C' due to prior errors.
INFO - Flow 'TEST_FLOW_A' is set to KILLED in 0 seconds
INFO - Job TEST_FLOW_A:JOB_A finished with status FAILED in 0 seconds
INFO - Submitting job 'TEST_FLOW_A:JOB_A' to run.

Fix:
I have tested the below code fix and the same is working. However, please review the same and let me know if my assumptions are correct. Fix code is highlighted in bold.

   **private boolean isInRetry(ExecutableNode node) {
		if (node.getRetries() > node.getAttempt()) {
			return true;
		} else {
			return false;
		}
	}**

  /**
   * Determines what the state of the next node should be. Returns null if the
   * node should not be run.
   *
   * @param node
   * @return
   */
  public Status getImpliedStatus(ExecutableNode node) {
    // If it's running or finished with 'SUCCEEDED', than don't even
    // bother starting this job.
    if (Status.isStatusRunning(node.getStatus())
        || node.getStatus() == Status.SUCCEEDED) {
      return null;
    }

    // Go through the node's dependencies. If all of the previous job's
    // statuses is finished and not FAILED or KILLED, than we can safely
    // run this job.
    ExecutableFlowBase flow = node.getParentFlow();
    boolean shouldKill = false;
    for (String dependency : node.getInNodes()) {
      ExecutableNode dependencyNode = flow.getExecutableNode(dependency);
      Status depStatus = dependencyNode.getStatus();
      if (!Status.isStatusFinished(depStatus) **|| (depStatus == Status.FAILED && isInRetry(dependencyNode))**) {
        return null;
[...]

PR #867, opened 2017-01-05
do not cancel the downstream jobs if the failed job is still in retry

hkiran:
Details about the issue and fix is present in #862

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -745,6 +745,20 @@ public class FlowRunner extends EventHandler implements Runnable {
     }
     ;
   }
+  
+  /**
+   * Check if the node is still in retry.
+   * 
+   * @param node
+   * @return
+   */
+  private boolean isInRetry(ExecutableNode node) {
+    if (node.getRetries() > node.getAttempt()) {
+	  return true;
+    } else {
+      return false;
+    }
+  }
 
   /**
    * Determines what the state of the next node should be. Returns null if the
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -770,7 +784,7 @@ public Status getImpliedStatus(ExecutableNode node) {
       ExecutableNode dependencyNode = flow.getExecutableNode(dependency);
       Status depStatus = dependencyNode.getStatus();
 
-      if (!Status.isStatusFinished(depStatus)) {
+      if (!Status.isStatusFinished(depStatus) || (depStatus == Status.FAILED && isInRetry(dependencyNode))) {
```
HappyRay: I am not sure why the node status will be failed if it is being retried.
azkaban.execapp.FlowRunner#retryJobIfPossible should have set the status to READY again.

HappyRay: Based on the example you gave in the issue #862, here is my interpretation of what may have happened, please correct me if I am wrong.

JobB finished successfully first, it is added to the finishedNodes.  azkaban.execapp.FlowRunner#progressGraph adds JobC to the nodesToCheck list. azkaban.execapp.FlowRunner#runReadyJob calls azkaban.execapp.FlowRunner#getImpliedStatus, at this time, jobA fails it sets its status to failed in azkaban.execapp.JobRunner#runJob. But it has yet to have a chance to be added to the finishedNodes list so that retryJobIfPossible has not been called to reset the status. JobC is canceld because it sees that jobA has failed.

HappyRay:
Ping. Are you still interested?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
hkiran: @HappyRay  Sorry for the delay in repose. 
Yes. Your understanding is correct. Ideally the downstream jobs should only be cancelled after all the retries are exhausted.

HappyRay: Thanks. Could you update the change description to include more details such as what I described above?

inramana:
Following up on Ray's comment- @hkiran

Andre-BC:
What can I do for this fix to be applied ?

PR #869, opened 2017-01-06
Refactoring Azkaban SLA Trigger check logic

suvodeep-pyne:
Context:
Azkaban Triggers are checked via the SlaChecker class. The logic is failing for certain conditions where user flows would keep running despite SLA expiry. Though the logic seems to be working for most cases, it seems to be quite complicated + confusing with lot of code duplication.

Changes:
The `SlaChecker` class has been rewritten with unit tests to cover all the relevant code paths. This ensures that the Checker works correctly. 

\+ updated gradle wrapper to match LinkedIn internal version

### DIFF
--- a/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
@@ -107,6 +107,14 @@ public class SlaChecker implements ConditionChecker {
     return null;
   }
 
+  /**
+   * Initialize the instance.
+   *
+   * Initializing the derived members depends on ExecutorManager to be set.
+   * Also, initializing the start time is possible when the job has already started.
+   *
+   * @return false on caught exceptions else true.
+   */
   private boolean init() {
     try {
       requireNonNull(executorManager);
--- a/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
@@ -32,210 +32,184 @@ import azkaban.sla.SlaOption;
 import azkaban.trigger.ConditionChecker;
 import azkaban.utils.Utils;
 
+import static java.util.Objects.requireNonNull;
+
+
 public class SlaChecker implements ConditionChecker {
+  public static final String type = "SlaChecker";
+  enum NodeType { FLOW, JOB }
+  enum ConditionType { FINISH, SUCCEED }
 
   private static final Logger logger = Logger.getLogger(SlaChecker.class);
-  public static final String type = "SlaChecker";
-
-  private String id;
-  private SlaOption slaOption;
-  private int execId;
-  private long checkTime = -1;
-
   private static ExecutorManagerAdapter executorManager;
+
+  private final String id;
+  private final SlaOption slaOption;
+  private final int execId;
+
+  private ExecutableFlow flow = null;
+  private ExecutableNode job = null;
+  private NodeType nodeType;
+  private ConditionType conditionType;
+  private long startTime = -1;
+  private long slaExpireTime = -1;
+
+  /**
+   * TODO remove static linking
+   * @param executorManagerAdapter
+   */
+  public static void setExecutorManager(ExecutorManagerAdapter executorManagerAdapter) {
+    executorManager = executorManagerAdapter;
+  }
 
   public SlaChecker(String id, SlaOption slaOption, int execId) {
     this.id = id;
     this.slaOption = slaOption;
     this.execId = execId;
+
+    retrieveSlaTypeProperties();
   }
 
-  public static void setExecutorManager(ExecutorManagerAdapter em) {
-    executorManager = em;
+  private void retrieveSlaTypeProperties() {
+    switch (slaOption.getType()) {
+      case SlaOption.TYPE_FLOW_FINISH:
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -32,210 +32,184 @@
...
   }
 
+  /**
+   * TODO Remove convoluted logic. Investigate the need for 2 separate methods for pass and fail test
+   * DO NOT DELETE METHOD. This is used by some weird reflection code using JEXL expression
+   *
+   * @return
+   */
+  @SuppressWarnings("unused")
   public Object isSlaFailed() {
```
chengren311: shouldn't it return boolean?

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -32,210 +32,184 @@
...
   }
 
+  /**
+   * TODO Remove convoluted logic. Investigate the need for 2 separate methods for pass and fail test
+   * DO NOT DELETE METHOD. This is used by some weird reflection code using JEXL expression
+   *
+   * @return
+   */
+  @SuppressWarnings("unused")
   public Object isSlaPassed() {
```
chengren311: same here

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
suvodeep-pyne: This is some weird legacy code which is called via JEXL. It's cryptic and will break if I make any function signature changes.

I think we might need to rewrite the Trigger Manager. This is kind of a spot fix.

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
suvodeep-pyne: Please see my comment above.

The APIs are kind of weird. there is a passed as well as failed API and they vary slightly. It is a quite confusing.

kunkun-tang:
Have you figured out why SLA expired ?

suvodeep-pyne:
@kunkun-tang 
I haven't been able to reproduce the exact bug since it happened once every 2 months. The trigger management seems to be working for normal cases.

The logic is that whenever SLA expires, Azkaban should launch the corresponding trigger action. There can be 2 places where SLA expire logic can fail. 
- Trigger Scheduling logic: The trigger was never launched
- The Sla Check logic didn't expire correctly

This change is to ensure that the SLA check performs as expected. Should we see this problem again, my strong suspect would be the trigger management code itself.

kunkun-tang:
Thanks. I see. @suvodeep-pyne

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -32,210 +32,184 @@
...
-    } else if (type.equals(SlaOption.TYPE_JOB_SUCCEED)) {
-      String jobName =
-          (String) slaOption.getInfo().get(SlaOption.INFO_JOB_NAME);
-      ExecutableNode node = flow.getExecutableNode(jobName);
-      if (node.getStartTime() < 0) {
-        return Boolean.FALSE;
+    return null;
+  }
+
+  private boolean init() {
```
suvodeep-pyne: Added.

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -32,210 +32,184 @@
...
-      if (checkTime < node.getStartTime()) {
-        ReadablePeriod duration =
-            Utils.parsePeriodString((String) slaOption.getInfo().get(
-                SlaOption.INFO_DURATION));
-        DateTime startTime = new DateTime(node.getStartTime());
-        DateTime nextCheckTime = startTime.plus(duration);
-        this.checkTime = nextCheckTime.getMillis();
+      // Start time needs to be initialized if unset
+      if (startTime < 0) {
+        long t = getStartTime();
```
suvodeep-pyne: init() is called on every API call. However, once inited, these code blocks ar not executed. The trigger part is a different code, this class is just to check the condition. I suspect the code which calls these APIs also controls whether to stop the trigger.

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -32,210 +32,184 @@
...
+    return isSlaFailed();
   }
 
+  /**
+   * TODO Remove convoluted logic. Investigate the need for 2 separate methods for pass and fail test
+   * DO NOT DELETE METHOD. This is used by some weird reflection code using JEXL expression
+   *
+   * @return
+   */
+  @SuppressWarnings("unused")
```
suvodeep-pyne: yes. there are multiple code paths and am not sure of all of them. The reason I haven't changed the public APIs is because of JEXL. If / when I touch the trigger management code itself, I can refactor the logic and clean up a lot of the complexities. For now, I'm leaving them as is.

HappyRay:
Discussed the PR in person. The eval method will be called by the scheduling thread in a regular internal. That part hasn't changed.

LGTM.

Thanks for adding unit tests and making the code easier to read and maintain!

Issue #856, opened 2016-12-15
date.js modifyStrToUnixCronSyntax() method returns incorrect nth day of week


jh3155:
Relating to the issue #845, nth day of week becomes n-1th day of week due to replacing all numbers in the cron expression that are in range 1-7 to 0-6. This does not look like a bug in later.js. One possible fix would be replacing the numbers up to the pound sign.

NaanProphet:
@kunkun-tang Happy new year! Looks like this bug is in date.js rather than later.js. If you can confirm, maybe @jh3155 can create a pull request?

kunkun-tang:
@jh3155 @NaanProphet Still confused. Could you have a concrete example show what bug it is ?

mariacioffi:
@kunkun-tang You could reference issue #845 for an example.

kunkun-tang:
@mariacioffi I couldn't see any correlation with #845. I'm able to see what happened on #845, but confused on this issue. If you are able to fix the bug #845 , feel free to open a pull request. I will review.

jh3155:
@kunkun-tang Happy new year!
I will make a pull request

Issue #845, opened 2016-12-01
Next 10 scheduled executions on schedule-panel.vm is incorrect for expressions containing # 


NaanProphet:
Suppose one wants to schedule a job for every third Tuesday of the month at 4:00. The cron expression would be 00 4 ? * TUE#3. Under the "next 10 scheduled executions" heading however, we see 
2016-12-13T04:00:00
2017-01-10T04:00:00
2017-02-14T04:00:00
2017-03-14T04:00:00
...
 It appears as though the jobs with be scheduled for every **second** Tuesday of the month. Once the schedule button is hit though, the next execution time shows that the job will execute on the correct days, in this case it would be the third Tuesday of the month or 2016-12-20T04:00:00.

kunkun-tang:
we use later.js to parse cron expression in UI, but use Quartz-Cron to trigger in our back-end service. The UI might have some bugs, but it will execute as you like if it is following the correct Quartz-Cron grammar.

kunkun-tang:
Closing as this a bug for later.js

PR #870, opened 2017-01-06
Fixing date.js function issue #845 and #856

mariacioffi:
This changes the method modifyStrToUnixCronSyntax in date.js so that the "#" syntax can be used in the day of week option and the correct dates will be displayed under "Next 10 scheduled executions" on the schedule-panel page.


mariacioffi:
I have added some test cases to test.js. Let me know if there is anything missing or any other modification I should make. Thanks.

HappyRay:
@mariacioffi Thanks for this PR.

For reference, # syntax is documented at [http://www.quartz-scheduler.org/api/2.2.1/org/quartz/CronExpression.html]

PR #872, opened 2017-01-10
Smtp email port fix

wangqiaoshi:
add AbstractMailerTest base on smtp_email_port_fix branch of @sathish316


HappyRay:
Please reference the existing PR on which yours bases.

HappyRay:
Please describe the PR in more details in your change description.

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by wqs on 17/1/9.
+ */
+public class AbstractMailerTest {
+
+    @Test
+    public void testCreateEmailMessage(){
```
HappyRay: Add a test for default port?

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by wqs on 17/1/9.
+ */
+public class AbstractMailerTest {
```
HappyRay: Add a similar test for Emailer?

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -0,0 +1,30 @@
+package azkaban.utils;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by wqs on 17/1/9.
```
HappyRay: Please remove the author information in the code. It's already in Git.

HappyRay:
#395

wangqiaoshi:
1.add EmailerTest  and AbstractMailerTest for test smtp.port 
2. remove the author information in the code of AbstractMailerTest.java
base on PR #395

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -1,18 +1,40 @@
+/*
+ * Copyright 2014 LinkedIn Corp.
```
HappyRay: Change to 2017

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -21,10 +43,23 @@ public void testCreateEmailMessage(){
...
         AbstractMailer mailer = new AbstractMailer(props);
-        List<String> senderList = new ArrayList<String>();
-        senderList.add("sender@xxx.com");
-        EmailMessage emailMessage = mailer.createEmailMessage("subject","text/html",senderList);
+        emailMessage = mailer.createEmailMessage("subject","text/html",senderList);
 
         assert emailMessage.getMailPort()==445;
+
+
+        //for default
```
HappyRay: Clarify that this is for testing default port?

azkaban-common/src/test/java/azkaban/utils/AbstractMailerTest.java
```diff
@@ -21,10 +43,23 @@ public void testCreateEmailMessage(){
...
+        Props defaultProps = new Props();
```
HappyRay: Consider making it a separate test? 
And reduce duplicate code?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -0,0 +1,56 @@
+package azkaban.utils;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by wqs on 17/1/11.
```
HappyRay: Remove author name.

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public class EmailerTest {
```
HappyRay: This test is almost identical to the AbstractMailer test. Can you think of a way to consolidate?

HappyRay:
Thanks for your quick response.

Please update your branch.

Did you run an end to end default port test?

wangqiaoshi:
Thanks for your quick response and help. @HappyRay 
1.AbstractMailer  add getMailPort function
2.modify  2014  to 2017 in AbstractMailerTest
3.change EmailerTest(Emailer load Props from config file)

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -1,56 +1,75 @@
...
-        List<String> senderList = new ArrayList<String>();
-        senderList.add("sender@xxx.com");
-        EmailMessage emailMessage = null;
+    public void testCreateEmailMessage() throws Exception{
+
+        URL url = Resources.getResource("conf/emailtest/azkaban.properties");
+        InputStream inputStream = new FileInputStream(url.getFile());
 
+        Properties properties = new Properties();
+        properties.load(inputStream);
```
HappyRay: What's the advantage of loading the properties from the resource file instead of your previous approach?

HappyRay: Your previous approach seems to be easier to read. The tests for Emailer and AbstractEmailer are very similar except the class.  Why do you use different styles for them? It would be great if you can consolidate them. But I am fine with some code duplication if the tests are still easy to read.

HappyRay:
Manually tested the default port in a test environment.

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
wangqiaoshi: The reason of loading the properties from the resource I modify is for  making  difference AbstractMailer  between Emailer.
Sorry,I do not understand what do you mean to consolidate the EmailTest code.

wangqiaoshi:
add `send email  manually` in MailerTest
move load properties for resource file in MailerTest

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -0,0 +1,118 @@
...
+        project.setVersion(123);
+    }
+
+
+
+
+    /**
+     * test send email
+     */
+    @Ignore
```
HappyRay: Why is this test ignored? If this test is not complete, I'd suggest removing it.

wangqiaoshi: Because sender and failAddr now is not real.if real,will remove @Ignore

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -0,0 +1,117 @@
...
+        Assert.assertEquals(0, loader.getErrors().size());
+        project.setFlows(loader.getFlowMap());
+        project.setVersion(123);
+    }
+
+
+
+
+    /**
+     * test send email
```
HappyRay: I saw that this test is no longer ignored. How does it work? Where is the email actually delivered?

I think it is a fine idea to create a test to make end to end testing email sending feature easier. In that case it should be clearly documented how that test should be customized to fit the test environment and run.

HappyRay: Looks like this test will throw an exception of socket time out error as I expected. But since the exception is caught and only logged, the test continue to succeed.

So even if the port number logic is long, this test will still pass. 

I am not sure the added test run time is worth the benefit. 

It would be great if you can add instruction/comment for the test to allow it to actually send a real email and disable this test so that it will only run manually. 

If not, I'd suggest removing it.

wangqiaoshi: I very much agree with your point of  adding comment for sending email test and  I will add comment for it.

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -67,16 +67,18 @@ public void setUp() throws Exception {
     }
 
 
-
-
     /**
-     * test send email
+     * (test send email)
```
HappyRay: How about clarifying further that this test is an integration test?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -67,16 +67,18 @@ public void setUp() throws Exception {
...
+     * if you want to run this case,
+     * please make sure these variable{host,mailPort,password,receiveAddr} effective.
```
HappyRay: Do you mean
please make sure these variable{host,mailPort,password,receiveAddr} are set to real values that would allow an email to be sent successfully?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -67,16 +67,18 @@ public void setUp() throws Exception {
...
+     *
```
HappyRay: I would prefer that the test to fail when these varaibles are not set or set incorrectly. But I am ok with the current implementation for now.

Maybe also mention that this test will currently succeed because email sending errors are caught. To verify email functionality is working, users need to manually verify that a real email is sent and received?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
wangqiaoshi: yes,it is.

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
wangqiaoshi: Well, these are needed

wangqiaoshi: and i add annotation for variable{host,mailPort,password,receiveAddr} .

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
wangqiaoshi: ok.

wangqiaoshi:
perfect annotation for testSendEmail of Emailer.

HappyRay:
Thanks!

PR #877, opened 2017-01-17
Revert "Upgraded Jetty from 6.1.26 to 9.3.14"

suvodeep-pyne:
Reverts azkaban/azkaban#858

Jetty 9.3.14 seems to have an issue while interacting via SSL with curl (7.19.7). Also tested with 9.4.0.v2016.12.08, it still seems to have an issue. While curl on mac (7.43.0) seems to work fine.

Reverting this should fix these issues while we figure out a solution for this.


suvodeep-pyne:
Thanks @HappyRay ! :)

Issue #878, opened 2017-01-18
ProjectManager setJobOverrideProperty throws null Exception


wangqiaoshi:
when add a job's properties before is not existing,then code
`String diffMessage = PropsUtils.getPropertyDiff(oldProps, prop);`
will throws null pointer exception.

I think  when oldProps is null,need new empty Props.

quote
https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/ProjectManager.java

PR #883, opened 2017-01-22
repair ProjectManager setJobOverrideProperty throws null Exception #878

wangqiaoshi:
in order to avoid throw exception when oldProps is null, add add `oldProps = new Props()` code.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -360,6 +360,7 @@ public class ProjectManager {
         projectLoader.fetchProjectProperty(project, prop.getSource());
 
     if (oldProps == null) {
+      oldProps = new Props();
       projectLoader.uploadProjectProperty(project, prop);
     } else {
       projectLoader.updateProjectProperty(project, prop);


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -360,6 +360,7 @@ public void setJobOverrideProperty(Project project, Props prop, String jobName,
         projectLoader.fetchProjectProperty(project, prop.getSource());
 
     if (oldProps == null) {
+      oldProps = new Props();
       projectLoader.uploadProjectProperty(project, prop);
```
kunkun-tang: @wangqiaoshi the if else looks redundant. Is it possible that we just move projectLoader.uploadProjectProperty(project, prop) outside of if else logics ?

wangqiaoshi: i agree with you point,now we don't need this logic,but later?

wangqiaoshi: yes,this code is redundant.the oldProps can not be null.
i remove it

HappyRay:
@kunkun-tang can this PR be merged if it is updated?

kunkun-tang:
I was revisiting this PR, but can not figure out how NullPointerException is brought by the exiting code. @wangqiaoshi Could you please give us a concrete example about this bug?

HappyRay:
I don't see the problem with the original code either. 
How would the new code work when the oldProps is null? Wouldn't the sql update command fail?

inramana:
@wangqiaoshi reminder to follow up on @kunkun-tang query.

PR #888, opened 2017-01-30
Have scheduled flow's job disabled/enabled status be picked up by next execution.

chengren311:
When manually start a flow, the initial job disabled/enabled state ( if a job is disabled or not ) should come from the last ( and only ) schedule of the flow if any modification to job states are made. The current behavior is that the initial state will be the state when the flow is uploaded. However when a job state is changed in a manual execution, the state is not persisted.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -174,7 +174,6 @@ public class FlowRunnerManager implements EventListener,
 
     this.validateProxyUser = azkabanProps.getBoolean("proxy.user.lock.down", false);
 
-    setActive(true);
 
     cleanerThread = new CleanerThread();
     cleanerThread.start();


kunkun-tang:
Have you evaluated this change in staging cluster?

chengren311:
i tested in solo mode

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -928,7 +928,7 @@ public void cancelExecutingJobs(ExecutableFlow exFlow, String userId,
     }
   }
 
-  private void applyDisabledJobs(List<Object> disabledJobs,
+  public static void applyDisabledJobs(List<Object> disabledJobs,
```
kunkun-tang: not sure whether the change is appropriate. A util class should be more suitable to be put.

chengren311: I think it's fine. IMO the class contains a few util method, and they are supposed to be declared as static as well.

suvodeep-pyne:
> The current behavior is that the initial state will be the state when the flow is uploaded. 

I think that's a good behavior. If you upload a new zip, you can pretty much change the entire flow and all its jobs. I am not entirely sure how will you preserve state if the jobs don't exist in the new flow ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
suvodeep-pyne: If that's the case, then they should be moved to a util class. `Utils` in general is a tricky thing in terms of coding practices. http://stackoverflow.com/questions/3340032/utility-classes-are-evil

HappyRay:
@suvodeep-pyne When a new flow is uploaded, I believe the current behavior is that everything will be reset. But the executed flow will use the updated version if the flow still exists.

kunkun-tang:
#892 re submitted this pull request.

PR #891, opened 2017-02-01
active is initialized to false when constructing FlowRunnerManager.

chengren311:
FlowRunnerManager should be inactive until explicitly activating it from API.


HappyRay:
Could you consider making the description more descriptive?
e.g.
Prevent a new inactive executor instance from deleting old project files.
And then describe the details.

Also explain that this variable only indicate if the executor is active or not. The flow manager will still serve requests regardless.

chengren311:
@HappyRay updated

PR #892, opened 2017-02-01
Have scheduled flow's job diabled/enabled status be picked up by next execution.

chengren311:
When manually start a flow, the initial job disabled/enabled state ( if a job is disabled or not ) should come from the last ( and only ) schedule of the flow if any modification to job states are made. The current behavior is that the initial state will be the state when the flow is uploaded. However when a job state is changed in a manual execution, the state is not persisted.


chengren311:
@kunkun-tang updated

HappyRay:
I don't see an automated test included. Is it possible to do?

PR #894, opened 2017-02-03
Use hard link in execution dir to prevent flow failure where executor is reading deleted project dirs.

chengren311:
Previously we use symlink in execution dir. Symlink will be invalidated and execution will thus fail when source project dir is deleted by other executors when multiple executors are running.


HappyRay:
* Have you consider using a command "cp -arl" instead?
* Could you add automated tests?
* The logic for running a shell command can probably be factored out to its own reusable component.
* Since we are not using soft link, I'd suggest we only add it when needed.

HappyRay:
Could you add more details to the description explaining why the other executor may delete the project files being used by another process?

HappyRay:
The design relies on external programs to call the new APIs to make sure only one executor instance on a given host is active. Only the active one will delete old project files. 

The following is still possible:
Instance A starts downloading a project.
Before A finishes, B is deployed and made active. B starts downloading the same project.
Because the download process will download the project to a temp location first and do a file move atomically, only one copy will be preserved.
How will this change work in that scenario?
Am I right that depending on timing, A may hard link to the first copy or the second? No deletion should take place since this is a new project.  I think it will be fine. What do you think? 

Could you add this design consideration in the change description as well?

chengren311:
@HappyRay  updated.
Have you considered using a command "cp -arl" instead? 
Thanks for suggestion! The current method works fine, and replacing it with new command requires extra overhead like testing/refactoring. So I will add it to todo item in the future if we want to refactor our code but for this time i think it's ok to stick with current one.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -84,9 +87,10 @@ public static String getSourcePathFromClass(Class<?> containedClass) {
   }
 
   /**
-   * Run a unix command that will symlink files, and recurse into directories.
+   * Run a unix command that will hard link files and recurse into directories.
```
HappyRay: Could you add a comment that with this change, the project directory can't be stored in a different partition from the partition that the executions directory is in?

It is probably better to add it to the change description as well.

HappyRay:
Does it make sense to add a test that test the deep link behavior? i.e. delete the source dir and test the link still works?

chengren311:
updated per comment @HappyRay

.travis.yml
```diff
@@ -3,4 +3,4 @@ languages: java
 jdk:
   - oraclejdk8
 sudo: false
-script: ./gradlew clean build
+script: ./gradlew clean build
```
suvodeep-pyne: How does this help ?

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -84,9 +87,10 @@ public static String getSourcePathFromClass(Class<?> containedClass) {
...
    */
-  public static void createDeepSymlink(File sourceDir, File destDir)
+
+  public static void createHardDeeplink(File sourceDir, File destDir)
```
suvodeep-pyne: I'm kinda confused here. 
Are we creating-a-deep-hardlink or creating-a-hard-deeplink? :|

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
```diff
@@ -35,40 +37,60 @@
...
+    FileUtils.deleteDirectory(destDir);
   }
 
   @Test
-  public void testSymlinkCopy() throws IOException {
-    FileIOUtils.createDeepSymlink(sourceDir, destDir);
+  public void testHardlinkCopy() throws IOException {
+    FileIOUtils.createHardDeeplink(sourceDir, destDir);
+    assertTrue(areDirsEqual(sourceDir, destDir, true));
+    FileUtils.deleteDirectory(destDir);
```
HappyRay: Great. Thanks for adding this test.

.travis.yml
chengren311: not sure why this's included..

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
HappyRay: Yes, I also think createDeepHardLink makes more sense.

chengren311:
The reason we don't consider cp -arl is that the command doesn’t work for mac, and mac doesn’t have an easy way to do recursive hard linking. If we use cp -arl only, it would be infeasible to do local debugging.

.travis.yml
suvodeep-pyne: I see. Let's revert this change as and when possible.
Not urgent by any means.

PR #895, opened 2017-02-03
set concurrent option to CONCURRENT_OPTION_SKIP

wyukawa:
"Prepare Execution" button is very useful when job failed and we retry.

But sometimes, two people push button and then two jobs are executed because default current option is CONCURRENT_OPTION_IGNORE.

In almost case, I think that CONCURRENT_OPTION_SKIP is better


suvodeep-pyne:
Hey @wyukawa ! I understand your usecase. However, can you make the default configurable? Then we can keep the behavior as is.  While you can set your custom default.

Let me know what you think.

wyukawa:
Ah, It seems good.
I agree with you.
By the way, FailureAction is also same situation. I want to set FINISH_ALL_POSSIBLE.
Anyway, thanks!

PR #852, opened 2016-12-12
Provide a way for Azkaban to specify links to external URLs (#851)

KyleFung:
URLs, which are referenced by names known as a "topic" are specified in `azkaban.properties`
For example, if you want an external URL under the topic of "arbitraryName", you would put the following line into `azkaban.properties`: 

```
azkaban.server.external.arbitraryName.url=http://someURL.com
```

Additionally, an external log viewer can then be specified using the parameter 

```
azkaban.server.external.logviewer.topic=arbitraryName
```

This line would reference the URL specified earlier in this description (http://someURL.com) as an external log viewer.

A similar behavior has been added for specifying external analyzers, ie.
```
azkaban.server.external.analyzer.topic=arbitraryName
```
Note that this commit is **NOT backwards compatible** and deprecates the mechanism created in (#657). That commit specifies external analyzers with the config
```
execution.external.link.url=http://someURL.com/?%url
```

This must now be specified as
```
azkaban.server.external.arbitraryName.url=http://someURL.com/?${url}
azkaban.server.external.analyzer.topic=arbitraryName
```

PR #896, opened 2017-02-03
Change Kafka message syncing to false

suvodeep-pyne:
Kafka message syncing ensures that the messages are in order. However, the problem is that the syncing also slows down the transaction rate of these messages drastically. This affects the service significantly and is a blocking call. Reverting this change which was added to #852


suvodeep-pyne:
@HappyRay 
can you please let me know if this looks good. I added as much detail as I know about the issues caused by `sync`. Please feel free to do the same. I just want this problem to be recorded well enough because it was not that intuitive / obvious to me.

HappyRay:
LGTM. Thanks

PR #904, opened 2017-02-09
Implementating interfaces to collect and send app metrics.

kunkun-tang:
Recomposing Event Handler interface by writing some default methods to let subclasses easily implement this interface. Checked backward compatibitliy, and it should a safe change. Constucting a Listener Set class to wrap up different singleton listener by exploiting multiton design pattern. The underlying benefits of this use is to create one factory-like class to generate different singletons.

Adding quite a few event types as a kind of metrics.

Having WebMetricsListener, ExecMetricsListener and CommonListener to handle received metrics event. Since the previous implementation (WEBMetricsRegister etc.) is quite redundant. So Reconstructing.

This commit will kick off having the following appcalition metrics:
*    The rate of uploading file chunks to DB
*    The rate of user Downloading a file
*    The rate of DB AZ Downloading Files in executor
*    The rate of DB connection in both web and executor
*    the rate of Rest Get Call in Webserver
*    the rate of Rest Post Call in Webserver


kunkun-tang:
this PR was evaluated in local laptop, and worked.

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -63,8 +71,14 @@ public AbstractJdbcLoader(Props props) {
     dataSource = DataSourceUtils.getDataSource(props);
   }
 
+  @Override
+  public synchronized HashSet<EventListener> getListeners() {
```
suvodeep-pyne: Could you please add some comments here on what's happening.

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -16,19 +16,27 @@
...
+import java.util.HashSet;
 
 import org.apache.commons.dbutils.DbUtils;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.dbutils.ResultSetHandler;
 
-import azkaban.utils.Props;
 
-public abstract class AbstractJdbcLoader {
+public abstract class AbstractJdbcLoader implements EventHandler {
```
suvodeep-pyne: Do you think the JDBC Loader is playing too many roles here?
What are your thoughts on this? https://en.wikipedia.org/wiki/Composition_over_inheritance

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -63,8 +71,14 @@ public AbstractJdbcLoader(Props props) {
...
   }
 
+  @Override
+  public synchronized HashSet<EventListener> getListeners() {
+    return MultitonListenerSet.getInstance(MultitonListenerSet.ListenerType.COMMON, CommonMetrics.INSTANCE).getListeners();
+  }
+
   protected Connection getDBConnection(boolean autoCommit) throws IOException {
     Connection connection = null;
+    fireEventListeners(Event.create(this, Event.Type.DB_CONNECTION));
```
suvodeep-pyne: Does getDBConnection always create connections? If yes, should we rename the method to something like `createDBConnnection` ?

azkaban-common/src/main/java/azkaban/event/Event.java
```diff
@@ -26,7 +26,13 @@
     JOB_FINISHED,
     JOB_STATUS_CHANGED,
     EXTERNAL_FLOW_UPDATED,
-    EXTERNAL_JOB_UPDATED
+    EXTERNAL_JOB_UPDATED,
```
suvodeep-pyne: Some comments on each kind of event would help. 
cosmetic: AZ_DOWNLOAD_FILE and USER_DOWNLOAD_FILE seemed a bit confusing? What files? Are we talking about project zips?

azkaban-common/src/main/java/azkaban/event/Event.java
```diff
@@ -41,6 +47,13 @@ private Event(Object runner, Type type, EventData eventData) {
     this.time = System.currentTimeMillis();
   }
 
+  private Event(Object runner, Type type) {
```
suvodeep-pyne: code duplication.
reuse previous constructor rather than implementing the same logic again?

azkaban-common/src/main/java/azkaban/event/Event.java
```diff
@@ -71,4 +84,15 @@ public static Event create(Object runner, Type type, EventData eventData) throws
...
 
+  /**
+   * Creates a new event.
+   *
+   * @param runner runner.
+   * @param type type.
+   * @return New Event instance.
+   * @throws NullPointerException if EventData is null.
+   */
+  public static Event create(Object runner, Type type) {
```
suvodeep-pyne: Both the constructor and the static create method are very similar. What's the advantage of having a private constructor here?

azkaban-common/src/main/java/azkaban/event/EventHandler.java
```diff
@@ -16,28 +16,33 @@
...
+ * entirely refactor this interface.
+ * Having default interface methods helps implementation classes easily write necessary logics.
+ *
+ * TODO: Refactor return type in getListeners method.
+ */
+public interface EventHandler {
 
-  public EventHandler() {
-  }
+  HashSet<EventListener> getListeners();
```
suvodeep-pyne: The common convention is to have `Interface I = new Implementation();` so probably `Set<EventListener>` instead of `HashSet` ?

azkaban-common/src/main/java/azkaban/event/EventHandler.java
```diff
@@ -16,28 +16,33 @@
...
   }
 
-  public void fireEventListeners(Event event) {
-    ArrayList<EventListener> listeners =
-        new ArrayList<EventListener>(this.listeners);
-    for (EventListener listener : listeners) {
-      listener.handleEvent(event);
+  default void fireEventListeners(Event event) {
+
+    synchronized (this) {
```
suvodeep-pyne: We are synchronizing on a large code block here. 
Will we face any performance issues?

azkaban-common/src/main/java/azkaban/event/EventHandler.java
```diff
@@ -16,28 +16,33 @@
...
+      for (EventListener listener : getListeners()) {
+        listener.handleEvent(event);
+      }
     }
   }
 
-  public void removeListener(EventListener listener) {
-    listeners.remove(listener);
+  default void removeListener(EventListener listener) {
+    getListeners().remove(listener);
```
suvodeep-pyne: I suppose these methods are being accessed in a multi threaded scenario. In that case we need some protection if we are modifying listeners.
Should we consider alternative like concurrent lists as opposed to managing this ourselves? 
http://stackoverflow.com/questions/6916385/is-there-a-concurrent-list-in-javas-jdk

azkaban-common/src/main/java/azkaban/event/MultitonListenerSet.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Served for recognizing which listener Type.
+   */
+  public enum ListenerType {
+    WEB, EXEC, COMMON
+  }
+
+  private static final Map<ListenerType, MultitonListenerSet> _instances = new HashMap<>();
+  private HashSet<EventListener> listeners = new HashSet<>();
```
chengren311: Set<EventListener> listeners maybe is better

kunkun-tang:
Closing this PR, as new one is created in #913

PR #905, opened 2017-02-13
Add stacktrace when exception occurs in ExecutorServlet because user can check log.

wyukawa:


HappyRay:
If I understand your change correctly, you are adding stack trace to the log. Previously only the exception message is logged. Am I right? If so, could you clarify in your commit message?

wyukawa:
You are right. I changed the commit message.

By the way,when I encountered the following error, I did't understand the cause well.
This is the reason why I sent this pull request.
```
2017-02-09 16:08:38 INFO  ExecutorServlet:120 - User  has called action log on 3285
2017-02-09 16:08:38 ERROR ExecutorServlet:191 - azkaban.executor.ExecutorManagerException: Running flow 3285 not found.
2017-02-09 16:08:38 INFO  ExecutorServlet:120 - User  has called action log on 3285
2017-02-09 16:08:38 ERROR ExecutorServlet:191 - azkaban.executor.ExecutorManagerException: Running flow 3285 not found.
```

At this time, azkaban job was successful, it was not problem.
I guess this error is the concurrency problem related to ConcurrentHashMap(runningFlows).

HappyRay:
Thanks

wyukawa:
Thanks!

PR #906, opened 2017-02-13
Creating a new log4j pattern to log the current host.

kunkun-tang:
The current log4j doesn't support delivering hostname inside a log message. We construct a custom log4j pattern to address this problem. Basically, this new pattern is integrated with an existing pattern about fixing exception. We didn't change the existing pattern (PatternLayoutEscape) created by @KyleFung . After rolling this out to all clusters, we could remove that class.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
+++ b/azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
@@ -1,0 +1,112 @@
+package azkaban.utils;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import org.apache.log4j.helpers.PatternConverter;
+import org.apache.log4j.helpers.PatternParser;
+import org.apache.log4j.PatternLayout;
+import org.apache.log4j.spi.LoggingEvent;
+
+
+/**
+ * When we use the log4j Kafka appender, it seems that the appender simply does not log the stack trace anywhere
+ * Seeing as the stack trace is a very important piece of information, we create our own PatternLayout class that
+ * appends the stack trace to the log message that reported it, so that all the information regarding that error
+ * can be found one in place. In addition, we create a new PatternParser to deliver logs with current host name.
+ */
+public class AzkabanPatternLayout extends PatternLayout {
+  protected String host;
+
+  public AzkabanPatternLayout(String s) {
+    super(s);
+  }
+
+  public AzkabanPatternLayout() {
+    super();
+  }
+
+  // TODO: There is no easy way to get Jetty Server's port by InetAddress API.
+  // We might need to find some other way to find this application's port
+  protected String getPort() {
+    return "";
+  }
+
+  protected String getHostname() {
+    if (host == null) {
+      try {
+        InetAddress addr = InetAddress.getLocalHost();
+        this.host = addr.getHostName();
+      } catch (UnknownHostException e) {
+
+        // Prevent exposing exceptions when something goes wrong.
+        this.host = "localhost";
+      }
+    }
+    return host;
+  }
+
+  @Override
+  protected PatternParser createPatternParser(String pattern) {
+    return new PatternParser(pattern) {
+
+      @Override
+      protected void finalizeConverter(char c) {
+        PatternConverter pc = null;
+
+        switch (c) {
+          case 'h':
+            pc = new PatternConverter() {
+              @Override
+              protected String convert(LoggingEvent event) {
+                return getHostname();
+              }
+            };
+            break;
+        }
+        if (pc==null)
+          super.finalizeConverter(c);
+        else
+          addConverter(pc);
+      }
+    };
+  }
+
+  @Override
+  public String format(final LoggingEvent event) {
+    if (event.getMessage() instanceof String) {
+      return super.format(appendStackTraceToEvent(event));
+    }
+    return super.format(event);
+  }
+
+  /**
+   * Create a copy of event, but append a stack trace to the message (if it exists).
+   * Then it escapes the backslashes, tabs, newlines and quotes in its message as we are sending it as JSON and we
+   * don't want any corruption of the JSON object.
+   */
+  private LoggingEvent appendStackTraceToEvent(LoggingEvent event) {
+    String message = event.getMessage().toString();
+    // If there is a stack trace available, print it out
+    if (event.getThrowableInformation() != null) {
+      String[] s = event.getThrowableStrRep();
+      for (String line: s) {
+        message += "\n" + line;
+      }
+    }
+    message = message
+        .replace("\\", "\\\\")
+        .replace("\n", "\\n")
+        .replace("\"", "\\\"")
+        .replace("\t", "\\t");
+
+    Throwable throwable = event.getThrowableInformation() == null ? null
+        : event.getThrowableInformation().getThrowable();
+    return new LoggingEvent(event.getFQNOfLoggerClass(),
+        event.getLogger(),
+        event.getTimeStamp(),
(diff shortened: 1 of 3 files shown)


kunkun-tang:
Referencing #854.

Evaluated in Staging server, and it worked.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  public AzkabanPatternLayout() {
+    super();
+  }
+
+  // TODO: There is no easy way to get Jetty Server's port by InetAddress API.
+  // We might need to find some other way to find this application's port
+  protected String getPort() {
+    return "";
```
suvodeep-pyne: The Azkaban Server object has methods for obtaining the port information. However, we need to check if accessing that class directly is the best option.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  protected String getHostname() {
```
suvodeep-pyne: This may be duplicate code. The server has this information.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.log4j.spi.LoggingEvent;
+
+
+/**
+ * When we use the log4j Kafka appender, it seems that the appender simply does not log the stack trace anywhere
+ * Seeing as the stack trace is a very important piece of information, we create our own PatternLayout class that
+ * appends the stack trace to the log message that reported it, so that all the information regarding that error
+ * can be found one in place. In addition, we create a new PatternParser to deliver logs with current host name.
+ */
+public class AzkabanPatternLayout extends PatternLayout {
```
suvodeep-pyne: Is this specifically for the Kafka Appender. Then we should name it in that context.

azkaban-common/src/test/java/azkaban/utils/AzkabanPatternLayoutTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+
+/**
+ * Test output of AzkabanPatternLayoutTest
+ * It should be appending stack traces, escaping new lines, quotes, tabs and backslashes
+ * This is necessary when we are logging these messages out as JSON objects
+ */
+public class AzkabanPatternLayoutTest {
```
suvodeep-pyne: I have to say. I really like the testing strategy. This is awesome!

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
kunkun-tang: Those information is not in a static field, right?

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
kunkun-tang: I was thinking this class could be supplemented with other custom patterns, and then could be used inside azkaban-server.log as well.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
HappyRay: Also we may use a VIP name of the host. Another reason to get the host name from one place.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import org.apache.log4j.helpers.PatternConverter;
+import org.apache.log4j.helpers.PatternParser;
+import org.apache.log4j.PatternLayout;
+import org.apache.log4j.spi.LoggingEvent;
+
+
+/**
```
HappyRay: Did you rename this file and class name? It would be best in the future to rename the file first in a separate commit to make the diff easier to track.

kunkun-tang: I didn't touch the original file. This file is newly created. using this new class needs conf change as well.

kunkun-tang:
Waiting for merging #908 , where VIP name is implemented to fetch.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Override
+  protected PatternParser createPatternParser(String pattern) {
+    return new PatternParser(pattern) {
+
+      @Override
+      protected void finalizeConverter(char c) {
+        PatternConverter pc = null;
+
+        switch (c) {
```
chengren311: if there's only one case here, why bother to use switch case?

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+   * Then it escapes the backslashes, tabs, newlines and quotes in its message as we are sending it as JSON and we
+   * don't want any corruption of the JSON object.
+   */
+  private LoggingEvent appendStackTraceToEvent(LoggingEvent event) {
+    String message = event.getMessage().toString();
+    // If there is a stack trace available, print it out
+    if (event.getThrowableInformation() != null) {
+      String[] s = event.getThrowableStrRep();
+      for (String line: s) {
+        message += "\n" + line;
```
chengren311: we should use stringbuffer here

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
```
chengren311: actually did you copy this code from somewhere? it seems to be duplicate of PatternLayoutEscaped.appendStackTraceToEvent

kunkun-tang: Yea. This change is also a part of refactor. I can't remove the original class directly as Production configurations still need it. Will remove duplicates afterward in a separate commit.

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
kunkun-tang: Could you please share more details how ?

azkaban-common/src/main/java/azkaban/utils/AzkabanPatternLayout.java
kunkun-tang: I intended to create the switch structure to help more options being added easily in future. Will might have AZ version, AZ port attached to every log message.

HappyRay:
@kunkun-tang what's your plan for this PR?

kunkun-tang:
I would think we can close it now. @HappyRay @ameyamk Let me know if this PR still has value.

HappyRay:
I think this PR can be postponed for now. We can pick it up again in the future. How about labeling it and closing it for now?

kunkun-tang:
Closing now. This PR will act as a reference if we want to design new log4j pattern in future.

PR #908, opened 2017-02-15
Allow users to specify hostname alias in getHost()

chengren311:
User can fill in their own alias generating logic in the newly added perl script in resource dir. The perl script will only be executed once in the first time executing getHost(). The result of execution of the script will be used as the hostname. If there's error in executing the script we will use regular hostname as fallback.

### DIFF
--- a/azkaban-common/src/main/resources/host_name_alias.perl
+++ b/azkaban-common/src/main/resources/host_name_alias.perl
@@ -1,0 +1,45 @@
+#!/usr/bin/env perl
+
+# A script to get the VIP name of the current host.
+
+use strict;
+use warnings;
+
+# Get a list of IP addresses that the host has.
+sub get_ip_addrs() {
+    my @ip_addrs = ();
+    # Use the absolute path of ifconfig
+    # since the script may be invoked without PATH being set.
+    for my $line (`/sbin/ifconfig -a`) {
+        # Parse the output of the ifconfig command
+        # looking for patterns of IP addresses
+        if ($line =~ /.*inet addr:(\d+\.\d+\.\d+\.\d+).*/) {
+            push @ip_addrs, $1;
+        }
+    }
+    @ip_addrs;
+}
+
+sub get_hostname_for_ip {
+    my ($ip) = @_;
+    my $result = `getent hosts $ip`;
+    chomp($result);
+    (split(/ /, $result))[-1];
+}
+
+sub get_vip_hostname {
+    my $vip_hostname;
+    for my $ip (get_ip_addrs()) {
+        next if $ip eq '127.0.0.1';
+        $vip_hostname = get_hostname_for_ip($ip);
+        next if $vip_hostname =~ /hcl\d+/;
+    }
+    $vip_hostname;
+}
+
+sub get_shortname {
+    my ($name) = @_;
+    (split(/\./, $name))[0];
+}
+
+print get_vip_hostname(), "\n";
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -16,8 +16,13 @@
 
 package azkaban.execapp;
 
+import azkaban.utils.FileIOUtils;
 import com.google.common.base.Throwables;
 
+import com.google.common.io.Resources;
+import java.net.URISyntaxException;
+import java.net.URL;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.joda.time.DateTimeZone;
 import org.mortbay.jetty.Connector;
@@ -79,6 +84,7 @@ import static azkaban.constants.ServerInternals.AZKABAN_EXECUTOR_PORT_FILENAME;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
+
 public class AzkabanExecutorServer {
   private static final String CUSTOM_JMX_ATTRIBUTE_PROCESSOR_PROPERTY = "jmx.attribute.processor.class";
   private static final Logger logger = Logger.getLogger(AzkabanExecutorServer.class);
@@ -91,6 +97,7 @@ public class AzkabanExecutorServer {
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
   private static final int DEFAULT_THREAD_NUMBER = 50;
 
+  private static String hostAlias = null;
   private static AzkabanExecutorServer app;
 
   private final ExecutorLoader executionLoader;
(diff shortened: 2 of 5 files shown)


azkaban-common/src/main/resources/host_name_alias.perl
```diff
@@ -0,0 +1,45 @@
...
+
+use strict;
+use warnings;
+
+# Get a list of IP addresses that the host has.
+sub get_ip_addrs() {
+    my @ip_addrs = ();
+    # Use the absolute path of ifconfig
+    # since the script may be invoked without PATH being set.
+    for my $line (`/sbin/ifconfig -a`) {
```
li-afaris: "ifconfig" is being replaced by the "ip" command in linux distributions.  This will not work on newer debian or redhat/centos 7 distros. 

Also instead of using backticks, you want to use the system function for executing external commands, ("perldoc -f system"). System this will perform a fork and wait for the child in the background.  Additionally you can right shift $? by 8 to get the exit code of the command called by system. 

Another ifconfig (or ip) command replacement suggestion would be to read /proc/net/tcp for IPv4 and convert the hex strings to IPs and /proc/net/if_inet6 for IPv6 addresses

azkaban-common/src/main/resources/host_name_alias.perl
```diff
@@ -0,0 +1,45 @@
...
+        # Parse the output of the ifconfig command
+        # looking for patterns of IP addresses
+        if ($line =~ /.*inet addr:(\d+\.\d+\.\d+\.\d+).*/) {
```
li-afaris: IPv4 regex matching should bound the octets its only possible for 1-3 digits per octet.  For example: '\d{1,3}\.\d{1,3}\.\d{1,3}\.d{1,3}'  You can can condense this idea further but loose readability when having fun with regular expressions: "m!((\d{1,3}\.){3})(\d+)!" 

My other question, is  what happens if you get a ipv6 address? It's a possibility these days

azkaban-common/src/main/resources/host_name_alias.perl
```diff
@@ -0,0 +1,45 @@
...
+    chomp($result);
+    (split(/ /, $result))[-1];
+}
+
+sub get_vip_hostname {
+    my $vip_hostname;
+    for my $ip (get_ip_addrs()) {
+        next if $ip eq '127.0.0.1';
+        $vip_hostname = get_hostname_for_ip($ip);
+        next if $vip_hostname =~ /hcl\d+/;
```
HappyRay: This logic is specific to a particular naming convention. 

What do you think about adding a config to the config file and allow an external system to set the hostname? Fall back to lost host name detection as you we now.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -91,6 +97,7 @@
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
   private static final int DEFAULT_THREAD_NUMBER = 50;
 
+  private static String hostAlias = null;
```
suvodeep-pyne: Couple of Questions:
1. Should we let the server run if the obtained `hostAlias` is null or should we crash the server (since metrics and everything will break)?
2. Can `hostAlias` change during execution? Else should it be a `final` variable?

azkaban-common/src/main/resources/host_name_alias.perl
li-afaris: > What do you think about adding a config to the config file and allow an external system to set the hostname? Fall back to local host name detection as you we now.

This would be a good feature to have and avoid having to scrape & detect ipv4 vs ipv6 addresses.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
li-afaris: > Should we let the server run if the obtained hostAlias is null or should we crash the server (since metrics and everything will break)?

I think logging an exception and crashing the server is the way to go.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -551,12 +559,29 @@ public Object getMBeanAttribute(ObjectName name, String attribute) {
...
    */
-  public String getHost() {
-    String host = "unkownHost";
+  public static String getHost() {
+    if(StringUtils.isEmpty(hostAlias)) {
+      URL scriptUrl = Resources.getResource(ServerInternals.HOST_NAME_ALIAS_SCRIPT);
+      if (scriptUrl != null) {
+        try {
+          File aliasFile = new File(scriptUrl.toURI());
+          hostAlias = FileIOUtils.runShellCommand("perl " + aliasFile, new File(".")).trim();
```
suvodeep-pyne: Questions:
 - What are the advantages of using perl to do this vs directly doing this in Java?
 - probably should be abstracted in a separate method / class since there's a lot going on to determine the vip?

Some comments on what's going on here might help.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -551,12 +559,29 @@ public Object getMBeanAttribute(ObjectName name, String attribute) {
...
+      if(StringUtils.isEmpty(hostAlias) || scriptUrl == null) {
+        hostAlias = "unknownAlias";
+      }
+    }
+
+    String host = "unknownHost";
     try {
-      host = InetAddress.getLocalHost().getCanonicalHostName();
+      if(hostAlias.equals("unknownAlias")) {
+        host = InetAddress.getLocalHost().getCanonicalHostName();
```
kunkun-tang: Curious, what does getCanonicalHostName return in this case? VIP name?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -551,12 +559,29 @@ public Object getMBeanAttribute(ObjectName name, String attribute) {
...
+          hostAlias = FileIOUtils.runShellCommand("perl " + aliasFile, new File(".")).trim();
+        } catch (Exception e) {
+          logger.error("Failed to get alias", e);
+        }
+      }
+      if(StringUtils.isEmpty(hostAlias) || scriptUrl == null) {
+        hostAlias = "unknownAlias";
+      }
+    }
+
```
kunkun-tang: Should we check the obtained URL a valid/legal VIP somewhere?  It might be helpful if perl gives a wrong vip name.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
shuzhang1989: fqdn name maybe

PR #911, opened 2017-02-16
Enable to set SLA Option when user schedules

wyukawa:
Although we can set execution options, we can't set SLA Option when user schedules.

If we can set SLA Option, it would be useful.


wyukawa:
@HappyRay Could you check this pull request?

kunkun-tang:
This change makes sense to me. Users need to set SLA options during scheduling.  Have you evaluated these changes @wyukawa ?

wyukawa:
Yes.
In my use case, I add a schedule through api.
In addition, I want to set SLA.

wyukawa:
rebase done

wyukawa:
ping?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -671,6 +676,11 @@ private void ajaxScheduleFlow(HttpServletRequest req,
     }
 
     List<SlaOption> slaOptions = null;
+    try {
+      slaOptions = getSlaOptionList(req, flowName);
+    } catch (Exception e) {
```
HappyRay: The getSlaOptionList throws a serveletException already. 
Could you explain why you would like to use a different error handling strategy?

wyukawa: I adapt the following coding style in ScheduleServlet
```
    ExecutionOptions flowOptions = null;
    try {
      flowOptions = HttpRequestUtils.parseFlowOptions(req);
      HttpRequestUtils.filterAdminOnlyFlowParams(userManager, flowOptions, user);
    } catch (Exception e) {
      ret.put("error", e.getMessage());
    }
```
Of course, it's possible to catch ServeletException, instead of Exception

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -138,27 +138,7 @@ private void ajaxSetSla(HttpServletRequest req, HashMap<String, Object> ret,
...
-          throw new ServletException(e);
-        }
-        if (sla != null) {
-          sla.getInfo().put(SlaOption.INFO_FLOW_NAME, sched.getFlowName());
-          sla.getInfo().put(SlaOption.INFO_EMAIL_LIST, slaEmails);
-          slaOptions.add(sla);
-        }
-      }
-
+      List<SlaOption> slaOptions = getSlaOptionList(req, sched.getFlowName());
```
wyukawa: Thanks!

wyukawa:
Thank you for comment!

First of all, let me explain my use case.

In my use case, I set schedule date through API, not using web browser because I want to maintein schedule informaion in GitHub.

Our Azkaban job management is the following
https://www.slideshare.net/wyukawa/azkabanen/17

And I set Failure Options when I set schdule data.

Although Azkaban default setting of Failure Options is "Finish Current Running", I want to set "Finish All Possible".

As a result, I use schedule API and then set schedule information and Failure Options at the same time.

I created Azkaban API Client "eboshi" (https://github.com/wyukawa/eboshi) and use the following way.
```
eboshi addCronSchedule --url http://localhost:8081 --username azkaban --password azkaban --project azkaban_project --flow azkaban_flow --cron '0 10 * * *' --option '{"failureAction":"finishPossible"}'
```

In addition to this, I want to set SLA option.

In other words, I want to set schedule information(for example, '0 10 * * *') and Failure Options and SLA option at the same time.

This is the reason why I send this pull request.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
HappyRay: Both styles have issues. Server should return 4xx error in this case. It's something we should fix in the future. I am interested in your thoughts on how to fix it.

HappyRay:
I understand your use case. Nice presentation. Thanks for being a champion for the project!

One of the benefits of keeping the API surface area small is to make automated testing simpler. We would like to have automated tests to cover all APIs. 

I would prefer returning the schedule id in this API which is useful anyway regardless. Is it too much work on the client side or multiple call latency a big issue in your case?

Any code refactoring/cleaning up contributions are always welcome.

wyukawa:
Thank you for comment.

Your thought is the following.

Current schedule APIs(ajaxScheduleFlow and ajaxScheduleCronFlow method) return the schedule Id.
So user can use SLA API(ajaxSetSla method ) by using this schedule Id.

Although user need to call two APIs, APIs are simple and easy to test.
And SLA should be set after a schedule is set.

Is that right?

If I understand correctly, you are right and I accept this pull request doesn't be merged.

>Is it too much work on the client side or multiple call latency a big issue in your case?

Maybe No

If I have a time, I will try to create SLA API document.

Thanks

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
wyukawa: Hmm, generally speaking, if user requests wrongly, server should return 4xx(for example, 400 Bad Request).
If there are some bugs, server should return 5xx.

But before you fix it in that way, I prefer to add stacktrace(for example, https://github.com/azkaban/azkaban/pull/890, https://github.com/azkaban/azkaban/pull/905)

Regarding stacktrace, I will to send pull request later.

HappyRay:
I am glad we are on the same page. :)
Any contribution to documentation will be appreciated. 
The refactoring change you made makes sense to me too.

Correction:
I checked the code and saw that scheduleId is returned by the ajaxScheduleCronFlow and ajaxScheduleFlow APIs. 

cc @kunkun-tang

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
wyukawa: Regarding stacktrace, I sent pull request.
https://github.com/azkaban/azkaban/pull/962

wyukawa:
Thank you for your clarification

PR #873, opened 2017-01-10
Fixing advanced filter by project on history page, addresses issue #403

mariacioffi:
Changed the SQL statement to avoid ambiguous column: enc_type. This should correct the problem.

PR #915, opened 2017-02-17
Fix advanced history query error

shelocks:
Fix advanced history query error and logging stacktrace when error occurs in HistoryServlet.


HappyRay:
Hi, sorry for the delay.

Could you provide more details in the repro steps in the change description and how you verified the fix worked?

I am not sure I understand the problem and the fix. Could you elaborate in the change description?

The logging change looks good. 
Per google style, LOGGER should be named logger. 
Please see the udpated readme for the contribution guide.
Could you open a separate PR for the logging change? We can merge that one quickly.

HappyRay:
@shelocks Ping. Are you still interested in this PR?

HappyRay:
Looks like it overlaps with #873

inramana:
Close once #1638 is merged.

PR #919, opened 2017-02-21
if jobtype.global.classpath or jobtype.classpath contain wildcard of …

wangqiaoshi:
fix loadJobTypeClassLoader if jobtype.global.classpath,jobtype.classpath contains wildcard of *.


HappyRay:
Could you give an example?
Could you provide automated tests?

PR #923, opened 2017-02-22
add basic el function and load external custom el functions.

wangqiaoshi:
basic function is :
timestamp,year,month,day,dateFormat,there function is for current time.

we can add custom el function.
azkaban.properties  configuration params:

Parameter                                           Description
azkaban.el.lib                                     Comma delimited list of  el lib path
azkaban.el.func.${namespace}        package.className,it will be put  into namespace

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
@@ -28,6 +28,7 @@ import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import azkaban.el.ElManager;
 import org.apache.commons.jexl2.Expression;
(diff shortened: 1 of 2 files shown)


kunkun-tang:
I don't understand what el stand for. what is its full name? @wangqiaoshi

wangqiaoshi:
express language.

kunkun-tang:
That must be a quite interesting feature if we could have it implemented. Your idea is to let users define custom EL functions, right?

kunkun-tang:
@suvodeep-pyne Could our new DSL parser have similar trait?

wangqiaoshi:
yes, i think it will be useful to timing job.

wangqiaoshi:
Its implementation is also based on jexl. @suvodeep-pyne

Issue #924, opened 2017-02-23
SQL lookup table needed for reporting


li-afaris:
I'd like to add a lookup table to assist with converting the numeric values in the execution_flows.status field to string values from the status enum datatype in Status.java.  

Also existing sql statements don't have guards in place to prevent overwriting existing tables.  A PR will follow shortly with these features.

PR #925, opened 2017-02-23
Issue #924 - new lookup table and guards on table create SQL statements.

li-afaris:
As described in [issue #924](https://github.com/azkaban/azkaban/issues/924), this PR is to add a lookup table to allow backend DB queries to expand the numeric values in the execution_flows.status field.   Additionally guards were put in place to prevent 'create table' statements from overwriting existing tables.  I also specified that InnoDB should be the default DB engine for each table.  (InnoDB is the default in MySQL 5.5.5 & newer.)

Testing was done manually.  After running 'gradlew distTar', I ran the schema script on a new database, followed by mysqldump to confirm the contents looked sane.   If there's a better way to test sql statements, I would be happy to test them.  

Thanks.

### DIFF
--- a/azkaban-sql/src/sql/create.execution_flows.sql
+++ b/azkaban-sql/src/sql/create.execution_flows.sql
@@ -1,4 +1,4 @@
-CREATE TABLE execution_flows (
+CREATE TABLE if NOT EXISTS execution_flows (
 	exec_id INT NOT NULL AUTO_INCREMENT,
 	project_id INT NOT NULL,
 	version INT NOT NULL,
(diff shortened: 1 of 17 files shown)


li-afaris:
@HappyRay requested I split this PR into multiple commits as several changes are included with this PR.  Will close this PR and submit new ones containing single changes.

azkaban-sql/src/sql/create.execution_flows.sql
```diff
@@ -13,10 +13,11 @@ CREATE TABLE execution_flows (
 	flow_data LONGBLOB,
 	executor_id INT DEFAULT NULL,
 	PRIMARY KEY (exec_id)
-);
+)
+ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
HappyRay: Is there a concern with moving charset to utf8? I remember a user asked for UTF8 support to store international characters ( maybe in project names?).

PR #927, opened 2017-02-23
Issue #924 - new sql lookup table

li-afaris:
As disscussed in #924, I'd like to add a lookup table to allow backend DB queries to expand the numeric values in the execution_flows.status field.   This table is external to the azkaban application and is being added as a convenience for external reporting.  

For example, using an external program I may want to query the execution_flows table to see how many failed jobs we had in the last month?  Currently, I would have to copy the  status enum datatype values from Status.java into my script.  With the proposed lookup table, I can perform a join on the status column during my sql query and have these values printed as part of the SQL return statement.

Testing was done manually. After running 'gradlew distTar', I ran the schema script on a new database to create the new table.  This was followed by mysqldump to confirm the contents looked sane.

### DIFF
--- a/azkaban-sql/src/sql/create.execution_flows_status.sql
+++ b/azkaban-sql/src/sql/create.execution_flows_status.sql
@@ -1,0 +1,25 @@
+-- -----------------------------------------------------------------
+-- Create lookup table for 'status' values in execution_flows table
+-- -----------------------------------------------------------------
+CREATE TABLE if NOT EXISTS execution_flows_status (
+	name VARCHAR(64),
+	status INT,
+	PRIMARY KEY (name)
+);
+-- -----------------------------------------------------------------
+-- Populate enum values into lookup table
+-- -----------------------------------------------------------------
+INSERT INTO execution_flows_status (name, status) VALUES 
+  ("READY",10),
+  ("PREPARING",20),
+  ("RUNNING",30),
+  ("PAUSED",40),
+  ("SUCCEEDED",50),
+  ("KILLED",60),
+  ("FAILED",70),
+  ("FAILED_FINISHING",80),
+  ("SKIPPED",90),
+  ("DISABLED",100),
+  ("QUEUED",110),
+  ("FAILED_SUCCEEDED",120),
+  ("CANCELED",130);


azkaban-sql/src/sql/create.execution_flows_status.sql
```diff
@@ -0,0 +1,25 @@
+-- -----------------------------------------------------------------
+-- Create lookup table for 'status' values in execution_flows table
+-- -----------------------------------------------------------------
+CREATE TABLE if NOT EXISTS execution_flows_status (
+	name VARCHAR(64),
+	status INT,
+	PRIMARY KEY (name)
+);
+-- -----------------------------------------------------------------
+-- Populate enum values into lookup table
```
HappyRay: Consider adding a comment to the code where the enum is defined that this sql script needs to be updated when the enum is changed. And there is no automated test due to the low return on investment.

azkaban-sql/src/sql/create.execution_flows_status.sql
```diff
@@ -0,0 +1,25 @@
+-- -----------------------------------------------------------------
+-- Create lookup table for 'status' values in execution_flows table
```
HappyRay: Consider adding a comment similar to the pull request description so that readers know what this table is for.

li-afaris:
I'm going to close this PR without merging with the upstream azkaban project.  After submitting this PR, I discovered the embedded H2 database has a different grammar then mysql for inserting values into a table, and this change will break azkaban in solo-mode.  

My suggestion for a future PR to use directories to isolate the schema files for each supported database type,  (currently this is how it's done with the Apache Hive project).  Lookup tables should be added after the isolation occurs.

HappyRay:
@li-afaris It looks like h2 DB should support this insert syntax. See
http://www.h2database.com/html/quickstart.html
It worked for me.
What problem did you observe?

HappyRay:
@li-afaris I found out what the problem was.

H2/SQL requires the single quote. MySQL seems to also accept the double quote.

The following worked for me:

-- -----------------------------------------------------------------
-- Populate enum values into lookup table
-- -----------------------------------------------------------------
INSERT INTO execution_flows_status (name, status) VALUES
  ('READY',10),
  ('PREPARING',20),
  ('RUNNING',30),
  ('PAUSED',40),
  ('SUCCEEDED',50),
  ('KILLED',60),
  ('FAILED',70),
  ('FAILED_FINISHING',80),
  ('SKIPPED',90),
  ('DISABLED',100),
  ('QUEUED',110),
  ('FAILED_SUCCEEDED',120),
  ('CANCELED',130);

===
Also, does it make more sense to make the status column the primary key so that it is indexed? The intended use case is to join based on the status code, not name. Right?

li-afaris:
Nice find, I didn't think about attempting to replace the double quote with a single.  And yes I should use the status column as the primary key.  I'll make time to reopen this & update later today. Thanks

PR #929, opened 2017-02-28
Check if the process being killed is dead in shutdown scripts

chengren311:
After killing the process, shutdown scripts will check whether the process is dead and retry a few times if not.


azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
```diff
@@ -5,15 +5,30 @@ installdir="$(dirname $0)/.."
...
+if [ -z $pid ]; then
+  echo "currentpid file doesn't exist in ${installdir}, shutdown completed"
+  exit 0
 fi
 
-rm  ${installdir}/currentpid
-rm  ${installdir}/executor.port
+for try in $(seq 1 $maxtry); do
+  if [ ! -z $pid ]; then
+    echo "Killing Exec Server. [pid: $pid, port: $port], $try th try"
```
HappyRay: It's hard to get the grammar right. How about just "count: $try"?

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
```diff
@@ -2,14 +2,33 @@
 # Shutdown script for azkaban web server
 installdir="$(dirname $0)/.."
 
+#currdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
HappyRay: Remove unused code?

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
```diff
@@ -2,14 +2,33 @@
...
+    else 
+      rm  ${installdir}/currentpid
+      echo "shutdown succeeded"
+      exit 0
+    fi
+  fi
+done
+
+echo "Error: Shutdown failed"
+exit 1
```
HappyRay: Will it fail the deployment?

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
```diff
@@ -2,14 +2,33 @@
...
-    echo "Error: Shutdown failed"
-    exit 1;
+if [ -z $pid ]; then
+  echo "currentpid file doesn't exist in ${installdir}, shutdown completed"
+  exit 0
 fi
 
-rm  ${installdir}/currentpid
-echo "done."
+for try in $(seq 1 $maxtry); do
```
HappyRay: The logics for the web server and executor are the same. 
Can you think of a way to consolidate the code?

azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
```diff
@@ -5,15 +5,30 @@ installdir="$(dirname $0)/.."
...
 port=`cat ${installdir}/executor.port`
 
-echo "Killing Executor. [pid: $pid, port: $port]"
-
-kill ${pid}
-if [ $? -ne 0 ]; then
-    echo "Error: Shutdown failed"
-    exit 1;
+maxtry=5
+if [ -z $pid ]; then
```
li-afaris: Could you add 'set -o nounset' guard against referencing undefined variables?  You will also want to add 'set -o errexit' to force an exit if any shell-outs  return a non-zero exit.  For example, if the cat command fails to find the currentpid file, the script will continue running and $pid is unset which is referenced on line 9.  By using both 'set -o' options, we can guard against this ocuring by forcing an early failure.

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
li-afaris: Would it be possible to refactor these script to use functions? It will allow for code consolidation and easier readability.  Additionally one could start writing unit tests using BATs (or other test framework).

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
chengren311: so web server shutdown fails here, deployment should also fail since starting another instance of web server is risky. But if the deployment fails because of this, we still have to manually kill the web server by doing kill -9. So I think we should have kill -9 as the final fallback in the script.

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
chengren311: the two scripts will be in two different dirs and might even be in two different machines. Would it be possible to have them share one consolidated shell script?

chengren311: probably we can, for example have the shared script in azkaban common, and when installing the RPM, the shared script will be put somewhere for these two scripts to refer to, but i don't think it's worth doing this, considering shutdown scripts are so rarely changed.

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
chengren311: done

azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
chengren311: done

chengren311:
@HappyRay will update the change description before merge.

azkaban-exec-server/src/main/bash/azkaban-executor-shutdown.sh
chengren311: @li-afaris i think 'set -o errexit' might cause undesirable results. For example, killing a non-existing process will exit the shell with error code which might be captured by deployment script treating it as a deployment failure, but in this case we actually want to proceed deployment.

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
```diff
@@ -1,15 +1,18 @@
 #!/usr/bin/env bash
 # Shutdown script for azkaban web server
-installdir="$(dirname $0)/.."
+set -o nounset
+source "$(dirname $0)/util.sh"
 
+installdir="$(dirname $0)/.."
+maxattempt=5
```
HappyRay: Shall we fail faster? If the process won't shut down within 10 seconds, I would expect a problem.   Is it better to fail the operation and let operator take some manual action sooner? At this point, the service is probably unavailable already. It's better to react fast.

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
HappyRay: Yes, kill -9 would be an option. However, on balance I would rather warn the operator if the process can't be shut down and allow the operator to take manual actions. Otherwise, the failure will go unnoticed and it won't be much better than just do kill -9 always.

chengren311:
@HappyRay we can make the process hang with kill -STOP ${PID}

HappyRay:
LGTM. I would suggest a shorter overall wait time, though. Maybe reduce the wait time between retries?

azkaban-web-server/src/main/bash/azkaban-web-shutdown.sh
chengren311: I will reduce maxattempt to 3, so that it can fail faster. But I don't think we should spend too much effort tweaking these numbers as long as they are reasonable.

chengren311:
I will reduce the retry number from 5 to 3.

HappyRay:
LGTM

PR #933, opened 2017-03-07
fix:932 Implement AzkabanDataSource's getParentLogger

darionyaphet:
`MySQLBasicDataSource` and `EmbeddedH2BasicDataSource` extend AzkabanDataSource class . 

It's need to implement getParentLogger()

### DIFF
--- a/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
+++ b/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
@@ -26,6 +26,7 @@ import org.apache.log4j.Logger;
 
 import java.sql.PreparedStatement;
 import java.sql.Connection;
+import java.sql.SQLFeatureNotSupportedException;
 
 import azkaban.utils.Props;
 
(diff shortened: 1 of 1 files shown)


kunkun-tang:
One thing I don't understand is that why compilation does not fail.

kunkun-tang:
I checked the recent dbcp common library 2.1.1 They already implemented a default getParentLogger method inside BasicDataSource, so that we don't need to take care of it our selves. We probably need to upgrade jdbc libraries at some point in future.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -229,6 +235,11 @@ public boolean allowsOnDuplicateKey() {
     public String getDBType() {
       return "h2";
     }
+
+    @Override
+    public java.util.logging.Logger getParentLogger() throws SQLFeatureNotSupportedException {
```
kunkun-tang: This method could be implemented inside azkabanDataSource, rather than implementing twice here.

PR #935, opened 2017-03-10
adding comments on Jetty Performance metrics

kunkun-tang:
Details what Jetty metrics is doing


azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -238,8 +238,15 @@ public AzkabanWebServer(Server server, Props props) throws Exception {
   private void startWebMetrics() throws Exception {
 
     MetricRegistry registry = MetricsManager.INSTANCE.getRegistry();
+
+    // the number of idle threads in Jetty thread pool
```
chengren311: drop "the"

kunkun-tang: Why do you not like the article here? I believe it is necessary.

kunkun-tang:
@chengren311 Added "The" in the following two comments.

PR #937, opened 2017-03-13
Retry the jobs on executor instead of killing them right away when executor host is running out of memory

chengren311:
This is a premature PR, just want to gather some early feedbacks.

Currently a job will be killed right away before actual execution if there's not enough memory. But as memory usage fluctuates, it's possible there will be enough memory later. Therefore we should retry the job if memory requirement is not met.


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -692,13 +693,12 @@ private Status runJob() {
...
       job.run();
     } catch (Throwable e) {
-
       if (props.getBoolean("job.succeed.on.failure", false)) {
         finalStatus = changeStatus(Status.FAILED_SUCCEEDED);
         logError("Job run failed, but will treat it like success.");
         logError(e.getMessage() + " cause: " + e.getCause(), e);
       } else {
-        finalStatus = changeStatus(Status.FAILED);
+        finalStatus = (e instanceof OutOfMemoryException) ? changeToFailureStatusWithCause(ExecutableNode.FailureCause.OOM) : changeStatus(Status.FAILED);
```
HappyRay: Would it be simpler to detect low memory situation in azkaban.execapp.JobRunner#run and retry after sleeping for some time?

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -55,9 +55,9 @@ public static void init(int memCheckInterval) {
    * Given Xms/Xmx values (in kb) used by java process, determine if system can
    * satisfy the memory request
    */
-  public synchronized static boolean canSystemGrantMemory(long xms, long xmx, long freeMemDecrAmt) {
+  public synchronized static boolean canSystemGrantMemory(long xms, long xmx, long freeMemDecrAmt) throws Exception {
     if (!memCheckEnabled) {
-      return true;
+      throw new Exception("memory check is disabled");
```
chengren311: this method returning true should only indicate memory is granted on request. Since we rely on the return value of this method to construct the retry logic.  If if memory check is not enabled,  it should throws an exception.

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
```diff
@@ -26,5 +26,12 @@
   public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
-  
+
+
+  // Memory check retry interval when OOM in ms
+  public static final long MEMORY_CHECK_INTERVAL = 1000*60*10;
```
chengren311: i think 1 min is too short, maybe 5 mins. Memory issue should't go away so soon.

HappyRay:
Could you add monitoring in a follow up change?

Please include in the change description how you manually tested the change. 

I understand automated testing in this code is difficult. Given the importance of this change, I am ok with checking it in now and get some benefit out of it faster. It would be great if you can add automated tests in the future. 

LGTM after you addressed the earlier comments.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
HappyRay: If the check is turned off, it seems reasonable to log that the memory is always granted without an exception.

In any case, I would suggest keeping non essential changes and refactoring changes separate.

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
ameyamk: + 1 for 1 minute interval - starving systems for too long is generally not a good idea - besides checking memory usage should be inexpensive operation too

chengren311: Discussed with @HappyRay  offline. We agreed that let's first start with short interval to avoid job long starving. We can adjust the parameter if we observe any outstanding issue.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
chengren311: updated. Might keep it as a separate change.

chengren311:
@HappyRay will add my test procedure in change description when merging the commits.

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
```diff
@@ -26,5 +26,12 @@
   public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
-  
+
+
+  // Memory check retry interval when OOM in ms
+  public static final long MEMORY_CHECK_INTERVAL = 1000*60*1;
```
shuzhang1989: add the unit into the variable name: MEMORY_CHECK_INTERVAL_MS

chengren311: good catch

PR #939, opened 2017-03-14
Creating a Project Spec File. Phase I

suvodeep-pyne:
Created a skeleton of the project spec YAML file which will can be used to fetch external resources. This adds a `ProjectSpec` POJO (Bean) and a corresponding loader. These classes will be used in subsequent PRs.

This is an incremental and idempotent change which does not affect the current behavior of Azkaban.


azkaban-common/src/main/java/azkaban/project/ProjectSpec.java
```diff
@@ -0,0 +1,50 @@
+package azkaban.project;
```
HappyRay: Please add a copyright notice.

azkaban-common/src/test/java/azkaban/project/ProjectSpecLoader.java
```diff
@@ -0,0 +1,14 @@
+package azkaban.project;
```
HappyRay: Please add a copyright notice to all new files.
Is there a tool to check and auto update copyright notices?

suvodeep-pyne: There should be a gradle plugin to do this. I have my IDE set up now.

PR #940, opened 2017-03-14
Creating the azkaban-spi sub-module and Storage Interface

suvodeep-pyne:
Creating the azkaban-spi package which is intended to be the contract against different Azkaban modules.
Added a Storage interface and related exception class hierarchy.

### DIFF
--- a/azkaban-spi/src/main/java/com/linkedin/azkaban/spi/AzkabanException.java
+++ b/azkaban-spi/src/main/java/com/linkedin/azkaban/spi/AzkabanException.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package com.linkedin.azkaban.spi;
+
+public class AzkabanException extends RuntimeException {
+  public AzkabanException(String message) {
+    this(message, null);
+  }
+
+  public AzkabanException(Throwable throwable) {
+    this(null, throwable);
+  }
+
+  public AzkabanException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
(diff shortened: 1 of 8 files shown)


azkaban-spi/src/main/java/com/linkedin/azkaban/spi/AzkabanException.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package com.linkedin.azkaban.spi;
+
+public class AzkabanException extends RuntimeException {
```
chengren311: why this has to be runtime exception?

azkaban-spi/src/main/java/com/linkedin/azkaban/spi/Storage.java
```diff
@@ -0,0 +1,59 @@
...
+  InputStream get(URI key);
+
+  /**
+   * Put an object into Storage against a key. If the key already exists, then it throws
+   * {@link KeyAlreadyExistsException;}.
+   *
+   * @param key The key is a URI pointing to the blob in Storage.
+   * @param is The input stream from which the value is read to the store. The value is read completely
+   */
+  void put(URI key, InputStream is) throws KeyAlreadyExistsException;
```
chengren311: any particular reason we throw an exception when there's duplicate key? why not just return false? and if we check duplicate key, why don't we check for delete?

suvodeep-pyne: @chengren311 

Thanks for the review. Both work. 

My thought process of having an exception is to force the user (of this interface) to check for the key upfront before using it. A return value can be easily ignored and the user won't notice that a storage request failed. This may create problems in our case where user is unware that an artifact upload failed. By having a `RuntimeException` this is bubbled and the user is forced to handle this.

This is different from a Java Map API where it is Ok to just replace the value and return the current one.

Regarding get and delete, I updated the API to throw exceptions as to keep it consistent.

azkaban-spi/src/main/java/com/linkedin/azkaban/spi/AzkabanException.java
suvodeep-pyne: This is kind of a debate and I don't have a strong opinion on this.

IMO Checked exceptions add more overhead in the source code than handle the actual problem. Normally people just bubble it up by wrapping it into a `RuntimeException`. I wanted to keep it simple. Also, this allows to not handle the exception if they are not capable of handling it.

suvodeep-pyne:
Thanks @chengren311 !

PR #941, opened 2017-03-15
Shutdown the process with kill -9 at the end in shutdown scripts

chengren311:
If all attempts to killing the web server/executor fail to kill the process, we will do a force kill (kill -9).

### DIFF
--- a/azkaban-common/src/main/bash/util.sh
+++ b/azkaban-common/src/main/bash/util.sh
@@ -34,6 +34,14 @@ function kill_process_with_retry {
      fi
     done
 
-   echo "Error: unable to kill process for $maxattempt attempt(s), shutdown failed"
-   return 1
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/bash/util.sh
```diff
@@ -34,6 +34,14 @@ function kill_process_with_retry {
...
+   kill -9 $pid
+   sleep 5
+   if [[ -n "$(ps -p $pid -o pid=)" ]]; then
+      echo "$pname is not dead ever after kill -9 [pid: $pid]"
+      return 1
+   else
+    echo "shutdown succeeded"
+    return 0
+   fi
 }
```
HappyRay: Could you add a new line?

devangorder: 'ever' should be 'even' too

chengren311: @devangorder good catch! thanks

PR #944, opened 2017-03-18
Insert azkaban.job.id before building job type

jamiesjc:
We've seen issues where 'azkaban.job.id' cannot be resolved when building job type:
"Failed to build job executor for job TestJobId_CommandCould not find variable substitution for variable(s) [command->azkaban.job.id]"
 'azkaban.job.id' is common to all job types. It should be inserted into jobProps before building job types.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -58,10 +58,6 @@ public class ProcessJob extends AbstractProcessJob {
   public ProcessJob(final String jobId, final Props sysProps,
       final Props jobProps, final Logger log) {
     super(jobId, sysProps, jobProps, log);
-
-    // this is in line with what other job types (hadoopJava, spark, pig, hive)
-    // is doing
-    jobProps.put(CommonJobProperties.JOB_ID, jobId);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -58,10 +58,6 @@
   public ProcessJob(final String jobId, final Props sysProps,
       final Props jobProps, final Logger log) {
     super(jobId, sysProps, jobProps, log);
-
-    // this is in line with what other job types (hadoopJava, spark, pig, hive)
-    // is doing
-    jobProps.put(CommonJobProperties.JOB_ID, jobId);
```
chengren311: quick question, isn't ProcessJob used by all job types? Then why is JOB_ID not found?

jamiesjc: In order to use ${azkaban.job.id} in job file or common.properties file, we need to resolve this variable when building job types. This happens earlier than configuring the JOB_ID in the jobProps. So at the time when it's building job types, it cannot find JOB_ID.

kunkun-tang:
Is this exception seen in all job types, including shell, javaProcess, hadoopJava?

jamiesjc:
Yes, I've tried adding ${azkaban.job.id} in both shell and javaProcess. The exception is seen on both. It should happen to all job types.

kunkun-tang:
@jamiesjc Thanks, I see. I'm still wondering how **shell/command** jobs know ${azkaban.job.id}. Is this stored in environment variable?

jamiesjc:
We need to explicitly put it in jobProps so that it can be resolved:
  props.put(CommonJobProperties.JOB_ID, this.jobId);

PR #947, opened 2017-03-21
Clean up executor's db entry and exit the application when executor's shutdown API is called.

chengren311:
When shutdown api is called, executor will deactivate itself, wait for all the job to finish, then call system.exit(), which triggers the shutdown hook. In shutdown hook, executor clears the database entry and forces a shutdown of all active threads.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
@@ -133,15 +133,15 @@ public interface ExecutorLoader {
 
   /**
    * <pre>
-   +   * Remove the executor from executors table.
-   +   * Note:-
-   +   * 1. throws an Exception in case of a SQL issue
-   +   * 2. throws an Exception if there is no executor in the table* </pre>
-   +   *
-   +   * @param host
-   +   * @param port
-   +   * @throws ExecutorManagerException
-   +   */
+   * Remove the executor from executors table.
+   * Note:-
+   * 1. throws an Exception in case of a SQL issue
+   * 2. throws an Exception if there is no executor in the table* </pre>
+   * </pre>
+   * @param host
+   * @param port
+   * @throws ExecutorManagerException
+   */
   void removeExecutor(String host, int port) throws ExecutorManagerException;
 
   /**
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
@@ -133,6 +133,19 @@ public interface ExecutorLoader {
 
   /**
    * <pre>
+   +   * Remove the executor from executors table.
+   +   * Note:-
+   +   * 1. throws an Exception in case of a SQL issue
+   +   * 2. throws an Exception if there is no executor in the table* </pre>
+   +   *
+   +   * @param host
+   +   * @param port
+   +   * @throws ExecutorManagerException
+   +   */
+  void removeExecutor(String host, int port) throws ExecutorManagerException;
+
+  /**
+   * <pre>
    * Log an event in executor_event audit table Note:- throws an Exception in
    * case of a SQL issue
    * Note: throws an Exception in case of a SQL issue
--- a/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
@@ -739,6 +739,21 @@ public class JdbcExecutorLoaderTest {
     Assert.assertFalse(fetchedExecutor.isActive());
   }
 
+  /* Test Removing Executor */
+  @Test
+  public void testRemovingExecutor() throws Exception {
+    if (!isTestSetup()) {
+      return;
+    }
+
+    ExecutorLoader loader = createLoader();
+    Executor executor = loader.addExecutor("localhost1", 12345);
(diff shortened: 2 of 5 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -611,12 +620,8 @@ public void shutdown() {
    */
   private void shutdownInternal() {
```
kunkun-tang: I'm wondering if this method could be leveraged/called when some severe problem occurs in the executor.

chengren311: then we can just kill the process, right?

kunkun-tang: I mean, the executor could kill itself.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -133,6 +133,19 @@ Executor addExecutor(String host, int port)
 
   /**
    * <pre>
+   +   * Remove the executor from executors table.
+   +   * Note:-
+   +   * 1. throws an Exception in case of a SQL issue
+   +   * 2. throws an Exception if there is no executor in the table* </pre>
+   +   *
```
chengren311: fixed

kunkun-tang:
LGTM.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -739,6 +739,21 @@ public void testExecutorInactivation() throws Exception {
...
+  public void testRemovingExecutor() throws Exception {
+    if (!isTestSetup()) {
+      return;
+    }
+
+    ExecutorLoader loader = createLoader();
+    Executor executor = loader.addExecutor("localhost1", 12345);
+    loader.removeExecutor("localhost1", 12345);
+
+    Executor fetchedExecutor = loader.fetchExecutor("localhost1", 12345);
```
suvodeep-pyne: May be its safer to assert that it exists before and then assert that is doesn't after delete.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -303,6 +303,15 @@ public Executor addExecutor(String host, int port)
   }
 
   @Override
+  public void removeExecutor(String host, int port) throws ExecutorManagerException {
```
suvodeep-pyne: nit:
I prefer using mockito rather than using a complete mocked class. You can spy on methods and validate calls etc. But this is up to you.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -379,7 +379,16 @@ public void run() {
...
+        String host = app.getHost();
+        int port = app.getPort();
+        try {
+          logger.info(String.format("Removing executor(host: %s, port: %s) entry from database...", host, port));
+          app.getExecutorLoader().removeExecutor(host, port);
+        } catch (ExecutorManagerException ex) {
+          logger.error(String.format("Exception when removing executor(host: %s, port: %s)", host, port), ex);
+        }
+
+        logger.info("Shutting down executor...");
```
suvodeep-pyne: Can we escalate this to `warn`. It is a major event.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -611,12 +620,8 @@ public void shutdown() {
...
   private void shutdownInternal() {
     getFlowRunnerManager().shutdown();
-    try {
-      shutdownNow();
-      logger.warn("Shutdown AzkabanExecutorServer complete");
-    } catch (Exception e) {
-      logger.error(e);
-    }
+    // trigger shutdown hook
+    System.exit(0);
```
suvodeep-pyne: Why is the shutdownNow call being skipped here?
How is the web server being killed ?

chengren311: it will trigger shutdown hook,  the executor will be shutdown there

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
chengren311: updated

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
chengren311: updated

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
chengren311: thanks, will take a look

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: Aah I see. Thanks for helping me understand this.

PR #949, opened 2017-03-21
Fetching External Dependencies

suvodeep-pyne:
Added a Dependency Manager class which would help in fetching dependencies for a project. The Manager class in turn delegates to a Dependency Fetcher class which downloads the dependency.

This is the first implementation which only downloads URLs. The downloaded URL is saved to a temporary directory using a SHA1 hash with metadata which is a properties file.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
+++ b/azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package azkaban.project;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Throwables;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Properties;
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
```diff
@@ -0,0 +1,216 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Properties;
+import javax.xml.bind.annotation.adapters.HexBinaryAdapter;
+import org.apache.log4j.Logger;
+
+import static com.google.common.base.Preconditions.*;
```
chengren311: please avoid wildcard importing

azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
```diff
@@ -0,0 +1,216 @@
...
+    // Download the resource
+    fetchUri(uri, tmpFilePath);
+
+    // key is the assumed to be unique and will be used to represent this data in Storage API
+    String key = requireNonNull(computeHash(tmpFilePath));
+
+    // Organize content
+    Path target = Paths.get(tmpDir().toString(), key);
+    if (Files.exists(target)) {
+      log.debug(String.format("Already Cached. Skipping [URI: %s, KEY: %s]", uri, key));
```
chengren311: why debug? not info?

azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  private void writeMetaData(Properties metadata, String key) {
+    try {
+      Path targetPropertiesPath = Paths.get(tmpDir().toString(), key + ".properties");
+      OutputStream out = new FileOutputStream(targetPropertiesPath.toFile());
+      metadata.store(out, "Metadata for key: " + key);
+    }
+    catch (Exception e ) {
+      e.printStackTrace();
```
chengren311: log instead of print

suvodeep-pyne: good catch! I'm actually gonna refactor all these try blocks. Handling this in each individual method doesn't make sense.

azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
suvodeep-pyne: This is gonna be the more common case. I don't want to pollute the logs for this.

azkaban-common/src/main/java/azkaban/project/DependencyFetcher.java
suvodeep-pyne: I agree wildcarding is not good. But it is not always bad.

In this case, this is kinda done intelligently by IntelliJ. If you see these classes, they are designed for static import. Example: Preconditions, Objects, Assert (in Junit).

It is a standard practice to wildcard import classes of these kind so that you can use the utilities easily and not have to write `MyAwesomeLibraryClassName.myUtilMethod()` every time. This in turn avoids multiple static import statements which kinda pollute your import paths.

suvodeep-pyne:
@chengren311 updated.

suvodeep-pyne:
holding this off for now.

PR #951, opened 2017-03-23
Flow Preparer Refactor

suvodeep-pyne:
Changes:
 - Extracted setupFlow method to FlowPreparer class
 - Simplified ProjectVersion class
 - All setup methods consolidated in FlowPreparer
 - Project Download and setup is also handled by FlowPreparer


azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,185 @@
...
+    this.projectsDir = projectsDir;
+    this.installedProjects = installedProjects;
+  }
+
+  /**
+   * Prepare the flow directory for execution.
+   *
+   * @param flow Executable Flow instance.
+   */
+  void setup(ExecutableFlow flow) {
```
chengren311: public?

suvodeep-pyne: It currently doesn't require any public access since it is used just by the executor to prepare a working directory.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,185 @@
...
+    // to set up.
+    final ProjectVersion projectVersion;
+    synchronized (installedProjects) {
+      projectVersion = installedProjects.computeIfAbsent(new Pair<>(flow.getProjectId(), flow.getVersion()),
+          k -> new ProjectVersion(flow.getProjectId(), flow.getVersion()));
+    }
+    return projectVersion;
+  }
+
+  private void cleanup(File execDir) {
```
chengren311: the cleanup method is basically equivalent of FileUtils.deleteDirectory(execDir), maybe we can remove it?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -427,6 +431,16 @@ private void cleanOlderProjects() {
     }
   }
 
+  public void deleteDirectory(ProjectVersion pv) throws IOException {
```
suvodeep-pyne: `ProjectVersion` is currently some data and some utils around projects. I am cleaning it up to be just a data class. I plan to rename it to `ProjectInfo` later.

Deletion and management of data is already done in `FlowRunnerManager`. Also, if you notice, the deleteDirectory method in turn calls a utility method. I agree it can be static. Will change that.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,185 @@
...
+
+      log.warn(String.format("Project Preparation complete. [%s]", pv));
+    } finally {
+
+      if (projectFileHandler != null) {
+        projectFileHandler.deleteLocalFile();
+      }
+
+      // Clean up: Remove tempDir if exists
+      if (tempDir.exists()) {
```
chengren311: just FileUtils.deleteDirectory(tempDir) will do, deleteDirectory already takes care of file non existing case.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: yes. it is simply a convenience method. Changed the code a bit.

Basically, I didn't want to handle IOException in a catch/finally block. Just made a method to make it look a little cleaner.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: fixed.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+    return projectVersion;
+  }
+
+  private void cleanup(File execDir) {
+    if (execDir != null) {
+      try {
+        FileUtils.deleteDirectory(execDir);
+      } catch (IOException e) {
+        Throwables.propagate(e);
```
HappyRay: Isn't this an anti-pattern described by https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

Shall we move away from using a deprecated method in general?

suvodeep-pyne: Interesting read. I kinda agree with the argument. Will refrain from using this. Updated PR.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -427,6 +431,16 @@ private void cleanOlderProjects() {
...
+    synchronized (pv) {
+      logger.warn("Deleting project: " + pv);
```
HappyRay: Why is it a warn or info?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -413,7 +417,7 @@ private void cleanOlderProjects() {
             try {
               logger.info("Removing old unused installed project "
                   + version.getProjectId() + ":" + version.getVersion());
-              version.deleteDirectory();
+              deleteDirectory(version);
               installedProjects.remove(new Pair<Integer, Integer>(version
                   .getProjectId(), version.getVersion()));
             } catch (IOException e) {
```
HappyRay: Should the exception be logged?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -112,6 +112,7 @@
   private final ExecutorLoader executorLoader;
   private final ProjectLoader projectLoader;
   private final JobTypeManager jobtypeManager;
+  private final FlowPreparer flowPreparer;
```
HappyRay: Does it need to be a member variable? It is currently only used in one method.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -530,46 +544,6 @@ private void configureFlowLevelMetrics(FlowRunner flowRunner) {
 
   }
 
-  private void setupFlow(ExecutableFlow flow) throws ExecutorManagerException {
```
HappyRay: Nice. This class is too big.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectVersion.java
```diff
@@ -16,34 +16,29 @@
...
+
   private File installedDir;
 
   public ProjectVersion(int projectId, int version) {
+    checkArgument(projectId > 0);
+    checkArgument(version > 0);
+
     this.projectId = projectId;
     this.version = version;
   }
```
HappyRay: Does it makes sense to  compute the installDir when the ProjectVersion is instantiated? This way we can have one constructor and make the installdDir final.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
```diff
@@ -0,0 +1,93 @@
...
+    when(executableFlow.getExecutionId()).thenReturn(12345);
+    when(executableFlow.getProjectId()).thenReturn(12);
+    when(executableFlow.getVersion()).thenReturn(34);
+
+    instance.setup(executableFlow);
+    File execDir = new File(executionsDir, "12345");
+    assertTrue(execDir.exists());
+    assertTrue(new File(execDir, SAMPLE_FLOW_01).exists());
+  }
+}
```
HappyRay: Add a trailing newline?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package azkaban.execapp;
```
HappyRay: Nice. Thanks for adding tests!

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @After
+  public void tearDown() throws Exception {
+    FileUtils.deleteDirectory(executionsDir);
+    FileUtils.deleteDirectory(projectsDir);
+  }
+
+  @Test
+  public void testSetupProject() throws Exception {
+    ProjectVersion pv = new ProjectVersion(12, 34, new File(projectsDir, "sample_project_01"));
```
HappyRay: In the production code, it appears that the project directory was not passed to the constructor. Shall the test match the usage in the production code?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,180 @@
...
+      checkState("zip".equals(projectFileHandler.getFileType()));
+
+      log.info("Downloading zip file.");
+      final File zipFile = requireNonNull(projectFileHandler.getLocalFile());
+      final ZipFile zip = new ZipFile(zipFile);
+      Utils.unzip(zip, tempDir);
+
+      Files.move(tempDir.toPath(), pv.getInstalledDir().toPath(), StandardCopyOption.ATOMIC_MOVE);
+
+      log.warn(String.format("Project Preparation complete. [%s]", pv));
```
HappyRay: Why log.warn not log.info?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Map;
+import java.util.zip.ZipFile;
+import org.apache.commons.io.FileUtils;
+import org.apache.log4j.Logger;
+
+import static com.google.common.base.Preconditions.*;
+import static java.util.Objects.*;
+
+
+public class FlowPreparer {
```
HappyRay: Please add a javadoc.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: It doesn't. I'm thinking of this class as a service. This gets instantiated with parameters that don't need to change per project. So `FlowRunnerManager` should depend on components like `FlowPreparer` and `FlowExecutor`. And they should work with some sort of `ExecutorConfig`. That's my thought process.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Oh I haven't touched this part yet. I'm attacking this class in parts.

Let me fix this anyway.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: `warn` is a stronger event that info. Deleting a project directory is an important event. Hence the categorization.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Thank you! Not only that. 
I thought of bringing the code for preparing flows at one place. So that I can update this easily.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectVersion.java
suvodeep-pyne: I have thought about this. But I want to push this for later. This class is heavily used and I didn't want to alter the flow in the first change.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
suvodeep-pyne: It is just a plain vanilla `set` method. I don't think this would change the behavior.

The main goal of the test is to test whether flow and project directories are setting up as expected.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: same reasoning.

`warn` is a stronger event that info. Project preparation happens once for every version for every executor. Just wanted to capture that with a higher log priority.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: I'm not sure what to add in the javadoc specifically.

If I add `FlowPreparer` prepares the flow or something: this is a perhaps redundant. I tried to name the methods and classes to be as self explanatory as possible so that documentation required is minimal. Also, I assume that these pieces of code will go through a lot of churn till they are stable. Documentation might be an overhead to maintain whilst this change. Should we add this later if at all? I don't see this to be an external facing API anyway.

HappyRay: Got it. That's fine.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
HappyRay: Got it. Warning seems to me like a situation that may not be normal or dangerous.  In this case, this is routine and normal. I don't have a strong opinion either way. We can adjust after we gain more experience working with the logs.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: I am interested in your thoughts on the future plan. We can discuss in person.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: Multiple refactoring is fine too.

Thanks.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: deleteDirectory seems to have side effects. Would making it static makes mocking it later more difficult?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: This is a destructive operation, warn seems to make sense to me.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: Agreed.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectVersion.java
HappyRay: Sounds good.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
HappyRay: Ok.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: I see what you mean. The problem is that the order is `info`, `warn` `error`! So, I really didn't have much of a choice :) . But I also wanted to highlight that project preparation complete was an important event.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: @HappyRay 
yes. didn't think about this. I have it `static` for now as per @chengren311 's suggestion. But we can always revert it back if the need arises.

PR #954, opened 2017-03-24
Improve error message of scheduleCronFlow API

wyukawa:
It is difficult to understand the error message of scheduleCronFlow API.

example

```
Flow test_flow cannot be found in project azkaban.project.Project@13846b77
```

In this case, I don't know the Azkaban project name.

I expect the following

```
Flow test_flow cannot be found in project test
```

This pull request is similar to https://github.com/azkaban/azkaban/pull/566


wyukawa:
Thanks!

PR #955, opened 2017-03-24
If a flow runs for longer than 10 days (configurable), kill it.

chengren311:
A flow running unnecessarily long usually means there's something wrong with the flow. We should kill it if it's longer than X days(10 days by default, but configurable in azkaban properties) to prevent it from keeping occupying system resource.


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -314,6 +321,20 @@ public void run() {
               lastExecutionDirCleanTime = currentTime;
             }
 
+            if (flowMaxRunningTimeInMins > 0 && currentTime - LONG_RUNNING_FLOW_KILLING_INTERVAL_MS > lastLongRunningFlowCleanTime) {
+              long now = System.currentTimeMillis();
+              logger.info(String.format("Killing long jobs running longer than %s mins", flowMaxRunningTimeInMins));
+              for(FlowRunner flowRunner : runningFlows.values()) {
```
shuzhang1989: space between for and (

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -314,6 +321,20 @@ public void run() {
...
+                logger.info("flowRunner: "+flowRunner.getExecutableFlow().getId()+": status: "+flowRunner.getExecutableFlow().getStatus());
```
shuzhang1989: space between +

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -271,10 +273,15 @@ public void setGlobalProps(Props globalProps) {
...
 
+    // Every 5 mins kill flows running longer than allowed max running time
+    private static final long LONG_RUNNING_FLOW_KILLING_INTERVAL_MS = 5 * 60 * 1000;
+
     private boolean shutdown = false;
     private long lastExecutionDirCleanTime = -1;
     private long lastOldProjectCleanTime = -1;
     private long lastRecentlyFinishedCleanTime = -1;
+    private long lastLongRunningFlowCleanTime = -1;
+    private final long flowMaxRunningTimeInMins = azkabanProps.getInt(ServerProperties.AZKABAN_MAX_FLOW_RUNNING_MINS, 60*24*10);
```
shuzhang1989: space between *

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -905,19 +911,15 @@ public void resume(String user) {
...
-    }
-    interrupt();
+    logger.info("Flow killed by " + user);
+    kill();
   }
 
-  private void kill() {
+  public void kill() {
     synchronized (mainSyncObj) {
+      if(flowKilled) return;
```
shuzhang1989: should we log when we got a double kill?

shuzhang1989:
@kunkun-tang i think these are valid thoughts. But for the old tasks stuck, do we need to have a gc like process to keep health check all tasks?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated

chengren311:
@kunkun-tang 
Running flow resides in old executor only.

I manually started a bunch of flows locally and checked they were killed on time.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
chengren311: it's a minor event so i think it's ok

suvodeep-pyne: @shuzhang1989 How do you suggest doing that?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -905,19 +911,15 @@ public void resume(String user) {
...
       logger.info("Kill has been called on flow " + execId);
-
+      flow.setStatus(Status.KILLED);
```
suvodeep-pyne: `setStatus` just sets the status right? Should we set it once all the process has been killed? This seems to be more wishful where we assume that we will kill the flow in the upcoming steps.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -905,19 +911,15 @@ public void resume(String user) {
...
-      updateFlow();
-    }
-    interrupt();
+    logger.info("Flow killed by " + user);
+    kill();
   }
 
-  private void kill() {
+  public void kill() {
     synchronized (mainSyncObj) {
```
suvodeep-pyne: `mainSyncObj` is not even a final member here. So, locking on it seems unsafe and not a best practice.

suvodeep-pyne: Also, I would ideally wrap the entire thing in a try catch block and log stuff when we are unable to kill jobs.

Things to think about:
You have 10 jobs running. killing the 4th one failed or threw an exception. What do we do? 
 - What's the status of the flow after that?
 - How do we handle failures?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -271,10 +273,15 @@ public void setGlobalProps(Props globalProps) {
     // Every 2 mins clean the recently finished list
     private static final long RECENTLY_FINISHED_INTERVAL_MS = 2 * 60 * 1000;
 
+    // Every 5 mins kill flows running longer than allowed max running time
+    private static final long LONG_RUNNING_FLOW_KILLING_INTERVAL_MS = 5 * 60 * 1000;
```
suvodeep-pyne: Why some time values in mins and other in ms? Can we make this consistent. Also, I would encourage looking at Java's `Duration` class.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -287,6 +294,15 @@ public void shutdown() {
       this.interrupt();
     }
 
+    private boolean isFlowRunningLongerThan(ExecutableFlow flow, long flowMaxRunningTimeInMins) {
+      Set<Status> nonFinishingStatusAfterFlowStarts = new HashSet<>();
+      nonFinishingStatusAfterFlowStarts.add(Status.RUNNING);
```
suvodeep-pyne: nit: You can shorten this with `new Hashet<>(Arrays.asList(...))`

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -314,6 +330,17 @@ public void run() {
               lastExecutionDirCleanTime = currentTime;
             }
 
+            if (flowMaxRunningTimeInMins > 0 && currentTime - LONG_RUNNING_FLOW_KILLING_INTERVAL_MS > lastLongRunningFlowCleanTime) {
```
suvodeep-pyne: Again, I would encourage using a class like `Duration` and objects for handling this rather than using primitives.

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
```diff
@@ -55,4 +55,8 @@
 
   // The property is used for the web server to get the host name of the executor when running in SOLO mode.
   public static final String EXECUTOR_HOST = "executor.host";
+
+  // Max flow running time in mins, server will kill flows running longer than this setting.
+  // if not set or <= 0, then there's no restriction on running time.
+  public static final String AZKABAN_MAX_FLOW_RUNNING_MINS = "azkaban.server.max.flow.running.mins";
```
HappyRay: How about azkaban.server.flow.max.running.mins so that we can group flow related server configs together?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: +1

chengren311: Will submit another PR for refactoring since there're multiple code pieces like this in this class. Just keeping them consistent for now.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: The allowed max running time for a flow is usually very long, not sure if it makes sense to use ms. @HappyRay what do you think?

HappyRay: I don't have a strong opinion either way. Since the name and the comment clearly indicate the unit is in minutes, I think it is a fine choice.

suvodeep-pyne: TL DR; There are some advantages to using classes like `Duration` as opposed to using primitives like `int` and `long` 

@chengren311 @HappyRay 
I totally understand the point you are trying to make but let me share a story. :)

This was narrated to me by Prof. Stroustrup himself. It's about a NASA probe which suffered a horrible crash probably more than a decade ago and costed more than $100 M. The controller code was written in C++. On investigation they found out that there were 2 components reading the same `double` constant which had a major role to play in the crash. I don't exactly remember the details but one of them was developed by the Americans and the other by some European country. The crash happened because the retro rockets which were supposed to stabilize and slowly land the probe stopped firing when the probe was 100 ft in the air. It was not a happy landing. On digging through the logs they found out the problem. As it turns out the Americans are fond of the pound system and the Europeans prefer the metric system. There was this evil `double` value which missed all conversions; slipped through the compiler and through code reviews and was passed as is to another component which expected a `double` but in different units. Rest is history :)

Here we have Azkaban! which is open source :D However, I would still prefer using a dedicated data structure for measuring time like `Duration` as opposed to using `long` or `int`. While this convention has been there for decades in C and other languages, it hasn't always yielded the best results.

Regarding intuition, `Duration.ofMinutes(5)` is more intuitive to me than a `5*60*1000` long with a comment. It is also very safe since the numbers are abstracted away in the API itself. The joda-time library has been extremely popular in the Java community and it was officially incorporated into Java 8 itself because of its simplicity and utility. Now, since we have Java 8 already, we can directly benefit from it instead of having variable *names* determine the *units*.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: alright. That works too.

HappyRay: Another PR sounds good to me too.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: Interesting story. Thanks!

Regarding the suggestion to use Duration, yes, I agree. I saw below in another comment that Cheng agrees too.

My comment was referring to 
" // Max flow running time in mins, server will kill flows running longer than this setting.
 +  // if not set or <= 0, then there's no restriction on running time.
 +  public static final String AZKABAN_MAX_FLOW_RUNNING_MINS = "azkaban.server.max.flow.running.mins";
"
This is a admin facing config variable name. I think having the unit in the name in minutes makes sense. Usage of Duration doesn't seem to apply in this context.

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
HappyRay: Also use xxx.minutes instead of xxx.mins to make it more explicit?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
mukund-thakur: @chengren311  is this configuration same for all flows. I was actually suggesting it to be passed during submission [#960]

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
chengren311: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: Agreed.
My comment was more directed against using primitives instead of dedicated classes for temporal variables in our internal code.

Regarding configuration parameters, `Duration` can be there as well. There are methods like [`java.time.Duration#parse`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) which uses the  ISO-8601 duration format `PnDTnHnMn.nS`. Example:
```
# Set max flow running time to 15 min
azkaban.server.max.flow.running=PT15M
```
That being said, I totally agree with you in retaining existing configuration. This may be useful if/when we create new ones.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
chengren311: checked the code a bit. runner.kill() calls job.cancel() which is implemented by jobtypes and does best effort killing, even if the cancel is not successful, runner.kill() marks the job as killed anyway. Similarly, the flow is marked as killed as long as user issues the kill command on it, but it doesn't guarantee the flow is actually killed.

chengren311: discussed offline. We should handle these cases appropriately. If runner.kill() fails for some reason, we should handle it catch the failure and continue to kill the rest.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
chengren311: see discussion above.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: @mukund-thakur we have flow/job level SLA feature. Maybe that's what you are looking for. Please let me know if that doesn't solve your problems.

chengren311:
@HappyRay will add unit test later.

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
```diff
@@ -55,4 +55,8 @@
 
   // The property is used for the web server to get the host name of the executor when running in SOLO mode.
   public static final String EXECUTOR_HOST = "executor.host";
+
+  // Max flow running time in mins, server will kill flows running longer than this setting.
+  // if not set or <= 0, then there's no restriction on running time.
+  public static final String AZKABAN_MAX_FLOW_RUNNING_MINS = "azkaban.server.flow.max.running.minutes";
```
juhoautio: The comment says _"if not set or <= 0, then there's no restriction on running time"_, but when I check the condition in FlowRunnerManager, the code is this:

```java
flowMaxRunningTimeInMins = azkabanProps.getInt(ServerProperties.AZKABAN_MAX_FLOW_RUNNING_MINS, 60 * 24 * 10);
```

This means that if the property is not set, the timeout actually defaults to 10 days?!

juhoautio:
1. Can we change the default to `-1` so that this feature is disabled unless the property is set?
1. Shouldn't this feature be documented?
1. In which file is this supposed to be set?

chengren311:
@juhoautio 
1. if the property is set to a number <= 0, then this feature is disabled.
2. it's documented http://azkaban.github.io/azkaban/docs/latest/ you can search azkaban.server.flow.max.running.minutes	
3. as you can see from document, it's set in executor's property file
4. Yeah, we merged multiple constants class into one class.

juhoautio:
> The reason we have default value as 10 days is we do want to kill flows running over long being unnoticed.

Isn't it a bit aggressive to have that as a default? Consider a user who has had a flow running happily for a couple of days and expects it to keep running.. Then it comes as a surprise that azkaban actually kills your flow after 10 days.

It's not only that it gives up tracking the execution, it even calls `cancel()` (which is kind of the right thing to do if Azkaban is going to stop tracking the execution, but..), which will definitely interrupt the running jobs in that execution.

Sorry, I couldn't find the sources for http://azkaban.github.io/azkaban/docs/latest/ in github. I wonder where it is..

> it's set in executor's property file

But the property is not present in the default executor props file. Having it there might help users because one usually reads through the props to see what they wanna customize when setting up Azkaban installation. I guess it should also be in the default props for solo Azkaban.

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
juhoautio: So @chengren311, what does it mean that "if not set .. there's no restriction on running time"? Isn't that comment wrong then?

chengren311:
@juhoautio the reason we want to have a max time on flow running time is in Linkedin there're non-trivial amount of flows that abandoned by their owners but keep occupying azkaban and hadoop resources. After talking to our users, we found 10 days are a reasonable number to us since generally no flow should actually run longer that 10 days. Why would 10 days seem too short to you?

you can check https://github.com/azkaban/azkaban/pull/1027 for the source

Thanks, it's a good idea to add this property in default prop file. We will follow up with that.

azkaban-common/src/main/java/azkaban/constants/ServerProperties.java
chengren311: sorry, this is my mistake, comment needs to be fixed.

juhoautio:
> Why would 10 days seem too short to you?

It's not as much about a specific limit. I think the execution time should be unlimited by default. That's the most natural behaviour. The problem is more like, 10 days is too much :) Because it can be easily missed, if you're testing that some job runs fine, maybe deploy/restart it once per week while still developing it.. then you consider that it's working fine, and leave it running there. But then it can come as a total surprise that the job gets suddenly auto-killed in the middle of the weekend and you get unexpected downtime.

For example we have a custom Azkaban job type that launches Apache Flink streaming jobs externally on AWS EMR. These streams are supposed to be going on forever unless touched. The job type also implements `cancel()`, which terminates the external EMR cluster. If the 10 day limit would be hit, the streaming job would be terminated, and we don't want that.

For people upgrading Azkaban from older version it can be easy to miss that they should add the new property to keep the existing behaviour. I think the optimal way to go about this would be to:
- Default to -1 in Java code (if not defined)
- Add to default props files, also set to -1 there as a "default"

At LinkedIn you could define the 10 day limit in props.

This is also related to _"Launching and monitoring long-living jobs (i.e. Samza, Storm, etc)"_ mentioned on [Azkaban 4.x Roadmap](https://github.com/azkaban/azkaban/wiki/Azkaban-4.x--Roadmap#future-ideas) – for that use case it just doesn't make sense to stop after 10 days. By the way if you have some concepting document about this roadmap item, it would be good to mention `azkaban.server.flow.max.running.minutes` in there because it probably needs to be considered regardless of what the default is.

ameyamk:
I think I agree. -1/ unlimited is more sound default choice.
@juhoautio you want to make a PR?

juhoautio:
Ok, I'll create a PR to change the default. That means that the current comment on AZKABAN_MAX_FLOW_RUNNING_MINS can be left unchanged (ping @chengren311).

Then the `gh-pages` branch needs an update as well – I will make a PR for that as well.

PR #956, opened 2017-03-27
Remove state out of web servers - remove Project cache

jamiesjc:
Currently, we are using "projectsById" and "projectsByName" to keep the cache info about projects. In order to scale out web servers and remove the state, we need to query DB directly for the project info instead of referring to the cache.


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -175,7 +150,16 @@ public Props getProps() {
   }
 
   public List<Project> getProjects() {
-    return new ArrayList<Project>(projectsById.values());
+    List<Project> projects;
```
chengren311: initialize projects

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -159,8 +134,8 @@ public Props getProps() {
       logger.error("Bad regex pattern " + regexPattern);
       return array;
     }
-
-    for (Project project : projectsById.values()) {
+    List<Project> projects = getProjects();
+    for (Project project : projects) {
```
chengren311: for (Project project : getProjects()), no need to introduce another variable

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -142,7 +116,8 @@ public Props getProps() {
 
   public List<Project> getGroupProjects(User user) {
     List<Project> array = new ArrayList<Project>();
```
chengren311: can you make the name more explicit? array doesn't indicate anything.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -187,7 +171,8 @@ public Props getProps() {
       logger.error("Bad regex pattern " + regexPattern);
       return allProjects;
     }
-    for (Project project : getProjects()) {
+    List<Project> projects = getProjects();
+    for (Project project : projects) {
```
chengren311: why do we need to introduce projects here? why not just for (Project project : getProjects())

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -196,21 +181,13 @@ public Props getProps() {
...
-
-    /**
      * Checks if a project is active using project_id
      *
      * @param name
      */
     public Boolean isActiveProject(int id) {
-        return projectsById.containsKey(id);
+      Project project = getProject(id);
+      return project != null;
```
chengren311: getProject(id) != null simpler

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -222,14 +199,11 @@ public Boolean isActiveProject(int id) {
...
-            try {
-                fetchedProject = projectLoader.fetchProjectByName(name);
-            } catch (ProjectManagerException e) {
-                logger.error("Could not load project from store.", e);
-            }
+        try {
+            fetchedProject = projectLoader.fetchProjectByName(name);
+            loadAllProjectFlows(fetchedProject);
+        } catch (ProjectManagerException e) {
+            logger.error("Could not load project from store.", e);
```
chengren311: log the project name as well.

kunkun-tang:
Just double check Is this a complete RB to remove project info out of cache? I thought we probably need to modify **class project** as well

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -196,21 +181,13 @@ public Props getProps() {
...
-    public Boolean isActiveProject(String name) {
-        return projectsByName.containsKey(name);
-    }
-
-    /**
      * Checks if a project is active using project_id
      *
      * @param name
      */
     public Boolean isActiveProject(int id) {
```
kunkun-tang: Looks like this method is not used any longer? Who will call it ?Should we just remove it? Besides, the method name sounds a bit confused here. We should detect exist here, rather than active, right?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
kunkun-tang: +1

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -118,17 +95,14 @@ private void loadAllProjectFlows(Project project) {
...
-
   public Props getProps() {
     return props;
   }
 
   public List<Project> getUserProjects(User user) {
     ArrayList<Project> array = new ArrayList<Project>();
-    for (Project project : projectsById.values()) {
+    List<Project> projects = getProjects();
+    for (Project project : projects) {
```
kunkun-tang: Should be safe to combine.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -243,14 +217,11 @@ public Project getProject(String name) {
...
-            fetchedProject = projectsById.get(id);
-        } else {
-            try {
-                fetchedProject = projectLoader.fetchProjectById(id);
-            } catch (ProjectManagerException e) {
-                logger.error("Could not load project from store.", e);
-            }
+        try {
+            fetchedProject = projectLoader.fetchProjectById(id);
+            loadAllProjectFlows(fetchedProject);
```
suvodeep-pyne: What does load all project flows do? Are there cases where I need the project and not load the flows?

If they need to always be done together. they should be encapsulated in a method probably in `ProjectLoader` if appropriate.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
suvodeep-pyne: This may not be required since you either populate the project or you throw a runtime exception.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: It is called in handlePurgeProject. A project is active means the boolean active is set to true in DB for that project. I'll add some comments about it.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: loadAllProjectFlows() will set the Map<string, Flow> flows for the project. getProject() is called in so many places, not able to enumerate them all...Most of them (>90%) will need to get the flow info, probably not all.

jamiesjc:
@chengren311 Project table has below data: | id | name                               | active | modified_time | create_time   | version | last_modified_by | description                          | enc_type | settings_blob. 
Moving cache info from web servers to DB might downgrade the performance a little bit. Currently the DB query takes about 0.02 sec. Consider our current scale and load, this shouldn't affect much. And we will have master-slave DB plan in the future to overcome potential issues.

jamiesjc:
@suvodeep-pyne 
The results of these methods are not cached. I'm planning to add some metrics to count how often these methods are called.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -118,64 +95,68 @@ private void loadAllProjectFlows(Project project) {
...
-    return array;
+    return userProjects;
   }
 
   public List<Project> getProjects() {
-    return new ArrayList<Project>(projectsById.values());
+    List<Project> projects;
+    try {
+      projects = projectLoader.fetchAllActiveProjects();
+      for (Project proj : projects) {
```
HappyRay: This is a more expensive operation. 

Is this information always needed when getProjects() is called?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -196,61 +177,44 @@ public Props getProps() {
...
-            try {
-                fetchedProject = projectLoader.fetchProjectById(id);
-            } catch (ProjectManagerException e) {
-                logger.error("Could not load project from store.", e);
-            }
+        try {
+            fetchedProject = projectLoader.fetchProjectById(id);
+            loadAllProjectFlows(fetchedProject);
+        } catch (ProjectManagerException e) {
+            logger.error("Could not load project from store.", e);
```
HappyRay: Add the project id to the log?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -268,16 +232,10 @@ public Project createProject(String projectName, String description,
           "Project names must start with a letter, followed by any number of letters, digits, '-' or '_'.");
     }
 
-    if (projectsByName.containsKey(projectName)) {
```
HappyRay: If the project already exists, a DB exception will be thrown now. Right? How is that new exception handled differently from the existing one?

HappyRay:
Could you describe how you test this change? I can't find existing unit tests for this class.

azkaban-common/src/test/java/azkaban/project/MockProjectLoader.java
```diff
@@ -37,30 +39,52 @@ public MockProjectLoader(File dir) {
...
+      }
+    }
+    return activeProjects;
   }
 
   @Override
   public Project fetchProjectById(int id) throws ProjectManagerException {
-    // TODO Auto-generated method stub
-    return null;
+    System.out.println("MockProjectLoader: fetch project by id " + id);
```
ameyamk: Any reason why we use System.out instead of logger?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -193,17 +193,23 @@ private Project fetchProjectByName(Connection connection, String name)
...
-        Project project = null;
+        Project project;
         ProjectResultHandler handler = new ProjectResultHandler();
+        // select active project from db first, if not exist, select inactive one.
+        // At most one active project with the same name exists in db.
         try {
             List<Project> projects =
                 runner.query(connection,
-                    ProjectResultHandler.SELECT_PROJECT_BY_NAME, handler, name);
+                      ProjectResultHandler.SELECT_ACTIVE_PROJECT_BY_NAME, handler, name);
```
HappyRay: Looks like this is an existing query. How was it used?

Is there a need to return inactive project?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: I think it is a good practice to always initialize local variables.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -196,61 +174,44 @@ public Props getProps() {
...
-        return projectsByName.containsKey(name);
-    }
-
-    /**
      * Checks if a project is active using project_id
      *
-     * @param name
+     * @param id
      */
     public Boolean isActiveProject(int id) {
```
HappyRay: How was this method used?

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.Test;
+
+public class ProjectManagerTest {
+
+  private File workingDir;
+
+  @Before
+  public void setUp() throws Exception {
+    System.out.println("projectManagerTest: create temp dir");
+    synchronized (this) {
```
HappyRay: Why does it need to be synchronized?

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+      workingDir = new File("_AzkabanTestDir_" + System.currentTimeMillis());
```
HappyRay: Is this helpful? http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    Project project1 = createProjectUtil(manager, projectName1, projectDescription1, user);
+
+    String projectName2 = "myTestFetchAllProjects_2";
+    String projectDescription2 = "This is to test fetching all projects: project_2";
+    Project project2 = createProjectUtil(manager, projectName2, projectDescription2, user);
+
+    List<Project> fetchedProjects = manager.getProjects();
+    Assert.assertTrue("Fetched project_1", fetchedProjects.contains(project1));
+    Assert.assertTrue("Fetched project_2", fetchedProjects.contains(project2));
+    System.out.println("Fetched all projects successfully");
```
HappyRay: Is this log necessary?

HappyRay:
What's the rollback strategy if something goes wrong?

HappyRay:
Are there places where later calls can use the previous DB call results in the context of the same user request? How does it affect consistency?
e.g. one user request translates to  get all projects, doing something else, get one project.

HappyRay:
If we refactor project class and separate out the mutable states from immutable states, can we cache project information?

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Assert.assertTrue("Project Id set", project.getId() > -1);
+    Assert.assertEquals("Project name", projectName, project.getName());
+    Assert.assertEquals("Project description", description,
+        project.getDescription());
+    Assert.assertTrue("Project is active", project.isActive());
+    return project;
+  }
+
+  private void assertProjectMemberEquals(Project p1, Project p2) {
```
HappyRay: Can this method be reused? 
azkaban.project.JdbcProjectLoaderTest#assertProjectMemberEquals

HappyRay: Also have you consider overwriting the equals method of the Project class?

HappyRay:
Removing the project related cache is not enough to scale out web servers. e.g.
azkaban.project.JdbcProjectLoader#createNewProject(java.sql.Connection, java.lang.String, java.lang.String, azkaban.user.User) there will be a race condition if two users try to create projects with the same name at the same time.
Currently the name column is not marked unique, it may work if we change it to unique and let one of the insert fail.

HappyRay:
I don't see anywhere in the non test code that azkaban.project.Project#isActive is used.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
HappyRay: see azkaban.project.Project#equals
What's the source member used for? It doesn't seem to be used.
azkaban.project.Project#source

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Test
+  public void testCreateProject() throws ProjectManagerException {
+    System.out.println("projectManagerTest: testCreateProject");
+    ProjectManager manager = createProjectManager();
+    String projectName = "myTestCreateProject";
+    String projectDescription = "This is to test creating a project";
+    User user = new User("testUser");
+    createProjectUtil(manager, projectName, projectDescription, user);
```
HappyRay: It wasn't clear to me at first where the assertions are.
Consider keeping the util function just as a helper to create projects but not do assertions.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+      }
+    }
+  }
+
+  @Test
+  public void testCreateProject() throws ProjectManagerException {
+    System.out.println("projectManagerTest: testCreateProject");
+    ProjectManager manager = createProjectManager();
+    String projectName = "myTestCreateProject";
+    String projectDescription = "This is to test creating a project";
```
HappyRay: Remove some code duplications?

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    User user = new User("testUser");
+    createProjectUtil(manager, projectName, projectDescription, user);
+    System.out.println("Created project successfully");
+  }
+
+  @Test
+  public void testRemoveProject() throws ProjectManagerException {
```
HappyRay: Since the method is simple
  public synchronized Project removeProject(Project project, User deleter)
    projectLoader.removeProject(project, deleter.getUserId());
    projectLoader.postEvent(project, EventType.DELETED, deleter.getUserId(),
        null);
    return project;
  }
Would using mockito to verify the internal calls be a better test? This will cover for example the user parameter is passed correctly.

HappyRay: Do you plan to update the unit tests for JdbcProjectLoader too? It's ok to do that in a separate pull request.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+  @Test
+  public void testUploadProject() throws ProjectManagerException {
+    System.out.println("projectManagerTest: testUploadProject");
+    ProjectManager manager = createProjectManager();
+    String projectName = "myTestUploadProject";
+    String projectDescription = "This is to test uploading a project";
+    User user = new User("testUser");
+    Project project = createProjectUtil(manager, projectName, projectDescription, user);
+
+    File testDir = new File("src/test/resources/project/testjob/testjob.zip");
```
HappyRay: Doesn't this code depend on the working directory to be set in certain way?

see http://stackoverflow.com/questions/28673651/how-to-get-the-path-of-src-test-resources-directory-in-junit for an alternative. You can find existing tests that use this technique.

azkaban-common/src/test/java/azkaban/project/MockProjectLoader.java
jamiesjc: The output of System.out will be redirected to a local file. This is already implemented for all other unit tests.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -193,17 +193,23 @@ private Project fetchProjectByName(Connection connection, String name)
         throws ProjectManagerException {
         QueryRunner runner = new QueryRunner();
         // Fetch the project
-        Project project = null;
+        Project project;
```
jamiesjc: Intellij gives this warning saying variable 'project' initializer null is redundant.  It will be assigned value later in project = projects.get(0); It there is any exception, then we don't even need to return project.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
jamiesjc: SELECT_ACTIVE_PROJECT_BY_NAME is used when creating a new project.
I also doubt whether we really need to return inactive project. In the future we can probably change the logic to remove inactive projects totally and only return active one. But for now, I just keep the existing logic in case something is breaking.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Intellij will give some optimization suggestions like the initialization is redundant and should be removed.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: It is used in handlePurgeProject() which will do the actual cleanup of the projects. If the project is still active, then it will return some error saying the project should be deleted before purging.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: Different threads may be creating and deleting the same directory simultaneously, causing some inconsistency issues.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: Thanks! This is very helpful. Seems much easier to implement. I'll give it a try.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: For different tests, the project names and descriptions are different. Might not be easy to combine those together?

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
HappyRay: Alternatively, you can consider this method:
http://stackoverflow.com/questions/2593154/get-a-resource-using-getresource

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
HappyRay: Got it. Make sense. Thanks

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
HappyRay: Understood.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: Good to know that compiler/IDE will take care of this. Thanks

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: azkaban.project.JdbcProjectLoader#createNewProject(java.sql.Connection, java.lang.String, java.lang.String, azkaban.user.User) will throw the same exception.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: I'll remove this after using the junit tempdir support.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: Changed.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: Removed.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: Changed.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
jamiesjc: done.

azkaban-common/src/test/java/azkaban/project/ProjectManagerTest.java
```diff
@@ -3,160 +3,180 @@
...
+    }).when(loader).removeProject(project1, user.getUserId());
 
-  @After
-  public void tearDown() throws IOException {
-    System.out.println("Teardown temp dir");
-    synchronized (this) {
-      if (workingDir != null) {
-        FileUtils.deleteDirectory(workingDir);
-        workingDir = null;
+    doAnswer(new Answer<Void>() {
```
HappyRay: Consider setting this only in the tests that need it.

Issue #957, opened 2017-03-28
azkaban-solo startup issues


li-afaris:
The startup script for azkaban-solo fails due to the 'set -o nounset ' flag in util.sh and unset variables in the start script.

`
$ ./gradlew build installDist
... build output truncated...

BUILD SUCCESSFUL

Total time: 51.111 secs
$ cd azkaban-solo-server/build/install/azkaban-solo-server; bin/azkaban-solo-start.sh
bin/azkaban-solo-start.sh: line 18: HADOOP_HOME: unbound variable
`

Will submit a patch shortly.

ameyamk:
+1 for fixing solo server. We need this to work perfectly!

kunkun-tang:
I guess this issue could be closed, as PR was merged.

PR #958, opened 2017-03-28
Issue #957 - fixes to azkaban-solo start & shunit2 unit tests

li-afaris:
This PR will fix the azkaban-solo startup problem reported in issue #957. Additionally, I have included a few shunit2 unit tests for utils.sh.  With future PR's, we should request the contributor make additions to the unit tests.

shunit2 can easily be installed with brew on os X or your favorite package manager.

### DIFF
--- a/azkaban-common/src/main/bash/util.sh
+++ b/azkaban-common/src/main/bash/util.sh
@@ -1,53 +1,50 @@
 #!/usr/bin/env bash
 # Common utils
-set -o nounset
+set -o nounset   # exit the script if you try to use an uninitialised variable
+set -o errexit   # exit the script if any statement returns a non-true return value
 
-
-# If process is still running then abort start up
-function abort_if_process_already_running {
-  installdir="$1"
-  pidfile="${installdir}/currentpid"
-  if [ -f ${pidfile} ]; then
-    pid=`cat ${pidfile}`
-    ps -p ${pid} 2>&1 1>nul && echo "Process already running [pid: ${pid}]. Aborting." && exit 1
-  fi
+#---
+# is_process_running: Checks if a process is running
+# args:               Process ID of running proccess
+# returns:            returns 0 if process is running, 1 if not found
+#---
+function is_process_running {
+  local  pid=$1
+  kill -0 $pid > /dev/null 2>&1 #exit code ($?) is 0 if pid is running, 1 if not running
+  local  status=$?              #because we are returning exit code, can use with if & no [ bracket 
+  return $status
 }
 
-# kill the process with retry
-# return 0 if kill succeeds or no process to kill,
-#        1 if kill fails
+#---
+# kill_process_with_retry: Reads pid file and attempts to kill the running process
+# args:                    None
+# returns:                 returns 0 if kill succeds or nothing to kill, 1 if kill fails
+#---
 function kill_process_with_retry {
-   pid="$1"
-   pname="$2"
-   maxattempt="$3"
-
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/bash/util.sh
```diff
@@ -1,57 +1,55 @@
...
+# kill_process_with_retry: Reads pid file and attempts to kill the running process
+# args:                    None
+# returns:                 returns 0 if kill succeds or nothing to kill, 1 if kill fails
+#---
 function kill_process_with_retry {
-   pid="$1"
-   pname="$2"
-   maxattempt="$3"
-
-   if [[ -z $pid ]]; then
```
HappyRay: With this change, the script will fail if the pid is NULL, right?

azkaban-common/src/test/bash/test_util.shunit2
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env shunit2
+# Make sure to "brew install shunit2" or use your favorite package manager to install
+# before running tests
+
+function oneTimeSetUp() {
```
HappyRay: Did you put the source inside a function to prevent accidental execution of the code from the external file?

azkaban-common/src/test/bash/test_util.shunit2
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env shunit2
```
HappyRay: Thanks for contributing the first bash tests!
It will be great if the tests can be run automatically like java tests. 
But this is a big improvement already.

azkaban-common/src/main/bash/util.sh
li-afaris: Yes that's correct.  My thinking was the calling script would first call is_process_running before calling kill_process_with_retry, but failed to update azkaban-solo-shutdown.sh to demonstrate this behavior.   Now I'm questioning if that's appropriate behavior.

I'll update this PR by placing a guard inside the kill_process_with_retry function to prevent killing non-existent pids.

azkaban-common/src/test/bash/test_util.shunit2
li-afaris: oneTimeSetup() is a shunit2 feature that is intended to prepare a environment for testing.  If temp directories are needed for testing, oneTimeSetup is when  they get created.   It seemed an appropriate place for sourcing the utils.sh file.  

FWIW: there's also oneTimeTeardown() useful for cleanup after tests are completed.  More info can be found here: https://github.com/kward/shunit2

azkaban-solo-server/src/main/bash/azkaban-solo-start.sh
```diff
@@ -33,21 +35,26 @@ else
...
+CLASSPATH=${CLASSPATH:""}  # Needed for set -o nounset above
 echo "CLASSPATH: ${CLASSPATH}";
 
-executorport=$(cat "${conf}/azkaban.properties" | grep executor.port | cut -d = -f 2)
+executorport=$(grep executor.port "${conf}/azkaban.properties" | cut -d = -f 2)
 serverpath=$(pwd)
 
-# Set the log4j configuration file
+AZKABAN_OPTS=" -Xmx512M -server -Djava.io.tmpdir=$tmpdir -Dexecutorport=${executorport} \
+    -Dserverpath=${serverpath} -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005"
```
HappyRay: Nice!

PR #959, opened 2017-03-28
Fix the race condition where two same flows are being submitted for execution.

chengren311:
1. Old code will execute two same flows if they are being submitted at the same time even though concurrent running option is disabled. We use combination of project name, flow name and a fixed suffix to synchronize on making sure submitting flow method is mutual exclusive to same flows' different submissions.
2. Add log before a scheduled flow starts.


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -934,7 +934,9 @@ public void cancelExecutingJobs(ExecutableFlow exFlow, String userId,
   @Override
   public String submitExecutableFlow(ExecutableFlow exflow, String userId)
     throws ExecutorManagerException {
-    synchronized (exflow) {
+
+    String exFlowKey = exflow.getProjectName() + "." + exflow.getId() + ".submitFlow";
+    synchronized (exFlowKey.intern()) {
```
suvodeep-pyne: This is smart but hacky! I would probably prefer an alternate solution that would be more obvious to read/understand.

chengren311: @suvodeep-pyne can you elaborate your thoughts? Why do you think it's hacky and less obvious? An alternate solution I can think of is to have a hashmap whose key is combination of project name and flow name, value is its associated lock. But the implementation is more complicated.

HappyRay: Is this method called from the TriggerScannerThread? 
If so, will it block other flows from being launched?

Also, the concurrent options need to be taken into consideration. e.g. if user specified skip option, the second flow should be skipped.

suvodeep-pyne: @chengren311 
You pretty much summarized my concerns in your above comment. I am good with the `intern()` approach as a hack for something better later on rather than use something that's breakable/buggy. One suggestion is to add your explanation comment in the synchronization logic code. I was thinking of the HashMap lock management approach but I agree it has a significant implementation overhead.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1827,6 +1834,7 @@ private void processQueuedFlows(long activeExecutorsRefreshWindow,
         if(exflow.getUpdateTime() > lastExecutorRefreshTime) {
           // put back in the queue
           queuedFlows.enqueue(exflow, reference);
+          runningCandidate = null;
```
suvodeep-pyne: @chengren311 
quick question: How does this fix the problem?

chengren311: @suvodeep-pyne sorry I am supposed to put it in the commit message, but will do it when merging. So there're two threads: producer thread which puts flow to be executed into the waiting queue, and consumer thread which polls from the queue and dispatch for execution. There' a short gap between consumer polls an flow(runningCandidate in the code) from the waiting queue and dispatching it, during which the flow is neither in waiting queue and running queue. At that time, producer checks the same flow and finds that it's not in any of the queue, assuming it's not running, then puts it in the waiting queue. Then two same flow are going to be dispatched. The fix is making the runningCandidate as shared variable so that it can be checked by producer as well.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -462,6 +464,9 @@ private void loadQueuedFlows() throws ExecutorManagerException {
     List<Integer> executionIds = new ArrayList<Integer>();
     executionIds.addAll(getRunningFlowsHelper(projectId, flowId,
       queuedFlows.getAllEntries()));
+    if(runningCandidate != null) {
```
HappyRay: Add space between if and (.

You may want to consider letting IDE format the modified methods.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -934,7 +939,9 @@ public void cancelExecutingJobs(ExecutableFlow exFlow, String userId,
   @Override
   public String submitExecutableFlow(ExecutableFlow exflow, String userId)
     throws ExecutorManagerException {
-    synchronized (exflow) {
+
+    String exFlowKey = exflow.getProjectName() + "." + exflow.getId() + ".submitFlow";
+    synchronized (exFlowKey.intern()) {
```
HappyRay: Is this change still needed until we have a better fix in place? This will only help with the race condition between scheduled flow and non-scheduled flow executions. 
The other change would fix the bigger vulnerability.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: i think it will fix the race condition when  same flow is submitted by API and schedule at the same time.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: thanks, updated the comment

suvodeep-pyne:
Other synchronization options:
 - Convert the `QueuedExecutions` to a `Queue` instead of a `PriorityQueue`. Perform a `peek` instead of dequeuing it right away. This doesn't work in a `PriorityQueue` since the head can change
 - Manage locking in `QueuedExecutions` itself.

PR #961, opened 2017-03-28
Added logging for classpath information while loading plugins

suvodeep-pyne:
Adding a log statement to capture plugin details and the classpath which is used to create the classloader for loading plugins.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -309,6 +309,7 @@ public class JobTypeManager {
     }
 
     // each job type can have a different class loader
+    logger.info(String.format("Classpath for plugin[dir: %s, JobType: %s]: %s", pluginDir, jobTypeName, resources));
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -309,6 +309,7 @@ private ClassLoader loadJobTypeClassLoader(File pluginDir,
     }
 
     // each job type can have a different class loader
+    logger.info(String.format("Classpath for plugin[dir: %s, JobType: %s]: %s", pluginDir, jobTypeName, resources));
```
chengren311: resources is a list, can it be printed out in this way? Or we need to join together as a string?

suvodeep-pyne: Any `Collection` class has `toString()` built in to do that.

chengren311: talked with @suvodeep-pyne. Collection type supports this.

Issue #603, opened 2016-01-26
How to set SLA for a flow using ajax api


vikramkone:
I tried this and it didn't work.. What am I doing wrong? There is no documentation for this api so this is based on code inspection of schedule servlet class.

`
curl -k https://HOST:PORT/schedule -d "ajax=setSla&scheduleId=2&slaEmails=foo@bar.com&settings=,06:00,SUCCESS,true,true" -b azkaban.browser.session.id=SESSION_ID`

vikramkone:
Figured this out by looking at the code..need to do this

` curl -k https://HOST:PORT/schedule -d "ajax=setSla&scheduleId=2&slaEmails=foo@bar.com&settings[0]=,06:00,SUCCESS,true,true" -b azkaban.browser.session.id=SESSION_ID`

moscovig:
It's not working

PR #965, opened 2017-03-31
Add setSla and slaInfo API documents

wyukawa:
setSla

Honestly, setSla document is difficult to write but better than nothing.

ref: https://github.com/azkaban/azkaban/issues/603

slaInfo


HappyRay:
Thanks!

BTW, we are planning to add these configurations in the flow config files so that they can be defined as part of the flow and checked into source control as the flow definition.

wyukawa:
Thanks!

>we are planning to add these configurations in the flow config files

Sounds good!

PR #967, opened 2017-03-31
change execution.dir.retention to execution.dir.retention.second

YuanGunGun:
use second to simple config

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -127,7 +127,7 @@ public class FlowRunnerManager implements EventListener,
   private Props globalProps;
 
   private long lastCleanerThreadCheckTime = -1;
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -147,7 +147,7 @@ public FlowRunnerManager(Props props, ExecutorLoader executorLoader,
     azkabanProps = props;
 
     // JobWrappingFactory.init(props, getClass().getClassLoader());
-    executionDirRetention = props.getLong("execution.dir.retention", executionDirRetention);
+    executionDirRetention = props.getLong("execution.dir.retention.second", executionDirRetention);
```
HappyRay: This is a backward incompatible change since the property name is part of the contract.

I understand the unit makes it less convenient to config this value. Is it a big problem for you?

You are welcome to improve the code by switching to use Duration class for the internal implementation.

PR #969, opened 2017-04-07
add metrics for jobs waiting when OOM

chengren311:
monitor the number of waiting jobs when OOM


azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -71,4 +73,12 @@ public void markFlowFail() {
   public void setDBConnectionTime(long milliseconds) {
     dbConnectionTime.set(milliseconds);
   }
+
+  /**
+   * Mark the occurrence of an job waiting event due to OOM
+   */
+  public void markJobWait() {
```
kunkun-tang: This name should be specific for OOM, right?

chengren311: updated

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -87,10 +88,13 @@ public void run() throws Exception {
           break;
         }
         if (attempt < ServerInternals.MEMORY_CHECK_RETRY_LIMIT) {
```
kunkun-tang: One thing still confusing me is that There looks duplicate checking logics in Line90 and line 84.
`attempt <= ServerInternals.MEMORY_CHECK_RETRY_LIMIT`

Are they really necessary?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -87,10 +88,13 @@ public void run() throws Exception {
...
-          info(String.format(oomMsg + ", sleep for %s secs and retry, attempt %s of %s", TimeUnit.MILLISECONDS.toSeconds(ServerInternals.MEMORY_CHECK_INTERVAL), attempt, ServerInternals.MEMORY_CHECK_RETRY_LIMIT));
+          info(String.format(oomMsg + ", sleep for %s secs and retry, attempt %s of %s", TimeUnit.MILLISECONDS.toSeconds(ServerInternals.MEMORY_CHECK_INTERVAL_MS), attempt, ServerInternals.MEMORY_CHECK_RETRY_LIMIT));
+          if (attempt == 1) {
+            CommonMetrics.INSTANCE.markOOMJobWait();
```
HappyRay: Is it better to track the current number of jobs blocked waiting for memory? This will also give us the information about how long the jobs wait.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
chengren311: we just want to skip waiting for last attempt

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
```diff
@@ -29,7 +29,7 @@
 
 
   // Memory check retry interval when OOM in ms
-  public static final long MEMORY_CHECK_INTERVAL = 1000*60*1;
+  public static final long MEMORY_CHECK_INTERVAL_MS = 1000 * 60 * 1;
```
chengren311: will update the description

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -31,6 +31,7 @@
 
   private Meter dbConnectionMeter;
   private Meter flowFailMeter;
+  private Meter OOMwaitingJobMeter;
```
chengren311: This requires moving ProcessJob and all its dependent class to executor package. I think it won't be that complicated since we can just move the files but keep the package name unchanged. But it's definitely a big change which should be in another commit.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -100,13 +103,15 @@ public void run() throws Exception {
...
           if(killed) {
+            CommonMetrics.INSTANCE.unmarkOOMJobWait();
             info(String.format("Job %s was killed while waiting for memory check retry", getId()));
             return;
           }
         }
       }
 
       if (!isMemGranted) {
+        CommonMetrics.INSTANCE.unmarkOOMJobWait();
```
kunkun-tang: Should we also have a logger.error message here besides an Exception.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
chengren311: updated

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
chengren311: good idea, updated

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
shuzhang1989: +1 for the good practice that one diff only does one thing

HappyRay: Consider using Duration class instead?

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -43,6 +43,7 @@
   private void setupAllMetrics() {
     dbConnectionMeter = MetricsUtility.addMeter("DB-Connection-meter", registry);
     flowFailMeter = MetricsUtility.addMeter("flow-fail-meter", registry);
+    MetricsUtility.addGauge("OOM-waiting-job-meter", registry, OOMWaitingJobCount::get);
```
HappyRay: This is a gauge not a meter. Will OOM-waiting-job-count be a better name?

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -71,4 +72,20 @@ public void markFlowFail() {
   public void setDBConnectionTime(long milliseconds) {
     dbConnectionTime.set(milliseconds);
   }
+
+  /**
+   * Mark the occurrence of an job waiting event due to OOM
+   */
+  public void markOOMJobWait() {
```
HappyRay: Similarly, would it be better to rename it " incrementOOMJobWaitCount"?

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -71,4 +72,20 @@ public void markFlowFail() {
...
+    OOMWaitingJobCount.incrementAndGet();
+  }
+
+  /**
+   * Unmark the occurrence of an job waiting event due to OOM
+   */
+  public void unmarkOOMJobWait() {
```
HappyRay: ditto

azkaban-common/src/main/java/azkaban/constants/ServerInternals.java
chengren311: yup, but will create another PR refactoring all of them.

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
chengren311: update

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
chengren311: updated

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
chengren311: updated

HappyRay:
LGTM. I hope you can move this logic to executor as discussed earlier.

PR #971, opened 2017-04-08
Ignore a few more files

HappyRay:
Ignore the following:
.DS_Store
.bundle/
Gemfile.lock
vendor/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,14 @@
 /bin
 bin/
 build/
+.DS_Store
+.bundle/
+Gemfile.lock
+vendor/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,9 +8,14 @@
 /bin
 bin/
 build/
+.DS_Store
```
kunkun-tang: Do we have a way to ignore all DS_Store filed in children folders? I guess this file might exist there.

HappyRay: This will ignore this file in all sub-folders.

PR #972, opened 2017-04-08
Ignore node modules

HappyRay:
They are created by the unit test process.


HappyRay:
Both will work. I have seen examples of both styles. 
I like adding separate .gitignore files to sub-directories better since it limits the scope and gives better and more explicit control.

PR #978, opened 2017-04-10
Set execution log name size to 255 to match filesystem limits

hanny24:
Filesystems typically support up to 255 characters as a filename, yet azkban itself imposes 128 characters. This might be quite limiting, especially when using deeply nested subflows.


HappyRay:
The code suggests that the name represents flow name+job name. It's not a file name.
And the column value is not used.

See 

azkaban.executor.JdbcExecutorLoader#fetchLogs
azkaban.executor.JdbcExecutorLoader.FetchLogsHandler

===
Did you run into issues with this limitation?

Why not go for 512 since this is varchar?

HappyRay:
@hanny24 ping, are you still interested in working on this?

HappyRay:
Close due to inactivity.

Issue #975, opened 2017-04-09
MomentJS timezone broken


dmeijboom:
I'm getting the following Javascript error in my console when trying to schedule a workflow:

![screenshot from 2017-04-09 16 21 28](https://cloud.githubusercontent.com/assets/3341876/24838044/a074a2ac-1d40-11e7-9b86-cfb52c9ccb3b.jpeg)

I think it has something to do with the fact that momentjs renamed it's timezone package from "moment-timezone-with-data-2010-2020.min.js" to "moment-timezone-with-data-2012-2022.js" while in Gradle it's still: https://github.com/azkaban/azkaban/blob/8de25c09d177234022dcd22ae1d502d59085709b/azkaban-web-server/build.gradle#L96

I'm not sure though because I don't have any experience with Gradle. Note that as of right now I'm not able to schedule workflows at all because of this bug.

ameyamk:
We are not able to reproduce the bug - but to get you unblocked you can try and schedule flows via API

kunkun-tang:
@dmeijboom You are likely right. I just rebuild the code, found that moment-timezone-with-data-2012-2022.js replaced the original file name. I will do a PR to fix this problem. For now, please use @ameyamk 's suggestion to send schedules. Thanks.

dmeijboom:
Thanks!
For now I've fixed it by manually downloading momentjs to the old filename.

kunkun-tang:
#979 fixed.

PR #979, opened 2017-04-10
Fixing js libraries version compatibility

kunkun-tang:
Issue #975 reported a bug that the name of moment-timezone js is not right, and gradle can not fetch it to distribution package.

Tried gradle build, and figured out that the latest moment-timezone js 0.5.13 is not compatible with 0.5.5, since the js file built's name doesn't keep the same.

We used to leverage `^` to specify compatible js libraries (see details in https://docs.npmjs.com/misc/semver). In this change, we disable it to in order to secure repo.

Gradle Build works, and target-version libraries are downloaded. Also, I made a bit evaluation on google chrome, and no problem.

### DIFF
--- a/azkaban-web-server/package.json
+++ b/azkaban-web-server/package.json
@@ -3,13 +3,13 @@
   "name": "azkaban",
   "description": "Azkaban 3 [![Build Status](http://img.shields.io/travis/azkaban/azkaban.svg?style=flat)](https://travis-ci.org/azkaban/azkaban) ========",
   "dependencies": {
-    "mocha": "^3.0.2",
-    "rewire": "^2.5.2",
-    "chai": "^3.5.0",
-    "later": "^1.2.0",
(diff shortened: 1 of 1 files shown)


kunkun-tang:
Fixing #975

HappyRay:
LGTM.
Please include more details in the change description when you merge as discussed in person.

HappyRay:
Here is my understanding of this fix:

Use fixed versions of dependent js libraries.  

Issue: 
We used ^<version number> to use compatible versions of the js libraries in our build process.  See https://docs.npmjs.com/misc/semver "Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4" for details on what compatible version means. In issue #975, the build pulled in moment-timezone 0.5.13 which is supposed to be backward compatible with what we specified: "moment-timezone": "^0.5.5". However, it was not the case. Based on this experience, it seems safer to specify the fixed versions and upgrade explicitly.

PR #983, opened 2017-04-12
use full moment-timezone js and add test

kunkun-tang:
We previously used timezone-2010-to-2020 js file, which wis working. However, its name got changed when moment timezone upgraded to 0.5.13. In this change, we replace the file by the full-time-range timezone js, which will permanently stay and should not be changed in future. Also, we add some moment js tests to guarantee moment js working.

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/test/test.js
+++ b/azkaban-web-server/src/web/js/azkaban/test/test.js
@@ -7,8 +7,11 @@ function addClass(el, newClass) {
 var rewire = require('rewire'),
     dateJs = rewire('../util/date.js');
 
-var chai = require('chai');
-var assert = chai.assert;
+var moment = require('moment-timezone');
+
+var chai = require('chai'),
+    assert = chai.assert;
+
 
 //Some Test instances
 describe('addClass', function() {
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/web/js/azkaban/test/test.js
```diff
@@ -7,8 +7,11 @@ function addClass(el, newClass) {
 var rewire = require('rewire'),
     dateJs = rewire('../util/date.js');
 
-var chai = require('chai');
-var assert = chai.assert;
+var moment = require('moment-timezone');
```
HappyRay: What do you think of separating independent tests into separate files?

HappyRay:
Thanks for fixing this and adding tests!

azkaban-web-server/src/web/js/azkaban/test/test.js
kunkun-tang: Yes. We should do it when we have large extensively tests. For now, we may keep them together.

HappyRay:
Also is there a way to fail the build if a file being copied doesn't exist?

kunkun-tang:
@HappyRay Unfortunately this way can not guarantee a file being always copied. I made a quick research. It looks like running nodejs and browser is quite different. In order to detect if we have an exact js file, I guess we probably need to set up browser integration test.

HappyRay:
Is there a gradle config that will fail the gradle build process when the file is copied?

PR #984, opened 2017-04-13
Ignore the top level "temp" directory

HappyRay:
"temp" directory is ignored so that it can be used to store temporary files such as local testing configurations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ build/
 Gemfile.lock
 vendor/
 
+# temp directory is ignored so that it can be used to store temporary files such as local testing configurations.
+temp/
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -12,6 +12,9 @@ build/
 Gemfile.lock
 vendor/
 
+# temp directory is ignored so that it can be used to store temporary files such as local testing configurations.
+temp/
```
kunkun-tang: I found that I also have this folder inside AZ folder. Do you know which azkaban module generate this folder?

HappyRay: I am surprised. Not sure. What happens if you remove it and build again?

PR #985, opened 2017-04-13
implementing DB retry logics

kunkun-tang:
when DB runs unstably (or, DB switch), Azkaban today fails to respond generally. Schedule, Execution, Fetch all fail because AZ can not talk to database. In this change, we implement an DB retry logics to enforce retry fetching potential connection, when AZ could not find a stable DB connection.

Still in progress. submit PR for discussion. The main problem is the timeout setting.

Haven't done test yet.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
+++ b/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
@@ -16,15 +16,19 @@
 
 package azkaban.database;
 
+
+import azkaban.constants.ServerInternals;
+import azkaban.utils.Props;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.SQLException;
 
+import javax.sql.DataSource;
+
+import org.apache.commons.dbcp.BasicDataSource;
 import org.apache.commons.dbutils.DbUtils;
 import org.apache.log4j.Logger;
-
-import java.sql.PreparedStatement;
-import java.sql.Connection;
-
-import azkaban.utils.Props;
 
 public class DataSourceUtils {
 
@@ -142,6 +146,45 @@ public class DataSourceUtils {
       setTestOnBorrow(true);
     }
 
+    /**
+     * This method overrides {@link BasicDataSource#getConnection()}, in order to have retry logics.
+     *
+     */
+    @Override
+    public Connection getConnection() throws SQLException {
+
+      /*
+       * when az server process launches, if the connection isn't valid (e.g., network problem, wrong password, etc),
+       * we fail the connection immediately.
+       */
+      if (getDataSource() == null) {
+        return createDataSource().getConnection();
+      }
+
+      Connection connection = null;
+      int retryAttempt = 0;
+      this.dataSource = null;
+      while (connection == null && retryAttempt < ServerInternals.MAX_DB_RETRY_COUNT) {
+        try {
+          /*
+           * when DB connection could not be fetched here, dbcp library will keep searching until a timeout defined in
+           * its code hardly.
+           */
+          connection = createDataSource().getConnection();
+          if(connection != null)
+            return connection;
+        } catch (Exception ex) {
+          logger.error("Failed to find DB connection", ex);
+          this.dataSource = null;
+          logger.info("waits 10 seconds and retry. No.Attempt = " + retryAttempt);
+//          sleepMillis(1000L);
+        } finally {
+          retryAttempt ++;
+        }
+      }
+      return connection;
+    }
+
     @Override
     public boolean allowsOnDuplicateKey() {
       return true;
@@ -150,6 +193,21 @@ public class DataSourceUtils {
     @Override
     public String getDBType() {
       return "mysql";
+    }
+
+    private void sleepMillis(long numMilli) {
+      try {
+        Thread.sleep(numMilli);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+    }
+
+    /*
+     * get the parent class's dataSource
+     */
+    private DataSource getDataSource() {
+      return dataSource;
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
+++ b/azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
@@ -18,6 +18,7 @@ package azkaban.database;
 
 
 import azkaban.constants.ServerInternals;
+import azkaban.metric.MetricReportManager;
 import azkaban.utils.Props;
 
 import java.sql.Connection;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -142,6 +146,45 @@ private MySQLBasicDataSource(String host, int port, String dbName,
...
+           * its code hardly.
+           */
+          connection = createDataSource().getConnection();
+          if(connection != null)
+            return connection;
+        } catch (Exception ex) {
+          logger.error("Failed to find DB connection", ex);
+          this.dataSource = null;
+          logger.info("waits 10 seconds and retry. No.Attempt = " + retryAttempt);
+//          sleepMillis(1000L);
```
ameyamk: please remove uncommented code.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -142,6 +146,45 @@ private MySQLBasicDataSource(String host, int port, String dbName,
...
+       * we fail the connection immediately.
+       */
+      if (getDataSource() == null) {
+        return createDataSource().getConnection();
+      }
+
+      Connection connection = null;
+      int retryAttempt = 0;
+      this.dataSource = null;
+      while (connection == null && retryAttempt < ServerInternals.MAX_DB_RETRY_COUNT) {
```
ameyamk: This method works all ok - but usually its a good idea to try exponential/ fibonacci backoff in such cases

This is a good read: https://carlosbecker.com/posts/exponential-backoff-java8/
This as well: https://github.com/rholder/guava-retrying

We can keep it simple and not necessarily have to use external dependency - but this is a good example of using exponential backoff

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -151,6 +194,21 @@ public boolean allowsOnDuplicateKey() {
     public String getDBType() {
       return "mysql";
     }
+
+    private void sleepMillis(long numMilli) {
+      try {
+        Thread.sleep(numMilli);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
```
ameyamk: why not log error as well?

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
ameyamk: can we add unit test to retry logic?

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -142,6 +146,45 @@ private MySQLBasicDataSource(String host, int port, String dbName,
...
+      Connection connection = null;
+      int retryAttempt = 0;
+      this.dataSource = null;
+      while (connection == null && retryAttempt < ServerInternals.MAX_DB_RETRY_COUNT) {
+        try {
+          /*
+           * when DB connection could not be fetched here, dbcp library will keep searching until a timeout defined in
+           * its code hardly.
+           */
+          connection = createDataSource().getConnection();
```
chengren311: How does createDataSource().getConnection() work internaly? Will it establish a new connection or just fetch one from existing connection pool? If establishing a new connection, will it be an issue when db fails, every thread which are querying the DB will try to establish a new connection?

kunkun-tang: This is a very good question! We indeed have this risk if we keep the today's MysqlDataSource class structure. Let me investigate some simple fix if possible to resolve this issue.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
kunkun-tang: Good catch.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
kunkun-tang: Will considering how to add test after the code is reviewed to be valid and efficient.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
kunkun-tang: I didn't know that. They are good example of good practice!

chengren311: and we don't need such high retry limit if we use exponential backoff.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
shuzhang1989: +1 preventing thundering herd

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -142,6 +149,61 @@ private MySQLBasicDataSource(String host, int port, String dbName,
       setTestOnBorrow(true);
     }
 
+    /**
+     * Get a singleton object for MySQL BasicDataSource
+     */
+    public static MySQLBasicDataSource getInstance(String host, int port, String dbName,
+        String user, String password, int numConnections) {
+      if (instance == null) {
+        synchronized (MySQLBasicDataSource.class) {
```
HappyRay: This check and set optimization is not safe. See the effective java book.
Please use Guice instead.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
HappyRay: I am not too concerned as long as there is an upper bound on the threads from AZ hosts that will call this method.

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -142,6 +149,61 @@ private MySQLBasicDataSource(String host, int port, String dbName,
...
+          if(connection != null)
+            return connection;
+        } catch (SQLException ex) {
+          this.dataSource = null;
+          logger.error("Failed to find DB connection. waits 1 minutes and retry. No.Attempt = " + retryAttempt, ex);
+        } catch (Exception ex) {
+          logger.error("Exception other than SQLException is thrown", ex);
+          return null;
+        } finally {
+          retryAttempt ++;
```
HappyRay: Will putting this statement inside the catch (SQLException ex) { block have the same effect?

azkaban-common/src/main/java/azkaban/database/DataSourceUtils.java
```diff
@@ -151,6 +213,13 @@ public boolean allowsOnDuplicateKey() {
     public String getDBType() {
       return "mysql";
     }
+
+    /*
+     * get the parent class's dataSource
+     */
+    private DataSource getDataSource() {
```
HappyRay: What's the benefit of having this wrapper?

kunkun-tang:
#1021 reimplements this feature.

PR #989, opened 2017-04-17
Refactored Azkaban Constants

suvodeep-pyne:
Azkaban Constants was split into 4 classes in a package called Constants. The current change consolidates all classes into 1 single class with inner classes.
The motivation for this change is consolidation.
 - The classes were named as *Properties, *Internals etc which is not easy to guess and find.
 - Secondly, a separate package for constants seems like an overkill. The state of the constants in the code is quite fragmented and scattered across the codebase. Having a package for constants promotes further scattering and makes it difficult to find where to put what.
 - Although having strict boundaries for storing configuration keys, property keys may be helpful. This is the reason those smaller classes were merged into a single large class.

 If the class becomes really large and unmaintainable, we could split `ConfigurationKeys` out of the class. However, today I think it will be more easy to manage and refactor the code if we consolidate global constants such as strings and numbers in one consolidated class as opposed to multiple classes in a constants package.


suvodeep-pyne:
@kunkun-tang 
I don't suppose so because these classes were created relatively recently (Nov 2016).

HappyRay:
LGTM

PR #991, opened 2017-04-18
Storage Stubs and Implementation + Guice Service Provider

suvodeep-pyne:
This patch introduces a `ServiceProvider` interface powered by `Guice` which is also a new dependency for Azkaban. This was required to implement some of the storage implementation code.

Storage API changes
 - refactored some of the storage APIs
 - Implemented Local storage for Azkaban + unit test
 - Included Stub for Database Storage.

None of these code paths are invoked anywhere in the codebase at this time. So this should not affect any existing code. Just wanted to break up my changes into smaller chunks to make it easier to review.


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.inject.Provides;
+import com.google.inject.Scopes;
+import com.google.inject.Singleton;
+import java.io.File;
+
+import static azkaban.storage.StorageImplementationType.*;
+
+
+public class AzkabanCommonModule extends AbstractModule {
+  public static final String AZKABAN_STORAGE_TYPE = "azkaban.storage.type";
```
HappyRay: Is it going to be used as a key in the server config file?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private final Props props;
```
HappyRay: Comment on what the props will contain?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Storage Implementation
+   * This can be any of the {@link StorageImplementationType} values in which case {@link StorageFactory} will create
```
HappyRay: Nice usage of @link

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,87 @@
...
+   * path required by a classloader.
+   *
+   * examples: LOCAL, DATABASE, azkaban.storage.MyFavStorage
+   *
+   */
+  private final String storageImplementation;
+
+  public AzkabanCommonModule(Props props) {
+    this.props = props;
+    this.storageImplementation = props.getString(AZKABAN_STORAGE_TYPE, LOCAL.name());
```
HappyRay: Is it better to make DB the default so that OSS users don't have to migrate?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,85 @@
...
+  public AzkabanCommonModule(Props props) {
+    this.props = props;
+    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, LOCAL.name());
+  }
+
+  @Override
+  protected void configure() {
+    bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
+    bind(Props.class).toInstance(props);
+    bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
```
HappyRay: Is in(Scopes.SINGLETON) necessary here since the provider has the singleton annotation already.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class AzkabanCommonModule extends AbstractModule {
+  private final Props props;
+  /**
+   * Storage Implementation
+   * This can be any of the {@link StorageImplementationType} values in which case {@link StorageFactory} will create
+   * the appropriate storage instance. Or one can feed in a custom implementation class using the full qualified
+   * path required by a classloader.
+   *
+   * examples: LOCAL, DATABASE, azkaban.storage.MyFavStorage
```
HappyRay: What do you think of always asking the class name in the config? It will simplify this code. It's the method used by other projects such as Hive, hadoop.

azkaban-common/src/main/java/azkaban/storage/DatabaseStorage.java
```diff
@@ -0,0 +1,49 @@
...
+
+import azkaban.project.JdbcProjectLoader;
+import azkaban.spi.Storage;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Properties;
+import javax.inject.Inject;
+
+
+public class DatabaseStorage implements Storage {
```
HappyRay: Add a class doc?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: I went with that approach initially. I agree that would simplify the code and can get rid of the enum, etc. The downside is that a hardcoded class reference is difficult to refactor. 

Tomorrow if we decide to rename or move the class all configs need to change. This might be important today since the codebase is going through a lot of churns.
Secondly, the name mapping kind of draws a strict line between what is supported out of the box versus what is supported externally.

What do you think ?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: Actually, both safeguards are helpful in a way. This line ensures that `Guice` creates only one Storage while the `provider` ensures that we have only one `LocalStorage`. In a fictitious case where someone decides to instantiate `LocalStorage` directly, it wouldn't create a second storage. 

But, I think we can get rid of the provider `@singleton` annotation since it is not consistent with others, like `DatabaseStorage`, etc.

azkaban-common/src/main/java/azkaban/storage/DatabaseStorage.java
suvodeep-pyne: updated PR.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
HappyRay: I like the idea. It makes sense. 
Thanks

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
HappyRay: It's fine either way with me. 
From the Guice doc: 
https://github.com/google/guice/wiki/Scopes
"If there's conflicting scopes on a type and in a bind() statement, the bind() statement's scope will be used. "

"The annotation is preferred because it allows the module to be reused in different types of applications. "

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -0,0 +1,118 @@
...
+    checkArgument(metadata.containsKey(PROJECT), "Incorrect metadata. project not found");
+    checkArgument(metadata.containsKey(VERSION), "Incorrect metadata. version not found");
+    checkArgument(metadata.containsKey(EXTENSION), "Incorrect metadata. extension not found");
+
+    final String projectId = requireNonNull(metadata.getProperty(PROJECT), "project is null");
+    final String version = requireNonNull(metadata.getProperty(VERSION), "version is null");
+
+    final File projectDir = new File(baseDirectory, projectId);
+    if (projectDir.mkdir()) {
+      log.info("Created project dir: " + projectDir.getAbsolutePath());
```
HappyRay: I think it will make debugging easier if we log consistently the project id in this class. Maybe create a helper log method?

e.g. 
log consistently with: project (<project id>) in the log line.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  private File getFile(URI key) {
+    return new File(baseDirectory, key.getPath());
+  }
+
+  @Override
+  public URI put(Properties metadata, InputStream is) {
+    checkArgument(metadata.containsKey(PROJECT), "Incorrect metadata. project not found");
```
HappyRay: Would it be better to use a strong type for the metadata parameter and do the validation at the object creation time?

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -0,0 +1,118 @@
...
+
+    final File projectDir = new File(baseDirectory, projectId);
+    if (projectDir.mkdir()) {
+      log.info("Created project dir: " + projectDir.getAbsolutePath());
+    }
+
+    final File targetFile = new File(projectDir, version + "." + metadata.getProperty(EXTENSION));
+
+    if (targetFile.exists()) {
+      throw new StorageException("Error in LocalStorage. Target file already exists. targetFile: " + targetFile);
```
HappyRay: Add the project id in the exception message to make debugging easier?

HappyRay:
What do you think of creating an issue for this PR and describe what we intend to do? 
This is a relatively big change. Having an issue to tie the PRs together may be useful. 

And it also helps us to communicate with the OSS community.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -0,0 +1,118 @@
...
+    return baseDirectory.toURI().relativize(targetFile.toURI());
+  }
+
+  @Override
+  public void delete(URI key) throws KeyDoesNotExistException {
+    throw new UnsupportedOperationException("delete has not been implemented.");
+  }
+
+  private static File validateBaseDirectory(File baseDirectory) {
+    if(!baseDirectory.exists()) {
```
HappyRay: It's a bit surprising to me that validation would have this side effect. Can you think of a better strategy?

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+    checkArgument(baseDirectory.isDirectory());
+    try {
+      File testFile = new File(baseDirectory, "_tmp");
+      /*
+       * Create and delete a dummy file in order to check file permissions. Maybe
+       * there is a safer way for this check.
+       */
+      testFile.createNewFile();
+      testFile.delete();
```
HappyRay: Should there be a finally block to delete the testFile? Or use https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

HappyRay: What do you think about moving the testing of writability to a utility class and test it separately. It can be reused in other places. And the name of the method can be used as a documentation.

azkaban-common/src/main/java/azkaban/storage/StorageConfig.java
```diff
@@ -0,0 +1,44 @@
...
+package azkaban.storage;
+
+import azkaban.Constants;
+import azkaban.utils.Props;
+import com.google.inject.Inject;
+
+
+public class StorageConfig {
+
+  private final Props props;
```
HappyRay: Why is it necessary to store the props?

HappyRay: What do you think of centralizing all the server configs into one config object in the future? That may make it easier to support hot reload of configs.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -15,18 +15,22 @@
...
 
 
-/**
- * This exception is thrown when there is an attempt to create a duplicate storage key via the {@link Storage}
- * interface.
- */
-public class KeyAlreadyExistsException extends StorageException {
-  public KeyAlreadyExistsException(URI key) {
-    super("Storage key already present: " + key);
+public class StorageManager {
```
HappyRay: The name storageManager seems a bit general. It doesn't manage the entire persistent layer.
Maybe use a more specific name? As long as we have the option to change the name, we can retain the flexibility.

Consider adding a class doc?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: Oh! I already removed the `@singleton` in the provider yesterday since it was inconsistent.

azkaban-common/src/main/java/azkaban/storage/StorageConfig.java
suvodeep-pyne: We already have `Props` for that and it is not easy to manage. I personally prefer modular configs since they are much smaller and easier to deal with and also address a specific context. However, this class is expected to go through some churn.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
suvodeep-pyne: It is designed to manage storage. It is a work in progress.

I envision this to be the single point of contact for all Azkaban APIs to interact with Storage.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Provides;
+import com.google.inject.Scopes;
+import com.google.inject.Singleton;
+import java.io.File;
+
+import static azkaban.storage.StorageImplementationType.*;
+
+
+public class AzkabanCommonModule extends AbstractModule {
```
kunkun-tang: Quick question. Don't understand why this class is called AZCommonModule, but all code is storage related?

suvodeep-pyne: It is difficult to introduce Guice into an already established project since the programming model is very different in both cases. The current approach is to create a single fat module `AzkabanCommonModule` and add all bindings in it and expose this through a `ServiceProvider` API. Thus, this class is intended to be a container for all bindings.

It contains storage bindings for now as a start and both as an example on how to use Guice and benefit from it.

suvodeep-pyne: As we refactor more and more code into this pattern, we can think about breaking up the module into smaller ones. But now, feel free to put everything (in `azkaban-common`) in this module.

suvodeep-pyne:
@HappyRay 
> What do you think of creating an issue for this PR and describe what we intend to do?
This is a relatively big change. Having an issue to tie the PRs together may be useful.

I think it is a good idea to tie these PRs together. FYI #995 

> What's in sample_flow_01.zip?

It is a test zip file which contains some sample flows which I created for testing. Simulates a real use case of Azkaban.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -0,0 +1,87 @@
...
+    this.storage = storage;
+  }
+
+  /**
+   * API to a project file into Azkaban Storage
+   *
+   * @param project           project ID
+   * @param fileExtension     extension of the file
+   * @param filename          name of the file
+   * @param localFile         local file
```
HappyRay: What does local file represent?

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -0,0 +1,87 @@
...
+      String filename,
+      File localFile,
+      String uploader) {
+    final StorageMetadata metadata = new StorageMetadata(
+        String.valueOf(project.getId()),
+        String.valueOf(getLatestVersion(project)),
+        fileExtension
+    );
+    log.info(String.format(
+        "Uploading project. Uploader: %s, Metadata:%s, filename: %s[%d bytes]",
```
HappyRay: Log project name consistently for easier debugging?

azkaban-common/src/test/java/azkaban/storage/LocalStorageTest.java
```diff
@@ -0,0 +1,84 @@
...
+    File expectedTargetFile = new File(BASE_DIRECTORY, new StringBuilder()
+        .append("testProjectId")
+        .append(File.separator)
+        .append("1.zip")
+        .toString()
+    );
+    assertTrue(expectedTargetFile.exists());
+    assertTrue(FileUtils.contentEquals(testFile, expectedTargetFile));
+
+    // test get
```
HappyRay: Consider keeping each test focused on one thing.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -42,6 +42,31 @@
 public class FileIOUtils {
   private final static Logger logger = Logger.getLogger(FileIOUtils.class);
 
+  /**
+   * Check if a directory is writable
+   *
+   * @param dir directory file object
+   * @return true if it is writable. false, otherwise
+   */
+  public static boolean isDirWritable(File dir) {
```
HappyRay: Is there a test for this method?

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
suvodeep-pyne: Currently the web server caches a local copy of the uploaded file before pushing to db. This is used to performing validations and computing hash etc.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
suvodeep-pyne: The problem is `name` can have collisions. `id` is unique. And we can fix this later.

azkaban-common/src/test/java/azkaban/storage/LocalStorageTest.java
suvodeep-pyne: This tests the entire flow. I do a put and then a get.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
suvodeep-pyne: No. How do you suggest we unit test this?

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
HappyRay: Sounds good.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
suvodeep-pyne: spoke offline. This can be unit tested by creating a dir and altering permissions using OS calls and verifying that we get the required results.

PR #992, opened 2017-04-18
Update Gauva,Error Prone. Fix errors

HappyRay:
The update flagged three errors.

"installedVersions.remove(versionKey);"
The type of the value doesn't match the actual type defined in the list.
Upon close inspection, it turns out that if the entry were
correctly removed, it would cause a logical error since the
for loop access the list by an index the list should not be modified
within the same loop.
Since this is only a local variable, it is ok not updating it after old project versions are deleted.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -447,8 +447,6 @@ public class FlowRunnerManager implements EventListener,
             } catch (IOException e) {
               logger.error(e);
             }
-
-            installedVersions.remove(versionKey);
           }
         }
       }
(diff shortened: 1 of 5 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -447,8 +447,6 @@ private void cleanOlderProjects() {
             } catch (IOException e) {
               logger.error(e);
             }
-
-            installedVersions.remove(versionKey);
```
chengren311: don't we need to remove the project version from installedVersions?

HappyRay: It will cause logical errors. Please see the change description.

HappyRay:
@suvodeep-pyne Yes, we should fix that dependency.
I tested manually the basics in our test cluster.

PR #993, opened 2017-04-18
Fix azkaban-solo-start.sh

jorizci:
azkaban-solo-start.sh had a problem starting when HADOOP_HOME environment variable was not present in the system

-Fixed default set typo in HADOOP_HOME, HIVE_HOME and CLASSPATH.


li-afaris:
Looks good and thank you for the patch.   Once the 'update branch' step is completed I'll merge the change.

PR #994, opened 2017-04-18
adding test to verify dbcp connection pool auto commit reset behavior

kunkun-tang:
In AZ code, some DB operations use autocommit=true conf but others don't. In this change, we try to figure out dbcp connection pool euto commit reset behavior by adding some test. The conclusion could be made after the test:
* The default connection comes with autocommit=true attribute.
* When a connection closes, this connection will be reset to autocommit=true before returning to the connection pool.

We bring in-memory derby database to test dbcp behavior.


azkaban-common/build.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
   compile('io.dropwizard.metrics:metrics-core:3.1.0')
   compile('io.dropwizard.metrics:metrics-jvm:3.1.0')
 
+  testCompile('org.apache.derby:derby:10.13.1.1')
```
kunkun-tang: @HappyRay I'm only able to find 10.12.1.1 in our internal artifactory. Should we downgrade the version?

HappyRay: Could you check if we can use javadb instead which is part of the JDK?

azkaban-common/build.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
   compile('io.dropwizard.metrics:metrics-core:3.1.0')
   compile('io.dropwizard.metrics:metrics-jvm:3.1.0')
 
+  testCompile files(System.getenv("JAVA_HOME") + "/db/lib/derby.jar")
```
kunkun-tang: @HappyRay Checked google resourced. Even though JDK comes with this jar, Java does not rely on it when being launched. So I come up this work around. What do you think?

HappyRay: In this case, how about we stick with H2 DB?

azkaban-common/src/test/java/azkaban/database/AzkabanConnectionPoolTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+  AzkabanDataSource derbyDataSource;
+
+  @Before
+  public void setup() throws Exception{
+    derbyDataSource = new EmbeddedDerbyBasicDataSource();
+  }
+
+  @Test
+  public void testCoonectionPoolAutoCommitRest() throws Exception {
```
HappyRay: Typo in connectionPool.

also do you mean "reset"?

azkaban-common/src/test/java/azkaban/database/AzkabanConnectionPoolTest.java
```diff
@@ -0,0 +1,83 @@
...
+    Connection connection = derbyDataSource.getConnection();
+    Assert.assertEquals(connection.getAutoCommit(), true);
```
HappyRay: This works.

It would be even better if each test stays focused and test only one thing if possible so that it is easier to read and debug.

azkaban-common/src/test/java/azkaban/database/AzkabanConnectionPoolTest.java
kunkun-tang: Yes. Modified.

azkaban-common/src/test/java/azkaban/database/AzkabanConnectionPoolTest.java
kunkun-tang: Tests are separated.

PR #996, opened 2017-04-19
Trigger refactored

chengren311:
1. improve logging for existing trigger management logic
2. remove unnecessary and unused code in existing trigger management logic
3. move flow SLA trigger from web server to executor so that SLA checking is performed on executor side but keep schedule trigger as it is in web server


azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -218,6 +230,13 @@ public void setSubmitTime(long submitTime) {
 
     flowObj.put(SUBMITTIME_PARAM, submitTime);
 
+    List<Map<String, Object>> slaOptions = new ArrayList<>();
+    for (SlaOption slaOption : this.getSlaOptions()) {
```
shuzhang1989: this.getSlaOptions().stream.foreach(s -> slaOptions.add(s.toObject());

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -191,6 +199,10 @@ public void setSubmitTime(long submitTime) {
     this.submitTime = submitTime;
   }
 
+  public void setSlaOptions(List<SlaOption> slaOptions) {
+    this.slaOptions = slaOptions;
```
shuzhang1989: is builder pattern preferred here?
public ExecutableFlow setSlaOptions() {
   this.slaOptions = slaOptions;
   return this;
}

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -259,7 +278,15 @@ public void fillExecutableFromMapObject(
     if (flowObj.containsKey(PROXYUSERS_PARAM)) {
       List<String> proxyUserList = flowObj.<String> getList(PROXYUSERS_PARAM);
       this.addAllProxyUsers(proxyUserList);
+    }
 
+    if (flowObj.containsKey(SLAOPTIONS_PARAM)) {
+      List<Map<String, Object>> slaOptionsObject = flowObj.getList(SLAOPTIONS_PARAM);
+      List<SlaOption> slaOptions = new ArrayList<>();
+      for(Map<String, Object> slaOption : slaOptionsObject) {
+        slaOptions.add(SlaOption.fromObject(slaOption));
```
shuzhang1989: List<SlaOptions> slaOptions = slaOptionsObject.stream().map(s -> SlaOption.fromObject(s)).collect(Collector.toList)

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -329,13 +299,13 @@ private void onTriggerExpire(Trigger t) throws TriggerManagerException {
       List<TriggerAction> expireActions = t.getExpireActions();
       for (TriggerAction action : expireActions) {
         try {
-          logger.info("Doing expire actions");
+          logger.info("Doing expire actions for "+ action.getDescription() + " for " + t);
```
shuzhang1989: do we prefer logger.info("Doing expire actions for {} for {}", action.getDescription(), t)? it makes code more readable

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -259,7 +278,15 @@ public void fillExecutableFromMapObject(
     if (flowObj.containsKey(PROXYUSERS_PARAM)) {
       List<String> proxyUserList = flowObj.<String> getList(PROXYUSERS_PARAM);
       this.addAllProxyUsers(proxyUserList);
+    }
 
+    if (flowObj.containsKey(SLAOPTIONS_PARAM)) {
+      List<Map<String, Object>> slaOptionsObject = flowObj.getList(SLAOPTIONS_PARAM);
+      List<SlaOption> slaOptions = new ArrayList<>();
+      for(Map<String, Object> slaOption : slaOptionsObject) {
```
kunkun-tang: Why not use for(Map<String, Object> slaOption : flowObj.getList(SLAOPTIONS_PARAM))

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -259,7 +278,15 @@ public void fillExecutableFromMapObject(
     if (flowObj.containsKey(PROXYUSERS_PARAM)) {
       List<String> proxyUserList = flowObj.<String> getList(PROXYUSERS_PARAM);
       this.addAllProxyUsers(proxyUserList);
+    }
 
+    if (flowObj.containsKey(SLAOPTIONS_PARAM)) {
+      List<Map<String, Object>> slaOptionsObject = flowObj.getList(SLAOPTIONS_PARAM);
```
kunkun-tang: Why this variable is called slaOptionObject? is it not a list?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -115,6 +115,7 @@ protected ActionTypeLoader getActionLoader() {
   }
 
   public void insertTrigger(Trigger t) throws TriggerManagerException {
+    logger.info("Inserting trigger " + t + " in TriggerManager");
```
kunkun-tang: Why do we need this? I believe we have some other logger info messages during inserting.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -218,6 +230,13 @@ public void setSubmitTime(long submitTime) {
 
     flowObj.put(SUBMITTIME_PARAM, submitTime);
 
+    List<Map<String, Object>> slaOptions = new ArrayList<>();
+    for (SlaOption slaOption : this.getSlaOptions()) {
+      slaOptions.add(slaOption.toObject());
+    }
+
+    flowObj.put(SLAOPTIONS_PARAM, slaOptions);
```
HappyRay: Is this map serialized and saved to the  DB? Could you elaborate how the backward compatibility is handled here and in the change description?

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,47 +120,6 @@ public void setContext(Map<String, Object> context) {
     this.context = context;
   }
 
-  public Trigger(long lastModifyTime, long submitTime, String submitUser,
```
HappyRay: Nice clean up.

Would it be better to keep the clean up code in a separate pull request?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -304,12 +274,12 @@ private void onTriggerTrigger(Trigger t) throws TriggerManagerException {
       List<TriggerAction> actions = t.getTriggerActions();
       for (TriggerAction action : actions) {
         try {
-          logger.info("Doing trigger actions");
+          logger.info("Doing trigger actions " + action.getDescription() + " for " + t);
```
HappyRay: This is great. It was very difficult to debug without the trigger id. Thanks!

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.PriorityBlockingQueue;
+import org.apache.log4j.Logger;
+
+
+public class FlowExecutionMonitorManager extends Thread {
```
HappyRay: Is it simpler to name it FlowExecutionMonitor?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
```diff
@@ -0,0 +1,116 @@
...
+      triggers.add(new Trigger(execId, triggerCond, expireCond, actions));
+    }
+  }
+
+  /**
+   * Check all triggers. If no trigger is available then wait.
+   * Remove expired triggers from the queue if any.
+   *
+   */
+  private void checkAllTriggers() throws InterruptedException {
```
suvodeep-pyne: So you dequeue all elements and then add them back after you exhaust the entire loop?

Let's consider a scenario
 - Trigger `A(time: 15)`
 - Trigger `B(time: 50)`

After processing A, the trigger time is set to lets say `A(time: 30)`

But your code will fire B first and then look for A. right? That means the trigger for A is missed. Correct me if I'm wrong here.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
suvodeep-pyne: Also, use composition instead of inheritance. I don't see why we should inherit a Java thread and all its internals for a class which has a much more specific use case.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
```diff
@@ -0,0 +1,116 @@
...
+      if (first == second) {
+        return 0;
+      } else if (first < second) {
+        return 1;
+      }
+      return -1;
+    }
+  }
+
+  public void addMonitor(int execId, List<SlaOption> slaOptions) {
```
suvodeep-pyne: nit:
I have read this logic earlier in the code. It still goes right over my head. Some doc here briefly describing what it does would be helpful. Else breaking it down into smaller functions would make it easy to read.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
suvodeep-pyne: @HappyRay 
I am thinking more on the lines of the name `TriggerManager`. This class is running through the triggers and firing them. It is not 'technically' monitoring a flow.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -194,6 +195,10 @@ public FlowRunnerManager(Props props, ExecutorLoader executorLoader,
             AzkabanExecutorServer.JOBTYPE_PLUGIN_DIR,
             JobTypeManager.DEFAULT_JOBTYPEPLUGINDIR), globalProps,
             parentClassLoader);
+
+    executionMonitorManager = new FlowExecutionMonitorManager();
```
suvodeep-pyne: nit:
This class can be Guice injected. I leave it to you if you wanna do this as part of this change. This is exactly what I would refer to as a class 'dependency'

azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
```diff
@@ -0,0 +1,94 @@
...
+
+import azkaban.trigger.Condition;
+import azkaban.trigger.TriggerAction;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import org.apache.log4j.Logger;
+
+
+public class Trigger implements Runnable {
```
suvodeep-pyne: +1. These classes seem to have similar data. We can use one to wrap the other. Or have a single Trigger class and have 'controller' kind of classes to work with these objects. thoughts?

azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Check if the trigger is expired and reset isExpired
+   * @return true if trigger is expired
+   */
+  public boolean isTriggerExpired() {
+    isExpired = isExpired || expireCondition.isMet();
+    return isExpired;
+  }
+
+  public String toString() {
```
suvodeep-pyne: nit: Use StringBuilder all the way?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowExecutionMonitorManagerTest.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Test
+  public void addMonitorTest() {
+    FlowExecutionMonitorManager monitorManager = createMonitorManager();
+    Trigger[] triggers = monitorManager.getAllTriggers();
+    Assert.assertEquals(3, triggers.length);
+
+    Set<String> expectedTriggerStr = new HashSet<>();
+    expectedTriggerStr.add("Trigger for execution 1 with trigger condition of slaFailChecker.isSlaFailed() and expire condition of slaPassChecker.isSlaPassed(), KillExecutionAction for 1");
```
suvodeep-pyne: I think a style checker would help here but the max length of a line is 120 characters.

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
```diff
@@ -239,63 +239,17 @@ public void doAction() throws Exception {
...
+
     try {
       logger.info("Invoking flow " + project.getName() + "." + flowName);
       executorManager.submitExecutableFlow(exflow, submitUser);
       logger.info("Invoked flow " + project.getName() + "." + flowName);
     } catch (ExecutorManagerException e) {
       throw new RuntimeException(e);
     }
-
-    // deal with sla
```
suvodeep-pyne: @chengren311 : where does this logic move to?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -258,33 +245,16 @@ public void run() {
     }
 
     private void checkAllTriggers() throws TriggerManagerException {
-      long now = System.currentTimeMillis();
-
       // sweep through the rest of them
       for (Trigger t : triggers) {
         try {
           scannerStage = "Checking for trigger " + t.getTriggerId();
```
chengren311: shouldSkip here is never used after setting by this piece code. So I just removed it.

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
chengren311: it's moved tolowExecutionMonitorManager.java#addMonitor which is in executor.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.PriorityBlockingQueue;
+import org.apache.log4j.Logger;
+
+
+public class FlowExecutionMonitorManager extends Thread {
+  private static Logger logger = Logger.getLogger(FlowExecutionMonitorManager.class);
+  private BlockingQueue<Trigger> triggers = null;
+  private static final Duration CHECK_INTERVAL = Duration.ofMillis(15 * 1000);
```
chengren311: I see. thanks. IMO this should be a fixed value, making it configurable exposes the complexity to user.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
chengren311: i don't think we are on the same page. Let's discuss offline.

azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
chengren311: We will deprecate old trigger when we have distributed scheduler ready.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
suvodeep-pyne: sounds good. btw. you have all sorts of helper methods - ofDays, ofHours, ofMinutes, etc.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
chengren311: lots of places in this code uses this pattern. created an independent refactor request to track.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowExecutionMonitorManager.java
chengren311: triggermanager is better to serve the current purpose. But intention is to move all flow-related monitoring functionality here. I will call it trigger manager for now.

chengren311:
splitting this change into multiple PRs. closing this one.

PR #997, opened 2017-04-20
[Prototype code] Storage API: Add hash to Storage Metadata

suvodeep-pyne:
Adding hash information to storage metadata since it may be required during storage.

**Please note** that the code changes here do not affect any module since they are not yet wired up and thus can't get executed. [No static code injection]


suvodeep-pyne:
pausing this for now.

PR #998, opened 2017-04-20
Implemented put API for Database Storage

suvodeep-pyne:
Refactored Upload code to route upload files via the new Storage APIs.

Changes
 - Implemented `put` API in `DatabaseStorage`
 - Wired StorageManager to take in all upload requests. All requests are routed to the `DatabaseStorage` by default.
 - Add injections for `ProjectManager`
 - Updated `ProjectLoader` API

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -49,7 +49,7 @@ public class AzkabanCommonModule extends AbstractModule {
 
   public AzkabanCommonModule(Props props) {
     this.props = props;
-    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, LOCAL.name());
+    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, DATABASE.name());
   }
 
   @Override
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -16,6 +16,7 @@
 
 package azkaban.project;
 
+import com.google.common.io.Files;
 import com.google.inject.Inject;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
@@ -350,20 +351,23 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
   }
 
   @Override
-  public void uploadProjectFile(Project project, int version, String filetype,
-      String filename, File localFile, String uploader)
+  public void uploadProjectFile(int projectId, int version, File localFile, String uploader)
       throws ProjectManagerException {
     long startMs = System.currentTimeMillis();
-    logger.info("Uploading to " + project.getName() + " version:" + version
-        + " file:" + filename);
+    logger.info(String.format("Uploading Project ID: %d file: %s [%d bytes]",
+        projectId, localFile.getName(), localFile.length()));
     Connection connection = getConnection();
 
     try {
-      uploadProjectFile(connection, project, version, filetype, filename,
-          localFile, uploader);
+      /* Update DB with new project info */
+      addProjectToProjectVersions(connection, projectId, version, localFile, uploader);
+
+      uploadProjectFile(connection, projectId, version, localFile);
+
       connection.commit();
-      logger.info("project " + project.getName() + " commiting upload " + localFile.getName()
-              + " took " + ((System.currentTimeMillis() - startMs) / 1000) + " seconds.");
+      long duration = (System.currentTimeMillis() - startMs) / 1000;
+      logger.info(String.format("Uploaded Project ID: %d file: %s [%d bytes] in %d sec",
+          projectId, localFile.getName(), localFile.length(), duration));
     } catch (SQLException e) {
       logger.error(e);
       throw new ProjectManagerException("Error getting DB connection.", e);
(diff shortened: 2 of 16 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -462,27 +488,32 @@ private void uploadProjectFile(Connection connection, Project project,
         size = bufferedStream.read(buffer);
       }
     } catch (IOException e) {
-      throw new ProjectManagerException("Error chunking file " + filename);
+      throw new ProjectManagerException("Error chunking file " + localFile.getName());
```
chengren311: we can add chunk here as well.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -56,22 +57,23 @@ private File getFile(URI key) {
...
+        metadata.getVersion() + "." + Files.getFileExtension(localFile.getName()));
 
     if (targetFile.exists()) {
       throw new StorageException(String.format(
           "Error in LocalStorage. Target file already exists. targetFile: %s, Metadata: %s",
           targetFile, metadata));
     }
     try {
-      FileUtils.copyInputStreamToFile(is, targetFile);
+      FileUtils.copyInputStreamToFile(new FileInputStream(localFile), targetFile);
```
chengren311: Should we need to close the FileInputStream being constructed here?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -193,18 +196,22 @@ public AzkabanWebServer() throws Exception {
    * Constructor
    */
   public AzkabanWebServer(Server server, Props props) throws Exception {
```
chengren311: should we make the constructor private since this class seems to be implemented as singleton

azkaban-spi/src/main/java/azkaban/spi/StorageMetadata.java
```diff
@@ -23,31 +23,31 @@
 
 
 public class StorageMetadata {
-  private final String projectId;
-  private final String version;
-  private final String extension;
+  private final int projectId;
+  private final int version;
+  private final String uploader;
```
chengren311: what is uploader here? why it's uploader a string?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
suvodeep-pyne: Good point. Will add it. I am not too worried since the chunking logic, etc will go away when we move to the new storage.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
suvodeep-pyne: Thanks. fixed it.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: The constructor is used externally. It is complicated and that's why I didn't refactor this part as part of this change.

azkaban-spi/src/main/java/azkaban/spi/StorageMetadata.java
suvodeep-pyne: It is a hack to get `DatabaseStorage` to work without too much trouble. It may not be required later on. However, adding more metadata / context should hurt the code.

suvodeep-pyne:
@chengren311 updated PR with your comments.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -36,7 +36,7 @@
    * @return
    * @throws ProjectManagerException
    */
-  public List<Project> fetchAllActiveProjects() throws ProjectManagerException;
```
chengren311: why? what if we want to access those methods outside of azkaban-common?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -159,12 +163,11 @@
...
-  // private ExecutorManagerAdapter executorManager;
-  private ExecutorManager executorManager;
-  private ScheduleManager scheduleManager;
-  private TriggerManager triggerManager;
+  private final UserManager userManager;
+  private final ProjectManager projectManager;
+  private final ExecutorManager executorManager;
+  private final ScheduleManager scheduleManager;
+  private final TriggerManager triggerManager;
   private Map<String, Alerter> alerters;
```
chengren311: should alerters be final as well?

chengren311: also MBeanServer mbeanServer and triggerPlugins

azkaban-common/src/main/java/azkaban/storage/DatabaseStorage.java
```diff
@@ -33,9 +34,12 @@
...
+  private final ProjectLoader projectLoader;
+
   @Inject
-  public DatabaseStorage(JdbcProjectLoader jdbcProjectLoader) {
+  public DatabaseStorage(ProjectLoader projectLoader) {
 
+    this.projectLoader = projectLoader;
   }
 
   @Override
```
chengren311: why is getter not implemented? should we throw NotImplementedException here?

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
suvodeep-pyne: ??
java interface methods are public by default.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: added final to the remaining ones. Although this is not a change about refactoring `AzkabanWebServer`. Thus, am not keen on making too many changes there.

azkaban-common/src/main/java/azkaban/storage/DatabaseStorage.java
suvodeep-pyne: because the pull request is on `put` :)
added `UnsupportedOperationException`

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -350,20 +351,23 @@ private synchronized Project createNewProject(Connection connection,
   }
 
   @Override
-  public void uploadProjectFile(Project project, int version, String filetype,
-      String filename, File localFile, String uploader)
+  public void uploadProjectFile(int projectId, int version, File localFile, String uploader)
```
suvodeep-pyne: Just wanted to add that the refactor in the JdbcProjectLoader code simply splits the upload method into multiple smaller methods.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -49,7 +49,7 @@
 
   public AzkabanCommonModule(Props props) {
     this.props = props;
-    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, LOCAL.name());
+    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, DATABASE.name());
```
kunkun-tang: I would prefer DATASTORAGE over DATABASE regarding naming.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -372,14 +376,43 @@ public void uploadProjectFile(Project project, int version, String filetype,
     }
   }
 
-  private void uploadProjectFile(Connection connection, Project project,
-      int version, String filetype, String filename, File localFile,
+  private void uploadProjectFile(Connection connection, int projectId, int version, File localFile)
```
kunkun-tang: Understand that we need to do version initialization in a separate metho, but can't see the benefits why we need a separate method to have the two steps below

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -193,18 +196,22 @@ public AzkabanWebServer() throws Exception {
    * Constructor
    */
   public AzkabanWebServer(Server server, Props props) throws Exception {
-    this.props = props;
+    this.props = requireNonNull(props);
```
kunkun-tang: May we also consider using java 8 type annotations, like @NonNull  ?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: This line changes the default storage implementation type from `LOCAL` file system storage to `DATABASE` which is the MySQL storage used today.

`DATASTORAGE` is generic and doesn't convey anything about the implementation detail.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
suvodeep-pyne: I don' think we do apart from that the outer api does connection handling, etc making the inner function simpler.

suvodeep-pyne: Oh, btw. This was an existing method wich I didn't remove. There are other refactors in my mind but I'll probably do that in stages.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: I think that's a good idea. btw can you point me to the annotation that you are referring to? I'm unable to find the annotation in our current classpath.

PR #1006, opened 2017-04-23
Fix ClassCanBeStatic warnings in common sub-project

HappyRay:
See http://errorprone.info/bugpattern/ClassCanBeStatic

e.g.
/Users/ruyang/oss/azkaban/azkaban-common/src/test/java/azkaban/utils/RestfulApiClientTest.java:47:
warning: [ClassCanBeStatic] Inner class is non-static but does not
reference enclosing class
  protected class MockRestfulApiClient extends RestfulApiClient<String>
  {
              ^
	          (see
		  http://errorprone.info/bugpattern/ClassCanBeStatic)
		    Did you mean 'protected static class
		    MockRestfulApiClient extends
		    RestfulApiClient<String> {'?

Fix:
Made these inner classes static inner classes.
Also removed protected from a few static classes introduced.

See http://stackoverflow.com/questions/24289070/why-we-should-not-use-protected-static-in-java

These classes only need package private access.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
@@ -36,7 +36,7 @@ import azkaban.utils.JSONUtils;
 
 public class SelectorTest {
   // mock executor object.
-  protected static class MockExecutorObject implements Comparable <MockExecutorObject>{
+  static class MockExecutorObject implements Comparable <MockExecutorObject>{
     public String name;
     public int    port;
     public double percentOfRemainingMemory;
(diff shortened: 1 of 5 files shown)


azkaban-common/src/test/java/azkaban/executor/SelectorTest.java
```diff
@@ -78,7 +78,7 @@ public int compareTo(MockExecutorObject o) {
   }
 
   // Mock flow object.
-  protected class MockFlowObject{
+  protected static class MockFlowObject{
```
suvodeep-pyne: `protected static` seems a bit confusing.
http://stackoverflow.com/questions/24289070/why-we-should-not-use-protected-static-in-java

Should we just make it package private or public (if it is accessed externally)?
There are other similar changes in this PR.

HappyRay: Good point. Will investigate.

HappyRay: Changed. Thanks.

PR #1011, opened 2017-04-24
Increasing the job cancel time from 5 sec to 30 sec

suvodeep-pyne:
For some jobs, eg hadoop related, 5 sec is not enough for the jobs to
finish their clean up and exit. Increasing the time to 30 sec in code
to temporarily mitigate issue.


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -40,7 +41,7 @@
 
   public static final String COMMAND = "command";
 
-  private static final long KILL_TIME_MS = 5000;
+  static final Duration KILL_TIME = Duration.ofSeconds(30);
```
HappyRay: Why not keeping this variable private?

PR #1013, opened 2017-04-25
Refactored AzkabanWebServer and AzkabanExecServer launch code

suvodeep-pyne:
This change refactors the initialization of the server instances. The current code in these classes is quite entangled. Breaking this up into multiple commits. This is a prerequisite to the upcoming Guice changes.
 - removed duplicate methods from Azkaban*Server classes
 - moved some of the references constants to `Constants` class
 - config was getting updated during server initialization. Fixed that. The same change is now applied when we create the config object `Props`
 - refactored IntelliJ suggested warnings which simplify the code
 - extracted methods and created smaller methods to make it easier to navigate and isolate code

Tested on solo server as well as web and executor separately on dev instances.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
+++ b/azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
@@ -33,6 +33,9 @@ import azkaban.user.UserManager;
 import azkaban.utils.Props;
 import azkaban.server.session.SessionCache;
 
+import static azkaban.Constants.*;
+
+
 public abstract class AzkabanServer {
   private static final Logger logger = Logger.getLogger(AzkabanServer.class);
   private static Props azkabanProperties = null;
@@ -47,12 +50,11 @@ public abstract class AzkabanServer {
   }
 
   public static Props loadProps(String[] args, OptionParser parser) {
-    ;
-    OptionSpec<String> configDirectory =
-        parser
-            .acceptsAll(Arrays.asList("c", "conf"),
-                "The conf directory for Azkaban.").withRequiredArg()
-            .describedAs("conf").ofType(String.class);
+    OptionSpec<String> configDirectory = parser.acceptsAll(
+        Arrays.asList("c", "conf"), "The conf directory for Azkaban.")
+        .withRequiredArg()
+        .describedAs("conf")
+        .ofType(String.class);
 
     // Grabbing the azkaban settings from the conf directory.
     Props azkabanSettings = null;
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
```diff
@@ -75,12 +77,27 @@ public static Props loadProps(String[] args, OptionParser parser) {
       azkabanSettings = loadConfigurationFromAzkabanHome();
     }
 
+    if (azkabanSettings != null) {
```
chengren311: i would prefer doing null check inside the method

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
```diff
@@ -118,17 +132,14 @@ private static Props loadConfigurationFromAzkabanHome() {
       logger.error("AZKABAN_HOME not set. Will try default.");
       return null;
     }
-
-    if (!new File(azkabanHome).isDirectory()
-        || !new File(azkabanHome).canRead()) {
+    if (!new File(azkabanHome).isDirectory() || !new File(azkabanHome).canRead()) {
```
chengren311: thanks for reformatting this. maybe we will need to rely on intellij plugin to do this.

HappyRay:
LGTM. 

Because a large block of code is moved around, it is hard to review. It would be great if you can describe in the change description the general strategy you used in this change.

HappyRay:
Thanks for the cleanup!

suvodeep-pyne:
@HappyRay 
Most of these are IntelliJ refactors and extract methods. I'll update the description.

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
HappyRay: What's the advantage?

Shall we throw an exception if the setting is null here? 
If so, I would suggest doing it in another pull request to keep refactoring and functional change separate.

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
HappyRay: Yes. Intellij can automatically reformat code.

@chengren311 Do you still plan to set up a code format standard?

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
suvodeep-pyne: This is a functional change in itself - perhaps the only one in the PR. But this is required for the refactor.

The existing code modifies the config while setting up the server. This creates a spaghetti flow and I had to disentangle this to make that flow somewhat linear. The null check is required else some test fails. I don't want to focus on that now. The current direction is to make the initialization simple so that I can introduce Guice at the root level. This would simplify refactoring drastically since a lot of the entanglement is due to static referencing of data/ members of one class from another. This will be evident in subsequent PRs

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
chengren311: yes, it's on my plate.

suvodeep-pyne: I think we need that. I didn't autoformat the entire file though - because we don't have a standard style yet. Once we establish that, then we can have a single format change and update everything. Also we can add style check validations to make our life simpler.

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
chengren311: I think validation of the argument inside the method is a better practice.

suvodeep-pyne: I agree it is. I'll fix it. We need to change the entire args processing code anyway. The current code flow is very unclean and needs to be restructured. I'll add validations when I implement the new classes. Treat this as a stop gap change.

PR #1015, opened 2017-04-25
Adding root level Guice injection to main server components

suvodeep-pyne:
Root level Guice injection has been added to web server, executor and
solo server. This change ensures that the main server classes are
injected as opposed to manually constructed. This can help us refactor
the code further and introduce the major classes as Guice dependencies.

The code has been tested on solo, web and executor instances both
locally and on the grid. Unit testing these modules are currently not
easy due to the code structure. However, Guicification will help us a
lot in moving towards that direction.


azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+public class WebServerProvider implements Provider<Server> {
+  private static final Logger logger = Logger.getLogger(WebServerProvider.class);
+  private static final int MAX_HEADER_BUFFER_SIZE = 10 * 1024 * 1024;
+
+  @Inject
+  private Props props;
+
+  @Override
+  public Server get() {
```
suvodeep-pyne: This code is not new. It is cut paste from `AzkabanWebServer`

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -68,6 +72,7 @@ protected void configure() {
     }
   }
 
+  @SuppressWarnings("unchecked")
```
HappyRay: Where does the warning come from?

suvodeep-pyne: Unchecked cast
```
       return (Class<? extends Storage>) Class.forName(storageImplementation);
```

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -1261,6 +1205,13 @@ public void close() {
     }
     scheduleManager.shutdown();
     executorManager.shutdown();
+    try {
+      server.stop();
+    } catch (Throwable t) {
```
chengren311: catch throwable or exception? just curious about the difference

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -681,18 +671,28 @@ public static void main(String[] args) throws Exception {
...
+    SERVICE_PROVIDER.setInjector(Guice.createInjector(
+        new AzkabanCommonModule(props),
+        new AzkabanWebServerModule()
+    ));
 
-    app = new AzkabanWebServer(server, azkabanSettings);
+    launch(props);
+  }
+
+  public static void launch(Props azkabanSettings) throws Exception {
```
suvodeep-pyne: This is being called by the solo server main. The solo server used the call the `main` methods for both web and exec servers.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: `Throwable` is the mother class of all `catch`able classes in java. `Exception` is a subclass of Throwable. The idea is to catch all errors at this point since we are trying to shutdown the server.

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -45,6 +45,7 @@
   public static final int MEMORY_CHECK_RETRY_LIMIT = 720;
   public static final int DEFAULT_PORT_NUMBER = 8081;
   public static final int DEFAULT_SSL_PORT_NUMBER = 8443;
+  public static final int DEFAULT_THREAD_NUMBER = 20;
```
suvodeep-pyne: Yes. This made sense when it was inside the web server code. updated it.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
HappyRay: Thanks

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -681,18 +671,28 @@ public static void main(String[] args) throws Exception {
...
+    /* This creates the Web Server instance */
+    app = SERVICE_PROVIDER.getInstance(AzkabanWebServer.class);
 
-    prepareAndStartServer(azkabanSettings, server);
+    // TODO refactor code into ServerProvider
```
HappyRay: In the future, please add your name to the todo comment. See our todo code convention.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
HappyRay: Why is catching throwable here better than catching just RuntimeException?

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
HappyRay: Thanks for the comment. It makes review much easier.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: I wasn't aware of it. Let's discuss offline.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: Just wanted to ensure that rest of the cleanup is executed no matter what.

HappyRay: Even when an Error is thrown?
see https://www.cis.upenn.edu/~bcpierce/courses/629/jdkdocs/api/java.lang.Error.html
An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it.

suvodeep-pyne: I see. Thanks ! 
I'll fix it in my next PR.

suvodeep-pyne:
Thanks @HappyRay

HappyRay:
This is great. I like the new code much better. Thanks!

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
suvodeep-pyne: @HappyRay 
The method call itself throws `Exception` so we cannot just catch `RuntimeException`. I have already made the suggested changes and pushed it #1019

Issue #1009, opened 2017-04-24
The common metrics test is flaky


HappyRay:
See https://travis-ci.org/azkaban/azkaban/builds/220548106
azkaban.metrics.CommonMetricsTest > testMarkDBConnectionMetrics FAILED
java.lang.AssertionError at CommonMetricsTest.java:48

PR #1018, opened 2017-04-25
Fix flaky metrics related tests

HappyRay:
#1009 #
Instead of using a reporter which is subject to the multi-thread
related timing issues, verify that the metrics registry has the
 correct information. We don't have to test that reporter works here.

Redesigned the tests to make the tests more explicit and more closely
reflect the actual usage pattern of the metrics related methods.
The existing shared test routines make strong assumptions about the
behavior of the functional methods passed to them. IMHO they make the
test code harder to read and are error-prone.

The meter related tests are a bit trickier to test since they
return double type.
The existing meter related tests only test the count not the actual
value reported in the AutoMetrics anyway.
Remove these tests for now until we have a better solution.

Removed the empty Executor Metrics test class.


ameyamk:
LGTM.

For meter based tests can we use more relaxed asserts? eg. instead of `assertTrue(1.55 == value)` use `asertTrue(1< value && value < 2)`

HappyRay:
I am thinking of using assertj's floating point assertions. 
But that is not the only problem. That part of the testing doesn't seem to be a high priority right now.  So I am delaying that part.

PR #1019, opened 2017-04-26
Wiring get calls via StorageManager

suvodeep-pyne:
This changes delegates all get calls via the `StorageManager`. By `get` I mean the process of fetching a project file (blob) from storage. By default, this storage is the database. There are some hacks in the code which can be fixed later as we go.

There are some refactors in the executor codebase. This was done to easily fetch the `StorageManager` class via Guice.

The changes have been verified on solo server locally and also on executors separately.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/storage/StorageManager.java
+++ b/azkaban-common/src/main/java/azkaban/storage/StorageManager.java
@@ -19,7 +19,6 @@ package azkaban.storage;
 
 import azkaban.project.Project;
 import azkaban.project.ProjectFileHandler;
-import azkaban.project.ProjectLoader;
 import azkaban.spi.Storage;
 import azkaban.spi.StorageMetadata;
 import azkaban.user.User;
@@ -36,12 +35,10 @@ public class StorageManager {
   private static final Logger log = Logger.getLogger(StorageManager.class);
 
   private final Storage storage;
-  private final ProjectLoader projectLoader;
 
   @Inject
-  public StorageManager(Storage storage, ProjectLoader projectLoader) {
+  public StorageManager(Storage storage) {
     this.storage = storage;
-    this.projectLoader = projectLoader;
   }
 
   /**
@@ -77,6 +74,10 @@ public class StorageManager {
    * @return Handler object containing hooks to fetched project file
    */
   public ProjectFileHandler getProjectFile(final int projectId, final int version) {
-    return projectLoader.getUploadedFile(projectId, version);
+    // TODO remove huge hack !
+    if (storage instanceof DatabaseStorage) {
+      return ((DatabaseStorage) storage).get(projectId, version);
+    }
+    throw new UnsupportedOperationException("Operation currently unsupported for other types.");
   }
 }
--- a/azkaban-common/src/main/java/azkaban/storage/StorageManager.java
+++ b/azkaban-common/src/main/java/azkaban/storage/StorageManager.java
@@ -75,7 +75,7 @@ public class StorageManager {
    */
   public ProjectFileHandler getProjectFile(final int projectId, final int version) {
     log.info(String.format("Fetching project file. project ID: %d version: %d", projectId, version));
-    // TODO remove huge hack !
+    // TODO spyne: remove huge hack ! There should not be any special handling for Database Storage.
     if (storage instanceof DatabaseStorage) {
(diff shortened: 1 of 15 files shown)


azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -71,4 +65,20 @@ public void uploadProject(
...
+   *
+   * @param projectId required project ID
+   * @param version version to be fetched
+   * @return Handler object containing hooks to fetched project file
+   */
+  public ProjectFileHandler getProjectFile(final int projectId, final int version) {
+    log.info(String.format("Fetching project file. project ID: %d version: %d", projectId, version));
+    // TODO spyne: remove huge hack ! There should not be any special handling for Database Storage.
+    if (storage instanceof DatabaseStorage) {
+      return ((DatabaseStorage) storage).get(projectId, version);
```
chengren311: what is the problem of just storage.get(projectId, version) ?

suvodeep-pyne: There is no problem as such. It is just that I would prefer storage key to get a single parameter, `URI` in this case.

For example in LOCAL or HDFS storage, I would get the artifact based on path.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -193,7 +197,7 @@ public FlowRunnerManager(Props props, ExecutorLoader executorLoader,
         new JobTypeManager(props.getString(
             AzkabanExecutorServer.JOBTYPE_PLUGIN_DIR,
             JobTypeManager.DEFAULT_JOBTYPEPLUGINDIR), globalProps,
-            parentClassLoader);
+            getClass().getClassLoader());
```
kunkun-tang: Don't understand. paranetClassLoader sounds indicating some class in upper level, not this clas?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -41,19 +41,17 @@
 public class FlowPreparer {
```
kunkun-tang: Looks like this class should not be called outside the its current package. Should we consider remove public?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -314,11 +309,6 @@ private void loadCustomJMXAttributeProcessor(Props props) {
     }
   }
 
-
-  public ProjectLoader getProjectLoader() {
-    return projectLoader;
-  }
-
   public ExecutorLoader getExecutorLoader() {
```
kunkun-tang: I'm seeing that ExecutorServlet calls this method. Should we inject executorLoader there in near future?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
suvodeep-pyne: by parent the code was referring to `AzkabanExecServer` class. There was no code initializing `FlowRunnerManager` with a different classloader. It was a simple new call. This means that that classloader for `AzkabanExecServer` and `FlowRunnerManager` will always be the same. Therefore sending it via argument seems redundant. This is also true further down the flow but I didn't want to refactor too much in this change since the PR is for a different purpose.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
suvodeep-pyne: I agree but by that logic, we may have to change the visibility of a lot of classes. I left it defaulted to public for now since that is the usual convention for classes.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
suvodeep-pyne: Yes. we should. There are a lot of classes initialized in `AzkabanExecServer` and other top level services just to pass arguments around. These wirings can simply be removed by using Guice.

PR #1021, opened 2017-04-26
Azkaban-db subproject introduction

kunkun-tang:
This patch introduces akzbana-db subproject, which defines new SQL
interfaces to interact with JDBC libraries. The goal of this design is
to have database retires more easily and add more monitoring metrics in
future.

Part of tests are created for review. Will have more tests later.

These code is not called by any production logics yet. 

todo: will move azkaban-sql to azkaban-db subproject

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -16,9 +16,6 @@
  */
 package azkaban;
 
-import azkaban.db.AzkabanDataSource;
-import azkaban.db.H2FileDataSource;
-import azkaban.db.MySQLDataSource;
 import azkaban.db.DatabaseOperator;
 import azkaban.db.DatabaseOperatorImpl;
 
@@ -27,7 +24,6 @@ import azkaban.project.ProjectLoader;
 import azkaban.spi.Storage;
 import azkaban.spi.StorageException;
 import azkaban.storage.LocalStorage;
-import azkaban.storage.StorageConfig;
 import azkaban.storage.StorageImplementationType;
 import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
@@ -37,7 +33,6 @@ import com.google.inject.Scopes;
 import java.io.File;
 import org.apache.commons.dbutils.QueryRunner;
 
-import static azkaban.storage.StorageImplementationType.*;
 
 /**
  * This Guice module is currently a one place container for all bindings in the current module. This is intended to
@@ -45,42 +40,28 @@ import static azkaban.storage.StorageImplementationType.*;
  * structuring of Guice components.
  */
 public class AzkabanCommonModule extends AbstractModule {
-  private final Props props;
-  /**
-   * Storage Implementation
-   * This can be any of the {@link StorageImplementationType} values in which case {@link StorageFactory} will create
-   * the appropriate storage instance. Or one can feed in a custom implementation class using the full qualified
-   * path required by a classloader.
-   *
-   * examples: LOCAL, DATABASE, azkaban.storage.MyFavStorage
-   *
-   */
-  private final String storageImplementation;
-
-  private final QueryRunner queryRunner;
+  private final AzkabanCommonModuleConfig config;
 
   public AzkabanCommonModule(Props props) {
-    this.props = props;
-    this.queryRunner = new QueryRunner(getDataSource(props));
-    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, DATABASE.name());
+    this.config = new AzkabanCommonModuleConfig(props);
   }
 
   @Override
   protected void configure() {
     bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
-    bind(Props.class).toInstance(props);
+    bind(Props.class).toInstance(config.getProps());
     bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
     bind(DatabaseOperator.class).to(DatabaseOperatorImpl.class).in(Scopes.SINGLETON);
     //todo kunkun-tang : Consider both H2 DataSource and MysqlDatasource case.
-    bind(QueryRunner.class).toInstance(queryRunner);
+    bind(QueryRunner.class).toInstance(config.getQueryRunner());
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
-    final StorageImplementationType type = StorageImplementationType.from(storageImplementation);
+    final StorageImplementationType type = StorageImplementationType.from(config.getStorageImplementation());
     if (type != null) {
       return type.getImplementationClass();
     } else {
-      return loadCustomStorageClass(storageImplementation);
+      return loadCustomStorageClass(config.getStorageImplementation());
     }
   }
 
@@ -95,29 +76,7 @@ public class AzkabanCommonModule extends AbstractModule {
 
   @Inject
   public @Provides
-  LocalStorage createLocalStorage(StorageConfig config) {
-    return new LocalStorage(new File(config.getBaseDirectoryPath()));
-  }
-
-  // todo kunkun-tang: the below method should moved out to azkaban-db module eventually.
-  // Today azkaban-db can not rely on Props, so we can not do it.
-  private AzkabanDataSource getDataSource(Props props) {
-    String databaseType = props.getString("database.type");
-
-    // todo kunkun-tang: temperaroy workaround to let service provider test work.
-    if(databaseType.equals("h2")) {
-      String path = props.getString("h2.path");
-      return new H2FileDataSource(path);
-    }
-    int port = props.getInt("mysql.port");
-    String host = props.getString("mysql.host");
-    String database = props.getString("mysql.database");
-    String user = props.getString("mysql.user");
-    String password = props.getString("mysql.password");
-    int numConnections = props.getInt("mysql.numconnections");
-
-    return MySQLDataSource.getInstance(host, port, database, user, password,
-            numConnections);
-
+  LocalStorage createLocalStorage(AzkabanCommonModuleConfig config) {
+    return new LocalStorage(new File(config.getLocalStorageBaseDirPath()));
   }
 }
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -19,8 +19,8 @@ package azkaban;
 import azkaban.db.AzkabanDataSource;
 import azkaban.db.H2FileDataSource;
 import azkaban.db.MySQLDataSource;
-import azkaban.db.AzDBOperator;
-import azkaban.db.AzDBOperatorImpl;
+import azkaban.db.DatabaseOperator;
+import azkaban.db.DatabaseOperatorImpl;
 
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
@@ -35,6 +35,7 @@ import com.google.inject.Inject;
 import com.google.inject.Provides;
 import com.google.inject.Scopes;
 import java.io.File;
+import org.apache.commons.dbutils.QueryRunner;
 
 import static azkaban.storage.StorageImplementationType.*;
 
@@ -56,11 +57,11 @@ public class AzkabanCommonModule extends AbstractModule {
    */
   private final String storageImplementation;
 
-  private final AzkabanDataSource dataSource;
+  private final QueryRunner queryRunner;
 
   public AzkabanCommonModule(Props props) {
     this.props = props;
-    this.dataSource = getDataSource(props);
+    this.queryRunner = new QueryRunner(getDataSource(props));
     this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, DATABASE.name());
   }
 
@@ -69,9 +70,9 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
     bind(Props.class).toInstance(props);
     bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
-    bind(AzDBOperator.class).to(AzDBOperatorImpl.class).in(Scopes.SINGLETON);
+    bind(DatabaseOperator.class).to(DatabaseOperatorImpl.class).in(Scopes.SINGLETON);
     //todo kunkun-tang : Consider both H2 DataSource and MysqlDatasource case.
-    bind(AzkabanDataSource.class).toInstance(dataSource);
+    bind(QueryRunner.class).toInstance(queryRunner);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
@@ -100,7 +101,7 @@ public class AzkabanCommonModule extends AbstractModule {
 
   // todo kunkun-tang: the below method should moved out to azkaban-db module eventually.
   // Today azkaban-db can not rely on Props, so we can not do it.
-  private static AzkabanDataSource getDataSource(Props props) {
+  private AzkabanDataSource getDataSource(Props props) {
     String databaseType = props.getString("database.type");
 
     // todo kunkun-tang: temperaroy workaround to let service provider test work.
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -16,6 +16,12 @@
  */
 package azkaban;
 
+import azkaban.db.AzkabanDataSource;
+import azkaban.db.H2FileDataSource;
+import azkaban.db.MySQLDataSource;
+import azkaban.db.AzDBOperator;
+import azkaban.db.AzDBOperatorImpl;
+
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
 import azkaban.spi.Storage;
@@ -28,7 +34,6 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Inject;
 import com.google.inject.Provides;
 import com.google.inject.Scopes;
-import com.google.inject.Singleton;
 import java.io.File;
 
 import static azkaban.storage.StorageImplementationType.*;
@@ -47,9 +52,12 @@ public class AzkabanCommonModule extends AbstractModule {
    */
   private final String storageImplementation;
 
+  private final AzkabanDataSource dataSource;
+
   public AzkabanCommonModule(Props props) {
     this.props = props;
     this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, LOCAL.name());
+    this.dataSource = getDataSource(props);
   }
 
   @Override
@@ -57,6 +65,9 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
     bind(Props.class).toInstance(props);
     bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
+    bind(AzDBOperator.class).to(AzDBOperatorImpl.class).in(Scopes.SINGLETON);
+    //todo kunkun-tang : Consider both H2 DataSource and MysqlDatasource case.
+    bind(AzkabanDataSource.class).toInstance(dataSource);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
@@ -80,5 +91,27 @@ public class AzkabanCommonModule extends AbstractModule {
   public @Provides
   LocalStorage createLocalStorage(StorageConfig config) {
     return new LocalStorage(new File(config.getBaseDirectoryPath()));
+  }
+
+  // todo kunkun-tang: the below method should moved out to azkaban-db module eventually.
+  // Today azkaban-db can not rely on Props, so we can not do it.
+  private static AzkabanDataSource getDataSource(Props props) {
+    String databaseType = props.getString("database.type");
+
+    // todo kunkun-tang: temperaroy workaround to let service provider test work.
+    if(databaseType.equals("h2")) {
+      String path = props.getString("h2.path");
+      return new H2FileDataSource(path);
+    }
+    int port = props.getInt("mysql.port");
+    String host = props.getString("mysql.host");
+    String database = props.getString("mysql.database");
+    String user = props.getString("mysql.user");
+    String password = props.getString("mysql.password");
+    int numConnections = props.getInt("mysql.numconnections");
+
+    return MySQLDataSource.getInstance(host, port, database, user, password,
+            numConnections);
+
   }
 }
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -38,7 +38,11 @@ import java.io.File;
 
 import static azkaban.storage.StorageImplementationType.*;
 
-
+/**
+ * This Guice module is currently a one place container for all bindings in the current module. This is intended to
+ * help during the migration process to Guice. Once this class starts growing we can move towards more modular
+ * structuring of Guice components.
+ */
 public class AzkabanCommonModule extends AbstractModule {
   private final Props props;
   /**
@@ -56,8 +60,8 @@ public class AzkabanCommonModule extends AbstractModule {
 
   public AzkabanCommonModule(Props props) {
     this.props = props;
-    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, LOCAL.name());
     this.dataSource = getDataSource(props);
+    this.storageImplementation = props.getString(Constants.ConfigurationKeys.AZKABAN_STORAGE_TYPE, DATABASE.name());
   }
 
   @Override
@@ -79,6 +83,7 @@ public class AzkabanCommonModule extends AbstractModule {
     }
   }
 
+  @SuppressWarnings("unchecked")
   private Class<? extends Storage> loadCustomStorageClass(String storageImplementation) {
     try {
       return (Class<? extends Storage>) Class.forName(storageImplementation);
--- a/azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
+++ b/azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package azkaban.db;
+
+import java.sql.SQLException;
+import org.apache.commons.dbutils.ResultSetHandler;
+
+/**
+ * This interface is to define Base Data Access Object contract for Azkaban. All azkaban
+ * DB related operations must be performed upon this interface. AZ DB operators all leverages
+ * QueryRunner interface.
+ *
+ * @see org.apache.commons.dbutils.QueryRunner
+ */
+public interface AzDBOperator {
+
+  /**
+   * returns the last id from a previous insert statement.
+   * Note: The implementation must catch SQL Exception itself.
+   *
+   * @return the last inserted id in mysql per connection.
+   */
+  Long getLastInsertId();
+
+  /**
+   * Executes the given AZ related SELECT SQL statements.
+   *
+   * @param sqlQuery The SQL query statement to execute.
+   * @param resultHandler The handler used to create the result object
+   * @param params Initialize the PreparedStatement's IN parameters
+   * @param <T> The type of object that the qeury handler returns
+   * @return The object returned by the handler.
+   * @throws SQLException
+   */
+  <T> T query(String sqlQuery,
+      ResultSetHandler<T> resultHandler,
+      Object...params) throws SQLException;
+
+  /**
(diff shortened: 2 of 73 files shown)


azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.commons.dbutils.ResultSetHandler;
+
+/**
+ * This interface is to define Base Data Access Object contract for Azkaban. All azkaban
+ * DB related operations must be performed upon this interface. AZ DB operators all leverages
+ * QueryRunner interface.
+ *
+ * @see org.apache.commons.dbutils.QueryRunner
+ */
+public interface AzDBOperator {
```
suvodeep-pyne: I don't think we have a naming convention yet, but this is probably difficult to guess or search even in IntelliJ. You can simply call it `AzkabanDatabaseOperator`. Or since we are already inside the Azkaban package, may be just `DatabaseOperator`. 

Having consecutive capitals make sense of abbreviations like `URI` or something.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -86,4 +97,26 @@ protected void configure() {
   LocalStorage createLocalStorage(StorageConfig config) {
     return new LocalStorage(new File(config.getBaseDirectoryPath()));
   }
+
+  // todo kunkun-tang: the below method should moved out to azkaban-db module eventually.
+  // Today azkaban-db can not rely on Props, so we can not do it.
+  private static AzkabanDataSource getDataSource(Props props) {
```
suvodeep-pyne: suggestion: remove static and make this a provider like the method just above this `createLocalStorage` The idea is to initialize only the minimal set of classes like config classes etc and leave the rest to Guice.

suvodeep-pyne: If you foresee this method to grow, then make it a provider class. example: `ServerProvider`

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -86,4 +97,26 @@ protected void configure() {
...
+    String databaseType = props.getString("database.type");
+
+    // todo kunkun-tang: temperaroy workaround to let service provider test work.
+    if(databaseType.equals("h2")) {
```
suvodeep-pyne: Do we need to specify hard coded settings for h2 and mysql? Are other database types not supported ?

azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Executes the given AZ related INSERT, UPDATE, or DELETE SQL statement.
+   *
+   * @param updateClause sql statements to execute
+   * @param params Initialize the PreparedStatement's IN parameters
+   * @return The number of rows updated.
+   * @throws SQLException
+   */
+  int update(String updateClause,
```
suvodeep-pyne: single line ?

azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * returns the last id from a previous insert statement.
+   * Note: The implementation must catch SQL Exception itself.
+   *
+   * @return the last inserted id in mysql per connection.
+   */
+  Long getLastInsertId();
+
+  /**
+   * Executes the given AZ related SELECT SQL statements.
```
suvodeep-pyne: Again, I think it would be better to use Azkaban instead of short names. `AZ` or `Az`, etc . The consistency helps.

suvodeep-pyne: SQL statements ? Can one provide multiple statements? If so, how?

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.commons.dbutils.ResultSetHandler;
+import org.apache.commons.dbutils.handlers.ScalarHandler;
+import com.google.inject.Inject;
+
+import static java.util.Objects.*;
+
+/**
+ * Implement AZ DB related operations. This class is thread safe.
+ */
+public class AzDBOperatorImpl implements AzDBOperator {
```
suvodeep-pyne: Just curious: What other kind of implementations do you foresee ?

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * transaction method Implementation.
+   *
+   * @param operations A sequence of DB operations
+   * @param <T> object type
+   * @return transaction result
+   * @throws SQLException
+   */
+  @Override
+  public <T> T transaction(SQLTransaction<T> operations) throws SQLException {
```
suvodeep-pyne: I really like the javadoc but it is Ok to leave out the stuff that is already documented in the interface.

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * update implementation. it will call {@link AzkabanDataSource#getConnection()} inside queryrunner.update.
+   *
+   * @param updateClause sql statements to execute
+   * @param params Initialize the PreparedStatement's IN parameters
+   * @return the number of rows being affected by update
+   * @throws SQLException
+   */
+  @Override
+  public int update(String updateClause,
```
suvodeep-pyne: single line ?

azkaban-db/src/main/java/azkaban/db/AzDBTransOperator.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+/**
+ * This interface is designed as an supplement of {@link AzDBOperator}, which do commit at the end of every query. Given
+ * this interface, users/callers (implementation code) should decide where to {@link Connection#commit()}
+ * based on their requirements.
+ *
+ * @see org.apache.commons.dbutils.QueryRunner
+ */
+public interface AzDBTransOperator {
```
suvodeep-pyne: Same comment as above

azkaban-db/src/main/java/azkaban/db/AzDBTransOperatorImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public <T> T query(String querySql, ResultSetHandler<T> resultHandler, Object... params) throws SQLException {
+    try{
+      return queryRunner.query(conn, querySql, resultHandler, params);
+    } catch (SQLException ex){
+      //RETRY Logic should be implemented here if needed.
+      throw ex;
+    } finally {
+      // Note: CAN NOT CLOSE CONNECTION HERE.
```
suvodeep-pyne: +1 for comment.

azkaban-db/src/main/java/azkaban/db/AzDBTransOperatorImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import java.sql.Connection;
+import java.sql.SQLException;
+import org.apache.commons.dbutils.QueryRunner;
+import org.apache.commons.dbutils.ResultSetHandler;
+
+
+/**
+ * This class should be only used inside {@link AzDBOperatorImpl}, then we remove public.
+ */
+class AzDBTransOperatorImpl implements AzDBTransOperator {
```
suvodeep-pyne: I don't completely understand how this works. Let's discuss offline.

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package azkaban.db;
+
+public class AzDBUtil {
```
suvodeep-pyne: call it `Constants`? I specifically avoid creating `Util` classes which become a static dumpground for people to dump random unrelated methods into.

I think we are redoing some of the infrastructure in this module. How about @HappyRay 's idea of pulling out a core package and depending on that?

azkaban-db/src/main/java/azkaban/db/AzkabanDataSource.java
```diff
@@ -0,0 +1,22 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.db;
+
+import org.apache.commons.dbcp2.BasicDataSource;
+
+public abstract class AzkabanDataSource extends BasicDataSource {
+  public abstract String getDBType();
```
suvodeep-pyne: Why do we need this ? And why String ?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * This method overrides {@link BasicDataSource#getConnection()}, in order to have retry logics.
+   *
+   */
+  @Override
+  public synchronized Connection getConnection() throws SQLException {
+
+      /*
```
suvodeep-pyne: comment is aligned incorrectly I think ?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,104 @@
...
+    setUrl(url);
+    setMaxTotal(numConnections);
+    setValidationQuery("/* ping */ select 1");
+    setTestOnBorrow(true);
+  }
+
+  /**
+   * Get a singleton object for MySQL BasicDataSource
+   */
+  public static MySQLDataSource getInstance(String host, int port, String dbName,
```
suvodeep-pyne: This should be Guice injected

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+public class AzDBOperatorImplTest {
+
+  private AzkabanDataSource datasource = new AzDBTestUtility.EmbeddedH2BasicDataSource();
+
+  private AzDBOperator dbOperator;
+  private QueryRunner queryRunner;
+  private Connection conn;
+
+  private ResultSetHandler<Integer> handler = rs -> {
+    if (!rs.next()) {
```
suvodeep-pyne: nit: can be shortened to a single line.

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+      return 0;
+    }
+
+    return rs.getInt(1);
+  };
+
+  private static final List<Integer> list = new ArrayList<>();
+
+  private static int index_1 = 3;
```
suvodeep-pyne: ??

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  @Test
+  public void testInvalidQuery() throws Exception {
+    int res = dbOperator.query("select * from blah where ? = ?", handler, "id", 3);
+    Assert.assertEquals(0, res);
+  }
+
+  @Test(expected = SQLException.class)
+  public void testTypoSqlStatement() throws Exception {
```
suvodeep-pyne: nice ! 👍

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    when(queryRunner.getDataSource()).thenReturn(mockDataSource);
+    when(mockDataSource.getConnection()).thenReturn(conn);
+
+    this.dbOperator = new AzDBOperatorImpl(queryRunner);
+
+    list.add(index_1);
+    list.add(index_2);
+
+    // valid query returns correct value
+    when(queryRunner.query("select * from blah where ? = ?", handler, "id", 2)).thenReturn(index_2);
```
suvodeep-pyne: I like the table name :)

azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
kunkun-tang: DatabaseOperator sounds good to me. changing.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
kunkun-tang: Thanks for suggestions.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
kunkun-tang: We don't support other databases for now.

azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
kunkun-tang: Yes.

azkaban-db/src/main/java/azkaban/db/AzDBOperator.java
kunkun-tang: Multiple statements are supported by `DatabaseTranOperator`.

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
kunkun-tang: This is a very good question. As we are able to control retry logics. we might have different retry strategies during implementation. Today, we enforce using DBUtils to do queries. We might consider unlock the restriction to allow more query libraries.

azkaban-db/src/main/java/azkaban/db/AzkabanDataSource.java
kunkun-tang: It could be useful when we print out a datasource's name.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: This style was recommended by intellij for some reason. Just googled. every line of comment should start by `*`.

Modified.

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
kunkun-tang: How it transform to one line? No idea about this.

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
kunkun-tang: :)

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public interface DatabaseOperator {
+
+  /**
+   * returns the last id from a previous insert statement.
+   * Note: The implementation must catch SQL Exception itself.
+   *
+   * @return the last inserted id in mysql per connection.
+   */
+  Long getLastInsertId();
```
chengren311: Long or long?

kunkun-tang: good catch!

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+   *
+   */
+  @Override
+  public Long getLastInsertId() {
+    // A default connection: autocommit = true.
+    long num = -1;
+    try {
+      num = ((Number) queryRunner.query("SELECT LAST_INSERT_ID();", new ScalarHandler<>(1))).longValue();
+    } catch (SQLException ex) {
+      logger.error("can not get last insertion ID", ex);
```
chengren311: should it swallow the exception or throw the exception. IMO throwing exception seems more reasonable.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.commons.dbutils.ResultSetHandler;
+
+/**
+ * This interface is to define Base Data Access Object contract for Azkaban. All azkaban
+ * DB related operations must be performed upon this interface. AZ DB operators all leverages
+ * QueryRunner interface.
+ *
+ * @see org.apache.commons.dbutils.QueryRunner
+ */
+public interface DatabaseOperator {
```
chengren311: is this necessary to create an interface here?

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Note: this queryRunner should include a concrete {@link AzkabanDataSource} inside.
+   *
+   * @param queryRunner
+   */
+  @Inject
+  public DatabaseOperatorImpl(QueryRunner queryRunner){
+    this.queryRunner = queryRunner;
+    requireNonNull(queryRunner.getDataSource(), "data source must not be null.");
```
chengren311: i think it's better to put requireNonNull(queryRunner.getDataSource(), "data source must not be null.") before this.queryRunner = queryRunner.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
chengren311: it would be good to log more detailed info like db, table name..

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * query method Implementation. it will call {@link AzkabanDataSource#getConnection()} inside queryrunner.query.
+   */
+  @Override
+  public <T> T query(String baseQuery, ResultSetHandler<T> resultHandler, Object...params) throws SQLException {
+    try{
+      return queryRunner.query(baseQuery, resultHandler, params);
+    } catch (SQLException ex){
+      // todo kunkun-tang: Retry logics should be implemented here.
+      logger.error("query failed", ex);
```
chengren311: same here. log db/table name/query

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    try{
+      conn = queryRunner.getDataSource().getConnection();
+      conn.setAutoCommit(false);
+      DatabaseTransOperator transOperator = new DatabaseTransOperatorImpl(queryRunner, conn);
+      T res = operations.execute(transOperator);
+      conn.commit();
+      return res;
+    } catch (SQLException ex) {
+      // todo kunkun-tang: Retry logics should be implemented here.
+      logger.error("transaction failed", ex);
```
chengren311: same here

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+   * @return the number of rows being affected by update
+   * @throws SQLException
+   */
+  @Override
+  public int update(String updateClause, Object...params) throws SQLException {
+    try{
+      return queryRunner.update(updateClause, params);
+    } catch (SQLException ex){
+      // todo kunkun-tang: Retry logics should be implemented here.
+      logger.error("update failed", ex);
```
chengren311: same here

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: If something wrong happens, I will return -1 rather than throws an exception. The advantage is the caller doesn't need to deal with exception inside code.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
kunkun-tang: Yes, we could have multiple DatabaseOperator Implementations.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: Changed.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: The difficulty is that we don't have db, table name things inside this method.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: Good catch! Changing.

suvodeep-pyne:
@kunkun-tang 
I wasn't aware that this change was checked in. There are some issues that need to be addressed.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: I can't find the todo tag. Could you point me to it?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,105 @@
...
+      String user, String password, int numConnections) {
+    if (instance == null) {
```
HappyRay: This is not a safe pattern. 
Please see "effective java" book for details.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,105 @@
...
+      synchronized (MySQLDataSource.class) {
+        if (instance == null) {
+          logger.info("Instantiating MetricReportManager");
```
HappyRay: Is this log statement right? I don't see metrics manager being instantiated here.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,105 @@
...
+           * when DB connection could not be fetched here, dbcp library will keep searching until a timeout defined in
+           * its code hardly.
+           */
+        connection = createDataSource().getConnection();
+        if(connection != null)
+          return connection;
+      } catch (SQLException ex) {
+        logger.error("Failed to find DB connection. waits 1 minutes and retry. No.Attempt = " + retryAttempt, ex);
+      } finally {
+        retryAttempt ++;
```
HappyRay: Does this statement need to be in the finally block? Can it be put in the catch block?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+    return null;
```
HappyRay: Is it better to re-throw the sql exception when it gives up?
What's the effect of returning null?

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
```diff
@@ -0,0 +1,21 @@
...
+  static final int MAX_DB_RETRY_COUNT = 5;
```
HappyRay: Why 5?
Is it better to retry for a long time? What's the downside?

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * transaction method Implementation.
+   *
+   */
+  @Override
+  public <T> T transaction(SQLTransaction<T> operations) throws SQLException {
+    Connection conn = null;
+    try{
+      conn = queryRunner.getDataSource().getConnection();
+      conn.setAutoCommit(false);
```
HappyRay: Consider adding a comment explaining that autocommit will be reset for the next connection. And refer to the unit test for that behavior.

azkaban-db/src/main/java/azkaban/db/DatabaseTransOperator.java
```diff
@@ -0,0 +1,61 @@
...
+package azkaban.db;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import org.apache.commons.dbutils.ResultSetHandler;
+
+
+/**
+ * This interface is designed as an supplement of {@link DatabaseOperator}, which do commit at the end of every query. Given
+ * this interface, users/callers (implementation code) should decide where to {@link Connection#commit()}
```
HappyRay: Do you mean "when to commit"?

azkaban-db/src/test/java/azkaban/db/DatabaseOperatorImplTest.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Test(expected = SQLException.class)
+  public void testTypoSqlStatement() throws Exception {
+    System.out.println("testTypoSqlStatement");
+    dbOperator.query("sele * from blah where ? = ?", handler, "id", 2);
+  }
+
+  @Test
+  public void testTransaction() throws Exception {
```
HappyRay: This test doesn't seem to verify that the change is actually committed which is a key part of this method.

Can you improve it?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: Good catch! already removed.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: not necessary. Either method looks working.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: You are probably right. will evaluate if we should do exception. Sounds a better idea.

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
kunkun-tang: I was still actively testing this feature. Should be a very big number when eventually enabling this feature.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: Will add it.

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
kunkun-tang: This issue is documented at #1337 to track.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: This issue is documented at #1337 to track.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: Synced up offline. When SQL exception occurs, the SQLexception will print the compelte sql query to track.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: Synced up offline. When SQL exception occurs, the SQLexception will print the compelte sql query to track.

azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java
kunkun-tang: Synced up offline. When SQL exception occurs, the SQLexception will print the compelte sql query to track.

azkaban-db/src/test/java/azkaban/db/AzDBOperatorImplTest.java
kunkun-tang: Just some random number should be fine.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: Comments were added in #1040 . Thanks.

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
kunkun-tang: Right. I already made it as single line.

azkaban-db/src/main/java/azkaban/db/AzDBOperatorImpl.java
kunkun-tang: Rigjt. I removed useless comments out of Impl class.

azkaban-db/src/test/java/azkaban/db/DatabaseOperatorImplTest.java
kunkun-tang: Yes, you are right. The essential part of the transaction is to use a single connection to do multiple things at one time. For instance, if we use transaction to insert1, update2, delete3. When exception occurs to delete3, we should verify `insert1` didn't evenly happen since rollback works at the connection level. Documented at #1337 to track.

azkaban-db/src/main/java/azkaban/db/DatabaseTransOperator.java
kunkun-tang: I already rewrote this introductory comments in #1040 .

HappyRay:
Is it better to make DatabaseTransOperatorImpl package private?

HappyRay:
Could you add a comment to explain the autocommit behavior and expectations in 
azkaban.db.DatabaseOperatorImpl#transaction
https://github.com/HappyRay/azkaban/blob/5391230029a6f2be63e9a198aacf0ac1f786a8d0/azkaban-db/src/main/java/azkaban/db/DatabaseOperatorImpl.java#L69

      conn.setAutoCommit(false);

PR #1022, opened 2017-04-26
Refactor storage code. Added HDFS storage stub

suvodeep-pyne:
Refactor: Remove `StorageConfig` and move everything to `AzkabanCommonModuleConfig`. This creates a config container which can be used for all classes in azkaban-common. If the config grows too large, we can always refactor later. Besides currently there is no config class for the modules.

Added HDFS storage implementation stub. Following changes will implement the APIs.


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -21,54 +21,40 @@
...
 
   @Override
   protected void configure() {
     bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
-    bind(Props.class).toInstance(props);
+    bind(Props.class).toInstance(config.getProps());
     bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
```
kunkun-tang: Can' we put this method to config as well?

gradle.properties
```diff
@@ -20,3 +20,10 @@ org.gradle.parallel=true
 
 #Allows generation of idea/eclipse metadata for a specific subproject and its upstream project dependencies
 ide.recursive=true
+
+
+#---------------------------------
+# Versions
+#---------------------------------
+hadoopVersion=2.6.1
+hiveVersion=0.13.1
```
kunkun-tang: Will we also use hive/Dali ?

azkaban-hadoop-security-plugin/build.gradle
```diff
@@ -1,12 +1,10 @@
...
-ext.hadoopVersion = "2.6.1"
-
 dependencies {
   compile project(":azkaban-common")
 
   compileOnly "org.apache.hadoop:hadoop-common:$hadoopVersion"
   compileOnly "org.apache.hadoop:hadoop-mapreduce-client-common:$hadoopVersion"
   compileOnly "org.apache.hadoop:hadoop-mapreduce-client-core:$hadoopVersion"
-  compileOnly "org.apache.hive:hive-metastore:0.13.1"
+  compileOnly "org.apache.hive:hive-metastore:$hiveVersion"
```
kunkun-tang: How metastore help us?

suvodeep-pyne: It's not. I'm just pulling the hadoop and hive versions out to `gradle.properties`.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: I was thinking of it initially. But I also don't want too much business logic inside a config class. I'm leaving class loader specific code inside the module because that's what a module should do.

gradle.properties
suvodeep-pyne: We do! as part of plugins! For now, this change is just pulling out these versions to an external file so that we can easily change these and also maintain a consistent version across all modules.

PR #1024, opened 2017-04-27
Storage schema changes and implementation. Working Local Storage.

suvodeep-pyne:
The `project_versions` table currently stores all metadata corresponding to every upload of a project zip. The newly introduced storage layer in azkaban requires the persistence of an additional field in the `project_versions` table `uri`.
```
CREATE TABLE project_versions (
...
	uri VARCHAR(512) DEFAULT NULL,
...
);
```

The changes integrate `uri` in the query APIs in `ProjectLoader` and modify the interfaces in a *backward incompatible* manner. The solo server SQL scripts have been updated accordingly and have been tested to work.

Existing installations of Azkaban *will* require a simple DB migration: `uri` field addition to `project_versions`.

This completes the implementation for `LOCAL` and `DATABASE` storage types. By default, the storage will continue to be `DATABASE`. (DB migration will still be required)

To change to `LOCAL` storage. Add this to `conf/azkaban.properties`
```
azkaban.storage.type=LOCAL
```

This has been tested to work on solo server and also tested on dev clusters

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -360,7 +360,7 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
 
     try {
       /* Update DB with new project info */
-      addProjectToProjectVersions(connection, projectId, version, localFile, uploader);
+      addProjectToProjectVersions(connection, projectId, version, localFile, uploader, null);
 
       uploadProjectFile(connection, projectId, version, localFile);
 
@@ -385,6 +385,21 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
     updateChunksInProjectVersions(connection, projectId, version, chunks);
   }
 
+  public void addProjectVersion(
+      int projectId,
+      int version,
+      File localFile,
+      String uploader,
+      String uri) throws ProjectManagerException {
+    try (Connection connection = getConnection()) {
+      addProjectToProjectVersions(connection, projectId, version, localFile, uploader, uri);
+    } catch (SQLException e) {
+      logger.error(e);
+      throw new ProjectManagerException(String.format("Add ProjectVersion failed. project id: %d version: %d",
+          projectId, version), e);
+    }
+  }
+
   /**
    * Insert a new version record to TABLE project_versions before uploading files.
    *
@@ -404,11 +419,8 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
    * When we upload a new project zip in day2, new file in day 2 will use the new version (proj_v + 1).
    * When file uploading completes, AZ will clean all old chunks in DB afterward.
    */
-  private void addProjectToProjectVersions(Connection connection,
-      int projectId,
-      int version,
-      File localFile,
-      String uploader) throws ProjectManagerException {
+  private void addProjectToProjectVersions(Connection connection, int projectId, int version, File localFile,
+      String uploader, String uri) throws ProjectManagerException {
     final long updateTime = System.currentTimeMillis();
     QueryRunner runner = new QueryRunner();
     logger.info("Creating message digest for upload " + localFile.getName());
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -385,6 +385,22 @@ private void uploadProjectFile(Connection connection, int projectId, int version
...
+      String uri) throws ProjectManagerException {
+    try (Connection connection = getConnection()) {
+      addProjectToProjectVersions(connection, projectId, version, localFile, uploader, uri);
+      connection.commit();
+    } catch (SQLException e) {
+      logger.error(e);
+      throw new ProjectManagerException(String.format("Add ProjectVersion failed. project id: %d version: %d",
+          projectId, version), e);
+    }
+  }
```
kunkun-tang: I believe, in this case, you should explicitly call `DBUtils.close(connection)` here. Otherwise, you could call createQueryRunner(), then you don't need to worry about conn.commit, and conn.close. In future, we will use DatabaseOperator to replace.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -404,11 +420,8 @@ private void uploadProjectFile(Connection connection, int projectId, int version
    * When we upload a new project zip in day2, new file in day 2 will use the new version (proj_v + 1).
    * When file uploading completes, AZ will clean all old chunks in DB afterward.
    */
-  private void addProjectToProjectVersions(Connection connection,
-      int projectId,
-      int version,
-      File localFile,
-      String uploader) throws ProjectManagerException {
+  private void addProjectToProjectVersions(Connection connection, int projectId, int version, File localFile,
```
kunkun-tang: Why last method has multiple line method signature, but this one does not.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -530,42 +544,41 @@ public ProjectFileHandler getUploadedFile(int projectId, int version)
...
+          ProjectVersionResultHandler.SELECT_PROJECT_VERSION, pfHandler, projectId, version);
+      if (projectFiles == null || projectFiles.isEmpty()) {
+        return null;
+      }
+      return projectFiles.get(0);
     } catch (SQLException e) {
       logger.error(e);
-      throw new ProjectManagerException(
-          "Query for uploaded file for project id " + projectId + " failed.", e);
+      throw new ProjectManagerException("Query for uploaded file for project id " + projectId + " failed.", e);
```
kunkun-tang: I think `DBUtils.close` should be called here as well.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -63,7 +90,23 @@ public void uploadProject(
...
     log.info(String.format("Adding archive to storage. Meta:%s File: %s[%d bytes]",
         metadata, localFile.getName(), localFile.length()));
-    storage.put(metadata, localFile);
+
+    /* upload to storage */
+    URI uri = storage.put(metadata, localFile);
+
+    /* Add metadata to db */
+    // TODO spyne: remove hack. Database storage should go through the same flow
+    if (!(storage instanceof DatabaseStorage)) {
```
kunkun-tang: Should addProjectVersion eventually be placed inside storage.put, right?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
suvodeep-pyne: @kunkun-tang 
I think `close()` will be automatically called on connection since it is being called in a [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) mechanism. Let me know if you still think it's a problem

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
suvodeep-pyne: I guess this was due to IntelliJ doing its thing based on code style. Fixed it.

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
suvodeep-pyne: same argument as above.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
suvodeep-pyne: No. The way I am approaching this design is that the storage shouldn't be bothered about putting metadata into Azkaban DB. It should simply take care of storing the file and returning a valid uri to reference it. Currently the metadata management is being done inside the `DatabaseStorage`. That's the reason for the hack.

PR #1026, opened 2017-04-27
Cleaning up warnings generated during build process

li-afaris:
This is to address the build warnings in issue#1025.

I focused on two areas related to build warnings.  
1) NPM
2) Override annotations.

#### NPM
While I was able to fix a licensing issue with NPM, there's still a version dependency issue with eonasdan-bootstrap-datetimepicker.  Instead of passing in a 0.4.x version of  moment-timezone, we are giving it a 0.5.x version which causes the dependency check to fail.  

#### Override
I got most of the override warnings but didn’t feel comfortable with blindly putting @Override everywhere the warning occurred.  Hence, there are still a remainder of Override warnings.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -624,6 +624,7 @@ public class ExecutorManager extends EventHandler implements
     }
   }
 
+  @Override
   public List<ExecutableFlow> getRecentlyFinishedFlows() {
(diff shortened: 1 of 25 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1225,6 +1226,7 @@ private void shutdown() {
       shutdown = true;
     }
 
+    @Override
     @SuppressWarnings("unchecked")
```
suvodeep-pyne: should we also get rid of `@SuppressWarnings`here ?

PR #1039, opened 2017-04-28
Refactor the code for getting OS free memory information

HappyRay:
Replaced SwapCache with SwapFree
See https://bbs.archlinux.org/viewtopic.php?id=133604
SwapCache doesn't represent the available swap space.

Switched the logger to lsf4j.

Will make another pull request to switch to Guice.

Fixed all the intellij and ErrorProne warnings.
including:

/Users/ruyang/oss/azkaban/azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java:44: warning: [FutureReturnValueIgnored] Return value of methods returning Future must be checked.
      scheduledExecutorService.scheduleAtFixedRate(new MemoryInfoReader(), 0, memCheckInterval, TimeUnit.SECONDS);
                                                  ^
    (see http://errorprone.info/bugpattern/FutureReturnValueIgnored)
  Did you mean '@SuppressWarnings("unused")'?
/Users/ruyang/oss/azkaban/azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java:108: warning: [DefaultCharset] Implicit use of the platform default charset, which can result in e.g. non-ASCII characters being silently replaced with '?' in many environments
      br = new BufferedReader(new FileReader(MEMINFO_FILE));
                              ^
    (see http://errorprone.info/bugpattern/DefaultCharset)

Testing:
Built on both Linux system and Mac.
Ran acceptance tests on the test cluster.


azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -22,34 +17,46 @@
  * All the memory size used in this function is in KB
  */
 public class SystemMemoryInfo {
-  private static final Logger logger = Logger.getLogger(SystemMemoryInfo.class);
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(SystemMemoryInfo.class);
```
suvodeep-pyne: I think slf4j is good. But should we do it as part of a global change?

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
```diff
@@ -22,34 +17,46 @@
...
+  private static final long LOW_MEM_THRESHOLD = 3L * 1024L * 1024L; //3 GB
+  // In case there is a problem reading the meminfo file, we want to "fail open".
+  private static long freeMemAmount = LOW_MEM_THRESHOLD * 100;
 
   private static ScheduledExecutorService scheduledExecutorService;
 
+  //todo HappyRay: switch to Guice
+  private static OsMemoryUtil util = new OsMemoryUtil();
+
+  @SuppressWarnings("FutureReturnValueIgnored")
```
suvodeep-pyne: so this is how it's ignored. Thanks 👍

azkaban-common/src/test/java/azkaban/utils/OsMemoryUtilTest.java
```diff
@@ -0,0 +1,60 @@
...
+    assertEquals(500, size);
+  }
+
+  @Test
+  public void parseIncorrectMemoryLine() {
+    String line = "MemFree:        ab kB";
+    long size = util.parseMemoryLine(line);
+    assertEquals(0, size);
+  }
+}
```
suvodeep-pyne: nit: missing newline

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -0,0 +1,110 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Utility class for getting system memory information
+ */
+class OsMemoryUtil {
```
HappyRay: I found the suggested name harder to read.
I am fine with the change if it is part of the naming convention. Could you point me to the naming convention doc that mandates this?

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(SystemMemoryInfo.class);
+
+  static final String MEMINFO_FILE = "/proc/meminfo";
```
HappyRay: No. It is never designed to work on Mac. The check will be disabled when the file doesn't exist.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private enum MemKey {MemFree, Buffers, Cached, SwapFree}
```
HappyRay: Will fix.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -0,0 +1,110 @@
...
+      logger.warn(errMsg);
+      totalFree = 0;
+    }
+    return totalFree;
+  }
+
+  /**
+   * Example file:
+   * $ cat /proc/meminfo
+   MemTotal:       65894008 kB
```
HappyRay: The IDE doesn't do it for me automatically. 
I didn't find the manual effort worth it.  I can change if you insist or suggest a way for tools to do it for me.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -0,0 +1,110 @@
...
+   Cached:          4290236 kB
+   SwapCached:            0 kB
+
+   * Make the method package private to make unit testing easier.
+   * Otherwise it can be made private.
+
+   * @param line the text for a memory usage statistics we are interested in
+   * @return size of the memory. unit kB. 0 if there is an error.
+   */
+  long parseMemoryLine(String line) {
```
HappyRay: Yes, same here. I did some research on the same.

Turns out that there is no platform independent way to do so.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
HappyRay: Why global change if we can do it incrementally?

azkaban-common/src/test/java/azkaban/utils/OsMemoryUtilTest.java
HappyRay: Will fix.

HappyRay: Found this http://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
suvodeep-pyne: That's cool. There is no mandate as such. 
There are examples like `sun.security.x509.IPAddressName` but there are counter examples like `com.sun.jmx.snmp.SnmpIpAddress` in Java itself.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
suvodeep-pyne: I see. So this is kind of Linux only. We can probably call that out then.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
suvodeep-pyne: yes. That's a pain. There is a quick way to do this though. Let's chat offline.

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
suvodeep-pyne: Having this change incrementally can create confusion. I am not sure whether this helps us migrate from one to another. Rather it may lead to a codebase with both loggers in use.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
HappyRay: Will add a comment.

HappyRay: Added.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
HappyRay: Changed to an array.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
HappyRay: The column editing mode works. Thansk!

azkaban-common/src/main/java/azkaban/utils/SystemMemoryInfo.java
HappyRay: We already have both in place. I think it will be good to require new code to use slf4j. What do you think?

suvodeep-pyne: yes. That we can enforce.

PR #1041, opened 2017-04-28
move sql files from azkaban-sql to azkaban-db

kunkun-tang:
This patch consolidate the azkaban-sql and azkaban-db modules into one place. Basically we move sql files from azkaban-sql into `azkaban-db/src/main/sql`. Tested by `gradle build`. Also, these files is able to appear in the right place (Distribution zip) after whatever `gradle installDist` or `gradle build`.


azkaban-db/build.gradle
```diff
@@ -27,3 +28,33 @@ dependencies {
...
+ * concat task
+ * Concatenate all the sql queries / files (except update.*.sql) and collect it into a single
+ * SQL file.
+ */
+task concat() << {
+  ext.destFile = 'build/sql/create-all-sql-' + version + '.sql';
+  ant.concat(destfile: destFile, fixlastline: 'yes') {
+    logger.info('Concating create scripts to ' + destFile)
+    fileset(dir: 'src/main/sql') {
+      exclude(name: 'update.*.sql')
```
jamiesjc: It seems there is no 'update' sql. Do we still need to exclude that?

jamiesjc:
LGTM

azkaban-db/build.gradle
kunkun-tang: Good Catch. Removing.

PR #1046, opened 2017-05-01
Implement HDFS Storage for Azkaban

suvodeep-pyne:
This patch implements HDFS storage for Azkaban.

Config parameters introduced:
```
# URI root for HDFS storage.
azkaban.storage.hdfs.root.uri=hdfs://localhost:9000/path/to/foo

# Hadoop conf dir
hadoop.conf.dir.path=/path/to/hadoop/etc/hadoop
```

Local and HDFS example project filepath: `<project_id>/<project_id>_<md5_hash>.zip`

### DIFF
--- a/azkaban-common/build.gradle
+++ b/azkaban-common/build.gradle
@@ -35,6 +35,11 @@ configurations {
   testCompileOnly.extendsFrom(compileOnly)
 }
 dependencies {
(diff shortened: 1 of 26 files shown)


suvodeep-pyne:
moving to #1069

PR #1047, opened 2017-05-02
Simplify the logic for getting system memory information

HappyRay:
There is no need to gather the information in a separate thread.
It's simpler just to read the file every time.

Switched to Guice to construct these classes.

Added unit tests.

Switched MEM_KEYS from array to ImmutableSet.

Changed some logging to error level from warning level since they are
unexpected conditions.


HappyRay:
This is a strange test failure on Travis:

"azkaban.utils.OsMemoryUtilTest > canReadMemInfoFileIfExists FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at azkaban.utils.OsMemoryUtilTest.canReadMemInfoFileIfExists(OsMemoryUtilTest.java:20)

258 tests completed, 1 failed, 21 skipped
:azkaban-common:test FAILED"

The same test worked fine on my Linux test machine.

And Travis didn't catch the real test failure somehow previously. See my second change set.

PR #1051, opened 2017-05-03
Changing Storage API key from URI to String

suvodeep-pyne:
Motivation: String is more generic and gives flexibility to the storage implementation to structure their own keys. 
For example, the storage layer may use a JSON string as an identifier. This is not currently a use case, but going forward it would probably be better to not impose an unnecessary restriction on a top level API.

Tested separately on dev clusters.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -391,9 +391,9 @@ public class JdbcProjectLoader extends AbstractJdbcLoader implements
       int version,
       File localFile,
       String uploader,
-      String uri) throws ProjectManagerException {
+      String storage_key) throws ProjectManagerException {
     try (Connection connection = getConnection()) {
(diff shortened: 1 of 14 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -426,7 +426,7 @@ private void addProjectToProjectVersions(Connection connection,
       int version,
       File localFile,
       String uploader,
-      String uri) throws ProjectManagerException {
+      String storage_key) throws ProjectManagerException {
```
ameyamk: storageLocation perhaps?

ameyamk: other idea is storageLocatorKey ?

suvodeep-pyne: updated to `resourceId`

suvodeep-pyne:
Thanks @chengren311 . updated PR description.

PR #1053, opened 2017-05-03
Add org.slf4j:slf4j-log4j12 run time dependency to select modules

HappyRay:
This allows loggers that use slf4j to log properly.

Also added log4j.properties files to test resources to log to console
in tests by default.

Symptom:
example warning when running tests:
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
log4j:WARN No appenders could be found for logger (azkaban.jobExecutor.JavaProcessJob).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.

Testing:
Here is the console output from running a test

2017/05/03 08:35:08.586 -0700 ERROR [DatabaseOperatorImpl] query failed
java.sql.SQLException

### DIFF
--- a/azkaban-common/build.gradle
+++ b/azkaban-common/build.gradle
@@ -70,6 +70,7 @@ dependencies {
   testCompile(project(':azkaban-test').sourceSets.test.output)
 
   testRuntime('com.h2database:h2:1.4.193')
+  testRuntime('org.slf4j:slf4j-log4j12:1.7.18')
 }
 
 tasks.withType(JavaCompile) {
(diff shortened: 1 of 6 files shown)


HappyRay:
solo server already has the file.

PR #1055, opened 2017-05-04
Minor Refactor. Added a unit test for Web server instantiation

suvodeep-pyne:
Changes:
 - Added `ExecutorLoader` binding to `AzkabanCommonModule`
 - Added a test to ensure that instantiation of WebServer is successful
 - Removed duplicate and derived dependencies from web-server build.gradle

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -19,6 +19,8 @@ package azkaban;
 import azkaban.db.DatabaseOperator;
 import azkaban.db.DatabaseOperatorImpl;
 
+import azkaban.executor.ExecutorLoader;
+import azkaban.executor.JdbcExecutorLoader;
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
 import azkaban.spi.Storage;
@@ -48,6 +50,7 @@ public class AzkabanCommonModule extends AbstractModule {
 
   @Override
   protected void configure() {
+    bind(ExecutorLoader.class).to(JdbcExecutorLoader.class).in(Scopes.SINGLETON);
     bind(ProjectLoader.class).to(JdbcProjectLoader.class).in(Scopes.SINGLETON);
     bind(Props.class).toInstance(config.getProps());
     bind(Storage.class).to(resolveStorageClassType()).in(Scopes.SINGLETON);
--- a/azkaban-web-server/build.gradle
+++ b/azkaban-web-server/build.gradle
@@ -66,9 +66,7 @@ dependencies {
   generateRestli('com.linkedin.pegasus:generator:' + pegasusVersion)
   generateRestli('com.linkedin.pegasus:restli-tools:' + pegasusVersion)
 
-  // Needed by Velocity at runtime
-  testRuntime('commons-collections:commons-collections:3.2.2')
-  testCompile('org.hamcrest:hamcrest-all:1.3')
+  testRuntime "com.h2database:h2:1.4.193"
 }
 
 sourceSets {
--- a/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package azkaban.webapp;
+
+import azkaban.AzkabanCommonModule;
+import azkaban.Constants;
+import azkaban.database.AzkabanDatabaseSetup;
+import azkaban.database.AzkabanDatabaseUpdater;
+import azkaban.executor.Executor;
+import azkaban.executor.ExecutorLoader;
+import azkaban.utils.Props;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static azkaban.ServiceProvider.*;
+import static azkaban.executor.ExecutorManager.*;
+import static java.util.Objects.*;
+import static org.apache.commons.io.FileUtils.*;
+import static org.junit.Assert.*;
+
+
+public class AzkabanWebServerTest {
+  public static final String AZKABAN_DB_SQL_PATH = "azkaban-db/src/main/sql";
+
+  private static final Props props = new Props();
+
+  private static String getUserManagerXmlFile() {
+    URL resource = AzkabanWebServerTest.class.getClassLoader().getResource("azkaban-users.xml");
+    return requireNonNull(resource).getPath();
+  }
+
+  private static String getSqlScriptsDir() throws IOException {
+    // Dummy because any resource file works.
+    final String dummyResourcePath = getUserManagerXmlFile();
+    Path resources = Paths.get(dummyResourcePath).getParent();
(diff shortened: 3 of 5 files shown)


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -48,6 +50,7 @@ public AzkabanCommonModule(Props props) {
 
   @Override
   protected void configure() {
+    bind(ExecutorLoader.class).to(JdbcExecutorLoader.class).in(Scopes.SINGLETON);
```
HappyRay: Please mention this functional change in the description.

azkaban-web-server/build.gradle
```diff
@@ -66,9 +66,7 @@ dependencies {
   generateRestli('com.linkedin.pegasus:generator:' + pegasusVersion)
   generateRestli('com.linkedin.pegasus:restli-tools:' + pegasusVersion)
 
-  // Needed by Velocity at runtime
```
HappyRay: Are they no longer needed? 
This change is not mentioned in the change description. This may make it harder to trace problems later.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.junit.Assert.*;
+
+
+public class AzkabanWebServerTest {
+  public static final String AZKABAN_DB_SQL_PATH = "azkaban-db/src/main/sql";
+
+  private static final Props props = new Props();
+
+  private static String getUserManagerXmlFile() {
+    URL resource = AzkabanWebServerTest.class.getClassLoader().getResource("azkaban-users.xml");
```
HappyRay: FYI I remember Gauva has a convenient class to load resources.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    return requireNonNull(resource).getPath();
```
HappyRay: Nice usage of assertions.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    Path resources = Paths.get(dummyResourcePath).getParent();
+    Path azkabanRoot = resources.getParent().getParent().getParent().getParent();
+
+    File sqlScriptDir = Paths.get(azkabanRoot.toString(), AZKABAN_DB_SQL_PATH).toFile();
+    return props.getString(AzkabanDatabaseSetup.DATABASE_SQL_SCRIPT_DIR, sqlScriptDir.getCanonicalPath());
+  }
+
+  @BeforeClass
+  public static void setUp() throws Exception {
+    tearDown();
```
HappyRay: Is this necessary?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    String sqlScriptsDir = getSqlScriptsDir();
+    props.put(AzkabanDatabaseSetup.DATABASE_SQL_SCRIPT_DIR, sqlScriptsDir);
+
+    props.put("database.type", "h2");
+    props.put("h2.path", "./h2");
+
+    props.put(AZKABAN_USE_MULTIPLE_EXECUTORS, "true");
```
kunkun-tang: This looks a bit strange. Haven't seen anywhere calling azkaban.use.multiple.executors?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @BeforeClass
+  public static void setUp() throws Exception {
+    tearDown();
+
+    String sqlScriptsDir = getSqlScriptsDir();
+    props.put(AzkabanDatabaseSetup.DATABASE_SQL_SCRIPT_DIR, sqlScriptsDir);
+
+    props.put("database.type", "h2");
+    props.put("h2.path", "./h2");
```
HappyRay: Why not use in memory h2 db for testing?

azkaban-web-server/build.gradle
```diff
@@ -66,9 +66,7 @@ dependencies {
...
-  testRuntime('commons-collections:commons-collections:3.2.2')
-  testCompile('org.hamcrest:hamcrest-all:1.3')
+  testRuntime "com.h2database:h2:1.4.193"
```
kunkun-tang: Just would like double check this library is in our artifactory, right?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
suvodeep-pyne: Yeah. these are all the hacks to just instantiate the web server. :|

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  private static String getSqlScriptsDir() throws IOException {
+    // Dummy because any resource file works.
+    final String dummyResourcePath = getUserManagerXmlFile();
+    Path resources = Paths.get(dummyResourcePath).getParent();
+    Path azkabanRoot = resources.getParent().getParent().getParent().getParent();
+
+    File sqlScriptDir = Paths.get(azkabanRoot.toString(), AZKABAN_DB_SQL_PATH).toFile();
+    return props.getString(AzkabanDatabaseSetup.DATABASE_SQL_SCRIPT_DIR, sqlScriptDir.getCanonicalPath());
```
HappyRay: Why do you need to care about what the entry is in the props? The test is in full control of what's in the props. Right?

azkaban-web-server/build.gradle
suvodeep-pyne: Yes. I am using the same h2 version as used elsewhere in the project. Haven't done any consolidation.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: Yes. I missed this. will add it.

azkaban-web-server/build.gradle
suvodeep-pyne: These are derived and duplicate dependencies. Removed it. updated description

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
suvodeep-pyne: tests can start from a clean state despite running into errors in previous execution

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
suvodeep-pyne: Honestly, I don't have much experience with h2. But I am not particularly concerned about performance in test code in a case where we don't have any SQL testing.

If we can make a default azkaban mode to pick up an in memory DB if not configured, that would be great.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
suvodeep-pyne: copy paste error. fixed it.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
HappyRay: Thanks

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
HappyRay: Isn't @AfterClass designed for that purpose?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
HappyRay: Yes, we should invest in making that easier since we are likely going to write more unit tests against DB.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
suvodeep-pyne: That's a valid argument. The reason I left it there is because during debugging or fixing code you may leave the state inconsistent. For example, if you stopped while running a test, you would still have files and directories present which will not be cleaned on the next run unless we explicitly clean it.

PR #1056, opened 2017-05-05
Refactored Solo server launch. Also added a unit test for Solo server

suvodeep-pyne:
Refactored Solo server to launch from an instance as opposed to launching from a static context. Added a unit test validating the same.


azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -339,28 +340,16 @@ public static void main(String[] args) throws Exception {
...
     /* Initialize Guice Injector */
-    SERVICE_PROVIDER.setInjector(Guice.createInjector(
-        new AzkabanCommonModule(props),
-        new AzkabanExecServerModule()
-    ));
+    final Injector injector = Guice.createInjector(new AzkabanCommonModule(props), new AzkabanExecServerModule());
+    SERVICE_PROVIDER.setInjector(injector);
 
-    launch(props);
+    launch(injector.getInstance(AzkabanExecutorServer.class));
```
jamiesjc: Just curious what is the intention/benefit of launching from server instance instead of props here?

suvodeep-pyne: Well, this is less bad than what it was. Let me explain.

Previously, solo server would call main methods of web and exec servers. I refactored that to call `launch` and created room for different injectors for solo, web and exec servers.

This goes one step ahead; such that we inject the Web server instance instead of properties. This is more `Guicy`. Since we are creating the service directly from Guice and sending it. Notice that we don't have to use the `SERVICE_PROVIDER` singleton. The lesser the calls to this, the better.

Future: As we integrate with Guice and refactor more code, we can get rid of the `SERVICE_PROVIDER` interface totally.

PR #1057, opened 2017-05-05
Storage Refactor. Changed filename to include hash of file

suvodeep-pyne:
Changes:
- refactored Storage API to throw IOException
- extracted hash computation code to StorageManager
- Changed ProjectLoader API to accept hash as parameter

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -33,7 +33,11 @@ import azkaban.trigger.JdbcTriggerImpl;
 import azkaban.trigger.TriggerLoader;
 import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
+import com.google.inject.Inject;
+import com.google.inject.Provides;
 import com.google.inject.Scopes;
+import com.google.inject.Singleton;
+import javax.sql.DataSource;
 import org.apache.commons.dbutils.QueryRunner;
 
 
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -26,9 +26,9 @@ import azkaban.executor.ExecutorLoader;
 import azkaban.executor.JdbcExecutorLoader;
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
+import azkaban.spi.AzkabanException;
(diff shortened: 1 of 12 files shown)


azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -36,34 +37,32 @@
 public class LocalStorage implements Storage {
   private static final Logger log = Logger.getLogger(LocalStorage.class);
 
-  final File baseDirectory;
+  final File rootDirectory;
 
-  public LocalStorage(File baseDirectory) {
-    this.baseDirectory = validateBaseDirectory(createIfDoesNotExist(baseDirectory));
+  @Inject
+  public LocalStorage(AzkabanCommonModuleConfig config) {
```
kunkun-tang: Is going to be a future convention for juice usage in Azkaban? Every class calls `AzkabanCommonModuleConfig ` ?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -76,10 +71,4 @@ protected void configure() {
       throw new StorageException(e);
     }
   }
-
-  @Inject
-  public @Provides
-  LocalStorage createLocalStorage(AzkabanCommonModuleConfig config) {
```
kunkun-tang: We don't need to bind `.class` when we move this out?

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
suvodeep-pyne: I think classes operating on a config is a standard convention. Just that our module is too big - as in common has too much stuff in it. But for the time being I think, it is easier to work with a common config container since it also contains props as well.

So my suggestion is to use this model if it applies - for example if you need a bunch of `props` values and need to extract information from them.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
suvodeep-pyne: No. Because I already have an `@inject` annotation in the class. Guice gets confused if there are no annotations and it has multiple constructors like we saw the other day.

Bindings are mandatory for interfaces where Guice doesn't know what implementation to use.

suvodeep-pyne:
Azkaban computes a md5 hash of each project zip file and stores it in the db for validation. This is just a refactor to expose the API in a way that is easy for the new StorageManager to consume it.

PR #1062, opened 2017-05-07
Fix bugs in resubmitting flow and fetching flow logs after flow finis…

jamiesjc:
After removing runningFlows cache from web server, running flow info will be fetched from DB directly by ExecutorManager. This includes the flow updateTime which is updated by FlowRunnerManager when the flow finishes. Currently the UpdaterThread in ExecutorManager will send update request to executor to get updated flow info. UpdateTime now is synced between web server and executor since they all update and get from DB instead of cache, the previous logic of comparing updateTime should be changed.
TODO: updaterThread in executorManager should be removed in the future to simplify the logic. handleAjaxUpdateRequest() should be deprecated as well.


azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
```diff
@@ -268,7 +268,10 @@ private void handleAjaxUpdateRequest(HttpServletRequest req,
         continue;
       }
 
-      if (flow.getUpdateTime() > updateTime) {
+      // After removing runningFlows cache from web server, flow info will be fetched from DB directly,
+      // updateTime will be synced between web server and executor. So below if condition will always be true.
+      // TODO: will remove UpdaterThread in executorManager and deprecate handleAjaxUpdateRequest()
```
HappyRay: Please follow the todo naming convention.

PR #1063, opened 2017-05-07
Exclude junit from the json-simple library

HappyRay:
This library brings in an older version of junit which causes
Intellij to pick this version somehow and caused compilation error
when building using Intellij.


suvodeep-pyne:
Thanks a lot for fixing this!

PR #1070, opened 2017-05-09
Upgrading Hive to 1.1.0

suvodeep-pyne:


gradle.properties
```diff
@@ -26,4 +26,4 @@ ide.recursive=true
 # Versions
 #---------------------------------
 hadoopVersion=2.6.1
-hiveVersion=0.13.1
+hiveVersion=1.1.0
```
HappyRay: How is this variable used?

PR #1071, opened 2017-05-09
Changing style template to Google IntelliJ style

suvodeep-pyne:
Switching to [Google style](https://github.com/google/styleguide). No code has been reformatted yet.

Instructions for setup are available here:
https://github.com/HPI-Information-Systems/Metanome/wiki/Installing-the-google-styleguide-settings-in-intellij-and-eclipse


HappyRay:
Looks Gobblin project uses a modified google java style as well.  http://gobblin.readthedocs.io/en/latest/developer-guide/CodingStyle/

HappyRay:
It would be great if you can add instruction on how to apply this style. 
Thanks!

suvodeep-pyne:
I added a link to the instructions. However, I think this is something very easily Google-able.

HappyRay:
Thanks @suvodeep-pyne 

Yes, I agree other users can search for this information. However if you would like your effect to have the maximum effect, it is better to make it as easy as possible for your users.

After following your link, it is still not clear to me how to apply the styles. So I searched again and found this instruction

https://www.jetbrains.com/help/idea/2017.1/configuring-code-style.html
This seems to be a more direct instruction from the vendor itself. 

And I have to turn off Linkedin CheckSytle in the other settings. It appears that the coding style setting doesn't replace the functionalities provided by checkstyle. e.g. According to https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names " Identifiers use only ASCII letters and digits, and, in a small number of cases noted below, underscores. Thus each valid identifier name is matched by the regular expression \w+ ."
When I added a _ in front of a member variable for testing, the ide didn't give me a warning.

Do you know if there is more that need to be done to enforce the coding styles?

Also, the https://www.jetbrains.com/help/idea/2017.1/configuring-code-style.html shows that intellij also supports  
http://editorconfig.org/
This seems attractive since it makes applying consistent styles across editors easier. 
However, it doesn't appear that the options are as extensive. 
It's worth looking into though.
cc @chengren311

suvodeep-pyne:
I think it makes sense to add a developer section to our readme regarding our recommended setup with the code style, etc. This would probably be more accessible than a PR.

I have seen the jetbrains link that you posted. Even though it has a ton of information about codestyles, I didn't find the piece of information that we needed for setting up a code style file. While the link I posted just mentions that piece.

I don't know about disabling LinkedIn checkstyle. I didn't have any such issues. I just switched my style file and it worked.

Thanks for sharing editorconfig. I wasn't aware about this. However, if you say that this isn't extensive enough, then I would stick with supporting IntelliJ as our preferred dev env.

HappyRay:
Following the intellij help page, I was able to import the xml file and copy it to the project. I haven't verified it is working though.

HappyRay:
Yes, updating the README is better than putting this information in a PR. When I suggested adding instructions, I meant adding instructions in a place where you think will be best to serve the purpose. :)

PR #1073, opened 2017-05-09
Gucify ExecutorManager

chengren311:


azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.log4j.Logger;
+
+
+public class AlerterHolder {
+  private final Map<String, Alerter> alerters;
+
+  private static Logger logger = Logger.getLogger(AlerterHolder.class);
```
suvodeep-pyne: I think migrating to a different logger in an incremental way is difficult. More so because people just copy the logger line from a different class. This needs to be done either all at once or one module at a time. and that dependency needs to be removed.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -0,0 +1,193 @@
...
+
+  public AlerterHolder() {
```
suvodeep-pyne: I think we can use the same below constructor at all places. The usages of this are in test code. We should either mock it or just fill it with empty props to get it to work.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -111,7 +112,7 @@
   private long lastThreadCheckTime = -1;
   private String updaterStage = "not started";
 
-  private Map<String, Alerter> alerters;
+  private AlerterHolder alerterHolder;
```
suvodeep-pyne: final ?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -159,7 +160,7 @@
   private final ExecutorManager executorManager;
   private final ScheduleManager scheduleManager;
   private final TriggerManager triggerManager;
-  private final Map<String, Alerter> alerters;
+  private final AlerterHolder alerters;
```
suvodeep-pyne: Let's remove alerters from web server as well.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
chengren311: updated

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
chengren311: updated

suvodeep-pyne:
Looks great! Thanks!

shuzhang1989:
nice work @chengren311 !

PR #1074, opened 2017-05-10
Add project name to bodies of Azkaban emails

reallocf:
Adding the project name to the body of Azkaban emails when flows succeed or fail. Will help users better identify succeeding/failing flows. Done at the request of a user.


azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
```diff
@@ -66,12 +66,12 @@ public boolean createFirstErrorMessage(ExecutableFlow flow,
     if (emailList != null && !emailList.isEmpty()) {
       message.addAllToAddress(emailList);
       message.setMimeType("text/html");
-      message.setSubject("Flow '" + flow.getFlowId() + "' has encountered a failure on "
-          + azkabanName);
+      message.setSubject("Flow '" + flow.getFlowId()
```
chengren311: i think we can have those strings as constant variable in this class and use string.format

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
```diff
@@ -130,12 +130,12 @@ public boolean createErrorEmail(ExecutableFlow flow, EmailMessage message,
...
-      message.setSubject("Flow '" + flow.getFlowId() + "' has failed on "
-          + azkabanName);
+      message.setSubject("Flow '" + flow.getFlowId()
+          + "' has failed on " + azkabanName);
 
       message.println("<h2 style=\"color:#FF0000\"> Execution '" + execId
-          + "' of flow '" + flow.getFlowId() + "' has failed on " + azkabanName
-          + "</h2>");
+          + "' of flow '" + flow.getFlowId() + "' of project '"
+          + flow.getProjectName() + "' has failed on " + azkabanName + "</h2>");
```
chengren311: looks "has encountered a failure on" and "has failed on" are telling the same thing, but having inconsistent expression might cause friction for user to search those emails in their inbox. 
should we have the same string template as a member variable for them?

azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java
```diff
@@ -95,7 +95,8 @@ public void createErrorEmail() throws Exception {
     executableFlow.setStatus(Status.FAILED);
     assertTrue(mailCreator.createErrorEmail(
         executableFlow, message, azkabanName, scheme, clientHostname, clientPortNumber));
-    assertEquals("Flow 'mail-creator-test' has failed on unit-tests", message.getSubject());
+    assertEquals("Flow 'mail-creator-test' has failed on unit-tests",
```
chengren311: why split into two lines?

reallocf: You're right - it shouldn't be two lines. I had added a change here and then removed it, but forgot to remove the newline. Thanks!

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
reallocf: I just followed the best practices I already saw in the file. Do you think it would be better to refactor the template using constant variables / string.format? What's the best practice now?

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
reallocf: createFirstErrorMessage and createErrorEmail are (from what I can tell) two distinct things. I've never actually seen the first one (createFirstErrorMessage) so for all I know that copy is never seen by users. But we can definitely change it to one/the other and then template it if you think that'd be a good idea!

PR #1076, opened 2017-05-10
Fix race condition caused by UpdaterThread in ExecutorManager

jamiesjc:


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1333,13 +1340,9 @@ public void run() {
...
-                    ExecutableFlow flow = e.getExecutableFlow();
-                    logger.error(e);
-
-                    if (flow != null) {
-                      logger.error("Finalizing flow " + flow.getExecutionId());
-                      finalizeFlows.add(flow);
-                    }
+                    // Currently just ignore the update error. Will remove UpdaterThread and
+                    // add separate clean up code to handle errors.
+                    logger.error(e + ". Update execution failed. Ignored.");
```
chengren311: logger.error("Update execution failed. Ignored. "， e);

PR #1077, opened 2017-05-10
[DO NOT MERGE] Migration Tool for transferring project file data from Database to Local Storage

suvodeep-pyne:
This PR implements a tool to transfer existing projects in Azkaban DB to Local Storage. A similar approach can be taken to transfer projects to HDFS. Or one can simply copy the local storage directory to HDFS to make it work.

Usage:
```
# migration command
java -jar migrateSql2Local.jar migrate.properties
```

Example migrate.properties
```
# mysql
database.type=mysql
mysql.port=3306
mysql.password=
mysql.user=azkaban
mysql.numconnections=100
mysql.host=
mysql.database=

#local storage
azkaban.storage.local.basedir=/data/azkaban/
```

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
@@ -16,6 +16,18 @@
 
 package azkaban.project;
 
+import azkaban.database.AbstractJdbcLoader;
+import azkaban.flow.Flow;
+import azkaban.project.ProjectLogEvent.EventType;
+import azkaban.user.Permission;
+import azkaban.user.User;
+import azkaban.utils.GZIPUtils;
+import azkaban.utils.JSONUtils;
+import azkaban.utils.Md5Hasher;
+import azkaban.utils.Pair;
+import azkaban.utils.Props;
+import azkaban.utils.PropsUtils;
+import azkaban.utils.Triple;
 import com.google.common.io.Files;
 import com.google.inject.Inject;
 import java.io.BufferedInputStream;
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -16,6 +16,18 @@
 
 package azkaban.project;
 
+import azkaban.database.AbstractJdbcLoader;
```
kunkun-tang: Oops. Actually I'm refactoring this class. Eventually I will use another class (`JdbcProjectImpl`) to replace this class. Of course, you can still use this. But I would probably change `AzkabanCommonModule` to allow `JdbcProjectImpl` to be the only one implmentation of `Projectloader`.

suvodeep-pyne: That's cool. I don't intend to merge this PR. I just wanted to have this patch available in case it becomes useful later on.

ameyamk: Can we have another module like tools or something where people can check in such tools? Not merging does not sound like the most optimal approach.
(Something to think about)

suvodeep-pyne: I agree. But this code involves some hacks with the JDBC interfaces which I don't intend to push to master. That's the reason why I published this as [my fork](https://github.com/suvodeep-pyne/azkaban/tree/migrate-tool).

PR #1040, opened 2017-04-28
new Jdbc Trigger class

kunkun-tang:
This patch is make a replacement to TriggerJDBCLoader class, in order to
call out new Azkaban-db interface/code. Bring juice to it. Fix the broken JDBC Test class
, and add in-memory database to test. Fix getLstInsertId method inside
akzban-db module.

Also, some other db related refactor is commited in order to allow test
work.

PR #1079, opened 2017-05-10
refactor JDBC Project class to leverage new azkaban-db

kunkun-tang:
This change is aimed at re-constructing JDBC project loader class to
allow db retry feature to be leveraged. In order to make this happen,
there are a few changes made in this patch:
* remove some never-use abstract method in ProjectLoader interface.
* guicify JdbcProjectImpl and make it as the only implementation to
Projectloader.
* Seprate All Project related JDBC handler classes to a ourside static class.
* Adding getDataSource method in azkaban-db interface 

Evaluated the basic uses of ProjectLoader.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
@@ -52,41 +52,41 @@ class JdbcProjectHandlerSet {
         "SELECT id, name, active, modified_time, create_time, version, last_modified_by, description, enc_type, settings_blob FROM projects WHERE name=? AND active=true";
 
     @Override
-    public List<Project> handle(ResultSet rs) throws SQLException {
+    public List<Project> handle(final ResultSet rs) throws SQLException {
       if (!rs.next()) {
         return Collections.emptyList();
       }
 
-      ArrayList<Project> projects = new ArrayList<>();
+      final ArrayList<Project> projects = new ArrayList<>();
       do {
-        int id = rs.getInt(1);
-        String name = rs.getString(2);
-        boolean active = rs.getBoolean(3);
-        long modifiedTime = rs.getLong(4);
-        long createTime = rs.getLong(5);
-        int version = rs.getInt(6);
-        String lastModifiedBy = rs.getString(7);
-        String description = rs.getString(8);
-        int encodingType = rs.getInt(9);
-        byte[] data = rs.getBytes(10);
+        final int id = rs.getInt(1);
+        final String name = rs.getString(2);
+        final boolean active = rs.getBoolean(3);
+        final long modifiedTime = rs.getLong(4);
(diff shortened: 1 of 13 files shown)


kunkun-tang:
This PR is similar to #1040, which also tries leveraging newly constructed azkaban-db module.

kunkun-tang:
rebased the latest code from master, and re-evaluated. Ready for review.

azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
```diff
@@ -0,0 +1,322 @@
...
+                resourceId);
+
+        handlers.add(handler);
+      } while (rs.next());
+
+      return handlers;
+    }
+  }
+
+  public static class IntHander implements ResultSetHandler<Integer> {
```
reallocf: IntHander or IntHandler?

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
```diff
@@ -785,7 +784,7 @@ public void removePermission(final Project project, final String name, final boo
 
   @Override
   public List<Triple<String, Boolean, Permission>> getProjectPermissions(
```
jamiesjc: This method doesn't seem to be used. Could be removed later.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -200,7 +200,7 @@ void updateProjectProperty(Project project, Props props)
   Props fetchProjectProperty(int projectId, int projectVer, String propsName)
       throws ProjectManagerException;
 
-  List<Triple<String, Boolean, Permission>> getProjectPermissions(int projectId)
```
jamiesjc: Same comment as above.

azkaban-common/src/test/java/azkaban/project/MockProjectLoader.java
```diff
@@ -231,7 +231,7 @@ public Props fetchProjectProperty(final int projectId, final int projectVer,
 
   @Override
   public List<Triple<String, Boolean, Permission>> getProjectPermissions(
```
jamiesjc: Same comment as above.

azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
kunkun-tang: Good Catch! Changing!

azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java
kunkun-tang: I searched and recalled that this method has to be used in my test. So I didn't remove it.

PR #1081, opened 2017-05-12
refactor trigger class

chengren311:
1. use builder pattern instead of constructor in trigger
2. add toString()


azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,120 +121,101 @@ public void setContext(Map<String, Object> context) {
...
+
+    private final Condition triggerCondition;
+    private final List<TriggerAction> actions;
+    private Condition expireCondition;
+    private List<TriggerAction> expireActions = new ArrayList<>();
+
+    private Map<String, Object> info = new HashMap<String, Object>();
+    private Map<String, Object> context = new HashMap<String, Object>();
+
+    public TriggerBuilder(String submitUser,
```
kunkun-tang: Why can't we do setSubmitUser also?

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,120 +121,101 @@ public void setContext(Map<String, Object> context) {
...
+      this.info = info;
+      return this;
+    }
+
+    public TriggerBuilder setContext(Map<String, Object> context) {
+      this.context = context;
+      return this;
+    }
+
+    public Trigger build() {
```
kunkun-tang: Is it a good practice to have some requreNonNull check here to guarantee building process doesn't miss some important parameters?

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: since submituser is a required parameter.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,120 +121,101 @@ public void setContext(Map<String, Object> context) {
...
+    private int triggerId = -1;
+    private long lastModifyTime;
+    private long submitTime;
+    private final String submitUser;
+    private final String source;
+    private final TriggerStatus status = TriggerStatus.READY;
+
+    private final Condition triggerCondition;
+    private final List<TriggerAction> actions;
+    private Condition expireCondition;
```
jamiesjc: Should expireCondition be final?

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -27,6 +27,7 @@
 
 import azkaban.utils.JSONUtils;
 
+
 public class Trigger {
```
jamiesjc: Do we need to set the members to final like submitTime, submitUser ...? They are only used by getter method.

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
```diff
@@ -283,9 +283,12 @@ public void doAction() throws Exception {
             actions.add(killAct);
           }
         }
-        Trigger slaTrigger =
-            new Trigger("azkaban_sla", "azkaban", triggerCond, expireCond,
-                actions);
+        Trigger slaTrigger = new Trigger.TriggerBuilder("azkaban_sla",
```
suvodeep-pyne: would be good to turn these into constants.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,120 +121,101 @@ public void setContext(Map<String, Object> context) {
...
+      return this;
+    }
+
+    public TriggerBuilder setContext(Map<String, Object> context) {
+      this.context = context;
+      return this;
+    }
+
+    public Trigger build() {
+      return new Trigger(triggerId,
```
suvodeep-pyne: Now that we are initializing everything in one shot, should we also make all trigger fields final?

suvodeep-pyne: nit: this is more like an fyi.

For some reason, this class has a private constructor which throws an exception! We can probably get rid of that since we already have another constructor. Also, one of the other ways of taking accepting parameters is to use a builder as a parameter itself and initialize all fields from it.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -120,120 +121,101 @@ public void setContext(Map<String, Object> context) {
...
   }
 
-  public Trigger(int triggerId, long lastModifyTime, long submitTime,
+  private Trigger(int triggerId, long lastModifyTime, long submitTime,
       String submitUser, String source, Condition triggerCondition,
       Condition expireCondition, List<TriggerAction> actions,
       List<TriggerAction> expireActions, Map<String, Object> info,
       Map<String, Object> context) {
-    this.triggerId = triggerId;
     this.lastModifyTime = lastModifyTime;
```
suvodeep-pyne: nit: probably good to do validations here. like `requireNonNull` checks, etc

suvodeep-pyne:
Looks good. minor comments.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: updated

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: updated

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
chengren311: already updated in the other PR

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: updated

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: updated

>

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
chengren311: discussed offline, i removed that private constructor and made fields without associated setters final.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -402,10 +393,19 @@ public static Trigger fromJson(Object obj) throws Exception {
         action.setContext(context);
       }
 
-      trigger =
```
jamiesjc: Is there a reason why the submitUser is explicitly set to "azkaban"?

chengren311: that's my mistake, updated in https://github.com/azkaban/azkaban/pull/1250

PR #1082, opened 2017-05-12
refactor TriggerManager

chengren311:
1. remove unused code
2. improve logging


azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -120,6 +120,7 @@ protected ActionTypeLoader getActionLoader() {
   }
 
   public void insertTrigger(Trigger t) throws TriggerManagerException {
+    logger.info("Inserting trigger " + t + " in TriggerManager");
```
shuzhang1989: is Trigger::toString() implemented? what will be the output?

chengren311: yep, it's added in https://github.com/azkaban/azkaban/pull/1081

PR #1093, opened 2017-05-17
Remove recently finished flows cache

jamiesjc:
In order to scale out web server, we need to move the existing cache from web server to database so that web server can be stateless. Currently, recentlyFinishedFlows cache keeps the flows which finished within 10 mins. This info is shown on Executing Flows - Recently Finished web page.
Changes made:
1. Removed the recentlyFinishedFlows cache.
2. Added new SQL query to fetch recentlyFinishedFlows which end within 10 mins and status is finished (SUCCEEDED/KILLED/FAILED). Execution_flows DB table is already indexed by end_time, so the query performance is good.
3. Added unit test. Also tested on solo server and testing cluster.


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -107,6 +104,8 @@
   // 12 weeks
   private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
       * 24 * 60 * 60 * 1000L;
+  // 10 mins recently finished threshold.
+  private long recentlyFinishedLifetimeMs = 600000;
```
chengren311: final/static and upper case

chengren311: maybe consider using duration ?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1364,6 +1382,52 @@ public String handle(ResultSet rs) throws SQLException {
...
+      do {
+        int id = rs.getInt(1);
+        int encodingType = rs.getInt(2);
+        byte[] data = rs.getBytes(3);
+
+        if (data != null) {
+          EncodingType encType = EncodingType.fromInteger(encodingType);
+          Object flowObj;
+          try {
+            if (encType == EncodingType.GZIP) {
```
chengren311: this piece of code is duplicate all over the place, can we make it a util? e.x it's also used in https://github.com/azkaban/azkaban/pull/1079

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -59,6 +59,7 @@
   private static final String user = "azkaban";
   private static final String password = "azkaban";
   private static final int numConnections = 10;
+  private static final long recentlyFinishedLifetimeMs = 5000;
```
chengren311: upper case. also using duration might be helpful

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -388,4 +388,10 @@ public Executor fetchExecutorByExecutionId(int execId) throws ExecutorManagerExc
   public void unassignExecutor(int executionId) throws ExecutorManagerException {
     executionExecutorMapping.remove(executionId);
   }
+
+  @Override
+  public List<ExecutableFlow> fetchRecentlyFinishedFlows(long lifeTimeMs)
+      throws ExecutorManagerException {
+    return null;
```
chengren311: i think we should return an empty list instead of null here.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
suvodeep-pyne: Thanks @chengren311 ! 
@jamiesjc please take a look at the `Duration` class for storing time constants.

jamiesjc: Fixed. Thanks!

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
suvodeep-pyne: nit:
Will the test take 5s to finish? Will it be possible to reduce the time? 
We can mock this out if required.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
suvodeep-pyne: I think we should probably delete this class :@

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
jamiesjc: Yes, tried reducing to 1s and it also works.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
jamiesjc: I agree. It's better to use Mockito to mock the executor loader. Currently this class is used in many test cases, but we can gradually retire it in the future.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
jamiesjc: Yes, I also noticed that. Plan to do it in a separate pull request.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -107,6 +105,8 @@
   // 12 weeks
   private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
       * 24 * 60 * 60 * 1000L;
+  // 10 mins recently finished threshold.
+  private static final Duration RECENTLY_FINISHED_LIFETIME = Duration.ofSeconds(600);
```
chengren311: why not ofMins(10) ?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1364,6 +1382,52 @@ public String handle(ResultSet rs) throws SQLException {
     }
   }
 
+  private static class FetchRecentlyFinishedFlows implements
```
kunkun-tang: Would FetchRecentlyFinishedFlowsHandler be a better name? though I know the handler classes in this file don't have this convention. What do you think?

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +875,30 @@ public void testFetchActiveFlowByExecId() throws Exception {
     Assert.assertEquals(flow1.getVersion(), execFlow1.getVersion());
   }
 
+  @Test
```
kunkun-tang: Good test strategy!

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -388,4 +388,10 @@ public Executor fetchExecutorByExecutionId(int execId) throws ExecutorManagerExc
   public void unassignExecutor(int executionId) throws ExecutorManagerException {
     executionExecutorMapping.remove(executionId);
   }
+
+  @Override
+  public List<ExecutableFlow> fetchRecentlyFinishedFlows(long lifeTimeMs)
+      throws ExecutorManagerException {
+    return new ArrayList<>();
```
kunkun-tang: Is Java smart enough to know the generic type here? Never knew this.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1364,6 +1382,52 @@ public String handle(ResultSet rs) throws SQLException {
...
+    ResultSetHandler<List<ExecutableFlow>> {
+    // Execution_flows table is already indexed by end_time
```
kunkun-tang: A bit confused what this comment want to tell readers? Should it be "Execution_flows table should index end_time"

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +874,29 @@ public void testFetchActiveFlowByExecId() throws Exception {
     Assert.assertEquals(flow1.getVersion(), execFlow1.getVersion());
```
HappyRay: See http://stackoverflow.com/questions/1371940/unit-testing-is-it-a-good-practice-to-have-assertions-in-setup-methods

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +874,29 @@ public void testFetchActiveFlowByExecId() throws Exception {
...
+    flow1.setStatus(Status.SUCCEEDED);
+    flow1.setEndTime(System.currentTimeMillis());
+    loader.updateExecutableFlow(flow1);
+    List<ExecutableFlow> flows = loader.fetchRecentlyFinishedFlows(recentlyFinishedLifetimeMs);
+    Assert.assertEquals(1, flows.size());
+    Assert.assertEquals(flow1.getExecutionId(), flows.get(0).getExecutionId());
+    Assert.assertEquals(flow1.getProjectName(), flows.get(0).getProjectName());
+    Assert.assertEquals(flow1.getFlowId(), flows.get(0).getFlowId());
+    Assert.assertEquals(flow1.getVersion(), flows.get(0).getVersion());
+    Thread.currentThread().sleep(recentlyFinishedLifetimeMs);
```
HappyRay: Why does the thread need to sleep here? 
It's important to keep a unit test run as fast as possible. in a few ms is the norm.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +874,29 @@ public void testFetchActiveFlowByExecId() throws Exception {
     Assert.assertEquals(flow1.getVersion(), execFlow1.getVersion());
   }
 
+  @Test
+  public void testFetchRecentlyFinishedFlows() throws Exception {
+    if (!isTestSetup()) {
```
HappyRay: My understanding is that isTestSetup will return false in Travis effectively disabling this test. Am I right?

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +874,29 @@ public void testFetchActiveFlowByExecId() throws Exception {
...
+    flow1.setEndTime(System.currentTimeMillis());
+    loader.updateExecutableFlow(flow1);
+    List<ExecutableFlow> flows = loader.fetchRecentlyFinishedFlows(recentlyFinishedLifetimeMs);
+    Assert.assertEquals(1, flows.size());
+    Assert.assertEquals(flow1.getExecutionId(), flows.get(0).getExecutionId());
+    Assert.assertEquals(flow1.getProjectName(), flows.get(0).getProjectName());
+    Assert.assertEquals(flow1.getFlowId(), flows.get(0).getFlowId());
+    Assert.assertEquals(flow1.getVersion(), flows.get(0).getVersion());
+    Thread.currentThread().sleep(recentlyFinishedLifetimeMs);
+    flows = loader.fetchRecentlyFinishedFlows(recentlyFinishedLifetimeMs);
```
HappyRay: Have you considered setting up an in-memory DB state and just test the fetch method? 
@kunkun-tang has introduced some tests using in-memory DB.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
jamiesjc: This assert is not inside setupDB(). It's in a separate test case.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
jamiesjc: Yes, in order to run all these tests in JdbcExecutorLoaderTest class locally, we need to modify the database, user and password to match our local mysql setup. In Travis, all the tests will be disabled because DB is not set up. Once we change it to in-memory database later, this JdbcExecutorLoaderTest class will be refactored and the tests won't have this setup limitation.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
jamiesjc: Yes, from the return type it is able to figure out. If you explicitly put the type ArrayList<ExecutableFlow>, intellij will suggest you to remove it.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
jamiesjc: I added this comment in case readers have doubt about the query performance.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
jamiesjc: Thanks! : )

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
HappyRay: Got it. I was confused by the diff display.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: +1

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +875,30 @@ public void testFetchActiveFlowByExecId() throws Exception {
...
+    flow1.setEndTime(System.currentTimeMillis());
+    loader.updateExecutableFlow(flow1);
+    List<ExecutableFlow> flows = loader.fetchRecentlyFinishedFlows(
+        RECENTLY_FINISHED_LIFETIME.toMillis());
+    Assert.assertEquals(1, flows.size());
+    Assert.assertEquals(flow1.getExecutionId(), flows.get(0).getExecutionId());
+    Assert.assertEquals(flow1.getProjectName(), flows.get(0).getProjectName());
+    Assert.assertEquals(flow1.getFlowId(), flows.get(0).getFlowId());
+    Assert.assertEquals(flow1.getVersion(), flows.get(0).getVersion());
+    Thread.currentThread().sleep(RECENTLY_FINISHED_LIFETIME.toMillis());
```
HappyRay: We would like unit tests to run as fast as possible.
It seems that in this test you can control the flow end time. If you set an end time e.g. 20 minutes in the past, the test doesn't need to sleep, does it?

Also I don't see a positive test case, i.e. there is a flow that finishes within 10 minutes.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: Changed. Thanks!

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
jamiesjc: Yes, I've tried using DateTimeUtils to set end time in the past, and it works. Thanks!
In line 890 within the same test, it is the positive scenario where it successfully fetches the recently finished flow. I will split the two tests for clarity.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -32,6 +32,9 @@ void uploadExecutableFlow(ExecutableFlow flow)
   ExecutableFlow fetchExecutableFlow(int execId)
       throws ExecutorManagerException;
 
+  List<ExecutableFlow> fetchRecentlyFinishedFlows(long lifeTimeMs)
```
HappyRay: It's not obvious what the lifeTimeMs means. Add a javadoc?

Does the name like "maxAge" more closely reflect the intent?

How about using the Duration instead here and convert to long when needed? At this level of abstraction, users of this API shouldn't share how this value is represented in the DB.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -641,7 +641,14 @@ private void getRunningFlowsIdsHelper(List<Integer> allIds,
 
   @Override
   public List<ExecutableFlow> getRecentlyFinishedFlows() {
-    return new ArrayList<ExecutableFlow>(recentlyFinished.values());
+    List<ExecutableFlow> flows = new ArrayList<>();
+    try {
+      flows = executorLoader.fetchRecentlyFinishedFlows(
+          RECENTLY_FINISHED_LIFETIME.toMillis());
+    } catch(ExecutorManagerException e) {
+      logger.error(e);
```
HappyRay: This will call
void error(Object message) {
which will not log the exception details such as call stack.
Always use this method to log an exception. 
void error(Object message, Throwable t) {

HappyRay: Should the server return a server error (5xx) instead?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1364,6 +1382,52 @@ public String handle(ResultSet rs) throws SQLException {
...
+    private static String FETCH_RECENTLY_FINISHED_FLOW =
```
HappyRay: What do you think about limiting the returned result to the most recent N flows instead regardless of how long ago they finish?

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
jamiesjc: Changed. Thanks!

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: Changed to error(Object message, Throwable t). I think it's ok to just keep the existing behavior to return an empty list if it fails to fetch recently finished flows. What is the intention of returning a server error?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
jamiesjc: That's also an option. But I think it's a little hard to decide the exact number limit for all clusters. On some clusters we might want to display more? Also I guess users have already got used to look at the flows finished in last 10 mins?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: Returning a server error will help with the monitoring system for example.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
HappyRay: I don't see the change yet. Did you submit it?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -107,6 +105,8 @@
   // 12 weeks
   private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
       * 24 * 60 * 60 * 1000L;
+  // 10 mins recently finished threshold.
```
HappyRay: Is this comment still necessary? The code seems to be clear enough.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -641,7 +641,14 @@ private void getRunningFlowsIdsHelper(List<Integer> allIds,
 
   @Override
   public List<ExecutableFlow> getRecentlyFinishedFlows() {
-    return new ArrayList<ExecutableFlow>(recentlyFinished.values());
+    List<ExecutableFlow> flows = new ArrayList<>();
+    try {
+      flows = executorLoader.fetchRecentlyFinishedFlows(
+          RECENTLY_FINISHED_LIFETIME);
+    } catch(ExecutorManagerException e) {
+      logger.error("Failed to fetch recently finished flows.", e);
```
HappyRay: I would suggest adding a todo here to fix the error handling.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1364,6 +1386,52 @@ public String handle(ResultSet rs) throws SQLException {
     }
   }
 
+  private static class FetchRecentlyFinishedFlows implements
```
HappyRay: Another strategy is to break up a larger pull request into multiple smaller ones.  Create an issue to tie them together. This can speed up the review process. In case there is a need to rollback, you have an option to roll back a portion of the changes.

e.g. adding a new method like this can go in first independently. Since it is not used yet, it is safe.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -873,6 +877,50 @@ public void testFetchActiveFlowByExecId() throws Exception {
...
+    }
+
+    ExecutorLoader loader = createLoader();
+    ExecutableFlow flow1 = TestUtils.createExecutableFlow("exectest1", "exec1");
+    loader.uploadExecutableFlow(flow1);
+    flow1.setStatus(Status.SUCCEEDED);
+    flow1.setEndTime(DateTimeUtils.currentTimeMillis());
+    loader.updateExecutableFlow(flow1);
+    //Mock flow finished time to be 2 min ago.
+    DateTimeUtils.setCurrentMillisOffset(-FLOW_FINISHED_TIME.toMillis());
```
HappyRay: Is it necessary to modify the current time? Isn't create a new time enough?

We should use the java8 java.time api instead of jodatime in the new code. 
https://stackoverflow.com/questions/29750221/is-joda-time-deprecated-with-java-8-date-and-time-api-java-time

PR #1101, opened 2017-05-19
HDFS Security bug

suvodeep-pyne:
Any Hadoop object used in the code needs to be wrapped in a doAs call. Otherwise, it uses the system default Kerberos ticket which is not refreshed by HdfsAuth module. This can result in authorization failure when the ticket expires.

Tested manually by ensuring that there is no existing valid Kerberos ticket on the system. The storage APIs seem to work.


HappyRay:
According to this article
https://community.hortonworks.com/articles/56702/a-secure-hdfs-client-example.html
you shouldn't need to use doas since you are not impersonating. What did I miss?

HappyRay:
What happens if you run the tests while making sure that system default Kerberos ticket cache is empty?

suvodeep-pyne:
@HappyRay It works? or probably I didn't understand your question.

HappyRay:
@suvodeep-pyne do you mean the code before this change worked at least for a while, while making sure that system default Kerberos ticket cache is empty?

HappyRay:
According to this http://stackoverflow.com/questions/34616676/should-i-call-ugi-checktgtandreloginfromkeytab-before-every-action-on-hadoop, 

"
Can I rely on the various Hadoop clients they call checkTGTAndReloginFromKeytab whenever it's needed?
You can rely on this if your application's usage pattern is to call the Hadoop clients, which in turn utilize Hadoop's RPC framework. You cannot rely on this if your application's usage pattern only calls the Hadoop REST APIs.
"
The renewal should be taken care of for you.

HappyRay:
This MR method doesn't use doas
org.apache.hadoop.mapreduce.JobSubmissionFiles#getStagingDir

```
    UserGroupInformation ugi = UserGroupInformation.getLoginUser();
    realUser = ugi.getShortUserName();
    currentUser = UserGroupInformation.getCurrentUser().getShortUserName();
    if (fs.exists(stagingArea)) {
      FileStatus fsStatus = fs.getFileStatus(stagingArea);
      String owner = fsStatus.getOwner();
```

azkaban-common/src/main/java/azkaban/storage/HdfsAuth.java
```diff
@@ -53,25 +58,47 @@ public HdfsAuth(Props props, Configuration conf) {
...
         throw new AzkabanException(String.format(
-            "Error: Unable to authorize to Hadoop. Principal: %s Keytab: %s", keytabPrincipal, keytabPath));
+            "Error: Unable to authorize to Hadoop. Principal: %s Keytab: %s", keytabPrincipal,
+            keytabPath));
       }
     }
   }
 
   private void login(String keytabPrincipal, String keytabPath) throws IOException {
     if (loggedInUser == null) {
```
HappyRay: The loggedInUser is not volatile and the method is not synchronized, the check is not guaranteed to work.

suvodeep-pyne:
> @suvodeep-pyne do you mean the code before this change worked at least for a while, while making sure that system default Kerberos ticket cache is empty?

By 'while' I think you are talking about till Kerberos ticket expires? No I haven't but I think we can rely on standard Hadoop APIs. There is similar code pattern inside Azkaban itself: https://github.com/suvodeep-pyne/azkaban/blob/d854fb04b71cfea0bc68dc03086e7128f4b4e7b4/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java#L219-L231

suvodeep-pyne:
> According to this http://stackoverflow.com/questions/34616676/should-i-call-ugi-checktgtandreloginfromkeytab-before-every-action-on-hadoop,

What are we trying to optimize here? The relogin code is a no op as is mentioned on the same page unless there is an actual login.

From the same link:
> It's fine to call UserGroupInformation#checkTGTAndReloginFromKeytab right before every action that needs to be authenticated. If the ticket is not close to expiration, then the method will be a no-op. If you're suspicious that your Kerberos infrastructure is sluggish, and you don't want client operations to pay the latency cost of re-login, then that would be a reason to do it in a separate background thread. Just be sure to stay a little bit ahead of the ticket's actual expiration time. You might borrow the logic inside UserGroupInformation for determining if a ticket is "close" to expiration. In practice, I've never personally seen the latency of re-login be problematic.

Personally, I think it is better that we manage the login and use Hadoop only for HDFS APIs. I think this coding pattern to `authorize` first and then perform action is pretty standard. I don't want to assume that for RPC calls, it works and when we plan to use REST we start having auth problems again.

azkaban-common/src/main/java/azkaban/storage/HdfsAuth.java
suvodeep-pyne: Thanks for pointing this out. Let me fix it.

suvodeep-pyne:
@HappyRay updated PR.

HappyRay:
The existing AZ code you pointed it "There is similar code pattern inside Azkaban itself: https://github.com/suvodeep-pyne/azkaban/blob/d854fb04b71cfea0bc68dc03086e7128f4b4e7b4/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java#L219-L231" seems to be the impersonation use case which requires doas. 

What you proposed may turn out to the right thing to do. There are still conflicting signals, however. I don't feel I understand the problem and usage pattern well enough yet. I am concerned about hidden problems if we don't completely understand the behavior. 

I haven't found any Hadoop documentation that says that doas is required for non-impersonation use case. 

I would expect a simpler pattern to work:  login from the keytab once when the service starts up and make hdfs calls as normal. 

Can you try the example in 
https://community.hortonworks.com/articles/56702/a-secure-hdfs-client-example.html
?
```
Configuration conf = new Configuration();
conf.set("fs.defaultFS", "hdfs://one.hdp:8020");
conf.set("hadoop.security.authentication", "kerberos");
 
UserGroupInformation.setConfiguration(conf);
UserGroupInformation.loginUserFromKeytab("hdfs-user@MYCORP.NET", 
   "/home/hdfs-user/hdfs-user.keytab");
 
FileSystem fs = FileSystem.get(conf);
FileStatus[] fsStatus = fs.listStatus(new Path("/"));
for(int i = 0; i < fsStatus.length; i++){
  System.out.println(fsStatus[i].getPath().toString());
}
```

suvodeep-pyne:
> Can you try the example in
https://community.hortonworks.com/articles/56702/a-secure-hdfs-client-example.html
?

I think we have a very similar if not identical implementation which doesn't work today. You can probably check the existing code? The difference that I see is the initial launch args where they are introducing JVM level JAAS config. Example:

```
java -Djava.security.auth.login.config=/home/hdfs-user/jaas.conf \
-Djava.security.krb5.conf=/etc/krb5.conf \
-Djavax.security.auth.useSubjectCredsOnly=false \
-cp "./hdfs-sample-1.0-SNAPSHOT.jar:/usr/hdp/current/hadoop-client/lib/*:/usr/hdp/current/hadoop-hdfs-client/*:/usr/hdp/current/hadoop-client/*"  \
```
Let me explain why I think this is not the best design. This sets global kerberos credentials which is not required since the Hadoop access is required by a very specific part of the app. If and when we make HDFS storage a plugin, it would require system properties to be set by the parent app on launch which I think is inconvenient.

The doAs implementation contains the credential use in a way which is convenient. And as @Victsm (Min) pointed out, even yarn uses a similar pattern even though they only have the 'yarn' user and there don't have an impersonating use case.

> What you proposed may turn out to the right thing to do. There are still conflicting signals, however. I don't feel I understand the problem and usage pattern well enough yet. I am concerned about hidden problems if we don't completely understand the behavior.

I think we already have a working + tested solution. These are precisely the concerns I raised when we were choosing Hadoop. There are some subtleties when using secure HDFS. And this has clearly introduced development + testing + setup delays. Can we leverage the Hadoop team to get a closure on this?

HappyRay:
I tried a modified example referenced above
see "Via Keytab" section. Note that this example doesn't require passing in a JAAS config file. It's very similar to your implementation before this change.

The code worked without problem as I expected for two days. 
I am curious why changing to doas would affect the behavior in your test.

Test code:
```
package ray.hadooptest;

import java.io.IOException;
import java.time.Duration;
import java.time.LocalDateTime;
import java.time.format.DateTimeFormatter;
import org.apache.hadoop.conf.Configuration;
import org.apache.hadoop.fs.FileStatus;
import org.apache.hadoop.fs.FileSystem;
import org.apache.hadoop.fs.Path;
import org.apache.hadoop.security.UserGroupInformation;


public class SimpleHDFSClient {
  public static void main(String[] args) throws IOException, InterruptedException {
    SimpleHDFSClient client = new SimpleHDFSClient();
    client.listRootFolder();
  }

  public void listRootFolder() throws IOException, InterruptedException {
    Configuration conf = new Configuration();

    UserGroupInformation.setConfiguration(conf);
    UserGroupInformation.loginUserFromKeytab("...", "...");

    FileSystem fs = FileSystem.get(conf);
    while (true) {
      printCurrentTime();

      FileStatus[] fsStatus = fs.listStatus(new Path("/"));
      for (int i = 0; i < fsStatus.length; i++) {
        System.out.println(fsStatus[i].getPath().toString());
      }
      Thread.sleep(Duration.ofMinutes(30).toMillis());
    }
  }

  private void printCurrentTime() {
    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss");
    LocalDateTime now = LocalDateTime.now();
    System.out.println(dtf.format(now)); //2016/11/16 12:08:43
  }

```

HappyRay:
Discussed with @suvodeep-pyne off line.

I am ok with merging this change for now to enable testing in test environment while investigating further why the earlier code didn't work.

suvodeep-pyne:
@HappyRay 
You were correct. There is a much simpler alternate solution. Found a bug in the existing code. Please check #1222 for the new simpler fix.

Issue #721, opened 2016-09-07
[FeaturDevelopment] Stop/End date for Scheduled Jobs


inoviavenkat:
Hi,

Currently, for an azkaban-job, the user can specify a start-date and the
period-of-interval
for recurrent executions.

In addition to these, it would be a very helpful to specify an stop-date
too, past which the job would
no more be triggered/executed. The current job definition does not support
an stop-date and I am
trying to implement it (picture attached)

However, I am badly stuck with the implementation due to lack of
documentation in the code and below
are my specific questions. Any ideas to move further from here or
connecting to me somebody who
has knowledge of these areas of code, would be much appreciated.

1) The job details entered on the web-UI (attached page) are persisted as a
blob
in the column 'data' in the database table named 'triggers'. Which java
class
reads this data further and translates them to equivalent
threads/processes. Is this
blob further used by the ExecutorServer to create some other entries in a
different table?

2) Which existing class/method is the best choice to modify with a tiny
check to ensure that
current execution time is not past the stop-date.
I am thinking it should be
*azkaban.trigger.builtin.BasicTimeChecker.eval() *
Any better ideas?

3) I did notice that
_azkaban.scheduler.TriggerBasedScheduleLoader.scheduleToTrigger(Schedule)_
is adding translating the schedule into a Trigger, which consists of a
TriggerCondition
and an ExpiryCondition, both of them composed of
_azkaban.trigger.builtin.BasicTimeChecker._
However, they are almost identical, differing only with the id
BasicTimeChecker_1 vs BasicTimeChecker_2.
Where are these two Condition actually called to perform the check?

Thanks in advance.
Venkatesh

+46707188865
![screen shot 2016-09-07 at 17 36 13](https://cloud.githubusercontent.com/assets/17123141/18319919/0c2f784e-7528-11e6-8378-a55bb4838eec.png)

fsi206914:
@inoviavenkat Thanks a lot for your feature request. Below  are the answers to your questions.

1). The blob is used by webserver only. Jobs in executor will only be triggered by webserver calling.
No other tables. 
2). 3). They are called in triggermanager class. There is a thread to check these conditions periodically. 

BTW, you might contribute to adding this feature after #719 is merged. Flexible Scheduling will be released soon.

ameyamk:
@inoviavenkat still working on this feature? we will be happy to help

inoviavenkat:
@ameyamk
[...]

PR #1103, opened 2017-05-20
EndDate Expiring Schedule Introduction

kunkun-tang:
Today, when we specify a flow schedule in Azkaban, the flow will run infinitely without termination as per the schedule. As #721 proposes, should schedule have an end date, the flow will have more flexibilities to be managed. In this PR, I implement the end date feature to allow users to specify the expiration date given a schedule. UI has not been changed yet in this patch (will not be implemented in near future), but people will be able to call API to leverage the new feature.

The API command sample:
`curl -k -d ajax=scheduleCronFlow -d projectName=wtwt -d flow=azkaban -d endSchedTime=14XXXXXXXXXX --data-urlencode cronExpression="0 * * ? * 6" -b "azkaban.browser.session.id=XXXXXXXXXXXXXX" http://localhost:8081/schedule`

Similar to what @inoviavenkat implemented, I add an EndTimeChecker, as part of Expire Condition checker. Also, Added one parameter `endSchedTime` to schedule API. If the users don't specify this parameter, Azkaban would use the default value `-1`, and the schedule will be still running infinitely. Besides, I add Unit tests as much as possible to guarantee that this change will not break out. 

Furthermore, I made comprehensive manual tests and am glad to share:
* We have flow1, flow2, flow3, flow4, corresponding schedule1, schedule2, schedule3, schedule4. Every schedule runs flow every minute. The 4 flows keep undispatched at this stage.
* Ignoring this PR, I started a web server running on 3.24.0. Triggering schedule1. it is working fine.
* Apply the patch, started web server having end date feature. 
  * Triggered regular schedule2 without specifying endDate. Both sche1 and sche2 were running fine.
  * Triggered schedule3 by specifying endtime (5 minutes later). schedule3 was initially running. was able to observe schedule3 expire then.
  * Triggered schedule4 by specifying endtime (15 minutes later). Shutdown Webserver before schedule4 expires.

* restarted web server with end date implementation. Schedule3 never run. Both schedule1 and schedule2 and schdule4 were running well. Schedule4 expires at correct time.

* I also try rolling back web server to 3.24.0 (without this PR). Schedule 2 , 3 and 4 were not loaded due to lacking EndTimeChecker, but schedule1 was running good. Even though rolled back web server can not identify those latest schedules, the exceptions don't break and web server keeps working.

Though this PR looks big, but more than 80% code is related to testing and refactoring

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
@@ -31,6 +31,7 @@ import azkaban.trigger.TriggerManager;
 import azkaban.trigger.TriggerManagerAdapter;
 import azkaban.trigger.TriggerManagerException;
 import azkaban.trigger.builtin.BasicTimeChecker;
+import azkaban.trigger.builtin.EndTimeChecker;
 import azkaban.trigger.builtin.ExecuteFlowAction;
 
 public class TriggerBasedScheduleLoader implements ScheduleLoader {
@@ -45,7 +46,7 @@ public class TriggerBasedScheduleLoader implements ScheduleLoader {
   private long lastUpdateTime = -1;
 
   public TriggerBasedScheduleLoader(TriggerManager triggerManager,
-      String triggerSource) {
+                                    String triggerSource) {
     this.triggerManager = triggerManager;
     this.triggerSource = triggerSource;
   }
@@ -99,16 +100,11 @@ public class TriggerBasedScheduleLoader implements ScheduleLoader {
 
   // if failed to trigger, auto expire?
   private Condition createExpireCondition(Schedule s) {
-    Map<String, ConditionChecker> checkers =
-        new HashMap<String, ConditionChecker>();
-    ConditionChecker checker =
-        new BasicTimeChecker("BasicTimeChecker_2", s.getFirstSchedTime(),
-            s.getTimezone(), s.isRecurring(), s.skipPastOccurrences(),
-            s.getPeriod(), s.getCronExpression());
+    Map<String, ConditionChecker> checkers = new HashMap<>();
+    ConditionChecker checker = new EndTimeChecker("EndTimeChecker_1", s.getEndSchedTime());
     checkers.put(checker.getId(), checker);
     String expr = checker.getId() + ".eval()";
-    Condition cond = new Condition(checkers, expr);
-    return cond;
+    return new Condition(checkers, expr);
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


HappyRay:
"Schedule 2 , 3 and 4 were not loaded due to lacking EndTimeChecker" 
This makes rolling back the change impossible without user impact. Can you think of ways to mitigate? e.g. allow these flows to run as if the end date is not specified at least until we feel confident enough about the change.

ameyamk:
End Date should be optional. We should have config/ policy from where we can enforce this. But as a default endDate should be optional. 
+1 to Ray.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -132,31 +128,24 @@ public void updateSchedule(Schedule s) throws ScheduleManagerException {
     }
   }
 
-  // TODO may need to add logic to filter out skip runs
+  // Todo kunkun-tang: Never used method, should be completely removed later.
   @Override
   public synchronized List<Schedule> loadSchedules()
```
chengren311: why not remove it now?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -167,15 +156,22 @@ private Schedule triggerToSchedule(Trigger t) throws ScheduleManagerException {
...
-              act.getExecutionOptions(), act.getSlaOptions(), ck.getCronExpression());
-      return s;
+    if (basicTimeChecker != null && act != null) {
+      return new Schedule(t.getTriggerId(),
+          act.getProjectId(),
+          act.getProjectName(),
+          act.getFlowName(),
+          t.getStatus().toString(),
+          basicTimeChecker.getFirstCheckTime(),
+          endTimeChecker == null?-1 : endTimeChecker.getEndCheckTime(),
```
chengren311: space

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -300,7 +312,7 @@ private void onTriggerTrigger(Trigger t) throws TriggerManagerException {
       }
     }
 
-    private void onTriggerExpire(Trigger t) throws TriggerManagerException {
+    private void onTriggerPause(Trigger t) throws TriggerManagerException {
```
chengren311: when will the trigger be expired if you change it to pause?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -759,6 +767,14 @@ private void ajaxScheduleCronFlow(HttpServletRequest req,
       ret.put("error", e.getMessage());
     }
 
+    long endSchedTime = getLongParam(req, "endSchedTime", -1L);
+    try {
+      // Todo kunkun-tang: Need to verify if passed end time is valid.
```
chengren311: ?

kunkun-tang:
Closing this due to roll back issue. Reopen this PR at #1110

PR #1105, opened 2017-05-22
Fix killed job status update and FlowRunnerTest

juhoautio:
Main fixes:

1. Make the `FlowRunnerTest` pass again.
    - However the test is so slow that I'm leaving the whole test class with `@Ignore`.

----

1. Allow creating an instance of `AzkabanExecutorServer` for testing without starting jetty server etc.

----

1. Log job logs in `MockExecutorLoader` instead of discarding them, so that it's possible to debug failures in tests.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
+++ b/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
@@ -34,7 +34,7 @@ import org.apache.log4j.Logger;
 public class MockExecutorLoader implements ExecutorLoader {
 
   private static final Logger logger = Logger
-          .getLogger(MockExecutorLoader.class);
+      .getLogger(MockExecutorLoader.class);
   HashMap<Integer, Integer> executionExecutorMapping =
       new HashMap<Integer, Integer>();
   HashMap<Integer, ExecutableFlow> flows =
--- a/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
+++ b/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
@@ -17,6 +17,7 @@
 package azkaban.executor;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
@@ -27,9 +28,13 @@ import azkaban.executor.ExecutorLogEvent.EventType;
 import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import org.apache.commons.io.FileUtils;
+import org.apache.log4j.Logger;
 
 public class MockExecutorLoader implements ExecutorLoader {
 
+  private static final Logger logger = Logger
+          .getLogger(MockExecutorLoader.class);
   HashMap<Integer, Integer> executionExecutorMapping =
       new HashMap<Integer, Integer>();
   HashMap<Integer, ExecutableFlow> flows =
@@ -97,7 +102,14 @@ public class MockExecutorLoader implements ExecutorLoader {
   @Override
   public void uploadLogFile(int execId, String name, int attempt, File... files)
       throws ExecutorManagerException {
-
+    for (File file : files) {
+      try {
+        String logs = FileUtils.readFileToString(file, "UTF-8");
+        logger.info("Uploaded log for [" + name + "]:[" + execId + "]:\n" + logs);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
   }
 
   @Override
--- a/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
+++ b/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
@@ -54,7 +54,8 @@ public class MockExecutorLoader implements ExecutorLoader {
   @Override
   public void uploadExecutableFlow(ExecutableFlow flow)
       throws ExecutorManagerException {
-    flows.put(flow.getExecutionId(), flow);
+    ExecutableFlow exFlow = ExecutableFlow.createExecutableFlowFromObject(flow.toObject());
+    flows.put(flow.getExecutionId(), exFlow);
     flowUpdateCount++;
   }
 
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
@@ -76,13 +76,13 @@ public class AzkabanExecServerModule extends AbstractModule {
     for (Connector connector : server.getConnectors()) {
       connector.setStatsOn(isStatsOn);
       logger.info(String.format(
-              "Jetty connector name: %s, default header buffer size: %d",
-              connector.getName(), connector.getHeaderBufferSize()));
+          "Jetty connector name: %s, default header buffer size: %d",
+          connector.getName(), connector.getHeaderBufferSize()));
       connector.setHeaderBufferSize(props.getInt("jetty.headerBufferSize",
-              DEFAULT_HEADER_BUFFER_SIZE));
+          DEFAULT_HEADER_BUFFER_SIZE));
       logger.info(String.format(
-              "Jetty connector name: %s, (if) new header buffer size: %d",
-              connector.getName(), connector.getHeaderBufferSize()));
+          "Jetty connector name: %s, (if) new header buffer size: %d",
+          connector.getName(), connector.getHeaderBufferSize()));
     }
 
     return server;
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
@@ -19,8 +19,19 @@ package azkaban.execapp;
 
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.JdbcExecutorLoader;
+import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
 import com.google.inject.Scopes;
+import org.apache.log4j.Logger;
+import org.mortbay.jetty.Connector;
+import org.mortbay.jetty.Server;
+import org.mortbay.jetty.servlet.Context;
+import org.mortbay.jetty.servlet.ServletHolder;
+import org.mortbay.thread.QueuedThreadPool;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
 
 
 /**
@@ -29,12 +40,65 @@ import com.google.inject.Scopes;
  * structuring of Guice components.
  */
 public class AzkabanExecServerModule extends AbstractModule {
+
+  private static final int DEFAULT_THREAD_NUMBER = 50;
+  private static final int DEFAULT_HEADER_BUFFER_SIZE = 4096;
+  private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
+
+  private static final Logger logger = Logger.getLogger(AzkabanExecServerModule.class);
+
   @Override
   protected void configure() {
     bind(ExecutorLoader.class).to(JdbcExecutorLoader.class).in(Scopes.SINGLETON);
     bind(AzkabanExecutorServer.class).in(Scopes.SINGLETON);
     bind(TriggerManager.class).in(Scopes.SINGLETON);
     bind(FlowRunnerManager.class).in(Scopes.SINGLETON);
+  }
 
+  @Provides
+  @Named("ExecServer")
+  @Singleton
+  private Server createJettyServer(Props props) {
+    int maxThreads = props.getInt("executor.maxThreads", DEFAULT_THREAD_NUMBER);
+
+    /*
+     * Default to a port number 0 (zero)
+     * The Jetty server automatically finds an unused port when the port number is set to zero
+     * TODO: This is using a highly outdated version of jetty [year 2010]. needs to be updated.
+     */
+    Server server = new Server(props.getInt("executor.port", 0));
+    QueuedThreadPool httpThreadPool = new QueuedThreadPool(maxThreads);
+    server.setThreadPool(httpThreadPool);
+
+    boolean isStatsOn = props.getBoolean("executor.connector.stats", true);
+    logger.info("Setting up connector with stats on: " + isStatsOn);
+
+    for (Connector connector : server.getConnectors()) {
+      connector.setStatsOn(isStatsOn);
+      logger.info(String.format(
+              "Jetty connector name: %s, default header buffer size: %d",
+              connector.getName(), connector.getHeaderBufferSize()));
+      connector.setHeaderBufferSize(props.getInt("jetty.headerBufferSize",
+              DEFAULT_HEADER_BUFFER_SIZE));
+      logger.info(String.format(
+              "Jetty connector name: %s, (if) new header buffer size: %d",
+              connector.getName(), connector.getHeaderBufferSize()));
+    }
+
+    return server;
+  }
+
+  @Provides
+  @Named("root")
+  @Singleton
+  private Context createRootContext(Server server) {
+    Context root = new Context(server, "/", Context.SESSIONS);
+    root.setMaxFormContentSize(MAX_FORM_CONTENT_SIZE);
+
+    root.addServlet(new ServletHolder(new ExecutorServlet()), "/executor");
+    root.addServlet(new ServletHolder(new JMXHttpServlet()), "/jmx");
+    root.addServlet(new ServletHolder(new StatsServlet()), "/stats");
+    root.addServlet(new ServletHolder(new ServerStatisticsServlet()), "/serverStatistics");
+    return root;
   }
 }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -94,6 +94,19 @@ public class AzkabanExecutorServer {
   private final ArrayList<ObjectName> registeredMBeans = new ArrayList<ObjectName>();
   private MBeanServer mbeanServer;
 
+  public static boolean testing = false;
+
+  public AzkabanExecutorServer(ExecutorLoader executionLoader, FlowRunnerManager runnerManager, Props props,
+                               Server server) {
+    if (!testing) {
+      throw new IllegalStateException("This constructor is only for testing");
+    }
+    this.executionLoader = executionLoader;
+    this.runnerManager = runnerManager;
+    this.props = props;
+    this.server = server;
+  }
+
   @Inject
   public AzkabanExecutorServer(Props props,
       ExecutorLoader executionLoader,
@@ -319,6 +332,10 @@ public class AzkabanExecutorServer {
    */
   public static AzkabanExecutorServer getApp() {
     return app;
+  }
+
+  public static void setApp(AzkabanExecutorServer app) {
+    AzkabanExecutorServer.app = app;
   }
 
   /**
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -39,6 +39,7 @@ import azkaban.server.AzkabanServer;
 import azkaban.utils.Props;
 import azkaban.utils.StdOutErrRedirect;
 import azkaban.utils.Utils;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.inject.Guice;
 import com.google.inject.Inject;
@@ -55,6 +56,7 @@ import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.TimeZone;
+import javax.inject.Named;
 import javax.management.MBeanInfo;
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
@@ -64,8 +66,6 @@ import org.joda.time.DateTimeZone;
 import org.mortbay.jetty.Connector;
 import org.mortbay.jetty.Server;
 import org.mortbay.jetty.servlet.Context;
-import org.mortbay.jetty.servlet.ServletHolder;
-import org.mortbay.thread.QueuedThreadPool;
 
 import static azkaban.Constants.*;
 import static azkaban.ServiceProvider.*;
@@ -75,14 +75,11 @@ import static java.util.Objects.*;
 public class AzkabanExecutorServer {
   private static final String CUSTOM_JMX_ATTRIBUTE_PROCESSOR_PROPERTY = "jmx.attribute.processor.class";
   private static final Logger logger = Logger.getLogger(AzkabanExecutorServer.class);
-  private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
 
   public static final String JOBTYPE_PLUGIN_DIR = "azkaban.jobtype.plugin.dir";
   public static final String METRIC_INTERVAL = "executor.metric.milisecinterval.";
-  public static final int DEFAULT_HEADER_BUFFER_SIZE = 4096;
 
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
-  private static final int DEFAULT_THREAD_NUMBER = 50;
 
   private static AzkabanExecutorServer app;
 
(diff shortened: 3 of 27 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -94,6 +94,19 @@
   private final ArrayList<ObjectName> registeredMBeans = new ArrayList<ObjectName>();
   private MBeanServer mbeanServer;
 
+  public static boolean testing = false;
```
HappyRay: Consider wrapping it in a setter method and annotate it with @visablefortesting and add a comment?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -94,6 +94,19 @@
...
+
+  public AzkabanExecutorServer(ExecutorLoader executionLoader, FlowRunnerManager runnerManager, Props props,
```
HappyRay: I wonder if with the recent introduction of Guice. There is a better way to construct different dependencies in tests so that we can avoid introducing this new test only constructor.

@suvodeep-pyne

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -506,6 +506,15 @@ private void finalizeAttachmentFile() {
    */
   @Override
   public void run() {
+    try {
+      doRun();
+    } catch (Exception e) {
+      e.printStackTrace();
```
HappyRay: Log to a logger instead?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -696,8 +705,13 @@ private Status runJob() {
         logError("Job run failed, but will treat it like success.");
         logError(e.getMessage() + " cause: " + e.getCause(), e);
       } else {
-        finalStatus = changeStatus(Status.FAILED);
-        logError("Job run failed!", e);
+        if (node.getStatus() == Status.KILLED) {
```
HappyRay: With this change, does the code in run method azkaban.execapp.JobRunner#run still apply?

    if (isKilled()) {
      // even if it's killed, there is a chance that the job failed is marked as
      // failure,
      // So we set it to KILLED to make sure we know that we forced kill it
      // rather than
      // it being a legitimate failure.
    }

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -225,14 +234,8 @@ public void exec1FailedKillAll() throws Exception {
     Assert.assertTrue(runner.isKilled());
 
     Assert.assertTrue(
-        "Expected flow " + Status.FAILED + " instead " + exFlow.getStatus(),
-        exFlow.getStatus() == Status.FAILED);
-
-    try {
-        Thread.sleep(500);
```
HappyRay: Thanks for improving this slow tests.
Any suggestion on how to make the tests run faster so that we can run them on every commit?

HappyRay:
@chengren311 could you also take a look?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
chengren311: +1

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -395,6 +381,45 @@ private void testAttempts(ExecutableFlow flow, String name, int attempt) {
...
+      try {
+        Thread.sleep(10);
+      } catch (InterruptedException e) {
+        // ignored
+      }
+    }
+    Assert.fail("Jobs didn't start in 5 seconds");
+  }
+
+  private boolean checkJobsStarted(FlowRunner runner, String[] jobs) {
```
chengren311: what does this method check?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
juhoautio: Can do, but let's first explore that Guice option you mentioned below.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
juhoautio: Ok I hope so. The reason for the pattern I've chosen is because of trying to avoid merge conflicts, if these changes would only be done in private fork. I'm all in for doing it properly now that it's going towards the public repo.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: Yes, sorry..

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: I'll give it a try. First of this test should use some custom job type for tests, instead of type=java jobs that have a lot of overhead in launching.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: The method checks the given list of job names on the runner to see if their status is something that means that they have been at least started, if not even finished.

Do you mean that javadoc should be added?

Personally I feel that this method name should be clear enough, and having a glance at the code is probably better than my sorry attempt to explain it above :)

juhoautio:
Added commit:

Use logger instead of printStackTrace()
- Using the server logger instead of job log, because if `JobRunner.doRun()` throws an exception, it might be that the field `JobRunner.logger` is null.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: I think it's the best bet to keep it.

I'm only improving the "exit status of `runJob()` method", but it's also possible that the job is killed before `runJob()` is called at all. And I don't think status setting is synchronized well enough at the moment, I believe if the timing of `kill()` overlaps with `runJob()` in a certain way, it could still cause the flapping of `KILLED->FAILED->KILLED`.

Those possible mis-sync issues are really rare corner cases though. My change here was enough to keep the test passing without random failures. If someone wants to take it further, that could be done in another PR.

juhoautio:
Added commit:

Use Guice to create `AzkabanExecutorServer` for testing:
- Single constructor
- Inject jetty Server instance from module

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: For that I added this commit:
Convert `FlowRunnerTest` to use custom job type instead of java process
- Using `InteractiveTestJob` instead of `azkaban.executor.SleepJavaJob`.
- This makes the test run a bit faster already.

Then I tried to continue from there, and got the test to run quite quickly by removing all the extra waiting time. But the test now reveals some rare race condition that I wasn't able to hunt down yet. The flow runner is getting stuck occasionally.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -29,12 +40,65 @@
...
   @Override
   protected void configure() {
     bind(ExecutorLoader.class).to(JdbcExecutorLoader.class).in(Scopes.SINGLETON);
     bind(AzkabanExecutorServer.class).in(Scopes.SINGLETON);
     bind(TriggerManager.class).in(Scopes.SINGLETON);
     bind(FlowRunnerManager.class).in(Scopes.SINGLETON);
+  }
+
+  @Provides
+  @Named("ExecServer")
```
HappyRay: Why does it need to use @Named? 
Does plain bind work too?
I'd prefer minimizing the usage of Guice's advanced features to reduce the learning curve needed and make the code easier to understand.

juhoautio: Because there's another Server (unnamed) created by the AzkabanWebServerModule.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -29,12 +40,65 @@
...
+  @Singleton
+  private Server createJettyServer(Props props) {
```
HappyRay: Just like how AzkabanExecutorServer is constructed, can't the Server instance be constructed in the same way without using a Guice provider? 
I'd prefer keeping the module file as simple as possible.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -75,34 +75,38 @@
...
   private final ArrayList<ObjectName> registeredMBeans = new ArrayList<ObjectName>();
   private MBeanServer mbeanServer;
 
   @Inject
   public AzkabanExecutorServer(Props props,
       ExecutorLoader executionLoader,
-      FlowRunnerManager runnerManager) throws Exception {
+      FlowRunnerManager runnerManager,
+      @Named("ExecServer") Server server,
+      @Named("root") Context root) throws Exception {
```
HappyRay: Why does the exeServer need direct access to root Context now?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: I'm getting close to finding where the problem lies, but couldn't figure it out yet.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
juhoautio: You mean creating a "provider class" instead? See how it's done in AzkabanWebServerModule.java:
https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java#L32

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
juhoautio: To call this on startup, and to avoid circular dependency:

```java
    root.setAttribute(Constants.AZKABAN_SERVLET_CONTEXT_KEY, this);
```

I think the proper way to refactor this would be to not make `AzkabanExecutorServer` depend on `Server` either, but I'm not sure if that can be easily done.

azkaban-exec-server/src/test/java/azkaban/execapp/fake/FakeApp.java
```diff
@@ -0,0 +1,14 @@
+package azkaban.execapp.fake;
+
+import azkaban.execapp.AzkabanExecutorServer;
+import azkaban.execapp.jmx.JmxJobMBeanManager;
+import azkaban.utils.Props;
+
+public class FakeApp extends AzkabanExecutorServer {
```
HappyRay: Is a FakeApp class needed? Can test use a Mockito mock instance instead?

azkaban-exec-server/src/test/java/azkaban/execapp/fake/FakeServer.java
```diff
@@ -0,0 +1,14 @@
+package azkaban.execapp.fake;
+
+import org.mortbay.jetty.Connector;
+import org.mortbay.jetty.Server;
+import org.mortbay.jetty.bio.SocketConnector;
+
+public class FakeServer extends Server {
```
HappyRay: Can test use a Mockito mock instance instead?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -69,9 +77,11 @@ public void setUp() throws Exception {
     jobtypeManager =
         new JobTypeManager(null, null, this.getClass().getClassLoader());
     JobTypePluginSet pluginSet = jobtypeManager.getJobTypePluginSet();
-    pluginSet.addPluginClass("java", JavaJob.class);
+    pluginSet.setCommonPluginLoadProps(AllJobExecutorTests.setUpCommonProps());
     pluginSet.addPluginClass("test", InteractiveTestJob.class);
     fakeProjectLoader = new MockProjectLoader(workingDir);
+    Utils.initServiceProvider();
+    AzkabanExecutorServer.setApp(new FakeApp());
```
HappyRay: How does FlowRunner depend on AzkabanExecutorServer? In the future, we should refactor FlowRunner so that lower layer components shouldn't depend on higher level components.

azkaban-exec-server/src/test/java/azkaban/execapp/fake/FakeApp.java
juhoautio: I prefer Mockito, but it wasn't in the dependencies yet when I started working on this :) I'll make the switch.

HappyRay:
Thank you very much for trying this difficult task of fixing these tests!

My understanding of the goal of this change is that:
Make FlowRunnerTest pass consistently so that we can re-enable it.
The intermediate job state change is not necessary but it has no user-visible impact. However, it makes FlowRunnerTest flaky.  Am I right?

This change is becoming bigger and more complex. It would be better if you can think of ways to make this change incrementally.

IMHO, the current test running time is already longer than I would like. Let's try to keep the running time to the minimum.

juhoautio:
> My understanding of the goal of this change is that:
> Make FlowRunnerTest pass consistently so that we can re-enable it.

I guess not a must if it can't be made run quickly enough, but I'm pretty sure we'll get there. But it will be useful to have it working even if only executed manually.

> The intermediate job state change is not necessary but it has no user-visible impact. However, it makes FlowRunnerTest flaky. Am I right?

Yeah, that's my understanding. The change was needed to make the test work consistently. I guess the test could be stabilized with some synchronization instead of this change, but still it seems better to not have this back-and-forth setting of job status. There could be some place in the non-test code, too, where the job status is checked, and if it's temporarily FAILED instead of KILLED, something could go wrong.

> It would be better if you can think of ways to make this change incrementally.

I'll split this once ready otherwise.

> Let's try to keep the running time to the minimum.

Naturally.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: Huh, I finally found the reason to those random test failures and was able to fix it. The bug was in test code. Added a commit: "Fix concurrency bug in MockExecutorLoader".

juhoautio:
At the moment the test runs in ~5 seconds – and it's consistently passing. That's mostly because the kill/cancel test cases still use a sleep of 1 seconds. Should get rid of those and then we can remove `@Ignore`.

juhoautio:
Now the test is quite fast.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
HappyRay: Yes.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -16,10 +16,13 @@
 
 package azkaban.execapp;
 
+import static org.junit.Assert.assertNotNull;
```
HappyRay: Please clean up the imports. There are some unused imports.

HappyRay:
This is great. Thanks!

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
HappyRay: The refactoring of ExecSever class is useful.
Is it possible to use a mock class instead in these tests so that we can separate that refactoring change into a separate pull request and finish this one first?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
HappyRay: Exec server and tests can use different Guice modules if needed.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
HappyRay: You can use azkaban.webapp.AzkabanWebServer#AzkabanWebServer(org.mortbay.jetty.Server, azkaban.utils.Props) as an example.

HappyRay:
I set a breakpoint in 
  @Provides
  @Named("ExecServer")
  @Singleton
  private Server createJettyServer(Props props) {

And it wasn't hit when running the flow runner tests as I expected.
Could you point me to the code path that has this dependency?

juhoautio:
The injected Server dependency comes into play when you launch the server for realz. FlowRunnerTest passes new FakeServer() instead.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: Sorry, I need to find better IDEA settings to catch these.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: +1

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
juhoautio: Yep, this is not a problem in test scope but actual running of azkaban solo server.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
HappyRay: No problem. I have to manually tell IDEA to organize the imports.

HappyRay:
Could you make a separate pull request for the injected server dependency change so that we don't delay this pull request? The rest of the change LGTM.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -54,7 +54,8 @@
   @Override
   public void uploadExecutableFlow(ExecutableFlow flow)
       throws ExecutorManagerException {
-    flows.put(flow.getExecutionId(), flow);
+    ExecutableFlow exFlow = ExecutableFlow.createExecutableFlowFromObject(flow.toObject());
```
HappyRay: Thanks for the detailed description. However, I am still having trouble understanding this problem and fix. 
The new instance of the exFlow is just a clone of the existing one, isn't it?

HappyRay: In this test, would a Mockto based Mock ExecutorLoader which does nothing work and with less complications?

HappyRay: Is this an indication of a real production bug? 
@chengren311

HappyRay: "the current flow status, which is not synchronized with the job runner threads"
Why isn't it synchronized?

juhoautio: > Why isn't it synchronized?

I guess it just doesn't matter too much. Job status is only changed by the designated JobRunner thread (at least in the cases that FlowRunnerTest covers). The FlowRunner thread keeps running until it sees that all jobs have reached a finished status before it ends the flow (and it will eventually see the changes in job statuses even though ExecutableNode.status is not synchronized).

Open questions maybe worth investigating:
- Does the FlowRunner set status for jobs that entirely crashed (exception in job thread)?
- Is it a problem that the flow is sometimes saved with some "outdated" job statuses?

juhoautio: > Is this an indication of a real production bug?

No! In production code nothing tangles with the job statuses. It's all done in a designated thread  per job.

juhoautio: > would a Mockto based Mock ExecutorLoader which does nothing work and with less complications?

It would be much better. I'll try to switch to that..

juhoautio: > I am still having trouble understanding this problem and fix.

I guess you checked the [commit message](https://github.com/azkaban/azkaban/pull/1105/commits/ef2c53d82f9c432425a12925395843b75146a340)?

To add a bit of background to that, this test uses MockExecutorLoader instead of JDBCExecutorLoader. The uploadExecutableFlow() method saves current flow state into DB (I'm not 100% sure if there's any another reason, but at least via the DB it becomes accessible to the Azkaban web server). The mock version doesn't use a DB but local maps to keep the flows and jobs. I guess this is so that in test you could verify in the test "what would it save in DB".

Now on to the problem posed by this. The MockExecutorLoader created clones of the jobs, so (non-synchronized) job state updates didn't mess with the job nodes. But it didn't clone the flow object. And the original flow object has references to the original job nodes, too! When uploadExecutableFlow() was called, it also updated the job statuses on the _original_ ExecutableNodes based on the status values know to the FlowRunner thread at that time (see the stack trace in fix commit message to see why this causes job statuses being set). It's quite clear that to "mock the DB" this test shouldn't modify the original flows/ExecutableNodes.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: > How does FlowRunner depend on AzkabanExecutorServer?

Maybe the only need to have `AzkabanExecutorServer` initialized comes from this line in `FlowRunner`:

    logger.info("Assigned executor : " + AzkabanExecutorServer.getApp().getExecutorHostPort());

I guess this could be avoided quite easily at least.

juhoautio:
Split this into smaller chunks.

HappyRay:
"Open questions maybe worth investigating:

Does the FlowRunner set status for jobs that entirely crashed (exception in job thread)?
"
When the job process crashes, I assume the return code of the process will be non zero and the job will be marked failed.

in azkaban.jobExecutor.utils.process.AzkabanProcess#run
      int exitCode = -1;
      try {
        exitCode = process.waitFor();
      } catch (InterruptedException e) {
        logger.info("Process interrupted. Exit code is " + exitCode, e);
      }

Am I right?

"
Is it a problem that the flow is sometimes saved with some "outdated" job statuses?"

I have not able to understand the sequence this happens. The jobrunner thread seems to share the same node instance with FlowRunner.  azkaban.execapp.JobRunner#run appears to set the status on the node object directly.

If your theory is true, I would expect this to be a problem. I remember seeing incorrect job status. I am curious if what you found is the cause. Could you help me understand this problem?

juhoautio:
> AzkabanProcess..

But isn't that just in the scope of those job types? My question was about when JobRunner fails.

juhoautio:
>> Is it a problem that the flow is sometimes saved with some "outdated" job statuses?

> I have not able to understand the sequence this happens.

In short, I guess this shouldn't really be any problem.

JobRunner thread updates job statuses to ExecutableNode. The "problem" is, that the FlowRunner thread occasionally saves flow state into the DB, and that chunk of data also includes the job statuses. Because ExecutableNode.status is not synchronized (or volatile), it can happen that JobRunner has already set the status to something new, but FlowRunner writes to DB with an older job status after that.

But AFAIK this is still eventually consistent, because FlowRunner can't finalize the flow before it has seen that all job statuses are finished. Once it sees them, it writes the flow state once more and this time with the final statuses for all jobs.

Also AFAIK the flow state is a separate snapshot that includes also the job statuses. Apart from that I guess jobs are also saved to DB separately, and there the job status is never lagging.

> I remember seeing incorrect job status

I can think of a case where executor node is terminated and on restart flow is just marked FAILED. In that case it may be that jobs are not cleaned up in the same way, and they can be for example left in RUNNING state forever. But this is not about synchronization of ExecutableNode.status.

HappyRay:
I think you are right. If the current azkaban.execapp.JobRunner#run throws a runtime exception, the flowRunner won't notice. Filed an issue #1123. Thanks! 

===
Now I understand what you mean by the node statuses are not synchronized.  Since the node status is not synchronized or marked volatile, changes made to them by the job runner thread is not guaranteed to be seen by the flow runner thread. What would cause the change to be seen eventually?  Shouldn't the node status field be marked volatile?

And because this test wrote the job status known to the flowrunner thread back to the node and it exposed this issue?

Or in the future shall we use one thread to handle a flow to avoid synchronization problems?

PR #1108, opened 2017-05-22
fix AZKABAN_OPTS handling in azkaban-{web,executor}-start.sh

sjakthol:
If the value contains spaces, the comparison causes an error to be logged. If the reference to AZKABAN_OPTS is wrapped in quotes, spaces are handled properly.


HappyRay:
LGTM.
Maybe switch to [[]] too? 
see http://robertmuth.blogspot.com/2012/08/better-bash-scripting-in-15-minutes.html
Favor [[]] (double brackets) over [] 
 @li-afaris  what do you think? 

also http://tldp.org/LDP/abs/html/varsubn.html

sjakthol:
Switched from [ to [[ in both places.

li-afaris:
Looks good.  thanks for the patch @sjakthol.

PR #1110, opened 2017-05-23
EndDate Expiring Schedule Introduction

kunkun-tang:
Today, when we specify a flow schedule in Azkaban, the flow will run infinitely without termination as per the schedule. As #721 proposes, should schedule have an end date, the flow will have more flexibilities to be managed. In this PR, I implement the end date feature to allow users to specify the expiration date given a schedule. UI has not been changed yet in this patch (will not be implemented in near future), but people will be able to call API to leverage the new feature.

The API command sample:
`curl -k -d ajax=scheduleCronFlow -d projectName=wtwt -d flow=azkaban -d endSchedTime=14XXXXXXXXXX --data-urlencode cronExpression="0 * * ? * 6" -b "azkaban.browser.session.id=XXXXXXXXXXXXXX" http://localhost:8081/schedule`

I reuse BasicTimeChecker, as part of Expire Condition, in order to not break Old Azkaban flows. Also, Added one parameter `endSchedTime` to schedule API. If the users don't specify this parameter, Azkaban would use the default value for `01/01/2050`, and the schedule will be almost running infinitely. Besides, I add Unit tests as much as possible to guarantee that this change will not break out. 

Furthermore, I made comprehensive manual tests and am glad to share:
* We have flow1, flow2, flow3, flow4, corresponding schedule1, schedule2, schedule3, schedule4. Every schedule runs flow every minute. The 4 flows keep undispatched at this stage.
* Ignoring this PR, I started a web server running on 3.24.0. Triggering schedule1. it is working fine.
* Apply the patch, started web server having end date feature. 
  * Triggered regular schedule2 without specifying endDate. Both sche1 and sche2 were running fine.
  * Triggered schedule3 by specifying endtime (5 minutes later). schedule3 was initially running. was able to observe schedule3 expire then.
  * Triggered schedule4 by specifying endtime (15 minutes later). Shutdown Webserver before schedule4 expires.

* restarted web server with end date implementation. Schedule3 never run. Both schedule1 and schedule2 and schdule4 were running well. Schedule4 expires at correct time.

* I also try rolling back web server to 3.24.0 (without this PR).All un-expired schedules are loaded correctly, and running fine. Rolled back web server can not identify Expire conditions, and schedules which have not expired yet will ignore expire conditions.

* Given expired schedule, like schedule3, I reschedule it without/with endSchedule. Then it is able to keep running.

Though this PR looks big, but most code is related to testing and refactoring


azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -47,6 +47,9 @@
   public static final int DEFAULT_SSL_PORT_NUMBER = 8443;
   public static final int DEFAULT_JETTY_MAX_THREAD_COUNT = 20;
 
+  // One Schedule's default End Time: 01/01/2050, 00:00:00, UTC
+  public static final long DEFAULT_SCHEDULE_END_EPOCH_TIME = 2524608000000L;
```
chengren311: what's your rationale to pick up this date?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -97,18 +98,14 @@ private Condition createTriggerCondition(Schedule s) {
...
   private Condition createExpireCondition(Schedule s) {
-    Map<String, ConditionChecker> checkers =
-        new HashMap<String, ConditionChecker>();
-    ConditionChecker checker =
-        new BasicTimeChecker("BasicTimeChecker_2", s.getFirstSchedTime(),
-            s.getTimezone(), s.isRecurring(), s.skipPastOccurrences(),
-            s.getPeriod(), s.getCronExpression());
+    Map<String, ConditionChecker> checkers = new HashMap<>();
+    ConditionChecker checker = new BasicTimeChecker("EndTimeChecker_1", s.getFirstSchedTime(),
+        s.getTimezone(), s.getEndSchedTime(),false, false,
```
chengren311: space

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -132,33 +129,29 @@ public void updateSchedule(Schedule s) throws ScheduleManagerException {
...
-    Condition triggerCond = t.getTriggerCondition();
-    Map<String, ConditionChecker> checkers = triggerCond.getCheckers();
-    BasicTimeChecker ck = null;
+    Map<String, ConditionChecker> checkers = t.getTriggerCondition().getCheckers();
+    BasicTimeChecker triggerTimeChecker = null;
     for (ConditionChecker checker : checkers.values()) {
       if (checker.getType().equals(BasicTimeChecker.type)) {
-        ck = (BasicTimeChecker) checker;
-        break;
+        triggerTimeChecker = (BasicTimeChecker) checker;
```
chengren311: break here?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -167,15 +160,24 @@ private Schedule triggerToSchedule(Trigger t) throws ScheduleManagerException {
...
-      return s;
+    if (triggerTimeChecker != null && act != null) {
+      return new Schedule(t.getTriggerId(),
+          act.getProjectId(),
+          act.getProjectName(),
+          act.getFlowName(),
+          t.getStatus().toString(),
+          triggerTimeChecker.getFirstCheckTime(),
+          // End Schedule Time
+          endTimeChecker == null? Constants.DEFAULT_SCHEDULE_END_EPOCH_TIME: endTimeChecker.getNextCheckTime(),
```
chengren311: space?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -97,18 +98,14 @@ private Condition createTriggerCondition(Schedule s) {
...
-  // if failed to trigger, auto expire?
   private Condition createExpireCondition(Schedule s) {
-    Map<String, ConditionChecker> checkers =
-        new HashMap<String, ConditionChecker>();
-    ConditionChecker checker =
-        new BasicTimeChecker("BasicTimeChecker_2", s.getFirstSchedTime(),
-            s.getTimezone(), s.isRecurring(), s.skipPastOccurrences(),
-            s.getPeriod(), s.getCronExpression());
+    Map<String, ConditionChecker> checkers = new HashMap<>();
+    ConditionChecker checker = new BasicTimeChecker("EndTimeChecker_1", s.getFirstSchedTime(),
```
chengren311: the name EndTimeChecker_1 is not self explanatory, can we give a better name?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -132,33 +129,29 @@ public void updateSchedule(Schedule s) throws ScheduleManagerException {
...
-        ck = (BasicTimeChecker) checker;
-        break;
+        triggerTimeChecker = (BasicTimeChecker) checker;
       }
     }
+
+    Map<String, ConditionChecker> expireCheckers = t.getExpireCondition().getCheckers();
+    BasicTimeChecker endTimeChecker = null;
+    for (ConditionChecker checker : expireCheckers.values()) {
+      if (checker.getType().equals(BasicTimeChecker.type)) {
```
chengren311: this for loop looks similar to the one above. can we simplify them?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -167,15 +160,24 @@ private Schedule triggerToSchedule(Trigger t) throws ScheduleManagerException {
...
-              act.getExecutionOptions(), act.getSlaOptions(), ck.getCronExpression());
-      return s;
+    if (triggerTimeChecker != null && act != null) {
+      return new Schedule(t.getTriggerId(),
+          act.getProjectId(),
+          act.getProjectName(),
+          act.getFlowName(),
+          t.getStatus().toString(),
+          triggerTimeChecker.getFirstCheckTime(),
+          // End Schedule Time
```
chengren311: is this comment necessary?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -257,10 +255,17 @@ private void checkAllTriggers() throws TriggerManagerException {
...
           if (t.getStatus().equals(TriggerStatus.READY)) {
-            if (t.triggerConditionMet()) {
+
+            /**
+             * Prior to this change, expiration condition should never be called though
+             * we have some related code here. ExpireCondition used the same BasicTimeChecker
+             * as triggerCondition do. As a consequence, we need to figure out a way to distinguish
+             * the previous ExpireCondition and this commit's ExpireCondition.
+             */
+            if (t.getExpireCondition().getExpression().contains("EndTimeChecker") && t.expireConditionMet()) {
```
chengren311: using string containing to tell if it's endtimechecker is tricky. can we have a better way to handle this?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -300,7 +305,7 @@ private void onTriggerTrigger(Trigger t) throws TriggerManagerException {
       }
     }
 
-    private void onTriggerExpire(Trigger t) throws TriggerManagerException {
+    private void onTriggerPause(Trigger t) throws TriggerManagerException {
```
chengren311: why pausing instead of expire?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
kunkun-tang: Thanks for your review. Do you have any good idea what name could be better here? @chengren311

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
kunkun-tang: This comment was intended for previous commit. I would remove out of here.

azkaban-common/src/main/java/azkaban/Constants.java
kunkun-tang: I would believe 01/01/2050 could be regarded as never-expire schedule.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
kunkun-tang: I know this is hacky. This hack is intended for distinguishing existing schedules (created before the release of this PR). In other words, I don't need this hack if AZ doesn't have any schedule today.
Becuase existing triggers have the triggerCondition same as expireCondition, we need to tell the difference by this thack.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
kunkun-tang: There exists pros and cons regarding this change.

kunkun-tang: There exists pros and cons regarding this change.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
kunkun-tang: Added

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -132,33 +129,29 @@ public void updateSchedule(Schedule s) throws ScheduleManagerException {
     }
   }
 
-  // TODO may need to add logic to filter out skip runs
   @Override
   public synchronized List<Schedule> loadSchedules()
```
kunkun-tang: removed.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
kunkun-tang: Why do we need to break here? The logics here is confusing, and I added a todo comment to clarify.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
kunkun-tang: consolidated.

azkaban-common/src/main/java/azkaban/Constants.java
chengren311: maybe consider:   
public static final long NON_EXPIRE_EPOCH_TIME =  Long.MAX_VALUE;

kunkun-tang: @chengren311 I can't see too much difference on this change. Since this code is already in production, I would like we don't change this number.

PR #1112, opened 2017-05-25
Refactor AzkabanExecutorServer

juhoautio:
As requested by @HappyRay on https://github.com/azkaban/azkaban/pull/1105, this part as a separate PR.

Extract some parts into providers and inject them to `AzkabanExecutorServer`, instead of `AzkabanExecutorServer` creating them.

I hope you agree that https://github.com/azkaban/azkaban/pull/1113 is the better way to do this.

PR #1113, opened 2017-05-25
Refactor AzkabanExecutorServer

juhoautio:
As requested by @HappyRay on #1105, this part as a separate PR.

Extract some parts into a Guice module and inject them to `AzkabanExecutorServer`, instead of `AzkabanExecutorServer` creating them.

I think this is a cleaner way than implementing Provider classes as in  https://github.com/azkaban/azkaban/pull/1112. Especially note the difference in `AzkabanExecServerModule`.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -19,8 +19,8 @@ package azkaban.execapp;
 import static azkaban.Constants.AZKABAN_EXECUTOR_PORT_FILENAME;
 import static azkaban.Constants.ConfigurationKeys;
 import static azkaban.ServiceProvider.SERVICE_PROVIDER;
-import static azkaban.execapp.GuiceConstants.EXEC_JETTY_SERVER;
-import static azkaban.execapp.GuiceConstants.EXEC_ROOT_CONTEXT;
+import static azkaban.execapp.ExecJettyServerModule.EXEC_JETTY_SERVER;
+import static azkaban.execapp.ExecJettyServerModule.EXEC_ROOT_CONTEXT;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -17,7 +17,10 @@
 package azkaban.execapp;
 
 import static azkaban.Constants.AZKABAN_EXECUTOR_PORT_FILENAME;
+import static azkaban.Constants.ConfigurationKeys;
 import static azkaban.ServiceProvider.SERVICE_PROVIDER;
+import static azkaban.execapp.GuiceConstants.EXEC_JETTY_SERVER;
+import static azkaban.execapp.GuiceConstants.EXEC_ROOT_CONTEXT;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
@@ -61,6 +64,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.TimeZone;
+import javax.inject.Named;
 import javax.management.MBeanInfo;
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
@@ -70,19 +74,14 @@ import org.joda.time.DateTimeZone;
 import org.mortbay.jetty.Connector;
 import org.mortbay.jetty.Server;
 import org.mortbay.jetty.servlet.Context;
-import org.mortbay.jetty.servlet.ServletHolder;
-import org.mortbay.thread.QueuedThreadPool;
 
 public class AzkabanExecutorServer {
 
   public static final String JOBTYPE_PLUGIN_DIR = "azkaban.jobtype.plugin.dir";
   public static final String METRIC_INTERVAL = "executor.metric.milisecinterval.";
-  public static final int DEFAULT_HEADER_BUFFER_SIZE = 4096;
   private static final String CUSTOM_JMX_ATTRIBUTE_PROCESSOR_PROPERTY = "jmx.attribute.processor.class";
   private static final Logger logger = Logger.getLogger(AzkabanExecutorServer.class);
-  private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
-  private static final int DEFAULT_THREAD_NUMBER = 50;
 
   private static AzkabanExecutorServer app;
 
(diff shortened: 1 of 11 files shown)


HappyRay:
I realize now that solo server has both exeserver and webserver in the same process. They need different Jetty server instances.

Is there any immediate benefit of having Guice create Jetty server instances?

I'd like @suvodeep-pyne to weight in. He is currently on vacation.

juhoautio:
No hurry with this one from me. But I think this is definitely a nice improvement. Allows passing mocks to the constructor in a unit test etc.

HappyRay:
For tests that need an executorServer instance, I would expect them to use a mock.
Having the ability to mock the JettyServer can be useful when testing the executorServer methods itself.

juhoautio:
Yes. Exactly.

juhoautio:
Rebased & conflicts resolved.

juhoautio:
How about this?

juhoautio:
I wonder if @suvodeep-pyne is available to see this yet?

chengren311:
@juhoautio Suvodeep is on vacation. Currently there're some other high priority items team is working on, but I can take a look when I get a chance, most likely this weekend.

juhoautio:
@suvodeep-pyne do you think this would be the way to go? I could resolve conflicts then..

juhoautio:
@suvodeep-pyne what do you say? I was originally asked to create this PR by some of the repo owners, and personally I would like to finish the job, because otherwise I would've done all of this in vain.

suvodeep-pyne:
@juhoautio, somehow I missed this totally. Let me take a look.

azkaban-exec-server/src/main/java/azkaban/execapp/JettyServerModule.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.inject.Named;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
+import org.mortbay.jetty.Connector;
+import org.mortbay.jetty.Server;
+import org.mortbay.jetty.servlet.Context;
+import org.mortbay.jetty.servlet.ServletHolder;
+import org.mortbay.thread.QueuedThreadPool;
+
+public class JettyServerModule extends AbstractModule {
```
suvodeep-pyne: This is more like `ExecJettyServerModule` I presume?

azkaban-exec-server/src/main/java/azkaban/execapp/GuiceConstants.java
```diff
@@ -0,0 +1,8 @@
+package azkaban.execapp;
+
+public class GuiceConstants {
+
+  public static final String EXEC_JETTY_SERVER = "ExecServer";
```
suvodeep-pyne: Can probably move the constants inside the module itself.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -539,12 +505,9 @@ public void shutdown() {
   }
 
   /**
-   * (internal API)
-   * Note: This should be run in a separate thread.
+   * (internal API) Note: This should be run in a separate thread.
```
suvodeep-pyne: Was this change intentional? Else can probably revert this

azkaban-exec-server/src/main/java/azkaban/execapp/JettyServerModule.java
juhoautio: So package name `execapp` is not enough? I can rename it of course if you say so.

azkaban-exec-server/src/main/java/azkaban/execapp/GuiceConstants.java
juhoautio: The reason is that otherwise one might easily create a circular dependency between a module class and the classes to instantiate (that use `@Named` in constructor args). Do you want me to move these any way?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
juhoautio: I suspect Save Actions plugin but can't reproduce. So I will revert it.

juhoautio:
Yeah, Save Actions plugin is modifying some of the javadocs in AzkabanExecutorServer. I reverted those changes manually for now, but that class should be processed eventually.

azkaban-exec-server/src/main/java/azkaban/execapp/GuiceConstants.java
suvodeep-pyne: As long as it is a static variable, I don't think it should create a circular dependency since you don't need to instantiate the class to load a static constant. Could you give it a try and let me know. One of the injection tests should blow up in that case.

suvodeep-pyne:
I see.

juhoautio:
I moved the constants as you requested, although I wouldn't endorse that kind of structure if it was up to me.

> As long as it is a static variable, I don't think it should create a circular dependency since you don't need to instantiate the class to load a static constant.

Ok, I didn't mean a circular dependency in guice terms. I meant simply on Java code level. Not sure if there is a circle currently, but at least it becomes a good probability to happen when we start importing binding names from modules into concrete classes.

Maybe a bigger argument for having the constants separataly is backed by DI itself. I mean, if we think about dependency injection with Guice more like as a principle, the idea is that dependencies can be provided from the outside. Now, when the AzkabanExecutorServer itself depends on a particular provider module (JettyServerModule), then it's not any more as much loose coupling, because you couldn't just remove JettyServerModule and make some other class provide the required injectables for AzkabanExecutorServer: because you also have to change the imports for the constants in AzkabanExecutorServer if JettyServerModule is deleted.

suvodeep-pyne:
TL;DR How many Guice constants do you think the Azkaban Executor needs? If it is a lot, let's have a class. Else, let's hold off until that time.

> Maybe a bigger argument for having the constants separataly is backed by DI itself. I mean, if we think about dependency injection with Guice more like as a principle, the idea is that dependencies can be provided from the outside.

I see your point now.

My take: I wasn't that keen on creating another constants class. We had 5 of them. People didn't know about it and we had (and still have) constants all over the place. I consolidated them together and we now have single `Constants` class. But having a giant `Constants` class isn't great. What's great is that now, everybody uses that single class.

Coming back to the point, I agree with your design and I think that should be the long term state when we have more named injections, etc. What I'm concerned about is a one off 2 member constants class. Some questions are:
 - How many Guice constants do we have in the immediate roadmap?
 - How many use cases of external injection, etc do we see in the near future?

I don't see too many coming and that is what motivates my current suggestion. I am all in to refactor this at any time when there is a need to do so. Do you plan to feed in your own `Server` and `Context` objects? I'm curious.

Regarding the circular dependency thing: I think having the constants in either the class (to be instantiated) or the module is not a good design. But I'm fine with either to start with. Let me also add that the `Guice` usage in Azkaban currently, is in a nascent stage. It is not set up in a manner to be flexible with external modules. For example, a lot of the main services are annotated with `@Singleton` in the class as opposed to the module. We may need to refactor that at some point and that's totally fine too!

That being said, I appreciate you being upfront with me on this.

juhoautio:
> How many Guice constants do we have in the immediate roadmap?
> How many use cases of external injection, etc do we see in the near future?
> Do you plan to feed in your own Server and Context objects?

I suppose none of these reasons apply currently :) I just like to avoid such dependencies in general. It might also help if these classes were subject to some refactoring in the future.

Seems like this PR would be ready to merge?

suvodeep-pyne:
> It might also help if these classes were subject to some refactoring in the future. 

Absolutely. We are in a major rewrite phase at the moment. However, we are doing it in an incremental/iterative fashion.

PR #1114, opened 2017-05-25
Faster JobCallbackRequestMakerTest

juhoautio:
This test used to take ~5s. Now it's ~0.3s.

Remaining slowest test is `azkaban.execapp.JobRunnerTest`: 2.513s.

All enabled tests now run in ~5s \o/

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackRequestMaker.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackRequestMaker.java
@@ -1,5 +1,11 @@
 package azkaban.execapp.event;
 
+import static azkaban.Constants.JobCallbackProperties.JOBCALLBACK_CONNECTION_TIMEOUT;
+import static azkaban.Constants.JobCallbackProperties.JOBCALLBACK_CONNECTION_REQUEST_TIMEOUT;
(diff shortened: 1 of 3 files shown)


juhoautio:
@HappyRay, could you check this please?

azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackRequestMaker.java
```diff
@@ -84,18 +90,14 @@ public static JobCallbackRequestMaker getInstance() {
   private JobCallbackRequestMaker(Props props) {
 
     int connectionRequestTimeout =
-        props.getInt("jobcallback.connection.request.timeout",
-            DEFAULT_TIME_OUT_MS);
+        props.getInt(JOBCALLBACK_CONNECTION_REQUEST_TIMEOUT, DEFAULT_TIME_OUT_MS);
```
HappyRay: Thank you for making them constant and moving them to the central location.

HappyRay:
Excellent. Thanks!

LGTM. Could you update the branch or do you mind that I update the branch so that we can merge?

It appears that you shortened the various Jetty timeout values. It would be great if you can describe what you did to speed up the tests.

juhoautio:
Updated branch.

How this was speeded up: This test was being run using the default timeout values. In the test the normal request time is so short that we can set lower timeouts in test props and still test the timeout scenarios reliably.

HappyRay:
Thanks!

PR #1115, opened 2017-05-25
Fix killed job status update and FlowRunnerTest

juhoautio:
This is the minimal change to make it happen. Background discussion at https://github.com/azkaban/azkaban/pull/1105.

----

1. Make the `FlowRunnerTest` pass again.
    - Remove `@Ignore`.

----

1. Others: see the diff.

----

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -278,6 +278,7 @@ public class FlowRunner extends EventHandler implements Runnable {
       this.watcher.setLogger(logger);
     }
 
+    // Avoid NPE in unit tests when the static app instance is not set
     if (AzkabanExecutorServer.getApp() != null) {
       logger.info("Assigned executor : " + AzkabanExecutorServer.getApp().getExecutorHostPort());
     }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -278,7 +278,9 @@ public class FlowRunner extends EventHandler implements Runnable {
       this.watcher.setLogger(logger);
     }
 
-    logger.info("Assigned executor : " + AzkabanExecutorServer.getApp().getExecutorHostPort());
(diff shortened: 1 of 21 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -278,7 +278,9 @@ private void setupFlowExecution() {
       this.watcher.setLogger(logger);
     }
 
-    logger.info("Assigned executor : " + AzkabanExecutorServer.getApp().getExecutorHostPort());
+    if (AzkabanExecutorServer.getApp() != null) {
+      logger.info("Assigned executor : " + AzkabanExecutorServer.getApp().getExecutorHostPort());
+    }
```
juhoautio: This is enough to make the test work – otherwise there's no need to have that initialized instance of `AzkabanExecutorServer` available via the static getter.

HappyRay: Could you add a comment that this check is for the unit test only?

HappyRay:
Does this pull request depend on #1116? Fix bug in MockExecutorLoader

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
juhoautio: Sure.

juhoautio:
No, this PR alone works fine.

HappyRay:
I thought the status synchronization issue caused this test to be flaky. Am I wrong?

juhoautio:
Yes. But I haven't been able to make this fail now after I switched to using Mockito instead of the mock class that tangled with the job statuses. It requires extreme conditions to have this test fail because of the sync issue alone. To sum it up, this test is working fine even with that sync issue still lurking in the background.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: LGTM after you add the comment.

juhoautio:
Added the comment, @HappyRay.

juhoautio:
Replaced by https://github.com/azkaban/azkaban/pull/1129

PR #1116, opened 2017-05-25
Fix bug in MockExecutorLoader

juhoautio:
Fix concurrency bug in MockExecutorLoader: it shouldn't modify the state of actual flows, so make a clone of the flows, too.

Also log job logs in `MockExecutorLoader` instead of discarding them, so that it's possible to debug failures in tests.

----

The problem was this (originally revealed when `FlowRunnerTest` still used this mock class instead of a Mockito mock):
- MockExecutorLoader kept a handle to the actual ExecutableFlow instance
- MockExecutorLoader.updateExecutableFlow set the job statuses based on the current flow status, which is not synchronized with the job runner threads
- So this wrote an old job status over the newer one in ExecutableNode

Stacktrace for that:

job4: setStatus SUCCEEDED -> QUEUED
	at azkaban.executor.ExecutableNode.setStatus(ExecutableNode.java:141)
	at azkaban.executor.ExecutableNode.applyUpdateObject(ExecutableNode.java:398)
	at azkaban.executor.ExecutableFlowBase.applyUpdateObject(ExecutableFlowBase.java:352)
	at azkaban.executor.ExecutableFlowBase.applyUpdateObject(ExecutableFlowBase.java:369)
	at azkaban.executor.MockExecutorLoader.updateExecutableFlow(MockExecutorLoader.java:120)
	at azkaban.execapp.FlowRunner.updateFlow(FlowRunner.java:322)
	at azkaban.execapp.FlowRunner.updateFlow(FlowRunner.java:316)
	at azkaban.execapp.FlowRunner.progressGraph(FlowRunner.java:552)
	at azkaban.execapp.FlowRunner.runFlow(FlowRunner.java:413)
	at azkaban.execapp.FlowRunner.doRun(FlowRunner.java:238)
	at azkaban.execapp.FlowRunner.run(FlowRunner.java:215)

Overlapping updates by 2 different threads could look something like this:

2017/05/24 12:20:54.695 +0300 INFO [JobRunner-job3-1] [ExecutableNode] job4: applyUpdateObject: RUNNING-> QUEUED
2017/05/24 12:20:55.700 +0300 INFO [JobRunner-job4-1] [JobRunner] job4: changeStatus: QUEUED -> SUCCEEDED
2017/05/24 12:20:55.701 +0300 INFO [FlowRunner-exec-1] [ExecutableNode] job4: applyUpdateObject: SUCCEEDED-> QUEUED

----

Background discussion at https://github.com/azkaban/azkaban/pull/1105.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
@@ -16,18 +16,15 @@
 
 package azkaban.executor;
 
-import azkaban.AzkabanCommonModule;
-import azkaban.ServiceProvider;
-import com.google.inject.Guice;
-import com.google.inject.Injector;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.junit.Assert;
 import org.junit.Test;
@@ -210,21 +207,20 @@ public class ExecutorManagerTest {
     manager.submitExecutableFlow(flow1, testUser.getUserId());
     manager.submitExecutableFlow(flow2, testUser.getUserId());
 
-    List<ExecutableFlow> testFlows = new LinkedList<ExecutableFlow>();
-    testFlows.add(flow1);
-    testFlows.add(flow2);
+    List<Integer> testFlows = Arrays.asList(flow1.getExecutionId(), flow2.getExecutionId());
 
     List<Pair<ExecutionReference, ExecutableFlow>> queuedFlowsDB =
       loader.fetchQueuedFlows();
     Assert.assertEquals(queuedFlowsDB.size(), testFlows.size());
     // Verify things are correctly setup in db
     for (Pair<ExecutionReference, ExecutableFlow> pair : queuedFlowsDB) {
-      Assert.assertTrue(testFlows.contains(pair.getSecond()));
+      Assert.assertTrue(testFlows.contains(pair.getSecond().getExecutionId()));
     }
 
     // Verify running flows using old definition of "running" flows i.e. a
     // non-dispatched flow is also considered running
-    List<ExecutableFlow> managerActiveFlows = manager.getRunningFlows();
+    List<Integer> managerActiveFlows = manager.getRunningFlows()
+        .stream().map(ExecutableFlow::getExecutionId).collect(Collectors.toList());
     Assert.assertTrue(managerActiveFlows.containsAll(testFlows)
       && testFlows.containsAll(managerActiveFlows));
 
--- a/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
+++ b/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
@@ -21,13 +21,22 @@ import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import org.apache.commons.io.FileUtils;
+import org.apache.log4j.Logger;
 
+/**
+ * Used in unit tests to mock the "DB layer" (the real implementation is JdbcExecutorLoader).
+ * Captures status updates of jobs and flows (in memory) so that they can be checked in tests.
+ */
 public class MockExecutorLoader implements ExecutorLoader {
+
+  private static final Logger logger = Logger.getLogger(MockExecutorLoader.class);
 
   Map<Integer, Integer> executionExecutorMapping = new ConcurrentHashMap<>();
   Map<Integer, ExecutableFlow> flows = new ConcurrentHashMap<>();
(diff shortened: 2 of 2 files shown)


azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -49,7 +54,8 @@
   @Override
   public void uploadExecutableFlow(ExecutableFlow flow)
       throws ExecutorManagerException {
-    flows.put(flow.getExecutionId(), flow);
+    ExecutableFlow exFlow = ExecutableFlow.createExecutableFlowFromObject(flow.toObject());
```
HappyRay: Why making a copy here is not obvious without reading this change history. Could you add a comment?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -209,21 +207,20 @@ public void testQueuedFlows() throws ExecutorManagerException, IOException {
     manager.submitExecutableFlow(flow1, testUser.getUserId());
     manager.submitExecutableFlow(flow2, testUser.getUserId());
 
-    List<ExecutableFlow> testFlows = new LinkedList<ExecutableFlow>();
-    testFlows.add(flow1);
-    testFlows.add(flow2);
+    List<Integer> testFlows = Arrays.asList(flow1.getExecutionId(), flow2.getExecutionId());
```
HappyRay: Why is this change necessary?

HappyRay:
Do you think it will be better to do all the following in the future?
1. fix the node status synchronization issue
2. replace MockExecutorLoader with Mockito mock objects when possible.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
juhoautio: I can, but the same thing (cloning) is already done for job nodes, though..

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
juhoautio: Because after cloning Object.equals() can't be used.

juhoautio:
Yes, definitely.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
HappyRay: Which code path did you refer to?

How is it going to help readers understand why cloning here is needed?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
HappyRay: I see.

HappyRay:
If the node status field is marked volatile, will that fix the status synchronization issue? Can we then use the existing test without this change to verify the behavior? Since the node captured in the mockloader is the same instance, overwriting the node status should be fine then. Am I right?

HappyRay:
If the node status field is marked volatile, will that fix the status synchronization issue? Can we then use the existing test without this change to verify the behavior? Since the node captured in the mockloader is the same instance, overwriting the node status should be fine then. Am I right?

juhoautio:
IMHO it's not good that this test would modify the ExecutableNodes any way. The point is to just monitor the calls, similar to Mockito. So it's good to remove this problematic behavior in any case. Yes, volatile modifier should stabilize it, too, but I would see that as a separate issue.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
juhoautio: This: https://github.com/juhoautio/azkaban/blob/8272797bf60cf4009e20300e490e7838573a457c/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java#L129

Maybe there could be a class level comment instead telling what the point of this class is (ie. capture status updates so that they can be checked in unit tests).

Should I also add comments in these places where cloning is done why it's needed?

HappyRay: It would be great if you can add comments in both places.

My understanding is that the class is simulating a DB layer which should be a copy of the original. Thus cloning is the right choice. Am I right?

HappyRay:
I understand now why this PR doesn't depend on the status sync fix. It's because the incorrect status is written to a copy of the original node object. Right?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
juhoautio: Yes, on point.

juhoautio:
Correct.

juhoautio:
@HappyRay good to go?

juhoautio:
@HappyRay: rebased & conflicts resolved.

HappyRay:
LGTM. I appreciate the comments you added to the code.

However mixing the style changes with the real changes makes it harder to see the real changes. I'd suggest keeping code style changes separately. This is going to be less of an issue once we complete code reformatting of the entire code base. 

Per @jamiesjc's request, we delayed reformatting some directories. This is one of them. She has some pending changes she would like to go in first.

Do you mind reverting the code style change and only keep the functional change if it is not too much trouble?

Sorry about the trouble. 

Also since the original bug as described in "The problem was this (revealed when running FlowRunnerTest that also uses this mock class): ... " has been fixed, should this change description be modified?

juhoautio:
> Do you mind reverting the code style change and only keep the functional change

Can do, no problem.

> we delayed reformatting some directories. This is one of them

I believe you mean `ExecutorManagerTest.java`.

juhoautio:
> Also since the original bug as described in "The problem was this (revealed when running FlowRunnerTest that also uses this mock class): ... " has been fixed, should this change description be modified?

I think the current info should be kept, I will just add that since then `FlowRunnerTest` has been changed to use `Mockito` instead of `MockExecutorLoader`.

juhoautio:
@HappyRay now without formatting changes.

HappyRay:
Thanks!
I mean the "executor" directory and the corresponding one in test.

PR #1119, opened 2017-05-25
EventData to always include nestedId

juhoautio:
This gives nicer debug logs because there's no stuff like `"Job null finished with status SUCCEEDED"`.
It's also cleaner to be able to pass the `ExecutableNode` to the constructor instead of having to always manually call `getStatus()` and `getNestedId()`.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/event/EventData.java
+++ b/azkaban-common/src/main/java/azkaban/event/EventData.java
@@ -15,20 +15,19 @@ public class EventData {
    * Creates a new EventData instance.
    *
    * @param status node status.
-   */
-  public EventData(Status status) {
-    this(status, null);
-  }
-
-  /**
-   * Creates a new EventData instance.
-   *
-   * @param status node status.
    * @param nestedId node id, corresponds to {@link ExecutableNode#getNestedId()}.
    */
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -15,20 +15,19 @@
...
    * Creates a new EventData instance.
-   *
-   * @param status node status.
-   * @param nestedId node id, corresponds to {@link ExecutableNode#getNestedId()}.
+   * @param node node.
    */
-  public EventData(Status status, String nestedId) {
-    this.status = status;
-    this.nestedId = nestedId;
+  public EventData(ExecutableNode node) {
```
chengren311: is that possible we only use one constructor? EventData(ExecutableNode node)  ?

juhoautio: I'd like to avoid that bigger change. For this I can easily see that I'm not breaking anything.

HappyRay:
LGTM.

@chengren311 any further comments?

chengren311:
LGTM too. thanks!

PR #1120, opened 2017-05-25
Synchronize ExecutableNode status accessors

juhoautio:
Needed at least in FlowRunnerTest, to be able to always get the right status when FlowRunner runs in another thread. Doesn't do any harm for production code, but synchronization might be also crucial there in some cases.

See also: https://github.com/azkaban/azkaban/pull/1115 (also https://github.com/juhoautio/azkaban/pull/2 where the test was failing quite often without this sync).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
@@ -120,7 +120,7 @@ public class ExecutableNode {
     this.id = id;
   }
 
-  public Status getStatus() {
+  public synchronized Status getStatus() {
     return status;
   }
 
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
```diff
@@ -132,7 +132,7 @@ public void setType(String type) {
     this.type = type;
   }
 
-  public void setStatus(Status status) {
+  public synchronized void setStatus(Status status) {
```
chengren311: I slightly prefer making status volatile.

juhoautio: Nice one 👍 Actually I tried to revert this change and run some tests but couldn't get them fail any more (I didn't add `volatile` yet). I would take a bit of more effort to figure out if there's a real need for any of this, but making it `volatile` any way sounds good to me.

PR #1121, opened 2017-05-26
Revert running flows cache

jamiesjc:
Reverted running flows cache and two fixes on top of the changes.
Ignored some unit tests for running flows cache change for now.
Kept query method fetchActiveFlowByExecId() for future usage.


jamiesjc:
Git revert the main code. Unit tests have some dependencies so did not revert. Just ignored some of the test cases. Did not revert Jdbc query for fetchActiveFlowByExecId() for future usage. Anyway this method is only called in test case for now so should be fine.

PR #1129, opened 2017-05-29
Fix killed job status update and FlowRunner tests

juhoautio:
Make `FlowRunnerTest` & `FlowRunnerTest2` work, and make them run fast.

Good to have the `FlowRunnerTest2` as well because it has useful test cases for the flow pause feature.

Based on https://github.com/azkaban/azkaban/pull/1115, but:
- Now also `FlowRunnerTest2` runs, and it's all in one commit.
- `FlowRunnerTest` is also now simpler because there's no need for additional `waitJobsStarted(..)` calls: `assertStatus()` handles both waiting and assertion.

----

This is the minimal change to just make the tests pass again. Background discussion can be seen at https://github.com/azkaban/azkaban/pull/1105.

1. Make the `FlowRunnerTest` and `FlowRunnerTest2` pass again.
    - Removed `@Ignore`.

1. Others: see the diff.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
@@ -16,6 +16,10 @@
 
 package azkaban.utils;
 
+import azkaban.executor.ExecutableFlowBase;
+import azkaban.flow.CommonJobProperties;
+import com.google.common.collect.MapDifference;
+import com.google.common.collect.Maps;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
@@ -27,33 +31,28 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import org.apache.commons.jexl2.Expression;
 import org.apache.commons.jexl2.JexlEngine;
 import org.apache.commons.jexl2.JexlException;
 import org.apache.commons.jexl2.MapContext;
 import org.apache.commons.lang.StringUtils;
-
 import org.apache.log4j.Logger;
 import org.joda.time.DateTime;
-
-import com.google.common.collect.Maps;
-import com.google.common.collect.MapDifference;
-
-import azkaban.executor.ExecutableFlowBase;
-import azkaban.flow.CommonJobProperties;
 
 public class PropsUtils {
 
   private static final Logger logger = Logger.getLogger(PropsUtils.class);
+  private static final Pattern VARIABLE_REPLACEMENT_PATTERN = Pattern
+      .compile("\\$\\{([a-zA-Z_.0-9]+)\\}");
+
   /**
-   * Load job schedules from the given directories ] * @param dir The directory
-   * to look in
+   * Load job schedules from the given directories
    *
+   * @param dir The directory to look in
    * @param suffixes File suffixes to load
    * @return The loaded set of schedules
    */
-  public static Props loadPropsInDir(File dir, String... suffixes) {
+  public static Props loadPropsInDir(final File dir, final String... suffixes) {
     return loadPropsInDir(null, dir, suffixes);
   }
 
@@ -65,35 +64,35 @@ public class PropsUtils {
    * @param suffixes File suffixes to load
    * @return The loaded set of schedules
    */
-  public static Props loadPropsInDir(Props parent, File dir, String... suffixes) {
+  public static Props loadPropsInDir(final Props parent, final File dir, final String... suffixes) {
     try {
-      Props props = new Props(parent);
-      File[] files = dir.listFiles();
+      final Props props = new Props(parent);
+      final File[] files = dir.listFiles();
       Arrays.sort(files);
       if (files != null) {
-        for (File f : files) {
+        for (final File f : files) {
           if (f.isFile() && endsWith(f, suffixes)) {
             props.putAll(new Props(null, f.getAbsolutePath()));
           }
         }
       }
       return props;
-    } catch (IOException e) {
+    } catch (final IOException e) {
       throw new RuntimeException("Error loading properties.", e);
     }
   }
 
-  public static Props loadProps(Props parent, File... propFiles) {
+  public static Props loadProps(final Props parent, final File... propFiles) {
     try {
       Props props = new Props(parent);
-      for (File f : propFiles) {
+      for (final File f : propFiles) {
         if (f.isFile()) {
           props = new Props(props, f);
         }
       }
 
       return props;
-    } catch (IOException e) {
+    } catch (final IOException e) {
       throw new RuntimeException("Error loading properties.", e);
     }
   }
@@ -105,9 +104,9 @@ public class PropsUtils {
    * @param suffixes The suffixes to load
    * @return The properties
    */
-  public static Props loadPropsInDirs(List<File> dirs, String... suffixes) {
-    Props props = new Props();
-    for (File dir : dirs) {
+  public static Props loadPropsInDirs(final List<File> dirs, final String... suffixes) {
+    final Props props = new Props();
+    for (final File dir : dirs) {
       props.putLocal(loadPropsInDir(dir, suffixes));
     }
     return props;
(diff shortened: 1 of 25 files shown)


HappyRay:
I tried to update your branch and finish the review before the large reformat change is merged so that you don't have to deal with merge conflicts.

However, two tests failed. See the Travis log:

"azkaban.execapp.FlowRunnerTest > exec1FailedFinishRest FAILED
    java.nio.channels.ClosedByInterruptException
        at java.nio.channels.spi.AbstractInterruptibleChannel.end(AbstractInterruptibleChannel.java:202)
        at sun.nio.ch.FileChannelImpl.size(FileChannelImpl.java:314)
        at org.apache.commons.io.FileUtils.doCopyFile(FileUtils.java:1142)
        at org.apache.commons.io.FileUtils.doCopyDirectory(FileUtils.java:1428)
        at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1389)
        at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1261)
        at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1230)
        at azkaban.execapp.FlowRunnerTest.prepareExecDir(FlowRunnerTest.java:365)
        at azkaban.execapp.FlowRunnerTest.exec1FailedFinishRest(FlowRunnerTest.java:260)
azkaban.execapp.FlowRunnerTest2 > testFailedFinishingFailure3 FAILED
    java.lang.IllegalStateException: joba1 wasn't added in testJobs map
        at azkaban.executor.InteractiveTestJob.getTestJob(InteractiveTestJob.java:48)
        at azkaban.execapp.FlowRunnerTest2.testFailedFinishingFailure3(FlowRunnerTest2.java:520)
74 tests completed, 2 failed, 22 skipped
"

HappyRay:
Could you carry over the details in the change description you had in #115?

HappyRay:
in azkaban.execapp.FlowRunnerTest2#testNormalFailure2

duplicate asserts:
    // After it starts up, only joba should be running
    assertStatus("joba", Status.RUNNING);
    assertStatus("joba", Status.RUNNING);

juhoautio:
Removed the duplicate assert.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -122,6 +122,9 @@ public void uploadExecutableNode(ExecutableNode node, Props inputParams)
   @Override
   public void updateExecutableNode(ExecutableNode node)
       throws ExecutorManagerException {
+    if (!nodes.containsKey(node.getId())) {
+      uploadExecutableNode(node, new Props());
+    }
```
juhoautio: After updating latest changes from `master`, the test was failing here because the node wasn't found in the map. That could be just concurrency issue, because here `nodes` is just a regular `HashMap` without thread safety.. Or then there's an actual change in behaviour that makes an update before initial upload. I wonder if that would be a problem when running against the actual DB layer..

juhoautio:
Oops, new merge conflicts.. Is it now a good time to resolve them, so that this could be merged before new conflicts arrive? :)

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -361,10 +362,20 @@ private void assertAttempts(String name, int attempt) {
   
   private ExecutableFlow prepareExecDir(File execDir, String flowName,
       int execId) throws IOException {
-    synchronized (this) {
-      // clean interrupted status
-      Thread.interrupted();
-      FileUtils.copyDirectory(execDir, workingDir);
+    // copyDirectory has been failing randomly with ClosedByInterruptException. have some retries.
```
HappyRay: Any idea why this is happening? Can you think of a better way?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
HappyRay: It seems that this error happened consistently. It would be good to find out if this is an indication of a real issue.

HappyRay:
We just applied code formatting changes. Sorry about the inconvenience.
Please see the readme for the updated instruction on how to auto-format code changes. 
Yes, it will be good to resolve conflicts.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
juhoautio: Yes, but I hope that wouldn't block getting these tests into master. This only started happening after some recent changes merged into master.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
juhoautio: Actually good question and I think that's because some tests still run `runner.run()` in the main unit test thread. Part of FlowRunner design is calling `interrupt()` on the FlowRunner thread to wake it up for work. Should be possible to get rid of this error by always running the FlowRunner in its own thread. I'll try switching to that and then can remove all these hacks from the test (including call to `Thread.interrupted();`).

HappyRay: As long as the tests are stable and you are working on improving it, I will be happy to accpet this pull request after the conflicts are resolved.

juhoautio:
So do you actually require [Intellij Save Actions Plugin](https://github.com/dubreuia/intellij-plugin-save-actions) to be applied on the code before submitting pull requests? Or is this formatting I have here fine as long as merge conflicts have been resolved?

juhoautio:
I took the save actions plugin into use, so the diff grew a bit. I like how it improves the code structure, though.

juhoautio:
> Could you carry over the details in the change description you had in #115?

Added (actually #1115).

juhoautio:
Looks like running the `FlowRunner` in its own thread in `FlowRunnerTest` was indeed enough to get rid of `ClosedByInterruptException` in `FileUtils.copyDirectory`.

azkaban-common/src/main/java/azkaban/utils/PropsUtils.java
```diff
@@ -229,8 +233,8 @@ private static String resolveVariableExpression(final String value) {
   }
 
   /**
-   * Function that looks for expressions to parse. It parses backwards to
-   * capture embedded expressions
+   * Function that looks for expressions to parse. It parses backwards to capture embedded
+   * expressions
```
juhoautio: Hmm do I have a different line length limit?

juhoautio: I'll try loading `intellij-java-google-style.xml` again, looks like it was changed, too.

juhoautio: No change, this is still how the javadoc lines are being cut. Leaving like this then.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
```diff
@@ -41,56 +46,54 @@
...
  *
- * jobd:innerFlow2
+ *  jobd:innerFlow2
  *
- * innerJobA
- * |
- * innerFlow2
+ *       innerJobA
+ *           |
+ *       innerFlow2
```
juhoautio: Auto-format doesn't treat this javadoc too well.. I fixed it manually for now ^

juhoautio:
Ok done with the cleanup.

juhoautio:
One more commit:

`MockExecutorLoader` to use thread safe maps
- This should be the proper fix to concurrency issue in `updateExecutableNode()`.

..but let's see how Travis likes it.

HappyRay:
Thanks!
These tests will provide much-needed safeguards against regressions.

HappyRay:
Oops, somehow Github didn't include the full change description and I didn't catch it when I merged it. At least we have the link back to this pull request. 
I will be more careful next time.

HappyRay:
@juhoautio The travis build on the master failed with the following error:

https://travis-ci.org/azkaban/azkaban/builds
https://travis-ci.org/azkaban/azkaban/builds/239316403

azkaban.execapp.FlowRunnerTest2 > testRetryOnFailure FAILED

    java.lang.AssertionError: Wrong status for [jobb:innerFlow] expected:<SKIPPED> but was:<DISABLED>

        at org.junit.Assert.assertEquals(Assert.java:118)

        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:137)

        at azkaban.execapp.FlowRunnerTest2.testRetryOnFailure(FlowRunnerTest2.java:661)

Looks like the test is still flaky. 
Could you take a look?
If necessary disable the flaky tests while investigating.

juhoautio:
Sorry, I didn't see that happen. If you like, could add @Ignore asap. I can take a look later.

HappyRay:
Thanks. I will if it blocks progress.

azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
```diff
@@ -24,7 +26,7 @@
 
 public class InteractiveTestJob extends AbstractProcessJob {
 
-  private static final ConcurrentHashMap<String, InteractiveTestJob> testJobs =
+  public static final ConcurrentHashMap<String, InteractiveTestJob> testJobs =
```
HappyRay: Why making it public?

I wonder why this is static too.

HappyRay:
The same test failed again.

testRetryOnFailure

https://travis-ci.org/azkaban/azkaban/builds/239344361?utm_source=email&utm_medium=notification
I am going to ignore this test for now.
Opened a ticket #1155

HappyRay:
Pull request to ignore two tests #1156

azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
juhoautio: Sorry for that. It was made public so that elsewhere one can do `synchronized (testJobs)` and `testJobs.wait()` to be notified immediately of an update. Of course better would be to have that method in `InteractiveTestJob` itself, so that `testJobs` could still be private.

Static: I guess that's just been the easy way for someone to implement this, but it definitely wouldn't have to be so.

HappyRay: np. I didn't catch this in earlier reviews.
Do you mind improving it later?

juhoautio: Sure, it's on me to do this properly.

HappyRay: Thanks

juhoautio: Now I realized why it is `static`: because this way we can affect "job type" behaviour in the tests. Basically to tell it to succeed or fail, when it's waiting for that kind of instruction (there are some test cases that wouldn't be possible otherwise). Other option would be to somehow expose the private field `Job job` in `JobRunner`, so that the test could get it, cast to `InteractiveTestJob` and call `succeedJob` or `failJob`. I'm not going to make any change to that.

juhoautio: Here's the PR for reverting public to private: https://github.com/azkaban/azkaban/pull/1163

PR #1132, opened 2017-05-31
Create index on status for execution_flows DB table.

jamiesjc:
This is to accelerate the query speed when fetching active flows by the status.

### DIFF
--- a/azkaban-db/src/main/sql/create.execution_flows.sql
+++ b/azkaban-db/src/main/sql/create.execution_flows.sql
@@ -20,3 +20,4 @@ CREATE INDEX ex_flows_end_time ON execution_flows(end_time);
 CREATE INDEX ex_flows_time_range ON execution_flows(start_time, end_time);
 CREATE INDEX ex_flows_flows ON execution_flows(project_id, flow_id);
 CREATE INDEX executor_id ON execution_flows(executor_id);
+CREATE INDEX ex_flows_staus ON execution_flows(status);


HappyRay:
@jamiesjc could you share the tests you did? e.g. the SQL explain result?
Reminder: find ways to notify users that they need to update their DB to avoid surprises. e.g. highlight it in the release note.
cc @ameyamk

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -19,4 +19,5 @@ CREATE INDEX ex_flows_start_time ON execution_flows(start_time);
 CREATE INDEX ex_flows_end_time ON execution_flows(end_time);
 CREATE INDEX ex_flows_time_range ON execution_flows(start_time, end_time);
 CREATE INDEX ex_flows_flows ON execution_flows(project_id, flow_id);
-CREATE INDEX executor_id ON execution_flows(executor_id);
\ No newline at end of file
+CREATE INDEX executor_id ON execution_flows(executor_id);
```
jamiesjc: I did not add this executor_id index. It's the same as the previous code. I guess it's showing as different because of the endline mark.

jamiesjc:
Thanks @HappyRay, I've updated previous doc with the SQL explain result. Also I'll make sure to highlight the DB update in the release note.

HappyRay:
I think what @suvodeep-pyne did in azkaban-db/src/main/sql/upgrade.3.20.0.to.3.22.0.sql 
https://github.com/HappyRay/azkaban/blob/b4c4051ce835742124c34312572c566e71d73862/azkaban-db/src/main/sql/upgrade.3.20.0.to.3.22.0.sql#L6-L6
is good in that it makes it easier to apply the DB update and more explicit. He also added comments.
Consider following his example?

PR #1136, opened 2017-06-01
fix the test cases which might be broken with HTML reformatting

reallocf:


HappyRay:
Could you replace 
replaceAll("\\s+", ""),
with a utility method that has a more meaningful name?

HappyRay:
This may be useful for this purpose

http://joel-costigliola.github.io/assertj/assertj-core-news.html#assertj-core-2.8.0-isEqualToNormalizingWhitespace

PR #1137, opened 2017-06-01
reformatting the entire codebase

reallocf:
1. reformatting the entire codebase based on our coding style
2. reorganize the order of import
3. rearrange member methods
also leverage save action plugin to refactor code(https://github.com/dubreuia/intellij-plugin-save-actions):
Add final to local variable
Add final to field
Remove explicit generic type for diamond
Qualify field access with this
Remove unused suppress warning annotation
Remove final from private method
Remove unnecessary semicolon
Add missing @Override annotations


HappyRay:
Yeah. Thanks!

HappyRay:
@reallocf Nice change descriptions including the intermediate changes. Thanks for including the link to the plugin that we decide to use.

PR #1138, opened 2017-06-01
fix access to input and output property files for python, ruby and script jobs

### LINKED ISSUE #319: Python job-type does not have access to JOB_PROP_FILE and JOB_OUTPUT_PROP_FILE
We can't use a Python job-type to receive data from a previous job or send data to the next job. 

LongArgJob's run method always returns a zero for builder.getEnv().size (line 65). My guess is the culprit is in the constructor, which sets the builder member variable without first calling initPropFiles(). 

  public LongArgJob(String jobid, String[] command, Props sysProps,
      Props jobProp, Logger log, Set<String> suppressedKeys) {
    super(jobid, sysProps, jobProp, log);

```
this.builder =
    new AzkabanProcessBuilder(command)
        .setEnv(getJobProps().getMapByPrefix(ENV_PREFIX))
        .setWorkingDir(getCwd()).setLogger(getLog());
appendProps(suppressedKeys);
```

  }

LongArgJob does call initPropFiles() in its run method. 

My guess regarding this bug is based on looking at ProcessJob's run method's lines 67-102 and comparing them to LongArgJob's run method. Lines 67-10-2 of ProcessJob's run method are:

```
File[] propFiles = initPropsFiles();
Map<String, String> envVars = getEnvironmentVariables();

for (String command : commands) {
  info("Command: " + command);
  AzkabanProcessBuilder builder =
      new AzkabanProcessBuilder(partitionCommandLine(command))
          .setEnv(envVars).setWorkingDir(getCwd()).setLogger(getLog());

  if (builder.getEnv().size() > 0) {
    info("Environment variables: " + builder.getEnv());
  }
  info("Working directory: " + builder.getWorkingDir());

  boolean success = false;
  this.process = builder.build();

  try {
    this.process.run();
    success = true;
  } catch (Throwable e) {
    for (File file : propFiles)
      if (file != null && file.exists())
        file.delete();
  } finally {
    this.process = null;
    info("Process completed "
        + (success ? "successfully" : "unsuccessfully") + " in "
        + ((System.currentTimeMillis() - startMs) / 1000) + " seconds.");
  }
}

// Get the output properties from this job.
```

  }

sjakthol:
The LongArgJob was setting the environmental variables before the input and output property files were created. This meant that the process that was started did not have access to the input or output property files.

These changes modify the LongArgJob so that the property files are created before the environment variables are collected and configured for the job.

As the environment was previously set during the initialization, subclasses had a chance to add their own environmental variables to the process builder before the process was started. To keep this backwards compatible, any variables that have been added to the environmental variable map before the job runs will be merged to the final environment overriding any statically defined variables.

Fixes #319

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/LongArgJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/LongArgJob.java
@@ -21,6 +21,7 @@ import azkaban.jobExecutor.utils.process.AzkabanProcessBuilder;
 import azkaban.utils.Props;
 import java.io.File;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import org.apache.log4j.Logger;
@@ -46,7 +47,6 @@ public abstract class LongArgJob extends AbstractProcessJob {
 
     this.builder =
         new AzkabanProcessBuilder(command)
-            .setEnv(getJobProps().getMapByPrefix(ENV_PREFIX))
             .setWorkingDir(getCwd()).setLogger(getLog());
     appendProps(suppressedKeys);
   }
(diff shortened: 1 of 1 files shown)


sjakthol:
Rebased the changes.

HappyRay:
Thanks for the PR.

Please see my comments in the linked issue #319.

Does the command job type suffer from the same issue?

sjakthol:
As commented on the issue #319, we would be able to live without these job types.

As far as I can remember, the command job does not suffer from this issue.

HappyRay:
@sjakthol 
Thanks! I didn't realize that exe-as-user is not supported for these job types. They are not maintained.

HappyRay:
We are removing these job types. #1431

PR #1146, opened 2017-06-03
Prevent IDE auto formatting from messing up ascii art in javadoc

HappyRay:
Issue:

 The javadoc of the FlowRunnerTest2 class uses ascii art to describe
 the DAG. When intellij auto format runs, it renders the art
 incorrectly.

 Fix:

 Wrapped the parts affected in ```<pre> </pre>``` blocks.

PR #1147, opened 2017-06-03
fix-intellij-warnings-inFlowRunnerTest2

HappyRay:


HappyRay:
Would merge after #1146 due to dependencies.
@juhoautio

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
```diff
@@ -110,7 +115,9 @@ public void setUp() throws Exception {
     if (this.workingDir.exists()) {
       FileUtils.deleteDirectory(this.workingDir);
     }
-    this.workingDir.mkdirs();
+    if (!this.workingDir.mkdirs()) {
+      throw new RuntimeException("Failed to create working directory: " + this.workingDir);
+    }
```
juhoautio: As `FileUtils` is already used on the previous "line", this could be replaced by a one-liner: `FileUtils.forceMkdir(this.workingDir);`

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
```diff
@@ -1081,17 +1083,9 @@ public void testPauseFailKill() throws Exception {
...
   }
 
-  private Thread runFlowRunnerInThread(final FlowRunner runner) {
+  private void runFlowRunnerInThread(final FlowRunner runner) {
     final Thread thread = new Thread(runner);
     thread.start();
-    return thread;
-  }
-
-  private void sleep(final long millisec) {
```
juhoautio: Ha! I remember thinking that now I can finally remove this method, but forgot about it.

juhoautio:
👍

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
HappyRay: Done. Thanks

HappyRay:
@juhoautio please take another look.

juhoautio:
Good stuff 👍 Indeed I wanted to avoid making too many changes in the original PR that was to make these tests pass, although I could see a lot that could've been cleaned up. No more IDE warnings then?

juhoautio:
For example look at this line:
https://github.com/azkaban/azkaban/blob/c55e88237ebc252eaedca6e2ea32e38cf853fca2/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java#L166-L173
– shouldn't need to do anything else but call `eventCollector.checkEventExists`, and that in turn should be simplified to user `Assert.*` methods instead of custom conditions + throws: https://github.com/azkaban/azkaban/blob/c55e88237ebc252eaedca6e2ea32e38cf853fca2/azkaban-exec-server/src/test/java/azkaban/execapp/EventCollectorListener.java#L83-L91

HappyRay:
I would prefer small incremental commits too. And I suggested to the team members to use issues to tie multiple commits/PRs together when needed. 

It's less work to combine this commit with the previous one #1146 for example. 
To make this model work well, however, we need to have quick turn arounds on code review, unit tests etc.

HappyRay:
Good suggestions on other improvements. I agree. Would you like to make these changes?

HappyRay:
There are many areas in this code base that can be improved. As long as we continue to improve, over time, I am optimistic that developers will become more productive working on this project.

HappyRay:
There is still a weak warning about some duplicated code left in this class. Since a fix is more involved, I will defer that.

juhoautio:
Yeah I can handle that Exception -> Assert change, that I mention in https://github.com/azkaban/azkaban/pull/1147#issuecomment-306003984.

HappyRay:
That will be great. Thanks

PR #1148, opened 2017-06-03
Fix JobRunner bug when job is cancelled early

juhoautio:
This bug was revealed by `JobRunnerTest#testDelayedExecutionCancelledJob`:

If job was cancelled early, there was no initial "upload" ([JobRunner.java#L590](https://github.com/azkaban/azkaban/blob/c55e88237ebc252eaedca6e2ea32e38cf853fca2/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L590)) for the `ExecutableNode`, just an eventual "update" ([JobRunner.java#L626](https://github.com/azkaban/azkaban/blob/c55e88237ebc252eaedca6e2ea32e38cf853fca2/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L626)).

This bug also kicks in when running against the DB, because there would be an `UPDATE` without a preceding `INSERT`. So nothing would be written into the DB, because the `WHERE` conditions of the `UPDATE` don't match any row.

Test console output **before** the fix:

```
Create temp dir
2017/06/04 01:14:28.643 +0300 INFO [main] [JobTypeManager] Loading plugin default job types
2017/06/04 01:14:28.736 +0300 INFO [JobRunner-testJob-1] [JobRunnerTest] Created file appender for job testJob
2017/06/04 01:14:28.736 +0300 INFO [JobRunner-testJob-1] [JobRunnerTest] Attached file appender for job testJob
2017/06/04 01:14:30.736 +0300 INFO [JobRunner-testJob-1] [JobRunnerTest] No attachment file for job testJob written.
2017/06/04 01:14:30.737 +0300 ERROR [JobRunner-testJob-1] [JobRunner] Unexpected exception
	at azkaban.executor.MockExecutorLoader.updateExecutableNode(MockExecutorLoader.java:121)
	at azkaban.execapp.JobRunner.writeStatus(JobRunner.java:395)
	at azkaban.execapp.JobRunner.doRun(JobRunner.java:626)
	at azkaban.execapp.JobRunner.run(JobRunner.java:558)
	at java.lang.Thread.run(Thread.java:745)
Exception in thread "JobRunner-testJob-1" java.lang.NullPointerException
	at azkaban.executor.MockExecutorLoader.updateExecutableNode(MockExecutorLoader.java:121)
	at azkaban.execapp.JobRunner.writeStatus(JobRunner.java:395)
	at azkaban.execapp.JobRunner.doRun(JobRunner.java:626)
	at azkaban.execapp.JobRunner.run(JobRunner.java:558)
	at java.lang.Thread.run(Thread.java:745)
Teardown temp dir
```

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -584,13 +584,9 @@ public class JobRunner extends EventHandler implements Runnable {
     // Start the node.
     this.node.setStartTime(System.currentTimeMillis());
     Status finalStatus = this.node.getStatus();
+    uploadExecutableNode();
     if (!errorFound && !isKilled()) {
       fireEvent(Event.create(this, Type.JOB_STARTED, new EventData(this.node)));
-      try {
-        this.loader.uploadExecutableNode(this.node, this.props);
-      } catch (final ExecutorManagerException e1) {
-        this.logger.error("Error writing initial node properties");
-      }
 
       final Status prepareStatus = prepareJob();
       if (prepareStatus != null) {
@@ -624,6 +620,14 @@ public class JobRunner extends EventHandler implements Runnable {
     finalizeLogFile(attemptNo);
     finalizeAttachmentFile();
     writeStatus();
+  }
+
+  private void uploadExecutableNode() {
+    try {
+      this.loader.uploadExecutableNode(this.node, this.props);
+    } catch (final ExecutorManagerException e1) {
+      this.logger.error("Error writing initial node properties");
+    }
   }
 
   private Status prepareJob() throws RuntimeException {
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -306,6 +306,7 @@ public class JobRunnerTest {
     Assert.assertTrue(outputProps == null);
     Assert.assertTrue(logFile.exists());
 
+    Assert.assertEquals(2L, (long) loader.getNodeUpdateCount("testJob"));
     eventCollector.checkEventExists(Type.JOB_FINISHED);
   }
 


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -626,6 +622,14 @@ private void doRun() {
     writeStatus();
   }
 
+  private void uploadExecutableNode() {
+    try {
+      this.loader.uploadExecutableNode(this.node, this.props);
+    } catch (final ExecutorManagerException e1) {
```
chengren311: Thanks for catching this issue!
Can we also refactor this method a bit?
e.x: final ExecutorManagerException e instead of e1
and log the exception message when catch the exception.

juhoautio: Yes indeed. Will do. And also pass the exception to the logger for a stack trace.

juhoautio: Although it makes me wonder does it make sense to continue at all if this fails? Maybe throw an exception instead?

HappyRay:
It's great to see tests catch this type of issues.
Thanks for this work!

How did you make the test pass before this bug is fixed?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: Yes, I have the same question. If in doubt, I am in favor of failing early. At least we understand the behavior better.

That said, I would suggest that we make small incremental changes. e.g. fix this issue first and then do further improvements.

juhoautio:
> How did you make the test pass before this bug is fixed?

Yeah, the test still passed, it just printed this error in the console and I happened to notice it.

This test only checks that a status change event to JOB_FINISHED was fired. It doesn't check the "saved" `ExecutableNode` status in `MockExecutorLoader`, so the test passed regardless of this error. And one part is of course that the error is being ignored in `JobRunner` instead of stopping there, which we discussed above.

juhoautio:
Did I understand correctly that you like to have this change as it is? And make those further improvements separately.. But let me know if you're still expecting something from me here.

HappyRay:
Yes.

HappyRay:
Oh, I forgot. 
Now that we discovered this bug. Can you think of ways to prevent a regression?
I'd prefer a targeted test that test one behavior at a time. It's easier to figure out what's wrong when a test fails.
Ideally we would like to have that test together with the fix.

juhoautio:
Added commit, assert that caught the error before fix in JobRunner. Next: resolving conflicts.

juhoautio:
@HappyRay, plz check again.

PR #1150, opened 2017-06-03
Simplify tests: EventCollectorListener & runner tests

juhoautio:
1st commit: "Simplify tests: usage of `EventCollectorListener#checkEventExists`".

- Using hamcrest matcher `is` to compare arrays, because it gives an informative error message, like:

```
java.lang.AssertionError:
Expected: is [<FLOW_STARTED>, <FLOW_FINISHED>]
     but: was [<FLOW_STARTED>]
```

- `EventCollectorListener #writeAllEvents` deleted because not used anywhere.

----

2nd commit "Test cleanup: better naming, more readable" is kind of optional, but I think it makes this even better.

----

3rd commit: "Test cleanup: simplify, remove duplicate code" (I was a bit lazy when creating the `FlowRunnerTestBase`).

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
@@ -22,10 +22,8 @@ public class FlowRunnerTestBase {
   public static boolean isStarted(final Status status) {
     if (status == Status.QUEUED) {
       return false;
-    } else if (!Status.isStatusFinished(status) && !Status.isStatusRunning(status)) {
-      return false;
     }
-    return true;
+    return Status.isStatusFinished(status) || Status.isStatusRunning(status);
   }
 
   public void assertThreadShutDown() {
@@ -104,17 +102,7 @@ public class FlowRunnerTestBase {
 
   protected void assertFlowStatus(final Status status) {
     final ExecutableFlow flow = this.runner.getExecutableFlow();
-    for (int i = 0; i < 100; i++) {
-      if (flow.getStatus() == status) {
-        break;
-      }
-      synchronized (EventCollectorListener.handleEvent) {
-        try {
-          EventCollectorListener.handleEvent.wait(10L);
-        } catch (final InterruptedException e) {
-        }
-      }
-    }
+    waitForStatus(flow, status);
     printStatuses(status, flow);
     assertEquals(status, flow.getStatus());
   }
@@ -123,6 +111,12 @@ public class FlowRunnerTestBase {
     final ExecutableFlow exFlow = this.runner.getExecutableFlow();
     final ExecutableNode node = exFlow.getExecutableNodePath(name);
     assertNotNull(name + " wasn't found", node);
+    waitForStatus(node, status);
+    printStatuses(status, node);
+    assertEquals("Wrong status for [" + name + "]", status, node.getStatus());
+  }
+
+  private void waitForStatus(ExecutableNode node, Status status) {
     for (int i = 0; i < 100; i++) {
       if (node.getStatus() == status) {
         break;
@@ -131,11 +125,10 @@ public class FlowRunnerTestBase {
         try {
           EventCollectorListener.handleEvent.wait(10L);
         } catch (final InterruptedException e) {
+          i--;
         }
       }
     }
-    printStatuses(status, node);
-    assertEquals("Wrong status for [" + name + "]", status, node.getStatus());
   }
 
   protected void printStatuses(final Status status, final ExecutableNode node) {
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
 import azkaban.event.Event;
+import azkaban.event.Event.Type;
 import azkaban.executor.ExecutableFlow;
 import azkaban.executor.ExecutableFlowBase;
 import azkaban.executor.ExecutableNode;
@@ -168,6 +169,10 @@ public class FlowRunnerTestBase {
     for (final String name : jobs) {
       InteractiveTestJob.getTestJob(name).succeedJob();
     }
+  }
+
+  protected void assertEvents(final Type... types) {
+    eventCollector.assertEvents(types);
   }
 
 }
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/EventCollectorListener.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/EventCollectorListener.java
@@ -63,7 +63,7 @@ public class EventCollectorListener implements EventListener {
     return true;
   }
 
-  public void checkEventExists(final Type... expected) {
+  public void assertEvents(final Type... expected) {
     Object[] captured = this.eventList.stream()
         .filter(event -> event.getRunner() != null)
         .map(event -> event.getType())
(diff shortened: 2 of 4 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
```diff
@@ -170,4 +171,8 @@ protected void succeedJobs(final String... jobs) {
     }
   }
 
+  protected void assertEvents(final Type... types) {
```
HappyRay: The method name change is fine.

What's the advantage for turning the eventCollector from a local variable into a member variable?
The existing style seems to be more explicit to me since I don't have to navigate to the base class to understand the logic and since this method is protected, I will need to pay attention to see if it is overwritten.

HappyRay:
Thanks for separating these commits.

Since we do squash and merge only, all the commits will be merged into one commit.

They seem to be independent enough even though one builds on another. Do you mind breaking them up into separate pull requests?

HappyRay:
Have you considered using assertj instead?

It seems to me that assertj is a better choice. I also got the same recommendation from a testing expert.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
juhoautio: I think it makes the test cases easier to read. Same pattern is already used for assertStatus. I don't see it as a problem that you need to navigate to the superclass if you want to see how it's implemented. That isn't much different than navigating to the method if it was at the bottom of the concrete test class - at least if using an IDE, like one usually does.

HappyRay: Navigation is not a big issue although I would prefer minimizing the redirection when possible. e.g.   protected void assertEvents(final Type... types) {
    eventCollector.assertEvents(types);
  }
is one of such redirection. Now I have to remember where the eventCollector member comes from and where it is modified. I am looking at the base class. But the member is modified in the subclasses.

The bigger issue is to reduce the mental overhead of figuring out the states.  I prefer limiting the scope of a variable as much as possible. e.g. What's the advantage for turning the eventCollector from a local variable into a member variable? as I asked earlier.
 I would prefer composition over inheritance. e.g. it looks like the FlowRunnerTestBase can be turned into a utility class with static methods and pass in the runner object when needed to the methods that need it. 

See my pull request #1151 for an example. the new prepareProject method is relatively simple to understand in terms of what are the inputs and outputs. 

Reference:
https://stackoverflow.com/questions/11343840/favor-composition-over-inheritance
And the "Effective Java" book item 14: favor-composition-over-inheritance

If you can break up this pull request into simpler multiple ones, we can merge the first commit while we discuss the rest. This is another advantage of smaller and independent pull requests i.e. we can treat different parts differently.

juhoautio: Ok, disclaimer first of all: I totally respect the principles that you have. I think it's easier for both of us that I just make the changes that you want, even if my own preferences would be different. Below I'll also answer your questions, because maybe I can learn something here.

> What's the advantage for turning the `eventCollector` from a local variable into a member variable?

That you don't need that extra arg on every single line that needs to assert events.

> prefer composition over inheritance

I recognize the problems with inheritance in general, but sometimes it's just the simplest way to go. In this case I personally don't see that it's causing any real confusion. In case of unit tests I feel more like working on a standalone script than part of bigger application architecture, and there I feel it's better to establish a kind of DSL that makes it easy to create and understand different test cases: easily seeing the essential parts of what's being tested. So I try to remove as much noise as possible.

I checked https://github.com/azkaban/azkaban/pull/1151 and it's fine, but even there every test class has this "much" duplicate code:
```java
this.flowMap = FlowRunnerTestUtil
    .prepareProject(this.project, dir, this.logger, this.workingDir);
```

If you want to add an optional argument or something like that, you need to update every class that calls this method. With inheritance you can just add a new field to the base class and deal with it without having to update the subclasses if they're not interested in that "new feature".

With inheritance all the other args could be protected fields in the base class, and maybe in `setUp` there would be:

```java
super.prepareProject(dir);
```

> avoid redirection

I'm pretty sure `eventCollector` could be created in `FlowRunnerTestBase#setUp` and `FlowRunnerTest#setUp` would call `super.setUp()`. Maybe it could even be `private` then. But maybe that doesn't really improve the structure from your perspective.

> If you can break up this pull request into simpler multiple ones, we can merge the first commit while we discuss the rest. This is another advantage of smaller and independent pull requests i.e. we can treat different parts differently.

That is very true, but will add quite much overhead (on my part) when the changes build upon each other. Any way I will try my best to separate different things into separate PRs.

juhoautio:
> Have you considered using assertj instead?

assertj is better in general, but in this case I'd like to stick with hamcrest, because:
- No need to add assertj as a gradle dependency (hamcrest is already there)
- In this particular case assertj API doesn't make it much simpler compared to using hamcrest API (and after all this is already written now – maybe the main benefit of assertj is that it's easy to write assertions with IDE)
- I don't know what kind of output assertj would give in the error case, but I know for sure that hamcrest now gives exactly what's the right thing to print

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
HappyRay: Yes, we can agree to disagree. I am hoping to learn from this discussion and your perspective too. It doesn't mean that we have to agree on everything.

With the help of IDEs, I find it easier nowadays to optimize for reading than writing. 

The code duplication you referenced is much less than before. Yes, I'd like to further simplify the code. 
"this.flowMap = FlowRunnerTestUtil
    .prepareProject(this.project, dir, this.logger, this.workingDir);
"
Again, I would prefer an incremental approach, especially when dealing with legacy code which I don't understand well enough.  I would welcome any suggestion on how to simplify this logic further. 

Regarding the overhead, I tried the approach myself see 
#1147 with the help of intellij and your quick review, it didn't take much longer to make two PRs than one.
I understand the overhead. Again I'd like to optimize for reading than writing. After all, we read the code more often than modifying it. 

Thanks again for your help and understanding!

HappyRay:
Hamcrest for this one is fine. 
I am reading more about assertj and will discuss with the team if we want to standardize on one assertion library for the new code.

juhoautio:
So created https://github.com/azkaban/azkaban/pull/1152 and https://github.com/azkaban/azkaban/pull/1153 to follow.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java
juhoautio: > optimize for reading than writing

Maybe I wasn't clear enough, but I didn't really mean that "easy writing" would mean "less writing". I was only aiming for maximal clarity.

I actually feel that my way here was about _optimizing for reading_: that you don't have a lot of noise in the code, just the essential information in the context.

For this kind of unit test where we test multiple different scenarios and workflows on the same class, I think that less noise is more important than avoiding redirection or inheritance.

I guess my thinking there is that you already know for example that `assertEvents` will check against the "current `eventCollector`" or that `assertStatus` will check against the "current flow" withouth having to every time echo the reference to event collector or flow runner.

HappyRay: "I guess my thinking there is that you already know for example that assertEvents will check against the "current eventCollector" or that assertStatus will check against the "current flow" withouth having to every time echo the reference to event collector or flow runner."
I see.
But the contract wasn't clear to me when I read it and I think I will have the same reaction the next time I read this code.
My personal experience is that my mental overhead goes up when I see code that relies on inheritance typically.

PR #1157, opened 2017-06-04
Test cleanup: simplify prepareProject usage

juhoautio:
Now takes String instead of File.

Discussed at https://github.com/azkaban/azkaban/pull/1151#discussion_r120014278.

Issue #1155, opened 2017-06-04
The test azkaban.execapp.FlowRunnerTest2#testRetryOnFailure is flaky


HappyRay:
The travis build on the master failed with the following error:

https://travis-ci.org/azkaban/azkaban/builds
https://travis-ci.org/azkaban/azkaban/builds/239316403

azkaban.execapp.FlowRunnerTest2 > testRetryOnFailure FAILED

java.lang.AssertionError: Wrong status for [jobb:innerFlow] expected:<SKIPPED> but was:<DISABLED>

    at org.junit.Assert.fail(Assert.java:88)

    at org.junit.Assert.failNotEquals(Assert.java:834)

    at org.junit.Assert.assertEquals(Assert.java:118)

    at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:137)

    at azkaban.execapp.FlowRunnerTest2.testRetryOnFailure(FlowRunnerTest2.java:661)
Looks like the test is still flaky.

related to  #1129

HappyRay:
Another failure when running locally.

Create temp dir
2017/06/04 10:20:37.812 -0700 INFO [main] [JobTypeManager] Loading plugin default job types
2017/06/04 10:20:39.871 -0700 INFO [main] [JmxJobMBeanManager] Initializing azkaban.execapp.jmx.JmxJobMBeanManager
2017/06/04 10:20:39.906 -0700 INFO [main] [FlowRunnerTest2] Adding test1.properties
2017/06/04 10:20:39.907 -0700 INFO [main] [FlowRunnerTest2] Adding test2.properties
2017/06/04 10:20:41.706 -0700 INFO [Thread-1] [jobf] Running execid:101 flow:jobf project:1 version:-1
2017/06/04 10:20:41.721 -0700 INFO [FlowRunner-exec-101] [jobf] Update active reference
2017/06/04 10:20:41.722 -0700 INFO [FlowRunner-exec-101] [jobf] Updating initial flow directory.
2017/06/04 10:20:41.722 -0700 INFO [FlowRunner-exec-101] [jobf] Fetching job and shared properties.
2017/06/04 10:20:41.731 -0700 INFO [FlowRunner-exec-101] [jobf] Starting flows
2017/06/04 10:20:41.790 -0700 INFO [FlowRunner-exec-101] [jobf] Running flow 'jobf'.
ExecutableFlow: jobf RUNNING
ExecutableNode: joba1 QUEUED
[...]

PR #1158, opened 2017-06-04
Test stability: FlowRunnerTest2

juhoautio:
**Let's have https://github.com/azkaban/azkaban/pull/1159 before this, ok?**

Calling `ExecutableFlowBase#isFlowFinished` from the test thread wasn't a good idea.

In rare cases it failed with this exception (see https://github.com/azkaban/azkaban/pull/1157#issuecomment-306055490):

```
azkaban.execapp.FlowRunnerTest2 > testCancel FAILED
at azkaban.executor.ExecutableFlowBase.isFlowFinished(ExecutableFlowBase.java:398)
at azkaban.execapp.FlowRunnerTestBase.lambda$assertThreadShutDown$0(FlowRunnerTestBase.java:32)
at azkaban.execapp.FlowRunnerTestBase$$Lambda$9/1157486615.apply(Unknown Source)
at azkaban.execapp.FlowRunnerTestBase.waitFlowRunner(FlowRunnerTestBase.java:42)
at azkaban.execapp.FlowRunnerTestBase.assertThreadShutDown(FlowRunnerTestBase.java:31)
at azkaban.execapp.FlowRunnerTest2.testCancel(FlowRunnerTest2.java:729)
```

Extract from `ExecutableFlowBase` (this method is actually deleted in this PR):

```java
  /**
   * Only returns true if the status of all finished nodes is true.
   */
  public boolean isFlowFinished() {
    for (final String end : getEndNodes()) {
      final ExecutableNode node = getExecutableNode(end);
      // THIS IS WHERE THE NPE HAPPENED, so node was null
      if (!Status.isStatusFinished(node.getStatus())) {
        return false;
      }
    }

    return true;
  }
```

It must be that the NPE happened because this method was being called from the test main thread, and there is no synchronization on the "end nodes". So it was somehow possible that the test got a name for an end node but then couldn't get the ExecutableNode to match it.

I used `isFlowFinished()` (it existed previously, I didn't add it) in the test because it seemed like a handy way to check that, but seems that it didn't work reliably. And now I realized that this method is not being used at all otherwise, so it can be just removed.

Now checking flow completion in a thread-safe way.

Also delete unused methods in `ExecutableFlowBase`:
- `isFlowFinished`
- `findNextJobsToRun`

**Make `ExecutableNode.status` volatile**
- This gives guarantees especially for unit tests that need to check for statuses in a multi-threaded environment.
- But also `FlowRunner` thread vs. `JobRunner` threads benefit from this, although I haven't been able to prove that there would be any critical issues there.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlowBase.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlowBase.java
@@ -389,62 +389,6 @@ public class ExecutableFlowBase extends ExecutableNode {
     }
   }
 
-  /**
-   * Only returns true if the status of all finished nodes is true.
-   */
-  public boolean isFlowFinished() {
-    for (final String end : getEndNodes()) {
-      final ExecutableNode node = getExecutableNode(end);
-      if (!Status.isStatusFinished(node.getStatus())) {
-        return false;
-      }
-    }
-
-    return true;
-  }
-
-  /**
-   * Finds all jobs which are ready to run. This occurs when all of its
-   * dependency nodes are finished running.
-   *
-   * It will also return any subflow that has been completed such that the
-   * FlowRunner can properly handle them.
-   */
-  public List<ExecutableNode> findNextJobsToRun() {
-    final ArrayList<ExecutableNode> jobsToRun = new ArrayList<>();
-
-    if (isFlowFinished() && !Status.isStatusFinished(getStatus())) {
-      jobsToRun.add(this);
-    } else {
-      nodeloop:
-      for (final ExecutableNode node : this.executableNodes.values()) {
-        if (Status.isStatusFinished(node.getStatus())) {
-          continue;
-        }
-
-        if ((node instanceof ExecutableFlowBase)
-            && Status.isStatusRunning(node.getStatus())) {
-          // If the flow is still running, we traverse into the flow
-          jobsToRun.addAll(((ExecutableFlowBase) node).findNextJobsToRun());
-        } else if (Status.isStatusRunning(node.getStatus())) {
-          continue;
-        } else {
-          for (final String dependency : node.getInNodes()) {
-            // We find that the outer-loop is unfinished.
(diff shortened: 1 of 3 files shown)


HappyRay:
#1155

juhoautio:
Regardless of the build failure, this change is for the better, because it narrows down the ways of checking statuses concurrently.

HappyRay:
Could you share more details on the problem and how the fix works?

HappyRay:
It seems to me that both the production code and tests are complex given all the trouble we have. Can you think of ways to simplfy? e.g. can we do without mutliple threads? I understand this may require a large refactoring...

juhoautio:
> Could you share more details on the problem and how the fix works?

Added to PR description.

juhoautio:
> Can you think of ways to simplfy? e.g. can we do without mutliple threads?

Sure there are different ways to simplify and make this more robust. Maybe the simplest way would be to keep the threads as they are but use some concurrent utilities to make things easier to manage.

HappyRay:
Please see another new failure.

Thanks for the explanation.
Why is the new code thread safe?

juhoautio:
> Why is the new code thread safe?

Because it just checks the flow status, not statuses of every job in the flow. And this is enough to check if the flow has finished.

juhoautio:
Added commit: "Make ExecutableNode.status volatile". Also added to PR description. After that change I just ran `FlowRunnerTest2` 1000 times in a row locally and didn't get a single test failure.

juhoautio:
Same thing with `FlowRunnerTest` – ran 1000 times locally without failures (after making `ExecutableNode.status` volatile).

juhoautio:
Now that I merged from master there are some of those `@Ignore`s added. Once this has been merged I can create a new PR to remove the `@Ignore`s.

juhoautio:
Made a separate PR for that volatile status field: https://github.com/azkaban/azkaban/pull/1159

HappyRay:
I see. 
The getStatus method is only thread-safe after the status field is made volatile. 

LGTM.

PR #1160, opened 2017-06-04
DON'T MERGE – test stability on Travis

juhoautio:
DO NOT MERGE! Run FlowRunnerTest & FlowRunnerTest2 ~~100x~~ 50x in a row to check if tests are stable or not. This is with the fixes proposed in https://github.com/azkaban/azkaban/pull/1158. Also removed `@Ignore`s in `FlowRunnerTest2` to cover it all.

PR #1162, opened 2017-06-04
Test stability: FlowRunnerTest2

juhoautio:
See also https://github.com/azkaban/azkaban/pull/1160.

Clear previous test jobs so that `.succeedJob()` is called on the retried job instance.
So now we can remove all `@Ignore`s.

----

How does this help?

When a job run is started, the created job type instance gets added into `InteractiveTestJob`'s map. If a job is retried, another instance is created, but for `InteractiveTestJob` the key=jobname is the same.

My commit message says:
> Clear previous test jobs so that .succeedJob() is called on the retried job instance.

So occasionally it happened that `InteractiveTestJob.getTestJob("jobb:innerJobB").succeedJob` was targeted at the original failed instance. After that the retry attempt instance was added into `InteractiveTestJob`'s map, but `succeedJob` was never called on it, so it was left in `RUNNING` state.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
@@ -17,6 +17,7 @@
 package azkaban.executor;
 
 import static azkaban.flow.CommonJobProperties.JOB_ATTEMPT;
+import static org.junit.Assert.assertNotNull;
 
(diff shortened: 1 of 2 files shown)


HappyRay:
I will need more time to read the test code to understand the behavior. 
Given the test results you have, I merged it first so that I don't delay you further from going to sleep.

Thanks!

PR #1163, opened 2017-06-05
Change InteractiveTestJob.testJobs back to private

juhoautio:
As discussed in https://github.com/azkaban/azkaban/pull/1129#pullrequestreview-41944642, this doesn't need to be `public`.


HappyRay:
Thanks

PR #1165, opened 2017-06-05
[DO NOT MERGE UNTIL DB IS UPDATED] Fetch active/queued flows from execution_flows table only.

jamiesjc:
1) Currently active/queued flows are fetched from the join result of both execution_flows and active_executing_flows tables. There is some inconsistency of the flow info between the two tables. We are planning to deprecate active_executing_flows table finally. This will pave the way for the future change of removing runningFlows and queuedFlows cache from web server. 
2) In order to improve the DB performance after this change, we need to create index on status for execution_flows table. See PR https://github.com/azkaban/azkaban/pull/1132 for the DB update change.
3) In current code, when dispatching flows in single executor mode, if the dispatch fails, flow reference is removed from active_executing_flows table, but flow status is not updated in execution_flows table, which will cause the inconsistency of the flow info in DB. So we need to fail the flow if it's not dispatched successfully and update its status in DB just as what we are doing in multi executor mode. 
4) Currently, updateTime in ExecutionReference is not used. (Only inserted into DB when submitting the flow and updated when executor starts to run the flow). Will deprecate together with the entire active_executing_flows table in the future.
4) Testing:
Used local mysql db to test fetching active/queued flows covering both valid and invalid cases in unit tests.
Verified the sql query in azkaban DBs, also did some cleanup on the remaining inconsistent DB info.
Tested the changes in testing cluster and verified web server can loadRunningFlows correctly when it starts.


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1025,8 +1025,11 @@ public String submitExecutableFlow(ExecutableFlow exflow, String userId)
           try {
             dispatch(reference, exflow, choosenExecutor);
           } catch (ExecutorManagerException e) {
-            executorLoader.removeActiveExecutableReference(reference
-              .getExecId());
+            // When flow dispatch fails, should update the flow status
+            // to FAILED in execution_flows DB table as well. Currently
+            // this logic is only implemented in multiExecutorMode but
```
HappyRay: Do you mean the single executor mode will be broken after this change?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1308,27 +1310,22 @@ public String handle(ResultSet rs) throws SQLException {
     ResultSetHandler<List<Pair<ExecutionReference, ExecutableFlow>>> {
     // Select queued unassigned flows
     private static String FETCH_QUEUED_EXECUTABLE_FLOW =
-      "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, "
-        + " ax.update_time axUpdateTime FROM execution_flows ex"
-        + " INNER JOIN"
-        + " active_executing_flows ax ON ex.exec_id = ax.exec_id"
-        + " Where ex.executor_id is NULL";
+        "SELECT exec_id, enc_type, flow_data FROM execution_flows"
+            + " Where executor_id is NULL AND status = ?";
```
HappyRay: Have you considered adding the status here instead of having the caller passing in a value? 
I think it will be easier to read and less likely for callers to pass in the wrong value. There is no need for different callers to pass in a different value. Right?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1369,19 +1363,17 @@ public String handle(ResultSet rs) throws SQLException {
...
       "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-        + "et.port port, ax.update_time axUpdateTime, et.id executorId, et.active executorStatus"
+        + "et.port port, et.id executorId, et.active executorStatus"
         + " FROM execution_flows ex"
         + " INNER JOIN "
-        + " active_executing_flows ax ON ex.exec_id = ax.exec_id"
-        + " INNER JOIN "
-        + " executors et ON ex.executor_id = et.id";
+        + " executors et ON ex.executor_id = et.id"
+        + " Where ex.status NOT IN (?, ?, ?)";
```
HappyRay: Similar comment as the other SQL query above.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1435,13 +1423,11 @@ public String handle(ResultSet rs) throws SQLException {
...
         "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-            + "et.port port, ax.update_time axUpdateTime, et.id executorId, et.active executorStatus"
+            + "et.port port, et.id executorId, et.active executorStatus"
             + " FROM execution_flows ex"
             + " INNER JOIN "
-            + " active_executing_flows ax ON ex.exec_id = ax.exec_id"
-            + " INNER JOIN "
             + " executors et ON ex.executor_id = et.id"
-            + " WHERE ax.exec_id = ?";
+            + " Where ex.exec_id = ? AND ex.status NOT IN (?, ?, ?)";
```
HappyRay: Similar comment as the other SQL query above.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -467,18 +467,17 @@ public void testFetchNoQueuedFlows() throws ExecutorManagerException,
     int port = 12345;
     Executor executor = loader.addExecutor(host, port);
 
+    // When flow is assigned an executor, it is no longer in queued flows
```
HappyRay: When a flow ....

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: Yes, currently the inconsistency issue already exists between active_executing_flows and executionf_flows table. After deprecating active_executing_flows table, this issue will be exposed in single executor mode when dispatch fails. So it is necessary to set the flow status to FAILED in execution_flows table when dispatch fails.

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
```diff
@@ -19,6 +19,7 @@
 public class ExecutionReference {
   private final int execId;
   private Executor executor;
+  //Todo jamiesjc: deprecate updateTime in ExecutionReference class gradually.
   private long updateTime;
```
chengren311: should we consider add deprecated annotation here?

chengren311: also can we remove setUpdateTime if it's no longer used?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
chengren311: why not use ex.status in (active, running, ....)? should be more straightforward.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
chengren311: same here

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
HappyRay: Is there a possibility to remove update time in a separate commit?

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
HappyRay: I have the same question.

HappyRay:
Thanks for the detailed change description.
Could you describe your test strategy as well?

azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
jamiesjc: In this PR, I'm only making the change not to read updateTime from active_executing_flows table when fetching active/queued flows. Plan to do the actual removal in a separate PR.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
jamiesjc: Will fix it, thanks!

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
jamiesjc: Because currently there are 13 statuses in total. 3 of them are finished statuses. So it's more concise in the query to just exclude the finished statuses.

HappyRay: But not all 13 apply to flow status. Right?
Not listing them explicitly may expose us to problems when an incorrect status somehow shows up.
Having it listed here also serves as documentation. Maybe have a central place to return such subsets?

jamiesjc:
In unit tests, I used local mysql db to test fetching active/queued flows covering both valid and invalid cases.
I verified the sql query in azkaban DBs, also did some cleanup on the remaining inconsistent DB info.
Plan to put the changes in testing cluster and see if web server can loadRunningFlows and loadQueuedFlows correctly when it starts.

HappyRay:
Thanks. Please add a testing section to the change description when you merge.

PR #1166, opened 2017-06-05
JobRunner uploadExecutableNode improvement

juhoautio:
I would suggest having https://github.com/azkaban/azkaban/pull/1167 instead of this, but this is of course a smaller change..

Include stack trace in error logging.

Based on this discussion / request by @chengren311: https://github.com/azkaban/azkaban/pull/1148#pullrequestreview-41936280


HappyRay:
#1167 LGTM.

@chengren311 what do you think?

PR #1167, opened 2017-06-05
JobRunner uploadExecutableNode improvement

juhoautio:
If initial upload fails:
- Don't run the Job instance, just report failed status
- At the end, try to update the node in DB any way (as FAILED), because _maybe_ it was created regardless of the caught exception
- Include stack trace in error logging

Based on this discussion / request by @chengren311 & @HappyRay: https://github.com/azkaban/azkaban/pull/1148#pullrequestreview-41936280


juhoautio:
Resolved conflicts.

HappyRay:
I changed the change description a bit when merging, please check.

juhoautio:
Good catch 👌

PR #1169, opened 2017-06-05
fix Triggermanager Guice bug

kunkun-tang:
I intended to guicify `TriggerManager` as part of last [PR](https://github.com/azkaban/azkaban/pull/1110/files#diff-f916ab7306f59ff720753937cf9a9ff1L62). I just realized that it was an incomplete Guice (mistake) without changes in AzkabanWebServer code. As a result, this PR made the fix and completely guicify `TriggerManager`.

Unless this PR gets merged, we will keep using outdated `JDBCTriggerLoader` code.


azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -194,8 +194,8 @@ public AzkabanWebServer(Server server, Props props) throws Exception {
     // TODO remove hack. Move injection to constructor
     executorManager = SERVICE_PROVIDER.getInstance(ExecutorManager.class);
```
suvodeep-pyne: At some point, we would like to get rid of these custom injections as well. Looks Ok for now.

PR #1170, opened 2017-06-06
Fix MutableConstantField ErrorProne warning

HappyRay:
azkaban/azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java:27:
warning: [MutableConstantField] Constant field declarations should use
the immutable type (such as ImmutableList) instead of the general
collection interface type (such as List)
  private static final Set<String> MEM_KEYS = ImmutableSet
                          ^
    (see http://errorprone.info/bugpattern/MutableConstantField)


HappyRay:
Thanks @reallocf

Issue #1122, opened 2017-05-26
Azkaban to have KILLING status transition


juhoautio:
We've had quite a lot of trouble with jobs/flows not terminating properly after execution has been killed. In that case it would be really helpful to see that the flow is still in "killing" state instead of "killed". Because if you kill an execution and expect it to have been terminated, but it hasn't, next day's scheduled run can be blocked by some job threads that were somehow hung (we mostly run everything in pipeline level 1 mode). Of course the real problem is somewhere else then (if a flow doesn't terminate properly). But this would at least give better visibility and save from most of the cases because if you notice that flow refuses to die you can restart the executor node.

Now it's just: RUNNING -> KILLED, although in reality some jobs can be still running, waiting for thread to exit, and what not.

So, add a new lifecycle phase so that it's: RUNNING -> KILLING -> KILLED.

I think it might be good if it's allowed to try killing the flow again while it's already being killed.. or if it makes more sense, while killing, azkaban could keep calling cancel() again on jobs that refuse to terminate, because it's possible that the first cancellation request failed for some reason.

From the perspective of blocking other jobs in pipeline mode, KILLING status should block newer executions in the same way that RUNNING status does.

juhoautio:
@HappyRay & co., any chance to get some early feedback based on https://github.com/juhoautio/azkaban/pull/2? This is basically a working implementation, but it depends on some other PRs that I'd like to get in first. Then I will make a cleaned up & rebased PR based on https://github.com/juhoautio/azkaban/pull/2.

HappyRay:
I am not against the idea of having this state given that administrators may choose to allow a relatively long timeout for a job to be terminated gracefully. 

As you suggested, we need to find a way to reliably terminate jobs and if we can't, notify the users. 

The current ProcessJob based jobs should force kill the job process see
azkaban.jobExecutor.ProcessJob#cancel

Do your jobs overwrite this method?

We typically avoid restarting executors since this will affect existing running flows. We start a new executor each time and leave the older executors running.

HappyRay:
FYI since you use pipeline option, here is a related issue we discovered recently while reading the code. #1125 "The pipeline concurrent option is not reliable in the multiple executor mode"

juhoautio:
[...]

PR #1172, opened 2017-06-06
New status: KILLING

juhoautio:
Implements https://github.com/azkaban/azkaban/issues/1122: _Azkaban to have KILLING status transition_.

- New status transition after `kill()` is called: `.. -> KILLING -> KILLED`.
- Don't set the status to `KILLED` before all job threads have exited.

Execution graph:

killing-1-graph

Execution Job List:

killing-2-job list

History:

killing-3-history

----

After this change, I would like to rename these private booleans:
- `JobRunner: killed -> cancelRequested`
- `FlowRunner: flowKilled -> killRequested`

Otherwise it's quite easy to assume that those `killed` booleans mean already having `KILLED` status, when it can also be `KILLING`.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -16,8 +16,8 @@
 
 package azkaban.executor;
 
-import java.util.HashMap;
-import java.util.Map;
+import com.google.common.collect.ImmutableMap;
+import java.util.Arrays;
 
 public enum Status {
   READY(10),
@@ -37,13 +37,8 @@ public enum Status {
   // status is TINYINT and in H2 DB the possible values are: -128 to 127
   // so trying to store CANCELLED in H2 fails at the moment
 
-  private static final Map<Integer, Status> numValMap = new HashMap<>();
-
-  static {
-    for (final Status status : Status.values()) {
-      numValMap.put(status.getNumVal(), status);
-    }
-  }
+  private static final ImmutableMap<Integer, Status> numValMap = Arrays.stream(Status.values())
+      .collect(ImmutableMap.toImmutableMap(status -> status.getNumVal(), status -> status));
 
   private final int numVal;
 
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -16,12 +16,16 @@
 
 package azkaban.executor;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public enum Status {
   READY(10),
   PREPARING(20),
   RUNNING(30),
   PAUSED(40),
   SUCCEEDED(50),
+  KILLING(55),
   KILLED(60),
   FAILED(70),
   FAILED_FINISHING(80),
@@ -30,6 +34,16 @@ public enum Status {
   QUEUED(110),
   FAILED_SUCCEEDED(120),
   CANCELLED(130);
+  // status is TINYINT and in H2 DB the possible values are: -128 to 127
+  // so trying to store CANCELLED in H2 fails at the moment
+
+  private static final Map<Integer, Status> numValMap = new HashMap<>();
+
+  static {
+    for (Status status : Status.values()) {
+      numValMap.put(status.getNumVal(), status);
+    }
+  }
 
   private final int numVal;
 
@@ -38,36 +52,10 @@ public enum Status {
   }
 
   public static Status fromInteger(final int x) {
-    switch (x) {
-      case 10:
-        return READY;
-      case 20:
-        return PREPARING;
-      case 30:
-        return RUNNING;
-      case 40:
-        return PAUSED;
-      case 50:
-        return SUCCEEDED;
-      case 60:
-        return KILLED;
-      case 70:
-        return FAILED;
-      case 80:
-        return FAILED_FINISHING;
-      case 90:
-        return SKIPPED;
-      case 100:
-        return DISABLED;
-      case 110:
-        return QUEUED;
-      case 120:
-        return FAILED_SUCCEEDED;
-      case 130:
-        return CANCELLED;
-      default:
-        return READY;
+    if (numValMap.containsKey (x)) {
+      return numValMap.get(x);
     }
+    return READY;
   }
 
   public static boolean isStatusFinished(final Status status) {
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -47,10 +47,7 @@ public enum Status {
   }
 
   public static Status fromInteger(final int x) {
-    if (numValMap.containsKey(x)) {
-      return numValMap.get(x);
-    }
-    return READY;
+    return numValMap.getOrDefault(x, READY);
   }
 
   public static boolean isStatusFinished(final Status status) {
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -40,7 +40,7 @@ public enum Status {
   private static final Map<Integer, Status> numValMap = new HashMap<>();
 
   static {
-    for (Status status : Status.values()) {
+    for (final Status status : Status.values()) {
       numValMap.put(status.getNumVal(), status);
     }
   }
@@ -52,7 +52,7 @@ public enum Status {
   }
 
   public static Status fromInteger(final int x) {
-    if (numValMap.containsKey (x)) {
+    if (numValMap.containsKey(x)) {
       return numValMap.get(x);
     }
     return READY;
--- a/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
@@ -63,6 +63,12 @@ public class InteractiveTestJob extends AbstractProcessJob {
     testJobs.clear();
   }
 
+  public static void clearTestJobs(final String... names) {
+    for (final String name : names) {
+      assertNotNull(testJobs.remove(name));
+    }
+  }
+
   @Override
   public void run() throws Exception {
     final String nestedFlowPath =
@@ -154,12 +160,6 @@ public class InteractiveTestJob extends AbstractProcessJob {
     info("Killing job");
     if (!this.ignoreCancel) {
       failJob();
-    }
-  }
-
-  public static void clearTestJobs(final String... names) {
-    for (String name : names) {
-      assertNotNull(testJobs.remove(name));
     }
   }
 }
--- a/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
@@ -33,6 +33,7 @@ public class InteractiveTestJob extends AbstractProcessJob {
   private Props generatedProperties = new Props();
   private boolean isWaiting = true;
   private boolean succeed = true;
+  private boolean ignoreCancel = false;
 
   public InteractiveTestJob(final String jobId, final Props sysProps, final Props jobProps,
       final Logger log) {
@@ -137,6 +138,12 @@ public class InteractiveTestJob extends AbstractProcessJob {
     }
   }
 
+  public void ignoreCancel() {
+    synchronized (this) {
+      this.ignoreCancel = true;
+    }
+  }
+
   @Override
   public Props getJobGeneratedProperties() {
     return this.generatedProperties;
@@ -145,7 +152,9 @@ public class InteractiveTestJob extends AbstractProcessJob {
   @Override
   public void cancel() throws InterruptedException {
     info("Killing job");
-    failJob();
+    if (!this.ignoreCancel) {
+      failJob();
+    }
   }
 
   public static void clearTestJobs(final String... names) {
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -112,13 +112,13 @@ public class FlowRunner extends EventHandler implements Runnable {
   private String jobLogFileSize = "5MB";
   private int jobLogNumFiles = 4;
 
-  private boolean flowPaused = false;
-  private boolean flowFailed = false;
-  private boolean flowFinished = false;
-  private boolean flowKilled = false;
+  private volatile boolean flowPaused = false;
+  private volatile boolean flowFailed = false;
+  private volatile boolean flowFinished = false;
+  private volatile boolean flowKilled = false;
 
   // The following is state that will trigger a retry of all failed jobs
-  private boolean retryFailedJobs = false;
+  private volatile boolean retryFailedJobs = false;
 
   /**
    * Constructor. This will create its own ExecutorService for thread pools
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -546,7 +546,7 @@ public class FlowRunner extends EventHandler implements Runnable {
   }
 
   private void propagateStatus(final ExecutableFlowBase base, final Status status) {
-    if (!Status.isStatusFinished(base.getStatus())) {
+    if (!Status.isStatusFinished(base.getStatus()) && base.getStatus() != Status.KILLING) {
       this.logger.info("Setting " + base.getNestedId() + " to " + status);
       base.setStatus(status);
       if (base.getParentFlow() != null) {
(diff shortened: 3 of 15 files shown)


HappyRay:
Thanks.

Given other priorities, I am afraid it will take me a while to review this change. @chengren311 do you have the bandwidth to review?

cc @ameyamk

chengren311:
@HappyRay I'm reviewing this change.

HappyRay:
Thanks @chengren311

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -30,6 +34,16 @@
   QUEUED(110),
   FAILED_SUCCEEDED(120),
   CANCELLED(130);
+  // status is TINYINT and in H2 DB the possible values are: -128 to 127
+  // so trying to store CANCELLED in H2 fails at the moment
+
+  private static final Map<Integer, Status> numValMap = new HashMap<>();
```
chengren311: what do you think of using ImmutableMap https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableMap.html?

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -38,36 +52,10 @@
...
-        return DISABLED;
-      case 110:
-        return QUEUED;
-      case 120:
-        return FAILED_SUCCEEDED;
-      case 130:
-        return CANCELLED;
-      default:
-        return READY;
+    if (numValMap.containsKey (x)) {
```
chengren311: this is nice! thanks!

azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
```diff
@@ -137,6 +138,12 @@ public void succeedJob(final Props generatedProperties) {
     }
   }
 
+  public void ignoreCancel() {
+    synchronized (this) {
+      this.ignoreCancel = true;
```
chengren311: again i think just making this boolean volatile or atomicboolean is sufficient, what do you think?

juhoautio: Maybe, but I was following the pattern used in this class. Let's have it like this and refactor all those flags to volatile in a later commit?

juhoautio: Actually the difference with the other flags is that when they're toggled there's also `this.notify();` so it's handy to cover both with the sync block..

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -602,9 +604,14 @@ private void finalizeFlow(final ExecutableFlowBase flow) {
     final long durationSec = (flow.getEndTime() - flow.getStartTime()) / 1000;
     switch (flow.getStatus()) {
       case FAILED_FINISHING:
-        this.logger.info("Setting flow '" + id + "' status to FAILED in "
+        Status status = isKilled() ? Status.KILLED : Status.FAILED;
+        this.logger.info("Setting flow '" + id + "' status to "+status+" in "
```
chengren311: did you enable save plugin to auto reformat code? it should auto add new space before/after + here.

azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
juhoautio: Want me to change `ignoreCancel` to volatile? Somehow I would be inclined to prefer the sync block because that it can be maybe easier to understand this class as a whole? When you don't need to pay special attention to that special volatile field and how it differs from those other flags.. Well, that refactoring to change all of them is still possible. Then it would be just the waits and the notifys that need to be inside a sync block.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
juhoautio: Nope, I didn't have it activated. I'm a bit afraid to activate save plugin because not all code is yet formatted (or is it now?), and if a previously unformatted class gets auto-formatted when I have also some manual changes in that file, I'm in trouble :) Usually I take care of formatting any way, this is plain mistake, sorry..

azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
chengren311: ok, i see. yeah you are right, let's keep it for now to be consistent and refactor it in a later commit.

azkaban-common/src/main/java/azkaban/executor/Status.java
juhoautio: I pushed separately here so that you can take a look:
- truly immutable: https://github.com/juhoautio/azkaban/commit/23c9cdd43e6f74ecefcd3e246dad148633261abe
- a bit of cheating but eventually immutable: https://github.com/juhoautio/azkaban/commit/bf0a7ee646270b67d0ba361bfec08dd54b41aa8d

Do you think it's worth adding this much extra code to make it immutable?

It's pretty much closed down like it is any way..

juhoautio: There must be some way to also do this with lesser code, using streams/collector.. Should I try that? Readability may be not so good on that?

juhoautio: Here's the way using streams – just 2 lines: https://github.com/juhoautio/azkaban/commit/b90c4969dacfae394604ce8e6384a9fb39cfa537

Please let me know which way you prefer, maybe even some 4th way? :)

juhoautio:
Thank you @chengren311! Fixed the bad formatting. Waiting for your thoughts on that ImmutableMap.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
juhoautio: I just tried and at least `ExecutorManager.java` (changed in this commit) would be totally destroyed by Save Action Plugin :D So it must be that not all files in master have been reformatted just yet.

juhoautio:
@chengren311, could I have your verdict on https://github.com/azkaban/azkaban/pull/1172#discussion_r120729789 ?

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -30,6 +34,16 @@
   QUEUED(110),
   FAILED_SUCCEEDED(120),
   CANCELLED(130);
+  // status is TINYINT and in H2 DB the possible values are: -128 to 127
```
HappyRay: Thanks for reporting the documenting this issue. 
Any suggestion on how to fix it?
I hope to use h2 DB more for unit testing.

HappyRay: And I am not sure why the status can't be just 1, 2, 3 etc

juhoautio: > Any suggestion on how to fix it?

I don't know, that might not be so trivial..

> why the status can't be just 1, 2, 3 etc

Yeah :D

juhoautio: How about altering the column type from TINYINT to INT? That's the simplest fix I can think of.

HappyRay: Yes, that's what I thought too. I was hoping someone smarter than me can come up with a better solution. :)

juhoautio:
Changed to `ImmutableMap`.. @chengren311: I've resolved all comments so far.

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -38,36 +47,10 @@
...
-        return DISABLED;
-      case 110:
-        return QUEUED;
-      case 120:
-        return FAILED_SUCCEEDED;
-      case 130:
-        return CANCELLED;
-      default:
-        return READY;
+    if (numValMap.containsKey(x)) {
```
chengren311: one line might be simpler:
return numValMap.containsKey(x) ? numValMap.get(x) : READY;

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -474,7 +474,8 @@ private boolean progressGraph() throws IOException {
     boolean jobsRun = false;
     for (final ExecutableNode node : nodesToCheck) {
       if (Status.isStatusFinished(node.getStatus())
```
chengren311: maybe consider putting all of the three conditions into a method called notReadyToRun() ?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -753,9 +753,10 @@ private void insertJobMetadata() {
   }
 
   private Status runJob() {
-    Status finalStatus = this.node.getStatus();
+    Status finalStatus;
     try {
       this.job.run();
+      finalStatus = this.node.getStatus();
```
chengren311: thanks for fixing this as well :)

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -778,7 +779,7 @@ private Status runJob() {
     }
 
     // If the job is still running, set the status to Success.
-    if (!Status.isStatusFinished(finalStatus)) {
+    if (!Status.isStatusFinished(finalStatus) && finalStatus != Status.KILLING) {
```
chengren311: maybe change the comment accordingly

azkaban-common/src/main/java/azkaban/executor/Status.java
juhoautio: If that change is done _only_ for h2, that should be fine? How do the h2 tables get created?

azkaban-common/src/main/java/azkaban/executor/Status.java
juhoautio: Yeah we can simply use `Map#getOrDefault` because ImmutableMap doesn't even allow null values.

----

`Map#getOrDefault`:

```java
  @Override
  public final V getOrDefault(@Nullable Object key, @Nullable V defaultValue) {
    V result = get(key);
    return (result != null) ? result : defaultValue;
  }
```

juhoautio:
@chengren311: now fixed per comments, rebased & resolved conflicts.

azkaban-common/src/main/java/azkaban/executor/Status.java
HappyRay: So far there is no need to create separate sql DDLs for h2 yet. 
I think it is simpler to change the schema for both DBs.

juhoautio: I was thinking "additional" DDL only. So the column would be altered for h2 when it's initialized for the first time. I would avoid any migrations required on existing installations on MySQL databases as long as possible.

juhoautio:
@chengren311 anything else?

juhoautio:
Hm.. bump? :)

juhoautio:
@chengren311 (same story as for the stuck pipeline fix..) I'd like to have this in use ASAP. What do you think, will you be able to review and get this merged on this side soon? I'd rather do it that way but can of course apply on our fork if not possible to prioritize here.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -591,9 +597,15 @@ private void finalizeFlow(final ExecutableFlowBase flow) {
     final long durationSec = (flow.getEndTime() - flow.getStartTime()) / 1000;
     switch (flow.getStatus()) {
       case FAILED_FINISHING:
-        this.logger.info("Setting flow '" + id + "' status to FAILED in "
+        final Status status = isKilled() ? Status.KILLED : Status.FAILED;
```
chengren311: is this change related with killing status or it's just a low hanging fruit of bug fix?

chengren311:
@juhoautio sorry for the delayed response and thanks for your active contribution! I am keeping reviewing this PR and will work with you on getting this merged ASAP. Thanks!

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
juhoautio: Sorry, that was a mistake. There's something more that's needed because of KILLING status, but this is not the correct way. As you can see some assertions had been also changed in the FlowRunnerTest2, but I hadn't really thought about those. I should have been more careful.

juhoautio:
Added commit:

Fix KILLED vs. FAILED status propagation

Bring back original assertions in FlowRunnerTest2.

azkaban-web-server/src/web/js/azkaban/view/exflow.js
```diff
@@ -447,6 +449,10 @@ var updaterFunction = function() {
 			// 2 min updates
 			setTimeout(function() {updaterFunction();}, 2*60*1000);
 		}
+		else if (data.status == "KILLING") {
+			// 20 s updates
+			setTimeout(function() {updaterFunction();}, 30*1000);
```
chengren311: so this is interval between each updaterFunction? can you help me understand better why it's 30 secs? also comment says 20s

chengren311:
@juhoautio also maybe i missed something from your PR, but where do we set job status from killing to killed?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -595,6 +601,11 @@ private void finalizeFlow(final ExecutableFlowBase flow) {
             + durationSec + " seconds");
         flow.setStatus(Status.FAILED);
         break;
+      case KILLING:
+        this.logger
+            .info("Setting flow '" + id + "' status to KILLED in " + durationSec + " seconds");
+        flow.setStatus(Status.KILLED);
```
chengren311: so we rely on finalizeFlow to set to killed status from killing. @jamiesjc will anything cause finalizeFlow not being called for a flow?

chengren311:
also how would killing status work with concurrent execution options?

azkaban-web-server/src/web/js/azkaban/view/exflow.js
juhoautio: I'll fix the comment.

I set it to 30 s because if the flow is KILLING we can expect that it will finish ~soon. So have a shorter status polling interval.

juhoautio:
> where do we set job status from killing to killed

This is handled by existing code in [JobRunner#doRun](https://github.com/azkaban/azkaban/blob/32cd294c65853401e50140fa3d8418459740a002/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L592-L599).

It's basically like this:

```java
    runJob();
    if (isKilled()) {
    }
```

juhoautio:
> also how would killing status work with concurrent execution options?

KILLING blocks in a pipeline the same way that RUNNING does. This may actually make the pipelining more robust in some situations. I think that before introducing KILLING status, RemoteFlowWatcher may have seen KILLED and deemed it as finished, allowing a pipelined job to be unblocked, even if the blocking job run had not completely ended yet.

juhoautio:
Tests revealed a race condition that can make the flow not terminate. 

I need to figure it out before this is ready for production.

I managed to reproduce the failure once locally, and for that test run I found this in the logs:

```
INFO [JobRunner-job6-1] [derived-member-data-2] Job job6 finished with status KILLING in 0 seconds
```

That should not happen, as this is logged in a branch of `FlowRunner.JobRunnerEventListener.handleEvent` that handles `event.getType() == Type.JOB_FINISHED`. When `handleEvent` is called, the status should already be KILLED?

juhoautio:
Ha, found & fixed. Added commit: _Fix a race condition that prevents setting KILLING->KILLED (by making JobRunner.killed volatile)_.

chengren311:
@juhoautio thanks for fixing this race condition!
How about FlowRunner's flowKilled? I suspect it has the same risk.

juhoautio:
Great catch.

Added commit:

```
Better thread safety in FlowRunner

Make the boolean flags volatile, because these are accessed from different threads.
```

chengren311:
thanks for working on this

jamiesjc:
@juhoautio Are you still interested in bringing back KILLING status?

juhoautio:
@jamiesjc definitely! Also it's been working without any problems on our deployments after those fixes that were made soon after introducing it.

juhoautio:
@jamiesjc is there some way to proceed right now?

jamiesjc:
@juhoautio Could you submit a new PR? We'll merge it and then deploy.

jamiesjc:
cc @HappyRay @chengren311

juhoautio:
Yes. I guess I'll just revert the previous revert commit and resolve possible conflicts.

juhoautio:
Here: https://github.com/azkaban/azkaban/pull/1509

PR #1179, opened 2017-06-07
Revert "uploadExecutableNode improvement (#1167)"

chengren311:
This would throw exception in case of uploading duplicate keys when two subflows contain jobs of same name. 
This reverts commit ffd9ebd4910cd9e8ac46494c90c14f4d8a2cd7a2.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -584,9 +584,8 @@ public class JobRunner extends EventHandler implements Runnable {
     // Start the node.
     this.node.setStartTime(System.currentTimeMillis());
     Status finalStatus = this.node.getStatus();
-    if (!uploadExecutableNode()) {
-      finalStatus = changeStatus(Status.FAILED);
-    } else if (!errorFound && !isKilled()) {
+    uploadExecutableNode();
+    if (!errorFound && !isKilled()) {
       fireEvent(Event.create(this, Type.JOB_STARTED, new EventData(this.node)));
 
       final Status prepareStatus = prepareJob();
@@ -623,13 +622,11 @@ public class JobRunner extends EventHandler implements Runnable {
     writeStatus();
   }
 
-  private boolean uploadExecutableNode() {
+  private void uploadExecutableNode() {
     try {
       this.loader.uploadExecutableNode(this.node, this.props);
-      return true;
     } catch (final ExecutorManagerException e) {
       this.logger.error("Error writing initial node properties", e);
-      return false;
     }
   }
 


juhoautio:
I don't get it. How i this any different?

juhoautio:
Ok, this changes it so that job is run even if upload fails. If that's normal, then shouldn't be logging an error about it. Better avoid upload when not needed, or even better: do it in a way that only inserts if it doesn't already exist?

chengren311:
@juhoautio 
i'm going to close this PR as we are putting a long-term fix which is to fix the db schema. We will add project id to the primary key set for execution_jobs.

juhoautio:
Isn't project id the same if two subflows (in a flow) contain the same job? So should use full nested id or something?

juhoautio:
But if this is currently affecting all such cases, isn't this a serious problem to have in master even for the time before a proper fix? Or are you doing that really soon now?

chengren311:
@juhoautio flow id i mean, current table's key is (execution id, job id, attempt). 
So it's a rare case since it only happens as an issue for two subflows contain two jobs with same name. And we are asking our DBA to push the schema change for us.

juhoautio:
Ok. For us that's not even a rare case. Would it be possible to reopen this fix & merge to master for the time being?

HappyRay:
@juhoautio Have you deployed a version that includes this change? 
@chengren311 is not familiar with the tests that failed after this change was made.

If this is important to you, do you mind submitting a pull request since you may be more familiar with the tests and this change?

HappyRay:
Do you use subflows extensively?

juhoautio:
Yeah subflows are also used. We're running a fork where #1167 has been reverted, so I haven't ever seen that problem in effect myself.

juhoautio:
> @chengren311 is not familiar with the tests that failed after this change was made.

@HappyRay, could you clarify what tests you mean? Also what is "this change" in your comment?

My understanding is that this isn't about any tests failing – at least not unit tests in this repo. The commit https://github.com/azkaban/azkaban/commit/ffd9ebd4910cd9e8ac46494c90c14f4d8a2cd7a2 was only done to "fail faster" (see the commit message there). I would still like to revert that commit until the job node key in DB is made unique also across sibling sub-flows. So I don't see a need to create a new pull request, because this (#1179), could be reopened for that purpose.

HappyRay:
See the test failures in https://travis-ci.org/azkaban/azkaban/builds/240580566?utm_source=github_status&utm_medium=notification

juhoautio:
I have no idea about those failures :O The only simple explanation I can think of is that Travis was running really slow, so that the maximum waiting time limits are not enough.. I applied this "revert" locally on current master and ran those two failed tests 1000 times in a row – no failures (`FlowRunnerTest2 > testFailedFinishingFailure3` & `FlowRunnerTest > execAndCancel`).

juhoautio:
To continue on that, we can of course make the waiting timeouts longer to add stability in case of unexpected test runner slowness. On the other hand I feel that the current extra 1-2 seconds of waiting is quite much already, when you're actively developing something, so that the tests are failing because your changes are not good yet.

Issue #990, opened 2017-04-17
[Security Flaw] Writing username and password to log file when logging in via the API


devangorder:
Currently the logRequest() function within azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java logs the full string of every API call. This includes the credentials used to login and obtain a session id for further API calls. For example, see below:

```
2017-04-17 14:50:39 INFO  LoginAbstractAzkabanServlet:171 - 0  -  "POST / action=login&username=azkaban_admin&password=PASSWORD HTTP/1.1" not-browser
2017-04-17 14:50:40 INFO  LoginAbstractAzkabanServlet:171 - 0 azkaban_admin "POST /executor ajax=reloadExecutors&session.id=7e5927e9-4c60-7465-9c26-897cd45b25d7 HTTP/1.1" not-browser
2017-04-17 14:50:40 INFO  ExecutorManager:194 - Initializing multi executors from database
```

This is a major security flaw as our admin and automation testing accounts frequently use the API, and we don't want those credentials being logged to the local file system in a plaintext file. Could we implement a solution to scrub the password value from the string before writing login requests to the log file? Does anyone else agree this is a security risk?

ameyamk:
Yes. Can you consider contributing the necessary patch?

devangorder:
I'd love to, but unfortunately my hands are currently tied by legal at my company and I'm not allowed to publish any of my code back yet. From what I can tell, it would be a simple update to that one function within that one file, if anyone else wants to take a stab at it. Once I get approval from legal I'll submit a pull request.

ameyamk:
ok no problem - We will try and fix it. Btw which company are you from? Can you share how you use Azkaban. Would love to get more details. Also - please share if you have any product ideas.

devangorder:
Sure, we currently use Azkaban as our batch scheduler for all our Java Batch processing jobs. In preprod, we have several instances set up, each managing multiple environments. For example, we have one Azkaban instance for all our dev regions, a separate Azkaban instance for our test regions, and then a separate Azkaban instance for our performance test regions. Each of these instances manage multiple environments, sending jobs to the correct executor depending on which project it is launched from. When a job is launched, it compared the project name to values stored in the properties file to determine which executor ID to send it to (internal code enhancement we have developed). This allows us to manage multiple independent environments within the same Azkaban console, without the worry of sending jobs to the wrong executor.

[...]

PR #1180, opened 2017-06-08
Removing sessionId from logger statements

reallocf:
Fixing error reported in LIHADOOP-26732.
sessionId shouldn't be logged anywhere. Looked through related files and couldn't find any other occurrences.
@HappyRay @ameyamk does this look sufficient?


HappyRay:
Could you take a look at #990 too?

PR #1181, opened 2017-06-08
Fixing ReferenceEquality warnings

reallocf:
http://errorprone.info/bugpattern/ReferenceEquality

Any thoughts on unintended consequences?


HappyRay:
Original warning:

"Servlet.java:111: warning: [ReferenceEquality] Comparison using reference equality instead of value equality
          begin == "" ? -1 : DateTimeFormat.forPattern(FILTER_BY_DATE_PATTERN)
                ^
    (see http://errorprone.info/bugpattern/ReferenceEquality)
  Did you mean '"".equals(begin) ? -1 : 
"

Could you add this info to the description when you merge?

The suggested fix above seems to be more concise.

Also see
https://stackoverflow.com/questions/7520432/what-is-the-difference-between-vs-equals-in-java

PR #1182, opened 2017-06-08
Deprecate schedule calendar page and related ajax calls

jamiesjc:
Several reasons for this change:
1. Schedule calendar page is not exposed and used by anyone. 
2. When loading schedule calendar page, it will trigger both loadFlow and loadHistory ajax calls. They will create .cache file and store in web server /cache folder, which will take unnecessary disk space. And the .cache files are not cleaned up properly. 
3. This is part of the steps to simplify updaterThread logic.


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1337,14 +1336,6 @@ public void run() {
               }
             }
 
-            for (ExecutableFlow flow : finishedFlows) {
-              if (flow.getScheduleId() >= 0
-                  && flow.getStatus() == Status.SUCCEEDED) {
-                ScheduleStatisticManager.invalidateCache(flow.getScheduleId(),
-                    cacheDir);
```
jamiesjc: In updaterThread, the original intention of this code was to delete the .cache file for the flow when it succeeded. However, there was some bug and flow.getScheduleId() was always -1 and never updated correctly. So .cache file was never deleted once created. So I removed this code and simplified the logic in updaterThread.

kunkun-tang: Great Explanation!

HappyRay:
Great to see "dead code" gone.

Thanks!

wyukawa:
@jamiesjc 

Oh, I use loadFlow API because I want to get all schedule data.

https://github.com/wyukawa/eboshi/blob/master/eboshi/schedule.py#L18

Is there alternative API?

wyukawa:
send pull request https://github.com/azkaban/azkaban/pull/1219

PR #1184, opened 2017-06-08
Switch from hamcrest to assertj library for EventCollectorListener

HappyRay:
assertj library has a few advantages.
see


juhoautio:
I tested what kinda message it prints in case of a failure, this:

```
org.junit.ComparisonFailure: 
Expected :[FLOW_STARTED]
Actual   :[FLOW_STARTED, FLOW_FINISHED]
```

It's cool.

PR #1185, opened 2017-06-08
Revert "uploadExecutableNode improvement (#1167)"

juhoautio:
(Same as the closed PR https://github.com/azkaban/azkaban/pull/1179)

This reverts commit ffd9ebd.

The problem is not in the "fails fast" strategy per se, but there's another bug that would cause a job in a subflow to fail instead of running it, when two subflows contain jobs of the same name. It's because current DB record key for executable node is (execution id, job id, attempt). So the problem only happens when two subflows contain two jobs with the same name.


juhoautio:
So, how about this? Quite clear that the occasional test failures were not caused by this change.

HappyRay:
@chengren311

PR #1187, opened 2017-06-08
Remove hamcrest assertion library

HappyRay:
We will standardize on assertj library, which has a few advantages.
see

This will speed up the Travis build process a bit.


HappyRay:
No, it is not.

PR #1188, opened 2017-06-08
refactor MysqlDataSource

kunkun-tang:
MysqlDataSource used to be a singleton class. However, we are trying to apply guicification to it. As one step of completing gucification to this class, we do a refactor. The reason we can not entirely gucify today:
* The class depends on Props, which is in `azkaban-common` module. `Azkaban-db` can not call `Props` today.

### DIFF
--- a/azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
+++ b/azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
@@ -25,10 +25,7 @@ public class MySQLDataSource extends AzkabanDataSource {
 
   private static final Logger logger = Logger.getLogger(MySQLDataSource.class);
 
-  private static volatile MySQLDataSource instance = null;
-
-  // TODO kunkun-tang: have guice inject working here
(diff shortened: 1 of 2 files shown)


azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -25,10 +25,7 @@
 
   private static final Logger logger = Logger.getLogger(MySQLDataSource.class);
 
-  private static volatile MySQLDataSource instance = null;
-
-  // TODO kunkun-tang: have guice inject working here
-  private MySQLDataSource(final String host, final int port, final String dbName,
+  public MySQLDataSource(final String host, final int port, final String dbName,
```
HappyRay: This class is indirectly created by Guice already. Right? 
And it doesn't require Guice's assistance to create an instance. So I think what you have is fine.

kunkun-tang: Yeah, probably.

PR #1189, opened 2017-06-08
Guicing MetricsManager, CommonMetrics, and WebMetrics and handling related changes

reallocf:
Initial motivation was to remove ImmutableEnumChecker warnings. Worked with @kunkun-tang to determine that Guicing the offending classes was the best way to proceed.

MetricsManager, CommonMetrics, and WebMetrics had to be guicified together in order to handle tests that used both CommonMetrics/MetricsManager or WebMetrics/MetricsManager.

CommonMetricsTest and WebMetricsTest both required refactoring to handle the new guicier versions.

Added utility method to MetricsTestUtility that initializes a new version of MetricsManager that's used in multiple tests. This occurred in both CommonMetricsTest and WebMetricsTest as well as LoginAbstractAzkabanServletTest.

I'm hoping these testing changes/guicing overall will allow for more parallel testing. Additional adjustment of testing classes is likely necessary to make that happen.

Changes to all other files replaced the old singleton enum syntax with guice syntax.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -123,7 +123,7 @@ public class AzkabanCommonModule extends AbstractModule {
     final String password = props.getString("mysql.password");
     final int numConnections = props.getInt("mysql.numconnections");
 
-    return MySQLDataSource.getInstance(host, port, database, user, password, numConnections);
+    return new MySQLDataSource(host, port, database, user, password, numConnections);
   }
 
   @Inject
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -30,6 +30,7 @@ import azkaban.executor.ExecutorLoader;
 import azkaban.executor.ExecutorManager;
 import azkaban.executor.JdbcExecutorLoader;
 import azkaban.metrics.CommonMetrics;
+import azkaban.metrics.MetricsManager;
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
 import azkaban.spi.AzkabanException;
@@ -86,6 +87,7 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(ExecutorManager.class).in(Scopes.SINGLETON);
     bind(AlerterHolder.class).in(Scopes.SINGLETON);
     bind(CommonMetrics.class).in(Scopes.SINGLETON);
+    bind(MetricsManager.class).in(Scopes.SINGLETON);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -88,6 +88,7 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(AlerterHolder.class).in(Scopes.SINGLETON);
     bind(CommonMetrics.class).in(Scopes.SINGLETON);
     bind(MetricsManager.class).in(Scopes.SINGLETON);
+    bind(MetricRegistry.class).in(Scopes.SINGLETON);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
@@ -164,9 +165,4 @@ public class AzkabanCommonModule extends AbstractModule {
     return new QueryRunner(dataSource);
   }
 
-  @Provides
-  @Singleton
-  public MetricRegistry createMetricRegistry() {
-    return new MetricRegistry();
-  }
 }
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -164,5 +164,4 @@ public class AzkabanCommonModule extends AbstractModule {
   public QueryRunner createQueryRunner(final AzkabanDataSource dataSource) {
     return new QueryRunner(dataSource);
   }
-
 }
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -29,8 +29,6 @@ import azkaban.executor.AlerterHolder;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.ExecutorManager;
 import azkaban.executor.JdbcExecutorLoader;
-import azkaban.metrics.CommonMetrics;
-import azkaban.metrics.MetricsManager;
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
 import azkaban.spi.AzkabanException;
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -29,6 +29,7 @@ import azkaban.executor.AlerterHolder;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.ExecutorManager;
 import azkaban.executor.JdbcExecutorLoader;
+import azkaban.metrics.CommonMetrics;
 import azkaban.project.JdbcProjectLoader;
 import azkaban.project.ProjectLoader;
 import azkaban.spi.AzkabanException;
@@ -38,6 +39,7 @@ import azkaban.storage.StorageImplementationType;
 import azkaban.trigger.JdbcTriggerImpl;
 import azkaban.trigger.TriggerLoader;
 import azkaban.utils.Props;
+import com.codahale.metrics.MetricRegistry;
 import com.google.inject.AbstractModule;
 import com.google.inject.Inject;
 import com.google.inject.Provides;
@@ -83,6 +85,7 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(DataSource.class).to(AzkabanDataSource.class);
     bind(ExecutorManager.class).in(Scopes.SINGLETON);
     bind(AlerterHolder.class).in(Scopes.SINGLETON);
+    bind(CommonMetrics.class).in(Scopes.SINGLETON);
   }
 
   public Class<? extends Storage> resolveStorageClassType() {
@@ -157,5 +160,11 @@ public class AzkabanCommonModule extends AbstractModule {
   @Provides
   public QueryRunner createQueryRunner(final AzkabanDataSource dataSource) {
     return new QueryRunner(dataSource);
+  }
+
+  @Provides
+  @Singleton
+  public MetricRegistry createMetricRegistry() {
+    return new MetricRegistry();
   }
 }
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -86,8 +86,6 @@ public class AzkabanCommonModule extends AbstractModule {
     bind(DataSource.class).to(AzkabanDataSource.class);
     bind(ExecutorManager.class).in(Scopes.SINGLETON);
     bind(AlerterHolder.class).in(Scopes.SINGLETON);
-    bind(CommonMetrics.class).in(Scopes.SINGLETON);
-    bind(MetricsManager.class).in(Scopes.SINGLETON);
     bind(MetricRegistry.class).in(Scopes.SINGLETON);
   }
 
--- a/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
+++ b/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
@@ -16,13 +16,10 @@
 
 package azkaban.database;
 
-import static azkaban.ServiceProvider.SERVICE_PROVIDER;
-
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.Props;
 import java.io.IOException;
 import java.sql.Connection;
-import javax.inject.Inject;
 import org.apache.commons.dbutils.DbUtils;
 import org.apache.commons.dbutils.QueryRunner;
 
@@ -32,7 +29,6 @@ public abstract class AbstractJdbcLoader {
   private final AzkabanDataSource dataSource;
   private final CommonMetrics commonMetrics;
 
-  @Inject
   public AbstractJdbcLoader(final Props props, final CommonMetrics commonMetrics) {
     this.dataSource = DataSourceUtils.getDataSource(props);
     this.commonMetrics = commonMetrics;
@@ -49,8 +45,7 @@ public abstract class AbstractJdbcLoader {
       DbUtils.closeQuietly(connection);
       throw new IOException("Error getting DB connection.", e);
     }
-    SERVICE_PROVIDER.getInstance(CommonMetrics.class)
-        .setDBConnectionTime(System.currentTimeMillis() - startMs);
+    this.commonMetrics.setDBConnectionTime(System.currentTimeMillis() - startMs);
     return connection;
   }
 
--- a/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
+++ b/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
@@ -22,6 +22,7 @@ import azkaban.metrics.CommonMetrics;
 import azkaban.utils.Props;
 import java.io.IOException;
 import java.sql.Connection;
+import javax.inject.Inject;
 import org.apache.commons.dbutils.DbUtils;
 import org.apache.commons.dbutils.QueryRunner;
 
@@ -29,14 +30,17 @@ import org.apache.commons.dbutils.QueryRunner;
 public abstract class AbstractJdbcLoader {
 
   private final AzkabanDataSource dataSource;
+  private final CommonMetrics commonMetrics;
 
-  public AbstractJdbcLoader(final Props props) {
+  @Inject
+  public AbstractJdbcLoader(final Props props, final CommonMetrics commonMetrics) {
     this.dataSource = DataSourceUtils.getDataSource(props);
+    this.commonMetrics = commonMetrics;
   }
 
   protected Connection getDBConnection(final boolean autoCommit) throws IOException {
     Connection connection = null;
-    SERVICE_PROVIDER.getInstance(CommonMetrics.class).markDBConnection();
+    this.commonMetrics.markDBConnection();
     final long startMs = System.currentTimeMillis();
     try {
       connection = this.dataSource.getConnection();
--- a/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
+++ b/azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
@@ -16,6 +16,8 @@
 
 package azkaban.database;
 
+import static azkaban.ServiceProvider.SERVICE_PROVIDER;
+
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.Props;
 import java.io.IOException;
@@ -34,7 +36,7 @@ public abstract class AbstractJdbcLoader {
 
   protected Connection getDBConnection(final boolean autoCommit) throws IOException {
     Connection connection = null;
-    CommonMetrics.INSTANCE.markDBConnection();
+    SERVICE_PROVIDER.getInstance(CommonMetrics.class).markDBConnection();
     final long startMs = System.currentTimeMillis();
     try {
       connection = this.dataSource.getConnection();
@@ -43,7 +45,8 @@ public abstract class AbstractJdbcLoader {
       DbUtils.closeQuietly(connection);
       throw new IOException("Error getting DB connection.", e);
     }
-    CommonMetrics.INSTANCE.setDBConnectionTime(System.currentTimeMillis() - startMs);
+    SERVICE_PROVIDER.getInstance(CommonMetrics.class)
+        .setDBConnectionTime(System.currentTimeMillis() - startMs);
     return connection;
   }
 
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -48,12 +48,15 @@ public class ProcessJob extends AbstractProcessJob {
   public static final String KRB5CCNAME = "KRB5CCNAME";
   private static final Duration KILL_TIME = Duration.ofSeconds(30);
   private static final String MEMCHECK_ENABLED = "memCheck.enabled";
+  private final CommonMetrics commonMetrics;
   private volatile AzkabanProcess process;
   private volatile boolean killed = false;
 
   public ProcessJob(final String jobId, final Props sysProps,
       final Props jobProps, final Logger log) {
     super(jobId, sysProps, jobProps, log);
+    // TODO: reallocf fully guicify CommonMetrics through ProcessJob dependents
+    this.commonMetrics = SERVICE_PROVIDER.getInstance(CommonMetrics.class);
   }
 
   /**
@@ -140,7 +143,7 @@ public class ProcessJob extends AbstractProcessJob {
         if (isMemGranted) {
           info(String.format("Memory granted for job %s", getId()));
           if (attempt > 1) {
-            SERVICE_PROVIDER.getInstance(CommonMetrics.class).decrementOOMJobWaitCount();
+            this.commonMetrics.decrementOOMJobWaitCount();
           }
           break;
         }
@@ -150,7 +153,7 @@ public class ProcessJob extends AbstractProcessJob {
                   Constants.MEMORY_CHECK_INTERVAL_MS), attempt,
               Constants.MEMORY_CHECK_RETRY_LIMIT));
           if (attempt == 1) {
-            SERVICE_PROVIDER.getInstance(CommonMetrics.class).incrementOOMJobWaitCount();
+            this.commonMetrics.incrementOOMJobWaitCount();
           }
           synchronized (this) {
             try {
@@ -161,7 +164,7 @@ public class ProcessJob extends AbstractProcessJob {
             }
           }
           if (this.killed) {
-            SERVICE_PROVIDER.getInstance(CommonMetrics.class).decrementOOMJobWaitCount();
+            this.commonMetrics.decrementOOMJobWaitCount();
             info(String.format("Job %s was killed while waiting for memory check retry", getId()));
             return;
           }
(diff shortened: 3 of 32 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -140,7 +140,7 @@ public void run() throws Exception {
         if (isMemGranted) {
           info(String.format("Memory granted for job %s", getId()));
           if (attempt > 1) {
-            CommonMetrics.INSTANCE.decrementOOMJobWaitCount();
+            SERVICE_PROVIDER.getInstance(CommonMetrics.class).decrementOOMJobWaitCount();
```
kunkun-tang: Not sure if this is a good practice. If me, I would have a CommonMetrics's instance in the Constructor of `ProcessJob`. Basically, I don't know the performance/efficiency of using `SERVICE_PROVIDER.getInstance`

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -16,26 +16,29 @@
...
   private final AtomicLong dbConnectionTime = new AtomicLong(0L);
   private final AtomicLong OOMWaitingJobCount = new AtomicLong(0L);
   private final MetricRegistry registry;
   private Meter dbConnectionMeter;
   private Meter flowFailMeter;
 
+  @Inject
   CommonMetrics() {
-    this.registry = MetricsManager.INSTANCE.getRegistry();
+    this.registry = SERVICE_PROVIDER.getInstance(MetricsManager.class).getRegistry();
```
kunkun-tang: Sort of believe We should have MetricManager in parameter given this case.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
```diff
@@ -29,7 +31,7 @@
   private final MetricRegistry registry;
 
   ExecMetrics() {
-    this.registry = MetricsManager.INSTANCE.getRegistry();
+    this.registry = SERVICE_PROVIDER.getInstance(MetricsManager.class).getRegistry();
```
kunkun-tang: Sort of believe We should have MetricManager in parameter given this case.

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -34,7 +36,7 @@ public AbstractJdbcLoader(final Props props) {
 
   protected Connection getDBConnection(final boolean autoCommit) throws IOException {
     Connection connection = null;
-    CommonMetrics.INSTANCE.markDBConnection();
+    SERVICE_PROVIDER.getInstance(CommonMetrics.class).markDBConnection();
```
HappyRay: Consider saving this reference in the constructor? This class can be considered a dependency of this class.

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
HappyRay: Why not take MetricRegistry directly as a dependency and have Guice create it?

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -28,15 +28,15 @@
...
  * The singleton class, MetricsManager, is the place to have MetricRegistry and ConsoleReporter in
  * this class. Also, web servers and executors can call {@link #startReporting(String, Props)} to
  * start reporting AZ metrics to remote metrics server.
  */
-public enum MetricsManager {
-  INSTANCE;
+public class MetricsManager {
 
   private static final Logger logger = Logger.getLogger(MetricsManager.class);
   private final MetricRegistry registry = new MetricRegistry();
```
HappyRay: Declare MetricRegistry as a dependency and have Guice create it for you?

azkaban-common/src/test/java/azkaban/metrics/CommonMetricsTest.java
```diff
@@ -16,41 +16,39 @@
...
+    testUtil = new MetricsTestUtility(
+        SERVICE_PROVIDER.getInstance(MetricsManager.class).getRegistry());
+    metrics = SERVICE_PROVIDER.getInstance(CommonMetrics.class);
   }
 
   @Test
   public void testDBConnectionTimeMetrics() {
-    this.metrics.setDBConnectionTime(14);
-    assertEquals(14, this.testUtil.getGaugeValue("dbConnectionTime"));
+    metrics.setDBConnectionTime(14);
```
HappyRay: Did you apply saveaction?

reallocf: I did - metrics is static now because JUnit requires setUp to be static in order for it to be annotated @ BeforeClass which causes metrics and testUtil to be static. So "this." is no longer recommended by intelliJ. Does that make sense?

azkaban-common/src/test/java/azkaban/metrics/CommonMetricsTest.java
```diff
@@ -16,41 +16,39 @@
...
-    // todo HappyRay: move MetricsManager, CommonMetrics to use Juice.
-    this.testUtil = new MetricsTestUtility(MetricsManager.INSTANCE.getRegistry());
-    this.metrics = CommonMetrics.INSTANCE;
+  private static MetricsTestUtility testUtil;
+  private static CommonMetrics metrics;
+
+  @BeforeClass
+  public static void setUp() {
+    // initialize new guice MetricsManager
+    MetricsTestUtility.initServiceProvider();
```
HappyRay: Here is an alternative:

CommonMetrics constructor takes an instance of MetricRegistry. In the production code, it is constructed using Guice.

In test code, simply new the instances you need without using Guice.

azkaban-common/src/test/java/azkaban/metrics/CommonMetricsTest.java
HappyRay: I see. Why do they need to be in BeforeClass?

HappyRay: Typically it is better to have tests to be independent of each other as possible i.e. create their own instances when needed.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
kunkun-tang: +1.

azkaban-common/src/test/java/azkaban/metrics/CommonMetricsTest.java
```diff
@@ -16,41 +16,39 @@
...
+    testUtil = new MetricsTestUtility(
+        SERVICE_PROVIDER.getInstance(MetricsManager.class).getRegistry());
+    metrics = SERVICE_PROVIDER.getInstance(CommonMetrics.class);
```
kunkun-tang: We probably need to move this statement (`metrics = SERV...`) into separate tests, since we are testing itself specifically.

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -16,25 +16,31 @@
 
 package azkaban.database;
 
+import static azkaban.ServiceProvider.SERVICE_PROVIDER;
+
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.Props;
 import java.io.IOException;
 import java.sql.Connection;
+import javax.inject.Inject;
```
HappyRay: What is this used for?

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -16,25 +16,31 @@
...
 
 public abstract class AbstractJdbcLoader {
 
   private final AzkabanDataSource dataSource;
+  private final CommonMetrics commonMetrics;
 
-  public AbstractJdbcLoader(final Props props) {
+  @Inject
+  public AbstractJdbcLoader(final Props props, final CommonMetrics commonMetrics) {
     this.dataSource = DataSourceUtils.getDataSource(props);
```
HappyRay: Not in this PR, but in the future, we should get datasource from Juice. 
@kunkun-tang

azkaban-common/src/main/java/azkaban/database/AbstractJdbcLoader.java
```diff
@@ -43,7 +49,8 @@ protected Connection getDBConnection(final boolean autoCommit) throws IOExceptio
       DbUtils.closeQuietly(connection);
       throw new IOException("Error getting DB connection.", e);
     }
-    CommonMetrics.INSTANCE.setDBConnectionTime(System.currentTimeMillis() - startMs);
+    SERVICE_PROVIDER.getInstance(CommonMetrics.class)
```
HappyRay: Shouldn't this use this.commonMetrics too?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
HappyRay: +1

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -16,26 +16,29 @@
...
   private final AtomicLong dbConnectionTime = new AtomicLong(0L);
   private final AtomicLong OOMWaitingJobCount = new AtomicLong(0L);
   private final MetricRegistry registry;
   private Meter dbConnectionMeter;
   private Meter flowFailMeter;
 
-  CommonMetrics() {
-    this.registry = MetricsManager.INSTANCE.getRegistry();
+  @Inject
+  public CommonMetrics(MetricsManager metricsManager) {
```
HappyRay: Why not depend on MetircsRegistry directly?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -66,7 +69,8 @@ private ExecutorManager createMultiExecutorManagerInstance(
 
     loader.addExecutor("localhost", 12345);
     loader.addExecutor("localhost", 12346);
-    return new ExecutorManager(props, loader, new AlerterHolder(props));
+    return new ExecutorManager(props, loader, new AlerterHolder(props),
+        new CommonMetrics(new MetricsManager(new MetricRegistry())));
```
HappyRay: Will a mock work here?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -705,7 +705,8 @@ private static Props loadConfigurationFromAzkabanHome() {
 
   private void startWebMetrics() throws Exception {
 
-    final MetricRegistry registry = MetricsManager.INSTANCE.getRegistry();
+    final MetricRegistry registry = SERVICE_PROVIDER.getInstance(MetricsManager.class)
```
HappyRay: Why not just get the MetricsRegistry directly?

HappyRay: Have you tried binding MetricsRegistry class to a singleton in a GuiceModule?

reallocf: This is being refactored to rely on MetricsManager dependency which is handled in constructor. The MetricsManager class is used explicitly on line 736 to take advantage of the startReporting method. Do you think it'd be better to depend on both MetricsManager and MetricsRegistry?

Also, MetricsRegistry is binded to a singleton locally, I'm resolving some consequences of that switch then I'll push a new version - thanks for the feedback! :+1:

HappyRay: I see. I would suggest listing all the direct dependencies directly to make the dependencies more explicit.

reallocf: That's a good point. I'll have it directly depend on both MetricsManager and MetricsRegistry then. Thanks again!

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -158,4 +163,10 @@ public FileSystem createHadoopFileSystem(final Configuration hadoopConf) {
   public QueryRunner createQueryRunner(final AzkabanDataSource dataSource) {
     return new QueryRunner(dataSource);
   }
+
+  @Provides
```
HappyRay: Why not 
    bind(MetricRegistry.class).in(Scopes.SINGLETON);
it also seems to work.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -28,24 +28,26 @@
 import java.lang.reflect.Constructor;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
```
HappyRay: Is this right? 
Not com.google.inject.Inject;?

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerTest.java
```diff
@@ -53,6 +59,20 @@ public JobTypeManagerTest() {
 
   @Before
   public void setUp() throws Exception {
+    // TODO: reallocf Remove injector creation/deletion when ProcessJob is fully guicified
+    final Props props = new Props();
+
+    final Injector injector = Guice.createInjector(new AbstractModule() {
```
HappyRay: Why is this change necessary?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -36,6 +36,6 @@ protected void configure() {
     bind(AzkabanExecutorServer.class).in(Scopes.SINGLETON);
     bind(TriggerManager.class).in(Scopes.SINGLETON);
     bind(FlowRunnerManager.class).in(Scopes.SINGLETON);
-
+    // TODO: add ExecMetrics when guicy
```
HappyRay: Is this Todo necessary?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -172,6 +174,8 @@ public AzkabanWebServer(final Server server, final Props props) throws Exception
     this.executorManager = SERVICE_PROVIDER.getInstance(ExecutorManager.class);
     this.projectManager = SERVICE_PROVIDER.getInstance(ProjectManager.class);
     this.triggerManager = SERVICE_PROVIDER.getInstance(TriggerManager.class);
+    this.metricsManager = SERVICE_PROVIDER.getInstance(MetricsManager.class);
+    this.registry = new MetricRegistry();
```
HappyRay: Shouldn't get it from Serivce_provider too?

azkaban-web-server/src/main/java/azkaban/webapp/WebMetrics.java
```diff
@@ -38,8 +37,13 @@
   // How long does user log fetch take when user call fetch-log api.
   private final AtomicLong logFetchLatency = new AtomicLong(0L);
 
-  WebMetrics() {
-    this.registry = MetricsManager.INSTANCE.getRegistry();
+  @Inject
+  WebMetrics(final MetricRegistry registry) {
+    // Add unit test coverage to guarantee no duplication
+    // Seeing this error after guicifying, but before binding as singleton in AzkabanWebServerModule:
```
HappyRay: Could you find out what the problem is and then remove this comment?

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTest.java
```diff
@@ -17,22 +17,46 @@
...
 import javax.servlet.http.HttpServletResponse;
+import org.junit.Before;
 import org.junit.Test;
 
 public class LoginAbstractAzkabanServletTest {
 
+  @Before
+  public void setUp() {
+    // TODO: reallocf remove setUp() when LoginAbstractServlet fully guicified
+    final Props props = new Props();
```
HappyRay: If this logic is need, please find a way to reduce the code duplication.

see azkaban.test.Utils#initServiceProvider

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
reallocf: This was a recommendation @kunkun-tang made because he said the usage was similar to QueryRunner

HappyRay: @kunkun-tang could you explain?

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
reallocf: Nice catch! Thank you!

HappyRay: In general, try to make small incremental changes. When a change becomes bigger, it is more likely to miss issues. It takes longer to finish reviews too.

reallocf: Definitely, I appreciate the feedback. This one kind of got away from me as I was dealing with unraveling the dependencies. I also didn't understand Guice very well going into it, so it was a lot of learning on the fly 😅

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
reallocf: This is a consequence of the change that @kunkun-tang suggested in AzkabanCommonModule.

azkaban-web-server/src/main/java/azkaban/webapp/WebMetrics.java
reallocf: Yes, the problem has already been fixed. I just included this because there was no unit test confirming that WebMetrics is a singleton, which allowed me to build and run and then find the bug while using my local version. Adding `bind(WebMetrics.class).in(Scopes.SINGLETON);` to AzkabanWebServerModule fixed the problem, but it would probably be good to have a unit test in case anybody removes it for whatever reason. I didn't think it'd be wise to include it in this (already very large) PR, so I added the comment. What do you think?

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerTest.java
reallocf: This was included in order to handle the partial guicing of the ProcessJob class where I leverage the SERVICE_PROVIDER instead of injecting it the right way.

The problem with doing it the right was is then I have to refactor every class that extends the ProcessJob and then refactor every test for the classes that extend the ProcessJob and then suddenly I'm refactoring another 10+ files. I'd rather break this up into separate PR's which is why I'm leveraging this bandaid for now. Do you think it'd be better to implement Guice throughout this dependency chain now?

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTest.java
reallocf: I can do that. Do you think it's a good idea to build that infrastructure when this code will (hopefully) be retired quickly?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -99,10 +99,11 @@
   @Inject
   public AzkabanExecutorServer(final Props props,
       final ExecutorLoader executionLoader,
-      final FlowRunnerManager runnerManager) throws Exception {
+      final FlowRunnerManager runnerManager, final MetricsManager metricsManager) throws Exception {
```
juhoautio: Would it be possible to not touch `AzkabanExecutorServer` yet? This earlier PR has conflicting changes: https://github.com/azkaban/azkaban/pull/1113. Or, if you're going even further with injection on this class, maybe you want to check out that PR so that you don't have to figure out the same things again.

azkaban-web-server/src/main/java/azkaban/webapp/WebMetrics.java
HappyRay: Sounds good. Thanks for thinking about how to prevent regressions.

HappyRay: IMHO, I prefer annotating the class itself with singleton if the class is designed to be used as singleton everywhere. It's less code and the information is centralized. Otherwise, readers have to check the module files too to find this information.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTest.java
HappyRay: Yes if it is not a lot of work.
Can you reuse the code I referenced?  see azkaban.test.Utils#initServiceProvider

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
HappyRay: I think this is a different change. I don't expect resolving conflict will be too hard.

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerTest.java
HappyRay: Yes, I am in favor of phased approach.

In this case, I replaced your Guice initiailization code with 
import static azkaban.test.Utils.initServiceProvider;

initServiceProvider();

And the tests in this file passed

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
juhoautio: True.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
kunkun-tang: I just replaced this line by `this.registry = SERVICE_PROVIDER.getInstance(MetricRegistry.class);` and didn't observe any issues. What did you see if you use `Serivce_provider ` ?

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
kunkun-tang: I was trying to google the difference between using `@provide` and `to bind`, but failed. Some online resources say that there'w no difference between the two approaches. I believe @suvodeep-pyne has better understanding on this case.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -162,7 +170,8 @@ public void testSetupExecutorsException() throws ExecutorManagerException {
     Executor executor1 = loader.addExecutor("localhost", 12345);
 
     ExecutorManager manager =
-      new ExecutorManager(props, loader, new AlerterHolder(props));
+      new ExecutorManager(props, loader, new AlerterHolder(props),
```
kunkun-tang: Is there some way that we could consolidate the code that set up an Executor Manager.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
reallocf: Should we just use to bind then in order to standardize on that?

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerTest.java
reallocf: Awesome, thank you for doing that! I'll make the update 👍

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTest.java
reallocf: Will do!

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
reallocf: :+1:

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -83,6 +86,9 @@ protected void configure() {
     bind(DataSource.class).to(AzkabanDataSource.class);
     bind(ExecutorManager.class).in(Scopes.SINGLETON);
     bind(AlerterHolder.class).in(Scopes.SINGLETON);
+    bind(CommonMetrics.class).in(Scopes.SINGLETON);
```
HappyRay: Since CommonMetrics is annotated as singleton already, I think you don't need this line here. same with MetricsManager.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
```diff
@@ -29,7 +31,8 @@
   private final MetricRegistry registry;
 
   ExecMetrics() {
-    this.registry = MetricsManager.INSTANCE.getRegistry();
+    // TODO: reallocf make guicy
+    this.registry = SERVICE_PROVIDER.getInstance(MetricsManager.class).getRegistry();
```
HappyRay: Why not directly get the MetricsRegistry instance?

azkaban-exec-server/src/test/java/azkaban/execapp/AzkabanExecutorServerTest.java
```diff
@@ -86,7 +87,9 @@ public void testInjection() throws Exception {
         new AzkabanCommonModule(props),
         new AzkabanExecServerModule()
     );
+    SERVICE_PROVIDER.unsetInjector();
```
HappyRay: Why is this change necessary?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
```diff
@@ -32,5 +32,6 @@
   protected void configure() {
     bind(Server.class).toProvider(WebServerProvider.class);
     bind(AzkabanWebServer.class).in(Scopes.SINGLETON);
+    bind(WebMetrics.class).in(Scopes.SINGLETON);
```
HappyRay: Same as the previous comment.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
reallocf: Would it be a good idea to standardize on the @Singleton annotation? Not in this PR, but in the future.

azkaban-exec-server/src/test/java/azkaban/execapp/AzkabanExecutorServerTest.java
reallocf: Looks like it isn't anymore! It was necessary when things were not guiced the right way, but it seems to be working now without the change. Thanks!!

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
HappyRay: This is an example of why Guice is helpful. Otherwise, we may have more similar code in the production code.

PR #1197, opened 2017-06-10
Fix deprecated api compiler warnings in HadoopSecurityManager_H_2_0

HappyRay:
Warning:(446, 20) java: cancelDelegationToken(org.apache.hadoop.security.token.Token<org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier>) in org.apache.hadoop.mapred.JobClient has been deprecated

Fixed by removing the now unused private methods.
Also corrected the log messages.
The previous log messages still show that the tokens are canceled while the code had been changed to not canceling the tokens.


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -515,14 +453,11 @@ public void cancelTokens(final File tokenFile, final String userToProxy, final L
           logger.info("Cancelling hive token.");
           cancelHiveToken(t, userToProxy);
         } else if (t.getKind().equals(new Text("RM_DELEGATION_TOKEN"))) {
-          logger.info("Cancelling mr job tracker token.");
-          // cancelMRJobTrackerToken(t, userToProxy);
+          logger.info("Ignore cancelling mr job tracker token request.");
         } else if (t.getKind().equals(new Text("HDFS_DELEGATION_TOKEN"))) {
-          logger.info("Cancelling namenode token.");
-          // cancelNameNodeToken(t, userToProxy);
```
jamiesjc: Maybe also remove cancelNameNodeToken() in line 405 since it's not used?

HappyRay: Good point. I missed it. Will do.

PR #1199, opened 2017-06-10
Allow db operator to have rollback feature

kunkun-tang:
When I was refactoring [this code](https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java#L317), I noticed that we need to consider DB connection rollback when we do db operations. In this patch, I add and implement the roll back interface, as part of follow-up JDBC refactors. I use default interface to complete the task, and the default value is `we don't do rollback except specified.`

Besides, I add another method `getDataSource`, since [this code](https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java#L682) need to check datasource's attribute before operations.


azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -36,27 +36,51 @@
...
    * commit every time.
    *
    * @param operations A sequence of DB operations
+   * @param enableRollback decides if we want to roll back the transaction when sql exception is thrown.
    * @param <T> The type of object that the operations returns. Note that T could be null
    * @return T The object returned by the SQL statement, expected by the caller
+   * @throws SQLException
    */
-  <T> T transaction(SQLTransaction<T> operations) throws SQLException;
+  <T> T transaction(SQLTransaction<T> operations, boolean enableRollback) throws SQLException;
```
HappyRay: When would rollback not be desirable?

kunkun-tang: Do you mean not enabling rollback by default? @HappyRay

PR #1200, opened 2017-06-10
Have method to check allowsOnDuplicateKey

kunkun-tang:
When I was refactoring [this code](https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/project/JdbcProjectLoader.java#L682), I noticed that we need to consider allowsOnDuplicateKey since the original Azkaban database modules has this. So I'm completing now.


kunkun-tang:
Will have a TODO comment (using standard SQL statements.) right there to remind us in future.

PR #1186, opened 2017-06-08
DON'T MERGE - test 50x revert #1167

juhoautio:
This is just to stress test #1185.

Run FlowRunner tests 50x to reveal random failures.

There had been 2 failures in this Travis build:
https://travis-ci.org/azkaban/azkaban/builds/240580566?utm_source=github_status&utm_medium=notification

```
azkaban.execapp.FlowRunnerTest > execAndCancel FAILED
    java.lang.AssertionError: Wrong status for [job4] expected:<KILLED> but was:<SUCCEEDED>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:117)
        at azkaban.execapp.FlowRunnerTest.execAndCancel(FlowRunnerTest.java:274)
azkaban.execapp.FlowRunnerTest2 > testFailedFinishingFailure3 FAILED
    java.lang.AssertionError: Wrong status for [jobb:innerJobC] expected:<RUNNING> but was:<READY>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:117)
        at azkaban.execapp.FlowRunnerTest2.testFailedFinishingFailure3(FlowRunnerTest2.java:517)
:azkaban-solo-server:check
:azkaban-solo-server:build
```

PR #1201, opened 2017-06-11
Define dependencies in one place

HappyRay:
This is to promote consistency and make updating dependencies easier.

see
https://discuss.gradle.org/t/what-is-the-proper-way-to-make-dependencies-consistent-in-a-multi-project-setup/17383/12

Also:

Define the common dependencies in the root build.gradle file.

            compile deps.log4j
            compile deps.guice
            compile deps.slf4j

            runtime deps.slf4jLog4j

Introduce common versions in the ext.versions map.


build.gradle
```diff
@@ -21,6 +21,62 @@ allprojects {
...
+        assertj             : 'org.assertj:assertj-core:3.8.0',
+        collections         : 'commons-collections:commons-collections:3.2.2',
+        commonsLang         : 'commons-lang:commons-lang:2.6',
+        dbcp2               : 'org.apache.commons:commons-dbcp2:2.1.1',
+        dbutils             : 'commons-dbutils:commons-dbutils:1.5',
+        fileupload          : 'commons-fileupload:commons-fileupload:1.2.1',
+        guava               : 'com.google.guava:guava:21.0',
+        guice               : 'com.google.inject:guice:4.1.0',
+        h2                  : 'com.h2database:h2:1.4.193',
+        hadoopAnnotations   : "org.apache.hadoop:hadoop-annotations:" + versions.hadoop,
```
juhoautio: IMHO this was a bit better:
`"org.apache.hadoop:hadoop-annotations:$hadoopVersion"`.

Is anything similar possible with the new structure? I don't know about gradle, but maybe it supports something like this:
`"org.apache.hadoop:hadoop-annotations:${versions.hadoop}"`
?

HappyRay: Yes,         hadoopAnnotations   : "org.apache.hadoop:hadoop-annotations:${versions.hadoop}",
also works.
Why do you think the $ version is better?

The + version requires less typing. And IMHO it is as readable. 
I don't have a strong opinion though. 
If you feel strongly about it, how about you create a PR after this is merged?

HappyRay:
The build failed again.

azkaban.execapp.FlowRunnerTest2 > testFailedFinishingFailure3 FAILED
    java.lang.AssertionError: Wrong status for [jobb] expected:<FAILED_FINISHING> but was:<RUNNING>
        at org.junit.Assert.assertEquals(Assert.java:118)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:117)
        at azkaban.execapp.FlowRunnerTest2.testFailedFinishingFailure3(FlowRunnerTest2.java:520)
:azkaban-solo-server:check
:azkaban-solo-server:build
:azkaban-common:check
:azkaban-common:build
74 tests completed, 1 failed, 22 skipped

@juhoautio I think we should disable this test file until it is stable. What do you think?

juhoautio:
Can you get the console output for the failed test on that Travis build somehow? Otherwise it's much harder to fix, when this can't be reproduced locally.. For example I ran both FlowRunner tests 2 x 50 times on Travis and there were no failures (see https://github.com/azkaban/azkaban/pull/1186). I also ran them 1000 x locally, no failures..

Last time that there was a Travis failure, the failed tests were:

azkaban.execapp.FlowRunnerTest > execAndCancel 
azkaban.execapp.FlowRunnerTest2 > testFailedFinishingFailure3 (this is the same that you reported as failed now again)

I would add `@Ignore` only on those two test methods, not on entire files.

build.gradle
juhoautio: Never mind, no strong opinion by me and can't really argue why I would prefer the $ way. I guess I just personally find it more readable when you can more easily see where a string starts and where it ends. Maybe someone even shared that view, because it was done with $ originally :)

juhoautio:
Here, maybe one last try before ignores :}
https://github.com/azkaban/azkaban/pull/1207

HappyRay:
Unfortunately, I tried and couldn't get the console output from Travis. The best I was able to do was to get the stack trace when a test failed. 

Please contribute the knowledge if you find a way.

build.gradle
```diff
@@ -21,6 +21,62 @@ allprojects {
...
 }
 
+ext.versions = [
+        hadoop: '2.6.1',
+        hive  : '1.1.0',
+        restli: '1.15.7',
+        slf4j : '1.7.18',
+]
+
+ext.deps = [
```
kunkun-tang: Could these dependencies be declared in a json file?

HappyRay: Why?

kunkun-tang: In my standard practice experience, dependencies are usually put in json file, like azkaban-web-server/package.json

HappyRay: This is gradle. And gradle has its own syntax.

HappyRay:
@juhoautio I found a way to log test console output
see https://gist.github.com/HappyRay/84119110f957dc114402178258723632

juhoautio:
Are you going to enable it for Travis builds? Otherwise it can be hard to catch problems if they occur rarely.. But I guess there haven't been more random failures after the latest stability fixes?

HappyRay:
No. The output is too noisy. 
It's a tool, however, to debug issues when needed. It can be enabled on a per pull request basis. 

In any case, the test should be reliable and repeatable (locally or in travis ) by design. I would prefer simplifying the current DAG engining design and tests.

PR #1206, opened 2017-06-11
Simplify the azkaban-test subprojet name to test

HappyRay:
The azkaban prefix is redundant and requires unnecessary
typing.

Keep the azkaban prefix in the jar file name.


HappyRay:
If this model works well and others agree this is a good idea, we can do the same to the other modules.

HappyRay:
It seems that git is smart enough to preserve the change history after the directory is renamed.

kunkun-tang:
Any good example how other open source project name modules? I would say I'm already familiar with current namings.

HappyRay:
The ones I have seen don't have the duplicated prefix including the examples given in the gradle documentation.

suvodeep-pyne:
@HappyRay 

> When did you test this behavior? I saw that the directory name is test in the artifactory and the artifact name is azkaban-test. And the .ivy file seems to be correct.
What problem did you observe?

I don't see a direct issue since the artifact got published successfully. It is rather the structure of the publish which I think is not conventional. Let me explain.

I need to fetch the artifact using the coordinates:
```
runtime 'com.linkedin.azkaban:test:3.29.0'
```
and the artifact downloaded is `com/linkedin/azkaban/test/3.29.0/azkaban-test-3.29.0.jar`

Normally, the name of the artifact and the jar exactly match. Here, that is no longer the case and thus might be confusing. So you are fetching the `test` jar but it's named as `azkaban-test`.

This also violates the convention of all the other code modules since all modules are named as `azkaban-*` except this one which now also publishes a jar with a different custom output path.

Regards

PR #1219, opened 2017-06-13
Restore loadFlow API without stats

wyukawa:
loadFlow API was removed in https://github.com/azkaban/azkaban/pull/1182

But, there is no API to  get all schedule data.

I think that loadFlow API without stats will be acceptable.


jamiesjc:
LGTM. If you still need to use that API, we would like to keep it for your convenience. Thanks for removing the stats as well.

wyukawa:
@HappyRay Could you please take a look at this pull request when you get a chance?

HappyRay:
@jamiesjc?

jamiesjc:
I've reviewed the change and it looks good. Will merge it now.

wyukawa:
Thanks!

PR #1223, opened 2017-06-16
Upgrade gradle wrapper to use version 4.0.2

HappyRay:
See
https://docs.gradle.org/current/release-notes.html

Updated using the command:

```
./gradlew wrapper
```


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,4 +1,4 @@
-#Sun May 07 08:40:37 PDT 2017
+#Sat Jun 17 07:21:26 PDT 2017
```
chengren311: why do we need this? can we just remove this line?

HappyRay: It's autogenerated. And it documents when the file was generated.
I don't think we should manually edit this file.

HappyRay:
Thanks. I will wait a bit until I hear back from the tools team on the potential impact for integration with internal toolings.

suvodeep-pyne:
@HappyRay should we merge this now?

suvodeep-pyne:
It seems `4.0.2` is the most stable in the `4.0.x` pipeline
https://docs.gradle.org/4.0.2/release-notes.html

No known issues. List of Releases: https://services.gradle.org/distributions/

build.gradle
```diff
@@ -171,11 +171,27 @@ subprojects {
     }
 }
 
+/*
```
suvodeep-pyne: This should probably be inserted at the top if at all? Do we have to include copyright notices for build scripts?

PR #1230, opened 2017-06-18
Change the travis config to run the check task instead of the build task

HappyRay:
Travis by default runs the gradle check task
 See
 https://docs.travis-ci.com/user/languages/java/#Projects-Using-Gradle

 We can skip assemble task with this change. If we discover a problem with assemble task later, we can fix it later. Given that
  changes to the Gradle build script is not very often and
  developers can run the build script locally when they do, it's a
  better trade off.
 see https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ languages: java
 jdk:
   - oraclejdk8
 sudo: false
+script: ./gradlew check
 
 # see https://docs.travis-ci.com/user/languages/java/#Projects-Using-Gradle
 before_cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ languages: java
 jdk:
   - oraclejdk8
 sudo: false
-script: ./gradlew clean build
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@ languages: java
 jdk:
   - oraclejdk8
 sudo: false
-script: ./gradlew clean build
+script: ./gradlew check
```
suvodeep-pyne: I am a bit skeptical about this. We have a lot of distribution code where we package stuff from a ton of locations. Will it be a good idea to stop exercising this code totally? I am wondering whether we could potentially run into surprises while publishing.

IMO, I would sleep better if I know that all targets build and are working fine. I can sacrifice a minute of build time for that. Let me know your thoughts.

HappyRay: Ok, if you don't feel comfortable with this change. I will close it.

Thanks

PR #1231, opened 2017-06-19
Enable parallel tests

HappyRay:
Testing:

Reduced the test time from 15 seconds to 12 seconds on my mac.

./gradlew azkaban-common:cleanTest azkaban-common:test

See
http://mrhaki.blogspot.com/2010/11/gradle-goodness-running-tests-in.html
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html
https://docs.gradle.org/current/userguide/java_plugin.html#sec:test_execution

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,8 +120,12 @@ subprojects {
             testCompile deps.assertj
             testCompile deps.junit
             testCompile deps.mockito
+        }
 
-
+        test {
+            // Set maxParallelForks to a large number and let gradle to force it to a the
+            // max-workers number when needed.
+            maxParallelForks =  12
         }
     }
 


build.gradle
```diff
@@ -120,8 +120,12 @@ subprojects {
             testCompile deps.assertj
             testCompile deps.junit
             testCompile deps.mockito
+        }
 
-
+        test {
+            // Set maxParallelForks to a large number and let gradle to force it to a the
+            // max-workers number when needed.
+            maxParallelForks =  12
```
chengren311: i'm curious why the performance gain is not so significant from 1 to 12 forks.
also maybe considering number of processor here could be more flexible:
e.x:  
maxParallelForks = Runtime.runtime.availableProcessors * 4

HappyRay: 1. gradle will reduce the number based on the max-worker config. On my mac, the actual number is 4.

2. I suspect not all tests are run concurrently. Even they do, the overall time is dictated by the slowest test. Good question. I am not sure exactly though. The improvement is observable consistently though.

Do you have any objection to this change?

chengren311: i see, thanks

burgerkingeater:
./gradlew azkaban-common:cleanTest azkaban-common:test

for me test running time is increased from 33s to 18s with this change.

PR #1235, opened 2017-06-19
Use RetryingHiveMetaStoreClient instead of HiveMetaStoreClient

suvodeep-pyne:
Azkaban currently uses `HiveMetaStoreClient` while communicating with the Hive Metastore ref class: `HadoopSecurityManager_H_2_0`
This client does not automatically retry/reconnect in case of transient failures, such as when a metastore is restarted. The RetryingHiveMetaStoreClient does automatically retry/reconnect in case of errors.

This change switches it to use the RetryingHiveMetaStoreClient (which is what the Hive CliDriver uses internally).

NOTE: This introduces a runtime dependency on `hive-exec` jar. Currently, this is set as a compile only dependency in Azkaban but it actually requires this jar in the runtime and expects to get it from the `HIVE_HOME` environment variable.

Ref: https://github.com/linkedin/gobblin/issues/1506


azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    msc.dropDatabase(dbName, true, true);
+    msc.createDatabase(db);
+    db = msc.getDatabase(dbName);
+    Assert.assertEquals(db.getName(), dbName);
+    Assert.assertEquals(db.getDescription(), description);
+    Assert.assertEquals(db.getLocationUri(), location);
+
+    // Clean up if the test is successful
+    cleanup();
```
chengren311: maybe put it in finally block?

suvodeep-pyne: I would prefer the cleanup to not take place if something goes wrong. That is the reason I didn't put it a `finally` block. Else, I would have simply used `@AfterTest` to ensure that the cleanup always takes place.

suvodeep-pyne:
This is a smart catch by Travis. Apparently [this](https://mvnrepository.com/artifact/org.pentaho/pentaho-aggdesigner-algorithm/5.1.5-jhyde) dependency comes transitively and is not present in maven central but present in jcenter. This was cached on my machine but it fails on github.

azkaban-hadoop-security-plugin/build.gradle
```diff
@@ -1,12 +1,20 @@
...
+
 dependencies {
     compile project(":azkaban-common")
 
     compileOnly deps.hadoopCommon
     compileOnly deps.hadoopMRClientCommon
     compileOnly deps.hadoopMRClientCore
     compileOnly deps.hiveMetastore
+    compileOnly(deps.hiveExec) {
+        exclude group: 'org.pentaho', module: 'pentaho-aggdesigner-algorithm'
```
HappyRay: Why is it necessary to exclude these modules?

azkaban-hadoop-security-plugin/build.gradle
```diff
@@ -1,12 +1,20 @@
 apply plugin: 'distribution'
 
+configurations {
+    testCompile.extendsFrom compileOnly
```
HappyRay: Why is this change necessary?
How is it different from modifying testCompileOnly?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
```
HappyRay: Nice. Thanks for remembering to add the notice. I configured Idea to add it automatically. Did you do the same?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * An implementation of {@link BasePooledObjectFactory} for {@link IMetaStoreClient}.
+ */
+public class HiveMetaStoreClientFactory extends BasePooledObjectFactory<IMetaStoreClient> {
+
+  public static final String HIVE_METASTORE_TOKEN_SIGNATURE = "hive.metastore.token.signature";
+  private static final Logger LOG = LoggerFactory.getLogger(HiveMetaStoreClientFactory.class);
```
HappyRay: The convention is to use "logger". 
LOG should be marked as violating the google style guide.

HappyRay: Did you set up the checkstyle check? We should standardize the setup.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+        return null;
+      }
+
+      try {
+        final HiveStorageHandler storageHandler =
+            HiveUtils.getStorageHandler(HiveMetaStoreClientFactory.this.hiveConf,
+                tbl.getParameters().get(META_TABLE_STORAGE));
+        return storageHandler == null ? null : storageHandler.getMetaHook();
+      } catch (final HiveException e) {
+        LOG.error(e.toString());
```
HappyRay: Should log the stack too?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+        throw new MetaException("Failed to get storage handler: " + e);
```
HappyRay: shouldn't it be , e to include the inner exception?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+
+    return RetryingMetaStoreClient
+        .getProxy(this.hiveConf, hookLoader, HiveMetaStoreClient.class.getName());
+  }
+
+  @Override
+  public IMetaStoreClient create() {
+    try {
+      return createMetaStoreClient();
+    } catch (final MetaException e) {
```
HappyRay: If the exception is caught here, why not catch the HiveException here?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
+import org.apache.hadoop.hive.ql.metadata.HiveUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * An implementation of {@link BasePooledObjectFactory} for {@link IMetaStoreClient}.
+ */
+public class HiveMetaStoreClientFactory extends BasePooledObjectFactory<IMetaStoreClient> {
```
HappyRay: Is this code new or is it moved from somewhere else?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -402,11 +403,10 @@ private void getToken(final String userToProxy) throws InterruptedException,
     }
   }
 
-  private void cancelHiveToken(final Token<? extends TokenIdentifier> t,
-      final String userToProxy) throws HadoopSecurityManagerException {
+  private void cancelHiveToken(final Token<? extends TokenIdentifier> t)
+      throws HadoopSecurityManagerException {
```
HappyRay: Is this an independent code clean up change?

build.gradle
```diff
@@ -46,6 +46,7 @@ ext.deps = [
         hadoopMRClientCommon: "org.apache.hadoop:hadoop-mapreduce-client-common:" + versions.hadoop,
         hadoopMRClientCore  : "org.apache.hadoop:hadoop-mapreduce-client-core:" + versions.hadoop,
         hiveMetastore       : "org.apache.hive:hive-metastore:" + versions.hive,
+        hiveExec            : "org.apache.hive:hive-exec:" + versions.hive,
```
HappyRay: Shouldn't it come before hiveMetastore?

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
HappyRay: Why is this approach better?

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    final String dbName = "test_db";
+    final String description = "test database";
+    final String location = "file:/tmp/" + dbName;
+    Database db = new Database(dbName, description, location, null);
+
+    msc.dropDatabase(dbName, true, true);
+    msc.createDatabase(db);
+    db = msc.getDatabase(dbName);
+    Assert.assertEquals(db.getName(), dbName);
```
HappyRay: Consider using assertJ. 
It would be good that new code standardizes on assertJ.

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
HappyRay: I saw you had a cleanup call in the beginning of the test already. Is it the idea to be able to inspect the state after the test fails and still be able to run the next test run with a clean state?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+
+    final HiveConf hiveConf = new HiveConf();
+    hiveConf.setVar(HiveConf.ConfVars.METASTOREURIS, hcatURI);
+    hiveConf.set(HIVE_METASTORE_TOKEN_SIGNATURE, hcatURI);
+
+    return hiveConf;
+  }
+
+  private IMetaStoreClient createMetaStoreClient() throws MetaException {
+    final HiveMetaHookLoader hookLoader = tbl -> {
```
HappyRay: I am not familiar with this hive code. Is there a common utility function or an overload that will do this?  It doesn't look like it is specific to our use case.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * An implementation of {@link BasePooledObjectFactory} for {@link IMetaStoreClient}.
+ */
+public class HiveMetaStoreClientFactory extends BasePooledObjectFactory<IMetaStoreClient> {
+
+  public static final String HIVE_METASTORE_TOKEN_SIGNATURE = "hive.metastore.token.signature";
+  private static final Logger LOG = LoggerFactory.getLogger(HiveMetaStoreClientFactory.class);
+  private final HiveConf hiveConf;
+
+  public HiveMetaStoreClientFactory() {
```
HappyRay: Why having three overloaded constructors when we only use one?

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
suvodeep-pyne: @HappyRay yes. That was my motivation.

azkaban-hadoop-security-plugin/build.gradle
suvodeep-pyne: Main cause: They are not present in maven central or jcenter.
Why we don't care? These modules are not required to compile the code or run the test cases.

azkaban-hadoop-security-plugin/build.gradle
suvodeep-pyne: I don't want to duplicate each dependency. IMO, this is better than maintaining 2 duplicate dependency lists. That exposes us to cases like what if somebody adds it to one and not the other for example.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: yes :) . It adds the copyright notice whenever I create a new class.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: Where is the convention mentioned? I use `log`. I find `log` to be shorter and yet equally expressive as `logger` but that's just my opinion. In that case the style guide should fail the build.

> Did you set up the checkstyle check?

Yes. I followed the development instructions around a week back. Did things change again?

suvodeep-pyne: This class was picked up from the Gobblin code base. Let me refactor the code a bit.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: yes.

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
suvodeep-pyne: Will take a look and update the PR.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: I am not sure about this. This was the suggestion from our internal teams to use this implementation. For now, we can go ahead with this and we can refactor if we find something that is easier.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: borrowed this code from Gobblin. Thought it may be useful since we have the same use case. but I agree this is more confusion than help. I'll get rid of them

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: The `HiveException` is inside a Lambda. I don't think you can propagate that directly. Correct me if I'm wrong here.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: See the description. The code has been copied from Gobblin.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: Regarding the `log` convention, I would just suggest using simply `log`
This is how lombok does it. link: https://projectlombok.org/features/log

HappyRay:
How can these dependencies be cached on your machine if they are not available in the repository?

azkaban-hadoop-security-plugin/build.gradle
HappyRay: This is odd. Did you check with Hive team?

HappyRay: Consider adding a comment in the code if this change needs to stay.

azkaban-hadoop-security-plugin/build.gradle
HappyRay: Don't you have the same problem if someone adds to compile a dependency this test module needs?

IMHO, it is better to stay with the default/convention to avoid surprises as much as possible.

If someone forgets to add a dependency, tests will fail. Right? Why isn't it good enough?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
HappyRay: I'd suggest you keep them separate in the future. It makes review harder. If you want a quicker review, keep the change as small and focused as possible.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: I saw the link to Gobblin. But I didn't click through.

HappyRay: It's not clear that Gobblin has to make these build changes. Should you have @erwa review this change?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: Do you have a reference to this behavior ( a test program, article etc?)

If this is the issue, how does "  throw new MetaException("Failed to get storage handler: " + e);" work then?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,90 @@
...
+public class HiveMetaStoreClientFactory extends BasePooledObjectFactory<IMetaStoreClient> {
+
+  private static final Logger log = LoggerFactory.getLogger(HiveMetaStoreClientFactory.class);
+  private final HiveConf hiveConf;
+
+  public HiveMetaStoreClientFactory(final HiveConf hiveConf) {
+    this.hiveConf = hiveConf;
+  }
+
+  private IMetaStoreClient createMetaStoreClient() throws MetaException {
```
HappyRay: I don't follow why this code is needed and what it is doing. 
Consider adding comments.

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.thrift.TException;
+import org.junit.Test;
+
+public class HiveMetaStoreClientFactoryTest {
+
+  public static final File METASTORE_DB_DIR = new File("metastore_db");
+  public static final File DERBY_LOG_FILE = new File("derby.log");
+
+  @Test
+  public void testCreate() throws TException, IOException {
```
HappyRay: How does the test verify the retry behavior?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: Either log or logger sounds good to me. The main problem is that it should be lower case. You can read the Google style doc. 
No, we haven't setup the proper style check instruction yet. Are you interested in helping?

HappyRay: How does Gobblin enforce code styles?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: Is the question clear?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: ?

azkaban-hadoop-security-plugin/src/test/java/azkaban/hive/HiveMetaStoreClientFactoryTest.java
HappyRay: Thanks for making the update. It will make the job of the reviewers easier if you leave a comment suggesting that the change is made. Otherwise, the reviewers may have a hard time remembering which ones have been addressed.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: Please follow up. Simpler code is better code. 
And we need to understand what the code does.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: I saw the change now. This is an example of why I found it hard to track which comment has been addressed. I wish Github has a feature like the Review Board has that can make it easier. Do you know such a feature in Github?

azkaban-hadoop-security-plugin/build.gradle
suvodeep-pyne: No, I don't think so. `testCompile` extends `compile` by default. But it doesn't seem to pull in from `compileOnly` which is why I added this line.

> IMHO, it is better to stay with the default/convention to avoid surprises as much as possible.

I find it intuitive to have the compileOnly classes available in my test classpath so that I can test my code! In fact, it is counter intuitive to me if my jar gets built successfully yet my tests don't compile.

Gradle configurations are built from file collections so that they can be extended, filtered and manipulated. So I don't think this should come as a surprise to a Gradle user. In fact, I think it might help to take a look at the task and configuration graph.
link: https://docs.gradle.org/3.3/userguide/java_plugin.html

> If someone forgets to add a dependency, tests will fail. Right? Why isn't it good enough?

This is good. But why do have to add it in 2 places is the question.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: The `MetaException` is part of the interface which the Lambda accepts. Pasting the interface below. 

```

/**
 * HiveMetaHookLoader is responsible for loading a {@link HiveMetaHook}
 * for a given table.
 */
public interface HiveMetaHookLoader {
  /**
   * Loads a hook for the specified table.
   *
   * @param tbl table of interest
   *
   * @return hook, or null if none registered
   */
  public HiveMetaHook getHook(Table tbl) throws MetaException;
}
```

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
suvodeep-pyne: I am not too concerned about this since it is simply instantiating a HMS client. But I'm totally open to getting some feedback on how we can do this better.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: I agree this is off context. I tried the other way and I see myself making less changes overall and it's counter productive for me. So, I would rather choose to make the code better and wait a little longer for the review. I tend to switch to other things and often forget / feel lazy to create a refactor patch before/after my code changes. 

I don't want to make a whole bunch of changes as well but definitely the ones that come to my eye. This one, for example, has been lurking around for a long time and I haven't touched it because I wanted to make a refactor PR that never happened.

suvodeep-pyne:
@HappyRay 
> How can these dependencies be cached on your machine if they are not available in the repository?
I remember making some extra effort in fetching dependencies from elsewhere. This was when I was working on the Hive codebase.

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: This just indicates that this method will throw this checked exception. Will it work if we change it to throws HiveExecption?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/HiveMetaStoreClientFactory.java
HappyRay: @erwa could you comment?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
HappyRay: "This one, for example, has been lurking around for a long time and I haven't touched it because I wanted to make a refactor PR that never happened." How can I help so that you can make it happen more often in the future? :) Any suggestion?

HappyRay:
Discussed off line, due to the risk of not having this fix, we agreed to check in this fix first. @suvodeep-pyne agreed to address the remaining review comments later.

HappyRay:
The test took almost 9 seconds to run on my mac. This is slow. Any idea how to speed it up?

HappyRay:
How did you test this feature work?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
HappyRay: I would take into consideration of the reviewers and more importantly the productivities of the future readers of this code and change history.

azkaban-hadoop-security-plugin/build.gradle
HappyRay: Modifying the sourceSets is not an issue and is expected. 
I thought extendsFrom would **replace** the existing superConfigs. After some experiments, I realized that it **adds** to the existing configs. 
This change is fine. 
Also since the build only packages the jar and not the run time dependencies, does it need to use compileOnly in the first place? What's your long term plan?

suvodeep-pyne: > Modifying the sourceSets is not an issue and is expected.

AFAIK, I didn't modify any source set. I don't understand the context here.

> I thought extendsFrom would replace the existing superConfigs. After some experiments, I realized that it adds to the existing configs.

Yes. It simply _extends_ the configuration/collection with another.

> Also since the build only packages the jar and not the run time dependencies, does it need to use compileOnly in the first place? What's your long term plan?

Correction: build packages the jar WITH the runtime dependencies. `compileOnly` means that this collection is only used for compiling the artifact and they are not added to runtime. It does not need to be `compileOnly` since HDFS storage introduced Hadoop as a compile level dependency at this time. But prior to that, Hadoop and Hive dependencies were not shipped with Azkaban.

The long term plan IMO should be to get rid of Hadoop, Hive etc dependencies from core Azkaban. HDFS Storage and Hadoop Security Manager code should run in a plugin module like environment.

suvodeep-pyne: @HappyRay 
Let me know if we should do probably a 30-min or 1-hour discussion on Gradle some time since a GitHub PR doesn't seem to be well suited for explaining Gradle intricacies.

suvodeep-pyne:
> The test took almost 9 seconds to run on my mac. This is slow. Any idea how to speed it up?

It is taking some seconds to run since it sets up a DB and brings up the Hive Metastore. One way of speeding it up would be to remove the cleanup before running the test. This would benefit second runs in which the HMS start up wouldn't have to create the local files every time.

suvodeep-pyne:
> How did you test this feature work?

I didn't test whether `RetryingHiveMetaStoreClient` works or not. This is a Hive class. I just tested whether existing Hive jobs are running fine. This would ensure that they are fetching the delegation tokens correctly.

azkaban-hadoop-security-plugin/build.gradle
HappyRay: Sounds good.
Sorry, I meant configurations, not sourceSet.

HappyRay:
That test is good for catching regressions. But it doesn't test the new functionality is working, does it?

Issue #717, opened 2016-08-29
Job in pipeline mode is blocked by another execution, where the job is disabled


juhoautio:
If a job is disabled for an execution, it seems that later execution for that job in pipeline mode is blocked for the entire flow execution. Job shouldn't be blocked on pipeline if the same job is not running or going to be run by a previously started execution.
## Steps to reproduce

Let's say we have a flow with two jobs in it `job_x -> job_y`
- Launch an execution, id=1, with `job_x` disabled
- Now `job_y` is running in execution 1
- Launch another execution, id=2, with `job_y` disabled
- Now `job_x` of execution 2 is "queued" and is said to be blocked by the same job in execution 1

The bug here is that `job_x` of 2 is blocked until execution 1 has finished completely, although execution 1 never runs `job_x`.

juhoautio:
Fixed by https://github.com/azkaban/azkaban/pull/1245

PR #1243, opened 2017-06-21
Fix and enable FlowRunnerPipelineTest

juhoautio:
Fix and enable FlowRunnerPipelineTest.

Also much less code and faster running.

Separate commits should offer some clarity:
1. first I just fixed the "app initialization" and test file path
1. then I converted the assertions to use the "blocking" versions that wait until the expected status is reached, and removed redundant sleeps
    - Here I also removed the sleep between starting the 1st and 2nd execution: the 2nd execution can be started when the first jobs of the 1st execution have reached RUNNING status. 
1. finally I also converted the flow level assertions to the "blocking" versions

PR #1244, opened 2017-06-21
Enable RemoteFlowWatcherTest

juhoautio:
Fixed failures. It's quick, too.

PR #1245, opened 2017-06-21
Fix bug stuck in pipeline mode

juhoautio:
Blocked by 2 other PRs, see https://github.com/azkaban/azkaban/pull/1248.

----

Fix bug where pipelined execution gets stuck

If you had executed a flow with some subflows disabled and then started another execution in pipeline mode, the next execution remained blocked forever by the jobs inside sub-flows, because they remained in READY state instead of being set to SKIPPED.

**This was a pretty nasty bug because for example next morning's scheduled run can get stuck after you've launched an execution with some jobs disabled on the previous day.**

I think it would make sense to clean this up also on the flow saving side: when subflow status is changed from DISABLED to SKIPPED, all jobs in the subflows would also be marked as SKIPPED.

Indeed was able to reproduce with the new test method `testRemoteFlowWatcherBlockingJobsLeftInReadyState` that it gets stuck – before this fix.

----

Seems like this also fixes the issue https://github.com/azkaban/azkaban/issues/717 "Job in pipeline mode is blocked by another execution, where the job is disabled" that I posted before. So that's the case where a disabled job is blocking a pipelined execution _until the blocking execution finishes all its enabled jobs_ (even when both executions are running on the same executor at the same time).

----

Builds on https://github.com/azkaban/azkaban/pull/1243 which made FlowRunnerPipelineTest run.

----

There's actually another ~bug that caused this to happen for us: it seems that sometimes azkaban-web sets the pipelineExecId to point to an execution that has finished hours ago (about 24 hours ago in our case – and also there had been another finished execution after that, why not pick that one – although it wouldn't make sense either because it had also finished a couple hours before the stuck execution started :P). That could be investigated as a separate issue, but as long as the fix in this PR is in place at least execution won't get stuck even if that happens. Also there's a valid case left any way for the RemoteFlowWatcher to be chosen: when the execution that azkaban-web chose for pipelineExecId happens to just finish before azkaban-executor starts running the new execution.

----

TODO: once the pre-requisite PRs have been merged, get this additional test into this PR: https://github.com/azkaban/azkaban/pull/1248. That test genuinely reproduces the problem: use a flow with subflows and disable some subflow. The jobs inside the subflow remain in READY state.

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
@@ -107,18 +107,13 @@ public class RemoteFlowWatcherTest {
 
     final EventCollectorListener eventCollector = new EventCollectorListener();
 
-    final File workingDir1 = setupDirectory();
+    final File workingDir1 = this.temporaryFolder.newFolder();
     final FlowRunner runner1 =
         createFlowRunner(workingDir1, loader, eventCollector, "exec1", 1, null,
             null);
-    // this doesn't have real significance here, but the case where some jobs are eventually left
-    // with READY status (although flow has a finished status) is when some jobs have been disabled
-    // if a subflow is skipped entirely, the subflow job is set to SKIPPED but all jobs inside it
-    // remain in READY state - TODO prove that with a test case that has subflows
-    runner1.getExecutableFlow().getExecutableNodePath("job4").setStatus(Status.DISABLED);
     final Thread runner1Thread = new Thread(runner1);
 
-    final File workingDir2 = setupDirectory();
+    final File workingDir2 = this.temporaryFolder.newFolder();
     final RemoteFlowWatcher watcher = new RemoteFlowWatcher(1, loader, 100);
     final FlowRunner runner2 =
         createFlowRunner(workingDir2, loader, eventCollector, "exec1", 2,
@@ -140,7 +135,7 @@ public class RemoteFlowWatcherTest {
     FileUtils.deleteDirectory(workingDir1);
     FileUtils.deleteDirectory(workingDir2);
 
-    testPipelineLevel2(runner1.getExecutableFlow(), runner2.getExecutableFlow(), true);
+    assertPipelineLevel2(runner1.getExecutableFlow(), runner2.getExecutableFlow(), true);
   }
 
   @Test
@@ -238,7 +233,8 @@ public class RemoteFlowWatcherTest {
       if (watchedNode == null) {
         continue;
       }
-      Assert.assertEquals(Status.SUCCEEDED, watchedNode.getStatus());
+      Assert.assertEquals(watchedNode.getStatus(),
+          job4Skipped && watchedNode.getId().equals("job4") ? Status.READY : Status.SUCCEEDED);
 
       long minDiff = Long.MAX_VALUE;
       for (final String watchedChild : watchedNode.getOutNodes()) {
@@ -246,7 +242,8 @@ public class RemoteFlowWatcherTest {
         if (child == null) {
           continue;
         }
-        Assert.assertEquals(child.getStatus(), Status.SUCCEEDED);
+        Assert.assertEquals(child.getStatus(),
+            job4Skipped && child.getId().equals("job4") ? Status.READY : Status.SUCCEEDED);
         final long diff = node.getStartTime() - child.getEndTime();
         minDiff = Math.min(minDiff, diff);
         Assert.assertTrue(
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
@@ -39,7 +39,6 @@ import azkaban.test.executions.ExecutionsTestUtil;
 import azkaban.utils.Props;
 import java.io.File;
 import java.io.IOException;
-import org.apache.commons.io.FileUtils;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -60,9 +59,6 @@ public class RemoteFlowWatcherTest {
     this.jobtypeManager =
         new JobTypeManager(null, null, this.getClass().getClassLoader());
     this.jobtypeManager.getJobTypePluginSet().addPluginClass("test", InteractiveTestJob.class);
-    Utils.initServiceProvider();
-    JmxJobMBeanManager.getInstance().initialize(new Props());
-    InteractiveTestJob.setQuickSuccess(true);
     Utils.initServiceProvider();
     JmxJobMBeanManager.getInstance().initialize(new Props());
     InteractiveTestJob.setQuickSuccess(true);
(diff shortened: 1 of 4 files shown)


juhoautio:
@HappyRay & co., I'd like to have this fix in use ASAP. What do you think, will you be able to review and get this merged on this side soon? I'd rather do it that way but can of course apply on our fork if not possible to prioritize here.

juhoautio:
Yeah, I got this proven with a unit test that runs a sub-flow. That's shared at #1248.

HappyRay:
@juhoautio Thanks for the pull requests! I haven't had time to review them in details yet. Based on your descriptions, I think they are good candidates to merge.

Also, thank you for making separate pull requests.

I don't have the bandwidth to review at this time. And I am going on vacation soon. We are also dealing with multiple production issues before the upcoming holiday. I've asked @ameyamk for help with resourcing.

azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
```diff
@@ -54,17 +55,21 @@
...
 
   @After
   public void tearDown() throws IOException {
+    InteractiveTestJob.resetQuickSuccess();
   }
 
   public File setupDirectory() throws IOException {
     System.out.println("Create temp dir");
-    final File workingDir = new File("_AzkabanTestDir_" + this.dirVal);
+    workingDir = new File("build/tmp/_AzkabanTestDir_" + this.dirVal);
```
HappyRay: Could you convert it to tempDir junit rule too?

azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
```diff
@@ -103,10 +107,51 @@ public void testBasicRemoteFlowWatcher() throws Exception {
...
+    final EventCollectorListener eventCollector = new EventCollectorListener();
+
+    final File workingDir1 = setupDirectory();
+    final FlowRunner runner1 =
+        createFlowRunner(workingDir1, loader, eventCollector, "exec1", 1, null,
+            null);
+    // this doesn't have real significance here, but the case where some jobs are eventually left
+    // with READY status (although flow has a finished status) is when some jobs have been disabled
+    // if a subflow is skipped entirely, the subflow job is set to SKIPPED but all jobs inside it
+    // remain in READY state - TODO prove that with a test case that has subflows
```
HappyRay: Is the TODO still valid?

azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
```diff
@@ -248,7 +296,7 @@ private FlowRunner createFlowRunner(final File workingDir, final ExecutorLoader
       final EventCollectorListener eventCollector, final String flowName, final int execId,
       final FlowWatcher watcher, final Integer pipeline, final Props azkabanProps)
       throws Exception {
-    final File testDir = new File("unit/executions/exectest1");
+    final File testDir = new File("../test/src/test/resources/azkaban/test/executions/exectest1");
```
HappyRay: Please use the wrapper method to get the file object.

HappyRay:
Am I right that 
#1243
#1244
need to be merged first?

HappyRay:
The code LGTM.

azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
juhoautio: I think this is already fixed in the pre-requisite PR that enables the RemoteFlowWatcherTest. Once that is merged I will update this PR.

azkaban-exec-server/src/test/java/azkaban/execapp/event/RemoteFlowWatcherTest.java
juhoautio: I think this is already fixed in the pre-requisite PR that enables the RemoteFlowWatcherTest. Once that is merged I will update this PR.

HappyRay:
@juhoautio what do you think about deprecating the support for level 1 and 2 pipeline modes and replace them with another option: run the flow after the previous one finishes. The current level1 and 2 pipeline modes are rarely used inside Linkedin and are not well understood and tested. This will greatly simplify the code. 
It will make the engine rewrite project simpler.

HappyRay:
I have started the DAG engine rewrite project by writing a prototype first. You can track it at https://github.com/HappyRay/azkaban/tree/dag-engine-refactoring

HappyRay:
Is it ok to merge this before #1243?

juhoautio:
> Is it ok to merge this before #1243?

Rather not, because #1243 provides the basis for having the unit test.

juhoautio:
Also `testBasicPipelineLevel1RunDisabledJobs` remains to be brought into this from https://github.com/azkaban/azkaban/pull/1248. I can do that once https://github.com/azkaban/azkaban/pull/1243 has been merged.

HappyRay:
Could you resolve the conflict?

juhoautio:
> what do you think about deprecating the support for level 1 and 2 pipeline modes and replace them with another option: run the flow after the previous one finishes.

I wouldn't remove that feature. It has been actually the default mode that we use. I would expect this to be generally useful, as long as users are aware of it.

The main reason for us is that we have some flows with a lot of jobs, and we don't want the next days run blocked entirely by the previous days flow if some jobs from that are running slow or are even stuck.

Another case where the job level pipelining is really useful is backfilling. A simple example is that there's a combination of a hive job and an SQL job that both take 1 hour to run and handle a single day. Consider also that there can only be one of each job running at the same time, so it's not possible to run the flow in parallel. Now, if you want to run this flow to backfill some 4 days, it can finish in 5 hours. If it would need to wait on flow level it would take 8 hours. The speed boost is of course bigger if there are more long-running sub-sequent jobs in the flow.

> The current level1 and 2 pipeline modes are rarely used inside Linkedin and are not well understood and tested.

It's true that it's a bit hard to understand. Maybe the terminology (level 1 / 2) could be deprecated, but keep the feature and give it more descriptive names. Like: `pipeline: flow level` and `pipeline: job level`?

> This will greatly simplify the code. It will make the engine rewrite project simpler.

I would tend to think that it should be possible to manage pipelining semi-individually – it shouldn't make the DAG processing code that much more complicated?

juhoautio:
> Could you resolve the conflict?

Yes, now that #1243 is in.

juhoautio:
Ready to roll again.

HappyRay:
> I wouldn't remove that feature. 

I understand. Thanks.

I plan to get some statistics on how this feature is used in Linkedin. And then decide if it is cost-effective to maintain this feature.

HappyRay:
> The main reason for us is that we have some flows with a lot of jobs, and we don't want the next days run blocked entirely by the previous days flow if some jobs from that are running slow or are even stuck.

Most of flows at Linkedin don't suffer from this problem. They can run concurrently. Is it difficult to modify your flows to allow them to run concurrently?

juhoautio:
Yes, it can be quite a lot of work.

PR #1246, opened 2017-06-21
fix end time schedule 

chengren311:
An issue was introduced in https://github.com/azkaban/azkaban/pull/1110 that causing schedule end time is set to a past date whenever SLA changes are made.
It's caused by incompatibility of new code https://github.com/azkaban/azkaban/pull/1110 with old schedule metadata. SLA change triggers reloading updated schedules into memory. But instead of using a default schedule end time, which we set to some day in year 2050, it always mistakenly uses a past date as the expiring time of the schedule. 
This fixes the issue by setting appropriate end time of the schedule.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
@@ -44,7 +44,7 @@ public class TriggerBasedScheduleLoader implements ScheduleLoader {
   private long lastUpdateTime = -1;
 
   public TriggerBasedScheduleLoader(final TriggerManager triggerManager,
-                                    final String triggerSource) {
+      final String triggerSource) {
     this.triggerManager = triggerManager;
     this.triggerSource = triggerSource;
   }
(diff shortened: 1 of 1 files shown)


HappyRay:
Is it a clean revert?

If so, please indicate so in the commit message.

reallocf:
This looks pretty good to me, but would like additional eyes @HappyRay

chengren311:
@HappyRay no, this is a patch of fix.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -44,7 +44,7 @@
   private long lastUpdateTime = -1;
 
   public TriggerBasedScheduleLoader(final TriggerManager triggerManager,
-                                    final String triggerSource) {
+      final String triggerSource) {
```
HappyRay: This looks like a formatting change.  Is it intentional? I though the earlier code was formatted properly by the save action.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -174,6 +176,14 @@ private BasicTimeChecker getBasicTimeChecker(final Map<String, ConditionChecker>
     return null;
   }
 
+  private BasicTimeChecker getEndTimeChecker(final Trigger t) {
```
shuzhang1989: will t be null?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
chengren311: yes. previously code was not formatted properly.

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
chengren311: nope

PR #1247, opened 2017-06-22
Ensure flow finished notification

juhoautio:
For example, if there would be an unexpected DB error in `updateFlow()`, ensure that flow finished event is sent any way. This is crucial in pipeline concurrent mode when `LocalFlowWatcher` is used, because it won't unblock the pipeline if it doesn't get this event. Also if this event is missed execution is left in the `runningFlows` map of `FlowRunnerManager`.


juhoautio:
Updated branch

PR #1250, opened 2017-06-23
fix submitted user name for trigger when deserializing from json blob

chengren311:
submitted user should be deserialized from json instead of a constant value.


HappyRay:
This doesn't look like a typo. Could you improve the change description?

chengren311:
@HappyRay updated

HappyRay:
Could you reference the pull request that introduced this bug?

chengren311:
@HappyRay added in the change description.

PR #1252, opened 2017-06-23
Disallowing users from logging in via URI QueryString

reallocf:
Related to #990 

Users currently can login by passing their credentials through the query string in the URI in this method:

`curl -X POST http://localhost:8081/?action=login\&username=azkaban\&password=azkaban`

This leads to sensitive information being logged in access logs and overall isn't a secure method of logging in.

This PR fails these attempts and returns an appropriate warning to users. It also stops the query string from being logged when these requests are made.

This PR includes testing to replicate this user behavior for future regression testing.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
@@ -353,14 +353,17 @@ public abstract class LoginAbstractAzkabanServlet extends
    * Disallows users from logging in by passing their
    * username and password via the request header where it'd be logged.
    *
-   * Login attempt is expected if the query string contains the word password and there is a
-   * parameter named password.
+   * This returns true if:
    *
-   * The query string will be empty if the password is passed incorrectly.
-   * The parameter map will contain a password for any login attempt.
+   * 1. The parameter map contains a key called password which indicates a login attempt.
+   * AND
+   * 2. The query string contains the word password meaning they passed their sensitive information
+   * via the query string instead of in the request body.
    *
-   * Returns false if req indicates an attempt to login.
-   * Otherwise returns true.
+   * Otherwise returns false.
+   *
+   * Example of illegal post request:
+   * curl -X POST http://localhost:8081/?action=login\&username=azkaban\&password=azkaban
    */
   private boolean isIllegalPostRequest(final HttpServletRequest req) {
     return (req.getParameterMap().containsKey("password") && req.getQueryString() != null &&
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
@@ -148,7 +148,7 @@ public abstract class LoginAbstractAzkabanServlet extends
     buf.append("\"");
     buf.append(req.getMethod()).append(" ");
     buf.append(req.getRequestURI()).append(" ");
-    if (req.getQueryString() != null && allowedPostRequest(req)) {
+    if (req.getQueryString() != null && !isIllegalPostRequest(req)) {
       buf.append(req.getQueryString()).append(" ");
     } else {
       buf.append("-").append(" ");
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -136,7 +136,8 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
   /**
    * Log out request - the format should be close to Apache access log format
    */
-  private void logRequest(final HttpServletRequest req, final Session session) {
+  private void logRequest(final HttpServletRequest req, final Session session,
+      final Boolean allowedQueryString) {
```
chengren311: why Boolean not boolean?

reallocf: I didn't know there was a difference 😄 haha. I'll switch it to boolean because that looks like it'll do the job. Thanks!

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -276,7 +277,13 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
       throws ServletException, IOException {
     Session session = getSessionFromRequest(req);
     this.webMetrics.markWebPostCall();
-    logRequest(req, session);
+    if (allowedPostRequest(req)) {
+      logRequest(req, session, true);
```
chengren311: i would prefer having allowedPostRequest inside logRequest so that we don't need add a boolean argument to logRequest.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -345,6 +352,26 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
...
+   *
+   * req.getQueryString() returns the string after the path in the URL.
+   * It is empty if data is passed in request body.
+   *
+   * Returns false if req.getQueryString() indicates an attempt to login.
+   * Otherwise returns true.
+   */
+  private boolean allowedPostRequest(final HttpServletRequest req) {
+    final String queryString = req.getQueryString();
+    if (queryString != null && queryString.contains("action=login")
```
chengren311: we can use req.getParameter instead of searching for string to do this.

chengren311: also this can be simplified in one line like:
return condA && condB && condC;

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -148,7 +148,7 @@ private void logRequest(final HttpServletRequest req, final Session session) {
     buf.append("\"");
     buf.append(req.getMethod()).append(" ");
     buf.append(req.getRequestURI()).append(" ");
-    if (req.getQueryString() != null) {
+    if (req.getQueryString() != null && allowedPostRequest(req)) {
```
chengren311: req.getQueryString() != null is not necessary since same check is already done in allowedPostRequest

reallocf: A && !(A && B && C)
--> A && (!A || !B || !C) de Morgan's law
--> (A && !A) || (A && !B) || (A && !C) distributive law
--> (A && !B) || (A && !C) because (A && !A) can never happen
--> A && (!B || !C) distributive law in reverse
And A && (!B || !C) != !(A && B && C) because 1, 0, 0 leads to the first being true and the second being false.
So we should need both. Does that make sense? (Also, yay boolean algebra! :smile:)

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -345,6 +349,24 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
...
+   * parameter named password.
+   *
+   * The query string will be empty if the password is passed incorrectly.
+   * The parameter map will contain a password for any login attempt.
+   *
+   * Returns false if req indicates an attempt to login.
+   * Otherwise returns true.
+   */
+  private boolean isIllegalPostRequest(final HttpServletRequest req) {
+    return (req.getParameterMap().containsKey("password") && req.getQueryString() != null &&
```
chengren311: would you mind adding more description in comment on why req.getQueryString().contains("password") is needed?

reallocf: :+1:

PR #1255, opened 2017-06-28
Refactor: Guice injection of Azkaban Web Server dependencies

suvodeep-pyne:
The current code uses the `SERVICE_PROVIDER` interface to fetch dependencies for the Web Server. Refactored the code to enable parameterized dependency injection from the construction which is the standard way of declaring dependencies. In the process, removed legacy constructor from `AzkabanWebServer` and removed legacy class `AzkabanServletContextListener` which is no longer being used.

Injection sanity tests already exist in the code. Tested solo server and it works as well.


azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
```diff
@@ -43,10 +44,10 @@
 
   private long lastUpdateTime = -1;
 
-  public TriggerBasedScheduleLoader(final TriggerManager triggerManager,
-      final String triggerSource) {
+  @Inject
+  public TriggerBasedScheduleLoader(final TriggerManager triggerManager) {
     this.triggerManager = triggerManager;
-    this.triggerSource = triggerSource;
+    this.triggerSource = ScheduleManager.triggerSource;
```
reallocf: How does Guice know to inject ScheduleManager here if it isn't a parameter of TriggerBasedScheduleLoader?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
```diff
@@ -111,9 +92,7 @@ public AzkabanServer getApplication() {
 
   @Override
   public void init(final ServletConfig config) throws ServletException {
-    this.application =
-        (AzkabanServer) config.getServletContext().getAttribute(
-            AzkabanServletContextListener.AZKABAN_SERVLET_CONTEXT_KEY);
+    this.application = SERVICE_PROVIDER.getInstance(AzkabanWebServer.class);
```
reallocf: Will we want to inject AzkabanWebServer in the future? Do you expect any complications when if we do so?

azkaban-common/src/main/java/azkaban/scheduler/TriggerBasedScheduleLoader.java
suvodeep-pyne: It doesn't. Actually `triggerSource` is a static variable. Let me fix this.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
```diff
@@ -28,9 +42,71 @@
...
   protected void configure() {
     bind(Server.class).toProvider(WebServerProvider.class);
     bind(AzkabanWebServer.class).in(Scopes.SINGLETON);
+    bind(ScheduleLoader.class).to(TriggerBasedScheduleLoader.class);
+  }
+
+  @Inject
+  @Singleton
+  @Provides
+  public UserManager loadUserManager(final Props props) {
```
reallocf: Some commenting here and for the function below would be useful - especially because most of us aren't guice wizards ⚡️

suvodeep-pyne: Ha Ha. Actually, this is just code moved from the `AzkabanWebServer` class to here. These methods are used by Guice to create these objects. Ref: https://github.com/google/guice/wiki/ProvidesMethods

See `AzkabanCommonModule` code for similar examples.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
suvodeep-pyne: I don't suspect that the class will be **injected** here. This should be a fetch since Guice maintains only a single instance of the web server class. Although, this may fail spectacularly if the injection happens while constructing the web server itself in which case this should throw up some sort of a circular dependency error. None of that seems to be happening though. Else the build should fail because of injection tests in `AzkabanWebServerTest`

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -142,53 +134,50 @@
...
+    this.server = requireNonNull(server);
+    this.executorManager = requireNonNull(executorManager);
+    this.projectManager = requireNonNull(projectManager);
+    this.triggerManager = requireNonNull(triggerManager);
+    this.metricsManager = requireNonNull(metricsManager);
+    this.registry = requireNonNull(metricRegistry);
+    this.sessionCache = requireNonNull(sessionCache);
+    this.userManager = requireNonNull(userManager);
+    this.scheduleManager = requireNonNull(scheduleManager);
+    this.velocityEngine = requireNonNull(velocityEngine);
```
reallocf: Should we check requireNonNull for all injected classes when we guice? Is this best practice for dependency injection?

suvodeep-pyne: Nope. This isn't about Guice. It's just defensive programming. 
It is good to check arguments to make sure that you have lesser surprises.

reallocf: That makes sense - do you think we should make that the standard for when we inject moving forward?

suvodeep-pyne: @reallocf 
short answer: yes.
long answer: This is not just about injection. If you are exposing public APIs or arguments, it is always better to state your requirements up front. This makes it fail fast and helps us figure out the problems earlier.

Issue #1256, opened 2017-06-28
prevent service users from running flows


li-afaris:
Currently the 'root' user is prevented from running flows.  This patch will enhance this feature by preventing a comma separated list of service users from running flows.  The comma separated list can be configured to setting 'users.to.blacklist=user1,user2,user3,...' in the commonprivate.properties file and defaults to 'root,azkaban'. 

A pull request will be arriving momentarily.

li-afaris:
Closing as this has been reviewed and merged a long time ago.

PR #1257, opened 2017-06-28
Issue #1256 - enabling config setting to prevent service users from running Azkaban flows

li-afaris:
As described in issue #1256, this will allow Azkaban operators to create a list of service accounts that Azkaban shall not run flows as.  Examples of service accounts would be found in /etc/passwd, (ie: mail, uucp, nobody).


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -203,16 +205,21 @@ public void run() throws Exception {
     String effectiveUser = null;
     final boolean isExecuteAsUser = this.sysProps.getBoolean(EXECUTE_AS_USER, true);
 
+    //Get list of users we never execute flows as. (ie: root, azkaban)
+    final String [] userBlackList = this.sysProps.getString(BLACK_LISTED_USERS, "root,azkaban").split(",");
```
reallocf: Creating a public static final String DEFAULT_BLACK_LISTED_USERS = "root,azkaban" up top might make it easier to read what's going on here at a quick glance and make adding to the default list a bit simpler.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -26,6 +26,7 @@
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.SystemMemoryInfo;
+import com.sun.xml.internal.fastinfoset.util.StringArray;
```
reallocf: Not needed :smile:

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
suvodeep-pyne: If this is a new configuration key, then you can add it to `Constants.ConfigurationKeys` instead. 

We don't yet have a good standard for placing defaults. Some are listed as `DEFAULTS_*` in `Constants class like `DEFAULT_SSL_PORT_NUMBER`.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -203,16 +205,21 @@ public void run() throws Exception {
...
     // which will change user from Azkaban to effectiveUser
     if (isExecuteAsUser) {
       final String nativeLibFolder = this.sysProps.getString(NATIVE_LIB_FOLDER);
       executeAsUserBinaryPath =
           String.format("%s/%s", nativeLibFolder, "execute-as-user");
       effectiveUser = getEffectiveUser(this.jobProps);
-      if ("root".equals(effectiveUser)) {
-        throw new RuntimeException(
-            "Not permitted to proxy as root through Azkaban");
+      for (String blackListedUser : userBlackList) {
```
suvodeep-pyne: Just make `userBlackList` a `Set` here and do a `contains()` check. 
nit: probably `blackListedUsers` instead of `userBlackList`?

reallocf: 👆 good idea with the Set

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
li-afaris: Intellij added this for me, and I didn't validate if it was needed. I went ahead and removed this import. Thanks for pointing this out.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
li-afaris: Thanks for the information about Constants.ConfigurationKeys.  I added the new key as suggested.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
li-afaris: I changed this to a Set as suggested.  Thanks

PR #1259, opened 2017-06-28
Fix: Azkaban does not fetch tokens from all HCAT servers

suvodeep-pyne:
The `IMetaStoreClientFactory` previously `HiveMetaStoreClientFactory` would always use the default `HiveConf` from the host machine to create a Metastore client. This is incorrect since, in order to fetch delegation tokens from different clusters, it needs to use the conf as per the remote cluster.

Changes
- removing the factory as a member and constructing it ad hoc in every case
- renamed the factory and removed the dependency on pooling factory (refactor)
- renamed the test and added an assert (refactor)

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/IMetaStoreClientFactory.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/IMetaStoreClientFactory.java
@@ -20,9 +20,7 @@ package azkaban.hive;
 
 import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.META_TABLE_STORAGE;
 
-import org.apache.commons.pool2.BasePooledObjectFactory;
-import org.apache.commons.pool2.PooledObject;
-import org.apache.commons.pool2.impl.DefaultPooledObject;
+import azkaban.spi.AzkabanException;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.HiveMetaHookLoader;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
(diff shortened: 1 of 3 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/IMetaStoreClientFactory.java
```diff
@@ -37,54 +35,37 @@
...
+  private static final Logger log = LoggerFactory.getLogger(IMetaStoreClientFactory.class);
   private final HiveConf hiveConf;
 
-  public HiveMetaStoreClientFactory(final HiveConf hiveConf) {
+  public IMetaStoreClientFactory(final HiveConf hiveConf) {
     this.hiveConf = hiveConf;
   }
 
-  private IMetaStoreClient createMetaStoreClient() throws MetaException {
+  public IMetaStoreClient create() {
```
sbeeram: Did you consider making HiveConf a parameter to create()? That way you can work with just one Factory instance.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -406,7 +403,7 @@ private void getToken(final String userToProxy) throws InterruptedException,
   private void cancelHiveToken(final Token<? extends TokenIdentifier> t)
       throws HadoopSecurityManagerException {
     try {
-      final IMetaStoreClient hiveClient = this.hiveMetaStoreClientFactory.create();
+      final IMetaStoreClient hiveClient = new IMetaStoreClientFactory(new HiveConf()).create();
```
sbeeram: It seems odd that cancelHiveToken is a local cluster operation. Do you know why thats the case?

azkaban-hadoop-security-plugin/src/main/java/azkaban/hive/IMetaStoreClientFactory.java
suvodeep-pyne: I know. I did that initially. Then I just decided to stick to the standard factory pattern and have a parameterless `create` method.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
suvodeep-pyne: Valid point. We currently don't have a mapping from token to a conf or a client. Thus, I think the cancellation of "other" delegation tokens should be broken. I created #1261 for this.

suvodeep-pyne:
scraping fix. DO NOT MERGE.

PR #1260, opened 2017-06-29
add unit test for #1253

chengren311:
1. add unit test for #1253
2. minor fix for ProcessJob to make getProgress() work. getProgress() will be used in the unit test.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -259,7 +259,6 @@ public class ProcessJob extends AbstractProcessJob {
         }
         throw new RuntimeException(e);
       } finally {
-        this.process = null;
         info("Process completed "
             + (success ? "successfully" : "unsuccessfully") + " in "
             + ((System.currentTimeMillis() - startMs) / 1000) + " seconds.");
--- a/azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
+++ b/azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
@@ -16,10 +16,16 @@
 
 package azkaban.jobExecutor;
 
+import static org.junit.Assert.assertEquals;
+
 import azkaban.flow.CommonJobProperties;
 import azkaban.utils.Props;
 import java.io.File;
 import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import org.apache.log4j.Logger;
 import org.junit.After;
 import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -259,7 +259,6 @@ public void run() throws Exception {
         }
         throw new RuntimeException(e);
       } finally {
-        this.process = null;
```
reallocf: Is there any value in setting the process to null like this?

azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
```diff
@@ -136,4 +142,55 @@ public void testPartitionCommand() throws Exception {
...
+    final ExecutorService executorService = Executors.newSingleThreadExecutor();
+    final Future future = executorService.submit(job);
+    Thread.sleep(2000);
+    try {
+      this.job.cancel();
+    } catch (final Exception e) {
+      this.log.error(e.getMessage());
+    }
+    future.get();
+    assertEquals(this.job.getProgress(), 0.0, 0.001);
```
suvodeep-pyne: nit: As @HappyRay suggested, can you try using assertj?

azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
```diff
@@ -136,4 +142,55 @@ public void testPartitionCommand() throws Exception {
...
+    public SleepBeforeRunJob(final String jobId, final Props sysProps, final Props jobProps,
+        final Logger log) {
+      super(jobId, sysProps, jobProps, log);
+    }
+
+    @Override
+    public void run() {
+      try {
+        info("sleep for 5 seconds before actually running the job");
+        Thread.sleep(5 * 1000);
```
suvodeep-pyne: Will this run for the entire 5 sec? I was wondering what happens if we cut the sleep times by a factor of 5. like 200ms and 1000ms.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
suvodeep-pyne: TL;DR This is beyond the scope of the PR. Mostly my thoughts regarding this class.

I think we need a better way of managing the status of a process here. Currently what we have is a `boolean` called `killed` but effectively what we need is some level of granularity and a way to express it - similar to `status`.
There are certain benefits of maintaining a proper `status` for a process like `NOT_STARTED`, `STARTED`, `COMPLETE` or `KILLED` or `FAILED`. The last 3 being terminal nodes; think of this like a state machine. Having a simple mapping between the job status to the process status may make things much easier. Isn't it? Because a job will fail if a process fails, etc.
Using the process as `null` or not `null` as signals may not be the most intuitive way of doing it. I think this class is more like a `ProcessLifeCycleManager` which creates, runs and manages the lifecycle of the process. Currently, it subclasses from `Job` which is weird. Maybe the `Job` should have `ProcessLifeCycleManager` in it since every job runs in its own process. We should probably think of how we can restructure this code to make it simpler and more flexible.

chengren311: @reallocf i don't think it's necessary. It would make getProgress() return 0 instead 1 as expected when process finishes.

azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
chengren311: this is basically to simulate the scenario where job preparation is stuck. 200ms/1000ms is too short for this purpose but i think 3secs is a reasonable number. let me change it to 3secs.

azkaban-common/src/test/java/azkaban/jobExecutor/ProcessJobTest.java
chengren311: updated

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
chengren311: @suvodeep-pyne thanks for your input. I think we can simplify the life cycle further by having only 2 status: RUNNING, NOT_RUNNING and let jobrunner to take care of other status like killed/complete/failed.

PR #1262, opened 2017-06-30
Update doc for executing flow with overided property

chengren311:


li-afaris:
Thanks for pointing out this feature with documentation.   LGTM.

PR #1269, opened 2017-07-10
Revert "Use RetryingHiveMetaStoreClient instead of HiveMetaStoreClient (#1235)

suvodeep-pyne:
The PR #1235 (commit 68e507208941659026ac15a1b242364b25f5fa31) introduces a bug which breaks Azkaban's capability to handle multiple HCAT servers. More specifically #1235 introduces a factory class for the Hive Client and caches the `HiveConf` during construction. When fetching delegation tokens from other HCAT servers, it accidentally uses the cached configuration instead of creating a new HiveConf specific to the target HCAT server. Hence it ends up fetching tokens from `HIVE_HOME` configured HCAT server instead of the target HCAT server leading to SASL errors.

Reverting the change. Confirmed that `HadoopSecurityManager_H_2_0` is identical to the version prior to #1235


suvodeep-pyne:
@HappyRay added.

PR #1270, opened 2017-07-10
Fix email sending failures

jamiesjc:
Users have reported some intermittent failures of sending emails. In order to fix that, we need to enhance our logic of exception handling and retry mechanism: 
1. When an exception is thrown during sending email messages, the SMTP connection needs to be closed properly. Otherwise it could cause potential issues to future SMTP connections.
2. Increase the retry count to 3 and timeout to 30 seconds for connectToSMTPServer and sendMessage.
3. Add more detailed logs for debugging sending email issues.

Tested the changes on solo server.
1. Reproduced the SocketTimeoutException issue by modifying the mail.host to some other value in azkaban.properties and verified the retry mechanism for connectToSMTPServer. Modified the email recipient address to verify the retry mechanism for sendMessage.
2. use "netstat -a | grep "smtp"" to verify the SMTP connection is always closed properly even when an exception happens.


azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
```diff
@@ -234,24 +234,35 @@ public void sendEmail() throws MessagingException {
...
-          this.logger.info("Email retry on SocketTimeoutException succeeded");
-        } catch (final MessagingException me) {
-          this.logger.error("Email retry on SocketTimeoutException failed", me);
-          throw me;
-        }
-      } else {
-        this.logger.error("Encountered issue while connecting to email server", ste);
-        throw ste;
+    int attempt;
+    for (attempt = 0; attempt < MAX_EMAIL_RETRY_COUNT; attempt++) {
```
reallocf: Breaking this into two helper methods (one for connecting to SMTP server and one for sending the message) might make this a bit more readable.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -56,10 +56,10 @@ public Emailer(final Props props) {
     this.mailSender = props.getString("mail.sender", "");
     this.tls = props.getString("mail.tls", "false");
 
-    final int mailTimeout = props.getInt("mail.timeout.millis", 10000);
+    final int mailTimeout = props.getInt("mail.timeout.millis", 30000);
     EmailMessage.setTimeout(mailTimeout);
     final int connectionTimeout =
-        props.getInt("mail.connection.timeout.millis", 10000);
+        props.getInt("mail.connection.timeout.millis", 30000);
```
jamiesjc: Yes, it will try 3 times with 30 sec timeout interval, so the total would be 90 sec.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -56,10 +56,10 @@ public Emailer(final Props props) {
     this.mailSender = props.getString("mail.sender", "");
     this.tls = props.getString("mail.tls", "false");
 
-    final int mailTimeout = props.getInt("mail.timeout.millis", 10000);
+    final int mailTimeout = props.getInt("mail.timeout.millis", 30000);
```
jamiesjc: We also get exceptions like "java.net.SocketTimeoutException: Read timed out". In this case, if we extend the timeout, it would increase the chance of a successful retry.

PR #1275, opened 2017-07-12
Flow 2.0 Stubs. Reading YAML flow files

suvodeep-pyne:
This patch introduces classes to deserialize YAML files. The `Bean` classes are used by `snake-yml` framework to read data. They are read in and translated into `AzkabanFlow` and `AzkabanJob` objects. Also included tests to validate the code.

Note: These code paths will not be exercised yet.


reallocf:
This is really cool :+1:

azkaban-common/src/test/resources/sample_flow20_01/sample_flow20.yml
```diff
@@ -0,0 +1,41 @@
...
+    # List the dependencies of the job
+    dependsOn:
+      - shell_pwd
+      - shell_echo
+      - shell_bash
+
+  - name: shell_echo
+    # Describe the type of the job
+    type: command
+    config:
```
kunkun-tang: I'm wondering why we use a nested config here.  Why not just put `command:` at the same level of `type:` ?

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
```diff
@@ -0,0 +1,73 @@
...
+
+package azkaban.project;
+
+import azkaban.utils.Props;
+import com.google.common.collect.ImmutableMap;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class AzkabanFlow extends AzkabanNode {
```
kunkun-tang: Do we store yml file to database? as part of zip package?

In other words, when AZ loads the yml file, where will it be stored?

azkaban-common/src/main/java/azkaban/project/NodeBean.java
```diff
@@ -0,0 +1,69 @@
...
+
+package azkaban.project;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The node bean is used by the YAML loader to deserialize DAG nodes
+ */
+public class NodeBean {
```
kunkun-tang: Have you considered the updating properties cases? People update properties in AZ web UI, and its yml file should be updated as well?

azkaban-common/src/test/resources/sample_flow20_01/sample_flow20.yml
```diff
@@ -0,0 +1,41 @@
+---
+# All flow level properties here
+config:
+  flow-level-parameter: value
+
+# This section defines the list of jobs
+# A node can be a job or a flow
+# In this example, all nodes are jobs
+nodes:
```
kunkun-tang: I remember @chengren311 was proposing some condition flow check. Is it easy for him to extend on this yml file?

suvodeep-pyne: Yes. You can refer the design doc.

azkaban-common/src/test/resources/sample_flow20_01/sample_flow20.yml
suvodeep-pyne: This is explained in the doc as well.
This allows us to categorize any job specific configuration inside a `config` section as opposed to the top level which is read by the Azkaban Framework.

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
suvodeep-pyne: `YML` files are intended to be stored in a zip. They define a flow similar to a job file today which defines a job.

azkaban-common/src/main/java/azkaban/project/NodeBean.java
suvodeep-pyne: Those will come later. I don't foresee that to be an issue. The current change is just to be able to read YML flows.

suvodeep-pyne:
@kunkun-tang This code is a POC code. Expect this to go through a lot of change. And as I mentioned earlier, this code will not be executed by Azkaban currently.

suvodeep-pyne:
@kunkun-tang 
Why do you think this should be in a new module? This is very closely tied to the project code.

azkaban-common/src/main/java/azkaban/project/AzkabanNode.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+package azkaban.project;
+
+import static java.util.Objects.requireNonNull;
+
+import azkaban.utils.Props;
+
+public abstract class AzkabanNode {
```
chengren311: does this have to be abstract class given no abstract method is present here?

suvodeep-pyne: The class is `abstract` by design. The idea is that a DAG `Node` can be either a `Job` or an embedded `Flow` or may be something else. Here job and flow are concrete classes but a node technically doesn't mean anything. Hence it is an abstract class.

##### Why is it not an interface?
It could be! Just that I think there might be some common code between a Job and a Flow which we can house here.

PR #1276, opened 2017-07-12
Metrics refactor

suvodeep-pyne:
Changes:
- The `MetricsManager` is the point of contact for any emission of metrics from Azkaban. The current code directly accesses the `MetricsRegistry` class from different places.
- The `MetricsUtility` class has been removed. The static methods have been converted to instance methods in MetricsManager.
- All instances of accessing `MetricsRegistry` directly have been removed except test code.
- Removed dead code: `ConsoleReporter` was never used


azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -40,28 +40,41 @@
 @Singleton
 public class MetricsManager {
 
-  private static final Logger logger = Logger.getLogger(MetricsManager.class);
+  private static final org.slf4j.Logger log = LoggerFactory.getLogger(MetricsManager.class);
```
reallocf: Would it be better to import org.slf4j.Logger as well?

suvodeep-pyne: yup. my bad. fixed.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -40,28 +41,41 @@
 @Singleton
 public class MetricsManager {
 
-  private static final Logger logger = Logger.getLogger(MetricsManager.class);
+  private static final Logger log = LoggerFactory.getLogger(MetricsManager.class);
```
kunkun-tang: Are we migrating to slf4j recently? I'm not aware of that.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -40,28 +41,41 @@
...
     this.registry.register("GC_Gauge", new GarbageCollectorMetricSet());
     this.registry.register("Thread_State_Gauge", new ThreadStatesGaugeSet());
   }
 
   /**
-   * Return the Metrics registry.
+   * A {@link Meter} measures the rate of events over time (e.g., “requests per second”).
+   * Here we track 1-minute moving averages.
+   */
+  public Meter addMeter(final String name) {
```
kunkun-tang: +1 on moving static method to here. We could modify the comment following the actual method below `getFifteenMinuteRate `

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -74,51 +88,21 @@ public synchronized void startReporting(final String reporterName, final Props p
...
-    }
-
-  }
-
-  /**
-   * Create a ConsoleReporter to the AZ Metrics registry.
-   *
-   * @param reportInterval time to wait between dumping metrics to the console
-   */
-  public synchronized void addConsoleReporter(final Duration reportInterval) {
```
kunkun-tang: Console Reporter was intended to help debug metric related code, even though we haven't tried it yet.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
suvodeep-pyne: Yes. That's what @HappyRay is pushing for as well. And I think it has some advantages since we already have it in our dependency.

azkaban-common/src/main/java/azkaban/metrics/MetricsManager.java
suvodeep-pyne: In such cases, probably we should expose a config parameter that can enable it. Since we are not using it I am thinking of removing it. let me know if you want to restore it. 

We can also probably add that as a reporting feature separately.

kunkun-tang: I was also thinking if metrics test could use it, but so far looks like it is not valued.

PR #1277, opened 2017-07-17
make passwordPlaceHolder and password error message configurable

kunkun-tang:
Some Azkaban instance needs 2FA authentications. In that case, the password field in login UI doesn't only include user's password. We modified the UI and make password placeholder configurable, such that Azkaban admins will be able to define.

Tested on localhost.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
@@ -105,7 +105,7 @@ public abstract class AbstractAzkabanServlet extends HttpServlet {
     this.name = props.getString("azkaban.name", "");
     this.label = props.getString("azkaban.label", "");
     this.color = props.getString("azkaban.color", "#FF0000");
-    this.passwordPlaceholder = props.getString("azkaban.password.placeholder", "password");
+    this.passwordPlaceholder = props.getString("azkaban.password.placeholder", "Password");
 
     if (this.application instanceof AzkabanWebServer) {
       final AzkabanWebServer server = (AzkabanWebServer) this.application;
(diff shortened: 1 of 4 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
```diff
@@ -103,6 +105,7 @@ public void init(final ServletConfig config) throws ServletException {
     this.name = props.getString("azkaban.name", "");
     this.label = props.getString("azkaban.label", "");
     this.color = props.getString("azkaban.color", "#FF0000");
+    this.passwordPlaceholder = props.getString("azkaban.password.placeholder", "password");
```
jamiesjc: Should the string be "Password" with capital letter "P" ?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
```diff
@@ -58,6 +58,8 @@
   private static final String AZKABAN_FAILURE_MESSAGE =
       "azkaban.failure.message";
   private static final long serialVersionUID = -1;
+
+  protected String passwordPlaceholder;
```
jamiesjc: Why passwordPlaceholder is added in AbstractAzkabanServlet instead of LoginAbstractAzkabanServlet?

kunkun-tang: The reason is that props is a local variable, and can only be seen in AbstractAzkabanServlet.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/AbstractAzkabanServlet.java
kunkun-tang: Thanks for find. Changing.

PR #1279, opened 2017-07-20
Add metrics to indicate flow dispatch failure/success count

jamiesjc:


azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -34,6 +34,8 @@
   private final MetricsManager metricsManager;
   private Meter dbConnectionMeter;
   private Meter flowFailMeter;
+  private Meter dispatchFailMeter;
```
suvodeep-pyne: I think we should have a better way of organizing metrics. Keeping all metrics together in one class seems like putting apples and oranges in one basket. Also, such organization of metrics hints that metrics events can be declared from anywhere else in the code. However, dispatch fail / success should only be controlled by the `ExecutorManager`. Thoughts?

kunkun-tang: Sounds a not bad idea, but it is difficult to do at today's codebase, since Executor still exists in common module.

HappyRay: I agree with @suvodeep-pyne.

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -34,6 +34,8 @@
...
+  private Meter dispatchSuccessMeter;
```
HappyRay: Aren't these metrics web server only? If so, should they be in azkaban.webapp.WebMetrics?

jamiesjc: These metrics are for executorManager, and executorManager is in azkaban-common. Unless we refactor executorManager and put it in web server, I cannot think of a way to do that properly.

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
jamiesjc: I agree. All the metrics should be refactored in a separate pull request.

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
HappyRay: I see. Thanks

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
suvodeep-pyne: Oh cool. @jamiesjc, In that case, should we just retain these Meters in `ExecutorManager` then instead? The existing metrics can be refactored separately in a different PR.

PR #1280, opened 2017-07-20
Delete execution directory on flow completion.

reallocf:
Right now we store the project information, all user-generated intermediary files, and the job logs in the /executions/<execution_id> on the executor. The job logs have already been moved to the database and the project information will be re-copied for each subsequent execution, so storing this data on disk isn't necessary.

This change should hopefully free up some disk space. The only downside is we won't be able to get user's flow logs if they do not get uploaded to the database due to some error, but after discussing with @HappyRay we agreed that it's worth it to reclaim the disk space.

Tested on solo-server with both normal flows and nested flows.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -59,6 +59,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
+import org.apache.commons.io.FileUtils;
 import org.apache.log4j.Appender;
 import org.apache.log4j.FileAppender;
 import org.apache.log4j.Layout;
@@ -221,6 +222,12 @@ public class FlowRunner extends EventHandler implements Runnable {
       this.logger.info("Setting end time for flow " + this.execId + " to "
           + System.currentTimeMillis());
       closeLogger();
+      this.logger.info("Deleting execution directory " + this.execDir.toString());
+      try {
+        FileUtils.deleteDirectory(this.execDir);
+      } catch (final Exception IOException) {
+        this.logger.error("Failed to delete execution directory " + this.execDir.toString());
+      }
 
       updateFlow();
       this.fireEventListeners(Event.create(this, Type.FLOW_FINISHED, new EventData(this.flow)));


suvodeep-pyne:
@reallocf 
> The only downside is we won't be able to get user's flow logs if they do not get uploaded to the database due to some error, 
Is there a way to ensure that the delete only happens if we have a successful database copy?

reallocf:
@suvodeep-pyne have any ideas on how to do this? I'm worried about adding complexity by having to pass around whether or not logs were uploaded successfully - particularly for nested flows.

@HappyRay what do you think? When we talked offline it seemed like this risk was worth it given how rarely database uploads fail, do you still think that's true?

chengren311:
@reallocf thanks for doing this. What is the current way of cleaning execution dirs? As far as I could remember, they are cleaned up by cleaner thread inside flowrunnermannger. And it's a polling based mode which cleans up every 1 hour on directories not been modified for 1 day.  
Why not just making the upper limit(1 day) shorter?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -221,6 +222,12 @@ public void run() {
       this.logger.info("Setting end time for flow " + this.execId + " to "
           + System.currentTimeMillis());
       closeLogger();
+      this.logger.info("Deleting execution directory " + this.execDir.toString());
```
HappyRay: Have you tested that this change works with embedded flows?

HappyRay:
I am ok with delaying adding more complexity until there is evidence that it is a big problem. I haven't seen such evidence yet. If DB is down for an extended period of time, we have a bigger problem.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
reallocf: Yes! This still works with embedded flows :+1:

reallocf:
After speaking offline with @suvodeep-pyne, we agreed it'd be better to set a configurable TTL for these files instead of deleting them. While we don't use these files often in our debugging (unless the problem is with uploading logs files to the DB) our users might use the /execution/<exec_id> directory to do their own testing then view the files via the HDFS viewer. While I don't think we have numbers regarding how often this is done, I can imagine users doing this.

I'll start working on the configurable TTL today, likely using the cleaner thread. @chengren311 I might bug you about how the thread works as I look more into it :smile:

reallocf:
@HappyRay how does that sound to you?

reallocf:
No longer going about this problem this way, so I'm closing this PR and opening another one with changes discussed offline.

PR #1281, opened 2017-07-20
Support for HDFS Quota Metrics 

suvodeep-pyne:
This patch provides the ability for Azkaban to monitor and report (via the internal Metrics framework) HDFS quota stats for a configured directory and also a polling interval. By default, the polling interval is 1 min.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/metrics/HdfsMetrics.java
+++ b/azkaban-common/src/main/java/azkaban/metrics/HdfsMetrics.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/metrics/HdfsMetrics.java
```diff
@@ -0,0 +1,117 @@
...
+        AZKABAN_STORAGE_HDFS_METRICS_ROOT_URI + " must be set.")
+        .getPath();
+    this.interval = config.getHdfsMetricsPollInterval();
+
+    log.info(String.format("Initializing HdfsMetrics with path: %s interval: %d ms",
+        this.path, this.interval.toMillis()));
+  }
+
+  public void registerMetrics() {
+    startFetcherThread();
```
HappyRay: Should the thread be started after the gauges are registered?

suvodeep-pyne:
Pausing this for now.

PR #1282, opened 2017-07-20
Updated documentation for block project uploads property.

rajagopr:
Updated documentation for property 'lockdown.upload.projects' to reflect the new behavior.


rajagopr:
Tested the changes locally by starting the gh-pages server.

PR #1283, opened 2017-07-20
fix guice initialize TriggerManager twice bug

kunkun-tang:
This bug was found in local solo server built by July 19th's master code. Our integration test tried to add trigger, which run every minutes. The strange observation is that this trigger never run. We debugged the code and found that the [TriggerScannerThread](https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java#L67) was launched twice, and newly added triggers were put into the should-not-exist thread and never run. I noticed that TriggerManager did not bind to Singleton in Guice Conf. I made this change, and the thread was not launched twice. The bug disappeared.

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
@@ -22,12 +22,14 @@ import static azkaban.executor.ExecutorManager.AZKABAN_USE_MULTIPLE_EXECUTORS;
 import static java.util.Objects.requireNonNull;
 import static org.apache.commons.io.FileUtils.deleteQuietly;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


kunkun-tang:
Thanks for reviewing. Merging.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -111,6 +113,11 @@ public void testInjection() throws Exception {
     executorLoader.updateExecutor(executor);
 
     assertNotNull(injector.getInstance(AzkabanWebServer.class));
+
+    //Test if triggermanager is singletonly guiced. If not, the below test will fail.
+    final TriggerManager triggerManager1 = requireNonNull(injector.getInstance(TriggerManager.class));
```
HappyRay: requireNonNull seems odd. Why not assert the instance is not null using junit assert or assertj?

PR #1287, opened 2017-07-24
Add metrics for sending email successs/failure.

jamiesjc:
Tested the metrics output/ingraph on localhost.
Guicified Emailer class and modified related test cases.
//TODO: refactor ExecutorManangerTest

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -23,6 +23,7 @@ import azkaban.executor.ExecutionOptions;
 import azkaban.executor.Status;
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.executor.mail.MailCreator;
+import azkaban.metrics.CommonMetrics;
 import azkaban.sla.SlaOption;
 import java.util.ArrayList;
 import java.util.List;
@@ -34,6 +35,7 @@ public class Emailer extends AbstractMailer implements Alerter {
   private static final String HTTPS = "https";
   private static final String HTTP = "http";
   private static final Logger logger = Logger.getLogger(Emailer.class);
+  private final CommonMetrics commonMetrics;
   private final String scheme;
   private final String clientHostname;
   private final String clientPortNumber;
@@ -46,8 +48,9 @@ public class Emailer extends AbstractMailer implements Alerter {
   private final String tls;
   private boolean testMode = false;
 
-  public Emailer(final Props props) {
+  public Emailer(final Props props, final CommonMetrics commonMetrics) {
     super(props);
+    this.commonMetrics = commonMetrics;
     this.azkabanName = props.getString("azkaban.name", "azkaban");
     this.mailHost = props.getString("mail.host", "localhost");
     this.mailPort = props.getInt("mail.port", DEFAULT_SMTP_PORT);
@@ -101,8 +104,10 @@ public class Emailer extends AbstractMailer implements Alerter {
       if (!this.testMode) {
         try {
           message.sendEmail();
+          this.commonMetrics.markSendEmailSuccess();
         } catch (final MessagingException e) {
           logger.error("Failed to send SLA email message" + slaMessage, e);
+          this.commonMetrics.markSendEmailFail();
         }
       }
     }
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -46,8 +48,9 @@
   private final String tls;
   private boolean testMode = false;
 
-  public Emailer(final Props props) {
+  public Emailer(final Props props, final CommonMetrics commonMetrics) {
     super(props);
+    this.commonMetrics = commonMetrics;
```
kunkun-tang: I'm revisiting this design. I was thinking if we should have requireNonNull to wrap commonMetrics. However, if we do that way, other program (external MP) can not use Emailer easily. I'm thinking if we could let contructor doesn't include the `commonMetrics`? Does it make sense?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -46,8 +48,9 @@
   private final String tls;
   private boolean testMode = false;
 
-  public Emailer(final Props props) {
+  public Emailer(final Props props, final CommonMetrics commonMetrics) {
```
kunkun-tang: Why not use Guice?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -72,8 +72,9 @@ private ExecutorManager createMultiExecutorManagerInstance(
 
     loader.addExecutor("localhost", 12345);
     loader.addExecutor("localhost", 12346);
-    return new ExecutorManager(props, loader, new AlerterHolder(props),
-        new CommonMetrics(new MetricsManager(new MetricRegistry())));
+    final CommonMetrics commonMetrics = new CommonMetrics(new MetricsManager(new MetricRegistry()));
```
kunkun-tang: Should we iniatialize guice `@before` any test, so that we can use Guice to get ExecutorManager?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
kunkun-tang: I just realized I directly use the class `EmailMessage`, rather than Emailer. It is safe to have Metrics in the constructor.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
jamiesjc: Refactored using Guice, thanks!

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -74,14 +77,16 @@ public void testSendEmail() throws Exception {
     Assert.assertNotNull(flow);
 
     final ExecutableFlow exFlow = new ExecutableFlow(this.project, flow);
-    final Emailer emailer = new Emailer(this.props);
+    final CommonMetrics commonMetrics = new CommonMetrics(new MetricsManager(new MetricRegistry()));
```
kunkun-tang: Should CommonMetrics be placed as a private member as it doesn't change? Just a minor comment. Either way works.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -118,15 +122,13 @@ public void testLocalExecutorScenario() throws ExecutorManagerException {
    */
   @Test
   public void testMultipleExecutorScenario() throws ExecutorManagerException {
-    final Props props = new Props();
-    props.put(ExecutorManager.AZKABAN_USE_MULTIPLE_EXECUTORS, "true");
+    this.props.put(ExecutorManager.AZKABAN_USE_MULTIPLE_EXECUTORS, "true");
```
kunkun-tang: In future refactor, I think we can have a setup method to have all properties. When some @Test doesn't need, we can do props.delete().

jamiesjc: Agree, thanks.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -186,9 +201,11 @@ public void sendSuccessEmail(final ExecutableFlow flow) {
     if (mailCreated && !this.testMode) {
       try {
         message.sendEmail();
+        this.commonMetrics.markSendEmailSuccess();
       } catch (final MessagingException e) {
         logger.error("Failed to send success email message for execution " + flow.getExecutionId(),
             e);
+        this.commonMetrics.markSendEmailFail();
```
HappyRay: Is there a way to centralize these metrics reporting? Inside sendEmail?

PR #1288, opened 2017-07-25
Fix value out of range exception when storing CANCELLED status in DB.

jamiesjc:
When running one of the integration tests - kill the flow right after it executes, we are seeing below error:

Caused by: java.sql.SQLException: Data truncation: Out of range value for column 'status' at row 1 Query: UPDATE execution_flows SET status=?,update_time=?,start_time=?,end_time=?,enc_type=?,flow_data=? WHERE exec_id=? Parameters: [130, 1500497708092, 1500497 ......

Status is TINYINT and the max value is 127, so trying to store value 130 in DB throws an exception. Changing CANCELLED from 130 to 125 fixes the problem.

This change is backward compatible since we never run into the situation where CANCELLED status was stored in DB in the past. This value has never been persisted. With the new change of KILLING status, it could happen that we need to store this CANCELLED status in DB now.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -33,9 +33,8 @@ public enum Status {
   DISABLED(100),
   QUEUED(110),
   FAILED_SUCCEEDED(120),
-  CANCELLED(130);
-  // status is TINYINT and in H2 DB the possible values are: -128 to 127
-  // so trying to store CANCELLED in H2 fails at the moment
+  CANCELLED(125);
+  // status is TINYINT in DB and the value ranges from -128 to 127
 
   private static final ImmutableMap<Integer, Status> numValMap = Arrays.stream(Status.values())
       .collect(ImmutableMap.toImmutableMap(status -> status.getNumVal(), status -> status));


HappyRay:
Could you add to the description why this change is backward compatible? I assume it is because this value should never have been persisted?

The previous error was not detected by unit tests. 
To prevent future regression, does it make sense to add a unit test to test writing all the possible job statuses into an in-memory h2 DB? 
e.g. when someone adds a new status with a value of 140 and somehow miss the comment, the test should fail.

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -33,9 +33,8 @@
   DISABLED(100),
   QUEUED(110),
   FAILED_SUCCEEDED(120),
-  CANCELLED(130);
-  // status is TINYINT and in H2 DB the possible values are: -128 to 127
-  // so trying to store CANCELLED in H2 fails at the moment
```
juhoautio: I think it's worth preserving the information that this limitation comes from H2. Otherwise one could check MySQL and be confused as it allows larger values.

HappyRay: According to this doc
https://dev.mysql.com/doc/refman/5.7/en/integer-types.html
https://stackoverflow.com/questions/12812362/mysql-tinyint-as-unsigned
The signed tynyint type has the same range. 
This is confirmed by @jamiesjc's tests which use mysql.

@juhoautio how did you reach the conclusion that mysql was different?

HappyRay:
According to this doc
https://dev.mysql.com/doc/refman/5.7/en/integer-types.html
The tinyint is not a standard sql type. 
Shall we move to a standard type such as smallint?

azkaban-common/src/main/java/azkaban/executor/Status.java
juhoautio: Oh, sorry. Probably I got confused with the max value on unsigned vs. signed tinyint.

jamiesjc:
Even though TINYINT is not a standard sql type, it is still supported by MySQL as an extension. Both TINYINT and BIGINT are widely used in our db table already. I would prefer keeping the existing sql type since they are working properly, otherwise, we need to change everywhere that is using TINYINT and BIGINT.

jamiesjc:
@kunkun-tang has refactored unit tests for JdbcExecutorLoader (https://github.com/azkaban/azkaban/pull/1291/files) and I will add unit tests for writing all possible statuses into H2 DB on top of his new changes.

azkaban-common/src/main/java/azkaban/executor/Status.java
HappyRay: Got it. Thanks

PR #1289, opened 2017-07-26
Fix for failure to kill the job caused by race condition.

jamiesjc:
When running integration test - killing a flow immediately after it starts, users encounter below exception intermittently and are not able to actually kill the flow, even though the status shows KILLING for some time and finally changes to KILLED. The status itself is misleading and users could miss the chance to retry killing the flow.

Logs:
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Kill has been called.
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Process has not yet started.
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Failed trying to cancel job. Maybe it hasn't started running yet or just finished.
25-07-2017 17:15:03 PDT IntegrationTest_jobCommand INFO - Process completed successfully in 10 seconds.
25-07-2017 17:15:03 PDT IntegrationTest_jobCommand INFO - Finishing job IntegrationTest_jobCommand attempt: 0 at 1501028103180 with status KILLED

Problem analysis:
1. run the process:
azkaban.jobExecutor.utils.process.AzkabanProcess#run
2. at the same time, kill the job:
azkaban.jobExecutor.ProcessJob#cancel
azkaban.jobExecutor.utils.process.AzkabanProcess#softKill 
azkaban.jobExecutor.utils.process.AzkabanProcess#checkStarted
here it throws exception, because the process just starts and hasn't reached java.util.concurrent.CountDownLatch#countDown yet, 
and thus this.startupLatch.getCount() != 0, which means process has not fully started yet. So the job couldn't be killed.

Solution:
By adding this.process.awaitStartup(), the kill thread will wait for ( startupLatch.await ) the run thread to decrement the count of the latch ( startupLatch.countDown ) and then release itself to continue with the kill.
Verified the fix with integration tests.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -374,6 +374,10 @@ public class ProcessJob extends AbstractProcessJob {
     if (this.process == null) {
       throw new IllegalStateException("Not started.");
     }
+
+    info("Wait for process to start.");
+    this.process.awaitStartup();
+
     final boolean processkilled = this.process
         .softKill(KILL_TIME.toMillis(), TimeUnit.MILLISECONDS);
     if (!processkilled) {


jamiesjc:
This fix won't work if azkaban.jobExecutor.ProcessJob#cancel is called, killed is set to true, and then process would skip azkaban.jobExecutor.utils.process.AzkabanProcess#run. Then azkaban.jobExecutor.utils.process.AzkabanProcess#awaitStartup will be waiting forever causing the problem. Need to figure out a better solution.

juhoautio:
Is the problem in ProcessJob only? If yes, sounds like it needs suitable synchronization between ProcessJob#run and ProcessJob#cancel, so that `ProcessJob.this.process` is not even created if the job has been cancelled.

jamiesjc:
Actually there could be several race conditions when killing the flow. One is between ProcessJob#run and Process#cancel as you mentioned above. Also another problem is in FlowRunner#kill(), it could happen that the JobRunner hasn't been added to the activeJobRunners yet when it tries to kill. In that case, the job wouldn't be killed as well.

HappyRay:
@jamiesjc It appears that the activeJobRunners list modification is synchronized with 
see azkaban.execapp.FlowRunner#runFlow and 
azkaban.execapp.FlowRunner#kill()
  public void kill() {

"it could happen that the JobRunner hasn't been added to the activeJobRunners yet when it tries to kill. In that case, the job wouldn't be killed as well."
Unless I miss something, it appears that the list won't be modified until the FlowRunner#kill returns.

jamiesjc:
In azkaban.execapp.FlowRunner#runFlow, before progressGraph, azkaban.execapp.FlowRunner#runReadyJob will modify activeJobRunners, and it is not synchronized. azkaban.execapp.FlowRunner#kill() could happen during runReadyJob and cause the race condition mentioned above.

HappyRay:
@jamiesjc Good analysis. This seems to be a valid race condition.

Do you mind filing an issue? Is the fix as simple as protecting this code with the same lock?

jamiesjc:
@HappyRay I've created an issue in https://github.com/azkaban/azkaban/issues/1311
I think adding the lock could solve the problem. I'll think about it and try it out, thanks!

HappyRay:
Thanks @jamiesjc !

PR #1291, opened 2017-07-26
refactor JDBC Execuloader class to leverage new azkaban-db

kunkun-tang:
Similar to  #1079 , this change is aimed at re-constructing JDBC Executor loader class to
allow db retry feature to be leveraged. Fixed all tests from outdated JdbcExecutorLoader class, and use in-memory h2 for test.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorImplTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorImplTest.java
@@ -60,7 +60,7 @@ public class JdbcExecutorImplTest {
     final String sqlScriptsDir = new File("../azkaban-db/src/main/sql/").getCanonicalPath();
     props.put("database.sql.scripts.dir", sqlScriptsDir);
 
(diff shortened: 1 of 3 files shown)


azkaban-common/src/test/java/azkaban/executor/JdbcExecutorImplTest.java
```diff
@@ -0,0 +1,632 @@
...
+  public void clearDB() {
+    try {
+      dbOperator.update("DELETE FROM executors");
+      dbOperator.update("DELETE FROM executor_events");
+      dbOperator.update("DELETE FROM execution_logs");
+      dbOperator.update("DELETE FROM execution_jobs");
+      dbOperator.update("DELETE FROM execution_flows");
+      dbOperator.update("DELETE FROM active_executing_flows");
+      dbOperator.update("DELETE FROM active_sla");
+    } catch (final SQLException e) {
```
HappyRay: Is the catch necessary? What if we allow the exception to bubble up?

PR #1292, opened 2017-07-26
Running chmod g+s on executions directory to enable proper execution cleanup

reallocf:
Setting the gid bit on the group for a directory causes all items created within that directory to inherit the group of the directory. So all items users create in their /execution/<exec_id> directory will automatically be a part of the azkaban group. This allows the azkaban cleanup thread to properly remove user-generated files/directories.

The gid bit is system-specific, so there is [no java standard library api](https://stackoverflow.com/questions/5045869/change-sticky-bit-with-java) for setting it. The solution I proposed spawns a subprocess that performs the chmod command. I don't think this is very clean, but I haven't been able to find a better way. Anybody have any ideas?

Another option would be to build this executionDirectory as part of our build process, but that isn't how we do things right now (tested by running deploy on holdem4 jenkins job without any of the other steps and verifying no /executions directory is created until system start).

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -214,7 +214,7 @@ public class FlowRunnerManager implements EventListener,
    * is system dependent, so the only way to set this bit is to start a new process and run
    * the shell command "chmod g+s " + execution directory name.
    *
-   * Note that this will only work on unix systems.
+   * Note that this should work on most Linux distributions and MacOS, but will not work on Windows.
    */
   private void setgidPermissionOnExecutionDirectory() throws IOException {
     logger.info("Creating subprocess to run shell command: chmod g+s "
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -162,6 +162,11 @@ public class FlowRunnerManager implements EventListener,
     this.executionDirectory = new File(props.getString("azkaban.execution.dir", "executions"));
     if (!this.executionDirectory.exists()) {
       this.executionDirectory.mkdirs();
+      // setting sticky bit on this.executionDirectory to allow cleanup thread to delete all
+      // user-generated files/directories
+      logger.info("Creating subprocess to run shell command: chmod g+s "
+          + this.executionDirectory.toString());
+      Runtime.getRuntime().exec("chmod g+s " + this.executionDirectory.toString());
     }
     this.projectDirectory = new File(props.getString("azkaban.project.dir", "projects"));
     if (!this.projectDirectory.exists()) {
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -162,11 +162,7 @@ public class FlowRunnerManager implements EventListener,
     this.executionDirectory = new File(props.getString("azkaban.execution.dir", "executions"));
     if (!this.executionDirectory.exists()) {
       this.executionDirectory.mkdirs();
-      // setting sticky bit on this.executionDirectory to allow cleanup thread to delete all
-      // user-generated files/directories
-      logger.info("Creating subprocess to run shell command: chmod g+s "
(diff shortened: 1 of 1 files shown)


reallocf:
Had conversation with Ray about this. I will be moving the change into an outside function and adding additional comments. I'll also be putting a little more thought into how to test this - possibly via the integration testing suite.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -162,6 +162,11 @@ public FlowRunnerManager(final Props props,
     this.executionDirectory = new File(props.getString("azkaban.execution.dir", "executions"));
     if (!this.executionDirectory.exists()) {
       this.executionDirectory.mkdirs();
+      // setting sticky bit on this.executionDirectory to allow cleanup thread to delete all
+      // user-generated files/directories
+      logger.info("Creating subprocess to run shell command: chmod g+s "
+          + this.executionDirectory.toString());
+      Runtime.getRuntime().exec("chmod g+s " + this.executionDirectory.toString());
```
HappyRay: This is not sticky bit, is it? This is setgid. Right?

see
https://docs.oracle.com/cd/E19683-01/806-4078/secfiles-69/index.html
"When setgid permission is applied to a directory, files that were created in this directory belong to the group to which the directory belongs, not the group to which the creating process belongs. Any user who has write and execute permissions in the directory can create a file there. However, the file belongs to the group that owns the directory, not to the user's group ownership.
"

reallocf: I think you're correct - thank you for the correction!

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -204,6 +205,23 @@ public FlowRunnerManager(final Props props,
             getClass().getClassLoader());
   }
 
+  /**
+   * Setting the gid bit on the execution directory forces all files/directories created within
+   * the directory to be a part of the azkaban group. Then, when users create their own files,
```
HappyRay: The service account name may not be azkaban. Consider using a more generic term.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -204,6 +205,23 @@ public FlowRunnerManager(final Props props,
...
+   * the azkaban cleanup thread can properly remove them.
+   *
+   * Java does not provide a standard library api for setting the gid bit because the gid bit
+   * is system dependent, so the only way to set this bit is to start a new process and run
+   * the shell command "chmod g+s " + execution directory name.
+   *
+   * Note that this will only work on unix systems.
```
HappyRay: Linux?

HappyRay:
In the pull request description, you still refer to "Setting the sticky bit ". 

LGTM otherwise.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
reallocf: It will also work on MacOS. It will definitely not work on Windows.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
reallocf: Good idea :+1:

HappyRay:
Reminder: please describe why it is hard to write an automated test for this change in your commit message.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -207,8 +207,8 @@ public FlowRunnerManager(final Props props,
 
   /**
    * Setting the gid bit on the execution directory forces all files/directories created within
-   * the directory to be a part of the azkaban group. Then, when users create their own files,
-   * the azkaban cleanup thread can properly remove them.
+   * the directory to be a part of the group associated with the azkaban process. Then, when users
+   * create their own files, the azkaban cleanup thread can properly remove them.
    *
    * Java does not provide a standard library api for setting the gid bit because the gid bit
    * is system dependent, so the only way to set this bit is to start a new process and run
```
HappyRay: Add that this logic doesn't work on windows and it works on Linux and MacOS?

PR #1295, opened 2017-07-28
Delete executions directory on executor shutdown to reclaim disk space

reallocf:
The execution directory can bloat to be relatively large, which could lead to problems if the executor is shutdown and stale executions remain sitting on the executor occupying disk space.

This change deletes the execution directory on executor shutdown. The directory itself is created on executor initialization, so the directory isn't necessary to reset the executor later.

Note that the execution directory will not be deleted until AFTER all flows are done if the executor is brought down via shutdown() - so it will not delete partially-written logs before they are uploaded to the database.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -196,6 +196,7 @@ public class AzkabanExecutorServer {
         logger.warn("Shutting down executor...");
         try {
           app.shutdownNow();
+          app.getFlowRunnerManager().deleteExecutionDirectory();
         } catch (final Exception e) {
           logger.error("Error while shutting down http server.", e);
         }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -739,7 +739,6 @@ public class FlowRunnerManager implements EventListener,
         logger.error(e);
       }
     }
-    deleteExecutionDirectory();
     logger.warn("Shutdown FlowRunnerManager complete.");
   }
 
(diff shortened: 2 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -732,6 +733,19 @@ public void shutdownNow() {
     logger.warn("Shutting down FlowRunnerManager now...");
     this.executorService.shutdownNow();
     this.triggerManager.shutdown();
+    deleteExecutionDirectory();
```
chengren311: why not deleting inside shutdown hook?

HappyRay: Why is it better?

HappyRay: Doesn't shutdownnow call shutdown which will call the delete function?

reallocf: @chengren311 so you're saying I should call this in the AzkabanExecutorServer's shutdownInternal/shutdownNow methods instead of in the FlowRunnerManager?

I figure it's better to do it here because the sequencing (all running jobs end, then delete execution directory) is more clear. What do you think?

reallocf: I don't see the shutdownNow method calling the shutdown method anywhere. But maybe it'd be better if we didn't delete the execution directory on shutdownNow because logs that were only partially created would be lost. What do you think?

HappyRay: My mistake, I mis-interpreted the call to the shotdown method for the triggerManager as the call for the FlowRunnerManager.

chengren311: @HappyRay it's simpler, you can compare the old commit with the latest one with suggested change.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -196,6 +196,7 @@ public void run() {
         logger.warn("Shutting down executor...");
         try {
           app.shutdownNow();
+          app.getFlowRunnerManager().deleteExecutionDirectory();
```
HappyRay: Is the shutdown hook invoked in the orderly shutdown process? How did you test it?

The deletion could be made internal to the flowRunnerManager in the previous commits. Why does it need to be made public?

HappyRay:
@reallocf did you find out why the triggerManager.shutdown is not called in azkaban.execapp.FlowRunnerManager#shutdown?

reallocf:
@HappyRay it turns out the shutdownInternal method in AzkabanExecutorServer which invokes the orderly shutdown process for FlowRunnerManager ends with System.exit(0) which calls the shutdown hook - so the orderly process calls the disorderly process. So the deletion will happen under all shutdown conditions.

PR #1296, opened 2017-07-28
increase default session duration to 7 days

chengren311:
As we introduced dynamic authentication mechanism(password is generated dynamically), longer cache TTL will reduce users' frustration of typing password again and again. Changing the default value is simpler than changing each installation's configuration file.


HappyRay:
Could you explain why you chose to change the default instead of changing the configuration for each installation?

See http://azkaban.github.io/azkaban/docs/latest/
session.time.to.live

chengren311:
@HappyRay given the large session cache size(10,000), 1 week is a more reasonable default number.

shuzhang1989:
@chengren311 what is the current cache utilization? we may need to check the current number to decide the default value.

chengren311:
@shuzhang1989 currently it only uses around 1000 buckets out of 10000(default max).

PR #1297, opened 2017-07-30
add required configuration for distribution run mode

es1220:
add required configuration for distribution run mode


HappyRay:
Could you add more details to your change description? I am not sure I follow.

es1220:
I ran to fail When I built `installDist` and ran unziped file in `build/distributions/`

For running distribute mode, 
I had added missing configuration such as `conf/azkaban.properties`, `conf/log4j.properties`, `conf/azkaban-users.xml` and added build step, copy conf dir, on `distributions`

This configuration is similar solo mode configuration and added comment for distribution mode setting (e.g. db, multi executor)

suvodeep-pyne:
@es1220 
I think they have different behaviors for different purposes. The solo server is intended to be a testbed for new users to play around with Azkaban. Thus it packages configuration along with it. And we don't recommend using it for production.

While the web and exec server packages are intended for a distributed production setup. Here, the expectation is that the binaries and configuration are to be kept separate. The default configuration on the web server does not help since the user will have to modify/replace the config every time on deployment.

Let me know what you think.

HappyRay:
Also adding a default log4j config in the default classpath is problematic since it may interfere with user's own log4j configuration.

es1220:
@suvodeep-pyne @HappyRay 
Thank you for your explanation.

The first time I installed Distributed Mode, it was hard to know what the server was missing because there were no files. The special log4j file is not logged.
It would be easier to complete the configuration if you add the necessary files to the guide document.

PR #1299, opened 2017-07-31
delete rentention logs every hour

kunkun-tang:
We observed the long deletion transaction in our production Azkaban cluster. In mysql, if delete transaction is large (The data going to be deleted is too big), it occupies too much undo space (buffer), which affect other transactions and other databases. So we change every day one time's deletion to every hour.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1639,12 +1639,12 @@ public class ExecutorManager extends EventHandler implements
   }
 
   /*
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1645,8 +1645,7 @@ public int getExecutableFlows(int projectId, String flowId, int from,
     // log file retention is 1 month.
 
     // check every day
```
HappyRay: Please update the comment to match.

kunkun-tang: Good catch! Fixing.

PR #1300, opened 2017-07-31
Revert "New status: KILLING (#1172)"

reallocf:
This reverts commit d1b836a.

Conflicts:
	azkaban-common/src/main/java/azkaban/executor/Status.java
	azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java

From @jamiesjc: When we kill the flow immediately after it starts in our integration test, it couldn't be killed due to some race condition. Users will see on the execution page that the job is in KILLING status but it actually never gets killed. And users cannot click kill button again during the KILLING period.
At the time when we kill, the process might not have started yet or the jobRunner has not yet been added to the activeJobRunners.
You can check more details in the PR description: #1289. We are still working on the fix.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1483,7 +1483,7 @@ public class ExecutorManager extends EventHandler implements
         continue;
         // case UNKNOWN:
       case READY:
-        node.setStatus(Status.KILLING);
+        node.setStatus(Status.KILLED);
         break;
       default:
         node.setStatus(Status.FAILED);
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -25,7 +25,6 @@ public enum Status {
   RUNNING(30),
   PAUSED(40),
   SUCCEEDED(50),
-  KILLING(55),
   KILLED(60),
   FAILED(70),
   FAILED_FINISHING(80),
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -34,7 +34,7 @@ public enum Status {
   QUEUED(110),
   FAILED_SUCCEEDED(120),
   CANCELLED(125);
-  // status is TINYINT and in H2 DB the possible values are: -128 to 127
+  // status is TINYINT in DB and the value ranges from -128 to 127
 
   private static final ImmutableMap<Integer, Status> numValMap = Arrays.stream(Status.values())
       .collect(ImmutableMap.toImmutableMap(status -> status.getNumVal(), status -> status));
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -34,7 +34,7 @@ public enum Status {
   QUEUED(110),
   FAILED_SUCCEEDED(120),
   CANCELLED(125);
-  // status is TINYINT in DB and the value ranges from -128 to 127
+  // status is TINYINT and in H2 DB the possible values are: -128 to 127
 
   private static final ImmutableMap<Integer, Status> numValMap = Arrays.stream(Status.values())
       .collect(ImmutableMap.toImmutableMap(status -> status.getNumVal(), status -> status));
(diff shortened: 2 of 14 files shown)


HappyRay:
@reallocf could you add the motivations for reverting this change in the change description? This way, @juhoautio or someone else can address the concerns and move forward later.

reallocf:
@HappyRay yes! I'll add some more details. I'm currently working with @jamiesjc and @chengren311 to determine if we actually need to revert this or not.

@juhoautio we think your change might have revealed an already existing bug. We're hoping to cut a new release (3.32.0) and are hoping to do so from master. If we were to revert your change, it would only be temporary to fix the related Azkaban bug before merging #1172 back in.

Will give more details shortly!

juhoautio:
What was it, maybe flow stuck in killing status? Did you get a thread dump on the executor node to see if some of the threads on that execution were still running? If you have a bug causing hung threads, at least you can restart the executor node to clean the flow status to FAILED. Of course if my change is a at fault, I'm sorry for that, and willing to fix the problem.

jamiesjc:
@juhoautio When we kill the flow immediately after it starts in our integration test, it couldn't be killed due to some race condition. Users will see on the execution page that the job is in KILLING status but it actually never gets killed. And users cannot click kill button again during the KILLING period. 
At the time when we kill, the process might not have started yet or the jobRunner has not yet been added to the activeJobRunners.  
You can check more details in the PR description: https://github.com/azkaban/azkaban/pull/1289. We are still working on the fix.

jamiesjc:
@juhoautio We really want to bring back your KILLING status change asap. If you are interested in working on the fix, we can discuss more details and share with you the integration test we have.

juhoautio:
I can fix it – maybe next week. Please share how to reproduce.

jamiesjc:
Here is an example of the test case to reproduce the problem. You need to replace the session.id, project and flow info to make it work.

testKillFlow.py
`def execute_flow():
    qparams = {'session.id': '92cbc945-6457-4ac1-a96e-7f575cd49359', 'ajax': 'executeFlow', 'project': 'az_acceptance_test',
    'flow': 'IntegrationTest'}
    headers = {'Accept': 'application/json'}
    azk_target_exec_url = 'http://localhost:8081/executor'
    jsonObj = requests.get(azk_target_exec_url, params=qparams, headers=headers, verify=True).json()
    if 'error' in jsonObj or 'execid' not in jsonObj:
        raise AzkabanError('Executing flows with concurrent option failed.')
    print jsonObj
    print "Execution id is %s." % jsonObj['execid']
    return jsonObj['execid']

def kill_flow(exec_id):
    qparams = {'session.id': '92cbc945-6457-4ac1-a96e-7f575cd49359', 'ajax': 'cancelFlow', 'execid': exec_id}
    headers = {'Accept': 'application/json'}
    azk_exec_details_url = 'http://localhost:8081/executor'
    jsonObj = requests.get(azk_exec_details_url, params=qparams, headers=headers, verify=True).json()
    print jsonObj
    if 'error' in jsonObj:
        raise AzkabanError('Failed to kill the flow.')

exec_id = execute_flow()
kill_flow(exec_id)`

jamiesjc:
Hi @juhoautio, @HappyRay has already proposed a fix for this race condition between jetty killing thread and jobRunner thread: https://github.com/azkaban/azkaban/pull/1310, also we've identified another potential race condition with killing thread and flowRunner thread https://github.com/azkaban/azkaban/issues/1311. Once the fix is done, we can work together to bring back you KILLING status change.

juhoautio:
After looking at the open fix candidates, it seems to me that showing the status as KILLING was at least better than showing KILLED. I mean, if the jobs were actually running in the background. As `Job#cancel` is apparently missed in the problematic scenario, as you wrote _"it actually never gets killed"_ – **but** I would assume that when the job eventually finishes or fails, the status will be moved from KILLING to some "finished" status. Am I right?

jamiesjc:
Yes, I agree. The status will be moved from KILLING to KILLED eventually. Currently, we don't have the status to indicate failure of killing. If we can make sure killing could always succeed, then we don't need to introduce the new status.

juhoautio:
Thanks for confirming.

Depending on the job type, killing could fail for various reasons any way. So I would say that killing is done on best effort basis. The underlying jobs might still manage to run completely even though they started around the same time when kill was triggered.

Because of this I would personally find it useful being able to click kill button again while the flow is killing, to try triggering Job#cancel again, in case there was a random error preventing the job to be cancelled.

Maybe a better idea is that Azkaban could automatically keep trying to call Job#cancel until it succeeds.

PR #1302, opened 2017-08-02
Update distributionUrl

Dshiv:
Was running gradlew build instalDist and kept getting a url error. This should fix it I believe.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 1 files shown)


HappyRay:
https://services.gradle.org/distributions/ seems to be accessible and ./gradlew works for me. 
Could you point to the documentation that suggests this alternative URL should be used?

Dshiv:
Sorry, It looks like there is a different issue with my setup a bit. I'm running the gradlew build installDist via docker with openjdk-8-jdk installed. At some point I was recieving errors saying that it could not download https://services.gradle.org/distributions/gradle-3.5.1-all.zip. I ran curl and saw the 302 HTTP response that that url gives. Closing.

Dshiv:
I apologize for my lack of due diligence before opening a pull request.

HappyRay:
No problem. Thanks for trying to fix an issue you observed.

PR #1305, opened 2017-08-03
Add default config to Emailer

jamiesjc:
Added default config in Emailer.
Removed unused configs in AbstractMailer.
Added test case on executor side for guicifying Emailer.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
@@ -22,9 +22,6 @@ public class AbstractMailer {
 
   public static final int DEFAULT_SMTP_PORT = 25;
   private static final int MB_IN_BYTES = 1048576;
-  private final String clientHostname;
-  private final int clientPort;
-  private final boolean usesSSL;
   private final boolean usesAuth;
 
   private final String mailHost;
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/utils/AbstractMailer.java
```diff
@@ -53,24 +48,6 @@ public AbstractMailer(final Props props) {
...
-          "https://" + this.clientHostname
-              + (this.clientPort == 443 ? "/" : ":" + this.clientPort + "/");
-    } else {
-      this.referenceURL =
-          "http://" + this.clientHostname
-              + (this.clientPort == 80 ? "/" : ":" + this.clientPort + "/");
-    }
-  }
-
-  public String getReferenceURL() {
```
kunkun-tang: 1). If me, I would not clean up this method, though Intellij can not identify its caller.
2). There are quite a few methods below like `getMailHost`, which are also never called up. I would clean them together in future.

jamiesjc: Thanks for the comments. I want to remove this method also because it's using clientHostname, clientPortName and usesSSL, which are props that don't have default values. Alternative way would be to provide default value for all those three props, but since this getReferenceURL is not used anywhere in the code, I think it's safe to remove. I've tested it on localhost and it works properly.

PR #1307, opened 2017-08-04
add ajax api to reset schedule's submitter(DO NOT MERGE)

chengren311:
1. add ajax api to reset schedule's submitter, e.x: http://localhost:8081/schedule?ajax=setScheduleSubmitter&scheduleId=2&submitter=someone
2. some refactoring by intellij save plugin
3. TriggerManager needs to persist updated trigger when updating trigger

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
@@ -47,11 +47,11 @@ public class Schedule {
   private final DateTimeZone timezone;
   private final long lastModifyTime;
   private final ReadablePeriod period;
-  private final String submitUser;
   private final String status;
   private final long submitTime;
   private final String cronExpression;
   private final boolean skipPastOccurrences = true;
+  private String submitUser;
   private int scheduleId;
   private long nextExecTime;
   private ExecutionOptions executionOptions;
@@ -242,6 +242,10 @@ public class Schedule {
 
   public String getSubmitUser() {
     return this.submitUser;
+  }
+
+  public void setSubmitUser(String submitUser) {
+    this.submitUser = submitUser;
   }
 
   public String getStatus() {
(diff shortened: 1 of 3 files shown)


HappyRay:
Why is this API needed? It seems odd to have the API to change the submitter without changing the schedule.

kunkun-tang:
@chengren311 It would be rational to check in this code to another branch, rather than master.

chengren311:
@HappyRay we can talk offline. In a nutshell, trigger data is also kept in memory, it's very tricky to backfill to db only. Providing API is safest and most convenient way I can think of.

chengren311:
@kunkun-tang agree, i will add DO NOT MERGE in the title of this PR.

azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
```diff
@@ -244,6 +244,10 @@ public String getSubmitUser() {
     return this.submitUser;
   }
 
+  public void setSubmitUser(String submitUser) {
+    this.submitUser = submitUser;
```
shuzhang1989: builder pattern?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -329,7 +336,8 @@ private void checkAllTriggers() throws TriggerManagerException {
              * as triggerCondition do. As a consequence, we need to figure out a way to distinguish
              * the previous ExpireCondition and this commit's ExpireCondition.
              */
-            if (t.getExpireCondition().getExpression().contains("EndTimeChecker") && t.expireConditionMet()) {
+            if (t.getExpireCondition().getExpression().contains("EndTimeChecker") && t
+                .expireConditionMet()) {
```
shuzhang1989: format

PR #1308, opened 2017-08-04
Change API's concurrent default option to skip

kunkun-tang:
Today, given a schedule, the default option for concurrent setting is `ignore`, that is, allowing concurrent executions whatever schedule triggers. This could be regarded as a bug in current AZ. The reason is that, In UI, the default schedule option is "skip" while API's default option is "ignore". The inconsistency between API and UI should be resolved.

This commit fixes the option to `skip`. When next schedule triggers, as long as the last execution hasn't finished, it won't trigger the next execution. As a consequence, this commit changes API's default option to "skip" and UI's default behavior is not affected. Tested on our staging cluster, and working.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
@@ -87,16 +87,17 @@ public class HttpRequestUtils {
     String concurrentOption = "skip";
     if (hasParam(req, "concurrentOption")) {
       concurrentOption = getParam(req, "concurrentOption");
-      execOptions.setConcurrentOption(concurrentOption);
-      if (concurrentOption.equals("pipeline")) {
-        final int pipelineLevel = getIntParam(req, "pipelineLevel");
-        execOptions.setPipelineLevel(pipelineLevel);
-      } else if (concurrentOption.equals("queue")) {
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -87,16 +87,17 @@ public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
     String concurrentOption = "skip";
```
reallocf: Can you just say:
`String concurrentOption = getParam(req, "concurrentOption", "skip");`
Or something like that so we don't need the if statement at all?

kunkun-tang: You are right!

HappyRay:
Please explain in the commit message the motivation for this change.

My understanding is that this is to make the API behavior consistent with the existing UI.

kunkun-tang:
PR was updated. @HappyRay @reallocf

HappyRay:
Thanks
Could you make it clear in the commit summary line that this change applies to the API?

kunkun-tang:
Thanks for your review. updated the description. @HappyRay

PR #1312, opened 2017-08-07
remove session IP constraint and shorten session TTL to 10 hours

chengren311:
Azkaban checks if the ip associated with the session remains the same, if not, it will ask user to log in again. IP constraint offers some security benefits by preventing other people interacting with azkaban by stealing your session, but it frustrates users by forcing user to login when IP is changed. 
A better alternative is to remove the IP constraint, but shorten session timeout from 24 hours to 10 hours. 10 hours is a short period not to compromise security too much. At the same time, most users only need to login once a day with it.

test methodology:
issue ajax api with same session id from two different hosts, both of them return successful results.


azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -239,21 +238,15 @@ private Session getSessionFromRequest(final HttpServletRequest req)
...
   }
 
-  private Session getSessionFromSessionId(final String sessionId, final String remoteIp) {
+  private Session getSessionFromSessionId(final String sessionId) {
     if (sessionId == null) {
       return null;
     }
 
-    final Session session = getApplication().getSessionCache().getSession(sessionId);
-    // Check if the IP's are equal. If not, we invalidate the sesson.
```
kunkun-tang: typo: `session`.

kunkun-tang:
Just a bit worried if this could be a serious security hole. As long as this session ID is leaked, anyone can use it to do many things.

HappyRay:
I am not aware of many web applications that enforces this IP check. If anyone knows good examples, I am interested in reading more about it.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
chengren311: it's already deleted

PR #1092, opened 2017-05-16
New SLA action type for killing a job

chengren311:
New SLA action type for killing a job
If the job misses the SLA, executor will retry the job based on the retry configuration of that job.

PR #1313, opened 2017-08-08
update doc for SLA change

chengren311:
update related doc for https://github.com/azkaban/azkaban/pull/1092

### DIFF
--- a/_includes/docs/latest/schedule-flow.html
+++ b/_includes/docs/latest/schedule-flow.html
@@ -26,7 +26,8 @@
 
 <h3>SLA</h3>
 
-<p>To add SLA notification or pre-emption, click on the SLA button. From here you can set the SLA alert emails. Rules can be added and applied to individual jobs or the flow itself. If duration threshold is exceeded, than an alert email can be set or the flow can be auto killed.</p>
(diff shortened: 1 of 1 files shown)


HappyRay:
Is it worth adding the version information when this change was introduced? Otherwise LGTM

chengren311:
@HappyRay is this necessary? it seems version is not mentioned anywhere in this doc.  This doc is always referring to latest version.

HappyRay:
It can help users who are not using the latest version. 
What's the downside?

HappyRay:
It's less of a concern for users who deploys new versions regularly.

chengren311:
downside is we change the convention/implication of this doc. If we do this, this will give users impression that If version is not specified, every az version works as documented. But it's not actually the case.

HappyRay:
Some projects publish versioned docs. Maybe something to consider in the future. 
Just a suggestion. 

We can evaluate this enhancement later.

chengren311:
@HappyRay ok, do you have other comments regarding this PR?

shuzhang1989:
lgtm, but agreed with @HappyRay that we need some sort of format guide of writing docs.

HappyRay:
LGTM

PR #1314, opened 2017-08-08
Refactoring Web Server and related code.

suvodeep-pyne:
The motivation is to make it easy to add more routes to the web server. A lot of static code exists which makes it difficult to access dependent classes directly.

Refactor details:
 - `prepareAndStartServer` and related methods were converted to instance method
 - deprecating the `app` member which is a static reference to `AzkabanWebServer`. Not removing it because it may be used by downstream code.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
@@ -122,6 +122,8 @@ public class AzkabanWebServer extends AzkabanServer {
   private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
   private static final String DEFAULT_STATIC_DIR = "";
+
+  @Deprecated
   private static AzkabanWebServer app;
   private final VelocityEngine velocityEngine;
 
@@ -136,8 +138,6 @@ public class AzkabanWebServer extends AzkabanServer {
   private final SessionCache sessionCache;
   private final List<ObjectName> registeredMBeans = new ArrayList<>();
 
-  //queuedThreadPool is mainly used to monitor jetty threadpool.
-  private QueuedThreadPool queuedThreadPool;
   private Map<String, TriggerPlugin> triggerPlugins;
   private MBeanServer mbeanServer;
 
@@ -184,13 +184,11 @@ public class AzkabanWebServer extends AzkabanServer {
     configureMBeanServer();
   }
 
+  @Deprecated
   public static AzkabanWebServer getInstance() {
     return app;
   }
 
-  /**
-   * Azkaban using Jetty
-   */
   public static void main(final String[] args) throws Exception {
     // Redirect all std out and err messages into log4j
     StdOutErrRedirect.redirectOutAndErrToLog();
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -97,22 +96,19 @@
 /**
  * The Azkaban Jetty server class
  *
- * Global azkaban properties for setup. All of them are optional unless
```
suvodeep-pyne: No content changed here.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -296,209 +273,6 @@ private static void validateDatabaseVersion(final Props azkabanSettings)
     }
   }
 
-  private static void configureRoutes(final Server server, final Props azkabanSettings)
```
suvodeep-pyne: This method moved below because of the save actions plugin. no change in logic. just parameter names were updated.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -123,6 +119,8 @@
   private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
   private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
   private static final String DEFAULT_STATIC_DIR = "";
+
+  @Deprecated
```
chengren311: why not removing it?

suvodeep-pyne: It is currently being accessed in other classes and probably downstream plugins like `TriggerPlugin` for example. It may break something.

azkaban-web-server/src/main/java/azkaban/webapp/TriggerPluginLoader.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.log4j.Logger;
+import org.apache.velocity.app.VelocityEngine;
+import org.mortbay.jetty.servlet.Context;
+
+public class TriggerPluginLoader {
+
+  private static final Logger log = Logger.getLogger(TriggerPluginLoader.class);
+  private final String pluginPath;
+
+  public TriggerPluginLoader(final Props props) {
```
HappyRay: Where is it used?

suvodeep-pyne: in the web server code. The `loadTriggerPlugins` method was extracted into a separate class. This requires further refactoring but phasing it for now.

HappyRay:
"The motivation is to make it easy to add more routes to the web server." Other than making this code cleaner, how does this change directly make it easier?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -518,45 +424,115 @@ private static void loadViewerPlugins(final Context root, final String pluginPat
     ve.addProperty("jar.resource.loader.path", jarResourcePath);
   }
 
-  /**
-   * Loads the Azkaban property file from the AZKABAN_HOME conf directory
-   */
-  private static Props loadConfigurationFromAzkabanHome() {
```
suvodeep-pyne: This method is not used and has been removed. The code below is copy paste from above except that the static modifier has been removed and references have been updated accordingly.

suvodeep-pyne:
@HappyRay The answer is now, the code is no longer static. Thus, if I want to add a `Status` Servlet for example and depend on a Guice injected service, I can do so. In a static context, I would have to access it in a round about way.

PR #1315, opened 2017-08-09
Refactoring the `ServerProvider` logic into separate methods

suvodeep-pyne:
No logic change has been introduced

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
@@ -65,7 +65,8 @@ public class WebServerProvider implements Provider<Server> {
     // setting stats configuration for connectors
     setStatsOnConnectors(server);
 
-    logger.info(String.format("Starting %sserver on port: %d", useSsl ? "SSL " : "", port));
+    logger.info(String.format(
+        "Starting %sserver on port: %d # Max threads: %d", useSsl ? "SSL " : "", port, maxThreads));
     return server;
   }
 
--- a/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
@@ -53,9 +53,6 @@ public class WebServerProvider implements Provider<Server> {
       final int sslPortNumber = this.props
           .getInt("jetty.ssl.port", Constants.DEFAULT_SSL_PORT_NUMBER);
       port = sslPortNumber;
-      logger.info("Setting up Jetty Https Server with port:" + sslPortNumber
-          + " and numThreads:" + maxThreads);
-
       server.addConnector(getSslSocketConnector(sslPortNumber));
     } else {
       port = this.props.getInt("jetty.port", Constants.DEFAULT_PORT_NUMBER);
--- a/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
@@ -45,53 +45,62 @@ public class WebServerProvider implements Provider<Server> {
 
     final int maxThreads = this.props
         .getInt("jetty.maxThreads", Constants.DEFAULT_JETTY_MAX_THREAD_COUNT);
-    final boolean isStatsOn = this.props.getBoolean("jetty.connector.stats", true);
-    logger.info("Setting up connector with stats on: " + isStatsOn);
 
-    final boolean ssl;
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -45,53 +45,62 @@ public Server get() {
...
-      connector.setPort(port);
-      connector.setHeaderBufferSize(MAX_HEADER_BUFFER_SIZE);
-      server.addConnector(connector);
+      server.addConnector(getSocketConnector(port));
     }
 
     // setting stats configuration for connectors
+    setStatsOnConnectors(server);
+
+    logger.info(String.format("Starting %sserver on port: %d", useSsl ? "SSL " : "", port));
```
chengren311: is the log correct, for useSsl case, should the port be sslPortNumber?

suvodeep-pyne: The `port` is updated with the `sslPortNumber` in that case. See line 57

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -45,53 +45,62 @@ public Server get() {
...
-      server.addConnector(secureConnector);
+      server.addConnector(getSslSocketConnector(sslPortNumber));
     } else {
-      ssl = false;
       port = this.props.getInt("jetty.port", Constants.DEFAULT_PORT_NUMBER);
-      final SocketConnector connector = new SocketConnector();
-      connector.setPort(port);
-      connector.setHeaderBufferSize(MAX_HEADER_BUFFER_SIZE);
-      server.addConnector(connector);
+      server.addConnector(getSocketConnector(port));
```
chengren311: should we also add log here as in the IF block?

suvodeep-pyne: Actually, on second thoughts, I think the log statement in the if block is not required since we are already logging it at the end of the method.

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -45,53 +45,62 @@ public Server get() {
...
-    logger.info(String.format("Starting %sserver on port: %d", ssl ? "SSL " : "", port));
-    return server;
+  private SocketConnector getSocketConnector(final int port) {
+    final SocketConnector connector = new SocketConnector();
+    connector.setPort(port);
+    connector.setHeaderBufferSize(MAX_HEADER_BUFFER_SIZE);
+    return connector;
+  }
+
+  private SslSocketConnector getSslSocketConnector(final int sslPortNumber) {
```
chengren311: is the method parameter necessary? it's reading from this.props just like other properties being set in the method, right?

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -45,53 +45,62 @@ public Server get() {
...
+    final boolean isStatsOn = this.props.getBoolean("jetty.connector.stats", true);
+    logger.info("Setting up connector with stats on: " + isStatsOn);
     for (final Connector connector : server.getConnectors()) {
       connector.setStatsOn(isStatsOn);
     }
+  }
 
-    logger.info(String.format("Starting %sserver on port: %d", ssl ? "SSL " : "", port));
-    return server;
+  private SocketConnector getSocketConnector(final int port) {
```
chengren311: same question as below

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
HappyRay: It's good to remind ourselves possible improvements.
I would suggest keeping the refactoring PR focused especially for code that doesn't have sufficient unit test coverage.

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
suvodeep-pyne: Agreed. The problem is that this method ideally needs 2 outputs: a Connector and a port. I'm keeping this as is for now. I didn't want to create a Method object or maintain a port member to do this.

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
suvodeep-pyne: Agreed.

Issue #1293, opened 2017-07-27
Missing primary key in project_events and executor_events tables


iberezovskiy:
The one of the requirement for Mysql replication setup is primary key in each tables. The mentioned tables (project_events and executor_events) don't have them. Could someone help with defining the primary keys for them and adding them to sql scripts?

One of the suggestions:

CREATE TABLE executor_events (
  executor_id INT NOT NULL,
  event_type TINYINT NOT NULL,
  event_time DATETIME NOT NULL,
  username VARCHAR(64),
  message VARCHAR(512)
  **PRIMARY KEY (executor_id, event_type, event_time)**
);

CREATE TABLE project_events (
	project_id INT NOT NULL,
	event_type TINYINT NOT NULL,
	event_time BIGINT NOT NULL,
	username VARCHAR(64),
	message VARCHAR(512)
        **PRIMARY KEY (project_id, event_type, event_time)**
);

kunkun-tang:
This is a good proposal. We can't think of the reason why can't we do this.

PR #1318, opened 2017-08-09
Add primary keys for executor_events and project_events tables

iberezovskiy:
Related issue: https://github.com/azkaban/azkaban/issues/1293

### DIFF
--- a/azkaban-db/src/main/sql/create.executor_events.sql
+++ b/azkaban-db/src/main/sql/create.executor_events.sql
@@ -1,9 +1,10 @@
 CREATE TABLE executor_events (
-  executor_id INT NOT NULL,
-  event_type TINYINT NOT NULL,
-  event_time DATETIME NOT NULL,
-  username VARCHAR(64),
-  message VARCHAR(512)
+    executor_id INT NOT NULL,
+    event_type TINYINT NOT NULL,
+    event_time DATETIME NOT NULL,
+    username VARCHAR(64),
+    message VARCHAR(512),
+    PRIMARY KEY(executor_id, event_type, event_time)
 );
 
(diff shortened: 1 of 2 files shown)


kunkun-tang:
@iberezovskiy Have you already replicated successfully after this change? Have you run into any issues?

iberezovskiy:
@kunkun-tang, I was able to validate Mysql replication setup and everything works fine. I didn't face any issues in Azkaban functionality.

HappyRay:
@kunkun-tang should an upgrade*.sql file be created for such a change?
Can this PR move forward?

kunkun-tang:
@HappyRay @iberezovskiy Our DBA checked our existing tables. There's a blocker to add the primary key to `project_events `.

The exception thrown:

> mysql> alter table project_events add PRIMARY KEY(project_id, event_type, event_time);
ERROR 1062 (23000): Duplicate entry '305-7-1463079491784' for key 'PRIMARY'

Looks like there exists duplicate rows with the primary key.

HappyRay:
Thanks @kunkun-tang.

Could you investigate why the duplication happens? Is it easy to fix?

HappyRay:
Is it an indication that the primary key design is flawed?

kunkun-tang:
I'm seeing all duplicates happening on EventType.Schedule.

>2017/08/18 01:12:19.622 +0000 INFO [ScheduleServlet] [Azkaban] User ..... has scheduled
2017/08/18 01:12:19.622 +0000 INFO [ScheduleServlet] [Azkaban] User ..... has scheduled

Is it possible that the user calling schedule api twice at the same time causing this.

PR #1320, opened 2017-08-09
New Status API for obtaining runtime information about Azkaban instance.

suvodeep-pyne:
Current features:
- version
- used Memory
- Max memory: XMX value
- Database availability: SELECT 1 check
- Map of current executors

Example:
```
{
    "version": "3.33.0-11-g1ac90f03",
    "usedMemory": 65144664,
    "xmx": 3817865216,
    "isDatabaseUp": true,
    "executorStatusMap": {
        "104": {
            "id": 104,
            "host": "abc.def.company.com",
            "port": 15462,
            "isActive": true
        }
    }
}
```

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/Status.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/Status.java
@@ -29,13 +29,19 @@ import java.util.Map;
 public class Status {
 
   private final String version;
+  private final String installationPath;
   private final long usedMemory, xmx;
   private final boolean isDatabaseUp;
   private final Map<Integer, Executor> executorStatusMap;
 
-  Status(final String version, final long usedMemory, final long xmx, final boolean isDatabaseUp,
+  Status(final String version,
+      final String installationPath,
+      final long usedMemory,
+      final long xmx,
+      final boolean isDatabaseUp,
       final Map<Integer, Executor> executorStatusMap) {
     this.version = version;
+    this.installationPath = installationPath;
     this.usedMemory = usedMemory;
     this.xmx = xmx;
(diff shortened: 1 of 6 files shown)


chengren311:
could you help me understand better on use case of the api? or explain it in the change description. Thanks!

suvodeep-pyne:
I think that is already in the description. It's a status API. It also
includes an example output. What are you looking for?

chengren311:
i'm curious about when it will be used.

suvodeep-pyne:
Both for ad hoc and systematically retrieving health data for the web
server. We already have a similar API for the executor.

azkaban-web-server/src/main/java/azkaban/webapp/Status.java
```diff
@@ -0,0 +1,44 @@
...
+import azkaban.executor.Executor;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+/**
+ * This POJO is used by GSON library to create a status JSON object. Certain warnings do not apply
+ * here.
+ */
+@SuppressWarnings({"FieldCanBeLocal", "unused"})
+public class Status {
```
chengren311: maybe use WebServerStatus to be more specific?

suvodeep-pyne: The reason I kept it `Status` is because the code is already in the web server module in the `azkaban.webapp` package. And this class is intended to provide the status of both the web and executor. Future changes will pull in the executor status into the web server.

So, it is kinda aimed at being the status of the entire system.

chengren311: i see, thanks

PR #1321, opened 2017-08-10
make job retry log more clear

chengren311:
previously if a job retries, log looks like this:
Starting job 1 attempt 1 at 1502324648910

it makes people misinterpret this as the 1st attempt of running the job, this PR changes the message: 
Starting job 1 retry 1 at 1502324648910

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -602,9 +602,8 @@ public class JobRunner extends EventHandler implements Runnable {
     }
 
     final int attemptNo = this.node.getAttempt();
-    logInfo("Finishing job " + this.jobId + " attempt: " + attemptNo + " at "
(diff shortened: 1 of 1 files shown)


HappyRay:
Did you run saveAction or reformat? The line length exceeded 100.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -602,9 +602,8 @@ private void doRun() {
     }
 
     final int attemptNo = this.node.getAttempt();
-    logInfo("Finishing job " + this.jobId + " attempt: " + attemptNo + " at "
+    logInfo("Finishing job " + this.jobId + (this.node.getAttempt() > 0 ? (" retry: " + attemptNo) : "") + " at "
```
HappyRay: Why not use attemptNo instead of another call to getAttempt?

Even better, consider using a private method to calculate this string. This logic is in two places.

chengren311: good point, PR: https://github.com/azkaban/azkaban/pull/1326

PR #1322, opened 2017-08-10
Adding pid to /status API

suvodeep-pyne:
Adding Process ID to status API info. The pid is read from the pid file which is configurable via `azkaban.pid.filename`. By default this value is `currentpid`.

Example:
```
{
    "version": "unknown",
    "pid": "69041",
    ...
}
```

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/Status.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/Status.java
@@ -29,18 +29,21 @@ import java.util.Map;
 public class Status {
 
   private final String version;
+  private final String pid;
   private final String installationPath;
   private final long usedMemory, xmx;
   private final boolean isDatabaseUp;
   private final Map<Integer, Executor> executorStatusMap;
 
   Status(final String version,
+      final String pid,
       final String installationPath,
       final long usedMemory,
       final long xmx,
       final boolean isDatabaseUp,
       final Map<Integer, Executor> executorStatusMap) {
     this.version = version;
+    this.pid = pid;
     this.installationPath = installationPath;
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/Status.java
```diff
@@ -29,18 +29,21 @@
...
 
   private final String version;
+  private final String pid;
   private final String installationPath;
   private final long usedMemory, xmx;
   private final boolean isDatabaseUp;
   private final Map<Integer, Executor> executorStatusMap;
 
   Status(final String version,
+      final String pid,
```
kunkun-tang: Looks like this pid is for Web server? Is that necessary to monitor? I believe executors' pid even deserves being monitored.

suvodeep-pyne: @kunkun-tang 
Yes. This is the pid for the web server. It is not about 'monitoring' the pid. This API is about reporting the current status of the application and pid is an important piece of information here. I agree the executor piece is more important and we do have plans to add that as well.

azkaban-web-server/src/main/java/azkaban/webapp/StatusService.java
```diff
@@ -66,13 +72,26 @@ public Status getStatus() {
...
     return new Status(version,
+        getPid(),
         getInstallationPath(),
         usedMemory,
         runtime.maxMemory(),
         getDbStatus(),
         getActiveExecutors());
   }
 
+  private String getPid() {
```
chengren311: this seems more like an util method, should we make it static?

PR #1324, opened 2017-08-10
Standardize on "System.currentTimeMillis()" within the ExecutorManager class

li-afaris:
This is a small change to standardize on using System.currentTimeMillis() within the ExecutorManager class.


HappyRay:
The Jodatime has facilities to mock time to make unit testing easier.

We should move to the new Java8 Time

PR #1325, opened 2017-08-10
Fixing known security bug via proper *nix permissioning

reallocf:
Right now, Azkaban leverages the execute-as-user script so that users' jobs are run via a process of the user's creation instead of as the Azkaban process. This is really important because it restricts users from interfering with Azkaban or reading secure information that Azkaban holds.

While this script works well, it has a bug in which it doesn't remove the "azkaban" group from the users list of groups. This is not secure. It did have a side-effect though that should be kept: it allowed users to write to their /execution/<exec_id>/src directory. This is desired because it gives users the ability to cache temporary data on the executor between jobs in a flow.

This PR does two things:
- It fixes the security bug in execute-as-user.c by calling the initgroups() function for the effective user.
- It uses the execute-as-user executable to set up the /execution/<exec_id>/src directory so that its permissions allow the user to work within it . It also makes sure that this is done in a way that can be cleaned up later by the Azkaban thread. It does this by executing the command: `chown <user>:azkaban`

This particular change will be hard to test with unit testing, but it will easy to test within the integration testing suite (I discussed this with @jamiesjc). It can be done completely with two test:
- The first one tries to write to the current working directory. If the flow concludes successfully (i.e. the file is successfully created), the test passes.
- The second tries to perform an "improper" action reserved for members of the "azkaban" group. If the flow fails (i.e. the action cannot be performed), the test passes.
Together I believe this will give sufficient test coverage for this part of the code and each test should be quick (one job flows that should terminate in < 1 second)

### DIFF
--- a/azkaban-common/src/main/c/execute-as-user.c
+++ b/azkaban-common/src/main/c/execute-as-user.c
@@ -39,18 +39,23 @@ int INVALID_INPUT = 30;
 
 /*
  *  Change the real and effective user and group from super user to the specified user
- *  
+ *
  *  Adopted from:
  *  ./hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/native/container-executor/impl/container-executor.c
- *  
+ *
  */
 
-int change_user(uid_t user, gid_t group) {
+int change_user(char *username, uid_t user, gid_t group) {
     if (user == getuid() && user == geteuid() &&
             group == getgid() && group == getegid()) {
         return 0;
     }
 
+    if (initgroups(username, group) != 0) {
+        fprintf(LOGFILE, "Error setting supplementary groups for user %s: %s\n",
+            user, strerror(errno));
+        return SETUID_OPER_FAILED;
+    }
     if (seteuid(0) != 0) {
         fprintf(LOGFILE, "unable to reacquire root - %s\n", strerror(errno));
         fprintf(LOGFILE, "Real: %d:%d; Effective: %d:%d\n",
@@ -100,7 +105,7 @@ int main(int argc, char **argv){
 
     // try to change user
     fprintf(LOGFILE, "Changing user: user: %s, uid: %d, gid: %d\n", uid, user_info->pw_uid, user_info->pw_gid);
-    int retval = change_user(user_info->pw_uid, user_info->pw_gid);
+    int retval = change_user(uid, user_info->pw_uid, user_info->pw_gid);
     if (retval != 0){
         fprintf(LOGFILE, "Error changing user to %s\n", uid);
         return SETUID_OPER_FAILED;
--- a/azkaban-common/src/main/c/execute-as-user.c
+++ b/azkaban-common/src/main/c/execute-as-user.c
@@ -53,7 +53,7 @@ int change_user(char *username, uid_t user, gid_t group) {
 
     if (initgroups(username, group) != 0) {
         fprintf(LOGFILE, "Error setting supplementary groups for user %s: %s\n",
-            user, strerror(errno));
+            username, strerror(errno));
         return SETUID_OPER_FAILED;
     }
     if (seteuid(0) != 0) {
--- a/azkaban-common/src/main/c/execute-as-user.c
+++ b/azkaban-common/src/main/c/execute-as-user.c
@@ -90,24 +90,24 @@ int main(int argc, char **argv){
     }
 
     if (argc < 3) {
-        fprintf(ERRORFILE, "Requires at least 3 variables: ./execute-as-user uid command [args]");
+        fprintf(ERRORFILE, "Requires at least 3 variables: ./execute-as-user username command [args]");
         return INVALID_INPUT;
     }
 
-    char *uid = argv[1];
+    char *username = argv[1];
 
     // gather information about user
-    struct passwd *user_info = getpwnam(uid);
+    struct passwd *user_info = getpwnam(username);
     if (user_info == NULL){
-        fprintf(LOGFILE, "user does not exist: %s", uid);
+        fprintf(LOGFILE, "user does not exist: %s", username);
         return USER_NOT_FOUND;
     }
 
     // try to change user
-    fprintf(LOGFILE, "Changing user: user: %s, uid: %d, gid: %d\n", uid, user_info->pw_uid, user_info->pw_gid);
-    int retval = change_user(uid, user_info->pw_uid, user_info->pw_gid);
+    fprintf(LOGFILE, "Changing user: user: %s, uid: %d, gid: %d\n", username, user_info->pw_uid, user_info->pw_gid);
+    int retval = change_user(username, user_info->pw_uid, user_info->pw_gid);
     if (retval != 0){
-        fprintf(LOGFILE, "Error changing user to %s\n", uid);
+        fprintf(LOGFILE, "Error changing user to %s\n", username);
         return SETUID_OPER_FAILED;
     }
 
--- a/azkaban-common/src/main/java/azkaban/Constants.java
+++ b/azkaban-common/src/main/java/azkaban/Constants.java
@@ -37,6 +37,9 @@ public class Constants {
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
 
+  // Path name of execute-as-user executable
+  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
+
   // Internal username used to perform SLA action
   public static final String AZKABAN_SLA_CHECKER_USERNAME = "azkaban_sla";
 
--- a/azkaban-common/src/main/java/azkaban/Constants.java
+++ b/azkaban-common/src/main/java/azkaban/Constants.java
@@ -37,9 +37,6 @@ public class Constants {
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
 
-  // Path name of execute-as-user executable
-  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
-
   // Internal username used to perform SLA action
   public static final String AZKABAN_SLA_CHECKER_USERNAME = "azkaban_sla";
 
@@ -89,6 +86,12 @@ public class Constants {
 
     // List of users we prevent azkaban from running flows as. (ie: root, azkaban)
     public static final String BLACK_LISTED_USERS = "azkaban.server.blacklist.users";
+
+    // Path name of execute-as-user executable
+    public static final String AZKABAN_SERVER_NATIVE_LIB_FOLDER = "azkaban.native.lib";
+
+    // Name of *nix group associated with the process running Azkaban
+    public static final String AZKABAN_SERVER_GROUP_NAME = "azkaban.group.name";
 
     // Legacy configs section, new configs should follow the naming convention of azkaban.server.<rest of the name> for server configs.
 
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -50,6 +50,7 @@ public class ProcessJob extends AbstractProcessJob {
 
   public static final String COMMAND = "command";
   public static final String AZKABAN_MEMORY_CHECK = "azkaban.memory.check";
+  // Use AZKABAN_SERVER_NATIVE_LIB_FOLDER Configuration Key instead
   @Deprecated
   public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
   public static final String EXECUTE_AS_USER = "execute.as.user";
@@ -230,7 +231,7 @@ public class ProcessJob extends AbstractProcessJob {
     // nativeLibFolder specifies the path for execute-as-user file,
     // which will change user from Azkaban to effectiveUser
     if (isExecuteAsUser) {
-      final String nativeLibFolder = this.sysProps.getString(NATIVE_LIB_FOLDER);
+      final String nativeLibFolder = this.sysProps.getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER);
       executeAsUserBinaryPath = String.format("%s/%s", nativeLibFolder, "execute-as-user");
       effectiveUser = getEffectiveUser(this.jobProps);
       // Throw exception if Azkaban tries to run flow as a prohibited user
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -16,7 +16,8 @@
 
 package azkaban.jobExecutor;
 
-import static azkaban.Constants.NATIVE_LIB_FOLDER;
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_GROUP_NAME;
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_NATIVE_LIB_FOLDER;
 import static azkaban.ServiceProvider.SERVICE_PROVIDER;
 
 import azkaban.Constants;
@@ -49,6 +50,8 @@ public class ProcessJob extends AbstractProcessJob {
 
   public static final String COMMAND = "command";
   public static final String AZKABAN_MEMORY_CHECK = "azkaban.memory.check";
+  @Deprecated
+  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
   public static final String EXECUTE_AS_USER = "execute.as.user";
   public static final String USER_TO_PROXY = "user.to.proxy";
   public static final String KRB5CCNAME = "KRB5CCNAME";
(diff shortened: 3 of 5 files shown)


azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -37,6 +37,9 @@
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
 
+  // Path name of execute-as-user executable
+  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
```
suvodeep-pyne: There is a configuration key with this name already. Let's have this in `Constants.ConfigurationKeys`

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -67,15 +70,6 @@
 public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
 
   /**
-   * TODO Remove duplicated constants from plugins.
-   *
-   * Azkaban plugins don't depend on a common submodule from which they both can inherit code. Thus,
-   * constants are copied around and any changes to the constant values will break Azkaban. This
-   * needs to be fixed as part of a plugin infrastructure implementation.
-   */
-  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
```
suvodeep-pyne: Removing this parameter may break downstream plugins. I suggest we avoid this refactor. Just mark this deprecated for now.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -46,12 +49,16 @@
 
   public static final String COMMAND = "command";
   public static final String AZKABAN_MEMORY_CHECK = "azkaban.memory.check";
-  public static final String NATIVE_LIB_FOLDER = "azkaban.native.lib";
```
suvodeep-pyne: Removing this parameter may break downstream plugins. I suggest we avoid this refactor. Just mark this deprecated for now.

reallocf: Good thinking. Do you think it's worth creating a NATIVE_LIB_FOLDER string in Constants then? Or should we just keep the strings that are local to ProcessJob and HadoopSecurityManager_H_2_0?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -46,12 +49,16 @@
...
   public static final String EXECUTE_AS_USER = "execute.as.user";
   public static final String USER_TO_PROXY = "user.to.proxy";
   public static final String KRB5CCNAME = "KRB5CCNAME";
   private static final Duration KILL_TIME = Duration.ofSeconds(30);
   private static final String MEMCHECK_ENABLED = "memCheck.enabled";
+  private static final String CHOWN = "chown";
+  private static final String CREATE_FILE = "touch";
+  private static final String GROUP_NAME = "azkaban";
```
suvodeep-pyne: This needs to be a configuration parameter defaulted to `azkaban`

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -230,6 +237,11 @@ public void run() throws Exception {
             String.format("Not permitted to proxy as '%s' through Azkaban", effectiveUser)
         );
       }
+      // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
+      // if the directory is not permissioned correctly already (should happen once per execution)
+      if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
```
suvodeep-pyne: This seems tricky. If the effective user is able to write to the directory, then we don't set up permissions? I think we do run into cases today where the effective user is able to write but we are unable to clean up. no?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -345,6 +357,46 @@ private String getEffectiveUser(final Props jobProps) {
...
+
+  /**
+   * Changes permission on current working directory so that the directory is owned by the user
+   * and the group remains azkaban.
+   *
+   * Leverages execute-as-user with "root" as the user to run the command.
+   *
+   * @param effectiveUser user/proxy user running the job
+   */
+  private void assignUserDirOwnership(final String effectiveUser) throws IOException {
```
suvodeep-pyne: Let's use the  `ExecuteAsUser` class for this. If it is sitting in a different package, let's pull it to `azkaban-common` since this is precisely what the wrapper class was created for.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
reallocf: With this change, the default has changed. Before, users could always write to their directory. Now - due to the change to execute-as-user.c - they won't be able to until we allow them to.

The clean up change is a different problem that should be taken care of in 3.32.1 because of #1292 :+1:

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
suvodeep-pyne: I would like to have all config keys in one place. But, I think refactoring any public static variable in these classes is a risk. Whether you want to keep things this way or refactor it and call deprecated is up to you. I'm fine with either.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
reallocf: This does use `ExecuteAsUser` - is there a way you think I can use it better? :smile:

And yes! I moved it to `azkaban-common`. Glad to know that was the correct approach - thanks!

suvodeep-pyne: Oh. Yes. Else it would create a cyclical build dependency! :)

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -385,10 +387,11 @@ private boolean canWriteInCurrentWorkingDirectory(final String effectiveUser)
    */
   private void assignUserDirOwnership(final String effectiveUser) throws IOException {
     final ExecuteAsUser executeAsUser = new ExecuteAsUser(
-        this.sysProps.getString(NATIVE_LIB_FOLDER));
+        this.sysProps.getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
+    final String groupName = this.sysProps.getString(AZKABAN_SERVER_GROUP_NAME, "azkaban");
     final List<String> changeOwnershipCommand = Arrays
-        .asList(CHOWN, effectiveUser + ":" + GROUP_NAME, getWorkingDirectory());
-    info("Change current working directory ownership to " + effectiveUser + ":" + GROUP_NAME + ".");
+        .asList(CHOWN, effectiveUser + ":" + groupName, getWorkingDirectory());
```
suvodeep-pyne: Possible direction: 
Have an OS Helper/Util class which leverages `ExecuteAsUser` class to execute OS level functions since we do plenty of that in Azkaban. We have an `OsMemoryUtil` for example.
```
// So OsHelper becomes our central place for doing these kinda stuff.
new OsHelper().chown(getWorkingDirectory(), user, group);

```
However, I don't think this needs to be done now.

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -39,18 +39,23 @@ int INVALID_INPUT = 30;
...
- *  
+ *
  *  Adopted from:
  *  ./hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-nodemanager/src/main/native/container-executor/impl/container-executor.c
- *  
+ *
  */
 
-int change_user(uid_t user, gid_t group) {
+int change_user(char *username, uid_t user, gid_t group) {
```
suvodeep-pyne: I think `uid_t` is a typedef of an `unsigned int`. `char*` might be incorrect. This should be evident if you do a `*username` somewhere - might result in garbage or seg fault. Correct me if am wrong.

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -39,18 +39,23 @@ int INVALID_INPUT = 30;
...
     if (user == getuid() && user == geteuid() &&
             group == getgid() && group == getegid()) {
         return 0;
     }
 
+    if (initgroups(username, group) != 0) {
+        fprintf(LOGFILE, "Error setting supplementary groups for user %s: %s\n",
```
suvodeep-pyne: Following up from the previous comment: I'm not sure how or whether this works.
```
printf("%s", user); // user is of type uid_t
```

azkaban-common/src/main/c/execute-as-user.c
suvodeep-pyne: And I would call it `effective_user` or something since that's we are comparing against in `geteuid()`

reallocf: This actually is supposed to be a char* being passed. The initgroups() function below takes in a char* and a gid_t where the char* is the username of the user. When "uid" is passed from the main function, it actually is a char* - it's kind of hard to tell in github.

azkaban-common/src/main/c/execute-as-user.c
reallocf: http://man7.org/linux/man-pages/man3/initgroups.3.html - takes in a char* so %s should work I think

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -100,7 +105,7 @@ int main(int argc, char **argv){
 
     // try to change user
     fprintf(LOGFILE, "Changing user: user: %s, uid: %d, gid: %d\n", uid, user_info->pw_uid, user_info->pw_gid);
-    int retval = change_user(user_info->pw_uid, user_info->pw_gid);
+    int retval = change_user(uid, user_info->pw_uid, user_info->pw_gid);
```
reallocf: uid is declared/initialized as a char* right above here on line 97 :+1:

azkaban-common/src/main/c/execute-as-user.c
suvodeep-pyne: Can we double check?

for `initgroups()` I saw the signature. But `get*uid()` returns `uid`s which are probably not `char*`. Did we test this?
https://docs.oracle.com/cd/E36784_01/html/E36872/geteuid-2.html

The `getpwnam()` function for example translates `uid` to `username`
http://man7.org/linux/man-pages/man3/getpwnam.3.html

azkaban-common/src/main/c/execute-as-user.c
suvodeep-pyne: I didn't quite get that. Let's chat tomorrow. This looks a bit weird.

reallocf: Oh! You're right, this should be `fprintf(LOGFILE, "Error setting supplementary groups for user %s: %s\n", username, strerror(errno));` instead of user. Nice catch!

azkaban-common/src/main/c/execute-as-user.c
```diff
@@ -100,7 +105,7 @@ int main(int argc, char **argv){
...
     if (retval != 0){
         fprintf(LOGFILE, "Error changing user to %s\n", uid);
```
suvodeep-pyne: Can we rename the local variable `uid` to `username` since it is not the `uid`.  Even the code
```
fprintf(ERRORFILE, "Requires at least 3 variables: ./execute-as-user uid command [args]");
```
is confusing since it mixes up `uid`, which is actually an `uint` vs `username` which is a string.

reallocf: Agreed - this confuses me to. Will do.

azkaban-common/src/main/c/execute-as-user.c
suvodeep-pyne: same comment here as below. Let's call it `username` instead of `uid`.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -66,15 +69,8 @@
...
 public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
 
-  /**
-   * TODO Remove duplicated constants from plugins.
-   *
-   * Azkaban plugins don't depend on a common submodule from which they both can inherit code. Thus,
-   * constants are copied around and any changes to the constant values will break Azkaban. This
-   * needs to be fixed as part of a plugin infrastructure implementation.
-   */
+  @Deprecated
```
suvodeep-pyne: Let's add a small comment on what needs to be used instead.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -46,12 +50,17 @@
 
   public static final String COMMAND = "command";
   public static final String AZKABAN_MEMORY_CHECK = "azkaban.memory.check";
+  @Deprecated
```
suvodeep-pyne: Let's add a comment on what needs to be used instead.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -46,12 +50,18 @@
 
   public static final String COMMAND = "command";
   public static final String AZKABAN_MEMORY_CHECK = "azkaban.memory.check";
+  // Use AZKABAN_SERVER_NATIVE_LIB_FOLDER Configuration Key instead
```
suvodeep-pyne: nit: Can mention the full reference:
`azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_NATIVE_LIB_FOLDER`

This helps IntelliJ track renames inside comments

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -345,6 +360,47 @@ private String getEffectiveUser(final Props jobProps) {
...
+  /**
+   * Changes permission on current working directory so that the directory is owned by the user
+   * and the group remains azkaban.
+   *
+   * Leverages execute-as-user with "root" as the user to run the command.
+   *
+   * @param effectiveUser user/proxy user running the job
+   */
+  private void assignUserDirOwnership(final String effectiveUser) throws IOException {
+    final ExecuteAsUser executeAsUser = new ExecuteAsUser(
```
HappyRay: It would be good in the future to get this object from Guice to make unit test easier.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -345,6 +360,47 @@ private String getEffectiveUser(final Props jobProps) {
...
+        this.sysProps.getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
+    final String groupName = this.sysProps.getString(AZKABAN_SERVER_GROUP_NAME, "azkaban");
+    final List<String> changeOwnershipCommand = Arrays
+        .asList(CHOWN, effectiveUser + ":" + groupName, getWorkingDirectory());
+    info("Change current working directory ownership to " + effectiveUser + ":" + groupName + ".");
+    final int result = executeAsUser.execute("root", changeOwnershipCommand);
+    if (result != 0) {
```
HappyRay: Should this throw an exception instead? Won't the job fail later which may be harder to debug?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -230,6 +240,11 @@ public void run() throws Exception {
             String.format("Not permitted to proxy as '%s' through Azkaban", effectiveUser)
         );
       }
+      // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
+      // if the directory is not permissioned correctly already (should happen once per execution)
+      if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
```
HappyRay: Wouldn't it be cleaner to create the directory with the proper permission in the first place? I understand you tried and had issues. It would be good to document what the issue is and fix it if possible.

reallocf: It would be good to document this - do you think it'd be best to do so here? Or in a comment?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
reallocf: Yes I agree, this should throw an error - I'll make another PR on Monday doing so.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
HappyRay: If you plan to address it soon, documenting it here is easier. Otherwise, I will leave it to you. Just consider how others can find and use this information to make documenting it worthwhile.

HappyRay:
Thanks for the detailed commit message!

I think a summary line can describe more details. e.g.
security: jobs should not inherit the Azkaban service group permissions

PR #1326, opened 2017-08-10
a little refactor on job retry log

chengren311:
1. apply save plugin
2. make duplicate log construction logic an util method


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -632,8 +636,8 @@ private Status prepareJob() throws RuntimeException {
         return null;
       }
 
-      logInfo("Starting job " + this.jobId + (this.node.getAttempt() > 0 ? (" retry: " + this.node.getAttempt()) : "") + " at "
-          + this.node.getStartTime());
+      logInfo("Starting job " + this.jobId + getNodeRetryLog() + " at " + this.node
+          .getStartTime());
```
shuzhang1989: + this.node.getStartTime();

chengren311: thanks

PR #1328, opened 2017-08-10
TriggerManager needs to persist updated trigger when updating trigger

chengren311:
Previously, TriggerManager won't persist updated schedule when schedule gets updated(e.x. add new SLA rule to a schedule). It gets persisted only when schedule gets triggered. So if we add a  SLA rule to a schedule or do a reschedule, and restart the web server after it, these updates will be lost.


HappyRay:
Nice found. Thanks for fixing this!

PR #1330, opened 2017-08-11
Fix race condition between Job#run & Job#kill

juhoautio:
# Change description candidate

There's a race condition which causes Job status to be first set to some other finished status and then subsequently to KILLED. Practical implication is that when a job is killed, it's possible that it sends a JOB_FINISHED event with some other status, although it is eventually marked as KILLED.

Obviously the final status of a job should nevertheless be "final" instead of "flapping" after a finished status has been set.

This race condition is fixed by synchronizing in both run & kill methods, so that the combination of these operations are wrapped into "a single atomic operation":

- checking the status and/or killed flag
- based on the previous check result(s), setting the status and/or killed flag

Kill method already had the synchronization in place, but run method was missing it.

# Further details, to be left available here in the pull request

I'm also demonstrating the problem with a couple "dirty" commits, as I couldn't think of an easy way to create a unit test that simulates this race condition.

## 1st commit: Prove job status synchronization bug when killing

When job is killed, it's possible that it sends a JOB_FINISHED event with some other status but is eventually marked as KILLED.

For example this code modification demonstrates how it sends JOB_FINISHED with Status.SUCCEEDED but then marks the job as KILLED.

It should be ensured that if job has finished it won't be marked as killed and on the other hand it isn't marked killed if it has finished first.

Extract from `FlowRunnerTest#execAndCancel` output:

```
2017/08/11 11:25:22.889 +0300 INFO [main] [derived-member-data] Killing 1 jobs.
kill() called..
Sleeping before marking as SUCCEEDED
!Status.isStatusFinished(finalStatus) = true
changeStatus(Status.SUCCEEDED)!!
isKilled() = false
finalStatus = SUCCEEDED
setting killed=true

2017/08/11 11:25:26.946 +0300 INFO [JobRunner-job1-1] [derived-member-data] Job job1 finished with status SUCCEEDED in 1 seconds
2017/08/11 11:25:26.948 +0300 INFO [FlowRunner-exec-1] [derived-member-data] Flow '' is set to KILLED in 4 seconds
2017/08/11 11:25:26.949 +0300 INFO [FlowRunner-exec-1] [derived-member-data] Finished Flow
2017/08/11 11:25:26.949 +0300 INFO [FlowRunner-exec-1] [derived-member-data] Setting end time for flow 1 to 1502439926949
setting Status.KILLED

testJob: job1 KILLED
```

## 2nd commit: Fix job status synchronization bug when killed

Now the test output has this, as expected:

```
kill() called..
2017/08/11 12:42:47.845 +0300 INFO [main] [derived-member-data] Flow killed by me
Sleeping before marking as SUCCEEDED
!Status.isStatusFinished(finalStatus) = true
setting killed=true
setting Status.KILLED
2017/08/11 12:42:51.912 +0300 INFO [JobRunner-job1-1] [derived-member-data] Job job1 finished with status KILLED in 4 seconds
isKilled() = true
Setting to Status.KILLED
finalStatus = KILLED
2017/08/11 12:42:51.919 +0300 INFO [FlowRunner-exec-1] [derived-member-data] Flow '' is set to KILLED in 4 seconds
2017/08/11 12:42:51.919 +0300 INFO [FlowRunner-exec-1] [derived-member-data] Finished Flow
testJob: job1 KILLED
```


juhoautio:
@HappyRay, could you or someone have a look at this?

> If you agree, I will make another PR with the fix alone.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -739,29 +741,33 @@ private Status runJob() {
...
     if (this.job != null) {
       this.node.setOutputProps(this.job.getJobGeneratedProperties());
     }
 
-    // If the job is still running, set the status to Success.
-    if (!Status.isStatusFinished(finalStatus)) {
-      finalStatus = changeStatus(Status.SUCCEEDED);
+    synchronized (this.syncObject) {
+      // If the job is still running, set the status to Success.
+      if (!Status.isStatusFinished(finalStatus) && !this.killed) {
```
HappyRay: When will the job status remain "unfinished" when it gets here? And if killed is true, who will update the final status?

juhoautio: > And if killed is true, who will update the final status?

Here (KILLED is a "final status", it doesn't have to be set again after this):
https://github.com/juhoautio/azkaban/blob/723a46ec93d6e52f178e5d2ab445d630099d0ad6/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L833

Although now that you mention it, I'm not sure what would happen if this row is hit:
https://github.com/juhoautio/azkaban/blob/723a46ec93d6e52f178e5d2ab445d630099d0ad6/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L822 – seems a bit weird, I think we should add `this.changeStatus(Status.KILLED)` before exiting with this special `return`.

HappyRay:
Here is my understanding of the problem you described. Please correct me if I am wrong.

Users issue kill command for a flow.
Due to race conditions, the job may finish successfully and registered event listeners are notified as such.
Finally, the kill thread marks the job as killed as the final status.

---
Thread synchronization is tricky to get right. I am afraid I don't quite follow your diagnostics and how your fix will solve this problem. 

---

I am thinking of refactoring the DAG processing engine to reduce the complexity and reduce the need for thread synchronization.  e.g. if there is one thread that is responsible for the job/flow status transition, it would be easier to reason about.
In the killing scenario, a successful return of the kill command only means the server has received and queued up the kill command successfully. The final outcome may still be either a job is killed or a job finishes successfully. The user or a user program would still need to poll the status to find out the final status. And there should not be a state transition from running -> killed -> success/failure.  It should either be running -> killed or running -> success/failure. 

cc @jamiesjc who has worked in this area.

HappyRay:
How big a problem do you think this is causing?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: > When will the job status remain "unfinished" when it gets here?

In the normal, successful case. As you can see, here the status is RUNNING:
https://github.com/juhoautio/azkaban/blob/723a46ec93d6e52f178e5d2ab445d630099d0ad6/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L740-L742 – the value of `finalStatus` is only changed before row 768 in case of an exception.

juhoautio:
> How big a problem do you think this is causing?

I think currently the worst result is that you get a job callback that says SUCCEEDED (also could be FAILED, I suppose), although if you go and look at the execution, eventually it may be marked as KILLED instead.

With the additional KILLING status patch applied it's much worse: hitting this bug causes the flow to be stuck because a job is left stuck in KILLING status.

juhoautio:
> I am afraid I don't quite follow your diagnostics and how your fix will solve this problem.

The core source of problem here is using the field `Job.killed` to decide how to set job status, but not synchronizing these operations. The code branch in `Job#kill` was already synchronized (it reads status, possibly writes killed flag and job status). However, `Job#runJob` was setting job status without synchronization (depending on the killed flag), so a race condition was possible.

juhoautio:
> e.g. if there is one thread that is responsible for the job/flow status transition, it would be easier to reason about.

Definitely. May be a lot of work though, so I would appreciate having this bug fix ASAP.

> In the killing scenario, a successful return of the kill command only means the server has received and queued up the kill command successfully. The final outcome may still be either a job is killed or a job finishes successfully. The user or a user program would still need to poll the status to find out the final status. And there should not be a state transition from running -> killed -> success/failure. It should either be running -> killed or running -> success/failure.

I suppose you describe the situation how it is without this fix. Yes, that's how I see it pretty much. Is this just you ensuring that you understand how it works, or do you mean to say that this wouldn't be a real problem?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: I see. 

If job.run throws an exception, the finalStatus variable will be set to Killed. So the !this.killed will not be evaluated.

If there is no exception when will the killed be true? And what's the expected state transition for this job?

HappyRay:
> I suppose you describe the situation how it is without this fix. Yes, that's how I see it pretty much. Is this just you ensuring that you understand how it works, or do you mean to say that this wouldn't be a real problem?

I described it as the desired state. 

---
The lack of synchronization does seem to be a problem.

Could you describe an example sequence to better help me understand the particular race condition? And how it will interact with the killing status change?

Are you seeing this issue in your production environment now?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: > If there is no exception when will the killed be true? And what's the expected state transition for this job?

Sorry but I'm not sure what you mean?

juhoautio:
> Could you describe an example sequence to better help me understand the particular race condition?

Actually I tried to show that in the 2 scenarios (buggy vs. synchronized) of my PR description. It might help reading it to know that `Job job1 finished with status..` is the line that tells what status was fired as a JOB_FINISHED event.

Please let me know if you were looking for some other form of expressing it.

> And how it will interact with the killing status change?

When the KILLING status change is applied, this line will set status to KILLING instead of KILLED:
https://github.com/juhoautio/azkaban/blob/723a46ec93d6e52f178e5d2ab445d630099d0ad6/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L833

Then it would be possible that Job.run returns initially as SUCCEEDED, while Job.kill is being executed, but then Job.kill proceeds, setting status to KILLING. If the job runner thread proceeds only at this point, this branch is skipped:
https://github.com/juhoautio/azkaban/blob/723a46ec93d6e52f178e5d2ab445d630099d0ad6/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L597-L604

As a result the job exits with status KILLING and flow won't be able to terminate.

> Are you seeing this issue in your production environment now?

It was seen, but it's been patched in our fork (as in the second commit of this PR).

juhoautio:
> I described it as the desired state.

Oh, now your description is clear to me (how could I think about it that much wrong..). Indeed that's what my fix here ensures.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: If this line is hit
https://github.com/HappyRay/azkaban/blob/0d59dd63f0d694ffd9ef16368fd3b2e9230946e8/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java#L745
The finalStatus variable is set, so Status.isStatusFinished(finalStatus) will return true.


Your change will only have an effect when this.job.run(); doesn't throw an exception. Right? 
>  if (!Status.isStatusFinished(finalStatus) && !this.killed) {

What scenario is this modification designed to handle?

HappyRay:
Yes, I have looked at your example commit. It's great to have a repeatable program to demonstrate the problem.  The current code's complexity doesn't make it easy for me to remember all the details in my head. I am not very familiar with your killing status change. It would take me some time to research the code to understand. 

In the mean time, it will help me understand if you can describe the sequence of thread A does ... Thread B.... and here is the race condition. Here is how ... affect the timing...

Here is an example of how I describe race conditions. It's a style that I use.
https://github.com/azkaban/azkaban/commit/79f1f9927c151e226e9db601eaffc161df05ea8a

HappyRay:
My local build failed when I synced to your first commit: Prove job status synchronization bug when killing juhoautio 8/11/17, 2:31 AM 
I am not able to run the test.

jamiesjc:
I've read through the changes and descriptions. Correct me if I misunderstood anything. @juhoautio 
The main problem here is that the kill thread is synchronized but the job run thread is not. And both threads will change the job status which causes the race condition. 
One race condition happens in below scenario:
1. jetty thread:
    azkaban.execapp.JobRunner#kill
    azkaban.executor.Status#isStatusFinished -> status is not finished, does not return
2. switch to jobRunner thread:
    azkaban.execapp.JobRunner#runJob
    azkaban.execapp.JobRunner#changeStatus(azkaban.executor.Status) -> change status to SUCCEEDED
    azkaban.execapp.JobRunner#fireEvent(azkaban.event.Event, boolean) -> fire JOB_FINISHED event
3. switch back to jetty kill thread:
    set this.killed to true
    azkaban.jobExecutor.Job#cancel -> do the actual kill
    azkaban.execapp.JobRunner#changeStatus(azkaban.executor.Status) -> change status to KILLED

jamiesjc:
The fix is mainly to synchronize the places where #changeStatus would modify the job status incorrectly from different threads, e.g. changing it to SUCCEEDED from jobRunner thread first and then to FAILED from jettry kill thread. I think the fix makes sense. But it's definitely not easy to understand and maintain in the future. I would vote for @HappyRay's suggestion of reducing the overall complexity of the code if this is something we want to move towards finally.

> I am thinking of refactoring the DAG processing engine to reduce the complexity and reduce the need for thread synchronization. e.g. if there is one thread that is responsible for the job/flow status transition, it would be easier to reason about.

HappyRay:
Thanks @jamiesjc 
I can follow the race condition you described.
How would the change fix this race condition? I don't see what's preventing the jetty thread from setting the final status to killed.

juhoautio:
> I don't see what's preventing the jetty thread from setting the final status to killed.

jetty thread returns if status is a finished status. On the other hand if jetty thread gets to setting the status, then jobRunner thread won't set it any more, because it also checks if the status is finished (or if the job is killed), before changing the status.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
HappyRay: The rest of the code uses isKilled(). Any reason to use this.killed directly here?

HappyRay: I think I understand this logic now. 
This code is to set the final status to success when there is no exception. It is possible that the kill thread is executed between this line and a successful return of the job.run() call. This check is to prevent the status to be set to success in this scenario.

HappyRay:
I didn't realize that the whole logic in the kill azkaban.execapp.JobRunner#kill method is within this lock synchronized (this.syncObject) {

Your fix makes sense to me now. 
The code is still overly complex since the lock is all over the place. 

Could you try to summarize your technique in preventing the race condition between the kill thread and the jobrunner thread?
My understanding is that you would check killed status before setting the final job status and place that check and set logic inside the jobrunner syncObject lock?

Thanks for your contribution and patience!

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: Yeah this should also use `isKilled()`, definitely.

juhoautio:
> Could you try to summarize your technique in preventing the race condition between the kill thread and the jobrunner thread?
My understanding is that you would check killed status before setting the final job status and place that check and set logic inside the jobrunner syncObject lock?

Replace "check killed status" with "check killed flag" and we're there.

To sum it up in other words, I fix the problem by synchronizing in run & kill methods so that the combination of these operations are wrapped into "a single atomic operation":
- checking the status and/or killed flag
- based on the previous check result(s), setting the status and/or killed flag

HappyRay:
@juhoautio I can review and merge it if you could you update the change description and update the branch.

juhoautio:
Would you squash the commits when merging? I guess I don't need to create a new PR in that case :)

HappyRay:
Yes, the default merge option now is to rebase. 

Could you suggest the change description you would like to use?

juhoautio:
> Could you suggest the change description you would like to use?

Ok, let's see what you think about this (I edited the PR description):
https://github.com/azkaban/azkaban/pull/1330#issue-249587802

Feel free to edit it if there's anything.

juhoautio:
Hm, looks like master branch build is failing.

juhoautio:
It looks like there was also a random failure with `FlowRunnerTest > exec1Failed`.

This new PR aims to stabilize it: https://github.com/azkaban/azkaban/pull/1364

PR #1332, opened 2017-08-12
Event reporting for azkaban events.

rajagopr:
**Summary**
With this change, support for reporting events of interest from azkaban has been provided.
The default implementation uses a kafka event reporter. Users can provide alternate
implementations of the reporter. To begin with, the following events are reported.

 1. FLOW_STARTED
 2. FLOW_FINISHED
 3. JOB_STARTED
 4. JOB_FINISHED

In future, this can be easily extended to report other events. The default event reporter
implementation uses gobblin-metrics, which provides conevenient methods for event creation
and submission. Gobblin-metrics also provides a schema for the events. The default
implementation uses the kafka async producer.

**Configuration changes**
Note: All changes must be applied to the executor server or the solo-server if in solo mode.

// Property is used to enable/disable event reporting with the default being false.
azkaban.event.reporting.enabled=false

// Alternate implementations of the reporter can be specified using this property.
azkaban.event.reporting.class=com.foo.EventReporterImpl

// Kafka topic name for the default implementation.
azkaban.event.reporting.kafka.topic=TestTopicName

// Kafka broker list for the default implementation.
azkaban.event.reporting.kafka.brokers=hostname.com:port_num

// Schema registry server for the default kafka implementation.
azkaban.event.reporting.kafka.schema.registry.url=schemaRegistryUrl.com:port/schema

**Testing Done**
End-to-end tests were performed and the following scenarios were verified.

1. Consumed the emitted kafka events and messages received were as expected.
2. Event reporter enabled/disabled scenario.
3. Kafka broker down scenario.
4. Reporter enabled, but kafka brokers undefined in the config scenario.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/Constants.java
+++ b/azkaban-common/src/main/java/azkaban/Constants.java
@@ -116,6 +116,17 @@ public class Constants {
     public static final String AZKABAN_KEYTAB_PATH = "azkaban.keytab.path";
     public static final String PROJECT_TEMP_DIR = "project.temp.dir";
 
+    // Event reporting properties
+    public static final String AZKABAN_EVENT_REPORTING_CLASS_PARAM =
+        "azkaban.event.reporting.class";
+    public static final String AZKABAN_EVENT_REPORTING_ENABLED = "azkaban.event.reporting.enabled";
+    public static final String AZKABAN_EVENT_REPORTING_KAFKA_BROKERS =
+        "azkaban.event.reporting.kafka.brokers";
+    public static final String AZKABAN_EVENT_REPORTING_KAFKA_TOPIC =
+        "azkaban.event.reporting.kafka.topic";
+    public static final String AZKABAN_EVENT_REPORTING_KAFKA_SCHEMA_REGISTRY_URL =
+        "azkaban.event.reporting.kafka.schema.registry.url";
+
     /*
      * The max number of artifacts retained per project.
      * Accepted Values:
--- a/azkaban-exec-server/build.gradle
+++ b/azkaban-exec-server/build.gradle
@@ -4,6 +4,7 @@ dependencies {
     compile(project(':azkaban-common'))
 
     compile deps.kafkaLog4jAppender
+    compile deps.gobblinKafka
 
     runtime(project(':azkaban-hadoop-security-plugin'))
 
@@ -11,6 +12,16 @@ dependencies {
     testCompile(project(':azkaban-common').sourceSets.test.output)
 
     testRuntime deps.h2
+}
+
+configurations.compile {
+    exclude group: 'com.linkedin.gobblin', module: 'gobblin-api'
+    exclude group: 'com.linkedin.gobblin', module: 'gobblin-metrics-graphite'
+    exclude group: 'com.linkedin.gobblin', module: 'gobblin-metrics-hadoop'
+    exclude group: 'com.linkedin.gobblin', module: 'gobblin-metrics-influxdb'
+    exclude group: 'com.linkedin.gobblin', module: 'gobblin-runtime'
+
+    exclude group: 'org.projectlombok', module: 'lombok'
 }
 
 distributions {
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
@@ -17,9 +17,21 @@
 
 package azkaban.execapp;
 
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_EVENT_REPORTING_CLASS_PARAM;
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_EVENT_REPORTING_ENABLED;
+
+import azkaban.execapp.reporter.AzkabanKafkaAvroEventReporter;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.JdbcExecutorLoader;
+import azkaban.spi.AzkabanEventReporter;
+import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
 
 /**
  * This Guice module is currently a one place container for all bindings in the current module. This
@@ -28,9 +40,47 @@ import com.google.inject.AbstractModule;
  */
 public class AzkabanExecServerModule extends AbstractModule {
 
+  private static final Logger logger = Logger.getLogger(AzkabanExecServerModule.class);
+
   @Override
   protected void configure() {
     install(new ExecJettyServerModule());
     bind(ExecutorLoader.class).to(JdbcExecutorLoader.class);
   }
+
+  @Inject
+  @Provides
+  @Singleton
+  public AzkabanEventReporter createAzkabanEventReporter(final Props props) {
+    final boolean eventReporterEnabled =
+        props.getBoolean(AZKABAN_EVENT_REPORTING_ENABLED, false);
+
+    if (!eventReporterEnabled) {
+      logger.info("Event reporter is not enabled");
+      return null;
+    }
+
+    final Class<?> eventReporterClass =
+        props.getClass(AZKABAN_EVENT_REPORTING_CLASS_PARAM, AzkabanKafkaAvroEventReporter.class);
+    if (eventReporterClass != null && eventReporterClass.getConstructors().length > 0) {
+      this.logger.info("Loading event reporter class " + eventReporterClass.getName());
+      try {
+        final Constructor<?> eventReporterClassConstructor =
+            eventReporterClass.getConstructor(Props.class);
+        return (AzkabanEventReporter) eventReporterClassConstructor.newInstance(props);
+      } catch (final InvocationTargetException e) {
+        this.logger.error(e.getTargetException().getMessage());
+        if (e.getTargetException() instanceof IllegalArgumentException) {
+          throw new IllegalArgumentException(e);
+        } else {
+          throw new RuntimeException(e);
+        }
+      } catch (final Exception e) {
+        this.logger.error("Could not instantiate EventReporter " + eventReporterClass.getName());
+        throw new RuntimeException(e);
+      }
+    }
+    return null;
+  }
+
 }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -16,9 +16,10 @@
 
 package azkaban.execapp;
 
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_HOST_NAME;
+
 import azkaban.ServiceProvider;
 import azkaban.event.Event;
-import azkaban.event.Event.Type;
 import azkaban.event.EventData;
 import azkaban.event.EventHandler;
 import azkaban.event.EventListener;
@@ -42,6 +43,8 @@ import azkaban.metric.MetricReportManager;
 import azkaban.project.ProjectLoader;
 import azkaban.project.ProjectManagerException;
 import azkaban.sla.SlaOption;
+import azkaban.spi.AzkabanEventReporter;
+import azkaban.spi.EventType;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
 import azkaban.utils.SwapQueue;
@@ -91,10 +94,12 @@ public class FlowRunner extends EventHandler implements Runnable {
   private final Props azkabanProps;
   private final Map<String, Props> sharedProps = new HashMap<>();
   private final JobRunnerEventListener listener = new JobRunnerEventListener();
+  private final FlowRunnerEventListener flowListener = new FlowRunnerEventListener();
   private final Set<JobRunner> activeJobRunners = Collections
       .newSetFromMap(new ConcurrentHashMap<JobRunner, Boolean>());
   // Thread safe swap queue for finishedExecutions.
   private final SwapQueue<ExecutableNode> finishedNodes;
+  private final AzkabanEventReporter azkabanEventReporter;
   private Logger logger;
   private Appender flowAppender;
   private File logFile;
@@ -104,21 +109,16 @@ public class FlowRunner extends EventHandler implements Runnable {
   // Used for pipelining
   private Integer pipelineLevel = null;
   private Integer pipelineExecId = null;
-
   // Watches external flows for execution.
   private FlowWatcher watcher = null;
-
   private Set<String> proxyUsers = null;
   private boolean validateUserProxy;
-
   private String jobLogFileSize = "5MB";
   private int jobLogNumFiles = 4;
-
   private boolean flowPaused = false;
   private boolean flowFailed = false;
   private boolean flowFinished = false;
   private boolean flowKilled = false;
-
   // The following is state that will trigger a retry of all failed jobs
   private boolean retryFailedJobs = false;
 
@@ -127,9 +127,10 @@ public class FlowRunner extends EventHandler implements Runnable {
    */
   public FlowRunner(final ExecutableFlow flow, final ExecutorLoader executorLoader,
       final ProjectLoader projectLoader, final JobTypeManager jobtypeManager,
-      final Props azkabanProps)
+      final Props azkabanProps, final AzkabanEventReporter azkabanEventReporter)
       throws ExecutorManagerException {
-    this(flow, executorLoader, projectLoader, jobtypeManager, null, azkabanProps);
+    this(flow, executorLoader, projectLoader, jobtypeManager, null, azkabanProps,
+        azkabanEventReporter);
   }
 
   /**
@@ -137,7 +138,8 @@ public class FlowRunner extends EventHandler implements Runnable {
    */
   public FlowRunner(final ExecutableFlow flow, final ExecutorLoader executorLoader,
       final ProjectLoader projectLoader, final JobTypeManager jobtypeManager,
-      final ExecutorService executorService, final Props azkabanProps)
+      final ExecutorService executorService, final Props azkabanProps,
+      final AzkabanEventReporter azkabanEventReporter)
       throws ExecutorManagerException {
     this.execId = flow.getExecutionId();
     this.flow = flow;
@@ -154,10 +156,15 @@ public class FlowRunner extends EventHandler implements Runnable {
     this.executorService = executorService;
     this.finishedNodes = new SwapQueue<>();
     this.azkabanProps = azkabanProps;
+    // Add the flow listener only if a non-null eventReporter is available.
+    if (azkabanEventReporter != null) {
+      this.addListener(this.flowListener);
+    }
 
     // Create logger and execution dir in flowRunner initialization instead of flow runtime to avoid NPE
     // where the uninitialized logger is used in flow preparing state
     createLogger(this.flow.getFlowId());
+    this.azkabanEventReporter = azkabanEventReporter;
   }
 
   public FlowRunner setFlowWatcher(final FlowWatcher watcher) {
@@ -201,7 +208,7 @@ public class FlowRunner extends EventHandler implements Runnable {
       loadAllProperties();
 
       this.fireEventListeners(
-          Event.create(this, Type.FLOW_STARTED, new EventData(this.getExecutableFlow())));
+          Event.create(this, EventType.FLOW_STARTED, new EventData(this.getExecutableFlow())));
       runFlow();
     } catch (final Throwable t) {
       if (this.logger != null) {
@@ -226,7 +233,8 @@ public class FlowRunner extends EventHandler implements Runnable {
         closeLogger();
         updateFlow();
       } finally {
-        this.fireEventListeners(Event.create(this, Type.FLOW_FINISHED, new EventData(this.flow)));
+        this.fireEventListeners(
+            Event.create(this, EventType.FLOW_FINISHED, new EventData(this.flow)));
       }
     }
   }
@@ -287,7 +295,6 @@ public class FlowRunner extends EventHandler implements Runnable {
       this.logger.error("Error updating flow.", e);
     }
   }
-
 
   /**
    * setup logger and execution dir for the flowId
@@ -559,7 +566,7 @@ public class FlowRunner extends EventHandler implements Runnable {
   private void finishExecutableNode(final ExecutableNode node) {
     this.finishedNodes.add(node);
     final EventData eventData = new EventData(node.getStatus(), node.getNestedId());
-    fireEventListeners(Event.create(this, Type.JOB_FINISHED, eventData));
+    fireEventListeners(Event.create(this, EventType.JOB_FINISHED, eventData));
   }
 
   private void finalizeFlow(final ExecutableFlowBase flow) {
@@ -1099,20 +1106,79 @@ public class FlowRunner extends EventHandler implements Runnable {
     return ImmutableSet.copyOf(this.activeJobRunners);
   }
 
+  // Class helps report the flow start and stop events.
+  private class FlowRunnerEventListener implements EventListener {
+
+    public FlowRunnerEventListener() {
+    }
+
+    private synchronized Map<String, String> getFlowMetadata(final FlowRunner flowRunner) {
+      final ExecutableFlow flow = flowRunner.getExecutableFlow();
+      final Props props = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class);
+      final Map<String, String> metaData = new HashMap<>();
+      metaData.put("flowName", flow.getId());
+      metaData.put("azkabanHost", props.getString(AZKABAN_SERVER_HOST_NAME, "unknown"));
+      metaData.put("projectName", flow.getProjectName());
+      metaData.put("submitUser", flow.getSubmitUser());
+      metaData.put("executionId", String.valueOf(flow.getExecutionId()));
+      metaData.put("startTime", String.valueOf(flow.getStartTime()));
+      metaData.put("submitTime", String.valueOf(flow.getSubmitTime()));
+      return metaData;
+    }
+
+    @Override
+    public synchronized void handleEvent(final Event event) {
+      if (event.getType() == EventType.FLOW_STARTED) {
+        final FlowRunner flowRunner = (FlowRunner) event.getRunner();
+        final ExecutableFlow flow = flowRunner.getExecutableFlow();
+        FlowRunner.this.logger.info("Flow started: " + flow.getId());
+        FlowRunner.this.azkabanEventReporter.report(event.getType(), getFlowMetadata(flowRunner));
+      } else if (event.getType() == EventType.FLOW_FINISHED) {
+        final FlowRunner flowRunner = (FlowRunner) event.getRunner();
+        final ExecutableFlow flow = flowRunner.getExecutableFlow();
+        FlowRunner.this.logger.info("Flow ended: " + flow.getId());
+        final Map<String, String> flowMetadata = getFlowMetadata(flowRunner);
+        flowMetadata.put("endTime", String.valueOf(flow.getEndTime()));
+        flowMetadata.put("flowStatus", flow.getStatus().name());
+        FlowRunner.this.azkabanEventReporter.report(event.getType(), flowMetadata);
+      }
+    }
+  }
+
   private class JobRunnerEventListener implements EventListener {
 
     public JobRunnerEventListener() {
     }
 
+    private synchronized Map<String, String> getJobMetadata(final JobRunner jobRunner) {
+      final ExecutableNode node = jobRunner.getNode();
+      final Props props = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class);
+      final Map<String, String> metaData = new HashMap<>();
+      metaData.put("jobId", node.getId());
+      metaData.put("executionID", String.valueOf(node.getExecutableFlow().getExecutionId()));
+      metaData.put("flowName", node.getExecutableFlow().getId());
+      metaData.put("startTime", String.valueOf(node.getStartTime()));
+      metaData.put("jobType", String.valueOf(node.getType()));
+      metaData.put("azkabanHost", props.getString(AZKABAN_SERVER_HOST_NAME, "unknown"));
+      metaData.put("jobProxyUser",
+          jobRunner.getProps().getString("user.to.proxy", null));
+      return metaData;
+    }
+
     @Override
     public synchronized void handleEvent(final Event event) {
-
-      if (event.getType() == Type.JOB_STATUS_CHANGED) {
+      if (event.getType() == EventType.JOB_STATUS_CHANGED) {
         updateFlow();
-      } else if (event.getType() == Type.JOB_FINISHED) {
-        final JobRunner runner = (JobRunner) event.getRunner();
-        final ExecutableNode node = runner.getNode();
+      } else if (event.getType() == EventType.JOB_FINISHED) {
         final EventData eventData = event.getData();
+        final JobRunner jobRunner = (JobRunner) event.getRunner();
+        final ExecutableNode node = jobRunner.getNode();
+        if (FlowRunner.this.azkabanEventReporter != null) {
+          final Map<String, String> jobMetadata = getJobMetadata(jobRunner);
+          jobMetadata.put("jobStatus", node.getStatus().name());
+          jobMetadata.put("endTime", String.valueOf(node.getEndTime()));
+          FlowRunner.this.azkabanEventReporter.report(event.getType(), jobMetadata);
+        }
         final long seconds = (node.getEndTime() - node.getStartTime()) / 1000;
         synchronized (FlowRunner.this.mainSyncObj) {
           FlowRunner.this.logger.info("Job " + eventData.getNestedId() + " finished with status "
@@ -1127,12 +1193,18 @@ public class FlowRunner extends EventHandler implements Runnable {
           }
 
           FlowRunner.this.finishedNodes.add(node);
-          FlowRunner.this.activeJobRunners.remove(runner);
+          FlowRunner.this.activeJobRunners.remove(jobRunner);
           node.getParentFlow().setUpdateTime(System.currentTimeMillis());
           interrupt();
           fireEventListeners(event);
         }
-      } else if (event.getType() == Type.JOB_STARTED) {
+      } else if (event.getType() == EventType.JOB_STARTED) {
+        final EventData eventData = event.getData();
+        FlowRunner.this.logger.info("Job Started: " + eventData.getNestedId());
+        if (FlowRunner.this.azkabanEventReporter != null) {
+          final JobRunner jobRunner = (JobRunner) event.getRunner();
+          FlowRunner.this.azkabanEventReporter.report(event.getType(), getJobMetadata(jobRunner));
+        }
         // add job level checker
         final TriggerManager triggerManager = ServiceProvider.SERVICE_PROVIDER
             .getInstance(TriggerManager.class);
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -35,6 +35,8 @@ import azkaban.project.ProjectLoader;
 import azkaban.project.ProjectWhitelist;
 import azkaban.project.ProjectWhitelist.WhitelistType;
 import azkaban.sla.SlaOption;
+import azkaban.spi.AzkabanEventReporter;
+import azkaban.spi.EventType;
 import azkaban.storage.StorageManager;
 import azkaban.utils.FileIOUtils;
 import azkaban.utils.FileIOUtils.JobMetaData;
@@ -61,6 +63,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.io.FileUtils;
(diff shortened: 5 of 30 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -601,10 +601,9 @@ private void doRun() {
       finalStatus = changeStatus(Status.KILLED);
     }
 
-    logInfo(
-        "Finishing job " + this.jobId + getNodeRetryLog() + " at " + this.node.getEndTime()
-            + " with status " + this.node.getStatus());
-
+    final int attemptNo = this.node.getAttempt();
```
HappyRay: Why this change? @chengren311 refactored this code earlier.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -637,7 +636,8 @@ private Status prepareJob() throws RuntimeException {
         return null;
       }
 
-      logInfo("Starting job " + this.jobId + getNodeRetryLog() + " at " + this.node.getStartTime());
+      logInfo("Starting job " + this.jobId + (this.node.getAttempt() > 0 ? (" retry: " + this.node.getAttempt()) : "") + " at "
```
HappyRay: dito

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
```diff
@@ -0,0 +1,121 @@
...
+        new AzkabanKafkaPusher(eventReportingKafkaBrokers, eventReportingKafkaTopic);
+
+    try {
+      this.kafkaAvroEventReporter = builder.withKafkaPusher(kafkaPusher)
+          .withSchemaRegistry(new KafkaAvroSchemaRegistry(this.kafkaProps))
+          .build(eventReportingKafkaBrokers, eventReportingKafkaTopic);
+      logger.info("Initialized a new kafkaAvroEventReporter");
+    } catch (final Throwable t) {
+      logger.error("Exception while initializing kafka reporter");
+      t.printStackTrace();
```
HappyRay: Please log the exception in the logger.

build.gradle
```diff
@@ -135,7 +146,9 @@ subprojects {
             compile deps.log4j
             compile deps.guice
             compile deps.slf4j
-
+            // See: https://github.com/rzwitserloot/lombok/issues/1301
```
HappyRay: Does it mean that any version of error_prone_core newer than 2.0.15 will also work?

build.gradle
```diff
@@ -135,7 +146,9 @@ subprojects {
...
+            // See: https://github.com/tbroyer/gradle-errorprone-plugin#usage
+            errorprone 'com.google.errorprone:error_prone_core:2.0.15'
```
HappyRay: I specified the errorprone version recently too. Your way of specifying the dependency seems to be simpler.  How did you verify the specified version was used?

build.gradle
HappyRay: Does this change require any additional change to the IDE to support lombok?

azkaban-exec-server/build.gradle
```diff
@@ -4,6 +4,7 @@ dependencies {
     compile(project(':azkaban-common'))
 
     compile deps.kafkaLog4jAppender
+    compile deps.gobblinKafka
```
HappyRay: I assume we need to use the Linkedin internal Kafka when used inside linkedin.
How would that work?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -30,11 +37,49 @@
  */
 public class AzkabanExecServerModule extends AbstractModule {
 
+  private static final Logger logger = Logger.getLogger(AzkabanExecServerModule.class);
+  private static final String EVENT_REPORTER_CLASS_PARAM = "event.reporter.class";
```
HappyRay: Please define all configuration parameters in the azkaban.Constants class. This makes it easier to identify all the configuration keys.

HappyRay: Why isn't it prefixed with event.reporting like other configs?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -30,11 +37,49 @@
...
+  private static final String EVENT_REPORTER_ENABLED = "event.reporter.enabled";
```
HappyRay: Why isn't it prefixed with event.reporting like other configs?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -30,11 +37,49 @@
...
   @Override
   protected void configure() {
     bind(ExecutorLoader.class).to(JdbcExecutorLoader.class);
     bind(AzkabanExecutorServer.class).in(Scopes.SINGLETON);
     bind(TriggerManager.class).in(Scopes.SINGLETON);
     bind(FlowRunnerManager.class).in(Scopes.SINGLETON);
   }
+
+  @Inject
+  @Provides
```
HappyRay: Is the instance designed to be a singleton? If so annotate with @Singleton?
https://github.com/google/guice/wiki/Scopes

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -30,11 +37,49 @@
...
+  public AzkabanEventReporter createAzkabanEventReporter(final Props props) {
+    final boolean eventReporterEnabled = props.getBoolean(EVENT_REPORTER_ENABLED, false);
+
+    if (!eventReporterEnabled) {
+      logger.info("Event reporter is not enabled");
+      return null;
+    }
+
+    final Class<?> eventReporterClass =
+        props.getClass(EVENT_REPORTER_CLASS_PARAM, null);
```
HappyRay: How about providing a default "AzkabanKafkaAvroEventReporter" here so that you don't have to have a special case?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -91,10 +92,12 @@
   private final Props azkabanProps;
   private final Map<String, Props> sharedProps = new HashMap<>();
   private final JobRunnerEventListener listener = new JobRunnerEventListener();
+  private final FlowRunnerEventListener flowListener = new FlowRunnerEventListener();
```
HappyRay: Consider getting the listener instance from Guice. It would be great to refactor FlowRunner to use Guice. For now getting the dependency from azkaban.ServiceProvider is fine.

HappyRay: Same for JobRunnerEventListener. Refactoring can be done in multiple PRs.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1091,20 +1098,88 @@ public int getExecutionId() {
...
+
+    public FlowRunnerEventListener() {
+    }
+
+    private synchronized Map<String, String> getFlowMetadata(final FlowRunner flowRunner) {
+      final ExecutableFlow flow = flowRunner.getExecutableFlow();
+      final Map<String, String> metaData = new HashMap<>();
+      metaData.put("flowName", flow.getId());
+      metaData.put("azkabanHost",
+          flowRunner.azkabanProps.getString("azkaban.server.hostname", "unknown"));
```
HappyRay: Use constant azkaban.Constants.ConfigurationKeys#AZKABAN_SERVER_HOST_NAME

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1091,20 +1098,88 @@ public int getExecutionId() {
...
         updateFlow();
       } else if (event.getType() == Type.JOB_FINISHED) {
-        final JobRunner runner = (JobRunner) event.getRunner();
-        final ExecutableNode node = runner.getNode();
         final EventData eventData = event.getData();
+        final JobRunner jobRunner = (JobRunner) event.getRunner();
+        final ExecutableNode node = jobRunner.getNode();
+        if (FlowRunner.this.azkabanEventReporter != null) {
+          final Map<String, String> jobMetadata = getJobMetadata(jobRunner);
+          // data only available post call to prepareJob method, and hence added only to the
```
HappyRay: Do you mean azakbanHost? 
I thought it is defined in the config. You should be able to access it anywhere via Guice.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1091,20 +1098,88 @@ public int getExecutionId() {
...
+
   private class JobRunnerEventListener implements EventListener {
 
     public JobRunnerEventListener() {
     }
 
+    private synchronized Map<String, String> getJobMetadata(final JobRunner jobRunner) {
+      final ExecutableNode node = jobRunner.getNode();
+      final Map<String, String> metaData = new HashMap<>();
+      metaData.put("jobId", node.getId());
```
HappyRay: The keys in the metaData become part of the contract. Shall they be defined in an interface?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -358,7 +362,7 @@ public void submitFlow(final int execId) throws ExecutorManagerException {
 
     final FlowRunner runner =
         new FlowRunner(flow, this.executorLoader, this.projectLoader, this.jobtypeManager,
-            this.azkabanProps);
+            this.azkabanProps, this.azkabanEventReporter);
```
HappyRay: Have you considered adding the reporter to the flowRunner instance here instead?
And check if it is null. If it is, don't add it so that the report call won't be called and extra information won't be logged.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanEventReporter.java
```diff
@@ -0,0 +1,20 @@
...
+ * By default, a KafkaAvroEventReporter is provided. Alternate implementations
+ * can be provided by setting the property 'event.reporter.class'.
+ *
+ * The constructor will be called with a azkaban.utils.Props object passed as
+ * the only parameter. If such a constructor doesn't exist, then the AzkabanEventReporter
+ * instantiation will fail.
+ */
+public interface AzkabanEventReporter {
+
+  boolean report(Event event, Map<String, String> metadata);
```
HappyRay: Should the content of metadata be defined? Or is it the intention that the metadata should be treated as a black box by the reporter?

HappyRay: Event class is an implementation detail and it includes methods you don't need for this interface. Consider defining an event interface for this purpose?

We have an spi ( service provider interface ) module. e.g. azkaban.spi.Storage
Should this API go into that module? Clients can depend on this module to get the API definitions.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Default implementation of the <code>AzkabanEventReporter</code> class.
+ * An instance of this class is injected via guice into the <code>FlowRunnerManager</code>
+ * class when event reporting is enabled.
+ */
+public class AzkabanKafkaAvroEventReporter implements AzkabanEventReporter {
+
+  private static final Logger logger = Logger.getLogger(AzkabanKafkaAvroEventReporter.class);
+  private static final String EVENT_REPORTING_KAFKA_BROKERS = "event.reporting.kafka.brokers";
```
HappyRay: Move to azkaban.Constants. Similar for other config keys.

HappyRay: Also, we start to prefix the keys with Azkaban to make it clear that these keys are defined by Azkaban instead of defined by plugin code.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
```diff
@@ -0,0 +1,121 @@
...
+        KafkaAvroEventReporter.Factory.forContext(this.rootContext);
+    final KafkaPusher kafkaPusher =
+        new AzkabanKafkaPusher(eventReportingKafkaBrokers, eventReportingKafkaTopic);
+
+    try {
+      this.kafkaAvroEventReporter = builder.withKafkaPusher(kafkaPusher)
+          .withSchemaRegistry(new KafkaAvroSchemaRegistry(this.kafkaProps))
+          .build(eventReportingKafkaBrokers, eventReportingKafkaTopic);
+      logger.info("Initialized a new kafkaAvroEventReporter");
+    } catch (final Throwable t) {
```
HappyRay: Is catching RuntimeException enough?

build.gradle
```diff
@@ -40,6 +40,16 @@ allprojects {
     repositories {
         mavenCentral()
         mavenLocal()
+        jcenter()
```
HappyRay: Why are these additions necessary?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1091,20 +1098,88 @@ public int getExecutionId() {
...
+        final ExecutableFlow flow = flowRunner.getExecutableFlow();
+        FlowRunner.this.logger.info("Flow started: " + flow.getId());
+        FlowRunner.this.azkabanEventReporter.report(event, getFlowMetadata(flowRunner));
+      } else if (event.getType() == Type.FLOW_FINISHED) {
+        final FlowRunner flowRunner = (FlowRunner) event.getRunner();
+        final ExecutableFlow flow = flowRunner.getExecutableFlow();
+        FlowRunner.this.logger.info("Flow ended: " + flow.getId());
+        final Map<String, String> flowMetadata = getFlowMetadata(flowRunner);
+        flowMetadata.put("endTime", String.valueOf(flow.getEndTime()));
+        FlowRunner.this.azkabanEventReporter.report(event, flowMetadata);
```
HappyRay: What happens if the reporter is Null here?

HappyRay:
Have you done end to end tests?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
rajagopr: Accidental, I guess while I rebased. Will remove this.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
rajagopr: Will remove this.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
rajagopr: Done.

build.gradle
rajagopr: versions higher than 2.0.15 did not work and resulted in this issue. The suggested fix was to set to 2.0.15 version.

rajagopr: It's used in gobblin-metrics, I did not try using a lombok annotation directly in my class. I could run azkaban tests without anything extra.

build.gradle
rajagopr: Until I got this version, I was not able to build azkaban with gobblin metrics applied. Setting to this version solved this issue. I cross checked my local cache and could see this version.

azkaban-exec-server/build.gradle
rajagopr: kafka library used by gobblin metrics works fine within LinkedIn. We are using in metric store as well and did not observe any issues by using this.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: Sure, will do.

rajagopr: Reporter(noun) does reporting(verb). The noun form felt more apt to me especially for the config defining the class. I agree with you for the consistency part. I can keep it all as reporting. Let me know.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: Same as previous comment.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: For the current usage, marking it with `@singleton` makes sense.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1091,20 +1098,88 @@ public int getExecutionId() {
     return this.execId;
   }
 
+  public Set<JobRunner> getActiveJobRunners() {
+    return ImmutableSet.copyOf(this.activeJobRunners);
+  }
+
+  // Class helps report the flow start and stop events.
+  private class FlowRunnerEventListener implements EventListener {
```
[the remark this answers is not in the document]
rajagopr: Ok, I will take a look at that.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Ok, will use that.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: Yes, good point. I will change that.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Ok, I will work on that in a separate PR.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Yes, I did mean azkabanHost. Though it's possible to access the hostname via something like `AzkabanExecutorServer.getApp().getHost()`, I found it difficult to test the same. The getApp method is static and mockito doesn't allow mocking of static methods. I did not want to introduce another library for this. Let me know if there's a better way.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Metadata can be treated as a blackbox by the reporter.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
rajagopr: Sure, will do.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
rajagopr: I do not follow. Are you saying not to initialize `FlowRunner` with the event reporter in-case the reporter is null. Still, I would need to check if it's available or not while handling the event? Or, not to add the flowListener if the event reporter is null? Please let me know.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanEventReporter.java
rajagopr: Ok, let me check on these. I will make the suggested changes.

rajagopr: Metadata can be treated as a black-box though.

build.gradle
rajagopr: gobblin metrics seems to bring in some dependencies which are all over the place. Had to add these find all the libraries.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Null case is handled at the beginning of the method and it shouldn't reach here if it's null.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
rajagopr: Yea, that should be enough. Will change...

rajagopr:
@HappyRay - Yes, I have done end-to-end tests and everything seemed to work fine. Especially the following were checked.
1. Consumed the emitted kafka events and messages looked fine.
2. Event reporter enabled/disabled scenario.
3. Kafka broker down scenario
4. Reporter enabled, but kafka brokers undefined in the config scenario

I will get back with the suggested changes.

build.gradle
HappyRay: Not able to upgrade is still an issue.
According to https://github.com/rzwitserloot/lombok/issues/1332 this issue is fixed. Could you give it a try?
Also after you merge in the latest change, please use one way to specify the error-prone version.

azkaban-exec-server/build.gradle
HappyRay: Could you double check with the Kafka and Gobblin team? The open source version may work. But I thought the internal version has to be used. It's good to be prepared even if we can use the open source version for now.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
HappyRay: The convention is to prefix all azkaban config keys with azkaban.
Reporting can be interpreted as a noun too such as financial reporting, right?
Something like
azkaban.event.reporting....?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: This refactoring is tricky. May have to introduce additional classes to hold state and the stateless part of the class can be injected via Guice.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: AzkabanExecutorServer.getApp() is deprecated. Don't use in the new code.
Try azkaban.ServiceProvider instead for now to get this information.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: Please define how the metadata is supposed to be used. 
The schema of the metadata is still part of the contract between the producer (azkaban) and the consumers of this information. Right? If so, where is this contract defined?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: I meant " not to add the flowListener if the event reporter is null". Does it make sense?

build.gradle
HappyRay: What are the additional dependencies?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: I see. See the earlier comment on not adding the reporter to the listener list if it is null.

build.gradle
HappyRay: It appears that this change brought in a large number of dependencies.

```
 ./gradlew :azkaban-exec-server:dependencies
+--- com.linkedin.gobblin:gobblin-kafka-08:0.11.0
|    +--- com.linkedin.gobblin:gobblin-api:0.11.0
|    |    +--- org.slf4j:jcl-over-slf4j:1.7.21
|    |    |    \--- org.slf4j:slf4j-api:1.7.21
|    |    +--- commons-io:commons-io:2.5
|    |    +--- io.reactivex.rxjava2:rxjava:2.1.0
|    |    |    \--- org.reactivestreams:reactive-streams:1.0.0
|    |    +--- org.apache.hadoop:hadoop-common:2.3.0 -> 2.6.1 (*)
|    |    +--- com.google.guava:guava:15.0 -> 23.0 (*)
|    |    +--- com.google.code.gson:gson:2.6.2 -> 2.8.1
|    |    +--- com.typesafe:config:1.2.1
|    |    +--- org.jasypt:jasypt:1.9.2
|    |    +--- org.slf4j:slf4j-api:1.7.21
|    |    +--- org.projectlombok:lombok:1.16.8
|    |    +--- org.reflections:reflections:0.9.10
|    |    |    +--- com.google.guava:guava:18.0 -> 23.0 (*)
|    |    |    +--- org.javassist:javassist:3.18.2-GA
|    |    |    \--- com.google.code.findbugs:annotations:2.0.1
|    |    +--- commons-codec:commons-codec:1.10
|    |    +--- org.apache.commons:commons-lang3:3.4
|    |    +--- org.slf4j:slf4j-log4j12:1.7.21 (*)
|    |    +--- joda-time:joda-time:2.9.3
|    |    +--- org.apache.hadoop:hadoop-mapreduce-client-core:2.3.0
|    |    |    +--- org.apache.hadoop:hadoop-yarn-common:2.3.0
|    |    |    |    +--- org.apache.hadoop:hadoop-yarn-api:2.3.0
|    |    |    |    |    +--- commons-lang:commons-lang:2.6
|    |    |    |    |    +--- com.google.guava:guava:11.0.2 -> 23.0 (*)
|    |    |
...
```
Many of the compile time dependencies are not esssential.
This is not ideal. It will increase the maintenance cost. Is the abstraction worth it? How difficult will it be to have a direct dependency on Kafka?

rajagopr: @HappyRay - I will check what's the alternate library to depend upon.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
rajagopr: done.

build.gradle
rajagopr: With the latest version of error-prone, this issue has gone away.

azkaban-exec-server/build.gradle
rajagopr: I have checked this with Aditya in the gobblin team and he confirmed that this is the right way to use it.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: done.

rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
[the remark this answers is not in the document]
rajagopr: We can do this refactoring as part of a separate PR.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: Are you suggesting to use an interface with constants? This is generally considered a bad practice. Refer [this](https://stackoverflow.com/questions/320588/interfaces-with-static-fields-in-java-for-sharing-constants/320642#320642). As part of the metadata consumers will get a map which is what we mentioned in the interface AzkabanEventReporter.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanEventReporter.java
rajagopr: Ok, I have created a separate event interface for this purpose and moved the event interface and AzkabanEventReporter interface to the azkaban.spi module.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
rajagopr: done.

azkaban-exec-server/src/main/java/azkaban/execapp/reporter/AzkabanKafkaAvroEventReporter.java
rajagopr: done.

build.gradle
rajagopr: Checked with aditya from the gobblin team and he confirmed that the usage is right.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
rajagopr: done.

rajagopr:
@HappyRay - I have made all the suggested changes except refactoring of the FlowRunner class which can be taken up separately.

Updating error-prone version to the latest (tried 2.1.1 and 2.1.0) did not work out well and I had to switch back to 2.0.15 to get the build going.

Please review.

HappyRay:
> Updating error-prone version to the latest (tried 2.1.1 and 2.1.0) did not work out well and I had to switch back to 2.0.15 to get the build going.


It's not a blocker. But could you try to find a way to overcome this limitation?

build.gradle
HappyRay: I am still concerned about the large additional dependency this change brought in.

What additional jars would this change bring in?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecServerModule.java
```diff
@@ -17,10 +17,21 @@
...
+import azkaban.execapp.reporter.AzkabanKafkaAvroEventReporter;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.JdbcExecutorLoader;
+import azkaban.spi.AzkabanEventReporter;
+import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
+import com.google.inject.Inject;
+import com.google.inject.Provides;
 import com.google.inject.Scopes;
+import com.google.inject.Singleton;
```
HappyRay: Use Javax. See the new coding convention. #1435

azkaban-exec-server/src/test/java/azkaban/execapp/AzkabanExecServerModuleTest.java
```diff
@@ -0,0 +1,110 @@
...
+    props.put("azkaban.event.reporting.enabled", "true");
+    props.put("azkaban.event.reporting.class",
+        "azkaban.execapp.AzkabanEventReporterTest1");
+    final AzkabanEventReporter azkabanEventReporter = azkabanExecServerModule
+        .createAzkabanEventReporter(props);
+    assertThat(azkabanEventReporter).isNotNull();
+    assertThat(azkabanEventReporter).isInstanceOf(AzkabanEventReporterTest1.class);
+  }
+
+  /**
```
HappyRay: Why is this test commented out?

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
```diff
@@ -0,0 +1,19 @@
+package azkaban.spi;
+
+import java.util.Map;
+
+/**
+ * Implement this interface to report flow and job events. Event reporter
+ * can be turned on by setting the property 'azkaban.event.reporting.enabled' to true.
```
HappyRay: Refer to the constant instead using javadoc syntax. https://stackoverflow.com/questions/17496038/javadoc-link-to-method-in-other-class ?

HappyRay:
Thanks for updating the PR. 

Could you resolve the conflicts?

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
```diff
@@ -0,0 +1,19 @@
...
+ * By default, a KafkaAvroEventReporter is provided. Alternate implementations
+ * can be provided by setting the property 'azkaban.event.reporting.class'.
+ *
+ * The constructor will be called with a azkaban.utils.Props object passed as
+ * the only parameter. If such a constructor doesn't exist, then the AzkabanEventReporter
+ * instantiation will fail.
+ */
+public interface AzkabanEventReporter {
+
+  boolean report(IEvent event, Map<String, String> metadata);
```
HappyRay: Why is it better than use EventType directly in this interface?

azkaban-exec-server/src/test/java/azkaban/execapp/AzkabanExecServerModuleTest.java
rajagopr: That was marked to fix as `createAzkabanEventReporter` had changes. It's fixed and enabled now.

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
rajagopr: I have referred to the constants and added `@code` marker. It's not possible to link to classes outside of a module using `@link` or `@see`.

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
rajagopr: Changed to use EventType directly.

rajagopr:
@HappyRay - Excluded the unwanted modules from gobblin including lombok which fixed the compatibility issue with error-prone. Fixed other suggested changes. Performed the tests again and everything worked as expected. I was able to consume the emitted kafka messages.

Please review.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
HappyRay: Yes, I've read the Effective Java book and remember this recommendation. My understanding is that it is not a problem if the constants are part of the contract.

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
HappyRay: Are you sure? I was able to use e.g. 
{@link azkaban.utils.AbstractMailer}

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -116,6 +116,17 @@
     public static final String AZKABAN_KEYTAB_PATH = "azkaban.keytab.path";
     public static final String PROJECT_TEMP_DIR = "project.temp.dir";
 
+    // Event reporting properties
```
HappyRay: Could you add more detailed comments to this sections to explain how to use these configs? This can be a separate PR.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -151,11 +154,13 @@ public FlowRunnerManager(final Props props,
       final ExecutorLoader executorLoader,
       final ProjectLoader projectLoader,
       final StorageManager storageManager,
-      final TriggerManager triggerManager) throws IOException {
+      final TriggerManager triggerManager,
+      @Nullable final AzkabanEventReporter azkabanEventReporter) throws IOException {
```
HappyRay: Nice usage of @Nullable.

azkaban-spi/src/main/java/azkaban/spi/AzkabanEventReporter.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.Map;
+
+/**
+ * Implement this interface to report flow and job events. Event reporter
+ * can be turned on by setting the property {@code AZKABAN_EVENT_REPORTING_ENABLED} to true.
+ *
+ * By default, a KafkaAvroEventReporter is provided. Alternate implementations
+ * can be provided by setting the property {@code AZKABAN_EVENT_REPORTING_CLASS_PARAM}
+ * <br><br>
```
HappyRay: The convention is to use <p> see the google style guide and use checkstyle to check. You can change it later.

HappyRay:
Thanks!

PR #1335, opened 2017-08-14
creating API and UI to compose and publish messages

kunkun-tang:
Whatever production issues (e.g., hadoop crash, DB down) hits (Or, is schedule to hit) Azkaban, we are thinking about creating a simple UI Note system to inform AZ users the production issues as soon as we can. In this code patch, a UI Note system is proposed to resolve this issue.

AZ admins are able to compose the error/warning message and publish the contents into AZ memory. Then AZ navigation banner will display the message/error to every Azkaban user who's accessing AZ page. The message will disappear after 30 seconds.
screen shot 2017-08-14 at 3 36 32 pm

Evaluated in my local laptop.

The figure (warning.png) in my PR is from [this repo](https://github.com/google/material-design-icons), which follows Apache 2.0 license.


reallocf:
What's the benefit of having the message disappear after 30 seconds? Wouldn't it be better to keep it live until an admin takes it down?

kunkun-tang:
I suppose users should be aware of that and don't need it after 30 seconds. Actually, even though it disappears even now, it will show up when the user navigates to another tab page, like history, proejcts. That said, as long as users click something, the message will show up.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Map;
+import javax.servlet.ServletConfig;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.apache.log4j.Logger;
+
+public class NoteServlet extends LoginAbstractAzkabanServlet {
+
+  private static final long serialVersionUID = 1L;
```
jamiesjc: What does this serialVersionUID mean? Is it used anywhere?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public void init(final ServletConfig config) throws ServletException {
+    super.init(config);
+    this.server = (AzkabanWebServer) getApplication();
+  }
+
+  @Override
+  protected void handleGet(final HttpServletRequest req, final HttpServletResponse resp,
+      final Session session) throws ServletException, IOException {
+    if(isAdmin(session.getUser())) {
```
jamiesjc: What do you think of adding some errorMsg in the page if a user without ADMIN permission tries to access this page?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
```diff
@@ -0,0 +1,125 @@
...
+      }
+    } catch (final Exception e) {
+      ret.put("error", e.getMessage());
+    }
+    this.writeJSON(resp, ret);
+  }
+
+  private void ajaxAddNotes(final HttpServletRequest req,
+                            final Map<String, Object> ret, final User user) throws ServletException,
+      TriggerManagerException {
```
jamiesjc: When will this throw TriggerManagerException?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private void handleAJAXAction(final HttpServletRequest req,
+      final HttpServletResponse resp, final Session session) throws ServletException,
+      IOException {
+    final HashMap<String, Object> ret = new HashMap<>();
+    final String ajaxName = getParam(req, "ajax");
+    try {
+      if (ajaxName.equals("addNote")) {
+        ajaxAddNotes(req, ret, session.getUser());
+      } else {
```
jamiesjc: If someone enters some invalid parameters in the API call, it will just remove the note right? For that reason, I think it would be better to verify `else if (ajaxName.equals("removeNote"))`

jamiesjc:
One UI improvement suggestion:
When users move their mouse on the banner, it's better to show something different so that they know they can click into it and redirect them to the GCN jira page.
Something like the hover selector https://www.w3schools.com/cssref/sel_hover.asp

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
```diff
@@ -0,0 +1,125 @@
...
+                               final Map<String, Object> ret, final User user) throws ServletException,
+      TriggerManagerException {
+    type = null;
+    message= null;
+    url = null;
+    logger.info("removing note from memory.");
+    ret.put("status", "success");
+  }
+
+  private boolean isAdmin(final User user) {
```
jamiesjc: It seems we already have similar method which does the permission check: azkaban.server.HttpRequestUtils#hasPermission. Can we reuse it instead of creating a new method?

kunkun-tang: Good catch! I will use the existing method.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
kunkun-tang: Actually, it should be never used. Removing this exception.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
kunkun-tang: Since this class extends `java.io.Serializable`, it is recommended to explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization.

From: [link](https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it)

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
kunkun-tang: Yes! You are right! modifying.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
kunkun-tang: Good suggestion. Adding.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/NoteServlet.java
suvodeep-pyne: Question: Generally, this is set to a random number instead of `1L`. Does that matter?

PR #1336, opened 2017-08-15
remove job runner from active job runner list when job finishes

chengren311:
activeJobRunners maintains all active job runners, so we should remove inactive job runners when job finishes. As a consequence, activeJobRunners will keep queuing up and querying activeJobRunners by job name will return invalid job runner. This PR fixes this issue.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -1119,6 +1119,7 @@ public class FlowRunner extends EventHandler implements Runnable {
           }
 
           FlowRunner.this.finishedNodes.add(node);
+          activeJobRunners.remove(runner);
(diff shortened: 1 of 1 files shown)


HappyRay:
Could you describe why this change is needed?

chengren311:
@HappyRay @kunkun-tang added more explanation. But I do think the change is self-explanatory.

HappyRay:
How did the issue manifest itself? What problems will users or operators observe?

chengren311:
@HappyRay the issue surfaced when jobs were not killed successfully by SLA in occasion cases. It turned out it's killed by an invalid job runner returned by querying the active job runner by job id.

HappyRay:
I am not sure I follow. What's the sequence of the events?

And before your change, how was the list cleaned up?

chengren311:
@HappyRay 
No one cleaned the list up.
We can talk offline.

HappyRay:
Thanks for adding additional context.

Issue #1333, opened 2017-08-14
Standardize injection annotations


HappyRay:
Proposal:

use 
import javax.inject.Inject;
import javax.inject.Singleton;

instead of
import com.google.inject.Inject;
import com.google.inject.Singleton;

see https://github.com/google/guice/wiki/JSR330

> Best Practices
> 
> Prefer JSR-330's annotations and Provider interface.
> 

see http://docs.oracle.com/javaee/6/api/javax/inject/package-summary.html

HappyRay:
Fixed by #1434 
and documented in the coding convention. #1435

PR #1346, opened 2017-08-16
StorageManager to clean up project artifacts based on config

suvodeep-pyne:
Added a config `azkaban.storage.artifact.max.retention` (default: 0)
Based on the storage parameter the files are removed and the project_versions table entries are removed.

```
    /*
     * The max number of artifacts retained per project.
     * Accepted Values:
     * - 0 : Save all artifacts. No clean up is done on storage.
     * - 1, 2, 3, ... (any +ve integer 'n') : Maintain 'n' latest versions in storage
     *
     * Note: Having an unacceptable value results in an exception and the service would REFUSE
     * to start.
     *
     * Example:
     * a) azkaban.storage.artifact.max.retention=all
     *    implies save all artifacts
     * b) azkaban.storage.artifact.max.retention=3
     *    implies save latest 3 versions saved in storage.
     **/
    public static final String AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION = "azkaban.storage.artifact.max.retention";

```

### DIFF
--- a/azkaban-common/src/main/java/azkaban/Constants.java
+++ b/azkaban-common/src/main/java/azkaban/Constants.java
@@ -110,6 +110,23 @@ public class Constants {
     public static final String AZKABAN_KERBEROS_PRINCIPAL = "azkaban.kerberos.principal";
     public static final String AZKABAN_KEYTAB_PATH = "azkaban.keytab.path";
     public static final String PROJECT_TEMP_DIR = "project.temp.dir";
+
+    /*
+     * The max number of artifacts retained per project.
+     * Accepted Values:
+     * - 0 : Save all artifacts. No clean up is done on storage.
+     * - 1, 2, 3, ... (any +ve integer 'n') : Maintain 'n' latest versions in storage
+     *
+     * Note: Having an unacceptable value results in an exception and the service would REFUSE
+     * to start.
+     *
+     * Example:
+     * a) azkaban.storage.artifact.max.retention=all
+     *    implies save all artifacts
+     * b) azkaban.storage.artifact.max.retention=3
+     *    implies save latest 3 versions saved in storage.
+     **/
+    public static final String AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION = "azkaban.storage.artifact.max.retention";
   }
 
   public static class FlowProperties {
--- a/azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
+++ b/azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package azkaban.storage;
+
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION;
+import static com.google.common.base.Preconditions.checkArgument;
+
+import azkaban.db.DatabaseOperator;
+import azkaban.spi.Storage;
+import azkaban.utils.Props;
+import com.google.common.annotations.VisibleForTesting;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
+
+@Singleton
+public class StorageCleaner {
+
+  // Delete records of all older versions
+  static final String SQL_DELETE_RESOURCE_ID = "DELETE FROM project_versions WHERE resource_id=?";
+
+  /**
+   * The query must sort the versions in reverse order for the cleanup operation to work correctly!
+   * TODO spyne: Refactor database storage cleanup to use this
+   *
+   * When using DatabaseStorage, resourceId is always NULL. Hence, those rows will currently be
+   * never cleaned up.
+   */
+  static final String SQL_FETCH_PVR = "SELECT resource_id FROM project_versions WHERE project_id=? AND resource_id IS NOT NULL ORDER BY version DESC";
+
+  private static final Logger log = Logger.getLogger(StorageCleaner.class);
+  private final DatabaseOperator databaseOperator;
+  private final int maxArtifactsPerProject;
+  private final Storage storage;
+
+  @Inject
+  public StorageCleaner(final Props props, final Storage storage,
+      final DatabaseOperator databaseOperator) {
+    this.storage = storage;
+    this.databaseOperator = databaseOperator;
+
+    this.maxArtifactsPerProject = props.getInt(AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION, 0);
+    checkArgument(this.maxArtifactsPerProject >= 0,
+        String.format("Invalid value for %s : %d", AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION,
+            this.maxArtifactsPerProject));
+
+    if (isCleanupPermitted()) {
+      log.info(String.format("%s Config: Max %d artifact(s) retained per project",
+          AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION, this.maxArtifactsPerProject));
+    } else {
+      log.warn("Project cleanup disabled. All artifacts will be stored.");
+    }
+  }
+
+  @VisibleForTesting
+  boolean isCleanupPermitted() {
+    return this.maxArtifactsPerProject > 0;
+  }
+
+  /**
+   * Remove all but last N artifacts as configured by AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION
+   *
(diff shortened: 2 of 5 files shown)


azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
```diff
@@ -0,0 +1,156 @@
...
+  @VisibleForTesting
+  boolean isCleanupPermitted() {
+    return this.maxArtifactsPerProject > 0;
+  }
+
+  /**
+   * Remove all but last N artifacts as configured by AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION
+   *
+   * Since multiple versions can share the same filename, the algo is to collect all filenames and
+   * from them, remove the latest ones. The remaining ones are deleted by the respective storage.
```
kunkun-tang: we remove old ones but keep latest ones, right?

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @param projectId project ID
+   */
+  public void cleanupProjectArtifacts(final int projectId) {
+    if (!isCleanupPermitted()) {
+      return;
+    }
+    final List<VersionResourcePair> pvrList = fetchPvrList(projectId);
```
kunkun-tang: Due to some legacy reasons, the current `project_version` table might be messy. It does not correspond what you observe in `project_files` table. For instance, project `az_test` might only have one entry in `project_files`, but it has too many entries in `project_versions`. Have you considered that? Not very sure about that because we changed schema by bringing `resouce_id`, but we need to be careful about this.

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -110,6 +110,17 @@
     public static final String AZKABAN_KERBEROS_PRINCIPAL = "azkaban.kerberos.principal";
     public static final String AZKABAN_KEYTAB_PATH = "azkaban.keytab.path";
     public static final String PROJECT_TEMP_DIR = "project.temp.dir";
+
+    /*
+     * The max number of artifacts retained per project.
+     * Example: If this config is set to 3, then we have at latest 3 versions saved in storage.
+     * Constraints:
+     * - Has to be Integer
+     * - Negative values are not allowed and will crash the app.
```
HappyRay: Please make it more explicit that the service won't start.

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -110,6 +110,17 @@
...
+     * - A value of 0 implies all artifacts will be stored [Default]
```
HappyRay: Is 1 a better default?

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
```diff
@@ -0,0 +1,156 @@
...
+    if (pvrList.size() <= this.maxArtifactsPerProject) {
+      return;
+    }
+    cleanUpStorageArtifacts(projectId, pvrList);
```
HappyRay: What happens when the web server crashes after this step?

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
```diff
@@ -0,0 +1,156 @@
...
+    log.warn(String.format("Deleting project artifacts [id: %d]: %s", projectId, allResourceIds));
+    allResourceIds.forEach(this.storage::delete);
+  }
+
+  private void cleanUpDatabase(final int projectId, final int lowestRetainedVersion) {
+    this.projectLoader.cleanOlderProjectVersion(projectId, lowestRetainedVersion);
+    try {
+      this.databaseOperator.update(SQL_DELETE_OLDER_VERSIONS, projectId, lowestRetainedVersion);
+    } catch (final SQLException e) {
+      log.error("Error while deleting version entries from DB for project ID: " + projectId, e);
```
HappyRay: When we get this exception, how do we recover?

HappyRay: And how do we plan to detect this error?

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -31,6 +31,7 @@
 import azkaban.utils.Md5Hasher;
 import azkaban.utils.Props;
 import com.google.inject.Inject;
+import com.google.inject.Singleton;
```
HappyRay: see #1333

azkaban-common/src/main/java/azkaban/Constants.java
reallocf: Can we have a value like None or False to imply all artifacts are stored?

Setting "azkaban.storage.artifact.max.retention=0" makes me think no artifacts are being retained.

Or maybe introduce a boolean true/false config corresponding to whether or not any artifacts are being cleaned?

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
suvodeep-pyne: The deletion process in `project_version` marks the project as `inactive` instead of deleting it. That's why project versions has a whole bunch of records but other tables do not.

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
suvodeep-pyne: I changed the code a bit here. The current code will simply catch the exception, log it and continue.

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
suvodeep-pyne: yes. Else, it would be a big problem!

suvodeep-pyne: Oh! now I get it. What I mean by the comment is that the latest ones are removed from the delete list. The remaining items in the list get removed.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
```diff
@@ -114,6 +119,8 @@ public void uploadProject(
       log.info(String.format("Added project metadata to DB. Meta:%s File: %s[%d bytes] URI: %s",
           metadata, localFile.getName(), localFile.length(), resourceId));
     }
+
```
suvodeep-pyne: I think if upload fails, then it probably will raise an exception and the delete code will not execute.

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
```diff
@@ -0,0 +1,156 @@
...
+        .collect(Collectors.toSet());
+    allResourceIds.removeAll(doNotDeleteSet);
+
+    log.warn(String.format("Deleting project artifacts [id: %d]: %s", projectId, allResourceIds));
+    allResourceIds.forEach(this.storage::delete);
+  }
+
+  private void cleanUpDatabase(final int projectId, final int lowestRetainedVersion) {
+    this.projectLoader.cleanOlderProjectVersion(projectId, lowestRetainedVersion);
+    try {
```
suvodeep-pyne: The database cleanup is tricky. There are several tables which are used to show data to the UI. This may be a bigger item but we need to improve the cleanup strategy there as well. Currently, this deals with just the storage cleanup.

azkaban-common/src/main/java/azkaban/Constants.java
suvodeep-pyne: done.

azkaban-common/src/main/java/azkaban/Constants.java
suvodeep-pyne: @HappyRay 
I would prefer to stick to `all (0)` as the default for now since it is in line with the current implementation. Therefore any new build doesn't suddenly change the behavior of the app. It's only when you turn on this knob, you see a behavior change.

@reallocf 
I like the idea! It is much more explicit. However, I'm a bit reluctant to introduce that complexity in an almost non-user facing, and probably seldom used config variable. This keeps it very crisp. Less validation. Less code.
The same goes for introducing another boolean config variable. Thoughts?

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
suvodeep-pyne: Even if we face this error, this shouldn't affect the execution of Azkaban. In the next upload, the clean up will be attempted again. The recovery piece was missing and has been added.

azkaban-common/src/main/java/azkaban/storage/StorageManager.java
suvodeep-pyne: updated.

azkaban-common/src/main/java/azkaban/Constants.java
HappyRay: I am fine with keeping 0 as the default.

kunkun-tang:
LGTM.

azkaban-common/src/main/java/azkaban/storage/StorageCleaner.java
HappyRay: So there is no harm in leaving the older version entries in the DB even when older storage files are deleted, right?

suvodeep-pyne: The storage cleanup needs to be done before the DB cleanup. The reason is that the reference to the files in the storage is maintained by a `key` which is present in the db in the `project_verions` table.

So every time we try to cleanup, we find out all the versions of a project and remove the entries that we need to retain. The rest of the items are deleted one by one using the delete API call.

So as long as those entries are in the DB, they will eventually be cleaned up. However, vice versa is not true. If the database entry doesn't exist, then Azkaban has no idea what items exist in the storage and cannot clean them up. This ensures a conservative cleanup and mandates that the storage is cleaned up before the metadata in the DB is removed.

So, to answer your question, there is no harm in leaving the older versions entries in the DB. If the files aren't present in the DB it will simply log it and continue.

PR #1351, opened 2017-08-16
Open access of /jmx and /stats page to all users.

jamiesjc:
Tested on localhost:
1. Add in azkaban-users.xml:
`<user username="azktest1" password="azktest1" group="azk1" roles="all-except-admin"/>`
2. Launch solo server with username azktest1. 
3. Verify azktest1 can access /jmx and /stats page.


jamiesjc:
In the current azkaban-users.xml configuration, `azkaban `user's role is admin. So it will always have the permission to access /jmx and /stats even without my change.

suvodeep-pyne:
Oh I see. Makes sense.

PR #1353, opened 2017-08-17
remove deprecated jdbc project classes

kunkun-tang:
#1079 introduced the new `JdbcProjectImpl` class to leverage new DB API. It has been in production for a while, so it is safe to remove deprecated code.


HappyRay:
Looks like the test failure is due to a flaky test
https://travis-ci.org/azkaban/azkaban/builds/265386286?utm_source=github_status&utm_medium=notification
I restarted it.

PR #1345, opened 2017-08-16
Create ExecutionFlowDBManager class to migrate to use new azkaban db module

kunkun-tang:
The goal of this code patch is to gradually migrate to new azkaban DB module. I construct a new class `ExecutorFlowDBManager ` to store Table `execution_flow` related DB operations. Basically, I move a few methods out from `JdbcExecutorLoader` to new class and leverage  new DB API.  There are 3 benefits for the new class:

* Split #1291 into a few commits, and will make reviews easier.
* Put Query Handler class and its caller together. In this way, make code clean and easy to read.
* Also easy to manage unit Tests.

I'm considering using `ExecutorFlowDBManager` directly to replace `JdbcExecutorLoader`, when it comes to `execution_flows` related operations. However, Since some classes like `FlowRunner` has not been guificied, it is not easy to add another dependency. We will do the follow-up refactor after executor loader is completed migrated.

Evaluated via Integrate Tests.

PR #1357, opened 2017-08-17
One refactor for Executor operations from JdbcExecutorLoader class

kunkun-tang:
This code patch is a follow-up of #1345. Similar to that PR, I start moving executor related operations. The first step is to create a delegate class and move all related methods to the new class. 

Note that **I didn't change anything, but only move methods from one place to another place**. After this PR, I will change the implementation inside, and add unit test. The benefits of this PR is to help team do review.


azkaban-common/src/main/java/azkaban/executor/ExecutorDBManager.java
```diff
@@ -0,0 +1,213 @@
...
+ * the License.
+ */
+
+package azkaban.executor;
+
+import azkaban.database.AbstractJdbcLoader;
+import azkaban.db.DatabaseOperator;
+import azkaban.metrics.CommonMetrics;
+import azkaban.utils.Props;
+import com.google.inject.Singleton;
```
suvodeep-pyne: As per @HappyRay suggestion, please use `javax.inject package`. check #1333

azkaban-common/src/main/java/azkaban/executor/ExecutorDBManager.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.inject.Inject;
+import org.apache.commons.dbutils.QueryRunner;
+import org.apache.commons.dbutils.ResultSetHandler;
+import org.apache.log4j.Logger;
+
+@Singleton
+public class ExecutorDBManager extends AbstractJdbcLoader {
```
suvodeep-pyne: This pattern is pretty much a DAO pattern. Should we rename it to `ExecutorDao` then?

https://stackoverflow.com/questions/19154202/data-access-object-dao-in-java
https://en.wikipedia.org/wiki/Data_access_object

suvodeep-pyne:
Another thing is to have tests for all the `DAO` methods. This can be a future thing but it's important since this is the class doing the actual DB work / has the core business logic.

azkaban-common/src/main/java/azkaban/executor/ExecutorDBManager.java
kunkun-tang: That's a good suggestion. I was considering using this name, but I didn't adopt it as I thought we don't have explicit `Executor` class, which sounded necessary for a DAO pattern to me. Now that there's no objection, I can rename it to DAO. Likely we could add and use Executor Class in future refactor.

azkaban-common/src/main/java/azkaban/executor/ExecutorDBManager.java
kunkun-tang: Modifying.

azkaban-common/src/main/java/azkaban/executor/ExecutorDBManager.java
suvodeep-pyne: Well, there is `azkaban.executor.Executor`. However, this is just a nitpick. I leave this to you.

This change looks good to go. I like the idea of moving the core data access layer to `azkaban-db`.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -32,14 +31,14 @@
 import org.apache.commons.dbutils.ResultSetHandler;
 import org.apache.log4j.Logger;
 
-@Singleton
-public class ExecutionFlowDBManager {
+@javax.inject.Singleton
```
suvodeep-pyne: why not import `javax.inject.Singleton` instead of using the entire path?

kunkun-tang: Good finding! I also noticed just now and am modifying.

HappyRay:
Could you run saveAction and run CheckStyle on new files?

PR #1359, opened 2017-08-17
Remove unused code and consolidate common logics

kunkun-tang:
This code patch does two refactor works:
* remove `FetchExecutableFlows`, as new ExecutionFlowDBManager already takes over this work.
* consolidate set up test DB in a common place, such that every test is able to call it.


azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -1402,70 +1387,6 @@ public String handle(final ResultSet rs) throws SQLException {
     }
   }
 
-  private static class FetchExecutableFlows implements
```
suvodeep-pyne: Is this dead code? Was it being used earlier?

suvodeep-pyne: Oh. got it. this refactor already went it.

PR #1361, opened 2017-08-18
Build, test: move test data files to a non-resource directory

HappyRay:
## Issues:
The test data files are under the resource directory in the test
submodule.
Gradle tests will use the jar file that contains the data files. This
makes it harder to use these files as files directly. As a result the
build script of other submodules has to use the build directory as
dependencies instead of using the standard project dependency.

Recent changes in intellij make the matter worse by insisting on using
its own build output directory. When compiling from intellij along
without building the test submodule first in gradle, the compilation
will fail because intellij can't find the dependent classes.

When navigating to the dependent classes defined in the test submodule,
intellij will jump to the decompiled class instead of the correct source
 files.

## Fix:
- Move the data files to a top level directory in the test submodule.
- Provide static methods to locate these files instead of having the
other submodules to directly hardcode the directory paths.
- Change gradle build files to use the project dependency.
- Move the java code in the test submodule to the main directory since
this code is not a test of itself but rather code to be used by other
modules.
- Add unit tests for the test module itself.


HappyRay:
reference:

https://stackoverflow.com/questions/5644011/multi-project-test-dependencies-with-gradle/21484721#21484721 This seems to be a better solution because it works better with intellij's dependency management system.  This is the approach I took in this change.

This is an alternative approach.

https://softnoise.wordpress.com/2014/09/07/gradle-sub-project-test-dependencies-in-multi-project-builds/

PR #1362, opened 2017-08-18
Continue DB API migration as a follow-up of #1357

kunkun-tang:
This code patch is a follow-up of #1357 . I replaced the DB operations by calling new DB API, and added Unite Tests.


azkaban-common/src/main/java/azkaban/executor/ExecutorDao.java
```diff
@@ -16,89 +16,68 @@
...
 
 @Singleton
-public class ExecutorDao extends AbstractJdbcLoader {
+public class ExecutorDao {
 
   private static final Logger logger = Logger.getLogger(ExecutorDao.class);
   private final DatabaseOperator dbOperator;
 
   @Inject
-  public ExecutorDao(final Props props, final CommonMetrics commonMetrics,
```
suvodeep-pyne: This is great! I really like the simplification of dependencies. Also +1 for removing the `extends` as well.

azkaban-common/src/test/java/azkaban/executor/ExecutorDaoTest.java
```diff
@@ -0,0 +1,190 @@
...
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ExecutorDaoTest {
```
suvodeep-pyne: Awesome work adding tests! 👍 👍

PR #1364, opened 2017-08-18
Improve InteractiveTestJob stability

juhoautio:
Wait for the job to appear for 10 seconds instead of 1 second. This should help avoid test failures when they're running slower than usual for some reason.

If wait is interrupted, have another round (i--) of trying, so that the method always waits _at least_ 10 seconds before it gives up.

----

This should fix random failures like:
```
azkaban.execapp.FlowRunnerTest > exec1Failed FAILED
    java.lang.IllegalStateException: job6 wasn't added in testJobs map
        at azkaban.executor.InteractiveTestJob.getTestJob(InteractiveTestJob.java:54)
        at azkaban.execapp.FlowRunnerTestBase.succeedJobs(FlowRunnerTestBase.java:164)
        at azkaban.execapp.FlowRunnerTest.exec1Failed(FlowRunnerTest.java:172)
:azkaban-solo-server:test FAILED
```

As seen on this Travis build:
https://travis-ci.org/azkaban/azkaban/builds/265699991?utm_source=github_status&utm_medium=notification


HappyRay:
Thanks for the PR!

PR #1366, opened 2017-08-18
Improve JobRunnerTest stability

juhoautio:
- Remove some redundant sleeping, replace with more eager waiting
- Add sleep to another place to increase stability

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -286,9 +286,11 @@ public class JobRunnerTest {
     final Thread thread = new Thread(runner);
     thread.start();
 
-    Thread.sleep(2000);
+    waitForStatus(node, Status.READY);
+    // sleep so that job has time to get into delayExecution() -> wait()
+    Thread.sleep(1000L);
     runner.kill();
-    Thread.sleep(500);
+    waitForStatus(node, Status.KILLED);
 
     eventCollector.handleEvent(Event.create(null, Event.Type.JOB_FINISHED, new EventData(node)));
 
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -286,9 +286,11 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
     final Thread thread = new Thread(runner);
     thread.start();
 
-    Thread.sleep(2000);
+    waitForStatus(node, Status.READY);
+    // sleep so that job has time to get into delayExecution() -> wait()
+    Thread.sleep(1000L);
```
juhoautio: This could also be the previous 2 seconds and offer maybe better stability, but on the other hand it will make the test run take longer than before, because below we need to add a new 1 s sleep.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -348,4 +352,19 @@ private JobRunner createJobRunner(final int execId, final String name, final int
     return runner;
   }
 
+  private void waitForStatus(ExecutableNode node, Status status) {
```
HappyRay: This method looks the same as https://github.com/HappyRay/azkaban/blob/96ad15a98af88adb1c17b7486802ed016bd109f1/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java#L120
It would be great to consolidate them.

HappyRay:
LGTM overall.

Do these two changes have to be in the same PR?

juhoautio:
It should work if the test fix is merged first separately. I didn't realize that at first. I can split this.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
juhoautio: Do you have a suggestion what to do? Create a new test helper class maybe?

HappyRay: Yes, that will be great.

HappyRay:
Yes, splitting them would be ideal.

Issue #1363, opened 2017-08-18
SLA with no action should not apply


chengren311:
Currently if we setup SLA but with no action specified like this:
![image](https://user-images.githubusercontent.com/1428327/29443519-8d40e2fa-838d-11e7-93d5-c87e06ecaf05.png) there will be no warning, and schedule page will show "has SLA" is true:
![image](https://user-images.githubusercontent.com/1428327/29443540-bc4286ee-838d-11e7-84d1-7c76d683c760.png)

Action should be required when setting up SLA.

PR #1370, opened 2017-08-20
disallow no action SLA

chengren311:
This PR is attempting to solve issue https://github.com/azkaban/azkaban/issues/1363.
Currently if we setup SLA but with no action specified, there will be no warning, and schedule page will show "has SLA" is true.
However, action should be required when setting up SLA. 

This PR added a non empty check on SLA actions when adding SLA for a schedule. Warning window will show up if action is not specified when setting SLA.
Old check was not checking the number of actions at all when updating SLA. The null check was used when uploading project logs.
Verified manually, but will automate the test in the internal integration test framework

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
@@ -204,14 +204,16 @@ public class ScheduleServlet extends LoginAbstractAzkabanServlet {
         }
       }
 
+      if (slaOptions.isEmpty()) {
+        throw new ScheduleManagerException(
+            String.format("SLA for schedule %s must have at least one action", scheduleId));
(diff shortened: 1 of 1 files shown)


HappyRay:
The code looks fine to me.

Could you describe the issue in the change description to make it easier for readers?

Could you describe more details of the fix e.g. what's wrong with the existing code and how you fixed it? 

What's your test strategy? Can it be automated?

chengren311:
@HappyRay updated the description.

HappyRay:
Thanks.

Am I right that the old code's check if an action is specified is ineffective since the list object itself will never be null? The new code checks if the list is empty instead. It would be good to include this details in the commit message as well when you merge.

chengren311:
@HappyRay the old check was not checking the number of actions at all when updating SLA. The null check was used when uploading project logs.

Issue #1371, opened 2017-08-20
status of the flow killed by job-level SLA should be killed rather than failed


chengren311:
Currently status of flow killed by flow-level SLA is killed, but job-level SLA is failed(see image below). We should make it "killed" as well for consistency.
![image](https://user-images.githubusercontent.com/1428327/29484109-b6ae3afe-846c-11e7-98b2-4051908ed46e.png)

PR #1372, opened 2017-08-21
Set flow status to killed when killed by violating job-level SLA

chengren311:
This PR is to solve issue https://github.com/azkaban/azkaban/issues/1371. 
When a job is killed by violating SLA and no more retries for that job is available, flow will be set as "Killed" instead of "Failed Finishing".

This PR also includes small reformatting done automatically by save plugin.


juhoautio:
> Currently there's no unit case covering FlowRunner#run

Huh? What about
https://github.com/azkaban/azkaban/blob/master/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
https://github.com/azkaban/azkaban/blob/master/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java

chengren311:
@juhoautio hah, i missed the FlowRunnerTest2.java, thanks!

PR #1373, opened 2017-08-21
Improve JobRunnerTest stability

juhoautio:
- Remove some redundant sleeping, replace with more eager waiting
- Add sleep to another place to increase stability
- Extracted common purpose method into a new helper class

Aims to fix this kind of random failure (happened in current master):
https://travis-ci.org/azkaban/azkaban/builds/266111102

```
azkaban.execapp.JobRunnerTest > testDelayedExecutionCancelledJob FAILED
    java.lang.AssertionError: expected:<2> but was:<1>
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at azkaban.execapp.JobRunnerTest.testDelayedExecutionCancelledJob(JobRunnerTest.java:309)
```

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -286,9 +286,11 @@ public class JobRunnerTest {
     final Thread thread = new Thread(runner);
     thread.start();
 
-    Thread.sleep(2000);
+    StatusTestUtils.waitForStatus(node, Status.READY);
+    // sleep so that job has time to get into delayExecution() -> wait()
+    Thread.sleep(1000L);
     runner.kill();
-    Thread.sleep(500);
+    StatusTestUtils.waitForStatus(node, Status.KILLED);
 
     eventCollector.handleEvent(Event.create(null, Event.Type.JOB_FINISHED, new EventData(node)));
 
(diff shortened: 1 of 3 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/StatusTestUtils.java
```diff
@@ -0,0 +1,23 @@
+package azkaban.execapp;
+
+import azkaban.executor.ExecutableNode;
+import azkaban.executor.Status;
+
+public class StatusTestUtils {
+
+  public static void waitForStatus(final ExecutableNode node, final Status status) {
```
HappyRay: Could you use this method to replace the same logic in 
https://github.com/HappyRay/azkaban/blob/96ad15a98af88adb1c17b7486802ed016bd109f1/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestBase.java#L120
It can be done in a separate PR.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -297,15 +299,18 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
...
+    Assert.assertTrue(node.getStartTime() - startTime <= 4000);
     Assert.assertTrue(runner.isKilled());
 
     final File logFile = new File(runner.getLogFilePath());
     final Props outputProps = runner.getNode().getOutputProps();
     Assert.assertTrue(outputProps == null);
     Assert.assertTrue(logFile.exists());
 
+    // sleep so that there's time to make the "DB update" for KILLED status
+    Thread.sleep(1000L);
```
HappyRay: An extra one second delay in test is not ideal. It would be great if you can think of ways to make it better.

HappyRay:
No luck. Enabled and disabled travis in the settings page. See if it helps.
https://travis-ci.org/azkaban/azkaban/pull_requests

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
juhoautio: Actually original had 2s + 0,5s. My version has 1s + 1s.

This sleep could be replaced with a "poll (with a timeout) until assert succeeds". Do you know of some standard way with junit or assertj?

azkaban-exec-server/src/test/java/azkaban/execapp/StatusTestUtils.java
juhoautio: Huh? That's what I did, see above :)

HappyRay: Oh, I read it wrong. Sorry.

PR #1377, opened 2017-08-21
Clear test dependency management by create azkaban-db-{api, test} modules

kunkun-tang:
This PR is a follow-up of #1361 , which aimed at clearing the submodule dependency management. Change being made in this pr:

* create azkaban-db-test and azkaban-db-api sub modules.
* move DB's `abstract` class to  azkaban-db-api , and move Test class to `azkaban-db-test`. In this way, we can call up azkaban-db-test in `azkaban-common` module.
* azkaban-db will rely on azkaban-db-api.


azkaban-common/build.gradle
```diff
@@ -34,6 +34,7 @@ model {
 dependencies {
     compile project(':azkaban-spi')
     compile project(':azkaban-db')
+    compile project(':azkaban-db-test')
```
HappyRay: Shouldn't it be testCompile?

azkaban-db/build.gradle
```diff
@@ -18,6 +18,9 @@ apply plugin: 'distribution'
 
 dependencies {
 
+    compile project(':azkaban-db-api')
+    compile project(':azkaban-db-test')
```
HappyRay: Shouldn't it be testCompile?

azkaban-common/build.gradle
kunkun-tang: Yes, you are tight!

azkaban-db/build.gradle
kunkun-tang: Yes, you are tight!

kunkun-tang:
We can discuss together with @HappyRay

PR #1378, opened 2017-08-21
create num execution DAO class

kunkun-tang:
As a follow-up of #1345, I'm rewriting fetchNumExecutableFlows and fetchNumExecutableNodes methods in order to leverage new DB API.

Added corresponding unit tests.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
@@ -1062,7 +1062,7 @@ public class JdbcExecutorLoaderTest {
     return new JdbcExecutorLoader(props,
         new CommonMetrics(new MetricsManager(new MetricRegistry())), null
         , null, null, null, null,
-        null, null);
+        null, null, null);
   }
 
   private boolean isTestSetup() {
--- a/azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.executor;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import azkaban.db.DatabaseOperator;
+import azkaban.test.Utils;
+import azkaban.utils.TestUtils;
+import java.sql.SQLException;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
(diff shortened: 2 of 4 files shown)


azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
```diff
@@ -0,0 +1,92 @@
...
+    flow2b.setStatus(Status.FAILED);
+    this.executionFlowDao.uploadExecutableFlow(flow2b);
+
+    final int count = this.numExecutionsDao.fetchNumExecutableFlows();
+    assertThat(count).isEqualTo(3);
+
+    final int flow2Count = this.numExecutionsDao.fetchNumExecutableFlows(1, "derived-member-data-2");
+    assertThat(flow2Count).isEqualTo(2);
+  }
+  @Test
```
suvodeep-pyne: nit: Can probably add a line in between methods.

azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
```diff
@@ -0,0 +1,92 @@
...
+import azkaban.test.Utils;
+import azkaban.utils.TestUtils;
+import java.sql.SQLException;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class NumExecutionsDaoTest {
```
suvodeep-pyne: Is this class testing 2 classes here? If yes, then it might be cleaner to split this into separate test classes. If this db setup code is common for too many classes, then we need to think of a way to consolidate them.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
```diff
@@ -1062,7 +1062,7 @@ private ExecutorLoader createLoader() {
     return new JdbcExecutorLoader(props,
         new CommonMetrics(new MetricsManager(new MetricRegistry())), null
         , null, null, null, null,
-        null, null);
+        null, null, null);
```
suvodeep-pyne: So the tests run only when the db exists? Else everything passes?

azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
```diff
@@ -0,0 +1,92 @@
...
+    dbOperator = Utils.initTestDB();
+  }
+
+  @AfterClass
+  public static void destroyDB() throws Exception {
+    try {
+      dbOperator.update("DROP ALL OBJECTS");
+      dbOperator.update("SHUTDOWN");
+    } catch (final SQLException e) {
+      e.printStackTrace();
```
suvodeep-pyne: Let's use a logger here.

azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
kunkun-tang: No, I was actually testing only one class, but need two class dependencies in test.

azkaban-common/src/test/java/azkaban/executor/JdbcExecutorLoaderTest.java
kunkun-tang: This class `JdbcExecutorLoaderTest` is going to be deprecated.

azkaban-common/src/test/java/azkaban/executor/NumExecutionsDaoTest.java
kunkun-tang: I didn't know we should also use logger for test, Any benefits? @suvodeep-pyne

kunkun-tang: Merging this now. If necessary, I can do a separate PR to fix this.

PR #1379, opened 2017-08-21
rewrite assign executor jdbc methods

kunkun-tang:
Still, this PR is a follow-up of #1345. Last but not least, I'm moving assignExecutor and UnAssignExecutor to a new DAO file.

The reason I don't create a new test file:
* Not usre how much in-memory h2 database consume resources. We could use only one test class (in-memory h2) to do testing.


azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -176,6 +182,47 @@ public void testFetchQueuedFlows() throws Exception {
...
+  }
+
+  /* Test exception when assigning a non-existent executor to a flow */
+  @Test
+  public void testAssignExecutorInvalidExecutor() throws Exception {
+    final ExecutableFlow flow = TestUtils.createExecutableFlow("exectest1", "exec1");
+    this.executionFlowDao.uploadExecutableFlow(flow);
+
+    assertThatThrownBy(
+        () -> this.assignExecutor.assignExecutor(flow.getExecutionId(), 1))
```
jamiesjc: Should it be assignExecutor(1, flow.getExecutionId())? Also how do you know 1 is a non-existent executor?

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -176,6 +182,47 @@ public void testFetchQueuedFlows() throws Exception {
...
+
+  /* Test exception when assigning an executor to a non-existent flow execution */
+  @Test
+  public void testAssignExecutorInvalidExecution() throws Exception{
+    final String host = "localhost";
+    final int port = 12345;
+    final Executor executor = this.executorDao.addExecutor(host, port);
+
+    assertThatThrownBy(
+        () -> this.assignExecutor.assignExecutor(2, executor.getId()))
```
jamiesjc: Similar to above comment. Need to switch the two arguments.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
kunkun-tang: Great Catch! fixed and added a comment.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
kunkun-tang: Great Catch! fixed and added a comment.

PR #1380, opened 2017-08-22
Fix build failure by suppressing MissingFail warning

jamiesjc:
Fix below build failure error:

image

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/restli/DeployProjectTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/restli/DeployProjectTest.java
@@ -49,6 +49,7 @@ public class DeployProjectTest {
   }
 
   @Test
+  @SuppressWarnings("MissingFail")
   public void testErrorDeploy() {
     final ProjectManagerResource resource = new ProjectManagerResource();
     final Map<String, ValidationReport> reports = new LinkedHashMap<>();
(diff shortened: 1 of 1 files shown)


kunkun-tang:
This test is quite flaky, and my mac never complains.

reallocf:
Is there any way we can pass this warning in an expected way so we don't have to suppress it? I feel like we should only suppress warnings that we really have to suppress.

Link to related errorprone bug pattern: http://errorprone.info/bugpattern/MissingFail

Look at the "To reduce false positives" section in the link above and they detail some ways to get around this test in a safe way without suppression.

Does that make sense?

HappyRay:
This seems to be a legitimate warning. I wonder why I never saw this warning in my local build and why Travis build would pass.

azkaban-web-server/src/test/java/azkaban/restli/DeployProjectTest.java
```diff
@@ -49,6 +49,7 @@ public void testWarnedDeploy() {
   }
 
   @Test
+  @SuppressWarnings("MissingFail")
```
HappyRay: Would adding an Assert.fail(); statement at the end of the try block as the examples in 
http://errorprone.info/bugpattern/MissingFail work better?

```java
 public void expectedException_withFail() {
    try {
      dummyMethod();
      Assert.fail();
    } catch (Exception expected) {
    }
  }
```

jamiesjc:
Thanks @reallocf and @HappyRay for the comments! I've fixed it using the suggested way. 
This is probably something new in the errorProne bug pattern, so it only starts to show in the latest build.

HappyRay:
Maybe this change introduced this check.
https://github.com/google/error-prone/commit/70e0ad796a4bad56e065a1880dbf880932dfcda2

PR #1381, opened 2017-08-22
hiding internals of azkaban-db tests in other module

kunkun-tang:
This PR is a follow-up of #1361 , which aimed at clearing the submodule dependency management. We explicitly depend on some internals from `azkaban-db`. It would be nice if we could hide this detail.

The solution originated from:
https://softnoise.wordpress.com/2014/09/07/gradle-sub-project-test-dependencies-in-multi-project-builds/

### DIFF
--- a/azkaban-db/build.gradle
+++ b/azkaban-db/build.gradle
@@ -16,6 +16,19 @@
  */
 apply plugin: 'distribution'
 
+task testJar(type: Jar, dependsOn: testClasses) {
+    from sourceSets.test.output
+    classifier = 'test'
+}
+
+configurations {
+    testOutput
+}
+
+artifacts {
+    testOutput testJar
+}
+
 dependencies {
 
     // todo kunkun-tang: consolidate dependencies in azkaban-common and azkaban-db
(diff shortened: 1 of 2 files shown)


azkaban-db/build.gradle
```diff
@@ -16,6 +16,19 @@
  */
 apply plugin: 'distribution'
 
+task testJar(type: Jar, dependsOn: testClasses) {
+    from sourceSets.test.output
+    classifier = 'test'
+}
+
+configurations {
```
HappyRay: Is it better to modified it to
configurations {
    testOutput.extendsFrom (testCompile)
}
as the article suggested?

suvodeep-pyne: @kunkun-tang 
Can you test what @HappyRay suggested. In that case, we may not even have to add a `testJar` task.

HappyRay: I think testJar task is still needed. 
This change will give it more flexibility.

kunkun-tang: Based on this article, `testJar` can wrap the test.outputs to a jar format, then hide details.

kunkun-tang: Resolved the fix.

PR #1382, opened 2017-08-22
build: use a fixed google errorprone version

HappyRay:
## Issue:
"The plugin adds an errorprone configuration that automatically uses the
 latest release of error-prone. You can override it to use a specific
 version "
see https://github.com/tbroyer/gradle-errorprone-plugin

It appears that by latest version it will use the last locally cached
version. This caused inconsistency in builds.

Also, it introduces additional variables in the build. 
e.g. see #1380 

## Fix:
Specify a fixed version

Will update to the latest version after the internal Artifactory
is updated with the latest version.

## Test:
Delete the gradle cache and run the build again and verify in the
console output that the specified version is downloaded.


HappyRay:
Yes, this is the same version as published in our internal artifactory.

PR #1384, opened 2017-08-23
The final commit for JDBC Executor rewrite

kunkun-tang:
Still, this PR is a follow-up of #1345. Last but not least, I'm migrating remained methods in JdbcExecutorloader to use new DB API. This commit should be the last one of the series.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ActiveExecutingFlowsDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ActiveExecutingFlowsDao.java
@@ -21,6 +21,8 @@ import java.sql.SQLException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+//TODO jamiesjc: This class is deprecated as we don't fetch active_execution_flow table any long.
+// So this clas should be removed onwards.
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/ActiveExecutingFlowsDao.java
```diff
@@ -16,52 +16,43 @@
...
 import azkaban.db.DatabaseOperator;
-import azkaban.metrics.CommonMetrics;
-import azkaban.utils.Props;
 import java.sql.SQLException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
-import org.apache.commons.dbutils.QueryRunner;
 
+//TODO jamiesjc: This class is deprecated as we don't fetch active_execution_flow table any long.
+// So this clas should be removed onwards.
```
suvodeep-pyne: typo in 'class'. Let's also mark it `@deprecated` then.

kunkun-tang: Good suggestion!

PR #1386, opened 2017-08-23
Switch Metastore calls to use RetryingMetastoreClient

herrlich:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -780,4 +789,27 @@ private Text getMRTokenRenewerInternal(final JobConf jobConf) throws IOException
         hsProxy.getConnectAddress());
   }
 
+  // we use this function to create retrying metastore client
```
sbeeram: Use Javadoc format for method level comments; it would be better to say, "Method to create a metastore client that retries on failures". Update the method name to be createRetryingMetastoreClient

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -780,4 +789,27 @@ private Text getMRTokenRenewerInternal(final JobConf jobConf) throws IOException
...
+  private IMetaStoreClient createMetaStoreClient(HiveConf hiveConf) throws MetaException {
+    // the hookLoader object is not used in this function, but we will keep it for potential storage handler settings.
```
sbeeram: The comment is misleading; what you could say is "Custom hook-loader to return a HiveMetaHook if the table is configured with a custom storage handler"

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -17,6 +17,7 @@
 package azkaban.security;
 
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_NATIVE_LIB_FOLDER;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.*;
```
suvodeep-pyne: We have just 1 constant that we are using from `hive_metastoreConstants` right? Should we just import that?

herrlich: Yes, it seems that we only have META_TABLE_STORAGE under that should be imported. But intellij auto-imported it as "org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.*". Should I change that and start a new pull request?

HappyRay: If you follow the developer setup instruction in the README, import the google style xml and use the latest Intellj, it appears that it will import individual classes. Could you give it a try?

herrlich: Yes that solves the problem. Already updated the code

PR #1392, opened 2017-08-25
format: auto format the db module

HappyRay:


HappyRay:
@suvodeep-pyne Thanks. 
I agree that these are drawbacks. On balance, I think consistency and the ability to take full advantage of IDE are worth the cost.

PR #1409, opened 2017-08-30
remove unused test file and DB methods [refactor]

kunkun-tang:
Since we already migrated to use new DB API, it is safe to delete unused methods and deprecated tests.


azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -44,20 +39,17 @@
...
-      final ExecutorDao executorDao,
-      final ExecutionJobDao executionJobDao,
-      final ExecutionLogsDao executionLogsDao,
-      final ExecutorEventsDao executorEventsDao,
-      final ActiveExecutingFlowsDao activeExecutingFlowsDao,
-      final FetchActiveFlowDao fetchActiveFlowDao,
-      final AssignExecutorDao assignExecutorDao,
-      final NumExecutionsDao numExecutionsDao) {
-    super(props, commonMetrics);
+  public JdbcExecutorLoader(final ExecutionFlowDao executionFlowDao,
```
HappyRay: The format seems to be different from the google style. Did you import and activate the new style config? See the readme.

kunkun-tang: I should be using Google Style. I just checked the details of the style. It should allow multiline when method declaration parameter is too long. We may check together to make sure google style is working.

kunkun-tang:
@HappyRay PR updated.

PR #1410, opened 2017-08-30
Fixing DB failover functionalities

kunkun-tang:
When Mysql failovers from master to slave, it is expected to get hang at every DB operation, and keeps retrying to find correct connection. The Mysql cluster is a master-slave achitecture, where slave is read-only. I was testing the new DB API in our staging AZ cluster and testing DB, I ran into the below issues:
* After Mysql failovers (from master to slave), AZ keeps having the previous Mysql DataSource (previous master), which had been switched to read-only. That said, the dataSource remains cached, and doesn't get refresh to be the new dataSource.

* The first step of Mysql failover is to enforce master to be read-only. By then, AZ is not able to write to DB. The running job logs was missing. The observation during this stage is that AZ keeps throwing SQLException:

> java.sql.SQLException: The MySQL server is running with the --read-only option so it cannot execute this statement Query: INSERT INTO execution_flows (project_id, flow_id, version, status, submit_time, submit_user, update_time) values (?,?,?,?,?,?,?) Parameters:...

The proposed solution in this PR is 
* check if the current Mysql is read-only. If it is, it will keep retrying. 
* if AZ can not find satisfied SQL connection, we create new concrete DataSource, rather than using existing one. As a result, we override createDataSource method.

Applying this code patch, We made an extensive test on testing cluster. During the DB failover, the only Exception I was seeing is:

> 2017/08/29 23:58:11.051 +0000 ERROR [MySQLDataSource] [Azkaban] Failed to find write-enabled DB connection. Wait 1 minutes and retry. No.Attempt = 1
java.sql.SQLException: Failed to find DB connection Or connection is read only.
        at azkaban.db.MySQLDataSource.getConnection(MySQLDataSource.java:75)
        at org.apache.commons.dbutils.AbstractQueryRunner.prepareConnection(AbstractQueryRunner.java:175)
....

During retrying, AZ UI gets a hang when I try accessing history page. No other exceptions.
After retrying 3 attempts, AZ is back to normal. I can not observe any failures.


azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
          * Every Attempt generates a thread-hanging-time, about 75 seconds, which is hard coded, and can not be changed.
```
HappyRay: Is 75 seconds correct?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
+      return value != 0;
+    }
+    throw new SQLException("can not fetch read only value from DB");
+  }
+
+  private void sleep(final long milliseconds) {
+    try {
+      Thread.sleep(milliseconds);
+    } catch (final InterruptedException e) {
+      logger.error("can not sleep", e);
```
HappyRay: Sleep interrupted?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
         retryAttempt++;
+        this.dataSource = null;
       }
     }
-    return null;
+    return connection;
+  }
+
+  /**
+   * This method keeps the original functionality from parent class. The difference is that we want to
```
HappyRay: This doesn't seem to be ideal. Is there no better way?

HappyRay: Have you tried calling invalidateConnection and maybe clear the connection pool too when the connection is readonly to force DBCP to reconnect to the new host?
https://commons.apache.org/proper/commons-dbcp/apidocs/src-html/org/apache/commons/dbcp2/BasicDataSource.html#line.1971

http://commons.apache.org/proper/commons-pool/api-2.4.2/org/apache/commons/pool2/impl/GenericObjectPool.html#clear--
https://commons.apache.org/proper/commons-dbcp/apidocs/org/apache/commons/dbcp2/BasicDataSource.html#getConnectionPool--

HappyRay: Another possibility is to wrap the datasource and proxy calls to it.
https://stackoverflow.com/questions/5007371/refresh-datasource-using-springdbcp

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: why does Connection getConnection()  has to be synchronized? Generally I think keeping the synchronize scope as small as possible especially getConnection will potentially hanging for a long period, and given createDataSource().getConnection() is already synchronized, should we not mark this method synchronized?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
+    final Statement stmt = conn.createStatement();
+    final ResultSet rs = stmt.executeQuery("SELECT @@global.read_only");
+    if (rs.next()) {
+      final int value = rs.getInt(1);
+      return value != 0;
+    }
+    throw new SQLException("can not fetch read only value from DB");
+  }
+
+  private void sleep(final long milliseconds) {
```
chengren311: consider https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Uninterruptibles.html#sleepUninterruptibly(long, java.util.concurrent.TimeUnit) ?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
+        } else {
           return connection;
         }
       } catch (final SQLException ex) {
-        logger.error(
-            "Failed to find DB connection. waits 1 minutes and retry. No.Attempt = " + retryAttempt,
-            ex);
-      } finally {
+
+        logger.error( "Failed to find write-enabled DB connection. Wait 1 minutes and retry."
```
chengren311: 1 minute

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
+            + " No.Attempt = " + retryAttempt, ex);
+        /**
+         * When database is completed down, DB connection fails to be fetched immediately. So we need
+         * to hang 1 minute for retry.
+         */
+        sleep(1000L * 60);
```
chengren311: for last retry should we skip the hanging to save some time?

chengren311: also if it runs out of retries due to db is read-only, who is responsible for closing the connection?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: also seems impl. of this method will catch all possible SQLException, when would this method throw SQLException?

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
```diff
@@ -18,5 +18,6 @@
 
 public class AzDBUtil {
 
-  static final int MAX_DB_RETRY_COUNT = 5;
+  // A very big Integer
+  static final int MAX_DB_RETRY_COUNT = 999999;
```
chengren311: if it keeps reconnecting for too long, how should we get alerted?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +65,126 @@ public synchronized Connection getConnection() throws SQLException {
...
           return connection;
         }
       } catch (final SQLException ex) {
-        logger.error(
-            "Failed to find DB connection. waits 1 minutes and retry. No.Attempt = " + retryAttempt,
-            ex);
-      } finally {
+
+        logger.error( "Failed to find write-enabled DB connection. Wait 1 minutes and retry."
+            + " No.Attempt = " + retryAttempt, ex);
```
chengren311: should it be (retryAttempt+1) to match with No.Attempt?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: When the connection can't be established during a failover for example, other threads who are trying to get a connection are likely run into the same issue.  I don't think the overhead of synchronizing this method call is too high.  It's safer to do so when in doubt.

> given createDataSource().getConnection() is already synchronized

Where is this documented?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: Why?

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: retry attempt starts from 0. Isn't correct?

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
kunkun-tang: This is a very good concern. I haven't thought about this much. The simple and straightforward solution to report retry Attempt count. If this number is too high, we should get alerted.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: Based on the current design, we will never run out of retries due to retry maximum limit is extremely high. We can start from this, and think about what we can do if retry number reaches a threshold.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: Good catch.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
kunkun-tang: * Did you mean `createDataSource()` is synchronized as well? The benefit of this design can be easily observed in the failover test. When multiple threads are trying to getConnection during DB failover, only one thread keep retrying. It made log very clear. 

* You are likely right. Let me remove the throw Exception signature.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: No need to reinvent the wheel

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: It's a bit nit. But if log shows No.attempt = 1, i think generally people will assume it's first attempt.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: Isn't the behavior different? Here we do want the sleep to be interruptable, right?

azkaban-db/src/main/java/azkaban/db/AzDBUtil.java
HappyRay: Good discussion.
Yes, some alerts would be useful.
+ @li-afaris

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: Good point.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: I think it would be better to fix it now than later because connection leak is hard to find out later.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: why does it has to be interruptable? maybe we can talk offline.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
chengren311: createDataSource() is implemented in this PR as synchronized right? getConnection() i might miss it, but we can synchronize on createDataSource().getConnection() only. My point is synchronization on a method is enforcing an object-level lock, in case when getConnection() is hanging, all other synchronized methods(currently there is none, but we might have them in the future ) will be blocked as well.

kunkun-tang: The parent class method signature has `SQLException`. Then we should keep it. I guess `createDataSource().getConnection()` might throw Exception.

kunkun-tang: After thinking, we can remove synchronized and give a try to guarantee normal cases performance.

HappyRay: Discussed with @kunkun-tang offline. This change does introduce another network call to check if the DB is readonly each time a connection is requested. Having this method synchronized does have some performance implications. 

org.apache.commons.dbcp2.BasicDataSource's methods seem to handle synchronization already.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
HappyRay: My question is why does it have to be uninterruptable.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -58,18 +60,57 @@ public synchronized Connection getConnection() throws SQLException {
...
+        /**
+         * invalidate connection and reconstruct it later. if remote IP address is not reachable,
+         * it will get hang for a while and throw exception.
+         */
+        try {
+          invalidateConnection(connection);
+        } catch (final Exception e) {
+          logger.error( "can not invalidate connection.", e);
+        }
+        logger.error( "Failed to find write-enabled DB connection. Wait 1 minutes and retry."
```
HappyRay: wait 15 seconds to match the code?

kunkun-tang: Thanks. Let me fix this.

azkaban-db/src/main/java/azkaban/db/MySQLDataSource.java
```diff
@@ -43,12 +44,13 @@ public MySQLDataSource(final String host, final int port, final String dbName,
...
+   * We don't make the call synchronized in order to guarantee normal cases performance.
    */
   @Override
-  public synchronized Connection getConnection() throws SQLException {
+  public Connection getConnection() throws SQLException {
 
     Connection connection = null;
-    int retryAttempt = 0;
+    int retryAttempt = 1;
     while (retryAttempt < AzDBUtil.MAX_DB_RETRY_COUNT) {
```
chengren311: retryAttempt <= AzDBUtil.MAX_DB_RETRY_COUNT

kunkun-tang: Thanks for finding! I can not see much difference since we will never reach `MAX_DB_RETRY_COUNT`. IMO, it should not be a big deal

Issue #1100, opened 2017-05-18
the executor server don't auto update active stat?


kris37:
when I start the executor server and then start the web server ,the web server trhow error No active executor found! in mysql DB, I find that at the begining of starting executor,the executors table not update the column 'active',it all the time is 0!

huangfushun:
have the same problem,  it seems caused by inconsistent host/port setting.  just try update "active" to 1 can work.

prokod:
Hi, Have you figured out maybe what exactly is the root cause ? We are running Azkaban 3.32.2.
Thx

kunkun-tang:
Hi @prokod @huangfushun , do you use only one executor instance in your AZ cluster setup? In  more general use cases, we have quite a few executors. When we spin up new executors (e.g., deployment), we don't activate them right away when launching executor process. We wrote a script to deactivate old executor and activate new executor at the same time. In that way, we can safely retire old executor and switch on to new executor.

prokod:
Hi @kunkun-tang In our vase we use two executors. I haven't though about the possibility to update active state through script prior to azkaban-web-server start. So in case that an executor is down from any reason it is being restarted and following it you run a script that updates state ? If so can you share this script ? Thx.

prokod:
BTW, I looked at the [documentation for executor setup](http://azkaban.github.io/azkaban/docs/latest/#executor-setup). It is written 
> Configure executors for Multiple Executor Mode
> ...
> Insert all executors into the mysql DB for executor setups
Why it is written so when in recent version the executor updates the DB when it starts up.
I suppose that this lead me to believe that if the executor behaviour was modified to register itself in the executor table it is also activating itself with startup.

kunkun-tang:
We don't update prior to web-server start.

Let me introduce how we do a regular deployment. Both old webserver and old executor(3.a) are up initially, and old executor is running jobs (we can not interrupt it). Executor Deployment starts:

1. install new Executor (`3.b`) on our box without activating.
2. Run job tests, which targets `3.b` and make sure `3.b` is working.
3. Activate `3.b`, and deactivate `3.a`. Even though `3.a` is deactivated, jobs still keep runnig there.
4. Deployment completes.

We would better run tests against new executor before activating it. Is it a bit clear now?

prokod:
Yes it is. Thank you for taking the time to explain that.

ameyamk:
[...]

PR #1415, opened 2017-08-31
Executor server doesn't auto update active stat in DB at startup fix

prokod:
To solve Executor server doesn't auto update active stat in DB at startup

#1100


prokod:
Build with tests passes on local machine. In Travis unfortunately not. Closing it

Issue #914, opened 2017-02-17
Create configuration for azkaban-web-server and azkaban-exec-server


wyukawa:
Hi

conf directory is not created when I build azkaban-web-server and azkaban-exec-server.

When I build azkaban-solo-server, conf directory is created.
But I encountered the problem which data in H2DB was broken, so I want to use azkaban-web-server and azkaban-exec-server.

Currently, there are following problems.

https://github.com/azkaban/azkaban/pull/827
https://github.com/azkaban/azkaban/issues/860

Of course, I can resolve this problems manually.

But if build.gradle is improved to resolve these problems, it would be nice.

Thanks

kunkun-tang:
Not very sure if web server and executor could use the same H2 DB file. do you know? @wyukawa

wyukawa:
Sorry, I'm not sure.

If I understand correctly,  web server and executor use the same MySQL DB.
Is that right?
I don't talk about azkaban solo mode, H2DB.

My request is to simplify to deploy azkaban(two server mode).

Thanks.

kunkun-tang:
@wyukawa I usually create a separate folder to put my conf files(mysql information), and specify the folder path as argument when web server process starts. In this case, this change should not help much. I don't know how many people could get benefit by your idea.

wyukawa:
Now, I do the following steps to deploy Azkaban after I build source codes and setup mysql.

```
mkdir azkaban-web-server-${version}/conf
vim azkaban-web-server-${version}/conf/azkaban.properties
vim azkaban-web-server-${version}/conf/azkaban-users.xml
vim azkaban-web-server-${version}/conf/log4j.properties
azkaban-web-server-${version}/bin/azkaban-web-start.sh
```

```
mkdir azkaban-exec-server-${version}/conf
vim azkaban-web-server-${version}/conf/azkaban.properties
vim azkaban-web-server-${version}/conf/log4j.properties
mkdir azkaban-exec-server-${version}/plugins/jobtypes/
vim azkaban-exec-server-${version}/plugins/jobtypes/commonprivate.properties
azkaban-exec-server-${version}/bin/azkaban-executor-start.sh
```

If this steps are simplified, it would be nice.

sunghyuk:
May I submit a PR for this?  
I want to just edit properties files after building the package.

HappyRay:
@sunghyuk  Thanks.

These packages are designed for production use. I think it makes more sense for the configs to be managed separated, possibly in a separate Git repo. 

[...]

Issue #860, opened 2016-12-27
newest code, web-server StackOverflowError


ld00000:
RUN bin/azkaban-web-start.sh    error

log:
```
Error: HADOOP_HOME is not set. Hadoop job types will not run properly.
bin/..
[...]

PR #1421, opened 2017-09-04
exit with error msg if log4j.properties file not exist.

sunghyuk:
web-server and executor-server redirected out and err stream to logstream therefore log4j.properties is mandatory.
it differs from solo-server.

```
public static void main(final String[] args) throws Exception {
  // Redirect all std out and err messages into log4j
  StdOutErrRedirect.redirectOutAndErrToLog();
```

so if log4j.properties file not exist, startup script should exit with file missing error and it more helpful than stackoverflow error. 

#860
#914


azkaban-exec-server/src/main/bash/azkaban-executor-start.sh
```diff
@@ -50,9 +50,12 @@ if [[ -z "$AZKABAN_OPTS" ]]; then
 fi
 # Set the log4j configuration file
 if [ -f $conf/log4j.properties ]; then
-  AZKABAN_OPTS="$AZKABAN_OPTS -Dlog4j.configuration=file:$conf/log4j.properties"
+  AZKABAN_OPTS="$AZKABAN_OPTS -Dlog4j.configuration=file:$conf/log4j.properties -Dlog4j.log.dir=$azkaban_dir/logs"
+else
+  echo "Exit with error: $conf/log4j.properties file must be exist."
```
HappyRay: " file doesn't exist"?

HappyRay:
LGTM otherwise.

azkaban-exec-server/src/main/bash/azkaban-executor-start.sh
sunghyuk: sorry, I'm not native. fix please

HappyRay: No problem.  Could you update your PR with the suggested change?

Issue #1130, opened 2017-05-29
Running azkaban-solo-server from in development mode


juhoautio:
Is there a standard way to quickly run azkaban-solo-server locally? Maybe even directly from IDEA?

I expect that it would be possible to run `azkaban.soloserver.AzkabanSingleServer`, maybe even without any args, and it would just work.

Do you see a way to have this kind of setup available in git so that solo server could run "out of the box" in IDEA?

What about `.less` files – could they be automatically compiled (if modified) to `.css` when running the server while developing?

----

I've used this kinda script "build-and-restart-azkaban.sh":

```sh
#!/bin/bash
if [ -f azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0/bin/azkaban-solo-shutdown.sh ]; then
    azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0/bin/azkaban-solo-shutdown.sh
fi
set -e
rm -rf azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0
./gradlew distTar
mv $(ls -t azkaban-solo-server/build/distributions/azkaban-solo-server-*gz | head -1) azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0.tar.gz
tar -xzf azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0.tar.gz -C azkaban-solo-server/build/distributions/
mv $(ls -dt azkaban-solo-server/build/distributions/*/ | head -1) azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0
cd azkaban-solo-server/build/distributions/azkaban-solo-server-3.0.0/
# fix joda-time version conflict for aws s3 library
rm lib/joda-time-2.0.jar
cp ~/.m2/repository/joda-time/joda-time/2.9.1/joda-time-2.9.1.jar lib/
bin/azkaban-solo-start.sh
```

juhoautio:
@HappyRay shared this with me, but it wasn't yet enough for me to run the solo-server successfully:

> I can start solo server in Intellij.

<img width="1027" alt="199ed2d6-4205-11e7-9114-dedffa91513e" src="https://cloud.githubusercontent.com/assets/4446608/26550691/9a9ceb96-4487-11e7-8d41-0f6e69453c2c.png">

> Here is my property file
`azkaban/temp/conf/azkaban.properties`:

```sh
# Azkaban Personalization Settings
azkaban.name=Local
[...]

PR #1423, opened 2017-09-04
Run single server from IDEA

juhoautio:
This way the server runs "out of the box". User doesn't have to configure anything, just launch AzkabanSingleServer with "Run as Java..".

On the first run `test/local-conf-templates/conf/azkaban.properties` is copied to `local/conf` (a git-ignored location).

When running the `AzkabanSingleServer`, working dir must be the `azkaban` project root
- At least for me this is the default in IntelliJ IDEA, so indeed it works out of the box
- So working dir can _not_ be for example `azkaban/azkaban-solo-server/` or `azkaban/temp/`

----

This is according to the plan that was discussed with @HappyRay in https://github.com/azkaban/azkaban/issues/1130#issuecomment-306006139.

Since that discussion I made one more conf fix:
- Set `azkaban.jobtype.plugin.dir=azkaban-solo-server/src/main/resources`
- That folder includes a `commonprivate.properties` that sets `execute.as.user=false`
- Otherwise `ProcessJob` jobs would fail with a complaint that `azkaban.native.lib` must be defined


.gitignore
```diff
@@ -13,8 +13,13 @@ Gemfile.lock
 vendor/
 **/velocity.log
 
-# temp directory is ignored so that it can be used to store temporary files such as local testing configurations.
+# temp & local directories are ignored so that they can be used to store temporary files such as local testing configurations.
 temp/
+local/
```
HappyRay: Is there a way not to generate these files in the project root? Put them into local directory instead? 

This is one of the reasons I put all my custom config files in the temp directory.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
```diff
@@ -47,9 +49,21 @@ public AzkabanSingleServer(final AzkabanWebServer webServer,
     this.executor = executor;
   }
 
-  public static void main(final String[] args) throws Exception {
+  public static void main(String[] args) throws Exception {
     log.info("Starting Azkaban Server");
 
+    // to enable "run out of the box for testing"
+    if (args.length == 0) {
+      final File templateFolder = new File("test/local-conf-templates");
```
HappyRay: Put this logic in a separate method?

test/local-conf-templates/conf/azkaban.properties
```diff
@@ -0,0 +1,55 @@
...
+database.sql.scripts.dir=azkaban-db/src/main/sql
+database.check.version=true
+database.auto.update.tables=true
+#mysql
+database.type=mysql
+mysql.port=3306
+mysql.host=localhost
+mysql.database=test
+mysql.numconnections=100
+mysql.user=
```
HappyRay: Users will still need to configure mysql DB. right?
Have you considered using H2 instead?

kunkun-tang: +1 on h2. I would believe `h2` is more suitable in this case.

test/local-conf-templates/conf/azkaban.properties
```diff
@@ -0,0 +1,55 @@
...
+# Azkaban Jetty server properties. Ignored in tomcat
+jetty.use.ssl=false
+jetty.ssl.port=8043
+jetty.maxThreads=25
+jetty.port=8081
+# Azkaban Executor settings
+executor.port=12321
+executor.maxThreads=50
+executor.flow.threads=30
+# mail settings
```
kunkun-tang: Now that we don't specify and use mail configs in solo server, should we just remove the 4 lines of email parameters.

test/local-conf-templates/conf/azkaban.properties
```diff
@@ -0,0 +1,55 @@
...
+# Velocity dev mode
+#velocity.dev.mode=false
+velocity.dev.mode=true
+# Azkaban Jetty server properties. Ignored in tomcat
+jetty.use.ssl=false
+jetty.ssl.port=8043
+jetty.maxThreads=25
+jetty.port=8081
+# Azkaban Executor settings
+executor.port=12321
```
kunkun-tang: Today executor will generate a random port, right? How this line can help?

.gitignore
juhoautio: That would be much better. The only problem is that user will have to set the working directory to `temp/` before running the server :( But hey, maybe there's a way to somehow prefix all file paths instead of writing to the root of the current working dir. I'll explore that possibility.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
juhoautio: Sure

test/local-conf-templates/conf/azkaban.properties
juhoautio: I think both should be supported so that it's easy to switch. What's important for me is that I can restart the server without clearing the database, but I guess that's possible with h2 as well.

test/local-conf-templates/conf/azkaban.properties
juhoautio: Having a fixed port makes it easier to create for example scripts that are executed against the "AJAX API" of a locally running Azkaban?

test/local-conf-templates/conf/azkaban.properties
juhoautio: What does it mean that "we don't specify and use mail configs in solo server"? I would find it handy to be able to test the mail settings for testing as well..

juhoautio:
This is the file tree it now creates when run:

nayttokuva 2017-09-07 kello 11 45 26

juhoautio:
Ready from my perspective. What do you say?

test/local-conf-templates/conf/azkaban.properties
HappyRay: 👍

test/local-conf-templates/conf/azkaban.properties
HappyRay: 👍

test/local-conf-templates/conf/azkaban.properties
HappyRay: Yes. H2 can be configured to be persistent.
And users can customize the config to use mysql.

HappyRay:
Thanks!

PR #1424, opened 2017-09-05
Remove the unused ExecutorManagerServlet class

HappyRay:


reallocf:
Do we know for certain that this class isn't consumed by any plugin?

HappyRay:
Thanks for asking.

I've searched the oss and internal plugin repos.

PR #1425, opened 2017-09-05
Guicify ExecutorApiClient

juhoautio:
Guicify ExecutorApiClient

- Create a singleton of ExecutorApiClient with Guice
- Mock it with Mockito in unit tests

----

I need this because I want to create more test cases in ExecutorManagerTest.java. For example this kind of real-life scenario:
1. web submits an execution to executor, executor accepts it
1. executor is restarted
1. web asks executor for the execution, executor returns an error "Flow does not exist"
1. web marks the execution as FAILED

So with this change I will be able to mock the executor responses.


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -82,6 +83,7 @@ protected void configure() {
     bind(DataSource.class).to(AzkabanDataSource.class);
     bind(ExecutorLoader.class).to(JdbcExecutorLoader.class);
     bind(MetricRegistry.class).in(Scopes.SINGLETON);
+    bind(ExecutorApiClient.class).in(Scopes.SINGLETON);
```
HappyRay: Please use Singleton annotation and remove this explicit binding.
We plan to remove existing such usage pattern.

kunkun-tang: You might want to refer to #1428 .

juhoautio:
Now using the `@Singleton` annotation.

PR #1426, opened 2017-09-05
Test for an interrupted execution (Flow does not exist)

juhoautio:
Test for an interrupted execution (Flow does not exist).

https://github.com/azkaban/azkaban/pull/1436 should be merged first.

So this PR is essentially just for the 2nd commit (and _some_ commits after that).


azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -255,20 +248,44 @@ public void testDuplicateQueuedFlows() throws ExecutorManagerException,
...
+    testSetUpForRunningFlows();
+    final ExecutableFlow flow1 = TestUtils.createExecutableFlow("exectest1", "exec1");
+    when(this.loader.fetchExecutableFlow(-1)).thenReturn(flow1);
+    when(this.apiClient.httpGet(any(URI.class), any())).thenAnswer(invocation -> {
+      final Map<String, String> params = paramsToMap(invocation.getArgumentAt(0, URI.class));
+      System.out.println("params = " + params);
+      if (ConnectorParams.UPDATE_ACTION.equals(params.get("action"))) {
+        // TODO too low level? how about extracting some further code into a "gateway"?
+        // then this could be done with ImmutableMap.of(..)
+        return "{\"updated\":[{\"executionId\":-1, \"error\": \"Flow does not exist\"}]}";
```
juhoautio: Yes, it's too low level. Here, improvement to be done first: https://github.com/azkaban/azkaban/pull/1430.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -264,24 +277,17 @@ public void testKillQueuedFlow() throws ExecutorManagerException, IOException {
...
-        // action=execute
-        return "{}";
-      }
-    });
+    doReturn(ImmutableMap.of(ConnectorParams.RESPONSE_UPDATED_FLOWS,
+        Arrays.asList(ImmutableMap.of(
+            ConnectorParams.UPDATE_MAP_EXEC_ID, -1,
+            "error", "Flow does not exist")))).when(this.apiGateway)
+        .callExecutorServer(anyString(), anyInt(), eq(ConnectorParams.UPDATE_ACTION),
+            any(), any(), any(), any());
```
juhoautio: Indeed it would be much nicer if instead of this we'd just have something like:

```java
.when(this.apiGateway).update(any(), any(), any());
```

juhoautio:
Replaced by https://github.com/azkaban/azkaban/pull/1438

PR #1427, opened 2017-09-05
Remove unused constants in the ExecutorManagerAdapter class

HappyRay:


HappyRay:
They were used in the removed class 
see #1424

HappyRay:
Thanks! @reallocf

HappyRay:
@reallocf Good catch. Yes. This is another reason that we should consider combining these two repos to get better compile time error checking.

PR #1428, opened 2017-09-06
Enforce using Singleton Annotation rather than binding statement

kunkun-tang:
There still exists a few places where we bind class to singleton in guice module class. As discussed a few weeks ago, we decided to enforce using `@Singleton` annotation, rather than the binding statement. The benefit is to manage and view singleton class easily.

This code patch tries to fix/replace this issue.


kunkun-tang:
I will replace all `com.google.Inject.Inject` by `com.javax.Inject` in the next PR.

PR #1429, opened 2017-09-06
fix sql concat error

sunghyuk:
upgrade.*.sql not excluded. (duplicated column error occurred if run create-all-sql script.  )


sunghyuk:
sorry for spam. typo in branch name.

HappyRay:
How is upgrade*.sql supposed to be used? 
When there is a DB schema change does the change need to go to two places?

PR #1430, opened 2017-09-06
Create ExecutorApiGateway

juhoautio:
This builds on https://github.com/azkaban/azkaban/pull/1425 so assume the first 2 commits as the baseline.

----

Create ExecutorApiGateway
- Extract the API calling code from ExecutorManager to a new class ExecutorApiGateway
- This way ExecutorManager doesn't handle raw JSON, the gateway always returns some higher level object (once again, this will allow making cleaner unit tests for ExecutorManager).


juhoautio:
What about taking this a step further and creating individual methods for different actions? For example instead of taking `ConnectorParams.UPDATE_ACTION` as an argument there would be a specific method `ExecutorApiGateway#update`?

juhoautio:
It doesn't work so well when building on commits that are then squashed.. Creating a new PR.

PR #1431, opened 2017-09-06
Remove LongArgJob based job types

HappyRay:
Remove Python, Ruby and Script job types and the LongArgJob class.

These job types are not documented and not maintained. e.g. they don't
support execute-as-user security feature which is a security hole.

Command job type can be used to provide the equivalent functionality.


HappyRay:
I have searched internal user code and have not found references to these job types.

PR #1432, opened 2017-09-06
Test refactor: just one prepareExecDir()

juhoautio:
As requested by @HappyRay in https://github.com/azkaban/azkaban/pull/1244#issuecomment-326722462.


azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestUtil.java
```diff
@@ -16,18 +16,20 @@
...
+import azkaban.utils.JSONUtils;
 import azkaban.utils.Props;
 import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 import org.apache.commons.io.FileUtils;
 import org.apache.log4j.Logger;
 
-class FlowRunnerTestUtil {
+public class FlowRunnerTestUtil {
```
HappyRay: Why isn't package private enough?

HappyRay:
Thanks for the PR!

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTestUtil.java
juhoautio: Because `RemoteFlowWatcherTest` is in `azkaban.execapp.event`. I tried, it can't see this class if it's package visible.

PR #1435, opened 2017-09-06
Add coding conventions

HappyRay:


HappyRay:
Thanks @juhoautio! 😄

chengren311:
I'm wondering what the convention of static methods are? Should they all be put into an util class?

HappyRay:
> I'm wondering what the convention of static methods are? Should they all be put into an util class?

The "favor small classes " convention applies too. 
There are cases where it's better to keep static methods in the same class e.g. to implement a factory pattern. 

I would also like to keep the list of coding conventions as small as possible to help people remember.

CONTRIBUTING.md
```diff
@@ -59,3 +61,46 @@ It allows us to do following when saving a file:
 * Remove unnecessary semicolon
 * Add missing @Override annotations
 
+## Coding conventions
```
kunkun-tang: Do we name conventions?

HappyRay: It's covered in the referenced Google style guide.

HappyRay:
> In this doc, do we need to state that Azkaban build doesn't allow warning. If we have to bring code with warning, we must suppress it.

Do you mean errorprone warnings? @reallocf has made the change to enforce the no warning rule. So I don't think it is necessary to mention it here.

reallocf:
I'd rather the standard be that we fix warnings instead of suppressing them :wink:

We can mention it if we want to. But it should definitely say that any suppressed warnings must be justified somewhere - in the commit message at a minimum and ideally in a comment within the actual code.

PR #1436, opened 2017-09-06
Refactor only: create ExecutorApiGateway

juhoautio:
- Extract the API calling code from ExecutorManager to a new class ExecutorApiGateway
- This way ExecutorManager doesn't handle raw JSON, the gateway always returns some higher level object (once again, this will allow making cleaner unit tests for ExecutorManager).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
@@ -38,65 +38,56 @@ public class ExecutorApiGateway {
     this.apiClient = apiClient;
   }
 
-  Map<String, Object> callExecutorServer(final ExecutableFlow exflow,
+  Map<String, Object> callWithExecutable(final ExecutableFlow exflow,
       final Executor executor, final String action) throws ExecutorManagerException {
-    try {
-      return callExecutorServer(executor.getHost(), executor.getPort(), action,
-          exflow.getExecutionId(), null, (Pair<String, String>[]) null);
-    } catch (final IOException e) {
-      throw new ExecutorManagerException(e);
-    }
+    return callWithExecutionId(executor.getHost(), executor.getPort(), action,
+        exflow.getExecutionId(), null, (Pair<String, String>[]) null);
   }
 
-  Map<String, Object> callExecutorServer(final ExecutionReference ref,
-      final String action, final Pair<String, String>... params)
-      throws ExecutorManagerException {
-    try {
-      return callExecutorServer(ref.getHost(), ref.getPort(), action,
-          ref.getExecId(), null, params);
-    } catch (final IOException e) {
-      throw new ExecutorManagerException(e);
-    }
+  Map<String, Object> callWithReference(final ExecutionReference ref, final String action,
+      final Pair<String, String>... params) throws ExecutorManagerException {
+    return callWithExecutionId(ref.getHost(), ref.getPort(), action, ref.getExecId(),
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
```
HappyRay: 2017

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -0,0 +1,139 @@
...
+public class ExecutorApiGateway {
+
+  private final ExecutorApiClient apiClient;
+
+  @Inject
+  public ExecutorApiGateway(final ExecutorApiClient apiClient) {
+    this.apiClient = apiClient;
+  }
+
+  Map<String, Object> callExecutorServer(final ExecutableFlow exflow,
```
HappyRay: Is there any reason not to give different names to these methods?

juhoautio: > Is there any reason not to give different names to these methods?

No, that would be really handy actually. I tried to keep the diff minimal for now though.

HappyRay: Make sense. Thanks! Do you mind making this improvement in a later PR?

It would be helpful for reviewers if you can mention that this code is just an exact copy of the original. It's hard to compare when the code is moved to another file.

juhoautio:
Resolved @HappyRay's comments.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
HappyRay: FYI I set up the copyright notice in intellij and it will auto update.
I plan to share how I did it in a wiki when I find time to do so.

HappyRay:
Thanks for making the suggested changes.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -0,0 +1,130 @@
...
+        paramList.addAll(Arrays.asList(params));
+      }
+
+      paramList
+          .add(new Pair<>(ConnectorParams.ACTION_PARAM, action));
+      paramList.add(new Pair<>(ConnectorParams.EXECID_PARAM, String
+          .valueOf(executionId)));
+      paramList.add(new Pair<>(ConnectorParams.USER_PARAM, user));
+
+      final Map<String, Object> jsonResponse =
```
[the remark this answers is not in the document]
juhoautio: > why not just return it directly?

Yeah, this was just moved from the existing code. But I'll change it.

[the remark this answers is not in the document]
HappyRay: IMHO, having a local variable is fine. Having a local variable makes it easier to debug. At least that is my experience.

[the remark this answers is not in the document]
HappyRay: I prefer keeping the changes to the minimum when moving code. Make improvement in separate PRs.

juhoautio:
@kunkun-tang: resolved your comment

Issue #1127, opened 2017-05-27
Automatic resuming of flows after azkaban-executor restart


juhoautio:
This is a new feature suggestion. Feedback much appreciated!

## Rationale

We need to support resuming flows and jobs after azkaban-executor is restarted. Currently Azkaban just marks interrupted flows and jobs as FAILED on startup. This is actually done silently without sending flow failure notifications or anything like that.

Reasons for "restarting" azkaban-executor include at least:
- Azkaban-executor java process crashed (OOM, for example)
- Azkaban version update requires a restart to apply changes (especially in a single executor or solo-azkaban setup)
- Hardware failure or some other problem requires replacing the azkaban-executor with an entirely new instance

## Design

Actions on startup for interrupted flows:
- Start execution for the flow again with the _same execution id and project version_
  - For example flow log should have all the history before the flow was interrupted + note that execution was resumed
- Jobs that support resume are started again with `JobState.isResume=true`
- Jobs that don't support resume are started over
  - So jobs that don't support resume get "1 extra retry" by default
  - This can be disabled by setting `resume.retryJobIfResumeNotImplemented=false`
- Jobs are resumed without incrementing the retry attempt count – they should have the attempt number that was running when flow was interrupted
- Add new flow level email notification option `resume.emails` (similar to the existing `failure.emails`)

`Job` API change:
- New optional method: `boolean Job#supportsResume`, default: `false`
- New optional method: `void Job#setState(JobState)` (this is not called unless `supportsResume` returns `true`)

New interface `JobState` with methods:
- `void JobState#setState(Props)`
  - This persists the given state `Props` into DB (key: flow/execution/job/attempt)
- `Props JobState#getState`
- `boolean JobState#isResume`
  - Jobs could basically get the same result by checking `JobState.getState() != null` in the beginning of the run, but I think that having a specific method is useful.
  - Naturally `isResume` will keep returning `false` for a regular (not resumed) execution even if the job saves something to the `JobState`.

This way the job can store some execution state that is fetched back for it before resuming.

## Notes about our use cases

Usually we just need to store an external job id after it has been launched, and when resuming we continue polling for the external status by id.

Resuming can also be useful for "locally running jobs" if they know how to continue from whatever state they have saved so far.

In practice we for example launch AWS EMR clusters/jobs externally with a custom job type for that purpose. We have some other custom job types as well that launch a job externally and then poll for that to complete.

[...]

PR #1438, opened 2017-09-07
Faster updater thread and test for an interrupted execution

juhoautio:
- Test for an interrupted execution ("Flow does not exist")
  - This is groundwork for _Automatic resuming of flows after azkaban-executor restart_ https://github.com/azkaban/azkaban/issues/1127)
- Thanks to this, found out that wait/notify was missing for dispatching new executions
  - Fixed that, which helps the new test `testNotFoundFlows` finish in ~100ms instead of >2000ms

edit: it's not dispatching that's faster, it's just the flow updater thread, but still.. I guess that has significance only on unit tests. In real life flows don't finish _that_ quickly, so usually there's always some sleeping rounds involved before updater thread will be able to do something about any execution.


HappyRay:
> Thanks to this, found out that wait/notify was missing for dispatching new executions
Fixed that, which helps the new test testNotFoundFlows finish in ~100ms instead of >2000ms
👍

juhoautio:
That test failure is unrelated to these changes, and it will be fixed by this other PR: https://github.com/azkaban/azkaban/pull/1440

juhoautio:
Could someone review this, please?

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -349,4 +386,18 @@ private void testSetUpForRunningFlows()
...
   }
+
+  private ExecutableFlow waitFlowFinished(final ExecutableFlow flow1) throws Exception {
+    ExecutableFlow fetchedFlow;
+    for (int i = 0; i < 1000; i++) {
+      fetchedFlow = this.loader.fetchExecutableFlow(flow1.getExecutionId());
+      if (fetchedFlow != null && Status.isStatusFinished(fetchedFlow.getStatus())) {
+        return fetchedFlow;
+      }
+      Thread.sleep(10L);
```
juhoautio: Actually now that this doesn't use any `notify` mechanism, can be replaced with the use of `Awaitility`. So blocked by https://github.com/azkaban/azkaban/pull/1442 to add that dependency.

juhoautio:
Hit this kind of hamcrest dependency conflict, so fixed it by switching to `mockito-core`: https://stackoverflow.com/questions/7869711/getting-nosuchmethoderror-org-hamcrest-matcher-describemismatch-when-running

juhoautio:
Now using Awaitility. In case of failure it prints message like:

      expected a finished status but was <READY> within 10 seconds.

azkaban-common/src/test/java/azkaban/utils/LambdaMatcher.java
```diff
@@ -0,0 +1,39 @@
+package azkaban.utils;
+
+import java.util.function.Function;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+
+public class LambdaMatcher<T> extends BaseMatcher<T> {
```
HappyRay: Please add javadoc for public interfaces such as classes per the style guide.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -26,6 +27,10 @@
 import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+import org.awaitility.Awaitility;
+import org.awaitility.core.ConditionFactory;
+import org.hamcrest.Matcher;
```
HappyRay: Does assertj provide similar functionalities? We favor assertj over hamcrest per the dev guide.

HappyRay: https://blog.jayway.com/2014/04/23/java-8-and-assertj-support-in-awaitility-1-6-0/

https://solidsoft.wordpress.com/2014/04/29/using-assertj-and-awaitility-together-thanks-to-java-8-and-lambdas/

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1385,6 +1385,9 @@ private void dispatch(final ExecutionReference reference, final ExecutableFlow e
     // move from flow to running flows
     this.runningFlows.put(exflow.getExecutionId(),
         new Pair<>(reference, exflow));
+    synchronized (this) {
```
HappyRay: Could you add a comment to explain why notifyAll is needed? It may not be obvious to readers.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -65,4 +70,17 @@ public static UserManager createTestXmlUserManager() {
     final UserManager manager = new XmlUserManager(props);
     return manager;
   }
+
```
HappyRay: Does it make sense to move these common test methods to the test module so that they can be used in all other modules more easily? e.g. the new az-core module can't depend on the common module.

build.gradle
```diff
@@ -99,7 +99,7 @@ ext.deps = [
         math3               : 'org.apache.commons:commons-math3:3.0',
         metricsCore         : 'io.dropwizard.metrics:metrics-core:3.1.0',
         metricsJvm          : 'io.dropwizard.metrics:metrics-jvm:3.1.0',
-        mockito             : 'org.mockito:mockito-all:1.10.19',
+        mockito             : 'org.mockito:mockito-core:1.10.19',
```
HappyRay: Is it done so that mockito doesn't bring its own version of Hamcrest which causes conflict with the intellij's builtin version?

HappyRay: Found this https://solidsoft.wordpress.com/2012/09/11/beyond-the-mockito-refcard-part-3-mockito-core-vs-mockito-all-in-mavengradle-based-projects/

Could you add a comment here so that people don't accidentally change back to -all?

HappyRay: Also https://github.com/mockito/mockito/issues/324
"mockito-all is deprecated and is no longer published starting with version 2.0.3-beta to not confuse users."

Looks like we can update the mockito itself. The latest is
https://mvnrepository.com/artifact/org.mockito/mockito-core/2.10.0
If you update the version, then no comment is necessary since the -all version is no longer published.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
juhoautio: I don't understand how to make that work. This doesn't compile:

```java
    Awaitility.await().until(() ->
        org.assertj.core.api.Assertions.assertThat(getFlowStatus(flow))
            .matches(Status::isStatusFinished)
    );
```

Also tried to take the example from the second [link](https://blog.jayway.com/2014/04/23/java-8-and-assertj-support-in-awaitility-1-6-0/) but it doesn't compile either:

```java
    Awaitility.await().until( () -> org.assertj.core.api.Assertions.assertThat(1).isEqualTo(1) );
```

Unless you know how to use this, I propose we go with the "standard" awaitility+hamcrest combo.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
juhoautio: Yes, why not.. Maybe there could be a separate pull request after _this_ to move `TestUtils` and the new `LambdaMatcher` into `test/`?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Added that comment but it starts on the next line so this comment remains to be shown.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
HappyRay: Below example works for me.

```java
  @Test
  public void test_wait() {
    Awaitility.await().untilAsserted(() -> assertThat(1).isEqualTo(1));
  }
```
See
https://github.com/awaitility/awaitility/wiki/Usage#version-3x-and-above

juhoautio: Hmm, I just came across this, let me try it:
https://github.com/awaitility/awaitility/wiki/Usage#version-3x-and-above-1

juhoautio: Yeah, this works:

```java
    TestUtils.await().untilAsserted(() -> assertThat(getFlowStatus(flow))
        .matches(Status::isStatusFinished, "a finished status"));
```

juhoautio:
Ok, comments resolved? Reverted adding `LambdaMatcher.java` and using `Awaitility.await().untilAsserted` instead \o/

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
HappyRay: That would be good.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -317,8 +316,7 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
     Assert.assertTrue(logFile.exists());
 
     // wait so that there's time to make the "DB update" for KILLED status
-    Awaitility.await().atMost(10L, TimeUnit.SECONDS).pollInterval(10L, TimeUnit.MILLISECONDS)
-        .until(() -> loader.getNodeUpdateCount("testJob"), is(2));
+    TestUtils.await().until(() -> loader.getNodeUpdateCount("testJob"), is(2));
```
HappyRay: It would be good to use assertJ here as well. It can be done in another PR.

juhoautio: Do you know out of your head how that would be done? If not never mind I could look it up.. Thanks for the great review once again 👍

HappyRay: Tracked in the issue #1471

PR #1439, opened 2017-09-08
Reproduce flaky tests: FlowRunnerTest2 > testNormalFailure2 & FlowRunnerTest > exec1FailedKillAll

juhoautio:
The test failure is:

```
azkaban.execapp.FlowRunnerTest2 > testNormalFailure2 FAILED
    java.lang.AssertionError: Wrong status for [jobb:innerJobB] expected:<CANCELLED> but was:<RUNNING>

azkaban.execapp.FlowRunnerTest2 > testNormalFailure2 STANDARD_ERROR
    testJob: joba1 FAILED
    testJob: jobd:innerJobA SUCCEEDED
    testJob: jobd:innerFlow2 RUNNING
    testJob: joba SUCCEEDED
    testJob: jobc RUNNING
    testJob: jobb:innerJobB RUNNING
    testJob: jobb:innerJobC RUNNING
    testJob: jobb:innerJobA SUCCEEDED
    ExecutableFlow: jobf FAILED_FINISHING
    ExecutableNode: joba1 FAILED
    ExecutableFlow: jobb RUNNING
    ExecutableNode: jobb:innerFlow READY
    ExecutableNode: jobb:innerJobB RUNNING
    ExecutableNode: jobb:innerJobA SUCCEEDED
    ExecutableNode: jobb:innerJobC RUNNING
    ExecutableNode: joba SUCCEEDED
    ExecutableFlow: jobd RUNNING
    ExecutableNode: jobd:innerJobA SUCCEEDED
    ExecutableNode: jobd:innerFlow2 RUNNING
    ExecutableNode: jobc RUNNING
    ExecutableNode: jobf READY
    ExecutableNode: jobe READY
```

 Analysis: when a job finishes, an event is fired about its status. The events are handled asynchronously in FlowRunner's progressGraph() loop. This test case is flaky because before ("jobb:innerJobA").succeedJob() and ("jobd:innerJobA").succeedJob() it waits for the node status of "joba1" to be FAILED, but it doesn't wait for that event to have been processed.

Issue #1389, opened 2017-08-25
Fix flaky test


suvodeep-pyne:
This test is flaky and failing a lot of times. Example: https://travis-ci.org/azkaban/azkaban/builds/268055504

```
azkaban.execapp.JobRunnerTest > testDelayedExecutionCancelledJob FAILED
    java.lang.AssertionError: expected:<2> but was:<1>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at azkaban.execapp.JobRunnerTest.testDelayedExecutionCancelledJob(JobRunnerTest.java:313)
```

reallocf:
:pray:

HappyRay:
@juhoautio any idea?

HappyRay:
Master failed with
the following error too today
https://travis-ci.org/azkaban/azkaban/builds/268499837?utm_source=email&utm_medium=notification

azkaban.execapp.FlowRunnerTest2 > testCancelOnFailure FAILED
    java.lang.AssertionError: Wrong status for [jobb:innerJobB] expected:<RUNNING> but was:<READY>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:117)
        at azkaban.execapp.FlowRunnerTest2.testCancelOnFailure(FlowRunnerTest2.java:576)

juhoautio:
1. JobRunnerTest#testDelayedExecutionCancelledJob
   - It might help to make the sleep longer before that assertion, for example 1 s -> 2 s. Any way I'm not sure if this will help.. What I know though is that it constantly fails with a similar-looking same error if the sleep is removed altogether.
   - Better yet, that assertion could be changed to a "polling assert" with a, say, 10 s timeout.
2. FlowRunnerTest2#testCancelOnFailure, that seems really strange to me.. I can't really imagine how that could happen. Makes me wonder, could it even be a real bug, not just a flaky test..

HappyRay:
Another failure today:

azkaban.execapp.FlowRunnerTest2 > testRetryOnFailure FAILED
[...]

PR #1440, opened 2017-09-08
Fix flaky FlowRunnerTest2 > testNormalFailure2

juhoautio:
(Part of fixing https://github.com/azkaban/azkaban/issues/1389)

- Wait that FAILED job event has been handled before succeeding the other jobs.
- Extra: make InteractiveTestJob more reliable with volatile fields

(See Travis build history of https://github.com/azkaban/azkaban/pull/1439 for proof)


jamiesjc:
Thanks for the fix! I was able to reproduce the same issue on my local branch.
Here is my understanding of the issue and your fix. Correct me if I misunderstand it.
In testNormalFailure2, we explicitly fail joba1. Ideally all the remaining jobs should change status to CANCELLED. However, it doesn't always happen because we only wait for joba1 status to become FAILED and do not wait for the JOB_FINISHED event. 
This event needs to be handled because it will add joba1 to finishedNodes, and in azkaban.execapp.FlowRunner#progressGraph, it will process the finishedNodes and set flowFailed to true. Once this flag is set, then the rest of the remaining jobs will become CANCELLED status in azkaban.execapp.FlowRunner#getImpliedStatus.

juhoautio:
> However, it doesn't always happen because we only wait for joba1 status to become FAILED and do not wait for the JOB_FINISHED event.

Yes, the problem was that the test didn't wait for the JOB_FINISHED event to have been handled for the failed job _before allowing those two other jobs to succeed_. These test jobs are just waiting in RUNNING state before `InteractiveTestJob#succeedJob` or `InteractiveTestJob#failJob` is called.

Just wanted to clarify that because in your description you didn't say anything about those `InteractiveTestJob#succeedJob` calls. But yes, I'm sure you understood the issue here.

PR #1442, opened 2017-09-08
Fix flaky JobRunnerTest > testDelayedExecutionCancelledJob

juhoautio:
(Part of fixing https://github.com/azkaban/azkaban/issues/1389)

- Now instead of fixed 1s sleep we wait for max 10s for the node update count to be 2.
- This lets us use the luxury of additional sleep time (1s->2s) before killing the flow, so that there's a better chance for job getting inside its delayExecution() block.

Test failure was:

```
azkaban.execapp.JobRunnerTest > testDelayedExecutionCancelledJob FAILED
    java.lang.AssertionError: expected:<2> but was:<1>
```

It was faithfully reproduced when I removed the 1s sleep before assert on getNodeUpdateCount().

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -16,6 +16,8 @@
 
 package azkaban.execapp;
 
+import static org.hamcrest.CoreMatchers.is;
+
 import azkaban.event.Event;
 import azkaban.event.EventData;
 import azkaban.executor.ExecutableFlow;
@@ -32,8 +34,10 @@ import azkaban.utils.Props;
 import java.io.File;
 import java.io.IOException;
 import java.util.HashSet;
+import java.util.concurrent.TimeUnit;
 import org.apache.commons.io.FileUtils;
 import org.apache.log4j.Logger;
+import org.awaitility.Awaitility;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -292,7 +296,7 @@ public class JobRunnerTest {
 
     StatusTestUtils.waitForStatus(node, Status.READY);
     // sleep so that job has time to get into delayExecution() -> wait()
-    Thread.sleep(1000L);
+    Thread.sleep(2000L);
     runner.kill();
     StatusTestUtils.waitForStatus(node, Status.KILLED);
 
@@ -303,8 +307,8 @@ public class JobRunnerTest {
         node.getStatus() == Status.KILLED);
     Assert.assertTrue(node.getStartTime() > 0 && node.getEndTime() > 0);
     Assert.assertTrue(node.getEndTime() - node.getStartTime() < 1000);
-    Assert.assertTrue(node.getStartTime() - startTime >= 1000);
-    Assert.assertTrue(node.getStartTime() - startTime <= 4000);
+    Assert.assertTrue(node.getStartTime() - startTime >= 2000);
+    Assert.assertTrue(node.getStartTime() - startTime <= 5000);
     Assert.assertTrue(runner.isKilled());
 
     final File logFile = new File(runner.getLogFilePath());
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -298,20 +302,19 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
...
     final Props outputProps = runner.getNode().getOutputProps();
     Assert.assertTrue(outputProps == null);
     Assert.assertTrue(logFile.exists());
 
-    // sleep so that there's time to make the "DB update" for KILLED status
-    Thread.sleep(1000L);
-    Assert.assertEquals(2L, loader.getNodeUpdateCount("testJob").longValue());
-    Assert.assertEquals(2L, (long) loader.getNodeUpdateCount("testJob"));
+    // wait so that there's time to make the "DB update" for KILLED status
+    Awaitility.await().atMost(10L, TimeUnit.SECONDS).pollInterval(10L, TimeUnit.MILLISECONDS)
```
jamiesjc: It's a good idea to introduce Awaitility in the tests!

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -288,7 +292,7 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
 
     StatusTestUtils.waitForStatus(node, Status.READY);
     // sleep so that job has time to get into delayExecution() -> wait()
-    Thread.sleep(1000L);
+    Thread.sleep(2000L);
```
jamiesjc: Is one second not long enough for the job to get into delayExecution()? I've never seen this test case failing in my local branch, so I'm wondering if this sleep time needs to be longer?

juhoautio: Change at your own risk, but on Travis it's a whole different world. There are unexpected delays and in many places 1 second sleep hasn't been enough. Not saying this particular line would be as sensitive to random delays, but just trying to play it safe.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
juhoautio: Kudos to @HappyRay for the suggestion.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -298,20 +302,19 @@ public void testDelayedExecutionCancelledJob() throws InterruptedException {
     Assert.assertTrue("Node status is " + node.getStatus(),
         node.getStatus() == Status.KILLED);
     Assert.assertTrue(node.getStartTime() > 0 && node.getEndTime() > 0);
-    Assert.assertTrue(node.getEndTime() - node.getStartTime() < 1000);
-    Assert.assertTrue(node.getStartTime() - startTime >= 1000);
-    Assert.assertTrue(node.getStartTime() - startTime <= 4000);
+    Assert.assertTrue(node.getEndTime() - node.getStartTime() < 2000);
```
jamiesjc: After increasing the sleep time to 2s, I can understand that (node.getStartTime - startTime) is changed, but why is (node.getEndTime - node.getStartTime) also affected?

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
HappyRay: Glad to know the library turns out to be useful. :)

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
juhoautio: Honestly I haven't given it any deeper thought. I'm merely changing it back to [how it used to be](https://github.com/azkaban/azkaban/commit/eb6bdd5e4bb25e1effcbe5333fc7768ec79b7274#diff-e0cbe4296eb29a4206100f4f2fbc2c28). Now that we have the Awaitility in place, having 2s instead of 1s is afforable.

jamiesjc: I agree with changing 1s to 2s in the sleep time. But the check in line 305 seems to be independent of this change. The job is configured to run for 1 sec in `createJobRunner(1, "testJob", 1, false, loader, eventCollector);`, so I think the original intention was to check whether the job finished within 1 sec. Also it seems it used to be `Assert.assertTrue(node.getEndTime() - node.getStartTime() < 1000);` from the link you shared?

juhoautio: Oh sorry, I can see I made a mistake there. Changing back to 1000.

juhoautio:
Resolved conflicts.

PR #1443, opened 2017-09-08
Flow.2.0_part1_loadAzkabanFlowFromYamlFile[In progress]

jamiesjc:
In flow.2.0 design, each flow is defined in a single yaml file. 
When uploading project, we need to convert the flow yaml file to azkabanFlow object and integrate it in the core Azkaban code.

The integration will happen in below steps:
1. Flow UI: load flow/job properties, resolve dependencies and construct flows from azkabanFlow. Ensure the flow DAG is showing properly from the UI after uploading the project, also flow properties are stored correctly in DB. 
2. Flow execution: Integrate azkabanFlow in executableFlow, load properties from azkabanFlow directly for flowRunner. Ensure flow execution is successful.

Step 1 is done in the PR. Verified the changes on solo server. 

Related PR to convert yaml file: https://github.com/azkaban/azkaban/pull/1275


azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
```diff
@@ -264,49 +330,58 @@ private void resolveDependencies() {
...
+  }
+
+  private void resolveDependenciesFromAzkabanFlow() {
+    for (final Node node : this.nodeMap.values()) {
+      resolveDependenciesFromList(
+          ((AzkabanJob)node.getAzkabanNode()).getDependsOn(), node);
+    }
+  }
+
+  private void resolveDependenciesFromList(List<String> dependencyList, Node node) {
```
jamiesjc: This part of the code is the same. Just moved it to a separate method.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -79,6 +82,14 @@ public static ExecutableFlow createExecutableFlowFromObject(final Object obj) {
     return exFlow;
   }
 
+  public AzkabanFlow getAzkabanFlow() {
+    return azkabanFlow;
+  }
+
+  public void setAzkabanFlow(AzkabanFlow azkabanFlow) {
```
chengren311: should AzkabanFlow be initialized from constructor?

azkaban-common/src/main/java/azkaban/flow/Flow.java
```diff
@@ -133,6 +137,14 @@ public static Flow flowFromObject(final Object object) {
     return properties;
   }
 
+  public AzkabanFlow getAzkabanFlow() {
+    return azkabanFlow;
+  }
+
+  public void setAzkabanFlow(AzkabanFlow azkabanFlow) {
```
chengren311: same as above

azkaban-common/src/main/java/azkaban/flow/Node.java
```diff
@@ -157,6 +160,14 @@ public void setEmbeddedFlowId(final String flowId) {
     this.embeddedFlowId = flowId;
   }
 
+  public AzkabanNode getAzkabanNode() {
+    return azkabanNode;
+  }
+
+  public void setAzkabanNode(AzkabanNode azkabanNode) {
```
chengren311: same as above

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -121,11 +122,16 @@
     // synchronization between uploads. Since we're already reloading the XML
     // config file and creating validator objects for each upload, this does
     // not add too much additional overhead.
+
+    // Todo jamiesjc: Check if MANIFEST.MF file exists in the zip,
+    // if yes, treat it as flow 2.0, otherwise flow 1.0
+    // String azkabanFlowVersion = Constants.AZKABAN_FLOW_VERSION_2_0;
+    String azkabanFlowVersion = null;
```
chengren311: debugging code?

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
```diff
@@ -135,14 +138,25 @@ public void loadProjectFlow(final Project project, final File baseDirectory) {
...
+    else if(azkabanFlowVersion.equals(Constants.AZKABAN_FLOW_VERSION_2_0)) {
+      // For azkaban.flow.2.0, yaml files will be used to describe flow info.
+      // First convert yaml files to azkabanFlows, and then load project from
+      // azkabanFlows. This is the first step to integrate azkabanFlows in the
+      // core azkaban code.
+      for (AzkabanFlow azkabanFlow : convertYamlFileToAzkabanFlows(baseDirectory)) {
+        loadProjectFromAzkabanFlow(azkabanFlow);
+      }
+      resolveDependenciesFromAzkabanFlow();
+    }
```
chengren311: better to fail here if it falls into else block

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
```diff
@@ -135,14 +138,25 @@ public void loadProjectFlow(final Project project, final File baseDirectory) {
...
+    else if(azkabanFlowVersion.equals(Constants.AZKABAN_FLOW_VERSION_2_0)) {
+      // For azkaban.flow.2.0, yaml files will be used to describe flow info.
+      // First convert yaml files to azkabanFlows, and then load project from
+      // azkabanFlows. This is the first step to integrate azkabanFlows in the
+      // core azkaban code.
+      for (AzkabanFlow azkabanFlow : convertYamlFileToAzkabanFlows(baseDirectory)) {
+        loadProjectFromAzkabanFlow(azkabanFlow);
+      }
+      resolveDependenciesFromAzkabanFlow();
```
chengren311: better to warn here if it hits else block

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
```diff
@@ -151,6 +165,58 @@ public void loadProjectFlow(final Project project, final File baseDirectory) {
 
   }
 
+  private ArrayList<AzkabanFlow> convertYamlFileToAzkabanFlows(File projectDir) {
+    ArrayList<AzkabanFlow> azkabanFlows = new ArrayList<>();
+    // Load all yaml files to AzkabanFlows in current project directory,
+    // Do not support loading from subdirectories for now.
+    final File[] yamlFiles = projectDir.listFiles(new SuffixFilter(YML_SUFFIX));
+    Arrays.sort(yamlFiles);
```
chengren311: why do we need sort here?

azkaban-common/src/main/java/azkaban/Constants.java
```diff
@@ -31,6 +31,8 @@
  * <p>Job level Properties keys to be put in the {@link JobProperties} class
  */
 public class Constants {
+  // Azkaban Flow Version
+  public static final String AZKABAN_FLOW_VERSION_2_0 = "azkaban.flow.2.0";
```
kunkun-tang: Should this constant name as `AZKABAN_FLOW_VERSION` rather than `AZKABAN_FLOW_VERSION_2_0 `?

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
kunkun-tang: I guess this might be necessary for migration purpose.

azkaban-common/src/main/java/azkaban/Constants.java
jamiesjc: I'm specifying the flow version here in case we have more versions in the future like AZKABAN_FLOW_VERSION_3_0.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
jamiesjc: Yes, this AzkabanFlow will only apply to flow.2.0, and we still need to keep backward compatibility.

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
jamiesjc: Agree, thanks!

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
jamiesjc: Probably not needed. I'm just following what the current code is doing for .properties files.

HappyRay:
@jamiesjc for big changes that require multiple PRs, it may be useful to create an issue and reference the issue from all the PRs. This way all the PRs can be tied together.

azkaban-common/src/main/java/azkaban/project/validator/ValidatorManager.java
```diff
@@ -25,7 +25,7 @@
    * of {@link ValidationReport} with the key being the validator's name and the value being the
    * {@link ValidationReport} generated by that validator.
    */
-  Map<String, ValidationReport> validate(Project project, File projectDir);
+  Map<String, ValidationReport> validate(Project project, File projectDir, String azkabanFlowVersion);
```
HappyRay: How do you plan to discover and update all the users of this interface?

PR #1444, opened 2017-09-09
Test stability: InteractiveTestJob default wait 5s->10s

juhoautio:
There may be some test job files that still set seconds=1, for example.


juhoautio:
Could someone look at this PR, please? Really small change :)

HappyRay:
Thanks!

PR #1445, opened 2017-09-09
Enable JobRunnerTest#testBasicRun

juhoautio:
This works, but it seems wrong that `JobRunner` test runs a java job that launches an additional JVM / java process. That should be the responsibility of `JavaProcessJobTest`, and I think it already does that.

If this test case serves any purpose, I think it should be converted to use the `InteractiveTestJob` which is simple and quick and lives within the test process itself.

Any way, the enabled test `testBasicRun` runs in 0,5-1s.

Opinions?

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
@@ -86,11 +86,13 @@ public class SleepJavaJob {
 
     final int sec = Integer.parseInt(this.seconds);
     System.out.println("Sec " + sec);
-    synchronized (this) {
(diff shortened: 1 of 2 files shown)


juhoautio:
Aha, this is just the normal case. It's worth having this test. Do you agree on changing this to use the `InteractiveTestJob` so that it can run as quick as possible?

HappyRay:
Yes. Sounds good to me.

HappyRay:
A unit test that runs close to 1s is too long. I would prefer converting first before enabling it.

juhoautio:
Replaced with https://github.com/azkaban/azkaban/pull/1469

PR #1446, opened 2017-09-09
Enable all in JobRunnerTest

juhoautio:
Enable all in `JobRunnerTest`. First commit separately in https://github.com/azkaban/azkaban/pull/1445.

This works but is a bit slow with all the sleeps in place:

nayttokuva 2017-09-09 kello 12 09 47

I think this can be made run quickly, but I wonder if these tests are useful or not :)

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
@@ -86,11 +86,13 @@ public class SleepJavaJob {
 
     final int sec = Integer.parseInt(this.seconds);
     System.out.println("Sec " + sec);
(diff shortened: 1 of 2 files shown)


juhoautio:
Now that I stopped to look, also all of these tests seem to be worth enabling. Do you agree on changing all of them to use the `InteractiveTestJob` so that it can run as quick as possible?

HappyRay:
Yes, if they can be fixed to run quickly and reliably.  Thanks.

juhoautio:
I will revisit this once https://github.com/azkaban/azkaban/pull/1469 has been dealt with.

juhoautio:
Replaced by https://github.com/azkaban/azkaban/pull/1479

PR #1450, opened 2017-09-10
Move execute-as-user c program to its own module

### LINKED ISSUE #1227: Move execute-as-user c program to its own sub project
It's not a shared component.

And the output is not packaged currently into the distribution.

see
http://azkaban.github.io/azkaban/docs/latest/

"
Execute-As-User
...
Generate an executable on the Azkaban box for azkaban-common/src/main/c/execute-as-user.c. it should be named execute-as-user Below is a sample approach
scp ./azkaban-common/src/main/c/execute-as-user.c onto the Azkaban box
run: gcc execute-as-user.c -o execute-as-user
run: chown root execute-as-user (you might need root privilege)
run: chmod 6050 execute-as-user (you might need root privilege)
"

HappyRay:
This module doesn't belong to the common module.
It also helps speeding up the build a bit by taking better advantage of
the gradle cache and project dependency management.

Fixes #1227

### DIFF
--- a/az-exec-util/build.gradle
+++ b/az-exec-util/build.gradle
@@ -13,15 +13,18 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-
-rootProject.name = 'azkaban'
-
-include 'az-exec-util'
-include 'azkaban-spi'
-include 'azkaban-db'
-include 'azkaban-common'
-include 'azkaban-exec-server'
-include 'azkaban-hadoop-security-plugin'
-include 'azkaban-solo-server'
-include 'azkaban-web-server'
-include 'test'
+apply plugin: 'c'
+model {
+    components {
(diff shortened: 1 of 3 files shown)


settings.gradle
```diff
@@ -12,11 +12,11 @@
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
  * License for the specific language governing permissions and limitations under
  * the License.
- *
  */
 
 rootProject.name = 'azkaban'
 
+include 'az-exec-util'
```
kunkun-tang: Will have more code check in this gradle sub project?

HappyRay: Maybe, there is currently no plan to do so.

HappyRay:
The build is just to make sure the c code can compile. It needs to be built on the target platform I think. See the main doc for details.

Issue #1448, opened 2017-09-10
Fix flaky Travis build: Gradle test executor error


HappyRay:
master branch builds failed today with the following:

```
Caused by: org.gradle.process.internal.ExecException: Process 'Gradle Test Executor 7' finished with non-zero exit value 137
	at org.gradle.process.internal.DefaultExecHandle$ExecResultImpl.assertNormalExitValue(DefaultExecHandle.java:380)
	at org.gradle.process.internal.worker.DefaultWorkerProcess.waitForStop(DefaultWorkerProcess.java:190)
	at org.gradle.process.internal.worker.DefaultWorkerProcessBuilder$MemoryRequestingWorkerProcess.waitForStop(DefaultWorkerProcessBuilder.java:228)
	at org.gradle.api.internal.tasks.testing.worker.ForkingTestClassProcessor.stop(ForkingTestClassProcessor.java:121)
	at org.gradle.api.internal.tasks.testing.processors.RestartEveryNTestClassProcessor.endBatch(RestartEveryNTestClassProcessor.java:63)
	at org.gradle.api.internal.tasks.testing.processors.RestartEveryNTestClassProcessor.stop(RestartEveryNTestClassProcessor.java:57)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.FailureHandlingDispatch.dispatch(FailureHandlingDispatch.java:29)
	... 10 more
    2017/09/10 00:57:30.100 +0000 INFO [Test worker] [AzkabanDatabaseSetup] Creating new table execution_logs version 3.10.1
Overall
  SUCCESS (0 tests, 0 passed, 0 failed, 0 skipped)
```

HappyRay:
It appears that there were 8 test executors' running. Maybe reducing the number may help.

PR #1458, opened 2017-09-13
build: Enable gradle scans on travis CI to debug #1448

inramana:
## Issue
#1448 
Travis CI builds sometimes fail with the following error

> Process 'Gradle Test Executor 6' finished with non-zero exit value 137 travis

@HappyRay alleviated some of the resource issues with #1402 but we are still seeing sporadic failures. 

## Fix
Gradle has a tool to investigate builds including metrics like memory, threads GC etc called [scans](https://scans.gradle.com/get-started) .This change enables build scans on travis to help debug failures and also provide insight into our builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ languages: java
 jdk:
   - oraclejdk8
 sudo: false
-script: ./gradlew clean build --no-daemon --info
+script: ./gradlew clean build --scan --no-daemon --info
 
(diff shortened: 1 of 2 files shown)


HappyRay:
You can reference another issue with #1448 syntax.
It would be great if you can summarize the issue here as well.

reallocf:
Looks interesting! Is this something that would be useful for us to run locally too?

PR #1459, opened 2017-09-13
refactor DatabaseSetup class for Quartz Test being developed [Refactor]

kunkun-tang:
constructing Quartz Test, which needs to create tables in h2-in-memory. This PR does:

* Remove unknown code
* simplify logic

### DIFF
--- a/azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
+++ b/azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
@@ -33,7 +33,9 @@ import org.apache.log4j.Logger;
  * This class is used for creating DB tables by specifying where the create scripts are. The
  * input to this class is a folder path, which includes all create table sql scripts. The script's
  * name should follow: create.[table_name].sql in order to be identified. This class is used for
- * unit test only for now. We need to fix some reliability issues if we rely on this class to
+ * unit test only for now.
+ *
+ * Todo kunkun-tang: We need to fix some reliability issues if we rely on this class to
  * create tables when launching AZ in future.
  */
 public class DatabaseSetup {
--- a/azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
+++ b/azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
@@ -23,46 +23,22 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 
-/**
- * TODO kunkun-tang: this class is quite messy now, needs to be refactored.
- */
 public class DatabaseSetup {
 
-  public static final String DATABASE_SQL_SCRIPT_DIR =
-      "database.sql.scripts.dir";
   private static final Logger logger = Logger
       .getLogger(DatabaseSetup.class);
-  private static final String DEFAULT_SCRIPT_PATH = "sql";
   private static final String CREATE_SCRIPT_PREFIX = "create.";
   private static final String SQL_SCRIPT_SUFFIX = ".sql";
 
-  private static final String INSERT_DB_PROPERTY =
-      "INSERT INTO properties (name, type, value, modified_time) values (?,?,?,?)";
-  private static final String UPDATE_DB_PROPERTY =
-      "UPDATE properties SET value=?,modified_time=? WHERE name=? AND type=?";
-
   private final AzkabanDataSource dataSource;
-  private final Set<String> missingTables = new HashSet<>();
-  private final Map<String, String> installedVersions = new HashMap<>();
-  private String version;
-  private boolean needsUpdating;
-
+  private final Set<String> tables = new HashSet<>();
   private String scriptPath = null;
-
-  public DatabaseSetup(final AzkabanDataSource ds) {
-    this.dataSource = ds;
-    if (this.scriptPath == null) {
-      this.scriptPath = DEFAULT_SCRIPT_PATH;
-    }
-  }
 
   public DatabaseSetup(final AzkabanDataSource ds, final String path) {
     this.dataSource = ds;
@@ -71,11 +47,11 @@ public class DatabaseSetup {
 
   public void updateDatabase()
       throws SQLException, IOException {
-    findMissingTables();
-    createNewTables();
+    collectAllTables();
+    createTables();
   }
 
-  private void findMissingTables() {
+  private void collectAllTables() {
     final File directory = new File(this.scriptPath);
     final File[] createScripts =
         directory.listFiles(new PrefixSuffixFileFilter(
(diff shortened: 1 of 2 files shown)


azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -85,84 +61,38 @@ private void findMissingTables() {
...
+        runTableScripts(conn, table);
       }
     } finally {
       conn.close();
     }
   }
 
+  private void runTableScripts(final Connection conn, final String table)
+      throws IOException, SQLException {
+    logger.info("Creating new table " + table );
```
chengren311: extra space, do you use save plugin?

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -190,11 +120,8 @@ public boolean accept(final File pathname) {
 
       final String name = pathname.getName();
       final int length = name.length();
-      if (this.suffix.length() > length || this.prefix.length() > length) {
-        return false;
-      }
-
-      return name.startsWith(this.prefix) && name.endsWith(this.suffix);
+      return !(this.suffix.length() > length || this.prefix.length() > length) && name
```
chengren311: why not this.suffix.length() <= length && this.prefix.length() <= length

HappyRay:
Thanks for the refactoring. Could you add a bit more details on what's changed? It seems that you removed some logic to do an incremental update? How is the behavior going to be different? Or what the new code is designed to do?

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -23,59 +23,35 @@
...
-import java.util.Map;
 import java.util.Set;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 
-/**
- * TODO kunkun-tang: this class is quite messy now, needs to be refactored.
- */
 public class DatabaseSetup {
```
HappyRay: Could you add a javadoc explaining what this class does? According to the style guide, every public interfaces need to have javadocs.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -85,84 +61,38 @@ private void findMissingTables() {
...
-            false);
-      }
-      for (final String table : this.missingTables) {
-        if (!table.equals("properties")) {
-          runTableScripts(conn, table, this.version, this.dataSource.getDBType(), false);
-          // update version as we have create a new table
-          this.installedVersions.put(table, this.version);
-        }
+      for (final String table : this.tables) {
+        runTableScripts(conn, table);
```
chengren311: if one table creation throws an exception, should the rest fail or proceed?

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Yes, I'm using save plugin.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Good find. I was adopting Intellij's suggestion from previous code, and not carefully checking that.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -23,59 +23,35 @@
...
-  private static final String UPDATE_DB_PROPERTY =
-      "UPDATE properties SET value=?,modified_time=? WHERE name=? AND type=?";
-
   private final AzkabanDataSource dataSource;
-  private final Set<String> missingTables = new HashSet<>();
-  private final Map<String, String> installedVersions = new HashMap<>();
-  private String version;
-  private boolean needsUpdating;
-
+  private final Set<String> tables = new HashSet<>();
```
chengren311: i don't think tables as member variable is necessary.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -23,59 +23,35 @@
...
   public DatabaseSetup(final AzkabanDataSource ds, final String path) {
     this.dataSource = ds;
     this.scriptPath = path;
   }
 
   public void updateDatabase()
       throws SQLException, IOException {
-    findMissingTables();
-    createNewTables();
+    collectAllTables();
```
chengren311: Do you think this is better so that we remove the tables from member variables?
Set<String> tables = collectAllTables();
createTables(tables);

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Thanks. Just added.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Yes, you should be right if we use this class for table create automation in production. This class is used for unit test only for now. We need to fix some reliability issues if we rely on this class to create tables when launching AZ in future. Updated java doc.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Modified.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Sounds a fair suggestion. Changing.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Should be right.

kunkun-tang:
@HappyRay The new code is designed for unit Testing use only for now. I haven't tested using new code  to create tables when launching actual AZ.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -23,59 +23,43 @@
...
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 import org.apache.commons.dbutils.QueryRunner;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 
 /**
- * TODO kunkun-tang: this class is quite messy now, needs to be refactored.
+ * This class is used for creating DB tables by specifying where the create scripts are. The
```
HappyRay: Please follow the coding style guide for how the summary line should look like.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
HappyRay: Why is it better to proceed? Isn't it better to fail fast? This is used in tests.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
HappyRay: +1

HappyRay:
Is azkaban-db/src/main/sql/database.properties still needed after this change?

HappyRay:
I saw that it is still used by azkaban.database.AzkabanDatabaseSetup
What's your plan for the azkaban.database package?
Do you plan to move it and consolidate with the .db package?

kunkun-tang:
I was considering consolidating the two packages, but am worried about if external users rely on `azkaban.database.AzkabanDatabaseSetup`. It is quite useful for launching soloserver in an empty Mysql DB today. We may support it in `zkaban.db.DatabaseSetup` future, then we should be able to retire it completely.

kunkun-tang:
The original database.AzkabanDatabaseSetup should still rely on `database.properties`. We can remove it after retiring database package completely.

PR #1460, opened 2017-09-14
google java style tuning for comments above a public class

kunkun-tang:


azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -30,12 +30,12 @@
 import org.apache.log4j.Logger;
 
 /**
- * This class is used for creating DB tables by specifying where the create scripts are. The
+ * Creating DB tables by specifying where the create scripts are. The
```
HappyRay: change to "Creates"

How about just "Creates DB tables."?

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
```diff
@@ -30,12 +30,12 @@
...
  * input to this class is a folder path, which includes all create table sql scripts. The script's
  * name should follow: create.[table_name].sql in order to be identified. This class is used for
  * unit test only for now.
  *
- * Todo kunkun-tang: We need to fix some reliability issues if we rely on this class to
+ * <p>Todo kunkun-tang: We need to fix some reliability issues if we rely on this class to
```
HappyRay: FYI, <p> is not required if you don't care about having a separate paragraph in the rendered javadoc.

kunkun-tang: Yes, I know that, but I would think we should a separate line for this.

azkaban-db/src/main/java/azkaban/db/DatabaseSetup.java
kunkun-tang: Modified as suggested.

PR #1461, opened 2017-09-14
move one method out of PropsUtils [Refactor]

kunkun-tang:
In order to move metrics foundation class to a common place, I am creating `az-core` sub module. Since Metrics relies on some utils class in `azkaban-common`, the first step of this series of refactor is to clean up some methods in `azkaban-common', and it facilitate me to do those refactor in the following PRs.

In this code patch:
* I move `addCommonFlowProperties` out of PropsUtils, so that we can move PropsUtil to az-core in the next PR easily. 
* Also, this method is not suitable to stay in `PropsUtils`.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
+++ b/azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


reallocf:
Is there an issue or a project or something here on Github that illustrates the long term motivation for this change? What's the purpose of `az-core` compared to `azkaban-common`?

kunkun-tang:
Let me create one issue. The original motivation is that azkaban-db needs to do monitoring, and can not call azkaban-common today. But there are more motivations beyond that.

reallocf:
Cool - thank you!

HappyRay:
I suggested the az- prefix so that it is easier to separate the new modules from legacy modules and allow us to apply different policies to them such as code style enforcement policies.

HappyRay:
Be careful with the Props class, it is used in the public interfaces. User code may depend on the class's package name.

kunkun-tang:
#1462

kunkun-tang:
@HappyRay That's a very important reminder! What if they still rely on `azkaban-common` jar? In that way, people will not be able to get the Props. So, should we incorporate az-core classes into azkaban-common jar as a workaround for now?

Issue #1462, opened 2017-09-14
We need az-core module to put common code


kunkun-tang:
`azkaban-common` sub module today is **too common**. The original authors put too much burden in this place. Any other module relied on  `azkaban-common` can not call code there, so it causes problems. We need to create az-core module, which should be the baseline foundation class for all other az modules. For instance, Metrics Registry class should be there, and all other modules can call it to register monitoring.

kunkun-tang:
#1461

PR #1463, opened 2017-09-14
Moving part of Util class to new az-core module

kunkun-tang:
This code patch is to create az-core sub module and move some utils class to here. The discussion of this proposal is [here](https://github.com/azkaban/azkaban/issues/1462).

Used Intellij to do the refactor.


HappyRay:
Should the test code be moved too? e.g. azkaban.utils.JsonUtilsTest

PR #1466, opened 2017-09-15
Refactor uploading project

jamiesjc:
1. azkaban.project.AzkabanProjectLoader#uploadProject method is too chunky and complicated. Break it up into small logical pieces so that it's easier to understand.
2. azkaban.project.DirectoryFlowLoader is mainly responsible for loading project flows. It should be separated from project validation which is used by external plugins like byteRay. 
3. Given a project directory, DirectoryFlowLoader should only care about extracting the flow/job info and populating it to the project.
4. After separating the DirectoryFlowLoader, it would be easier to create and implement the new flowLoader for Flow 2.0 design.


azkaban-common/src/main/java/azkaban/project/validator/ValidatorManager.java
```diff
@@ -28,10 +28,9 @@
   Map<String, ValidationReport> validate(Project project, File projectDir);
 
   /**
-   * The ValidatorManager should have a default validator which checks for the most essential
-   * components of a project. The ValidatorManager should always load the default validator. This
-   * method returns the default validator of this ValidatorManager.
+   * ValidatorManager will not have any default validator.
    */
+  @Deprecated
   ProjectValidator getDefaultValidator();
```
jamiesjc: Found one internal team is using this method. Will coordinate with them about the migration solution.

kunkun-tang: Good find!

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -80,8 +80,46 @@
...
     log.info("Uploading files to " + project.getName());
 
-    // Unzip.
-    File file = null;
+    // Since props is an instance variable of ProjectManager, and each
+    // invocation to the uploadProject manager needs to pass a different
+    // value for the PROJECT_ARCHIVE_FILE_PATH key, it is necessary to
+    // create a new instance of Props to make sure these different values
+    // are isolated from each other.
+    final Props prop = new Props(this.props);
```
chengren311: new Props(this.props.getParent(), this.props) ?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -80,8 +80,46 @@
...
+    // Check the validation report.
+    if(!isReportStatusValid(reports, project, file)) {
+      return reports;
+    }
+
+    // Upload the project to DB and storage.
+    persistProject(project, archive, uploader);
+
+    // Clean up project directories and old installations.
+    cleanUpProject(project, file);
```
chengren311: should cleanup be in finally block?

azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -128,7 +131,7 @@ public Flow getFlow(final String flowId) {
   }
 
   public void setFlows(final Map<String, Flow> flows) {
-    this.flows = flows;
+    this.flows = new HashMap<>(flows);
```
chengren311: if you want them to be truly immutable, maybe consider this.flows =  Collections.unmodifiableMap(flows), same for jobPropsMap and propsList

azkaban-common/src/main/java/azkaban/project/validator/XmlValidatorManager.java
```diff
@@ -238,7 +231,7 @@ private void parseItemTag(final Node node, final Props props) {
 
   @Override
   public ProjectValidator getDefaultValidator() {
-    return this.validators.get(DEFAULT_VALIDATOR_KEY);
+    return null;
```
chengren311: why null?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Here prop doesn't have to be exactly the same. It can just inherit the parent prop. I think either way is fine. I didn't change the original implementation here, just moved it to a new method.

azkaban-common/src/main/java/azkaban/project/Project.java
jamiesjc: Actually, flows can be mutable. When uploading project in db, it will set the project id and project version for all the flows in the project. See azkaban.project.AzkabanProjectLoader#persistProject

azkaban-common/src/main/java/azkaban/project/validator/XmlValidatorManager.java
jamiesjc: With the new change, there won't be any default validator. Anyway it will return null if you try to get the default validator from validators map. We plan to deprecate getDefaultValidator() method.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: I can put deleting temp directory to a finally block. But for cleaning up old project installations, it should happen only when uploading new project is successful.

azkaban-common/src/main/java/azkaban/project/Project.java
jamiesjc: I'll change it to ImmutableMap since only the value inside flow object will be modified, but the map itself won't be modified. Thanks!

azkaban-common/src/main/java/azkaban/project/validator/ValidatorManager.java
jamiesjc: Already discussed with the team and they agree to use the old azkaban common module jar while we are developing new features. When the Flow 2.0 DSL change is rolled out, they will migrate to the new implementation or we provide some external API to them depending on the number of use cases then.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -49,6 +48,7 @@
 class AzkabanProjectLoader {
 
   private static final Logger log = LoggerFactory.getLogger(AzkabanProjectLoader.class);
+  private static final String DIRECTORY_FLOW_REPORT_KEY = "Directory Flow";
```
HappyRay: Is this part of the user-facing contract? Where is it used?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -49,6 +48,7 @@
 class AzkabanProjectLoader {
```
HappyRay: As you refactor, it would be good to add missing class docs.

HappyRay: New code should follow the dev guide as much as possible.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -79,10 +79,53 @@
       final File archive, final String fileType, final User uploader, final Props additionalProps)
       throws ProjectManagerException {
     log.info("Uploading files to " + project.getName());
+    final Map<String, ValidationReport> reports;
+
+    // Since props is an instance variable of ProjectManager, and each
```
HappyRay: I am not sure I follow the comment. Let's chat off line.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -79,10 +79,53 @@
...
+    // value for the PROJECT_ARCHIVE_FILE_PATH key, it is necessary to
+    // create a new instance of Props to make sure these different values
+    // are isolated from each other.
+    final Props prop = new Props(this.props);
+    prop.putAll(additionalProps);
 
-    // Unzip.
     File file = null;
     try {
+      // Unzip the project.
```
HappyRay: Consider breaking up this method to smaller units.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -79,10 +79,53 @@
...
+      file = unzipProject(archive, fileType);
+
+      // Validate the project, load external validators to do the validation.
```
HappyRay: Is this comment needed? 

> load external validators to do the validation.

It seems to be an implementation detail. Do readers need to know here?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -166,33 +200,37 @@
...
-    return reports;
+  private void cleanUpProjectTempDir(final File file) {
+    log.info("Cleaning up temp files.");
+    try {
+      if (file != null) {
+        FileUtils.deleteDirectory(file);
+      }
+    } catch (final IOException e) {
+      file.deleteOnExit();
+      e.printStackTrace();
```
HappyRay: log exceptions properly.

azkaban-common/src/main/java/azkaban/project/validator/XmlValidatorManager.java
HappyRay: Is it ok to remove this method?

HappyRay:
Thanks for the work. The new code is more readable.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Will change it, thanks!

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: I don't think it's a user-facing contract. It is used for displaying the error/warn messages to users when there is something wrong in uploading projects. It used to be: `public static final String DEFAULT_VALIDATOR_KEY = "Directory Flow";`. I just want to keep the same UI display.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Will remove, thanks!

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Will remove it, thanks!

azkaban-common/src/main/java/azkaban/project/validator/XmlValidatorManager.java
jamiesjc: Since this method is overriding the one in ValidatorManager interface, we need to remove both together. Currently I put @Deprecated in azkaban.project.validator.ValidatorManager#getDefaultValidator, but I think it's ok to remove it entirely.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Will do, thanks!

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
HappyRay: Got it. Thanks

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
HappyRay: Discussed offline. I understand it now.
In general, I think the Props are overused in the current code.

PR #1468, opened 2017-09-16
Creating DB Metrics

kunkun-tang:
The goal of this PR is to build metrics in `azkaban-db` module. The most important metrics are DB connection monitorings. Here's the steps to complete the task:
1. Create DBMetrics class;
2. make AzkabanDataSource Guiceable, in order to Inject DBMetrics. Adding Binding method in `azkabanCommonModule` to guice it;
3. Moving Common MetricsTestUtility method to Core module, so that db module can call it. Also, add test dependency gradle code.
4. After that, add 3 metrics to MysqlDataSource.

Tested locally.


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -72,9 +69,9 @@ public AzkabanCommonModule(final Props props) {
   protected void configure() {
     install(new AzkabanCoreModule(this.props));
     bind(Storage.class).to(resolveStorageClassType());
+    bind(AzkabanDataSource.class).to(resolveDataSourceType());
```
HappyRay: Why not implement a provider that returns an instance of AzkabanDataSource?

HappyRay: Never mind. I understand now. You would like to use Guice to construct the DataSource instances instead of using "new" in the provider.

kunkun-tang: Yes. Thanks to Guice, we don't need to actually new a class. How powerful Guice is!

PR #1473, opened 2017-09-16
Use assertj instead of hamcrest in JobRunnerTest

### LINKED ISSUE #1471: Use assertj instead of hamcrest in JobRunnerTest
`TestUtils.await().until(() -> loader.getNodeUpdateCount("testJob"), is(2));`

^ `is()` there is imported from hamcrest. Use assertj instead. Not sure how..

https://github.com/azkaban/azkaban/blob/master/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java#L319

HappyRay:
Fix #1471


HappyRay:
The verbosity seems to be comparable. Thanks to autocompletion, I don't have the type more.

I personally find assertj version easier to understand.

PR #1478, opened 2017-09-17
Build: remove eclipse gradle plugin

HappyRay:
The dev setup now depends on Intellij specific plugins, styles etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -114,7 +114,6 @@ ext.deps = [
 
 subprojects {
     apply plugin: 'java'
-    apply plugin: 'eclipse'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -114,7 +114,6 @@ ext.deps = [
 
 subprojects {
     apply plugin: 'java'
-    apply plugin: 'eclipse'
```
kunkun-tang: Can Eclipse support Azkaban code afterward?

HappyRay: I am not sure. I don't use Eclipse. 
If we don't actively use it, how do we know it is working?
If there is a need for it, someone should send a PR.

Issue #1451, opened 2017-09-11
Missing documentation and UI color for FAILED_SUCCEEDED


juhoautio:
`job.succeed.on.failure=true` allows continuing the flow execution over to children of a job even if it failed.

This is a really useful feature and **should be documented**.

It works! But it's **missing the UI color** (so uses the black/graphite default color in the graph view):

<img width="415" alt="nayttokuva 2017-09-11 kello 13 28 34" src="https://user-images.githubusercontent.com/4446608/30270557-a4f6a0e4-96f5-11e7-9497-8e8807dbadd9.png">

<img width="533" alt="nayttokuva 2017-09-11 kello 13 31 46" src="https://user-images.githubusercontent.com/4446608/30270789-9b8a8dda-96f6-11e7-86fb-17c53d82450e.png">

Mouse-over:

<img width="524" alt="nayttokuva 2017-09-11 kello 13 27 18" src="https://user-images.githubusercontent.com/4446608/30270654-fbc2e75c-96f5-11e7-86a1-c18818265af9.png">

Worse for Job List, because it's just blank, you might not notice that this job didn't succeed:

<img width="1154" alt="nayttokuva 2017-09-11 kello 13 29 07" src="https://user-images.githubusercontent.com/4446608/30270655-00e39812-96f6-11e7-82ef-2736e6e4bdec.png">

Finally, **in my opinion the eventual flow status should also be marked to be FAILED_SUCCEEDED** if any of the executed jobs are. As the flow FAILED_SUCCEEDED should only override SUCCEEDED, though. For example FAILED and KILLED would still override FAILED_SUCCEEDED as the flow status.

I'm willing to fix these issues – please advise, do you agree with my suggestions?

wyukawa:
I have thought about job.succeed.on.failure=true in some case.
for example, although job A doesn't depend on job B, I want to execute A after B to avoid high load average.

In this case, this setting is useful.

but, if job A fails, the eventual flow status become to be SUCCEEDED.

 In my opinion the eventual flow status should not be marked to be SUCCEEDED.

Anyway, job.succeed.on.failure option should be documented.
 
 Thanks

HappyRay:
@juhoautio Thanks for the suggestion.

We are thinking about introducing a more general form: conditional flow in the future. When we do, this will likely be a special case and we may remove this config from the code. So I would hesitate to document it now.

juhoautio:
Ok. Can we add the missing color though? What about allowing retries before proceeding?

[...]

PR #1483, opened 2017-09-19
Add UI colors for FAILED_SUCCEEDED

juhoautio:
FAILED_SUCCEEDED status was missing color in the UI, so add it.

The issue https://github.com/azkaban/azkaban/issues/1451 has some further details.

nayttokuva 2017-09-19 kello 22 37 04

nayttokuva 2017-09-19 kello 22 43 14

nayttokuva 2017-09-19 kello 22 37 25

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/historypage.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/historypage.vm
@@ -251,6 +251,8 @@
                     <option value=90>Skipped</option>
                     <option value=100>Disabled</option>
                     <option value=110>Queued</option>
+                    <option value=120>Failed, treated as success</option>
+                    <option value=125>Cancelled</option>
                   </select>
                 </div>
               </div>
--- a/azkaban-web-server/src/main/less/azkaban-graph.less
+++ b/azkaban-web-server/src/main/less/azkaban-graph.less
@@ -103,6 +103,15 @@
   fill: #FFF;
 }
 
+.FAILED_SUCCEEDED > g > rect {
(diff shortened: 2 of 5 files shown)


azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/historypage.vm
```diff
@@ -251,6 +251,8 @@
                     <option value=90>Skipped</option>
                     <option value=100>Disabled</option>
                     <option value=110>Queued</option>
+                    <option value=120>Failed, treated as success</option>
```
kunkun-tang: Is the message too long to be displayed ?

juhoautio: I added screenshots in the description, please check it out. I don't think it's too bad. I think this label makes it clear to the users. Can you think of something else that would be as clear as this, without being clumsy? It's a special case after all, won't be too common to appear in practice.

juhoautio:
I'm not so sure about the color though. Would you agree on orange/yellow? To me that would be more like "WARN log level" ie. something between ok & failed. Using the light red (as in cancelled) is not as good now that I think of it.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/historypage.vm
kunkun-tang: I didn't check out the screenshot. Looks good. +1 on the long message for detailed explanation. Otherwise, users should not understand what does FAILED_SUCCEEDED mean ? I always get confused, too.

PR #1484, opened 2017-09-19
Fix broken HTML in SLA message

juhoautio:
Using `</br>` is a "closing tag", doesn't exit. The self-sufficient linebreak tag is `<br/>`.

Save Actions plugin moved the imports around a bit.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/sla/SlaOption.java
+++ b/azkaban-common/src/main/java/azkaban/sla/SlaOption.java
@@ -17,6 +17,10 @@
 package azkaban.sla;
 
 import azkaban.executor.ExecutableFlow;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/sla/SlaOption.java
```diff
@@ -97,11 +98,11 @@ public static String createSlaMessage(final SlaOption slaOption, final Executabl
           (String) slaOption.getInfo().get(SlaOption.INFO_DURATION);
       final String basicinfo =
           "SLA Alert: Your flow " + flowName + " failed to FINISH within "
-              + duration + "</br>";
```
kunkun-tang: Good Catch! Looks like this has been failing for a long time.

HappyRay:
@kunkun-tang any reason this PR is not merged?

kunkun-tang:
Merging. @HappyRay

PR #1485, opened 2017-09-19
Include job/flow name in SLA alert title

juhoautio:
This way already the e-mail title will tell which a) job level SLA, or b) flow level SLA, was triggered.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -101,7 +101,7 @@ public class Emailer extends AbstractMailer implements Alerter {
   }
 
   private void sendSlaAlertEmail(final SlaOption slaOption, final String slaMessage) {
-    final String subject = "Sla Violation Alert on " + getAzkabanName();
+    final String subject = "SLA violation for " + getJobOrFlowName(slaOption) + " on " + getAzkabanName();
     final String body = slaMessage;
     final List<String> emailList =
         (List<String>) slaOption.getInfo().get(SlaOption.INFO_EMAIL_LIST);
@@ -120,6 +120,15 @@ public class Emailer extends AbstractMailer implements Alerter {
           this.commonMetrics.markSendEmailFail();
         }
       }
+    }
+  }
+
+  private String getJobOrFlowName(SlaOption slaOption) {
+    String jobName = (String) slaOption.getInfo().get(SlaOption.INFO_JOB_NAME);
+    if (org.apache.commons.lang.StringUtils.isNotBlank(jobName)) {
+      return jobName;
+    } else {
+      return (String) slaOption.getInfo().get(SlaOption.INFO_FLOW_NAME);
     }
   }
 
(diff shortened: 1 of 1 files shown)


kunkun-tang:
@chengren311  Could you please review the sla related change? You should know more than me.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -123,6 +123,15 @@ private void sendSlaAlertEmail(final SlaOption slaOption, final String slaMessag
     }
   }
 
+  private String getJobOrFlowName(SlaOption slaOption) {
+    String jobName = (String) slaOption.getInfo().get(SlaOption.INFO_JOB_NAME);
+    if (org.apache.commons.lang.StringUtils.isNotBlank(jobName)) {
```
chengren311: isNotEmpty to be safer?

chengren311: consider importing org.apache.commons.lang.StringUtils ?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -123,6 +123,15 @@ private void sendSlaAlertEmail(final SlaOption slaOption, final String slaMessag
...
+      return jobName;
```
chengren311: maybe return flowName+"."+jobName if job name exists to be more clear?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
juhoautio: Not possible because there's a class called `StringUtils` in the same package (`azkaban.utils`) as `Emailer`.

juhoautio: Why would we want to include a "job name" that is whitespace only? I don't think that should be even possible.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
juhoautio: Will do. I had thought that it would be either-or, but apparently flow name is available in SLA info also for job level SLAs.

juhoautio:
@chengren311: updated

juhoautio:
Thanks for the review @chengren311 and thanks for accepting this change 🙇

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -123,6 +123,16 @@ private void sendSlaAlertEmail(final SlaOption slaOption, final String slaMessag
     }
   }
 
+  private String getJobOrFlowName(SlaOption slaOption) {
```
HappyRay: Could you add unit tests? 
The email component is not unit test friendly yet. We would like to refactor it.
I think components of this change such as this method can be unit tested at least.

juhoautio: Umm.. do you mean that _you_ will refactor it to be unit test friendly (before writing a test)?

HappyRay: @jamiesjc plans to refactor the email component.
I am suggesting that even before that takes place, new changes should have automated test coverage as much as possible.
It's on the todo list to get code coverage report integrated with the review process.

Issue #1486, opened 2017-09-19
Rewrite the DAG handling code


HappyRay:
The current DAG processing logic is complex and hard to maintain.
e.g. we discovered many race conditions. Code review becomes difficult since it is hard to reason about the correctness. 

We should simplify the logic, make it easier to understand and maintain.

PR #1487, opened 2017-09-19
[Prototype. do not merge] Dag engine refactoring

HappyRay:
This is work in progress. Submit this review to make it easier for others to see the code.
#1486


azkaban-exec-server/src/main/java/azkaban/dag/Status.java
```diff
@@ -0,0 +1,48 @@
...
+enum Status {
+  READY, // ready to run
+  DISABLED, // disabled by users. Treat as the node has the status of success
+  BLOCKED, // temporarily blocked. Need to be unblocked by another external event
+  RUNNING,
+  SUCCESS,
+  FAILURE,
+
+  // doesn't run because one of the nodes it depends on fails or is killed. Apples to a node only.
+  CANCELED,
```
juhoautio: `azkaban.executor.Status` has `CANCELLED` with 2 L's. Just saying..

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
```diff
@@ -0,0 +1,48 @@
...
+
+enum Status {
+  READY, // ready to run
+  DISABLED, // disabled by users. Treat as the node has the status of success
+  BLOCKED, // temporarily blocked. Need to be unblocked by another external event
+  RUNNING,
+  SUCCESS,
+  FAILURE,
+
+  // doesn't run because one of the nodes it depends on fails or is killed. Apples to a node only.
```
juhoautio: Applies :)

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  void markKilled() {
+    assert (this.status == Status.KILLING);
+    changeStatus(Status.KILLED);
+    this.flow.checkFinished();
+  }
+
+  boolean isInTerminalState() {
+    return Status.isTerminal(this.status);
```
juhoautio: Could these methods be simplified to be on the enum instances themselves? For example this line would become:

```java
this.status.isTerminal();
```

Maybe that would even mean that there's no need for this method (`Node#isInTerminalState`).

This has always bothered me with the static methods of `azkaban.executor.Status`: `isStatusRunning` and `isStatusFinished`..

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -0,0 +1,196 @@
...
+ * Node in a DAG: Directed acyclic graph.
+ */
+class Node {
+
+  private final String name;
+
+  private final NodeProcessor nodeProcessor;
+
+  // The nodes that this node depends on.
+  private final List<Node> parents = new ArrayList<>();
```
juhoautio: This terminology can be confusing because if this is a sub-flow, parent should be the enclosing flow. So this field could be renamed to "dependencies", and below "children" -> "dependents"?

azkaban-exec-server/src/main/java/azkaban/dag/Flow.java
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * Gets all the nodes that are ready to run.
+   *
+   * @return a list of nodes that are ready to run
+   */
+  private List<Node> getReadyNodes() {
+    final List<Node> readyNodes = new ArrayList<>();
+    for (final Node node : this.nodes) {
+      if (node.isReady()) {
```
juhoautio: Just checking, I notice you have at least some unit test for sub-flows too, but I wonder if this works correctly? The thing is, you can't start a sub-flow "alone" even if it doesn't have any direct dependencies. It should be started once it's enclosing parent flow has been allowed to start. Maybe you are going to simplify this so that the parent flow node is added as a dependency to the starting nodes of a sub-flow?

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -0,0 +1,196 @@
...
+    for (final Node child : this.children) {
+      child.check();
+    }
+    this.flow.checkFinished();
+  }
+
+  /**
+   * Checks if all the dependencies are met and run if they are.
+   */
+  private void check() {
```
juhoautio: Could be called `runIfAllowed`.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -0,0 +1,196 @@
...
+      if (!node.isSuccessEffectively()) {
+        return;
+      }
+    }
+
+    run();
+  }
+
+  /**
+   * @return true if the dependency requirement is considered satisfied
```
juhoautio: Maybe add to the end of that sentence: "..for the nodes that depend on this node". Without that it seems a bit confusing, I guess because of the word "dependency".

azkaban-exec-server/src/main/java/azkaban/dag/NodeProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package azkaban.dag;
+
+public interface NodeProcessor {
+
+  /**
+   * Changes the status of the node.
+   *
+   * Typically a processor implementation should handle the RUNNING and KILLING status by staring
```
juhoautio: starting

azkaban-exec-server/src/test/java/azkaban/dag/DagServiceTest.java
```diff
@@ -0,0 +1,306 @@
...
+   *   sfb ( subflow. "sb" stands for sub flow. This prefix is used in a node name.)
+   *   |
+   *   c
+   *
+   * subflow:
+   * a b
+   * </pre>
+   */
+  @Test
+  public void simple_subflow_success_case() throws Exception {
```
juhoautio: Why no camel case? :)

azkaban-exec-server/src/test/java/azkaban/dag/NullFlowProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+ * the License.
+ */
+
+package azkaban.dag;
+
+public class NullFlowProcessor implements FlowProcessor {
+
+  @Override
+  public void changeStatus(final Flow flow, final Status status) {
+    return;
```
juhoautio: Wouldn't a default Mockito mock do exactly the same?

azkaban-exec-server/src/main/java/azkaban/dag/Flow.java
```diff
@@ -0,0 +1,141 @@
...
+        case KILLED:
+          killed = true;
+          break;
+        default:
+          break;
+      }
+    }
+
+    if (failed) {
+      // If there are both failures and killed nodes, the flow status should be failed.
```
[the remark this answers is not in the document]
HappyRay: Make sense. I will update it. Thanks.

In either case, the job status should reflect what actually happened.

azkaban-exec-server/src/main/java/azkaban/dag/Flow.java
HappyRay: The sub-flows are handled at a higher level as demonstrated by the unit tests. 
At the DAG engine level, it doesn't have the sub-flow concept. This is to separate the concerns so that each layer's code can be simplified. 
The higher layer will make sure the sub-flow won't start until the parent flow tells it to start. This happens as the node representing the sub-flow is asked to run.

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: Oops. I can't rely on spell checking too much, can I?

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: I didn't even notice. https://www.grammarly.com/blog/canceled-vs-cancelled/

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: As mentioned earlier, at this layer, there is no concept of sub-flows.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: Good suggestion. Thanks

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: Let me try your suggestion.

azkaban-exec-server/src/test/java/azkaban/dag/DagServiceTest.java
HappyRay: Because camel case doesn't' work well with long names. Google style guide permits it.

azkaban-exec-server/src/test/java/azkaban/dag/NullFlowProcessor.java
HappyRay: Good point.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: Removed this method.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: I like your suggestion. Changed.

azkaban-exec-server/src/main/java/azkaban/dag/NodeProcessor.java
HappyRay: done

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: done

azkaban-exec-server/src/test/java/azkaban/dag/NullFlowProcessor.java
HappyRay: done

HappyRay:
State machine diagram

azkaban dag state machine diagram - page 1

HappyRay:
related papers:
https://blog.acolyer.org/2017/12/04/ffwd-delegation-is-much-faster-than-you-think/
https://blog.acolyer.org/2015/06/05/scalability-but-at-what-cost/

azkaban-exec-server/src/test/java/azkaban/dag/DagServiceTest.java
HappyRay: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names
allows such usage:
 
“Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in lowerCamelCase. One typical pattern is <methodUnderTest>_<state>, for example pop_emptyStack. There is no One Correct Way to name test methods.
“

HappyRay:
Replaced by #1755

Issue #1488, opened 2017-09-19
Leverage Quartz to enhance Azkaban Scheduler


kunkun-tang:
Writing down some existing issues:

* Web server deployment misses schedules
```
The regular Web server deployment has two steps:
Shutdown old Web server; 
Start new web server. 
```
The gap between the two steps generates web server unavailability time ranging from 15 seconds to 5 minutes. All AZ schedules are missed during this window. More than that, flow owners are not aware of that since AZ silently ignores them and does not print any messages.

* One flow can have only one schedule
Today one flow exists only one schedule. Users would like to have two schedules with different flow-level parameters. Not only that, multiple cron per schedule are also expected. 
https://github.com/azkaban/azkaban/issues/650

* Schedules need the ability to be Paused/Resumed
When users debug a flow, they need to pause the schedule. After debugging, the schedule should be restored to normal status.

PR #1489, opened 2017-09-19
Creating quartz scheduler

kunkun-tang:
Inspired by #1488 , this PR proposes bringing quartz to AZ. Quartz Scheduler provides 3 modes of job stores: RAMJobStore, JDBCJobStore, and TerraCottaJobStore. In order to persist the job statuses out of Application memory regardless of downtime, we recommend using `JDBC JobStore`. In fact, users can specify the mode in quartz properties.

This code patch creates Quartz server and manages Quartz API (e.g., start, pause, shut down) at class QuartzScheduler. We directly use h2-in-memory DB to directly test JDBC Quartz in unit test.


kunkun-tang:
@chengren311 You might want to take a look at correponding tests in this PR. Let me know if you have any questions.

HappyRay:
Could you add more details to the change description? Duplication between the change description and the issue is ok IMHO. It makes it easier for readers.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -151,6 +151,9 @@
      *    implies save latest 3 versions saved in storage.
      **/
     public static final String AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION = "azkaban.storage.artifact.max.retention";
+
+    // enable Quartz Scheduler if true.
+    public static final String ENABLE_QUARTZ= "enable.quartz";
```
HappyRay: prefix it with azkaban.
Maybe
azkaban.server.schedule.enable_quartz?

azkaban-db/src/main/sql/create.quartz-tables-all.sql
```diff
@@ -0,0 +1,165 @@
+-- This file collects all quartz table create statement required for quartz 2.2.1
```
HappyRay: The current convention is to create one .sql file per table. 
And the database setup code relies on it.

Is there any reason to break this convention?

azkaban-web-server/src/test/java/azkaban/scheduler/TestQuartzJob.java
```diff
@@ -0,0 +1,56 @@
+/*
```
HappyRay: Other tests are named with xxxTest.java.

azkaban-web-server/src/test/resources/create.quartz-all-tables.sql
```diff
@@ -0,0 +1,165 @@
+-- All necessary quartz table create statement collection for unit test purpose
```
HappyRay: If the sql statements are changed incorrectly and this file is not updated, the tests may pass.

Can you think of ways to remove this duplication?

az-core/src/main/java/azkaban/Constants.java
kunkun-tang: why do we need a long name? I like `azkaban.enable_quartz`?

azkaban-db/src/main/sql/create.quartz-tables-all.sql
kunkun-tang: these tables originally reside in one file together. I would believe we should not manage/modify these sql statements after those tables are built, as we should not change quartz code, right? In this way, why not just have a big file, which just serves as a reference to users.

azkaban-web-server/src/test/java/azkaban/scheduler/TestQuartzJob.java
kunkun-tang: This is a not test, right? Is SampleQuartzJob a good name?

azkaban-web-server/src/test/resources/create.quartz-all-tables.sql
kunkun-tang: I can not think of situations where we will change quartz SQL statements. Do I miss something?

azkaban-web-server/src/main/java/azkaban/scheduler/AbstractQuartzJob.java
```diff
@@ -0,0 +1,35 @@
...
+
+package azkaban.scheduler;
+
+import java.io.Serializable;
+import org.quartz.Job;
+import org.quartz.JobExecutionContext;
+
+public abstract class AbstractQuartzJob implements Job {
+
+  protected static <T extends Serializable> T asT(final Object service, final Class<T> type) {
```
chengren311: please document these methods

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzJobDescription.java
```diff
@@ -0,0 +1,67 @@
...
+  public QuartzJobDescription(final Class<T> jobClass,
+      final String groupName,
+      final Map<String, ? extends Serializable> contextMap) {
+
+    /**
+     * This check is necessary for raw type. Please see test
+     * {@link QuartzJobDescriptionTest#testCreateQuartzJobDescription2}
+     */
+    if (jobClass.getSuperclass() != AbstractQuartzJob.class) {
+      throw new RuntimeException("jobClass must extend AbstractQuartzJob class");
```
chengren311: i would prefer a more meaningful exception type, maybe like illegalargumentexception?

az-core/src/main/java/azkaban/Constants.java
HappyRay: I am less worried about the length. We can copy and paste easily.
IMHO a more descriptive name is better. We would like to maintain a clear and consistent name hierarchy and namespace going forward.

azkaban-db/src/main/sql/create.quartz-tables-all.sql
HappyRay: I see. 
So you are ok with breaking the database setup code then?
Could you come up with a plan for how to evolve DB schemas?

azkaban-web-server/src/test/java/azkaban/scheduler/TestQuartzJob.java
HappyRay: I see. Then the name is fine.

azkaban-web-server/src/test/resources/create.quartz-all-tables.sql
HappyRay: When we update Quartz, the sql statements may change. Am I right?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzJobDescription.java
kunkun-tang: I found that CastException seems more appropriate. Updated

azkaban-web-server/src/main/java/azkaban/scheduler/AbstractQuartzJob.java
kunkun-tang: Wrote some java doc on top of it.

az-core/src/main/java/azkaban/Constants.java
kunkun-tang: Updated.

azkaban-db/src/main/sql/create.quartz-tables-all.sql
kunkun-tang: What do you mean by breaking database setup code? I believe new DatabaseSetup should be able to deal with both the two cases: 1. Sql folders; 2. one Sql file.

azkaban-web-server/src/test/resources/create.quartz-all-tables.sql
kunkun-tang: Upgrade Quartz! Yes, that's quite possible.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+    final StdSchedulerFactory schedulerFactory =
+        new StdSchedulerFactory(azProps.toProperties());
+    this.scheduler = schedulerFactory.getScheduler();
+  }
+
+  public void start() {
+    try {
+      this.scheduler.start();
+    } catch (final SchedulerException e) {
+      logger.warn("Error starting Quartz scheduler: ", e);
```
chengren311: warn seems too relaxing for this kind of exception. maybe error?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+    } catch (final SchedulerException e) {
+      logger.error("Exception shutting down scheduler: ", e);
+    }
+  }
+
+  public void unregisterJob(final String groupName) throws SchedulerException {
+    if(!ifJobExist(groupName)) {
+      logger.warn("can not find job with " + groupName + " in quartz.");
+    } else {
+      this.scheduler.deleteJob(new JobKey("job1", groupName));
```
chengren311: why job1?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+        .build();
+
+    this.scheduler.scheduleJob(job, trigger);
+    logger.info("Quartz Schedule with jobDetail " + job.getDescription() + " is registered.");
+  }
+
+
+  public boolean ifJobExist(final String groupName) throws SchedulerException {
+    final Set<JobKey> jobKeySet = this.scheduler.getJobKeys(GroupMatcher.jobGroupEquals(groupName));
+    return jobKeySet !=null && jobKeySet.size() > 0;
```
chengren311: space before null

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  Scheduler getScheduler() {
```
chengren311: access modifiers?

azkaban-web-server/src/main/java/azkaban/scheduler/AbstractQuartzJob.java
```diff
@@ -0,0 +1,38 @@
...
+package azkaban.scheduler;
+
+import java.io.Serializable;
+import org.quartz.Job;
+import org.quartz.JobExecutionContext;
+
+public abstract class AbstractQuartzJob implements Job {
+
+  /**
+   * Cast the object to the original one with the type. The object must extend Serializable.
```
chengren311: It is still not clear to me the purpose of the method with the description.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    if (!CronExpression.isValidExpression(cronExpression)) {
+      throw new SchedulerException("The cron expression string <" +  cronExpression + "> is not valid.");
+    }
+
+    // TODO kunkun-tang: Today all JobDetail should use "job1", and we will modify this when we
+    // start supporting multi schedules per flow.
+    final JobDetail job = JobBuilder.newJob(jobDescription.getJobClass())
+        .withIdentity("job1", jobDescription.getGroupName()).build();
```
chengren311: then can we make job1 a constant?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -184,8 +189,13 @@ public AzkabanWebServer(final Props props,
       DateTimeZone.setDefault(DateTimeZone.forID(timezone));
       logger.info("Setting timezone to " + timezone);
     }
-
     configureMBeanServer();
+
+    // Start Quartz Server
```
chengren311: the comment is unnecessary.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -184,8 +189,13 @@ public AzkabanWebServer(final Props props,
...
+    if(props.containsKey(ConfigurationKeys.ENABLE_QUARTZ) && props.getBoolean(ConfigurationKeys
+        .ENABLE_QUARTZ)) {
+      quartzScheduler.start();
```
chengren311: should we start the scheduler in constructor or prepareAndStartServer()?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
kunkun-tang: Good suggestion. Adding

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
kunkun-tang: What's the benefits over that approach?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
kunkun-tang: Adding public modifier here. I will add a shut down hook with web server stop.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
kunkun-tang: added.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
kunkun-tang: explained in a comment.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
kunkun-tang: Seems that shutdown doesn't need to `getScheduler`. For now, this method is never called by anyone.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
chengren311: to separate preparation work from constructor.

kunkun-tang: Moving to prepareAndStartServer()

PR #1490, opened 2017-09-19
Adding missing datasource singleton annoataion

kunkun-tang:
Both `H2FileDataSource ` and `MySQLDataSource ` should be singleton. The reason behind is that We would like all database operations to share the same connection pool to guarantee performance. 

Besides, removing singleton from `DatabaseOperator` as no states exist there. It is safe to have multiple DatabaseOperator instances.

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
@@ -148,7 +148,9 @@ public class AzkabanWebServerTest {
     assertSingleton(ExecutorEventsDao.class, injector);
     assertSingleton(ActiveExecutingFlowsDao.class, injector);
     assertSingleton(FetchActiveFlowDao.class, injector);
+    assertSingleton(AzkabanWebServer.class, injector);
     assertSingleton(H2FileDataSource.class, injector);
+
     SERVICE_PROVIDER.unsetInjector();
   }
 }
--- a/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
@@ -28,6 +28,7 @@ import azkaban.AzkabanCommonModule;
 import azkaban.database.AzkabanDatabaseSetup;
 import azkaban.database.AzkabanDatabaseUpdater;
 import azkaban.db.DatabaseOperator;
+import azkaban.db.H2FileDataSource;
 import azkaban.executor.ActiveExecutingFlowsDao;
 import azkaban.executor.AlerterHolder;
 import azkaban.executor.ExecutionFlowDao;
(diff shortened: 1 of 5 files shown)


HappyRay:
Do they need to be Singleton?
Is there a problem the way they are?
Singleton should be a special case, not a norm.

kunkun-tang:
@HappyRay I don't think we necessarily create multiple data source instances. One datasource should be able to manage all connections. Actually, I just realized it's already a singleton, since DatabaseOperator is set to singleton, and it is the only place to call AzkabanDataSource.

https://github.com/azkaban/azkaban/blob/b4d893fd7fd02fe50c090a1de55e6baf26ee5108/azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java#L37-L38

kunkun-tang:
Updated description.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -147,8 +148,7 @@ public void testInjection() throws Exception {
     assertSingleton(ExecutorEventsDao.class, injector);
     assertSingleton(ActiveExecutingFlowsDao.class, injector);
     assertSingleton(FetchActiveFlowDao.class, injector);
-    assertSingleton(AzkabanWebServer.class, injector);
-
+    assertSingleton(H2FileDataSource.class, injector);
```
HappyRay: How about the mysql datasource?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
```diff
@@ -147,8 +148,7 @@ public void testInjection() throws Exception {
     assertSingleton(ExecutorEventsDao.class, injector);
     assertSingleton(ActiveExecutingFlowsDao.class, injector);
     assertSingleton(FetchActiveFlowDao.class, injector);
-    assertSingleton(AzkabanWebServer.class, injector);
```
HappyRay: Why this change?

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
kunkun-tang: Mysql dataSource cannot be added here, since Guice only have h2 Props, which has conflicts against Mysql Props.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
kunkun-tang: This should be a typo. Adding back.

azkaban-web-server/src/test/java/azkaban/webapp/AzkabanWebServerTest.java
HappyRay: Can it be tested separately? It can be in a separate PR.

PR #1491, opened 2017-09-19
Enable new configuration other_hcat_clusters in AZ to talk to Metastores in HA mode

herrlich:
The logic that Azkaban uses to get the delegation token from ‘other_hcat_locations’ is trying to connect to all hcat servers in that field, and getting the delegation token from each hcat server. This is could cause delegation token confusion problems. Thus, we decide to introduce a new property field called “other_hcat_clusters”, which can also be set in the “properties” field in the workflow file. This field requires users to group the hcat servers to different clusters, in the form as “cluster1hcat01, cluster1cat02;cluster2hcat01,cluster2hcat02”. In our implementation, this string will be split by semicolon, and each group(cluster) will be traversed until we can get the delegation token from one of the machines in that cluster.


az-core/src/main/java/azkaban/utils/Props.java
```diff
@@ -430,6 +430,13 @@ public String getString(final String key) {
   }
 
   /**
+   * Returns a list of clusters with the comma as the separator of the value
+   */
```
beeramsunitha: Provide an example to clarify whats parsed by the method.

beeramsunitha: Add unit tests to cover different cases.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -122,6 +122,7 @@
    * supported, use comma to separate the values, values are case insensitive.
    */
   private static final String EXTRA_HCAT_LOCATION = "other_hcat_location";
+  private static final String EXTRA_HCAT_CLUSTERS = "other_hcat_clusters";
```
beeramsunitha: Update documentation to provide context around this (so the reader knows how it differes from EXTRA_HCAT_LOCATIONS). Also, we should mark EXTRA_HCAT_LOCATIONS as deprecated and ask users to switch to *_CLUSTERS to take advantage of Metastore HA.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -525,24 +527,42 @@ public synchronized void prefetchToken(final File tokenFile,
 
         cred.addToken(hcatToken.getService(), hcatToken);
 
-        // check and see if user specified the extra hcat locations we need to
-        // look at and fetch token.
-        final List<String> extraHcatLocations =
-            props.getStringList(EXTRA_HCAT_LOCATION);
-        if (Collections.EMPTY_LIST != extraHcatLocations) {
-          logger.info("Need to pre-fetch extra metaStore tokens from hive.");
+        // Added support for extra_hcat_clusters LIHADOOP-29840
```
beeramsunitha: Remove references to internal jiras

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -525,24 +527,42 @@ public synchronized void prefetchToken(final File tokenFile,
...
-            hiveConf.set(HiveConf.ConfVars.METASTOREURIS.varname, thriftUrl);
-            hcatToken =
-                fetchHcatToken(userToProxy, hiveConf, thriftUrl, logger);
+            hiveConf.set(HiveConf.ConfVars.METASTOREURIS.varname, thriftUrls);
+            hcatToken = fetchHcatToken(userToProxy, hiveConf, thriftUrls, logger);
             cred.addToken(hcatToken.getService(), hcatToken);
           }
-
+        } else {
+          // check and see if user specified the extra hcat locations we need to
```
beeramsunitha: Clarify (explicitly) that _LOCATIONS conf is processed only if _CLUSTERS is not provided.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
herrlich: Sorry forgot that

az-core/src/main/java/azkaban/utils/Props.java
kunkun-tang: +1 on Unit Test! You might need to create a test file under az-core module, as we don't have one.

az-core/src/test/java/azkaban/utils/PropsTest.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2014 LinkedIn Corp.
```
HappyRay: See the dev guide for how to update the copyright automatically.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -121,7 +121,17 @@
...
+  @Deprecated
   private static final String EXTRA_HCAT_LOCATION = "other_hcat_location";
+  /**
+   * this parameter is used to replace EXTRA_HCAT_LOCATION that could fail when one of the uris is not available.
+   * EXTRA_HCAT_CLUSTERS has the following format:
+   * other_hcat_clusters = "(thrift://hcat1:port,thrift://hcat2:port),(thrift://hcat3:port,thrift://hcat4:port)"
+   * Each string in the parenthesis is regarded as a "cluster", which we traverse until we get a delegation token
+   * The uris(hcat servers) in a "cluster" ensures HA is provided.
+   */
+  private static final String EXTRA_HCAT_CLUSTERS = "other_hcat_clusters";
```
HappyRay: Please move the keys to the Constant class. See the dev guide.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -525,24 +536,43 @@ public synchronized void prefetchToken(final File tokenFile,
...
-            hiveConf.set(HiveConf.ConfVars.METASTOREURIS.varname, thriftUrl);
-            hcatToken =
-                fetchHcatToken(userToProxy, hiveConf, thriftUrl, logger);
+            hiveConf.set(HiveConf.ConfVars.METASTOREURIS.varname, thriftUrls);
+            hcatToken = fetchHcatToken(userToProxy, hiveConf, thriftUrls, logger);
             cred.addToken(hcatToken.getService(), hcatToken);
           }
-
+        } else {
+          // If EXTRA_HCAT_CLUSTERS is not specified, we will continue with EXTRA_HCAT_LOCATION
```
beeramsunitha: Update it to say "Only if EXTRA_HCAT_CLUSTERS"

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -121,7 +121,17 @@
...
    * supported, use comma to separate the values, values are case insensitive.
    */
+  // Use EXTRA_HCAT_CLUSTERS instead
+  @Deprecated
   private static final String EXTRA_HCAT_LOCATION = "other_hcat_location";
+  /**
+   * this parameter is used to replace EXTRA_HCAT_LOCATION that could fail when one of the uris is not available.
+   * EXTRA_HCAT_CLUSTERS has the following format:
+   * other_hcat_clusters = "(thrift://hcat1:port,thrift://hcat2:port),(thrift://hcat3:port,thrift://hcat4:port)"
+   * Each string in the parenthesis is regarded as a "cluster", which we traverse until we get a delegation token
```
beeramsunitha: The comment is misleading; the code should not stop when it gets a delegation token from one cluster; it should get delegation tokens from all clusters.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -121,7 +121,17 @@
    * default one the system should pre-fetch hcat token from. Note: Multiple thrift uris are
    * supported, use comma to separate the values, values are case insensitive.
    */
+  // Use EXTRA_HCAT_CLUSTERS instead
```
beeramsunitha: Move the comment about using clusters into the block comment above.

az-core/src/test/java/azkaban/utils/PropsTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class PropsTest {
+  Props p = new Props();
+  private static final String EXTRA_HCAT_CLUSTERS = "other_hcat_clusters";
+
+  /* Test for getStringListFromCluster(String s)*/
+  @Test
+  public void testSplit1() {
+    String s1 = "(thrift://hcat1:port,thrift://hcat2:port),(thrift://hcat3:port,thrift://hcat4:port)";
+    p.put(EXTRA_HCAT_CLUSTERS, s1);
+    List<String> s2 = Arrays.asList("", "thrift://hcat1:port,thrift://hcat2:port" , "thrift://hcat3:port,thrift://hcat4:port");
```
beeramsunitha: We shouldn't be returning an empty string. The list should only  have 2 entries...

az-core/src/test/java/azkaban/utils/PropsTest.java
herrlich: Copied the unit test template from another file, will modify that.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -167,6 +167,15 @@
...
     public static final String AZKABAN_JOB_LOGGING_KAFKA_ENABLE = "azkaban.job.logging.kafka.enable";
+
+    /*
+     * this parameter is used to replace EXTRA_HCAT_LOCATION that could fail when one of the uris is not available.
+     * EXTRA_HCAT_CLUSTERS has the following format:
+     * other_hcat_clusters = "thrift://hcat1:port,thrift://hcat2:port;thrift://hcat3:port,thrift://hcat4:port"
+     * Each string in the parenthesis is regarded as a "cluster", and we will get a delegation token from each cluster.
+     * The uris(hcat servers) in a "cluster" ensures HA is provided.
+     **/
+    public static final String EXTRA_HCAT_CLUSTERS = "other_hcat_clusters";
```
kunkun-tang: Should this property be put under `JobProperties` class. Also, you might want to use `azkaban.job.other_hcat_clusters`

az-core/src/main/java/azkaban/utils/Props.java
```diff
@@ -430,6 +431,23 @@ public String getString(final String key) {
   }
 
   /**
+   * Returns a list of clusters with the comma as the separator of the value
+   * e.g., for input string: "thrift://hcat1:port,thrift://hcat2:port;thrift://hcat3:port,thrift://hcat4:port;"
```
kunkun-tang: What if the string has some format error? What will happen? Like 
```
thrift:hcat1:port;thrift:hcat2:port
```

az-core/src/test/java/azkaban/utils/PropsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    String s3 = "thrift://hcat1:port,thrift://hcat2:port     ;      thrift://hcat3:port,thrift://hcat4:port;";
+    p.put(EXTRA_HCAT_CLUSTERS, s3);
+    List<String> s4 = Arrays.asList( "thrift://hcat1:port,thrift://hcat2:port" , "thrift://hcat3:port,thrift://hcat4:port");
+    Assert.assertTrue(p.getStringListFromCluster(EXTRA_HCAT_CLUSTERS).equals(s4));
+
+    String s5 = "thrift://hcat1:port,thrift://hcat2:port";
+    p.put(EXTRA_HCAT_CLUSTERS, s5);
+    List<String> s6 = Arrays.asList("thrift://hcat1:port,thrift://hcat2:port");
+    Assert.assertTrue(p.getStringListFromCluster(EXTRA_HCAT_CLUSTERS).equals(s6));
```
kunkun-tang: Is there any assertFalse cases?

az-core/src/main/java/azkaban/Constants.java
kunkun-tang: Or, is `other_hcat_cluster`s an existing configuration now?

HappyRay: How about azkaban.job.hive.other_hcat_clusters?
Please move the existing config to this class too while you are refactoring.

HappyRay:
Please describe in more details why this config is needed? The relationship with the existing config and the plan for the existing config.

az-core/src/main/java/azkaban/utils/Props.java
herrlich: In this way, I think hive will throw an exception.

az-core/src/test/java/azkaban/utils/PropsTest.java
herrlich: Discussed offline

Issue #1481, opened 2017-09-18
Web Server: duplicate flow parameters after closing the popup.


armisael:
This is a low priority bug, but most likely an easy one to fix, and quite disturbing sometimes: basically, when you `Prepare execution`, close the popup, and open it again, overridden properties appear duplicated in the `Flow parameters` tab.

### How to reproduce it
1) Open any flow, and click `Schedule / Execute Flow`;
1) Select the `Flow parameters` tab;
1) Override a property of your choice (say `foo = bar`);
1) Run the flow
1) Once execution is completed (either with a success or with a failure), open it and click `Prepare Execution`;
1) Close the popup by either hitting `Esc` or clicking the `Cancel` button;
1) Click on `Prepare Execution` again;
1) Select the `Flow parameters` tab.

Expected behaviour: the `foo = bar` overridden parameter appears only once.
Actual behaviour: it appears twice.

HappyRay:
Thanks for reporting this issue and the detailed repro steps!

Are you interested in fixing this issue? Please see the updated dev guide for instructions.
https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md

armisael:
Hi,

I worked on this today, and managed to write a fix. I'll push it later tomorrow, after a few more checks.

I also took the chance to fix another small bug on the flow parameters input, I hope it won't be a problem if I push them together.

HappyRay:
Cool. Thanks.

I would suggest a separate PR for unrelated changes. It will increase your chance of getting a quicker review.

armisael:
Sure, I'll open another bug then, and submit two PRs. Thanks

HappyRay:
Thanks!

PR #1493, opened 2017-09-20
fix bug #1481: avoid duplicate properties when closing/reopening dialog

armisael:
Fix duplicate properties in flow execution and job edit when dialog is closed and reopened.

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/job-edit.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/job-edit.js
@@ -86,12 +86,14 @@ azkaban.JobEditView = Backbone.View.extend({
       mythis.overrideParams = overrideParams;
       mythis.generalParams = generalParams;
 
-      for (var okey in overrideParams) {
-        if (okey != 'type' && okey != 'dependencies') {
-          var row = handleAddRow();
-          var td = $(row).find('span');
-          $(td[0]).text(okey);
-          $(td[1]).text(overrideParams[okey]);
+      if (overrideParams && $(".editRow").length == 0)  {
+        for (var okey in overrideParams) {
+          if (okey != 'type' && okey != 'dependencies') {
+            var row = handleAddRow();
+            var td = $(row).find('span');
+            $(td[0]).text(okey);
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/web/js/azkaban/view/job-edit.js
```diff
@@ -86,12 +86,14 @@ azkaban.JobEditView = Backbone.View.extend({
       mythis.overrideParams = overrideParams;
       mythis.generalParams = generalParams;
 
-      for (var okey in overrideParams) {
-        if (okey != 'type' && okey != 'dependencies') {
-          var row = handleAddRow();
-          var td = $(row).find('span');
-          $(td[0]).text(okey);
-          $(td[1]).text(overrideParams[okey]);
+      if (overrideParams && $(".editRow").length == 0)  {
```
kunkun-tang: Based on what I understand, your fix in `flow-execute-dialog.js` seems to fix the flow parameter display issue in #1481. I'm wondering what this change is for. Does Job property Editing have similar issue?

armisael: Ah right! I forgot to mention it: yes, Job Editing had the same issue, I discovered it while checking the usages of the `. editRow` class. Since I was on it, I fixed it as well.

kunkun-tang: Thanks! Please fill in some description in the description panel, which will serve as the commit message.

kunkun-tang: Um, I tried but failed to reproduce the job editing bug. What scenario will generage this issue?

armisael: Apparently the bug appears only if you close the dialog by clicking outside of it; if you hit `Cancel` instead, the properties correctly appear only once.

kunkun-tang: be able to reproduce! Thanks for repairing! Could you please fill in some description? Then we are good to merge.

Issue #1492, opened 2017-09-20
Web Server: flow parameters cleared on click


armisael:
When overriding a flow parameter, clicking on the input while writing makes the value to be cleared. This makes copying flow parameter very troublesome.

### How to reproduce it
1) Open any flow, and click `Schedule / Execute Flow`;
1) Select the `Flow parameters` tab;
1) Input a property of your choice (say `foo = bar`);
1) Now click on `foo` to start editing, then click on the input again.

Expected behaviour: nothing special happens.
Actual behaviour: the input is cleared.

PR #1494, opened 2017-09-20
fix bug #1492: handle edit parameter only when clicking on span

armisael:


azkaban-web-server/src/web/js/azkaban/view/flow-execute-dialog.js
```diff
@@ -374,6 +374,8 @@ azkaban.EditTableView = Backbone.View.extend({
   },
 
   handleEditColumn: function (evt) {
+    if (evt.target.tagName == "INPUT")
```
kunkun-tang: Hi @armisael , I'm not very familiar with jQuery usage here. What's the difference between `evt.target` and `evt.currentTarget` ? 
And this looks like not a bug, as the previous authors designed this behavior to clear out test when the user clicks the "INPUT".

armisael: I'm not really familiar with jQuery as well, but as far as I understand, `evt.target` is what triggered the event (what you clicked on), while `evt.currentTarget` is the element on which the event is bound. Since the `input` is inside the `span`, when you click on the input `evt.target` points on it, while `evt.currentTarget` points to the `span`.

I don't think clearing the input was the intended behaviour of the authors: when not in "editing mode" in the DOM there is no input, only the span, containing the value of the property; when you click on the span, an input is created to allow you to edit it. The authors here decided to copy the value onto the input, and clear the span, instead of simply hiding it.

Now, since the input is created inside the span, and the event is bound on the latter, when clicking on the input the event is propagated to the parent, the trigger is fired again, and the procedure to create the input is re-executed: this time though, the span contains no value: the input is (re)created with an empty text, leading to this frustrating behaviour: you can't easily copy/paste part of the input by clicking on it.

kunkun-tang: Got it. Thanks for fixing.

PR #1500, opened 2017-09-22
Create directory flow loader for yaml files. Fixed an issue in #1466

jamiesjc:
1. In Flow 2.0 design, we want to have new flowLoader to load yaml files to azkabanFlow from project zip directory.
2. Used flowLoaderFactory to create flowLoader based on azkaban flow version, i.e. create DirectoryFlowLoader if the version is null (current implementation), and create DirectoryYamlFlowLoader if the version is 2.0 (future implementation). Guicified flowLoaderFactory.
3. Both DirectoryFlowLoader and DirectoryYamlFlowLoader extend FlowLoader class, share some common methods and implement their own logic of loading project flows. This part is still in progress.
4. Refactored some code and added some test cases.  
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Some updates:
1. Changed FlowLoader to interface and have DirectoryFlowLoader and DirectoryYamlFlowLoader implement the FlowLoader interface.
2. Moved some common methods of DirectoryFlowLoader and DirectoryYamlFlowLoader to the new FlowLoaderUtils class.
2. Fixed one potential issue of inconsistency between project cache and DB during project uploading introduced in https://github.com/azkaban/azkaban/pull/1466. If we update the project cache during loading project flows and it encounters loading errors, then we would not proceed to save project info to DB. This would cause the inconsistency issue. So we need to update project cache only after it's saved to DB.


azkaban-common/src/main/java/azkaban/project/FlowLoader.java
```diff
@@ -0,0 +1,139 @@
...
+    return report;
+  }
+
+  /**
+   * Adds email properties to a flow.
+   *
+   * @param flow the flow
+   * @param prop the prop
+   */
+  public void addEmailPropsToFlow(final Flow flow, final Props prop) {
```
jamiesjc: This part is the same. Just moved the original code to a new method.

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
```diff
@@ -0,0 +1,139 @@
...
+   * Checks job properties.
+   *
+   * @param project the project
+   */
+  public abstract void checkJobProperties(final Project project);
+
+  /**
+   * Implements Suffix filter.
+   */
+  public static class SuffixFilter implements FileFilter {
```
jamiesjc: This part is the same, just moved to a new class.

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
```diff
@@ -0,0 +1,50 @@
...
+  public static final String FLOW_NAME = "basic_flow";
+  private Project project;
+
+  @Before
+  public void setUp() {
+    this.project = new Project(12, "myTestProject");
+  }
+
+  @Test
+  public void testLoadYamlFileFromDirectory() {
```
kunkun-tang: Why it is named by `testLoadYamlFileFromDirectory` ? Does it mean it will include all `.flow` files in the directory?

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
jamiesjc: It seems not recommended to use protected static class. If needed in the future, we can move this suffixFilter to a util class.
https://stackoverflow.com/questions/24289070/why-we-should-not-use-protected-static-in-java

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -0,0 +1,59 @@
...
+import azkaban.Constants;
+import azkaban.utils.Props;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Factory class to generate flow loaders.
+ */
+@Singleton
+public class FlowLoaderFactory {
```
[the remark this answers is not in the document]
jamiesjc: Different flow loader will be created runtime based on the manifest file inside project zip. It would be difficult to guicify it in AzkabanCommonModule.

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
jamiesjc: In this test case, I just loaded one .flow file. I've added a new test case to load all .flow files from the directory.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -0,0 +1,83 @@
...
+  private void convertYamlFiles(final File projectDir) {
+    // Todo jamiesjc: convert project yaml file. It will contain properties for all flows.
+
+    //covert flow yaml files
+    final File[] flowFiles = projectDir.listFiles(new SuffixFilter(FLOW_YAML_SUFFIX));
+    for (final File file : flowFiles) {
+      final FlowBeanLoader loader = new FlowBeanLoader();
+      try {
+        final FlowBean flowBean = loader.load(file);
+        final AzkabanFlow azkabanFlow = loader.toAzkabanFlow(loader.getFlowName(file), flowBean);
```
chengren311: when do we validate the config file?

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Creates flow loader based on different flow versions.
+   *
+   * @param flowVersion the flow version
+   * @return the flow loader
+   */
+  public FlowLoader createFlowLoader(final String flowVersion) throws ProjectManagerException {
+    if (flowVersion == null) {
```
chengren311: should flowVersion 1.0 also fall into this case?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: We will validate the flow yaml file in FlowBeanLoader which loads the flowBean to azkabanFlow. azkaban.project.FlowBeanLoader#validate. This validation part will be implemented later.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -32,6 +32,9 @@
  */
 public class Constants {
 
+  // Azkaban Flow Versions
+  public static final String AZKABAN_FLOW_VERSION_20 = "2.0";
```
HappyRay: version_2? or version_2_0?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -95,13 +97,12 @@
...
       reports = validateProject(project, archive, file, prop);
 
-      // Todo jamiesjc: in Flow 2.0, we need to create new flowLoader class and
-      // call new method to load the project flows.
-      // Need to guicify it later so that we can mock flowLoader in the tests.
-      // Load the project flows.
-      final DirectoryFlowLoader directoryFlowLoader = new DirectoryFlowLoader(prop);
-      reports.put(DIRECTORY_FLOW_REPORT_KEY,
-          directoryFlowLoader.loadProject(project, file));
+      // Todo jamiesjc: need to check if manifest file exists in project zip,
```
HappyRay: Can this logic be encapsulated in loader class?

azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
```diff
@@ -302,34 +285,7 @@ private void buildFlowsFromDependencies() {
...
-                Collections.EMPTY_LIST);
-        for (String email : notifyEmailList) {
-          email = email.toLowerCase();
-          successEmail.add(email);
-          failureEmail.add(email);
-        }
-
-        flow.addFailureEmails(failureEmail);
-        flow.addSuccessEmails(successEmail);
+        addEmailPropsToFlow(flow, jobProp);
```
HappyRay: Nice.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import org.apache.log4j.Logger;
+
+/**
+ * Loads yaml files to flows from project directory.
+ */
+public class DirectoryYamlFlowLoader extends FlowLoader {
+
+  private static final Logger logger = Logger.getLogger(DirectoryYamlFlowLoader.class);
```
HappyRay: Use slf4j logger instead? See the dev guide and tips wiki.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -0,0 +1,83 @@
...
+  private static final String PROJECT_YAML_SUFFIX = ".project";
```
HappyRay: yaml is an implementation detail. Does it have to be part of the name?
project_file_suffix and flow_file_suffix?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -0,0 +1,83 @@
...
+        final Props flowProps = azkabanFlow.getProps();
+        addEmailPropsToFlow(flow, flowProps);
+      } catch (final FileNotFoundException e) {
+        logger.error("Error loading flow yaml files", e);
+      }
+    }
+  }
+
+  @Override
+  public void checkJobProperties(final Project project) {
```
HappyRay: What is this method used for?

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Loads flows from project directory.
+ */
+public abstract class FlowLoader {
```
HappyRay: Have you considered using the composition instead of inheritance see the dev guide.

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
HappyRay: Thanks for the comment.

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public boolean accept(final File pathname) {
+      final String name = pathname.getName();
+
+      return pathname.isFile() && !pathname.isHidden()
+          && name.length() > this.suffix.length() && name.endsWith(this.suffix);
+    }
+  }
+}
```
HappyRay: A new line is missing.

Did you use the saveAction plugin?

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+import azkaban.Constants;
+import azkaban.utils.Props;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Factory class to generate flow loaders.
+ */
+@Singleton
```
HappyRay: Does it need to be a singleton?

test/execution-test-data/basicflowyamltest/basic_flow.flow
```diff
@@ -0,0 +1,40 @@
...
+  # The job definition is like a YAMLified version of properties file
+  # with one major difference. All custom properties are now clubbed together
+  # in a config section in the definition.
+  # The first line describes the name of the job
+  - name: shell_end
+    # Describe the type of the job
+    type: noop
+
+    # List the dependencies of the job
+    dependsOn:
```
HappyRay: Per our earlier discussion, do you plan to change it to "dependedBy" later?

az-core/src/main/java/azkaban/Constants.java
jamiesjc: Version_2_0 seems better, I'll change it.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Good idea, I'll move this logic into the loader class.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Sure, I'll change it.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Will change, thanks!

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: It is used to check the job properties like memory size, and job callback related properties. http://azkaban.github.io/azkaban/docs/latest/#common-configurations

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
jamiesjc: We only need one FlowLoaderFactory instance, so I think it can be singleton. Any reason not to use singleton here?

test/execution-test-data/basicflowyamltest/basic_flow.flow
jamiesjc: Yes, I'll change it to "dependedBy" in a separate PR later.

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
jamiesjc: I've enabled saveAction, but it doesn't seem to fix this automatically. I'll add the new line manually.

azkaban-common/src/main/java/azkaban/project/FlowLoader.java
jamiesjc: I've thought about using composition. However, I don't think it will benefit from using composition in this case. Flowloader does not actually have any major implementation of loading project flows. So even if we add it as a reference inside DirectoryFlowLoader and DirectoryYamlFlowLoader, it doesn't make any difference. I think the better way to do it would be to change FlowLoader to an interface with one loadProject method inside. Then DirectoryFlowLoader and DirectoryYamlFlowLoader can extend the interface and implement their own loadProject method. We can move some common method like addEmailPropsToFlow to some util class. What do you think of the new approach? We can discuss it offline.

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
HappyRay: See https://github.com/google/guice/wiki/Scopes
"Singletons are popular in Java applications but they don't provide much value, especially when dependency injection is involved. Although singletons save object creation (and later garbage collection), initialization of the singleton requires synchronization; getting a handle to the single initialized instance only requires reading a volatile. Singletons are most useful for:

stateful objects, such as configuration or counters
objects that are expensive to construct or lookup
objects that tie up resources, such as a database connection pool."

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -95,13 +97,11 @@
 
       reports = validateProject(project, archive, file, prop);
```
HappyRay: If there is validation error, is it safe to load and save the project?

jamiesjc: We want to return as many validation results as possible to users, including the errors found in loading project. If there is any error during validation or loading project, we will just return the report and not save it to DB. So I think this should be safe.

Issue #1502, opened 2017-09-22
Azkaban-solo-server fails to start due to missing hadoop dependencies


inramana:
Due to #1499 we stopped including hadoop dependencies as part of azkaban-solo-server.
This results in a failure to start for the solo server with the following error:

`Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.getDeclaredMethods(Class.java:1975)
	at com.google.inject.internal.ProviderMethodsModule.getProviderMethods(ProviderMethodsModule.java:132)
	at com.google.inject.internal.ProviderMethodsModule.configure(ProviderMethodsModule.java:123)
	at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:340)
	at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:349)
	at com.google.inject.spi.Elements.getElements(Elements.java:110)
	at com.google.inject.internal.InjectorShell$Builder.build(InjectorShell.java:138)
	at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:104)
	at com.google.inject.Guice.createInjector(Guice.java:99)
	at com.google.inject.Guice.createInjector(Guice.java:73)
	at com.google.inject.Guice.createInjector(Guice.java:62)
	at azkaban.soloserver.AzkabanSingleServer.main(AzkabanSingleServer.java:75)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.conf.Configuration
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 14 more`

PR #1499, opened 2017-09-21
build: change hadoop gradle dependencies to compileOnly

inramana:
## Issue
#1498 
With change #1069 we introduced hadoop dependencies in azkaban-common/build.gradle which resulted in hadoop jars being present in the lib directory. 
Also with change #1332 we introduced transitive hadoop dependencies through the gobblinKafka dependency. 
When testing against different hadoop versions we noticed azkaban was pulling in its own hadoop jars into the distribution instead of using what's available on the hadoop cluster.

## Fix
For change #1069 
Gradle has a [compileOnly](https://blog.gradle.org/introducing-compile-only-dependencies) dependency option which is a true compile only dependency as opposed to compile which extend into runtime.
We also needed to add the same dependencies as a testCompile dependency tests. 

For change #1332 
we made two changes, the compile has an exclude condition thereby excluding any transitive hadoop dependencies. Similar to the previous change we also needed to add the dependencies as testCompile 

Verified manually that the distribution lib does not contain hadoop jars.

PR #1503, opened 2017-09-22
Fix azkaban-solo-server by including hadoop dependencies in build.gradle

inramana:
## Issue
#1502
With #1499 we broke azkaban-solo-server due to missing hadoop dependencies in solo server which were being transitively included.
## Fix
Include hadoop dependencies for azkaban-solo-server
Verified by bringing up azkaban-solo-server locally and running flows against it.


HappyRay:
For simple PRs, a separate issue is not required. It's recommended to have an issue if a fix requires multiple PRs.

PR #1504, opened 2017-09-26
Fixing regression where JOB_OUTPUT_PROP_FILE was no longer writable by user process

reallocf:
JOB_OUTPUT_PROP_FILE is a file used to pass properties between jobs. You can read about it here: http://azkaban.github.io/azkaban/docs/latest/#job-configuration

#1325 introduced a regression where JOB_OUTPUT_PROP_FILE became no longer writable due to an upgraded execute-as-user executable that stopped user processes from having the azkaban group permission. This file needs to be written to so properties can be passed from one job to another.

This fix assigns ownership of JOB_OUTPUT_PROP_FILE and JOB_PROP_FILE to `<username>:azkaban` to guarantee user ability to write to this file. It is uncertain whether or not JOB_PROP_FILE is intended to be written to, but I believe it's a good idea to do allow it to be in order to prevent regressions.

This is covered by an integration test that simply tries to write to the JOB_OUTPUT_PROP_FILE to confirm that it is doable. This is already in master of azkaban-auto-test.

This change will more drastically take effect when the execute-as-user executable is upgraded to its newest version.

This was tested on solo-server and holdem4.

I'll continue to explore JOB_OUTPUT_PROP_FILE and JOB_PROP_FILE usage and implementation so we can make sure these features are fully covered by tests and can be leveraged successfully by our users.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -387,7 +387,7 @@ public class ProcessJob extends AbstractProcessJob {
 
   /**
    * Changes permissions on file/directory so that the file/directory is owned by the user and
-   * the group remains azkaban.
+   * the group remains the azkaban service account name.
    *
    * Leverages execute-as-user with "root" as the user to run the command.
    *
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -243,7 +243,14 @@ public class ProcessJob extends AbstractProcessJob {
       // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
       // if the directory is not permissioned correctly already (should happen once per execution)
       if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
-        assignUserDirOwnership(effectiveUser);
+        info("Changing current working directory ownership");
+        assignUserFileOwnership(effectiveUser, getWorkingDirectory());
+      }
+      // Set property file permissions to <uid>:azkaban so user can write to their prop files
+      // in order to pass properties from one job to another
+      for (final File propFile : propFiles) {
+        info("Changing properties files ownership");
+        assignUserFileOwnership(effectiveUser, propFile.getAbsolutePath());
       }
     }
 
@@ -379,20 +386,22 @@ public class ProcessJob extends AbstractProcessJob {
   }
 
   /**
-   * Changes permission on current working directory so that the directory is owned by the user and
+   * Changes permissions on file/directory so that the file/directory is owned by the user and
    * the group remains azkaban.
    *
    * Leverages execute-as-user with "root" as the user to run the command.
    *
    * @param effectiveUser user/proxy user running the job
+   * @param fileName the name of the file whose permissions will be changed
    */
-  private void assignUserDirOwnership(final String effectiveUser) throws Exception {
+  private void assignUserFileOwnership(final String effectiveUser, final String fileName) throws
+      Exception {
     final ExecuteAsUser executeAsUser = new ExecuteAsUser(
         this.sysProps.getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
     final String groupName = this.sysProps.getString(AZKABAN_SERVER_GROUP_NAME, "azkaban");
     final List<String> changeOwnershipCommand = Arrays
-        .asList(CHOWN, effectiveUser + ":" + groupName, getWorkingDirectory());
-    info("Change current working directory ownership to " + effectiveUser + ":" + groupName + ".");
+        .asList(CHOWN, effectiveUser + ":" + groupName, fileName);
+    info("Change ownership of " + fileName + " to " + effectiveUser + ":" + groupName + ".");
     final int result = executeAsUser.execute("root", changeOwnershipCommand);
     if (result != 0) {
       handleError("Failed to change current working directory ownership. Error code: " + Integer


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -243,7 +243,14 @@ public void run() throws Exception {
       // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
       // if the directory is not permissioned correctly already (should happen once per execution)
       if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
-        assignUserDirOwnership(effectiveUser);
+        info("Changing current working directory ownership");
+        assignUserFileOwnership(effectiveUser, getWorkingDirectory());
```
kunkun-tang: I'm seeing the method `getWorkingDirectory` requires a public modifier, but `protected` should be good enough.

reallocf: `getWorkingDirectory` is legacy code (last updated in 2012) and I'm concerned there might be plugins or something that use it. So I'd rather not update it in this PR. Does that make sense?

kunkun-tang: I see!

HappyRay:
>   became no longer writable due to an upgraded execute-as-user executable that stopped user processes from having the azkaban group permission.

A bit more detail will be useful. e.g. the file is written to by the job process. Since the file can only be written to .... It worked before because ....

HappyRay:
> It is uncertain whether or not JOB_PROP_FILE is intended to be written to, but I believe it's a good idea to do allow it to be in order to prevent regressions.

I am pretty confident based on the doc that this file is not supposed to be written to by the user job process. Given that all the files are currently readable by "others", I don't think we need to change the permission of this file. We can discuss in person.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -379,20 +386,22 @@ private boolean canWriteInCurrentWorkingDirectory(final String effectiveUser)
   }
 
   /**
-   * Changes permission on current working directory so that the directory is owned by the user and
+   * Changes permissions on file/directory so that the file/directory is owned by the user and
    * the group remains azkaban.
```
HappyRay: remains the azkaban service account name?

reallocf:
@HappyRay updated my comment to provide a bit more information and to say that I'll be continuing to explore these features.

HappyRay:
LGTM. Please update a comment in the code too.

kunkun-tang:
Thanks @reallocf ! JOB_OUTPUT_PROP_FILE and JOB_PROP_FILE usage are very important to my Security work as well.

reallocf:
Looks like I need another ship it @HappyRay. Is that due to the new review requirement?

Thanks!

prokod:
Hi @reallocf I have tested 3.36.0 on real flow and unfortunately it blows on before start of second job in the flow because the permissions of the working dir were changed from `azkaban:azkaban` to `user:azkaban` on the first job.  
Now it blows when it tries to creat a log file for the second job. Something like this:  
```
java.io.FileNotFoundException: /opt/azkaban-exec-server-3.36.0/executions/28/forwardFlow/_job.28.preprocessing-middomain-forward_calcCheck.log (Permission denied)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.&lt;init&gt;(FileOutputStream.java:213)
	at java.io.FileOutputStream.&lt;init&gt;(FileOutputStream.java:133)
	at org.apache.log4j.FileAppender.setFile(FileAppender.java:294)
	at org.apache.log4j.RollingFileAppender.setFile(RollingFileAppender.java:207)
	at org.apache.log4j.FileAppender.&lt;init&gt;(FileAppender.java:110)
	at org.apache.log4j.RollingFileAppender.&lt;init&gt;(RollingFileAppender.java:79)
	at azkaban.execapp.JobRunner.createFileAppender(JobRunner.java:317)
	at azkaban.execapp.JobRunner.createLogger(JobRunner.java:266)
	at azkaban.execapp.JobRunner.doRun(JobRunner.java:569)
	at azkaban.execapp.JobRunner.run(JobRunner.java:552)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
```

I have fixed this on my local copy of the azkaban project by including in the `finally` block `assignUserFileOwnership` call with user 'azkaban', essentially reverting ownership. If you would like I can submit an MR for that, but it is simple as:
```
} finally {
        info("Reverting current working directory ownership");
        assignUserFileOwnership("azkaban", getWorkingDirectory());
        info("Process completed "
            + (this.success ? "successfully" : "unsuccessfully") + " in "
            + ((System.currentTimeMillis() - startMs) / 1000) + " seconds.");
      }
```
in `ProcessJob#run`

reallocf:
Hey @prokod, that does make sense but I'm surprised that you have to do it..

The "azkaban" user should still have read/write access on the current working directory because it should still be owned by the "azkaban" group. Are you not seeing that? How is the "azkaban" process permissioning the exec directories/files inside of them?

prokod:
I have it set up as follows
**In systemd I start the exec server like this:**
```
[Unit]
Description=Azkaban executor service
...
[Service]
Type=forking
User=azkaban
WorkingDirectory=/opt/azkaban-exec-server-3.36.0
...
ExecStart=/opt/azkaban-exec-server-3.36.0/bin/azkaban-executor-start.sh
...
```
The owner of the exec service is `azkaban` when `ps`ing  
  
**The permissioning the exec directories/files inside the failed execution:**
First time created under `executions` the permissioning is `azkaban:azkaban`  
After `projBuildInfo` job is finished the working directory looks like this:  
```
$ pwd
/opt/azkaban-exec-server-3.36.0/executions/28
$ ls -la forwardFlow/
total 48
drwxr-sr-x. 2 scheduler azkaban 4096 Oct  4 21:59 .
drwxr-sr-x. 3 azkaban   azkaban 4096 Oct  4 22:42 ..
-rw-r--r--. 1 azkaban   azkaban 4035 Oct  4 21:59 _job.28.preprocessing-middomain-forward_projBuildInfo.log
-rw-r--r--. 3 azkaban   azkaban  783 Oct  4 21:59 preprocessing-middomain-forward_calcCheck.job
-rw-r--r--. 3 azkaban   azkaban  489 Oct  4 21:59 preprocessing-middomain-forward_common.properties
-rw-r--r--. 3 azkaban   azkaban  888 Oct  4 21:59 preprocessing-middomain-forward_copyToAms.job
-rw-r--r--. 3 azkaban   azkaban  133 Oct  4 21:59 preprocessing-middomain-forward.job
-rw-r--r--. 3 azkaban   azkaban  651 Oct  4 21:59 preprocessing-middomain-forward_midDomainShuffle.job
-rw-r--r--. 3 azkaban   azkaban  262 Oct  4 21:59 preprocessing-middomain-forward_projBuildInfo.job
-rw-r--r--. 1 scheduler azkaban    0 Oct  4 21:59 preprocessing-middomain-forward_projBuildInfo_output_6921027372125729455_tmp
-rw-r--r--. 1 scheduler azkaban 4040 Oct  4 21:59 preprocessing-middomain-forward_projBuildInfo_props_399184297853803228_tmp
-rw-r--r--. 3 azkaban   azkaban  630 Oct  4 21:59 preprocessing-middomain-forward_results2hive.job
-rw-r--r--. 3 azkaban   azkaban  365 Oct  4 21:59 preprocessing-middomain-forward_writeMetadata.job
```

Do you see something wrong with the setup? Any other pointers?

reallocf:
Yes, I do see a difference.

In our executions directory, the generated files are created with the following permissions:

```
-rw-rw-r--. 4 user azkaban 5.0K Oct 5 16:28 FILE-NAME_props_2186224269708212984_tmp
```
So the azkaban group also has write permissions. This is important both for writing the log files and also for cleanup afterwards.

I'm not exactly sure how that's configured because I don't think we do anything special with the Azkaban process to do that.. I imagine it's a system setting. Maybe @li-afaris can speak to it?

prokod:
Ok. so setting `umask 002` did the trick for me.
Thx.

PR #1505, opened 2017-09-26
Flow Trigger Dependency Interface

chengren311:
related feature request: #1526
This PR provides skeleton for building dependency check.

Plugin needs to implement two interfaces: DependencyCheck and DependencyInstanceContext for their custom data dependency type.


azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
```diff
@@ -0,0 +1,46 @@
...
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+public interface DependencyCheck {
+
+
+  /**
+   * run this dependency check asynchronously return immediately when the check starts
```
HappyRay: Just say the call should not block?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  DependencyInstanceContext run(DependencyInstanceConfig config);
+
+  /**
+   * kill the dependency instance
+   */
+  void kill(DependencyInstanceContext depContext);
+
+  /**
+   * shutdown the dependency instance. Clean up resource if needed.
```
HappyRay: Shutdown the DependencyCheck plugin?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  void shutdown();
+
+  /**
+   * Initialize the dependency instance.
```
HappyRay: plugin not instance?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class DependencyInstanceConfig {
```
HappyRay: Why isn't it an interface?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceContext.java
```diff
@@ -0,0 +1,25 @@
...
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+/**
+ * This class holds the context info
+ * for a running dependency instance.
+ */
+public class DependencyInstanceContext {
```
HappyRay: Why isn't it an interface?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyPluginConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class DependencyPluginConfig {
```
HappyRay: Why isn't it an interface?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceConfig.java
chengren311: I don't see there will be multiple implementations for DependencyInstanceConfig.

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceContext.java
chengren311: I still haven't thought about the implementation detail yet. Will create an interface once I see a need.

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyPluginConfig.java
chengren311: I don't see there will be multiple implementations for it.

HappyRay: Since we may not have direct control over the plugin code, using an interface gives us more flexibility to evolve this type.

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyCheck.java
chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyPluginConfig.java
chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceContext.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+/**
+ * This interface provides getExecId() method
```
kunkun-tang: Why these java doc lines are short compared to other interfaces.

chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyPluginConfig.java
```diff
@@ -0,0 +1,22 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+public interface DependencyPluginConfig {
+
+  String get(final String key);
```
HappyRay: What's the behavior when the key doesn't exist?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceContext.java
```diff
@@ -0,0 +1,28 @@
...
+ * to get execution id for a running dependency instance.
+ * The implementation classes should hold context
+ * information for a running dependency instance.
+ */
+public interface DependencyInstanceContext {
+
+  int getExecId();
```
HappyRay: How is this id used?

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyPluginConfig.java
chengren311: it will return null.

azkaban-web-server/src/main/java/azkaban/flowtrigger/DependencyInstanceContext.java
chengren311: it's used to identify a dependency instance. Just like flow execution id.

chengren311:
@HappyRay do you have other feedback?

juhoautio:
I know DependencyInstanceConfig is an interface on purpose, but is there anything special compared to job Props? Could it be a more general interface (AzkabanProps?) that would have all the nice public methods that the Props class currently provides?

kunkun-tang:
@juhoautio This is a good question. Any insights? @chengren311

chengren311:
@juhoautio @kunkun-tang Props is overly used everywhere. In this case we are trying to remove the dependency on Props for few reasons: 1. Props is too complicated for the user case here. Props's main purpose is for inherited properties, but here we just want to key value map. 2. DependencyInstanceConfig will be used in external plugins. If they depends on Props, it will make us hard to evolve Props while keeping compatibility .

kunkun-tang:
Understand more now, but still don't get the difference between `DependencyInstanceConfig` and `DependencyPluginConfig`. Can they be consolidated by one Map?

juhoautio:
My question was more like: could you make an interface that has the convenient accessors that Props currently provides and use that interface here. Eventually there could be a new Job API that also uses it instead of the concrete Props class.

HappyRay:
@juhoautio could you give examples of the convenient accessors you think will be useful?

juhoautio:
These are really basic & generally useful:
```java
String getString(final String key, final String defaultValue)
boolean getBoolean(final String key, final boolean defaultValue)
boolean getBoolean(final String key)
long getLong(final String name, final long defaultValue)
long getLong(final String name)
int getInt(final String name, final int defaultValue)
int getInt(final String name)
double getDouble(final String name, final double defaultValue)
double getDouble(final String name)
```

We have also used this a lot but I think it would be more convenient for this to return some kind of "sub-config" object instead:
```java
Map<String, String> getMapByPrefix(final String prefix)
```

Then there are methods like these that can be useful sometimes but I'm not sure if they deserve to be there or not:

```java
public int size()
List<String> getStringList(final String key)
List<String> getStringList(final String key, String separator..
```

HappyRay:
I understand. I prefer having many focused smaller composable classes/interfaces.

Implementing these convenient methods on top of a key value string map like interface would be simple.

If there is enough demand, we can create a shared common library if such a library doesn't already exist in the open source world.

Issue #1508, opened 2017-09-27
Graph visualisation breaks in some rare cases


armisael:
## How to reproduce
1) Create a project starting from this [_project.zip](https://github.com/azkaban/azkaban/files/1337700/_project.zip) (it's one of our production flows, I had to obfuscate job names, sorry for that);
1) Open the job named `wwww-wwwww-wwwww`
1) Expand the nested flow `ccccc-cc-ccccc-ccccccc`

This appears:
![screenshot 2017-09-27 17 05 08](https://user-images.githubusercontent.com/71387/30921141-08d3d84a-a3a6-11e7-88ea-8ca0b57a9210.png).

I have a fix ready for this, I will push it soon.

HappyRay:
Thanks.

PR #1511, opened 2017-09-28
fix bug #1508: fix graph visualisation when exploding nested flows

armisael:
The broken graph attached to the bug #1508 happens because `animatePolylineEdge` fails: it assumes that each `edge` has always a number of `edge.oldpoints` equals to the number of `newPoints`, which sometimes fails.

`newPoints` is fed by the `moveNodeEdges`: it iterates through all the `edge.guides` (the segments composing the lines connecting the graph nodes) and adds the starting and ending points; this makes sense, because the assumption `edge.guides.length == edge.points.length - 2` should always hold.

The bug happens because this assumption is broken: in `layoutGraph` we sometimes add a new segment to `edge.guides` but we forget to update the points accordingly.

This patch fixes the problem: when a new segment is added to an edge (see the code after the `// Add gap` comment) we also add an extra point, therefore assuring the assumption holds, and allowing the `animatePolylineEdge` to work properly.

The result:
screenshot 2017-09-28 09 41 53

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/util/layout.js
+++ b/azkaban-web-server/src/web/js/azkaban/util/layout.js
@@ -212,22 +212,28 @@ function layoutGraph(nodes, edges, hmargin) {
       var destX = nodeMap[dest].x;
 
       var guides = [];
+      var pointArray = new Array();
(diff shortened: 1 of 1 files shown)


HappyRay:
@armisael Thanks!

Issue #1506, opened 2017-09-26
Rename ajax action loadFlow to fetchSchedules and document it


juhoautio:
On the ScheduleServlet there's an ajax action [loadFlow](https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L93-L94) that returns a list of all schedules. [It doesn't take any parameters](https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L108-L132).

The name is really misleading. I suggest renaming it to `fetchSchedules` (we can also keep the old `loadFlow` for backward compatibility).

Then it would be good to document it in http://azkaban.github.io/azkaban/docs/latest/#ajax-api as new sub-section `Fetch all schedules`.

kunkun-tang:
Sounds good! How about `fetchAllSchedules` ?

HappyRay:
If it is not documented, it is considered internal, probably to support the UI.

I would suggest updating the name and callers. I suspect it is called by some js files. 

We would like to improve our API in the near future. I hesitate to expand our public API set now.

juhoautio:
> How about fetchAllSchedules ?

I thought that `fetchSchedules` would be nicer because you can assume that without any additional params it returns everything. And then there's room for expansion: it would be easy to add support for some filter parameters like project id, and the name `fetchSchedules` would still work well.

juhoautio:
> I would suggest updating the name and callers

I'd like that. Want me to make a PR?

wyukawa:
I use loadFlow API to get all schedules, but actually I think that the name is misleading...

kunkun-tang:
@juhoautio Sounds good. Filter can be an option.

HappyRay:
I'd welcome a PR.

juhoautio:
It's been renamed to `fetchSchedules` but `loadFlow` is still kept as an alias for compatibility.

PR #1515, opened 2017-09-28
Rename ajax loadFlow to fetchSchedules

juhoautio:
Because that's what it does.

Keep alternative ajax action name 'loadFlow' for backward compatibility.

This resolves issue #1506.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
@@ -90,8 +90,9 @@ public class ScheduleServlet extends LoginAbstractAzkabanServlet {
       ajaxSlaInfo(req, ret, session.getUser());
     } else if (ajaxName.equals("setSla")) {
       ajaxSetSla(req, ret, session.getUser());
-    } else if (ajaxName.equals("loadFlow")) {
-      ajaxLoadFlows(req, ret, session.getUser());
+    // alias loadFlow is preserved for backward compatibility
+    } else if (ajaxName.equals("fetchSchedules") || ajaxName.equals("loadFlow")) {
+      ajaxFetchSchedules(ret);
     } else if (ajaxName.equals("scheduleFlow")) {
       ajaxScheduleFlow(req, ret, session.getUser());
     } else if (ajaxName.equals("scheduleCronFlow")) {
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -90,8 +90,8 @@ private void handleAJAXAction(final HttpServletRequest req,
       ajaxSlaInfo(req, ret, session.getUser());
     } else if (ajaxName.equals("setSla")) {
       ajaxSetSla(req, ret, session.getUser());
-    } else if (ajaxName.equals("loadFlow")) {
```
chengren311: The change itself makes sense, but my concern is it's not backward compatible.

juhoautio: Yes. I suggested making it backward compatible but understood from @HappyRay's comment that he'd prefer a "blunt rename" :) ..it would be as simple as this:

```java
// alias loadFlow is preserved for backward compatibility
} else if (ajaxName.equals("fetchSchedules") || ajaxName.equals("loadFlow")) {
```

Should I change to this?

kunkun-tang: +1 on this.

juhoautio: Oh, this comment is "on the wrong side" so it won't go away :) But i have updated it, see the diff, please.

juhoautio:
The test failure seems to be coming from master:

```
azkaban.execapp.FlowRunnerTest > exec1FailedKillAll FAILED
    java.lang.AssertionError: Flow didn't reach expected status
        at azkaban.execapp.FlowRunnerTestBase.waitFlowRunner(FlowRunnerTestBase.java:79)
        at azkaban.execapp.FlowRunnerTestBase.waitFlowRunner(FlowRunnerTestBase.java:63)
        at azkaban.execapp.FlowRunnerTestBase.assertThreadShutDown(FlowRunnerTestBase.java:45)
        at azkaban.execapp.FlowRunnerTest.exec1FailedKillAll(FlowRunnerTest.java:186)
```

This PR will fix it: https://github.com/azkaban/azkaban/pull/1518

juhoautio:
@chengren311 @kunkun-tang the build is passing, as expected..

kunkun-tang:
Before merging, want to double check. Looks like front-end code never calls this ajax method? @juhoautio

juhoautio:
Front-end doesn't call it.

I'm _not_ planning to add documentation at this point, becaue @HappyRay commented in the issue:

> We would like to improve our API in the near future. I hesitate to expand our public API set now.

PR #1516, opened 2017-09-28
create hadoop module

kunkun-tang:
#1499 makes Hadoop dependencies compileOnly, which caused web-server and exec-server can not initialize Guice successfully if users don't specify their own Hadoop jars in the classpath.
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.getDeclaredMethods(Class.java:1975)
	at com.google.inject.internal.ProviderMethodsModule.getProviderMethods(ProviderMethodsModule.java:132)
	at com.google.inject.internal.ProviderMethodsModule.configure(ProviderMethodsModule.java:123)
	at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:340)
	at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:349)
	at com.google.inject.spi.Elements.getElements(Elements.java:110)
	at com.google.inject.internal.InjectorShell$Builder.build(InjectorShell.java:138)
	at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:104)
	at com.google.inject.Guice.createInjector(Guice.java:99)
	at com.google.inject.Guice.createInjector(Guice.java:73)
	at com.google.inject.Guice.createInjector(Guice.java:62)
	at azkaban.soloserver.AzkabanSingleServer.main(AzkabanSingleServer.java:75)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.conf.Configuration
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 14 more
```

This patch creates a new HadoopModule, which places Hadoop-related dependencies. we only install this module when Hadoop related injection is needed. Today only HDFS_Storage needs it.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -16,10 +16,6 @@
  */
 package azkaban;
 
-import static azkaban.Constants.ConfigurationKeys.HADOOP_CONF_DIR_PATH;
-import static com.google.common.base.Preconditions.checkArgument;
-import static java.util.Objects.requireNonNull;
-
 import azkaban.db.AzkabanDataSource;
 import azkaban.db.H2FileDataSource;
 import azkaban.db.MySQLDataSource;
@@ -27,23 +23,15 @@ import azkaban.executor.ExecutorLoader;
 import azkaban.executor.JdbcExecutorLoader;
 import azkaban.project.JdbcProjectImpl;
 import azkaban.project.ProjectLoader;
-import azkaban.spi.AzkabanException;
 import azkaban.spi.Storage;
 import azkaban.spi.StorageException;
-import azkaban.storage.HdfsAuth;
 import azkaban.storage.StorageImplementationType;
 import azkaban.trigger.JdbcTriggerImpl;
 import azkaban.trigger.TriggerLoader;
 import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
-import java.io.File;
-import java.io.IOException;
-import javax.inject.Inject;
-import javax.inject.Singleton;
 import org.apache.commons.dbutils.QueryRunner;
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FileSystem;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -78,6 +66,9 @@ public class AzkabanCommonModule extends AbstractModule {
   public Class<? extends Storage> resolveStorageClassType() {
     final StorageImplementationType type = StorageImplementationType
         .from(this.config.getStorageImplementation());
+    if (type == StorageImplementationType.HDFS) {
+      install(new HadoopModule(this.props));
+    }
     if (type != null) {
       return type.getImplementationClass();
     } else {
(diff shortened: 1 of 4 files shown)


HappyRay:
Please clarify in the change description that this is only a problem if users don't supply their own Hadoop jars in the classpath.

How did you test?
Is this code path covered in an automated test, unit test?

HappyRay:
Thanks for linking the related changes.

kunkun-tang:
Description Updated. Unit Test should not cover HDFS coverage, and I evaluated the generated jar in staging cluster, and it works. @HappyRay

HappyRay:
Any reason this can't be unit tested?

kunkun-tang:
Yes, it should be unit testable. Added another method by providing Hdfs related props to test Hadoop Injections. @HappyRay

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -78,6 +66,9 @@ protected void configure() {
   public Class<? extends Storage> resolveStorageClassType() {
     final StorageImplementationType type = StorageImplementationType
         .from(this.config.getStorageImplementation());
+    if (type == StorageImplementationType.HDFS) {
+      install(new HadoopModule(this.props));
```
chengren311: why should we install here? this seems irrelevant to this method's intention.

kunkun-tang: I was considering this question where we should put installModule, but can not find a good approach. First off, install module cannot be put under AzkabanServer class when injector is initialized, because Guice.createInjector can not optionally install a module. The second alternative location is inside `configure` method in AzkabanCommonModule. It will incur duplicate code to resolve storage type, which is same logic inside the method `resolveStorageClassType`. The final alternative is just here. Only when we provide HDFSStorage class, we install this module. I have some idea how we can refactor HDFSStorage bindings, but it will incur a big code change, and we can do it in future.

azkaban-common/src/test/java/azkaban/ServiceProviderTest.java
```diff
@@ -71,7 +76,34 @@ public void testInjections() throws Exception {
     assertThat(injector.getInstance(StorageManager.class)).isNotNull();
     assertThat(injector.getInstance(DatabaseStorage.class)).isNotNull();
     assertThat(injector.getInstance(LocalStorage.class)).isNotNull();
+    assertThatThrownBy(
```
HappyRay: Is this assert necessary?

azkaban-common/src/test/java/azkaban/ServiceProviderTest.java
```diff
@@ -71,7 +76,34 @@ public void testInjections() throws Exception {
...
+        () -> injector.getInstance(HdfsStorage.class))
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessageContaining("Guice configuration errors");
+
     assertThat(injector.getInstance(Storage.class)).isNotNull();
     assertThat(injector.getInstance(DatabaseOperator.class)).isNotNull();
   }
+
+  @Test
+  public void testHadoopInjection() throws Exception {
```
HappyRay: Thanks for adding a test.

azkaban-common/src/test/java/azkaban/ServiceProviderTest.java
kunkun-tang: The ideal test should test out if runtime needs to resolve hadoop jars. However, it cannot be straightforwardly tested. This test can guarantee that Hadoop module is not installed, so I think it has values.

PR #1517, opened 2017-09-29
Revert "Event reporting for azkaban events. (#1332)"

rajagopr:
This reverts commit be6972a146ab769fa535a1788f6cc82c0c7b87ce.

Reason: Unable to release within LinkedIn due to licensing issues emerged from using gobblin-library.


HappyRay:
In case you haven't considered this option, you could have reverted the default event listener implementation part only. This will unblock the release also but has the benefit of reducing the chance of a conflict when you introduce it again. 
It's one reason I would encourage breaking up a large change into smaller ones.

rajagopr:
@HappyRay - I did notice your point. However, unblocking the release sounded pretty urgent and 'revert all' seemed the quick thing to do.

HappyRay:
Got it. Make sense. Thanks.

PR #1509, opened 2017-09-27
Bring back "New status: KILLING (#1172)"

juhoautio:
Originally added in PR https://github.com/azkaban/azkaban/pull/1172m which was then reverted by https://github.com/azkaban/azkaban/pull/1300.

Now bringing the KILLING feature back.

This reverts commit 92aff731f4e852674154c2a2d2ebf7bb1d9c03ec

After `git revert 92aff731f4e852674154c2a2d2ebf7bb1d9c03ec`, resolved conflicts in:
	azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
	azkaban-common/src/test/java/azkaban/executor/InteractiveTestJob.java
	azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
	azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
	azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/historypage.vm
	azkaban-web-server/src/web/js/azkaban/view/exflow.js

PR #1518, opened 2017-09-29
Fix bug in killed quick finish

juhoautio:
If a job is actually not run at all, but it has been marked as killed, the right thing to do is set the status to KILLED. Otherwise the flow would get stuck because this job would send the JOB_FINISHED event with the unfinished status: KILLING.

This was caught by a unit test on a failed Travis. Details can be found in https://github.com/azkaban/azkaban/pull/1509#issuecomment-333196277.


juhoautio:
Really sorry for introducing this bug in the first place. I should've checked this more carefully.. I guess it was so rare that it was never revealed by the tests before, but I'm glad it was caught this time!

PR #1520, opened 2017-09-30
fix upload package bug

acyouzi:
when i upload a zip format package and the file extension is not zip. This will cause the database to generate an incorrect record in the project_versions table. 

Code for determining file type in azkaban.webapp.servlet.ProjectManagerServlet # ajaxHandleUpload is:

    if (contentType != null
          && (contentType.startsWith(APPLICATION_ZIP_MIME_TYPE)
          || contentType.startsWith("application/x-zip-compressed") || contentType
          .startsWith("application/octet-stream"))) {
        type = "zip";
      }

But when inserting project_versions only according to the file extension to determine the type (azkaban.project.JdbcProjectImpl#addProjectToProjectVersions)

    final String INSERT_PROJECT_VERSION = "INSERT INTO project_versions "
        + "(project_id, version, upload_time, uploader, file_type, file_name, md5, num_chunks, resource_id) values "
        + "(?,?,?,?,?,?,?,?,?)";

    try {
      /*
       * As we don't know the num_chunks before uploading the file, we initialize it to 0,
       * and will update it after uploading completes.
       */
      transOperator.update(INSERT_PROJECT_VERSION, projectId, version, updateTime, uploader,
          Files.getFileExtension(localFile.getName()), localFile.getName(), md5, 0, resourceId);
    } catch (final SQLException e) {
      final String msg = String
          .format("Error initializing project id: %d version: %d ", projectId, version);
      logger.error(msg, e);
      throw new ProjectManagerException(msg, e);
    }

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
@@ -112,7 +112,7 @@ class AzkabanProjectLoader {
     }
 
     // Upload the project to DB and storage.
-    persistProject(project, loader, archive, uploader);
+    persistProject(project, loader, archive, uploader, fileType);
 
     // Clean up project old installations after new project is uploaded successfully.
     cleanUpProjectOldInstallations(project);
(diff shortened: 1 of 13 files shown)


chengren311:
@acyouzi thanks for the PR. IMO, we should enforce package must have "zip" as the extension. This will make things easier.

PR #1523, opened 2017-10-03
schema change for execution_jobs

chengren311:
1. add flow_id to key set of execution_jobs table
2. drop two unnecessary indexes.

Previously primary key set of execution_jobs is (execution_id, job_id, attempt).
This is under the assumption where no job shares the same name within a flow. However this is not the case when embedded flow comes into picture, which allows user to embed a smaller flow into multiple places inside a bigger flow.

with new keyset, two indexes becomes a subset of the key set, so we can drop them.

### DIFF
--- a/azkaban-db/src/main/sql/create.execution_jobs.sql
+++ b/azkaban-db/src/main/sql/create.execution_jobs.sql
@@ -11,12 +11,8 @@ CREATE TABLE execution_jobs (
   input_params  LONGBLOB,
   output_params LONGBLOB,
   attachments   LONGBLOB,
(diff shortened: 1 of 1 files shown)


juhoautio:
@chengren311 & @kunkun-tang: shouldn't this have been done as a "schema migration" instead? Something like this: 
https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/test/resources/sql/update.execution_jobs.2.1.sql

How are users expected to make any required DB changes when upgrading to a newer version?

HappyRay:
The existing schema migration method appears to be broken already.

I asked @kunkun-tang to come up with a new strategy.

PR #1524, opened 2017-10-04
Event reporting for azkaban events - No default implementation

rajagopr:
**Summary:**
With this change, support for reporting events of interest from azkaban has been provided.
Users should provide an implementation of the AzkabanEventReporter interface.
To begin with, the following events are reported.

 1. FLOW_STARTED
 2. FLOW_FINISHED
 3. JOB_STARTED
 4. JOB_FINISHED

In future, this can be easily extended to report other events.

**Configuration changes:**
Note: All changes must be applied to the executor server or the solo-server if in solo mode.

// Property is used to enable/disable event reporting with the default being false.
azkaban.event.reporting.enabled=false

// Implementations of the reporter to be specified using this property.
azkaban.event.reporting.class=com.foo.EventReporterImpl

// Kafka topic name for the default implementation.
azkaban.event.reporting.kafka.topic=TestTopicName

// Kafka broker list for the default implementation.
azkaban.event.reporting.kafka.brokers=hostname.com:port_num

// Schema registry server for the default kafka implementation.
azkaban.event.reporting.kafka.schema.registry.url=schemaRegistryUrl.com:port/schema

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -127,6 +127,17 @@ public class Constants {
     public static final String AZKABAN_KEYTAB_PATH = "azkaban.keytab.path";
     public static final String PROJECT_TEMP_DIR = "project.temp.dir";
 
+    // Event reporting properties
+    public static final String AZKABAN_EVENT_REPORTING_CLASS_PARAM =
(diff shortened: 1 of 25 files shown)


HappyRay:
This is the platform part of the earlier PR #1332 which was reverted.

PR #1525, opened 2017-10-04
Flow 2.0 design - embedded flows

jamiesjc:
In Flow 2.0 design, each flow is defined in its own YAML file. We want to support embedded flows, which are basically sub-flows nested inside the parent flow.
There are two approaches to define embedded flows in the new YAML format:
One is to define sub-flows inside the parent flow in the same YAML file. The other is to define sub-flows in separate YAML files and reference sub-flow names in the parent flow YAML file.
In this PR, I'm proposing the first approach.
The flow YAML file will be converted to an azkabanFlow containing a lists azkabanNodes, which can be either an azkabanJob or an azkabanFlow (sub-flow). Once the NodeBean is defined,  loading flow YAML files to azkabanFlow objects can be done in a recursive way.
The implementation is simple and straightforward and proves to be working by unit tests. We can explore the second option later as a future enhancement.


azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
```diff
@@ -18,33 +18,37 @@
...
 import java.util.Map;
 
 public class AzkabanFlow extends AzkabanNode {
 
   private final Map<String, AzkabanNode> nodes;
 
   private AzkabanFlow(final String name, final Props props,
-      final Map<String, AzkabanNode> nodes) {
-    super(name, props);
+      final Map<String, AzkabanNode> nodes, final List<String> dependsOn) {
```
chengren311: maybe parentFlow is more intuitive name?

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
```diff
@@ -18,33 +18,37 @@
...
-    return (AzkabanJob) nodes.get(name);
+  public AzkabanNode getNode(final String name) {
+    return this.nodes.get(name);
   }
 
   public static class AzkabanFlowBuilder {
 
     private String name;
     private Props props;
+    private List<String> dependsOn;
```
chengren311: same here

azkaban-common/src/main/java/azkaban/project/NodeBean.java
```diff
@@ -17,18 +17,20 @@
...
+ * Used by the YAML loader to deserialize DAG nodes in the flow
  */
-public class NodeBean {
+public class NodeBean implements Serializable {
 
   private String name;
   private Map<String, String> config;
   private List<String> dependsOn;
   private String type;
+  private List<NodeBean> nodes;
```
chengren311: maybe a more specific name?

jamiesjc: This name has to match what is defined in the flow YAML file. Since a node can either be a job or a sub-flow, I think it would be ok to have a generic name here.

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
```diff
@@ -18,33 +18,37 @@
...
 import azkaban.utils.Props;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class AzkabanFlow extends AzkabanNode {
```
HappyRay: Please include javadoc for public interfaces.

azkaban-common/src/test/resources/sample_flow.flow
```diff
@@ -1,3 +1,6 @@
+---
+name: sample_flow
+type: flow
```
HappyRay: Is the type here necessary? Doesn't a .flow file always represent a flow type?

jamiesjc: Originally I thought it was easy to implement because both parent flow and embedded flows could be loaded to azkabanFlow in the same way if we define the same structure. But I think from user's perspective, it would be redundant info if we put type flow here. Same as the flow name. I'll remove it in a separate PR, thanks!

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
jamiesjc: Sure, I will add it in a separate PR.

azkaban-common/src/test/resources/sample_flow.flow
jamiesjc: @HappyRay Submitted the changes in another PR https://github.com/azkaban/azkaban/pull/1531. Can you take a look?

PR #1532, opened 2017-10-12
add onKilled to callback interface

chengren311:
onKilled will be called when dependency context is killed successfully. Before that, dependency context is in "Killing" status.


chengren311:
this branch will used for development of data trigger feature. Will submit a series of PRs when it's ready.

PR #1534, opened 2017-10-13
Flow 2.0 design - Convert AzkabanFlow to Flow. Load project YAML file.

jamiesjc:
1. In Flow 2.0 design, we load project flows from flow YAML files and convert them to azkabanFlow objects. In order to display the flow graph in the UI correctly, we need to further convert azkabanFlow to Flow object with all the necessary parameters filled in, like nodes and edges. 
2. With the flow level definition, we don't need to rely on an endNode to resolve the DAG. The new design will be able to resolve the DAG even without an endNode.
3. Project YAML file is also introduced in the Flow 2.0 design. It contains azkaban flow version. Only when "Azkaban-Flow-Version: 2.0" is specified in the project YAML file, then we treat it as Flow 2.0 design. If the project YAML file does not exist, we just keep the current way of loading project flows from job and properties files.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -33,7 +33,7 @@ package azkaban;
 public class Constants {
 
   // Azkaban Flow Versions
-  public static final double VERSION_2_0 = 2.0;
+  public static final double AZKABAN_FLOW_VERSION_2_0 = 2.0;
 
   // Flow 2.0 file suffix
   public static final String PROJECT_FILE_SUFFIX = ".project";
@@ -68,8 +68,8 @@ public class Constants {
 
   public static class ConfigurationKeys {
 
-    // Configures Azkaban Flow Version
-    public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
+    // Configures Azkaban Flow Version in project YAML file
+    public static final String AZKABAN_FLOW_VERSION = "azkaban-flow-version";
 
     // These properties are configurable through azkaban.properties
     public static final String AZKABAN_PID_FILENAME = "azkaban.pid.filename";
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -33,15 +33,14 @@ package azkaban;
 public class Constants {
 
   // Azkaban Flow Versions
-  public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
-  public static final Double VERSION_2_0 = 2.0;
+  public static final double VERSION_2_0 = 2.0;
 
   // Flow 2.0 file suffix
   public static final String PROJECT_FILE_SUFFIX = ".project";
   public static final String FLOW_FILE_SUFFIX = ".flow";
 
   // Flow 2.0 node type
-  public static final String NODE_TYPE_FLOW = "flow";
+  public static final String FLOW_NODE_TYPE = "flow";
 
   // Names and paths of various file names to configure Azkaban
   public static final String AZKABAN_PROPERTIES_FILE = "azkaban.properties";
@@ -68,6 +67,9 @@ public class Constants {
   public static final long DEFAULT_SCHEDULE_END_EPOCH_TIME = 2524608000000L;
 
   public static class ConfigurationKeys {
+
+    // Configures Azkaban Flow Version
+    public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
 
     // These properties are configurable through azkaban.properties
     public static final String AZKABAN_PID_FILENAME = "azkaban.pid.filename";
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -33,11 +33,15 @@ package azkaban;
 public class Constants {
 
   // Azkaban Flow Versions
-  public static final String AZKABAN_FLOW_VERSION_2_0 = "2.0";
+  public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
+  public static final Double VERSION_2_0 = 2.0;
 
   // Flow 2.0 file suffix
   public static final String PROJECT_FILE_SUFFIX = ".project";
   public static final String FLOW_FILE_SUFFIX = ".flow";
+
+  // Flow 2.0 node type
+  public static final String NODE_TYPE_FLOW = "flow";
 
   // Names and paths of various file names to configure Azkaban
   public static final String AZKABAN_PROPERTIES_FILE = "azkaban.properties";
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
@@ -17,6 +17,7 @@
 
 package azkaban.project;
 
+import azkaban.Constants;
 import azkaban.utils.Props;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -36,7 +37,7 @@ public class AzkabanFlow extends AzkabanNode {
 
   private AzkabanFlow(final String name, final Props props,
       final Map<String, AzkabanNode> nodes, final List<String> dependsOn) {
-    super(name, props, dependsOn);
+    super(name, Constants.NODE_TYPE_FLOW, props, dependsOn);
     this.nodes = nodes;
   }
 
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
@@ -37,7 +37,7 @@ public class AzkabanFlow extends AzkabanNode {
 
   private AzkabanFlow(final String name, final Props props,
       final Map<String, AzkabanNode> nodes, final List<String> dependsOn) {
-    super(name, Constants.NODE_TYPE_FLOW, props, dependsOn);
+    super(name, Constants.FLOW_NODE_TYPE, props, dependsOn);
     this.nodes = nodes;
   }
 
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
@@ -184,35 +184,39 @@ class AzkabanProjectLoader {
   private void persistProject(final Project project, final FlowLoader loader, final File archive,
       final User uploader) throws ProjectManagerException {
     synchronized (project) {
+      final int newVersion = this.projectLoader.getLatestProjectVersion(project) + 1;
+      final Map<String, Flow> flows = loader.getFlowMap();
+      for (final Flow flow : flows.values()) {
+        flow.setProjectId(project.getId());
+        flow.setVersion(newVersion);
+      }
+
+      this.storageManager.uploadProject(project, newVersion, archive, uploader);
+
+      log.info("Uploading flow to db " + archive.getName());
+      this.projectLoader.uploadFlows(project, newVersion, flows.values());
+      log.info("Changing project versions " + archive.getName());
+      this.projectLoader.changeProjectVersion(project, newVersion,
+          uploader.getUserId());
+      project.setFlows(flows);
+
       if (loader instanceof DirectoryFlowLoader) {
         final DirectoryFlowLoader directoryFlowLoader = (DirectoryFlowLoader) loader;
-        final int newVersion = this.projectLoader.getLatestProjectVersion(project) + 1;
-        final Map<String, Flow> flows = directoryFlowLoader.getFlowMap();
-        for (final Flow flow : flows.values()) {
-          flow.setProjectId(project.getId());
-          flow.setVersion(newVersion);
-        }
-
-        this.storageManager.uploadProject(project, newVersion, archive, uploader);
-
-        log.info("Uploading flow to db " + archive.getName());
-        this.projectLoader.uploadFlows(project, newVersion, flows.values());
-        log.info("Changing project versions " + archive.getName());
-        this.projectLoader.changeProjectVersion(project, newVersion,
-            uploader.getUserId());
-        project.setFlows(flows);
         log.info("Uploading Job properties");
         this.projectLoader.uploadProjectProperties(project, new ArrayList<>(
             directoryFlowLoader.getJobPropsMap().values()));
         log.info("Uploading Props properties");
         this.projectLoader.uploadProjectProperties(project, directoryFlowLoader.getPropsList());
-        this.projectLoader.postEvent(project, EventType.UPLOADED, uploader.getUserId(),
-            "Uploaded project files zip " + archive.getName());
+
       } else if (loader instanceof DirectoryYamlFlowLoader) {
         // Todo jamiesjc: upload yaml file to DB as a blob
+
       } else {
         throw new ProjectManagerException("Invalid type of flow loader.");
       }
+
+      this.projectLoader.postEvent(project, EventType.UPLOADED, uploader.getUserId(),
+          "Uploaded project files zip " + archive.getName());
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
@@ -193,9 +193,9 @@ class AzkabanProjectLoader {
 
       this.storageManager.uploadProject(project, newVersion, archive, uploader);
 
-      log.info("Uploading flow to db " + archive.getName());
+      log.info("Uploading flow to db for project " + archive.getName());
       this.projectLoader.uploadFlows(project, newVersion, flows.values());
-      log.info("Changing project versions " + archive.getName());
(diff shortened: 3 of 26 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -33,12 +33,16 @@
 public class Constants {
 
   // Azkaban Flow Versions
-  public static final String AZKABAN_FLOW_VERSION_2_0 = "2.0";
+  public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
```
chengren311: this seems to be a property key name, so should it be a inside configurationkey.java?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -33,12 +33,16 @@
...
+  public static final Double VERSION_2_0 = 2.0;
```
chengren311: why not just double instead of Double class?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -33,12 +33,16 @@
...
 
   // Flow 2.0 file suffix
   public static final String PROJECT_FILE_SUFFIX = ".project";
   public static final String FLOW_FILE_SUFFIX = ".flow";
 
+  // Flow 2.0 node type
+  public static final String NODE_TYPE_FLOW = "flow";
```
chengren311: just call NODE_TYPE?

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
```diff
@@ -36,7 +37,7 @@
 
   private AzkabanFlow(final String name, final Props props,
       final Map<String, AzkabanNode> nodes, final List<String> dependsOn) {
-    super(name, props, dependsOn);
+    super(name, Constants.NODE_TYPE_FLOW, props, dependsOn);
```
chengren311: i think it's slightly better to add one more argument in AzkabanFlow constructor called type, and let builder handle the default parameter passing.

az-core/src/main/java/azkaban/Constants.java
jamiesjc: Yes you are right. I'll move it inside ConfigurationKeys class.

az-core/src/main/java/azkaban/Constants.java
jamiesjc: Will change to double, thanks!

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -184,35 +184,39 @@ private boolean isReportStatusValid(final Map<String, ValidationReport> reports,
...
+      final int newVersion = this.projectLoader.getLatestProjectVersion(project) + 1;
+      final Map<String, Flow> flows = loader.getFlowMap();
+      for (final Flow flow : flows.values()) {
+        flow.setProjectId(project.getId());
+        flow.setVersion(newVersion);
+      }
+
+      this.storageManager.uploadProject(project, newVersion, archive, uploader);
+
+      log.info("Uploading flow to db " + archive.getName());
```
chengren311: "Uploading flow"  + archive.getName + " to db"

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
jamiesjc: What's the advantage of doing it?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
chengren311: are we uploading flow or project here?

jamiesjc: Uploading flows to DB project_flows table. 
archive.getName is the project zip name. The logging message is not very clear.

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
```diff
@@ -37,31 +52,136 @@ public void setUp() {
...
+      if (edge.getError() != null) {
+        Assert.assertEquals(DEPENDENCY_NOT_FOUND_ERROR, edge.getError());
+      }
+    }
+
+    // Invalid flow 2: Cycles found.
+    Assert.assertTrue(loader.getFlowMap().containsKey(INVALID_FLOW_2));
+    final Flow flow2 = loader.getFlowMap().get(INVALID_FLOW_2);
+
+    Assert.assertEquals(4, flow2.getNodes().size());
```
kunkun-tang: When parsing yaml runs into any issue, it looks like nodes are identified anyhow. What's the intention behind this behavior?

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
```diff
@@ -37,31 +52,136 @@ public void setUp() {
...
+    loader.loadProjectFlow(this.project, ExecutionsTestUtil.getFlowDir(INVALID_FLOW_YAML_DIR));
+    Assert.assertEquals(2, loader.getErrors().size());
+    Assert.assertEquals(2, loader.getFlowMap().size());
+    Assert.assertEquals(2, loader.getEdgeMap().size());
+
+    // Invalid flow 1: Dependency not found.
+    Assert.assertTrue(loader.getFlowMap().containsKey(INVALID_FLOW_1));
+    final Flow flow1 = loader.getFlowMap().get(INVALID_FLOW_1);
+
+    Assert.assertEquals(3, flow1.getNodes().size());
```
kunkun-tang: If I were to write this test, I would write a common assert method to consolidate all similar check statements.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -40,6 +44,7 @@
   private final Props props;
   private final Set<String> errors = new HashSet<>();
```
kunkun-tang: Question: I'm seeing there exist a few common variables in class DirectoryYamlFlowLoader and class DirectoryFlowLoader, why not make FlowLoader an Abstract class rather than Interface?

azkaban-common/src/main/java/azkaban/project/AzkabanFlow.java
chengren311: since builder is externally visible It would be more flexible if we want to pass the type from outside in the future.

jamiesjc: Since the type of AzkabanFlow can only be "flow", is there a need for it to be passed from outside?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Good question. Several reasons why interface is a better option here:
1. The two classes don't share common implementations. They need to implement their own method `loadProjectFLow` in their own different ways. So defining the method in the interface should be sufficient.
2. Some variables might be common to the two classes but some are not. If we need to add a new type of FlowLoader, it won't necessarily share the same variables. In this case, making it abstract class for sharing common variables might not be flexible.
Actually in my previous PR, I used Abstract class first and then switched to interface. You can refer to the discussions below:
https://github.com/azkaban/azkaban/pull/1500/commits/30744ec9e375f03ce12525e64b50d2c8b1899f21#diff-33dbf7e0198cf8b60e328d5fde7921acR35

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
jamiesjc: Thanks for the suggestion. I'll consolidate the common check statements.

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
jamiesjc: Actually it happens in two steps, parsing the YAML file to AzkabanFlow first and then converting AzkabanFlow to Flow object. In this case, parsing the YAML file is successful, so we can identify the nodes. However, when converting AzkabanFlow to Flow, we resolve the dependency and detect an error. The error will be returned back in the report showing something wrong in the flow to users.

azkaban-common/src/test/java/azkaban/project/FlowLoaderFactoryTest.java
```diff
@@ -43,13 +37,25 @@ public void testCreateDirectoryFlowLoader() {
...
   @Test
-  @Ignore
   public void testCreateDirectoryYamlFlowLoader() {
     final FlowLoaderFactory loaderFactory = new FlowLoaderFactory(new Props(null));
     final File projectDir = ExecutionsTestUtil.getFlowDir(FLOW_20_TEST_DIRECTORY);
     final FlowLoader loader = loaderFactory.createFlowLoader(projectDir);
     Assert.assertTrue(loader instanceof DirectoryYamlFlowLoader);
   }
+
+  @Test(expected = ProjectManagerException.class)
```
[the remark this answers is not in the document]
jamiesjc: Thanks! Will fix it.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -65,6 +68,9 @@
 
   public static class ConfigurationKeys {
 
+    // Configures Azkaban Flow Version
+    public static final String AZKABAN_FLOW_VERSION = "Azkaban-Flow-Version";
```
chengren311: should the naming convention be dot-separated?

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
```diff
@@ -56,131 +57,109 @@ public void testLoadBasicYamlFile() {
     final DirectoryYamlFlowLoader loader = new DirectoryYamlFlowLoader(new Props());
 
     loader.loadProjectFlow(this.project, ExecutionsTestUtil.getFlowDir(BASIC_FLOW_YAML_DIR));
-    Assert.assertEquals(0, loader.getErrors().size());
-    Assert.assertEquals(1, loader.getFlowMap().size());
-    Assert.assertTrue(loader.getFlowMap().containsKey(BASIC_FLOW_1));
+    assertThat(loader.getErrors().size()).isEqualTo(0);
```
kunkun-tang: My previous suggestion was to consolidate these duplicate code since you always need to test loader.getFlowMap().size() in every @test method. Also, flow.getNodes().size(), flow.getErrors().size().  So, why not consolidate them into one method with a number of arguments.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -33,12 +33,15 @@
 public class Constants {
 
   // Azkaban Flow Versions
-  public static final String AZKABAN_FLOW_VERSION_2_0 = "2.0";
+  public static final double VERSION_2_0 = 2.0;
```
chengren311: isn't AZKABAN_FLOW_VERSION_2_0 more explicit?

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -41,21 +46,42 @@ public FlowLoaderFactory(final Props props) {
...
-    if (flowVersion == null) {
-      return new DirectoryFlowLoader(this.props);
-    } else if (flowVersion.equals(Constants.AZKABAN_FLOW_VERSION_2_0)) {
-      return new DirectoryYamlFlowLoader(this.props);
+
+    final File[] projectFileList = projectDir.listFiles(new SuffixFilter(Constants
+        .PROJECT_FILE_SUFFIX));
+    if (projectFileList != null && projectFileList.length > 0) {
+      if (projectFileList.length > 1) {
+        throw new ProjectManagerException("Duplicate project YAML files found in the project "
```
chengren311: might be better to put projectDir in the exception message to be more explicit.

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -41,21 +46,42 @@ public FlowLoaderFactory(final Props props) {
...
+            + "directory. Only one is allowed.");
+      }
+
+      final Map<String, Object> projectYamlParser;
```
chengren311: might call it parsedProjectProperties or something, fitting purpose of this variable more?

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -41,21 +46,42 @@ public FlowLoaderFactory(final Props props) {
...
-    // and create FlowLoader based on different flow versions specified in the manifest file.
-    final String flowVersion = null;
-    if (flowVersion == null) {
-      return new DirectoryFlowLoader(this.props);
-    } else if (flowVersion.equals(Constants.AZKABAN_FLOW_VERSION_2_0)) {
-      return new DirectoryYamlFlowLoader(this.props);
+
+    final File[] projectFileList = projectDir.listFiles(new SuffixFilter(Constants
+        .PROJECT_FILE_SUFFIX));
+    if (projectFileList != null && projectFileList.length > 0) {
```
chengren311: based on https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles(java.io.FileFilter)
returning null means the input dir is not a dir or IO error happens, in this case should we still fall back to DirectoryFlowLoader ?

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
jamiesjc: There is no restrict naming convention for the YAML file. But I found "key-name", "key_name" and "key name" are valid formats from below links:
https://learnxinyminutes.com/docs/yaml/
http://ess.khhq.net/wiki/YAML_Tutorial
https://bitbucket.org/asomov/snakeyaml/wiki/Documentation#markdown-header-yaml-syntax
Also I think "azkaban.flow.version" might be a little confusing since users might think it means the actual version of a flow file.

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
jamiesjc: If listFiles returns null, we should throw Exception. I'll change it, thanks!

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -41,21 +46,42 @@ public FlowLoaderFactory(final Props props) {
...
+      if (projectFileList.length > 1) {
+        throw new ProjectManagerException("Duplicate project YAML files found in the project "
+            + "directory. Only one is allowed.");
+      }
+
+      final Map<String, Object> projectYamlParser;
+      try {
+        projectYamlParser = (Map<String, Object>) new Yaml()
+            .load(new FileInputStream(projectFileList[0]));
```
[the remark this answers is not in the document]
jamiesjc: I will use try-with-resources to handle the closing of the input stream.

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
```diff
@@ -41,21 +46,42 @@ public FlowLoaderFactory(final Props props) {
...
+      } catch (final FileNotFoundException e) {
+        throw new ProjectManagerException("Error reading project YAML file.", e);
+      }
+
+      if (!projectYamlParser.containsKey(Constants.ConfigurationKeys.AZKABAN_FLOW_VERSION)) {
```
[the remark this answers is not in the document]
jamiesjc: I don't think it's necessary to have a default version. If project YAML file exists but doesn't specify the version, then there should be something wrong with it and we should not arbitrarily parse it.

azkaban-common/src/main/java/azkaban/project/FlowLoaderFactory.java
jamiesjc: ProjectDir is a temp dir that's created to unzip the project zip and will be deleted after used. So I don't think we need to print it in the message.

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
jamiesjc: Sure, I just submitted the changes to consolidate the test case. Thanks for the comment!

azkaban-common/src/test/java/azkaban/project/DirectoryYamlFlowLoaderTest.java
```diff
@@ -37,31 +53,67 @@ public void setUp() {
   }
 
   @Test
-  public void testLoadYamlFileFromDirectory() {
+  public void testLoadBasicYamlFile() {
```
chengren311: test dir without any yaml files?
test dir with yaml file without version specified?
test dir with yaml file with invalid version?

jamiesjc: Added the test case, thanks!

Issue #1433, opened 2017-09-06
Remove code duplication in tests: createFlowRunner


juhoautio:
Raised by @HappyRay: https://github.com/azkaban/azkaban/pull/1244#issuecomment-326722462

All of these tests have methods by that name, which overlap more or less:
```
FlowRunnerPipelineTest.java
FlowRunnerPropertyResolutionTest.java
FlowRunnerTest.java
FlowRunnerTest2.java
LocalFlowWatcherTest.java
RemoteFlowWatcherTest.java
```

Combine them as much as possible and move to `FlowRunnerTestUtil`.

Let's have this merged first to avoid conflicts: https://github.com/azkaban/azkaban/pull/1432

juhoautio:
Also the FlowRunnerPipelineTest has some `createFlowRunner` methods, so let's also have this first: https://github.com/azkaban/azkaban/pull/1243

juhoautio:
@ameyamk please let me know if there's review resources available for this change. Otherwise we could close the issue.

ameyamk:
I am sorry Juho. We are slammed right now.

HappyRay:
@juhoautio if you submit a PR for this issue, I will review it. Thanks

PR #1536, opened 2017-10-13
Issue #1433 part 2/2: Remove code duplication in FlowWatcher tests

juhoautio:
This completes fixing issue https://github.com/azkaban/azkaban/issues/1433 (1/2 was PR https://github.com/azkaban/azkaban/pull/1533).

"Extras":
- LocalFlowWatcherTest was actually ignored; I made it run quickly and enabled it.
- I also found & fixed a bug in both FlowWatcher tests: they were asserting pipeline level 1 twice, despite the test names. I fixed those "level 2" tests to run with and assert pipeline level 2 instead of level 1.


HappyRay:
It appears that GitHub will offer the commit message sometimes as the final commit message instead of the pull request description. 
I should have copied the pull request description when I merge.

PR #1541, opened 2017-10-24
Creating a Credential interface to inject user secrets into custom credential object [Work In Progress]

kunkun-tang:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -802,20 +830,21 @@ private Text getMRTokenRenewerInternal(final JobConf jobConf) throws IOException
   /**
    * Method to create a metastore client that retries on failures
    */
-  private IMetaStoreClient createRetryingMetaStoreClient(HiveConf hiveConf) throws MetaException {
+  private IMetaStoreClient createRetryingMetaStoreClient(final HiveConf hiveConf)
```
kunkun-tang: Those are format changes automatically forced by Intellij save-action plugin

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/Credential.java
```diff
@@ -0,0 +1,31 @@
...
+ * Facilitate to register user secret key into custom credential object.
+ */
+public interface Credential {
+
+  /**
+   * register custom secret keys on behalf of an user into Credentials.
+   *
+   * @param user find out the user's credential and register it.
+   */
+  public void register(String user);
```
eogren: Should this take the Credentials object as an explicit parameter?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/Credential.java
```diff
@@ -0,0 +1,31 @@
...
+ * the License.
+ *
+ */
+
+package azkaban.security;
+
+/**
+ * Facilitate to register user secret key into custom credential object.
+ */
+public interface Credential {
```
eogren: I'd probably call this CredentialProvider or something like that

kunkun-tang: `CredentialProvider ` Sounds better. Thanks.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/Credential.java
kunkun-tang: I was considering to have Credentials (e.g. hadoop.Crednetials) as an explicit parameter. I think we might provide more APIs rather than `register` here later on. We could write the API to get token/credentials from this interface. In that case, may we consider add the object when we do constructor?

kunkun-tang:
This PR is just an example how credential provider works. Proposing #1552, which should be merged.

Issue #1526, opened 2017-10-04
Trigger a flow based on data dependency availability


chengren311:
It is common for a scheduled flow to have data dependencies. With the data availability check feature for Azkaban flows, it will be easier for jobs in an Azkaban workflow to start only when the data they depend on becomes available. 

Motivation
1. Improve user productivity by standardizing how Azkaban flows specify data dependencies. Users don’t have to build their own jobs that check if the data their flows depend on are available.  

2. Reduce overall flow latency and error rate. Some flows are scheduled based on estimated data availability time. The flows will fail when the data is not available when they start. To reduce such failure rates, users will add some buffer time e.g. schedule a flow to start at 10 AM when the dependent dataset is typically available at 8 AM. Despite the buffer, the flow may still fail if the data is delayed for longer than 2 hours.  This feature allows these flows to be scheduled at an earlier time than they currently do so. In the example, the flow can be scheduled to run at 8 AM without worrying out failures due to data availability.  By removing the failures caused by data delays, the feature makes it easier to see true flow logical failures. 

3. Improve overall infrastructure efficiency and stability.  Azkaban can check the data availability efficiently. See the detailed design section below for details. A typical implementation of a custom job that checks for data availability relies on polling HDFS/Hive Metastore. A bad implementation or too many such jobs running concurrently can put pressure on the HDFS namenode/ Hive Metastore. 

4. Encourage users to provide the data dependency information as metadata that Azkaban can understand. Azkaban can leverage this information in the future to gain more flexibility or optimize the system. E.g. schedule the flow in a cluster where the data lives. 
bring the data to the cluster that the flow runs. This can happen if the flow relies on GPU in a certain Hadoop cluster for example.

juhoautio:
👍  for this first of all.

I'm mostly curious about how this will be reflected in the UI and what it will looks like in job configuration. Would you have any examples to share about those?

After all, from my perspective it maybe seems like this interface may not be as useful as it can be limiting when compared to custom polling scripts.. With an explicit job to poll for data it is quite clearly visible from the Job List where the flow is going, how long it has been "blocked by a data dependency", and why (you can log details in the job log of the data polling job). Personally I would really hope to have a similar level of visibility in the UI for the dependency checks.

ameyamk:
Yes - we spent lot of time thinking if data dependency should be modeled as as a job - or should it be designed at a flow level. After lot of thinking we have decided to design this at a flow level. There are couple of reasons for this:

[...]

PR #1543, opened 2017-10-24
make ExecutableFlow initialization methods common util methods

chengren311:
This PR consolidates some methods of initialization of ExecutableFlow which used in multiple places  into common static util methods. 
The motivation is to let flow trigger leverage those method to start an executable flow(https://github.com/azkaban/azkaban/issues/1526).

Also changed the azkaban.utils.TestUtils#createExecutableFlow to azkaban.utils.TestUtils#createTestExecutableFlow to avoid name duplication.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -924,9 +924,8 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
       return;
     }
 
-    final ExecutableFlow exflow = new ExecutableFlow(project, flow);
+    final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
     exflow.setSubmitUser(user.getUserId());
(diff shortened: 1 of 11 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -924,9 +924,8 @@ private void ajaxExecuteFlow(final HttpServletRequest req,
       return;
     }
 
-    final ExecutableFlow exflow = new ExecutableFlow(project, flow);
+    final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
```
kunkun-tang: I remember we have a method with the same name in test module. May we change the method name there, like `createTestExecutableFlow`?

azkaban.utils.TestUtils#createExecutableFlow

reallocf: :+1:

reallocf:
LGTM except for the thing @kunkun-tang pointed out. Once that's fixed I'll give a ship it :+1:.

chengren311:
@reallocf updated

PR #1544, opened 2017-10-24
Flow 2.0 design - Create project_flow_files DB table.

jamiesjc:
1. Created project_flow_files DB table to store versioned flow yaml files.
2. Supported uploading flow files to DB and fetching them.
3. Verified DB operations in unit tests.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
@@ -333,4 +333,26 @@ class JdbcProjectHandlerSet {
       return rs.getInt(1);
     }
   }
+
+  public static class FlowFileResultHandler implements ResultSetHandler<List<byte[]>> {
+
+    public static String SELECT_FLOW_FILE =
+        "SELECT flow_file FROM project_flow_files WHERE "
+            + "project_id=? AND project_version=? AND flow_name=? AND flow_version=?";
+
+    @Override
+    public List<byte[]> handle(final ResultSet rs) throws SQLException {
+      if (!rs.next()) {
+        return Collections.emptyList();
+      }
+
+      final ArrayList<byte[]> data = new ArrayList<>();
+      do {
+        final byte[] bytes = rs.getBytes(1);
+        data.add(bytes);
+      } while (rs.next());
+
+      return data;
+    }
+  }
 }
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
@@ -346,7 +346,7 @@ class JdbcProjectHandlerSet {
         return Collections.emptyList();
       }
 
-      final ArrayList<byte[]> data = new ArrayList<>();
+      final List<byte[]> data = new ArrayList<>();
       do {
         final byte[] bytes = rs.getBytes(1);
         data.add(bytes);
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -24,11 +24,12 @@ import static azkaban.project.JdbcProjectHandlerSet.ProjectPropertiesResultsHand
 import static azkaban.project.JdbcProjectHandlerSet.ProjectResultHandler;
 import static azkaban.project.JdbcProjectHandlerSet.ProjectVersionResultHandler;
 
-import azkaban.db.EncodingType;
 import azkaban.db.DatabaseOperator;
 import azkaban.db.DatabaseTransOperator;
+import azkaban.db.EncodingType;
 import azkaban.db.SQLTransaction;
 import azkaban.flow.Flow;
+import azkaban.project.JdbcProjectHandlerSet.FlowFileResultHandler;
 import azkaban.project.ProjectLogEvent.EventType;
 import azkaban.user.Permission;
 import azkaban.user.User;
@@ -40,8 +41,6 @@ import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
 import azkaban.utils.Triple;
 import com.google.common.io.Files;
-import javax.inject.Inject;
-import javax.inject.Singleton;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
@@ -54,6 +53,8 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.inject.Inject;
+import javax.inject.Singleton;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 
@@ -69,6 +70,8 @@ public class JdbcProjectImpl implements ProjectLoader {
   private static final Logger logger = Logger.getLogger(JdbcProjectImpl.class);
 
   private static final int CHUCK_SIZE = 1024 * 1024 * 10;
+  // Flow yaml files are usually small, set size limitation to 10 MB should be sufficient for now.
+  private static final int MAX_FLOW_FILE_SIZE = 1024 * 1024 * 10;
   private final DatabaseOperator dbOperator;
   private final File tempDir;
   private final EncodingType defaultEncodingType = EncodingType.GZIP;
@@ -950,6 +953,7 @@ public class JdbcProjectImpl implements ProjectLoader {
     final String DELETE_PROPERTIES = "DELETE FROM project_properties WHERE project_id=? AND version<?";
     final String DELETE_PROJECT_FILES = "DELETE FROM project_files WHERE project_id=? AND version<?";
     final String UPDATE_PROJECT_VERSIONS = "UPDATE project_versions SET num_chunks=0 WHERE project_id=? AND version<?";
+    // Todo jamiesjc: delete flow files
 
     final SQLTransaction<Integer> cleanOlderProjectTransaction = transOperator -> {
       transOperator.update(DELETE_FLOW, projectId, version);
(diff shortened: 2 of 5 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
```diff
@@ -333,4 +333,26 @@ public Integer handle(final ResultSet rs) throws SQLException {
...
+        "SELECT flow_file FROM project_flow_files WHERE "
+            + "project_id=? AND project_version=? AND flow_name=? AND flow_version=?";
+
+    @Override
+    public List<byte[]> handle(final ResultSet rs) throws SQLException {
+      if (!rs.next()) {
+        return Collections.emptyList();
+      }
+
+      final ArrayList<byte[]> data = new ArrayList<>();
```
chengren311: List<byte[]> ?

chengren311: why not while(rs.next()) {
},
then we don't need to have the !rs.next() check in the first place.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -69,6 +70,8 @@
   private static final Logger logger = Logger.getLogger(JdbcProjectImpl.class);
 
   private static final int CHUCK_SIZE = 1024 * 1024 * 10;
+  // Flow yaml files are usually small, set size limitation to 10 MB should be sufficient for now.
+  private static final int MAX_FLOW_FILE_SIZE = 1024 * 1024 * 10;
```
chengren311: MAX_FLOW_FILE_SIZE_IN_BYTES?

azkaban-db/src/main/sql/create.project_flow_files.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE project_flow_files (
+  project_id        INT          NOT NULL,
+  project_version   INT          NOT NULL,
+  flow_name         VARCHAR(128) NOT NULL,
```
chengren311: is 128 our convention?

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
```diff
@@ -352,6 +397,7 @@ public void clearDB() {
       dbOperator.update("DELETE FROM project_flows");
       dbOperator.update("DELETE FROM project_files");
       dbOperator.update("DELETE FROM project_events");
+      dbOperator.update("DELETE FROM project_flow_files");
```
chengren311: nit: truncate should be faster than delete

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
       throw new ProjectManagerException("clean older project transaction failed", e);
     }
   }
+
+  @Override
+  public void uploadFlowFile(final int projectId, final int projectVersion, final int flowVersion,
```
kunkun-tang: Will projectVersion always change when we upload a new zip?

kunkun-tang: If the project has two ymls. If one yml changes, will only one flowVersion change?

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -200,4 +200,17 @@ Props fetchProjectProperty(int projectId, int projectVer, String propsName)
       throws ProjectManagerException;
 
   void updateProjectSettings(Project project) throws ProjectManagerException;
+
+  /**
+   * Uploads flow file.
+   */
+  void uploadFlowFile(int projectId, int projectVersion, int flowVersion,
```
kunkun-tang: Can we have the same argument order as `getUploadedFlowFile`?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+      final File flowFile) throws ProjectManagerException {
+    logger.info(String
+        .format(
+            "Uploading flow file %s, version %d for project %d, version %d, file length is [%d bytes]",
+            flowFile.getName(), flowVersion, projectId, projectVersion, flowFile.length()));
+
+    if (flowFile.length() > MAX_FLOW_FILE_SIZE) {
```
kunkun-tang: What will happen when a user submits a right size zip after a too big zip failure? Will flowVersion increase one? Have you considered moving the logic at the beginning of uploading yml logic, like before launching database operations?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+      } catch (final SQLException e) {
+        throw new ProjectManagerException(
+            "Error uploading flow file " + flowFile.getName() + ", version " + flowVersion + ".",
+            e);
+      }
+    } catch (final IOException e) {
+      throw new ProjectManagerException(
+          String.format(
+              "Error reading flow file %s, version: %d, length: [%d bytes].",
+              flowFile.getName(), flowVersion, flowFile.length()));
```
kunkun-tang: why we don't throw exception here?

azkaban-db/src/main/sql/create.project_flow_files.sql
kunkun-tang: We should double check on this.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: 1. When we upload a new project zip, projectVersion will change.
2. Say the project has two flow yaml files. If one yaml file changes and user uploads a new project zip, then flowVersion will change for both yaml files. If one yaml file is changed through UI override, then we only change the flowVersion for this yaml file and won't affect the other one.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: FlowVersion will not increase because the big file is not uploaded to DB. 
User will first query DB to get the latest flow version (will be implemented in the next PR), and then call this method to upload new version of flow file to DB. This file length check happens before the uploading DB operation.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: I've thrown ProjectManagerException here. What's your concern?

azkaban-db/src/main/sql/create.project_flow_files.sql
jamiesjc: Yes, in our current DB table, flow_id/flow_name is 128.

azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
jamiesjc: The difference between the two approach is that one will return Collections.emptyList and the other will return new ArrayList<>() if the query result is empty. And there is some performance advantage of returning emptyList:
https://stackoverflow.com/questions/5552258/collections-emptylist-vs-new-instance
Also, it is a commonly used approach in our current code, so I just followed the convention.

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
jamiesjc: Thanks for sharing the knowledge. I think truncate is better in this case.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
jamiesjc: They don't have exactly the same arguments, but I'll try to align them.

azkaban-db/src/main/sql/create.project_flow_files.sql
kunkun-tang: Discussed offline. We can remain it to be 128 as same as flow_id in other tables.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
kunkun-tang: Good. Didn't see the throw statement.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+            .query(FlowFileResultHandler.SELECT_FLOW_FILE, handler,
+                projectId, projectVersion, flowName, flowVersion);
+      } catch (final SQLException e) {
+        logger.error(e);
+        throw new ProjectManagerException("Failed to query uploaded flow file " + flowName + ".",
+            e);
+      }
+
+      try {
+        bufferedStream.write(data.get(0));
```
chengren311: it will throw index out of bound exception when data is empty when corresponding file is not found.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+  }
+
+  @Override
+  public File getUploadedFlowFile(final int projectId, final int projectVersion,
+      final int flowVersion, final String flowName) throws ProjectManagerException {
+    final FlowFileResultHandler handler = new FlowFileResultHandler();
+
+    final List<byte[]> data;
+    // Todo jamiesjc: delete the flow file after used.
+    final File file = new File(this.tempDir, flowName);
```
chengren311: if getUploadedFlowFile of same flow file is being called concurrently, will the same file an intertwine of multiple threads's result?
  File.createTempFile() might be worth taking a look.

chengren311: BTW, you can also do file.deleteOnExit().

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+    } catch (final IOException e) {
+      throw new ProjectManagerException(
+          String.format(
+              "Error reading flow file %s, version: %d, length: [%d bytes].",
+              flowFile.getName(), flowVersion, flowFile.length()));
+    }
+  }
+
+  @Override
+  public File getUploadedFlowFile(final int projectId, final int projectVersion,
```
chengren311: what do you think of the argument order like this (projectId, projectVersion, flowName, flowVersion)? since It's good to keep the consistency of argument order(id, version, id, version).

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -200,4 +200,17 @@ Props fetchProjectProperty(int projectId, int projectVer, String propsName)
       throws ProjectManagerException;
 
   void updateProjectSettings(Project project) throws ProjectManagerException;
+
+  /**
+   * Uploads flow file.
+   */
+  void uploadFlowFile(int projectId, int projectVersion, int flowVersion, File flowFile)
```
chengren311: same here about the argument order.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -968,4 +972,75 @@ public void cleanOlderProjectVersion(final int projectId, final int version)
...
+      final int flowVersion, final String flowName) throws ProjectManagerException {
```
chengren311: flowName might be confusing since (flowName+".flow") is what it really means. Outside user without the context will pass a flowId or flowName based on the argument name but it will return nothing. Might be calling it flowFileName, and mention in the method doc that it has to have a suffix of  ".flow" .

jamiesjc: Agree. I'll change it to flowFileName.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: FlowName actually refers to the flow file name, not the flowId. The name seems a little confusing here.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: Thanks for pointing it out! I think creating a temp dir might be better because the file name has to be specified as the flow name which is used during parsing of the flow yaml file.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: Good catch!

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
jamiesjc: I'm ok with either order. I chose this argument order based on the sequence in the project_flow_fiels DB table. 
project_id | project_version | flow_name | flow_version | modified_time | flow_file

PR #1546, opened 2017-10-26
Flow 2.0 design - Upload flow YAML files to DB.

jamiesjc:
1. Add DB operation to get latest flow version from project_flow_files table.
2. Upload flow YAML files to DB table during uploading project zip.


azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -209,8 +212,13 @@ private void persistProject(final Project project, final FlowLoader loader, fina
         this.projectLoader.uploadProjectProperties(project, directoryFlowLoader.getPropsList());
 
       } else if (loader instanceof DirectoryYamlFlowLoader) {
-        // Todo jamiesjc: upload yaml file to DB as a blob
-
+        final File[] flowFiles = projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX));
+        for (final File file : flowFiles) {
+          final int newFlowVersion = this.projectLoader
+              .getLatestFlowVersion(project.getId(), newProjectVersion, file.getName()) + 1;
+          this.projectLoader
```
chengren311: as discussed, maybe we should persist the name of the uploader as well.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -209,8 +212,13 @@ private void persistProject(final Project project, final FlowLoader loader, fina
         this.projectLoader.uploadProjectProperties(project, directoryFlowLoader.getPropsList());
 
       } else if (loader instanceof DirectoryYamlFlowLoader) {
-        // Todo jamiesjc: upload yaml file to DB as a blob
-
+        final File[] flowFiles = projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX));
+        for (final File file : flowFiles) {
```
chengren311: one line?
for (final File file : projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX)))

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: The name of uploader is already persisted in project_versions DB table. See azkaban.project.ProjectLoader#changeProjectVersion.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: Will make the change together with next PR.

chengren311:
Can we upload in-memory object to db instead of the raw file? Then when reading flow config from db, we don't need to do all of these processing(download to a temp file, converting the temp file to an object, then delete the temp file).

Issue #1522, opened 2017-10-02
Issue on inherited Parameters for embedded flows


armisael:
Let's say we have a common job that is used by two embedded flows, which simply set some parameters when running it. Let's also say that such parameters are also used by other jobs, and therefore must placed inside a `.properties` file to avoid duplication.

[This project](https://github.com/azkaban/azkaban/files/1348950/_project.zip) shows an example of this:

```
.
|____common                    // folder for the common jobs
| |____echo.job                // common job, it requires a `foo` parameter
|____flow-a
| |____default.properties      // contains `foo.value=hello world`
| |____do-echo-a.job           // embedding of echo.job: it sets `foo=${foo.value}`
|____flow-b
| |____default.properties      // contains `foo.value=foo bar`
| |____do-echo-b.job           // embedding of echo.job: it sets `foo=${foo.value}`
|____main.job
```

This structure is allowed on azkaban 2.5: when `do-echo-a.job` is run, the parameter is expanded, and `foo=hello world` will be used when executing `echo.job`. Similarly, when `do-echo-b.job` is run, `foo=foo bar` will be used when executing `echo.job`.

In azkaban 3.x this is not allowed anymore. Expansion of parameters in the `do-echo-*.job` is performed based on the location of `echo.job`, and not of the flow itself; when `foo` is expanded, then, the property `foo.value` is searched among the properties define at `common` folder level, even if they are defined on a `.job` file in `flow-a` or `flow-b`.

Is this behaviour intended?

HappyRay:
How does do-echo-a.job reference echo.job?

armisael:
As an embedded flow:
```type=flow
flow.name=echo

foo=${foo.value}```

You can download the project and give it a try, it's linked in the issue description.

jamiesjc:
@armisael @HappyRay This issue could be related to an old PR https://github.com/azkaban/azkaban/commit/d114439dbc1f5d7f948888f7240e9a92a7cc1661.
In azkaban.execapp.FlowRunner#prepareJobProperties,
only if a node is not an instance of ExecutableFlow, it will load the sharedProps for the node. In this scenario, the sharedProps are defined in the default.properties files and they are not populated into the node since it is an instance of ExecutableFlow, and thus fail to be resolved in azkaban.utils.PropsUtils#resolveProps, causing the failure.
I tried removing `if (!(node instanceof ExecutableFlowBase))` and the flow could run successfully. This might not be the ideal fix, but it points out where the issue locates.

[...]

PR #1547, opened 2017-10-27
propose bugfix for #1522

armisael:
This required more work than I expected.

Currently for nested flows we totally ignore `.properties` files, due to a `if (!(node instanceof ExecutableFlowBase))`; what we want to achieve instead is to use them when evaluating properties of the flow itself, but to ignore them in inner jobs.

The only way to achieve this is to introduce a level of "visibility" for properties: `.properties` files of a flow will be marked as private, and used only on the flow itself; for inner jobs they will be ignored thanks to the new implementation of `clone` and `resolveProps`.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/Props.java
+++ b/az-core/src/main/java/azkaban/utils/Props.java
@@ -45,6 +45,7 @@ import org.apache.log4j.Logger;
 public class Props {
 
   private final Map<String, String> _current;
+  private boolean isPrivate;
   private Props _parent;
   private String source = null;
 
@@ -59,8 +60,13 @@ public class Props {
    * Constructor for empty Props with parent override.
    */
   public Props(final Props parent) {
+    this(parent, false);
+  }
+
+  public Props(final Props parent, final boolean isPrivate) {
     this._current = new HashMap<>();
     this._parent = parent;
+    this.isPrivate = isPrivate;
   }
 
   /**
@@ -154,7 +160,11 @@ public class Props {
    * Clones the Props p object and all of its parents.
    */
   public static Props clone(final Props p) {
-    return copyNext(p);
+    return Props.clone(p, false);
+  }
+
+  public static Props clone(final Props p, final boolean onlyPublic) {
+    return copyNext(p, onlyPublic);
   }
 
   /**
@@ -162,17 +172,34 @@ public class Props {
    * @param source
    * @return
    */
-  private static Props copyNext(final Props source) {
-    Props priorNodeCopy = null;
-    if (source.getParent() != null) {
-      priorNodeCopy = copyNext(source.getParent());
+  private static Props copyNext(final Props source, final boolean onlyPublic) {
+    Props parent = source.getParent();
+    while (parent != null && onlyPublic && parent.isPrivate) {
+      parent = parent.getParent();
     }
-    final Props dest = new Props(priorNodeCopy);
+
+    Props priorNodeCopy = null;
+    if (parent != null) {
+      priorNodeCopy = copyNext(parent, onlyPublic);
+    }
+
+    final Props dest = new Props(priorNodeCopy, source.isPrivate);
+    if (onlyPublic && source.isPrivate) {
+      return dest;
+    }
+
     for (final String key : source.localKeySet()) {
       dest.put(key, source.get(key));
     }
 
     return dest;
+  }
+
+  public void makePrivate() {
+    this.isPrivate = true;
+    if (this._parent != null) {
+      this._parent.makePrivate();
+    }
   }
 
   /**
@@ -254,8 +281,8 @@ public class Props {
   }
 
   /**
-   * Put the given string value for the string key. This method performs any variable substitution
-   * in the value replacing any occurance of ${name} with the value of get("name").
+   * Put the given string value for the string key. This method *does not* perform any variable
+   * substitution; this is done by PropsUtils.resolveProps; any ${name} variable will be kept.
    *
    * @param key The key to put the value to
    * @param value The value to do substitution on and store
@@ -267,9 +294,8 @@ public class Props {
   }
 
   /**
-   * Put the given Properties into the Props. This method performs any variable substitution in the
-   * value replacing any occurrence of ${name} with the value of get("name"). get() is called first
-   * on the Props and next on the Properties object.
+   * Put the given Properties into the Props. This method *does not* perform any variable
+   * substitution; this is done by PropsUtils.resolveProps; any ${name} variable will be kept.
    *
    * @param properties The properties to put
    * @throws IllegalArgumentException If the variable given for substitution is not a valid key in
@@ -436,10 +462,10 @@ public class Props {
    * we will get ["thrift://hcat1:port,thrift://hcat2:port", "thrift://hcat3:port,thrift://hcat4:port"] as output
    */
   public List<String> getStringListFromCluster(final String key) {
-    List<String> curlist = getStringList(key, "\\s*;\\s*");
+    final List<String> curlist = getStringList(key, "\\s*;\\s*");
     // remove empty elements in the array
-    for (Iterator<String> iter = curlist.listIterator(); iter.hasNext(); ) {
-      String a = iter.next();
+    for (final Iterator<String> iter = curlist.listIterator(); iter.hasNext(); ) {
+      final String a = iter.next();
       if (a.length() == 0) {
         iter.remove();
       }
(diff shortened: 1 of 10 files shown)


jamiesjc:
Thanks for the fix! @armisael 
I'm still trying to understand the whole idea. So what does it mean by "use them when evaluating properties of the flow itself, but to ignore them in inner jobs." For .properties files, shouldn't they apply to all the flows and also jobs inside the flows?

armisael:
Hi @jamiesjc, thank you for taking the time to have a look.

> For .properties files, shouldn't they apply to all the flows and also jobs inside the flows?

That's what I thought when I started working on this; to achieve this it was enough to remove the `if` statement you mentioned in #1522; there is a test, thought, that due to this change was failing, and it makes sense.

So, let's reason about this with the concept of "scoping". This is the current implementation of property scoping for a job:
1) properties defined in `.properties` files in the job folders, and up;
1) properties of the parent flow (if any);
1) output properties from jobs on which the current job depends on;
1) properties on the source of the job.

Now, if we replace the 2nd point recursively, we get:
1) properties defined in `.properties` files in the job folders, and up;
1) properties of the parent flow (if any):
   1) properties defined in `.properties` files in the parent job folders, and up;
   1) output properties from jobs on which the parent depends on;
   1) properties on the source of the parent.
1) output properties from jobs on which the current job depends on;
1) properties on the source of the job.

The reason that `if` statement was added is `2.i`: if the parent job is in a different folder than the inner job, the properties coming from the parent `.properties` files are out of scope to the job itself, and should be taken out; indeed, that `if` statement was specifically removing them for nested flow nodes.

This lead to bug #1522 though: properties coming from those files should not be simply removed from any scope: they should be removed from the scope of inner flows only; they should still be used for evaluating properties of the flow node itself. Take the example shown in bug #1522: both `do-echo-a` and `do-echo-b` are nested flows, and both set `foo=${foo.value}`; `foo.value` is a variable defined in the `default.properties` file located in their folders; the inner job `echo` should therefore receive the value of `foo`, but not the value of `foo.value` (it is defined in a `.properties` file that is out of scope). We can't simply ignore reading `default.properties`, though, since its content is needed to evaluate `foo=${foo.value}`.

We had therefore to implement a new concept of "visibility" of properties: `.properties` files of a flow should be used only to evaluate properties on the flow node, and be ignored for inner jobs; in the code this is represented by the `isPrivate` flag: private properties are filtered out when reading properties from parent flows; to do so, we had to:
- add a `makePrivate` method on `Props`;
- add a `getPublicKeySet` method on `Props`, which is used inside `PropsUtils.resolveProps` to resolve only public properties;
- add a `clone` method on `Props` that allows to specify whether to clone all the properties, or only the public ones (through the `onlyPublic` parameter): this is used in `FlowRunner` to get only the public properties of the parent flow.

HappyRay:
I haven't read this PR in detail. But it looks complicated.

@jamiesjc is working on a different yaml based representation of flows. Is that design subject to the same issue?

@armisael how urgent is it for you to solve this issue?

jamiesjc:
@armisael Thanks for the detailed explanation. So if I'm understanding it correctly, the issue happens when the inner flow exists in a different folder than the job (with the type flow) which references it? I'm not sure if this is a valid scenario that we need to support currently. 
@HappyRay In the new YAML based representation, inner flows are embedded inside the same parent flow file, so there is no such issue.

armisael:
@jamiesjc you are correct, this happens only when inner flow exists in a different folder.
@HappyRay this is blocking our migration to azk 3.0, so it would be greatly appreciated.

I understand you are developing a new way of representing flows; I guess you are not simply dumping the current `.job`-based flows, otherwise upgrading to the new azkaban with YAML support would be a big pain. This said, this issue will still be there, regardless of the new YAML implementation.

The real question is what @jamiesjc said: do you want to support this scenario? As stated in #1522, this is the only way to execute the same job/flow with different parameters in `.properties` files. This is not that unlikely, we are doing it quite a lot. Workarounds are possible, but quite nasty.

jamiesjc:
@armisael I'd like to discuss how we define the scope of job properties, mainly on point 1 and 2. From 1st point, a job could inherit all the properties from .properties files in its current folder and parent folders. From 2nd point, a job could inherit all the properties from its parent flow. 
I can understand your intention of the fix, but I couldn't think of a use case where it could cause problems if the inner job inherits everything from its parent flow including the .properties files in the parent flow folder. It would be great if you could share any use case like this. The properties are resolved for the parent flow first and they should be part of the parent flow already. I don't think there is a need to over complicate the issue by hiding some of the properties for inner jobs when they inherit from the parent flow. Maybe I'm wrong, but I think the complicated logic is more prone to potential bugs and hard to maintain. 
Today Azkaban does not define per flow properties. Users have to create sub-folders to limit the scope of a flow and its properties as a workaround. This is a little nasty and could cause some problems like this one. We are working on the new flow level design which could solve this problem. At the same time, we'd like to support the current code by fixing critical bugs, but just to let you know this part of the logic would eventually be deprecated.

armisael:
Thank you @jamiesjc for the answer, and sorry for the late reply, I've been away for a few days.

I'm really curious to see the new YAML-based way of defining flows, I'm not a big fan of the current implementation, even if after so many years I got used to it. As soon as it's able to represent complex flows as we currently can, I think it will be a great addition (and I'm sure YAML composition with anchors will come in handy).

About your question, there are some cases where the naïve implementation (inheriting all the parent properties) fails; you can try it out: just get on `master` branch, comment the `if` statement of `FlowRunner.java:652` and run the tests: one of those defined in `FlowRunnerPropertyResolutionTest` fails.

The problem is shown in `execpropstest`: `job4` should get property `props4=shared4` from `subdir/shared.properties`, but it gets `props4=moo4` from `moo.properties`.

These are the properties read from every source:
```
1. from properties files
  key=props1 value=shared1
  key=props2 value=shared2
  key=props3 value=moo3
  key=props4 value=shared4
  key=props5 value=moo5
  key=props6 value=shared6
  key=props8 value=shared8
2. from parent flow
  key=props1 value=shared1
  key=props2 value=shared2
  key=props3 value=moo3
  key=props4 value=moo4
  key=props5 value=innerflow5
  key=props6 value=innerflow6
  key=props7 value=flow7
  key=props8 value=innerflow8
  key=props9 value=gjob9
  key=props10 value=gjob10
3. from output props of dependencies
  key=props7 value=g2job7
  key=props9 value=g2job9
4. from job source
  key=props8 value=job8
  key=props9 value=job9
```

As you can see, at step 1 the property is correct: `props4=shared4`; at step 2, though, `props4=moo4` overwrites the correct value; this is due to the fact that it is a property file of the parent flow, which gets (wrongly) priority over the property file of the inner flow.

This thing is a subtle variant of the "scoping" issue I stated before.

HTH

jamiesjc:
@armisael I did some experiments on that test case and also saw the scoping issue you pointed out. To be honest, the existing rule of property priority is really complicated and could cause confusion. You have a very deep understanding of the issue and your fix looks good to me. But I would prefer simplifying the current logic instead of applying more complicated fix on top of the existing code. 
In the new flow YAML file, we would get rid of the shared properties files. Each flow will have its own flow properties defined inside. Sub-directories will not be needed. The tradeoff would be some duplications in the flow properties. Today we already have some internal tool to handle this duplication. For external users, we are trying to make it easy and convenient to migrate as well.

armisael:
Hi,

I understand your point; it's sad because this makes impossible for us to migrate all of our flows without a strong intervention, but your code, your rules. The good thing of the current implementation is that when things don't work, they crash instead of silently passing bad properties. This at least will help us identifying where the issues are.

That said, I hope the new YAML-based flows will be expressive enough to represent complex structures; simplicity and readability are very important, but if the old way of defining flows is buggy, and the new one is not able to cover all the use cases, it would rule out azkaban for most users (properties duplication could be an issue is some cases, for example). I'm sure you considered analysing how others tackled the same issue (the first coming to my mind is Ansible, for example), so I'll wish you all the best, I'm really looking forward to seeing this new feature come alive.

HappyRay:
@armisael 

At LinkedIn, we use the following DSL to generate AZ flow files.
https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop

The new YAML based flows assume code/config sharing will be handled by a higher level DSL. Will this approach work for you?

armisael:
@HappyRay the DSL looks for sure promising! Will it be included in the azkaban project, then?

HappyRay:
Yes, we are thinking of merging these two projects.

I hope people will come up with different DSLs that would best suit their needs. Here is an example I know.
https://github.com/aslotnick/azflow
The new YAML based representation hopefully would make creating such DSLs easier. @jamiesjc has submitted a few pull requests related to that feature. Please feel free to ask her if you have questions.

PR #1549, opened 2017-11-02
Flow 2.0 design - Get props from flow YAML file.

jamiesjc:
1. Support API to get flows and job properties from flow YAML file. This is needed when executing the flows and jobs.
2. With the new YAML file format, there is no need to specify the parent flow name as a prefix for the jobs. YAML format provides a structured way to distinguish sub-flows or jobs with same names between different parent flows. To achieve that, we choose to adopt ":" as the delimiter for job path. For example, the job path in an embedded flow could have the internal naming as "flow1: subflow2: subflow3: job4".  As a result, ":" will be restricted for use in the flow or job names. We will add validation to that later.
3. Today we are showing sub-flows in the project UI page together with the parent flow. This is not ideal because sub-flows should not be treated as independent flows and exposed for executing separately. This breaks the concept of a sub-flow. As an improvement, we will not display sub-flows as individual flows in the project UI page.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
@@ -24,6 +24,7 @@ import azkaban.flow.Node;
 import azkaban.project.FlowLoaderUtils.SuffixFilter;
 import azkaban.project.validator.ValidationReport;
 import azkaban.utils.Props;
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.util.ArrayList;
@@ -121,9 +122,7 @@ public class DirectoryYamlFlowLoader implements FlowLoader {
     FlowLoaderUtils.addEmailPropsToFlow(flow, props);
     props.setSource(flowFile.getName());
 
-    final List<FlowProps> flowPropsList = new ArrayList<>();
-    flowPropsList.add(new FlowProps(props));
-    flow.addAllFlowProperties(flowPropsList);
+    flow.addAllFlowProperties(ImmutableList.of(new FlowProps(props)));
 
     // Convert azkabanNodes to nodes inside the flow.
     azkabanFlow.getNodes().values().stream()
@@ -151,10 +150,10 @@ public class DirectoryYamlFlowLoader implements FlowLoader {
     if (azkabanNode.getType().equals(Constants.FLOW_NODE_TYPE)) {
       final String embeddedFlowId = flowName + Constants.PATH_DELIMITER + node.getId();
       node.setEmbeddedFlowId(embeddedFlowId);
-      final Flow flow_node = convertAzkabanFlowToFlow((AzkabanFlow) azkabanNode, embeddedFlowId,
+      final Flow flowNode = convertAzkabanFlowToFlow((AzkabanFlow) azkabanNode, embeddedFlowId,
           flowFile);
-      flow_node.setEmbeddedFlow(true);
-      this.flowMap.put(flow_node.getId(), flow_node);
+      flowNode.setEmbeddedFlow(true);
+      this.flowMap.put(flowNode.getId(), flowNode);
     }
     return node;
   }
--- a/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
@@ -19,6 +19,7 @@ package azkaban.project;
 import azkaban.Constants;
 import azkaban.flow.Edge;
 import azkaban.flow.Flow;
+import azkaban.flow.FlowProps;
 import azkaban.flow.Node;
 import azkaban.project.FlowLoaderUtils.SuffixFilter;
 import azkaban.project.validator.ValidationReport;
@@ -100,13 +101,12 @@ public class DirectoryYamlFlowLoader implements FlowLoader {
   private void convertYamlFiles(final File projectDir) {
     // Todo jamiesjc: convert project yaml file.
 
-    final File[] flowFiles = projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX));
-    for (final File file : flowFiles) {
+    for (final File file : projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX))) {
       final NodeBeanLoader loader = new NodeBeanLoader();
       try {
         final NodeBean nodeBean = loader.load(file);
         final AzkabanFlow azkabanFlow = (AzkabanFlow) loader.toAzkabanNode(nodeBean);
-        final Flow flow = convertAzkabanFlowToFlow(azkabanFlow);
+        final Flow flow = convertAzkabanFlowToFlow(azkabanFlow, azkabanFlow.getName(), file);
         this.flowMap.put(flow.getId(), flow);
       } catch (final FileNotFoundException e) {
         logger.error("Error loading flow yaml files", e);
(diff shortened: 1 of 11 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1585,7 +1587,12 @@ private void handleProjectPage(final HttpServletRequest req,
           page.add("exec", false);
         }
 
-        final List<Flow> flows = project.getFlows();
+        final List<Flow> flows = new ArrayList<>();
+        for (final Flow flow : project.getFlows()) {
```
kunkun-tang: nit: This can be made short by `project.getFlows().forEach()`

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -100,32 +101,38 @@ public ValidationReport loadProjectFlow(final Project project, final File projec
...
       } catch (final FileNotFoundException e) {
         logger.error("Error loading flow yaml files", e);
       }
     }
   }
 
-  private Flow convertAzkabanFlowToFlow(final AzkabanFlow azkabanFlow) {
-    final Flow flow = new Flow(azkabanFlow.getName());
+  private Flow convertAzkabanFlowToFlow(final AzkabanFlow azkabanFlow, final String flowName,
+      final File flowFile) {
```
chengren311: flowName is not necessary as an additional method argument since it can be obtained from azkabanFlow.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -100,32 +101,38 @@ public ValidationReport loadProjectFlow(final Project project, final File projec
...
+    final Flow flow = new Flow(flowName);
     final Props props = azkabanFlow.getProps();
     FlowLoaderUtils.addEmailPropsToFlow(flow, props);
+    props.setSource(flowFile.getName());
```
chengren311: should we set path instead of name as source?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -100,32 +101,38 @@ public ValidationReport loadProjectFlow(final Project project, final File projec
...
+
+    final List<FlowProps> flowPropsList = new ArrayList<>();
+    flowPropsList.add(new FlowProps(props));
+    flow.addAllFlowProperties(flowPropsList);
```
chengren311: one line:
addAllFlowProperties(ImmutableList.of(new FlowProps(props)));

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: For embedded flows, the flowName is encoded with delimiter ":" internally and it's different from the azkabanFlow name itself which is mapped directly from the flow yaml file. For example, an embedded flow "embedded_flow1" inside parent flow "simple_flow" would have "simple_flow:embedded_flow1" as its flowName as compared to the azkabanFlow name "embedded_flow1". This is to distinguish embedded flows with same names inside different parent flow files.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -134,40 +141,46 @@ private Flow convertAzkabanFlowToFlow(final AzkabanFlow azkabanFlow) {
     return flow;
   }
 
-  private Node convertAzkabanNodeToNode(final AzkabanNode azkabanNode) {
+  private Node convertAzkabanNodeToNode(final AzkabanNode azkabanNode, final String flowName,
+      final File flowFile) {
     final Node node = new Node(azkabanNode.getName());
     node.setType(azkabanNode.getType());
+    node.setPropsSource(flowFile.getName());
+    node.setJobSource(flowFile.getName());
```
chengren311: path or name?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -134,40 +141,46 @@ private Flow convertAzkabanFlowToFlow(final AzkabanFlow azkabanFlow) {
...
 
     if (azkabanNode.getType().equals(Constants.FLOW_NODE_TYPE)) {
-      node.setEmbeddedFlowId(azkabanNode.getName());
-      final Flow flow_node = convertAzkabanFlowToFlow((AzkabanFlow) azkabanNode);
+      final String embeddedFlowId = flowName + Constants.PATH_DELIMITER + node.getId();
+      node.setEmbeddedFlowId(embeddedFlowId);
+      final Flow flow_node = convertAzkabanFlowToFlow((AzkabanFlow) azkabanNode, embeddedFlowId,
```
chengren311: flowNode?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
jamiesjc: I'll improve it, thanks!

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -15,22 +15,82 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(FlowLoaderUtils.class);
+
+  /**
+   * Gets flow or job props from flow yaml file.
+   *
+   * @param path the flow or job path delimited by ":", e.g. "flow:subflow1:subflow2:job3"
+   * @param flowFile the flow yaml file
+   * @return the props from yaml file
+   */
+  public static Props getPropsFromYamlFile(final String path, final File flowFile) {
```
chengren311: What do you think of passing nodeBean inside instead of the original file? Otherwise, If one day we want to implement getFlowTriggerFromYamlFile, it will need to parse the yaml file again.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Name is enough because all the flow yaml files are under the same project directory. And we can retrieve the flow file from DB based on their names. We don't need to support subdirectories in the new design so there is no need to specify the path. Even today we store .job and .properties file name as the source instead of the path.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Will change, thanks!

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Yes, will rename.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
jamiesjc: Will getFlowTrigger be called together with getProps? If not, then we need to store the parsed nodeBean somewhere if you want to reuse it. But this might not work well if someone overrides the yaml file. I think reading the yaml file from DB and parsing the file is necessary since the file could potentially change. Also I think parsing the yaml file to NodeBean is straightforward and shouldn't be costly. Correct me if I'm wrong.

chengren311: '''
Will getFlowTrigger be called together with getProps
'''
No

jamiesjc: If it's really needed for optimization in the future, we can split this method into two parts. But again the main concern is the flow file itself might change, it is better to read from DB directly instead of caching the parsed value in nodeBean.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Name should be good.

PR #1552, opened 2017-11-07
Proposing CredentialProvider interface

kunkun-tang:
Creating Interface in order to let custom credential provider register user's secret keys. 

One sample use case: Today we don't have a secure way to pass on credentials into Hadoop. This interface allows Azkaban to fetch user's credential from external system (like certificate authority), and add it into Hadoop Job Context. Then submitted hadoop job will be able to use it.

Today we only have one method, but could have more in future.

This PR only proposes an interface, and the next step will be to use this interface in `HadoopSecurityManager_H_2_0` class.

Note: the changes in HadoopSecurityManager_H_2_0.java are automatically made by `save action` IntelliJ plugin.


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -161,6 +161,8 @@
 
     // enable Quartz Scheduler if true.
     public static final String ENABLE_QUARTZ= "azkaban.server.schedule.enable_quartz";
+
+    public static final String CUSTOM_CREDENTIAL_NAME = "azkaban.security.credential";
```
jamiesjc: Where would this be configured? In azkaban.properties?

kunkun-tang: No, this will be configured under jobtype/commonprivate.properties.

jamiesjc:
What is the custom credential provider? Can you give a little more detail on what you are planning to do with this new interface?

kunkun-tang:
Description Updated. Thanks. @jamiesjc

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/CredentialProvider.java
```diff
@@ -0,0 +1,31 @@
...
+
+package azkaban.security;
+
+/**
+ * Fetch and inject user secret key into custom credential object.
+ */
+public interface CredentialProvider {
+
+  /**
+   * register custom secret keys on behalf of an user into Credentials.
```
chengren311: lowercase credentials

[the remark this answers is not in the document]
kunkun-tang: corrected!

PR #1554, opened 2017-11-09
Flow 2.0 design - Resolve properties for flow executions.

jamiesjc:
1. When a flow starts, it needs to load all the flow properties. Same for a job. With the new flow YAML file, both flow and job properties are defined in a single flow YAML file for each flow. There is no need to create sub-directories as a workaround to handle per flow properties. We don't need to load properties from common properties files since the common properties would be duplicated to each flow file. (The duplication is handled in our internal plugin which uses DSL). After migrating from .job and .properties files to YAML flow files, the logic of resolving properties becomes simpler. 
2. azkaban.execapp.FlowRunner#isAzkabanFlowVersion20 is a flag to indicate whether we should interpret the flow as the new Flow 2.0 definition. Instead of creating a new table to store this info, we can just check the project_flow_files DB table during flow execution. If the YAML flow file is already uploaded to DB table project_flow_files, we know that it's interpreted as Flow 2.0. For now, we just set this flag to false and will enable this logic once DB change is rolled out.


azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -219,4 +219,10 @@ File getUploadedFlowFile(int projectId, int projectVersion, int flowVersion, Str
   int getLatestFlowVersion(int projectId, int projectVersion, String flowName)
       throws ProjectManagerException;
 
+  /**
+   * Gets all flow files that's uploaded.
```
chengren311: grammer

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -693,44 +722,95 @@ private void customizeJobProperties(final Props props) {
...
-      this.logger.error("Error loading job override property for job "
-          + node.getId());
-    }
+    if (this.isAzkabanFlowVersion20) {
+      props = loadJobPropsFromYamlFile(node);
+      if (props == null) {
+        this.logger.info("Job props loaded from yaml file is empty for job " + node.getId());
+        return props;
+      }
+    } else {
```
jamiesjc: Below code in the else condition is the same, just copied over.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
jamiesjc: Will change, thanks!

jamiesjc:
@inramana Thanks for the comments. It would be ideal to split the two paths, however, in the current code, it might not be easy to achieve that because the two paths share most of the logic in flowRunner but differ only in some places when loading properties. So I chose to split the two paths inside loading properties. The idea is the same, just have to split them in a more detailed way in multiple places instead of splitting the entire code path.

PR #1555, opened 2017-11-09
Update Utils.java

Free0xFF:
delimiter is not always comma.


HappyRay:
@Free0xFF Thanks for the contribution! 
Could you add a unit test?

PR #1556, opened 2017-11-10
Create a job setting to enable leveraging custom credential provider

kunkun-tang:
This PR is a follow-up of #1552 . We create a job setting "azkaban.job.enable.ssl". When it is enabled, we use Java reflection to initialize the credential provider object, and run the method.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -342,7 +342,7 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
     String credentialClassName = "unknown class";
       try {
         credentialClassName = props
-            .get(Constants.ConfigurationKeys.CUSTOM_CREDENTIAL_NAME);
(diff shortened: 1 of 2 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -334,6 +337,30 @@ public boolean shouldProxy() {
     return this.shouldProxy;
   }
 
+  private void registerCustomCredential(final Props props, final Credentials hadoopCred, final
+  String userToProxy) {
+    String credentialClassName = "unknown class";
+      try {
+        if (!props.containsKey(Constants.ConfigurationKeys.CUSTOM_CREDENTIAL_NAME)) {
```
reallocf: If you use props.getString on line 347 then I don't think you need this if statement because if Constants.ConfigurationKeys.CUSTOM_CREDENTIAL_NAME isn't defined then an UndefinedPropertyException would be thrown.

kunkun-tang: You are right!

PR #1558, opened 2017-11-14
Flow 2.0 design - Load flow trigger from YAML file.

jamiesjc:
Added flow trigger section in flow YAML file.
Added FlowTriggerBean and TriggerDependencyBean class as JavaBeans to load YAML files.
Provided API to get flow trigger from flow YAML file.
Todo: add validation for flow trigger.


azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -71,6 +72,7 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
           .setDependsOn(nodeBean.getDependsOn())
           .setNodes(
               nodeBean.getNodes().stream().map(this::toAzkabanNode).collect(Collectors.toList()))
+          .setFlowTrigger(toFlowTrigger(nodeBean.getTrigger()))
```
kunkun-tang: Nit: In fact, if we use build pattern, we don't need to explicitly call this method with "set{blabla}", because almost every method other than `build` should be called `set`.

azkaban-common/src/test/java/azkaban/project/NodeBeanLoaderTest.java
```diff
@@ -129,23 +139,22 @@ public void testToBasicAzkabanFlow() throws Exception {
...
-    assertThat(shellEcho.getProps().get(TYPE_COMMAND)).isEqualTo(ECHO_COMMAND);
+    final Props props = new Props();
+    props.put(Constants.NODE_TYPE, Constants.FLOW_NODE_TYPE);
+    props.put(FLOW_CONFIG_KEY, FLOW_CONFIG_VALUE);
+    validateAzkabanNode(flow, BASIC_FLOW_NAME, Constants.FLOW_NODE_TYPE, props, Arrays.asList
+        (SHELL_END, SHELL_PWD, SHELL_ECHO, SHELL_BASH), null);
+
+    final Props props1 = new Props();
+    props1.put(Constants.NODE_TYPE, TYPE_NOOP);
+    validateAzkabanNode(flow.getNode(SHELL_END), SHELL_END, TYPE_NOOP, props1, null,
```
kunkun-tang: Good consolidation

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
jamiesjc: Thanks for the suggestion. I can remove "set" in the method name.

azkaban-common/src/main/java/azkaban/project/FlowTriggerBean.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Java bean loaded from YAML file to represent a flow trigger.
+ */
+public class FlowTriggerBean {
+
+  private long maxWaitMins;
+  private Map<String, String> schedule;
```
chengren311: Why is it a map?

chengren311: why map?

azkaban-common/src/main/java/azkaban/project/FlowTriggerBean.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Java bean loaded from YAML file to represent a flow trigger.
+ */
+public class FlowTriggerBean {
+
+  private long maxWaitMins;
```
chengren311: int would be enough

azkaban-common/src/main/java/azkaban/project/FlowTriggerBean.java
jamiesjc: Because there are two key-value pairs for schedule in YAML file: 
  schedule:                    
    type: cron
    value: 0 0 1 ? * *

azkaban-common/src/main/java/azkaban/project/FlowTriggerBean.java
jamiesjc: ok

test/execution-test-data/flowtriggeryamltest/flow_trigger.flow
```diff
@@ -0,0 +1,58 @@
+---
+# Flow trigger
+trigger:
+  maxWaitMins: 5
+  schedule:
+    type: cron
+    value: 0 0 1 ? * *
+
+  triggerDependencies:
```
chengren311: can we handle the case of no triggerDependencies?

jamiesjc: Yes I can add the support for no triggerDependencies. Also do we need to support no params case?

Issue #1559, opened 2017-11-14
Error writing out logs for job on a retry attempt


juhoautio:
Got this error on the 1st retry, ie. attempt=1:

```
Duplicate entry '802364-push-tokens-daily-hive-1-0' for key 'PRIMARY'
Query: INSERT INTO execution_logs
```

Also as a result, job logs were missing for attempt #0.

I figured that the only way that this could happen is the initial attempt (0) logs were incorrectly written to the DB with with attempt=1.

I was able to verify how this happens by running `FlowRunnerTest.execRetries` with debugger, some manual edits & a suitable breakpoint:

![nayttokuva 2017-11-14 kello 10 16 39](https://user-images.githubusercontent.com/4446608/32769696-a178c7d2-c925-11e7-8ba9-81ab0798a074.png)

All related lines from the flow log:
```
14-11-2017 01:00:40 UTC push-tokens-daily-flow INFO - Submitting job 'push-tokens-daily-hive' to run.
14-11-2017 01:00:40 UTC push-tokens-daily-flow INFO - Created file appender for job push-tokens-daily-hive
14-11-2017 01:00:40 UTC push-tokens-daily-flow INFO - Attached file appender for job push-tokens-daily-hive

14-11-2017 01:00:40 UTC push-tokens-daily-flow INFO - Job Started: push-tokens-daily-hive
14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - Job push-tokens-daily-hive finished with status FAILED in 10928 seconds
14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - Job 'push-tokens-daily-hive' will be retried. Attempt 0 of 3

14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - Submitting job 'push-tokens-daily-hive' to run.
14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - Created file appender for job push-tokens-daily-hive
14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - Attached file appender for job push-tokens-daily-hive
14-11-2017 04:02:49 UTC push-tokens-daily-flow INFO - No attachment file for job push-tokens-daily-hive written.
14-11-2017 04:12:49 UTC push-tokens-daily-flow INFO - Job Started: push-tokens-daily-hive
14-11-2017 06:17:14 UTC push-tokens-daily-flow INFO - Job push-tokens-daily-hive finished with status SUCCEEDED in 7465 seconds

[...]

PR #1560, opened 2017-11-14
Fix issue #1559: Error writing out logs for job on a retry attempt

juhoautio:
Fix https://github.com/azkaban/azkaban/issues/1559.

After making `attempt` volatile, `attempt++` had to be synchronized because of this check:
http://errorprone.info/bugpattern/NonAtomicVolatileUpdate

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
@@ -20,6 +20,7 @@ import azkaban.flow.Node;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
 import azkaban.utils.TypedMapWrapper;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -28,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Base Executable that nodes and flows are based.
@@ -47,6 +49,7 @@ public class ExecutableNode {
   public static final String OUTPUT_PROPS_PARAM = "outputProps";
   public static final String ATTEMPT_PARAM = "attempt";
   public static final String PASTATTEMPTS_PARAM = "pastAttempts";
+  private final AtomicInteger attempt = new AtomicInteger(0);
   private String id;
   private String type = null;
   private volatile Status status = Status.READY;
@@ -54,7 +57,6 @@ public class ExecutableNode {
   private volatile long endTime = -1;
   private long updateTime = -1;
   private volatile boolean killedBySLA = false;
-
   // Path to Job File
   private String jobSource;
   // Path to top level props file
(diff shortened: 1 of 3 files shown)


juhoautio:
@HappyRay @chengren311 anyone to review?

juhoautio:
Or maybe @kunkun-tang, @jamiesjc ?

azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
```diff
@@ -63,7 +63,7 @@
   private Set<String> outNodes = new HashSet<>();
   private Props inputProps;
   private Props outputProps;
-  private int attempt = 0;
+  private volatile int attempt = 0;
```
kunkun-tang: Why not use AtomicInteger?

juhoautio: Yeah I don't mind, both work.. Do you? :)

kunkun-tang: In my view, synchronized (this) should be too aggressive, because it locks out the whole object?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -609,11 +609,15 @@ private void doRun() {
...
-        new EventData(finalStatus, this.node.getNestedId())), false);
-    finalizeLogFile(this.node.getAttempt());
-    finalizeAttachmentFile();
-    writeStatus();
+    try {
+      finalizeLogFile(this.node.getAttempt());
+      finalizeAttachmentFile();
+      writeStatus();
+    } finally {
+      // note that FlowRunner thread does node.attempt++ when it receives the JOB_FINISHED event
```
kunkun-tang: Do you mean there still exist possibilities that writing log to db would fail due to race condition?

azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
juhoautio: Ok, I can change it.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
juhoautio: I'm not aware of any specific case at the moment, but it's really important to guarantee that JOB_FINISHED event is fired. Otherwise the flow execution remains stuck in RUNNING state without ever finishing. For example, if writing the logs fails because of any random reason (for example DB connectivity) it's not as severe as having the flow get stuck.

azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
juhoautio: I was basically aiming at a less intrusive change. You'll see soon how much more complex the change becomes with AtomicInteger.

juhoautio:
@kunkun-tang, I've added two commits.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
kunkun-tang: Makes sense. Thanks for your contribution!

PR #1565, opened 2017-11-18
Flow 2.0 design - Override Job Properties

jamiesjc:
1. Support web servlet calls to get properties from flow YAML files and display the correct job/flow properties on UI page.
2. Support overriding job properties from UI page. Basically, read the latest flow file from project_flow_files DB table, override the properties and then insert a new row back into the DB table.


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -688,17 +688,22 @@ private void ajaxFetchJobInfo(final Project project, final HashMap<String, Objec
       return;
     }
 
-    final Props prop;
+    Props prop;
     try {
-      prop = this.projectManager.getProperties(project, node.getJobSource());
+      prop = this.projectManager.getProperties(project, flow, jobName, node.getJobSource());
```
kunkun-tang: Is this job Prop? Can we detail the name as jobProp?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -688,17 +688,22 @@ private void ajaxFetchJobInfo(final Project project, final HashMap<String, Objec
...
     } catch (final ProjectManagerException e) {
       ret.put("error", "Failed to retrieve job properties!");
       return;
     }
 
+    if (prop == null) {
```
kunkun-tang: Why necessary to new here?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -688,17 +688,22 @@ private void ajaxFetchJobInfo(final Project project, final HashMap<String, Objec
...
+      prop = new Props();
+    }
+
     Props overrideProp;
     try {
-      overrideProp = this.projectManager.getJobOverrideProperty(project, jobName);
+      overrideProp = this.projectManager
```
kunkun-tang: Is this prop still for Job? Can we override Flow Properties in the long run?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
jamiesjc: Yes, here it's job prop.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
jamiesjc: Because if prop is null, below method `prop.getKeySet()` at line 716 would throw null pointer exception.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
jamiesjc: Here this overrideProp is for job override to keep consistent with the current implementation. But you are right, we can support overriding flow properties in the same way as well.

azkaban-common/src/main/java/azkaban/flow/Flow.java
```diff
@@ -59,6 +60,7 @@ public static Flow flowFromObject(final Object object) {
     final String id = (String) flowObject.get("id");
     final Boolean layedout = (Boolean) flowObject.get("layedout");
     final Boolean isEmbeddedFlow = (Boolean) flowObject.get("embeddedFlow");
+    final Boolean isAzkabanFlowVersion20 = (Boolean) flowObject.get("azkabanFlowVersion20");
```
chengren311: why not just keeping azFlowVersion in flowObject?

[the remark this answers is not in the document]
jamiesjc: Yes, we can store the actual flowVersion so that it can be extended to support future flow versions like azkabanFlowVersion30.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -38,6 +46,81 @@
...
   private static final Logger logger = LoggerFactory.getLogger(FlowLoaderUtils.class);
 
+  /**
+   * Sets props in flow yaml file.
+   *
+   * @param path the flow or job path delimited by ":", e.g. "flow:subflow1:subflow2:job3"
+   * @param flowFile the flow yaml file
+   * @param prop the props to set
+   */
+  public static void setPropsInYamlFile(final String path, final File flowFile, final Props prop) {
```
chengren311: Will the method be called by multiple threads? if so is it thread safe?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
+    }
+    return props;
   }
 
-  public Props getJobOverrideProperty(final Project project, final String jobName)
-      throws ProjectManagerException {
-    return this.projectLoader.fetchProjectProperty(project, jobName + ".jor");
+  public Props getProperties(final Project project, final Flow flow, final String jobName,
+      final String source) throws ProjectManagerException {
+    if (Double.compare(flow.getAzkabanFlowVersion(), Constants.AZKABAN_FLOW_VERSION_2_0) == 0) {
```
chengren311: consider to put the check in the util class like FlowUtils? i guess it will be used in multiple cases? but not every one will remember to use Double.compare to compare two doubles. At least for me it's not occasionally to just use "=="  :)

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
 
-  public void setJobOverrideProperty(final Project project, final Props prop, final String jobName,
-      final User modifier)
-      throws ProjectManagerException {
-    prop.setSource(jobName + ".jor");
-    final Props oldProps =
-        this.projectLoader.fetchProjectProperty(project, prop.getSource());
+  public Props getJobOverrideProperty(final Project project, final Flow flow, final String jobName,
+      final String source) throws ProjectManagerException {
+    if (Double.compare(flow.getAzkabanFlowVersion(), Constants.AZKABAN_FLOW_VERSION_2_0) == 0) {
```
chengren311: same as above, put the check in a util method

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
+    }
   }
 
-  public void setJobOverrideProperty(final Project project, final Props prop, final String jobName,
-      final User modifier)
-      throws ProjectManagerException {
-    prop.setSource(jobName + ".jor");
-    final Props oldProps =
-        this.projectLoader.fetchProjectProperty(project, prop.getSource());
+  public Props getJobOverrideProperty(final Project project, final Flow flow, final String jobName,
```
chengren311: same here, we don't need the whole flow object.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
+    }
+  }
 
-    if (oldProps == null) {
-      this.projectLoader.uploadProjectProperty(project, prop);
+  public void setJobOverrideProperty(final Project project, final Flow flow, final Props prop,
+      final String jobName, final String source, final User modifier)
+      throws ProjectManagerException {
+    final Props oldProps;
+    if (Double.compare(flow.getAzkabanFlowVersion(), Constants.AZKABAN_FLOW_VERSION_2_0) == 0) {
```
chengren311: same as above

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
+
+      FlowLoaderUtils.setPropsInYamlFile(path, flowFile, prop);
+      this.projectLoader
+          .uploadFlowFile(flow.getProjectId(), flow.getVersion(), flowVersion + 1, flowFile);
+      if (flowFile != null && flowFile.exists()) {
+        flowFile.delete();
+      }
     } else {
-      this.projectLoader.updateProjectProperty(project, prop);
+      prop.setSource(jobName + ".jor");
```
chengren311: what is ".jor" ? can we make it as a constant variable a meaningful name?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
...
+      oldProps = this.projectLoader.fetchProjectProperty(project, prop.getSource());
+
+      if (oldProps == null) {
+        this.projectLoader.uploadProjectProperty(project, prop);
```
chengren311: if block and else block are exactly the same. Should it be uploadProjectProperty(project, oldProps)?

chengren311: BTW, you can use "? : " to reduce the if else block to one liner.

azkaban-common/src/main/java/azkaban/flow/Flow.java
```diff
@@ -48,6 +48,7 @@
 
   private boolean isLayedOut = false;
   private boolean isEmbeddedFlow = false;
+  private double azkabanFlowVersion;
```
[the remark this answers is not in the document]
jamiesjc: Currently no, you need to retrieve it from Flow. Any reason you want to get it from Project?  
The default value of azkabanFlowVersion is 0.0

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -119,6 +119,7 @@ private void convertYamlFiles(final File projectDir) {
   private Flow convertAzkabanFlowToFlow(final AzkabanFlow azkabanFlow, final String flowName,
       final File flowFile) {
     final Flow flow = new Flow(flowName);
+    flow.setAzkabanFlowVersion(Constants.AZKABAN_FLOW_VERSION_2_0);
```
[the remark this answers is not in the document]
jamiesjc: Sure, I'll add that.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Actually one is `upload` the other is `update` if you look at the two methods carefully. Anyway, I just copied this part from the original code without any change.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: I think it means job override. I can make it a constant variable.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +320,67 @@ public void updateProjectDescription(final Project project, final String descrip
     return this.projectLoader.getProjectEvents(project, results, skip);
   }
 
-  public Props getProperties(final Project project, final String source)
-      throws ProjectManagerException {
-    return this.projectLoader.fetchProjectProperty(project, source);
+  public Props getPropertiesFromFlowFile(final Flow flow, final String jobName, final String source,
```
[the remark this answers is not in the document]
jamiesjc: Agree. 
Here source is the flow file name. I think it also makes sense to have flow file name as the argument. Probably I can change the argument name to flowFileName.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Sure.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
jamiesjc: Good question! Yes it will be called by multiple threads when overriding job properties. If two threads are writing flow files into DB with the same primary key (project id, project version, flow name, flow version), then the second thread would fail in DB insertion. This is better than adding synchronization, which is blocking and could be slow. It makes sense to explicitly fail the second thread so that the user is aware of the conflict in writing to the same file and could retry again.

azkaban-common/src/main/java/azkaban/flow/Flow.java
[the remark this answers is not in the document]
chengren311: shouldn't the default value be a valid version?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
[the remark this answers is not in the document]
jamiesjc: It could be more reusable. But if we extract project id, project version, flow id, flow version from flow object, then the argument list would become relatively long (6 arguments in total), which might not be a good coding practice.
`When a function seems to need more than two or three arguments, it is likely that some of those arguments ought to be wrapped into a class of their own.` (Excerpt From: Martin, Robert C. “Clean Code: A Handbook of Agile Software Craftsmanship.” iBooks.)

chengren311:
thanks, will review it later.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -148,6 +232,33 @@ public static ValidationReport generateFlowLoaderReport(final Set<String> errors
     return report;
   }
 
+  /**
+   * Clean up temp dir.
+   *
+   * @param tempDir the temp dir
+   */
+  public static void cleanUpTempDir(final File tempDir) {
+    logger.info("Cleaning up temp directory.");
```
chengren311: would be great to log the temp dir full path as well.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -148,6 +232,33 @@ public static ValidationReport generateFlowLoaderReport(final Set<String> errors
     return report;
   }
 
+  /**
+   * Clean up temp dir.
+   *
+   * @param tempDir the temp dir
+   */
+  public static void cleanUpTempDir(final File tempDir) {
```
chengren311: might just call it cleanupdir since the method can be applied to any kind of  dir deletion.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -319,27 +321,79 @@ public void updateProjectDescription(final Project project, final String descrip
...
+      return getPropertiesFromFlowFile(flow, jobName, source, flowVersion);
+    } else {
+      return this.projectLoader
+          .fetchProjectProperty(project, jobName + Constants.JOB_OVERRIDE_SUFFIX);
+    }
+  }
 
-    if (oldProps == null) {
-      this.projectLoader.uploadProjectProperty(project, prop);
+  public void setJobOverrideProperty(final Project project, final Flow flow, final Props prop,
```
chengren311: when are the overrode properties becoming effective? Next execution of the flow or immediately?

jamiesjc: It depends on whether the particular job has already started in the current flow execution. If the override properties are set before the job starts to execute, then it will take effect immediately. If set after the job has started, then it will become effective in the next execution.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
jamiesjc: Sure.

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
jamiesjc: The temp dir path would just be some random string. Any reason why we need to log it?

jamiesjc: Actually, I think it's ok to delete this log message since it does not provide much useful info.

Issue #1566, opened 2017-11-18
azkaban.jobExecutor.utils.process.ProcessFailureException


encrylife:
command file:
type=javaprocess
java.class=AzkabanExample
working.dir=/home/ubuntu/Downloads/azkaban/azkaban-exec-server/job
but it not work
logs:
18-11-2017 22:23:44 CST command INFO - Starting job command at 1511015024964
18-11-2017 22:23:44 CST command INFO - azkaban.webserver.url property was not set
18-11-2017 22:23:44 CST command INFO - job JVM args: -Dazkaban.flowid=command -Dazkaban.execid=31 -Dazkaban.jobid=command
18-11-2017 22:23:44 CST command INFO - Building javaprocess job executor. 
18-11-2017 22:23:45 CST command INFO - No classpath specified. Trying to load classes from /home/ubuntu/Downloads/azkaban/azkaban-exec-server/job
18-11-2017 22:23:45 CST command INFO - 1 commands to execute.
18-11-2017 22:23:45 CST command INFO - cwd=/home/ubuntu/Downloads/azkaban/azkaban-exec-server/job
18-11-2017 22:23:45 CST command INFO - effective user is: test
18-11-2017 22:23:45 CST command INFO - Command: java -Dazkaban.flowid=command -Dazkaban.execid=31 -Dazkaban.jobid=command -Xms64M -Xmx256M -cp  AzkabanExample 
18-11-2017 22:23:45 CST command INFO - Environment variables: {JOB_OUTPUT_PROP_FILE=/home/ubuntu/Downloads/azkaban/azkaban-exec-server/job/command_output_3561948938160028661_tmp, JOB_PROP_FILE=/home/ubuntu/Downloads/azkaban/azkaban-exec-server/job/command_props_6040937614697965721_tmp, KRB5CCNAME=/tmp/krb5cc__ttt__command__command__31__test, JOB_NAME=command}
18-11-2017 22:23:45 CST command INFO - Working directory: /home/ubuntu/Downloads/azkaban/azkaban-exec-server/job
what happen?
I copy the 
 java -Dazkaban.flowid=command -Dazkaban.execid=31 -Dazkaban.jobid=command -Xms64M -Xmx256M -cp  AzkabanExample to commandline  it still failed,but if I remove -cp arg it work well.
what's wrong? command file error?

juhoautio:
[...]

PR #1567, opened 2017-11-18
javaprocess classpath improvements

juhoautio:
(relates to issue https://github.com/azkaban/azkaban/issues/1566)

Classpath improvements for the `javaprocess` job type:

- Tests for `javaprocess` job with empty or non-existing `classpath` property
- Fail the job with a helpful message if there's nothing to add as `-cp` arg (instead of trying to run `java` command that's doomed to fail)
- If `classpath` property exists, but has an empty value, try to scan the folder for `.jar` files (as was already the case when `classpath` property doesn't exist)
- Improve error message when `ProcessJob` setup fails: print the top-level exception message first, not the cause of it
- For some reason `getClassPathParam()` method wasn't being used. Now it is used again, with a small modification, though.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
@@ -59,7 +59,7 @@ public class JavaProcessJob extends ProcessJob {
     command += getJVMArguments() + " ";
     command += "-Xms" + getInitialMemorySize() + " ";
     command += "-Xmx" + getMaxMemorySize() + " ";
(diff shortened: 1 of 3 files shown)


juhoautio:
Could someone have a look? This is old but still valid. It would be nice if this improvement would be taken in, so that the effort wouldn't have been in vain.

juhoautio:
@burgerkingeater how about this one? The more tests the merrier..

juhoautio:
@burgerkingeater bump

burgerkingeater:
@juhoautio  sorry for the late reply, i will try to review it this week.

PR #1568, opened 2017-11-22
Validate nodebean and check job properties.

jamiesjc:
1. Validate nodebean against duplicate nodes and dependency undefined.
2. Check job properties during loading flows.
3. Added unit tests to verify the changes.


azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -149,6 +154,52 @@ public static ValidationReport generateFlowLoaderReport(final Set<String> errors
...
 
+  /**
+   * Check job properties.
+   *
+   * @param projectId the project id
+   * @param props the server props
+   * @param jobPropsMap the job props map
+   * @param errors the errors
+   */
+  public static void checkJobProperties(final int projectId, final Props props,
```
jamiesjc: The content of this method remains the same, just moved from DirectoryFlowLoader to the common util class FlowLoaderUtils, which is used by both DirectoryFlowLoader and DirectoryYamlFlowLoader.

Issue #1569, opened 2017-11-23
Startup fails: missing tables that need to be installed: quartz-tables-all


juhoautio:
- Open project in IDEA
- Run `AzkabanSingleServer`
- See it fail:

> 2017/11/23 16:47:42.235 +0200 INFO [AzkabanDatabaseSetup] [Azkaban] **The following are missing tables that need to be installed**
2017/11/23 16:47:42.235 +0200 INFO [AzkabanDatabaseSetup] [Azkaban]  **quartz-tables-all**
2017/11/23 16:47:42.235 +0200 ERROR [AzkabanWebServer] [Azkaban] **Exiting with error**.

Full console output below.

```
2017/11/23 16:47:39.390 +0200 INFO [AzkabanWebServer] [Azkaban] Starting Azkaban Server
2017/11/23 16:47:39.398 +0200 INFO [AzkabanWebServer] [Azkaban] Using conf at /Users/JuhoAutio/ideaprojects/azkaban-public/local/conf
2017/11/23 16:47:39.424 +0200 INFO [AzkabanServer] [Azkaban] Loading azkaban settings file from local/conf
2017/11/23 16:47:39.424 +0200 INFO [AzkabanServer] [Azkaban] Loading azkaban properties file
2017/11/23 16:47:39.427 +0200 INFO [AzkabanDatabaseUpdater] [Azkaban] Use scripting directory azkaban-db/src/main/sql
2017/11/23 16:47:39.428 +0200 INFO [AzkabanDatabaseUpdater] [Azkaban] Will auto update any changes.
2017/11/23 16:47:39.439 +0200 INFO [DataSourceUtils] [Azkaban] h2 DB path: /Users/JuhoAutio/ideaprojects/azkaban-public/local/h2
2017/11/23 16:47:39.470 +0200 INFO [AzkabanDatabaseSetup] [Azkaban] Searching for installed tables
2017/11/23 16:47:40.195 +0200 INFO [AzkabanDatabaseSetup] [Azkaban] Searching for table versions in the properties table
2017/11/23 16:47:40.272 +0200 INFO [AzkabanDatabaseUpdater] [Azkaban] Need to update the db.
2017/11/23 16:47:40.273 +0200 INFO [AzkabanDatabaseSetup] [Azkaban] The following are installed tables
2017/11/23 16:47:40.273 +0200 INFO [AzkabanDatabaseSetup] [Azkaban]  projects version:3.10.1
[...]

PR #1570, opened 2017-11-23
Temporary fix for startup failure (#1569)

juhoautio:
https://github.com/azkaban/azkaban/issues/1569

Issue #1569: "Startup fails: missing tables that need to be installed: quartz-tables-all"

This fix allows `AzkabanSingleServer` to start, and I was also able to run jobs after that.


juhoautio:
@kunkun-tang could you review? I can see that you added the problematic file: https://github.com/azkaban/azkaban/blob/master/azkaban-db/src/main/sql/create.quartz-tables-all.sql

Of course it would be best to fix this properly, but a quick fix like this would be useful in the meanwhile.

kunkun-tang:
Thanks for the finding and fix. I am just about to address this issue. Taking a look

juhoautio:
@kunkun-tang, I added the `@Deprecate` tag.

PR #1573, opened 2017-12-01
Enhance retry logic for sending emails.

jamiesjc:
1. Instead of only retrying for MessagingException, retry for all Exception during connecting to SMTP server and sending messages.
2. Add more logging messages to indicate sending email success or failure.
3. Increase retry count from 3 to 5.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
+++ b/azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
@@ -40,7 +40,7 @@ import org.apache.log4j.Logger;
 public class EmailMessage {
 
   private static final String protocol = "smtp";
-  private static final int MAX_EMAIL_RETRY_COUNT = 3;
+  private static final int MAX_EMAIL_RETRY_COUNT = 5;
   private static int _mailTimeout = 10000;
   private static int _connectionTimeout = 10000;
   private static long _totalAttachmentMaxSizeInByte = 1024 * 1024 * 1024; // 1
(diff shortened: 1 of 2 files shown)


jakhani:
Code change looks good to me. Just one question, why are we not passing retry count from config?

jamiesjc:
@jakhani Thanks for the comment. Yes, we can pass the retry count from config. We can add that later if we want this to be configured differently for different clusters.

azkaban-common/src/main/java/azkaban/utils/EmailMessage.java
```diff
@@ -267,8 +267,8 @@ private void retrySendMessage(final SMTPTransport t, final Message message)
       try {
         t.sendMessage(message, message.getRecipients(Message.RecipientType.TO));
         return;
-      } catch (final MessagingException sme) {
-        this.logger.error("Sending email messages failed, attempt: " + attempt, sme);
+      } catch (final Exception e) {
+        this.logger.error("Sending email messages failed, attempt: " + attempt, e);
```
chengren311: more clear logging: "Sending email messages failed, attempt: " + (attempt+1)

chengren311: should we have a sleep interval between each retry?

jamiesjc: There are sleep intervals for mailTimeout and ConnectionTimeout:
    final int mailTimeout = props.getInt("mail.timeout.millis", 30000);
    final int connectionTimeout =
        props.getInt("mail.connection.timeout.millis", 30000);

PR #1574, opened 2017-12-05
Address comments in PR# 1544

jamiesjc:
PR #1544

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -40,6 +40,7 @@ import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
 import azkaban.utils.Triple;
+import azkaban.utils.Utils;
 import com.google.common.io.Files;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
@@ -55,6 +56,7 @@ import java.util.List;
 import java.util.Map;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 
@@ -974,8 +976,8 @@ public class JdbcProjectImpl implements ProjectLoader {
   }
 
   @Override
-  public void uploadFlowFile(final int projectId, final int projectVersion, final int flowVersion,
-      final File flowFile) throws ProjectManagerException {
+  public void uploadFlowFile(final int projectId, final int projectVersion, final File flowFile,
+      final int flowVersion) throws ProjectManagerException {
     logger.info(String
         .format(
             "Uploading flow file %s, version %d for project %d, version %d, file length is [%d bytes]",
@@ -1015,31 +1017,37 @@ public class JdbcProjectImpl implements ProjectLoader {
 
   @Override
   public File getUploadedFlowFile(final int projectId, final int projectVersion,
-      final int flowVersion, final String flowName) throws ProjectManagerException {
+      final String flowFileName,
+      final int flowVersion) throws ProjectManagerException, IOException {
     final FlowFileResultHandler handler = new FlowFileResultHandler();
 
     final List<byte[]> data;
-    // Todo jamiesjc: delete the flow file after used.
-    final File file = new File(this.tempDir, flowName);
+    final File tempDir = Utils.createTempDir(this.tempDir);
+    final File file = new File(tempDir, flowFileName);
     try (final FileOutputStream output = new FileOutputStream(file);
         final BufferedOutputStream bufferedStream = new BufferedOutputStream(output)) {
       try {
         data = this.dbOperator
             .query(FlowFileResultHandler.SELECT_FLOW_FILE, handler,
-                projectId, projectVersion, flowName, flowVersion);
+                projectId, projectVersion, flowFileName, flowVersion);
       } catch (final SQLException e) {
         logger.error(e);
-        throw new ProjectManagerException("Failed to query uploaded flow file " + flowName + ".",
-            e);
+        if (tempDir.exists()) {
+          FileUtils.deleteDirectory(tempDir);
+        }
+        throw new ProjectManagerException(
+            "Failed to query uploaded flow file " + flowFileName + ".", e);
       }
 
-      try {
+      if (data != null && !data.isEmpty()) {
         bufferedStream.write(data.get(0));
-      } catch (final IOException e) {
-        throw new ProjectManagerException("Error writing flow file" + flowName, e);
       }
     } catch (final IOException e) {
-      throw new ProjectManagerException("Error creating output stream for flow file" + flowName, e);
+      if (tempDir.exists()) {
+        FileUtils.deleteDirectory(tempDir);
+      }
+      throw new ProjectManagerException(
+          "Error writing to output stream for flow file" + flowFileName, e);
     }
     return file;
   }
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -40,7 +40,6 @@ import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
 import azkaban.utils.Triple;
-import azkaban.utils.Utils;
 import com.google.common.io.Files;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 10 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -1015,31 +1017,37 @@ public void uploadFlowFile(final int projectId, final int projectVersion, final
...
   public File getUploadedFlowFile(final int projectId, final int projectVersion,
-      final int flowVersion, final String flowName) throws ProjectManagerException {
+      final String flowFileName,
+      final int flowVersion) throws ProjectManagerException, IOException {
     final FlowFileResultHandler handler = new FlowFileResultHandler();
 
     final List<byte[]> data;
-    // Todo jamiesjc: delete the flow file after used.
-    final File file = new File(this.tempDir, flowName);
+    final File tempDir = Utils.createTempDir(this.tempDir);
```
chengren311: it's azkaban utils? I'm inclined towards using public OSS/JDK libs providing the method of the same purpose. They are more trustworthy and well tested.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -1015,31 +1017,37 @@ public void uploadFlowFile(final int projectId, final int projectVersion, final
...
+    final File file = new File(tempDir, flowFileName);
```
chengren311: would be good to add some comments explaining the motivation of creating a temp dir.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -1015,31 +1017,37 @@ public void uploadFlowFile(final int projectId, final int projectVersion, final
...
 
-      try {
+      if (data != null && !data.isEmpty()) {
         bufferedStream.write(data.get(0));
-      } catch (final IOException e) {
-        throw new ProjectManagerException("Error writing flow file" + flowName, e);
       }
     } catch (final IOException e) {
-      throw new ProjectManagerException("Error creating output stream for flow file" + flowName, e);
+      if (tempDir.exists()) {
```
chengren311: code piece duplicating as above. would be good to wrap the exception handling logic into a method.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
chengren311: when will the temp dir get deleted? It's not a good idea to just delete on exit since linux has hard limit on the number of dir/file contained in a dir.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
chengren311: we also need to cover case where temp dir fails to be created for any reason. In this case might just throw an IOexception?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -1015,31 +1017,37 @@ public void uploadFlowFile(final int projectId, final int projectVersion, final
...
-            e);
+        if (tempDir.exists()) {
+          FileUtils.deleteDirectory(tempDir);
+        }
+        throw new ProjectManagerException(
+            "Failed to query uploaded flow file " + flowFileName + ".", e);
       }
 
-      try {
+      if (data != null && !data.isEmpty()) {
```
chengren311: can yaml parser handle empty file?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: Agreed. Changed to public jdk libs. 
Added the exception handling for failure to create temp dir.
Thanks!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: Refactored some of the code and handled the deletion of temp dir.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: Yaml parser can handle empty file. But we still need to explicitly check the parsed result is not null.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
jamiesjc: I've refactored the code in a new commit. Thanks for all the comments!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -1015,31 +1015,37 @@ public void uploadFlowFile(final int projectId, final int projectVersion, final
 
   @Override
   public File getUploadedFlowFile(final int projectId, final int projectVersion,
-      final int flowVersion, final String flowName) throws ProjectManagerException {
+      final String flowFileName, final int flowVersion, final File tempDir)
```
chengren311: why is tempDir an argument for this method?

jamiesjc: This is to delete the tempDir after the returned flow file is no longer used. We cannot do the deletion inside this method because we still need to reference the flow file in the tempDir.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -768,46 +772,46 @@ private Props loadJobProps(final ExecutableNode node) throws IOException {
     return props;
   }
 
-  private Props loadFlowPropsFromYamlFile() {
-    final File flowFile = getFlowFile();
-    final Props flowProps = FlowLoaderUtils.getPropsFromYamlFile(this.flow.getId(), flowFile);
-    if (flowFile != null && flowFile.exists()) {
-      flowFile.delete();
+  private Props loadPropsFromYamlFileUtil(final String path) {
```
kunkun-tang: What cause you to change the method name? To me, `loadPropsFromYaml` sounds good enough.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -768,46 +772,46 @@ private Props loadJobProps(final ExecutableNode node) throws IOException {
...
-    if (flowFile != null && flowFile.exists()) {
-      flowFile.delete();
+    try {
+      props = FlowLoaderUtils.getPropsFromYamlFile(path, getFlowFile(tempDir));
+    } catch (final Exception e) {
+      this.logger.error("Failed to get props from flow file. " + e);
+    } finally {
+      if (tempDir != null && tempDir.exists()) {
+        try {
+          FileUtils.deleteDirectory(tempDir);
```
kunkun-tang: 1). Don't understand why we need to create a temp folder to do these things. 

2). Can we find a way to merge the two try-catch to reduce redundant code?

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -35,11 +34,16 @@
  */
 public class NodeBeanLoader {
 
-  public NodeBean load(final File flowFile) throws FileNotFoundException {
+  public NodeBean load(final File flowFile) throws Exception {
+    checkArgument(flowFile != null);
```
kunkun-tang: Just another way to handle: checkArgument(flow != null && flowFile.exists());

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
jamiesjc: 1. We need to create the separate temp directory for each flow file to avoid overwriting the same file by multiple threads concurrently. And the reason why I create temp directory instead of temp file is that the file name will be interpreted as the flow name. And a temp file name is basically some random string combination, so we cannot use.  I've left some comments in azkaban.project.JdbcProjectImpl#getUploadedFlowFile.
2. Do you mean combining createTempDir and getPropsFromYamlFile in the same try-catch block? The reason I'm separating them out is for clear logging and debugging under different exceptions. But you are right they can be combined together, in this way, we will combine the logging message together as well.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
jamiesjc: I'm ok with either name.

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
jamiesjc: Will consolidate the code, thanks!

PR #1580, opened 2017-12-08
Refactor Emailer.java

chengren311:
Create a sendEmail method to be used more generally not just limited to failure/success/sla emails. This method will be used by Flow trigger to alert user on flow trigger failure with customized email subject/body.
Since currently Emailer.java can only be tested via integration test, I didn't write unit test but verified by sending a email with the new method in local mode.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -121,10 +121,10 @@ public class Emailer extends AbstractMailer implements Alerter {
       if (!this.testMode) {
         try {
           message.sendEmail();
-          logger.info("Sent SLA email message " + body);
+          logger.info("Sent email message " + body);
           this.commonMetrics.markSendEmailSuccess();
         } catch (final Exception e) {
-          logger.error("Failed to send SLA email message " + body, e);
+          logger.error("Failed to send email message " + body, e);
           this.commonMetrics.markSendEmailFail();
         }
       }
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -74,16 +74,20 @@ public class Emailer extends AbstractMailer implements Alerter {
 
     EmailMessage.setTotalAttachmentMaxSize(getAttachmentMaxSize());
 
-    this.clientHostname = props.getString(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME, props.getString("jetty.hostname", "localhost"));
+    this.clientHostname = props.getString(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME,
+        props.getString("jetty.hostname", "localhost"));
 
     if (props.getBoolean("jetty.use.ssl", true)) {
       this.scheme = HTTPS;
-      this.clientPortNumber = Integer.toString(props.getInt(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_SSL_PORT, props.getInt("jetty.ssl.port",
-          Constants.DEFAULT_SSL_PORT_NUMBER)));
+      this.clientPortNumber = Integer.toString(props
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -74,16 +74,20 @@ public Emailer(final Props props, final CommonMetrics commonMetrics) {
...
 
-    this.clientHostname = props.getString(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME, props.getString("jetty.hostname", "localhost"));
+    this.clientHostname = props.getString(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME,
+        props.getString("jetty.hostname", "localhost"));
 
     if (props.getBoolean("jetty.use.ssl", true)) {
       this.scheme = HTTPS;
-      this.clientPortNumber = Integer.toString(props.getInt(ConfigurationKeys.AZKABAN_WEBSERVER_EXTERNAL_SSL_PORT, props.getInt("jetty.ssl.port",
-          Constants.DEFAULT_SSL_PORT_NUMBER)));
+      this.clientPortNumber = Integer.toString(props
```
chengren311: auto formatted.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -113,10 +124,10 @@ private void sendSlaAlertEmail(final SlaOption slaOption, final String slaMessag
       if (!this.testMode) {
         try {
           message.sendEmail();
-          logger.info("Sent SLA email message " + slaMessage);
+          logger.info("Sent email message " + body);
```
jamiesjc: Would it be helpful to keep logging messages for sending sla email somewhere for debugging purpose?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -100,10 +104,17 @@ public Emailer(final Props props, final CommonMetrics commonMetrics) {
...
+        "SLA violation for " + getJobOrFlowName(slaOption) + " on " + getAzkabanName();
     final List<String> emailList =
         (List<String>) slaOption.getInfo().get(SlaOption.INFO_EMAIL_LIST);
+    sendEmail(emailList, subject, slaMessage);
+  }
+
+  /**
+   * Send an email to the specified email list
+   */
+  public void sendEmail(final List<String> emailList, final String subject, final String body) {
```
jamiesjc: We already have azkaban.utils.EmailMessage#sendEmail method. Consider another method name to avoid confusion?

chengren311: it's fine since there're in different class and different argument. In addition, I couldn't think a better name than sendEmail :)

azkaban-common/src/main/java/azkaban/utils/Emailer.java
chengren311: we can still grep "Sent email message" from the log to debug.

jamiesjc: As you mentioned, this would be a more general method used also in other places like the flow trigger. How could we distinguish whether the email message is for sla or flow trigger then?

chengren311: added logging message in sendSlaAlertEmail

HappyRay:
"Since currently Emailer.java can only be tested via integration test, I didn't write unit test but verified by sending a email with the new method in local mode."

Not having automated tests for new code is ok for a PR only if tests will be added shortly after.

I would suggest refactoring email component. This piece can be improved anyway. @jamiesjc and I had discussions about it. She planned to do it but it hasn't become a priority.

Since you modified the code, you are on the hook now. :)

chengren311:
@HappyRay https://github.com/kirviq/dumbster just did some research on unit testing on sending emails, this could be helpful.

HappyRay:
Other choices I found:

Java mail testing:

http://www.icegreen.com/greenmail/
https://stackoverflow.com/questions/1033076/mail-server-for-javamail-testing

HappyRay:
To make the unit tests effective, I think the mail module needs to be refactored.

PR #1583, opened 2017-12-14
use java lib instead of external sh commands to do recursive hard linking

chengren311:
Linux has restriction on shell argument length, it's likely the command to do hard linking project dir of a lot of subdir and files will exceed the restriction, causing flow setup failure. The fix is replacing the logic with java native file API.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
+++ b/azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
@@ -79,7 +79,6 @@ public class FileIOUtilsTest {
     file5.createNewFile();
     file6.createNewFile();
     file7.createNewFile();
-    createBigDir(this.baseDir.getAbsolutePath());
 
     byte[] fileData = new byte[]{1, 2, 3};
     FileOutputStream out = new FileOutputStream(file1);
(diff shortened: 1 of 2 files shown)


HappyRay:
Please include tests if they don't exist already. 
Also please explain how you verified that this new code will scale with the number of the files per project, unlike the existing code.

chengren311:
@HappyRay thanks, there're existing unit tests for that.

reallocf:
Can we create a test that tries to link a file with an impossibly long name? The ARG_MAX on the systems is 2621440 - if we generate a string with a name longer than that it can help handle regressions.. or maybe there's better for our integration testing suite?

chengren311:
@reallocf that's a good idea. I added the associated unit test.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
```diff
@@ -38,16 +39,47 @@
...
+   * exceed the allowed limit.
+   */
+  private void createBigDir(final String path) throws IOException {
+    final String verylongprefix =
+        "123123123123123123123123123123123113123111111111111111111111111111111111111111111111111111"
+            + "111111111111111111111111111111111111111111111111111111111111111111111111111111111111"
+            + "11231312312312312312313121111111111111111111111111111111111111111111111111";
+
+    for (int i = 1; i <= 400; i++) {
+      final File tmpDir = new File(path + "/" + verylongprefix + "dir" + i);
```
HappyRay: Would the new Path API be easier to use than the old File API?

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
```diff
@@ -38,16 +39,47 @@
...
+    final File file7 = new File(file4.getAbsolutePath() + "/c2.out");
+
     file1.createNewFile();
     file2.mkdir();
     file3.createNewFile();
+    file4.mkdir();
+    file5.createNewFile();
+    file6.createNewFile();
+    file7.createNewFile();
+    createBigDir(this.baseDir.getAbsolutePath());
```
HappyRay: Thanks for improving the tests.

I'd suggest a separate test for the special long argument case. This will keep the tests more focused and intention clearer.

chengren311: updated

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
chengren311: my personal feeling is it's slightly more convenient since no string concatenation needs to be done. I replaced a few of them to new API to get the taste of it but not all since there are too much existing usage of old API..

chengren311:
@HappyRay  let me know if you have more feedback.

PR #1586, opened 2017-12-19
Add azkabanFlowVersion in executableFlow

jamiesjc:
Moved azkabanFlowVersion from FlowRunner to ExecutableFlow. 
Added default value for azkabanFlowVersion.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/TypedMapWrapper.java
+++ b/azkaban-common/src/main/java/azkaban/utils/TypedMapWrapper.java
@@ -91,6 +91,24 @@ public class TypedMapWrapper<K, V> {
     }
   }
 
+  public Double getDouble(final K key) {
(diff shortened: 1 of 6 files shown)


jamiesjc:
@kunkun-tang Updated the description, thanks!

azkaban-common/src/main/java/azkaban/utils/TypedMapWrapper.java
```diff
@@ -91,6 +91,24 @@ public Long getLong(final K key, final Long defaultVal) {
     }
   }
 
+  public Double getDouble(final K key) {
```
kunkun-tang: I studied this class `TypeMapWrapper`, and it is used for metadata wrapping. Java Generics is not necessary, as AZ always initialize a <String, Object> instance. We may refactor it in future.

PR #1590, opened 2017-12-21
Enforcing num of concurrent runs quota given a Flow Executable

kunkun-tang:
A couple of production issues warned us that we should enforce the MAX number of concurrent executions given a flow executable. In case people accidentally schedule flows per minute or endlessly submit flows from the client side, this PR proposes to implement a simple quota to prevent it happening.

Unit Test is added.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -67,6 +67,8 @@ public class ExecutorManager extends EventHandler implements
 
   public static final String AZKABAN_USE_MULTIPLE_EXECUTORS =
       "azkaban.use.multiple.executors";
+  public static final String AZKABAN_MAX_CONCURRENT_RUNS_ONEFLOW =
+      "azkaban.max.concurrent.runs.oneflow";
   static final String AZKABAN_EXECUTOR_SELECTOR_FILTERS =
       "azkaban.executorselector.filters";
   static final String AZKABAN_EXECUTOR_SELECTOR_COMPARATOR_PREFIX =
@@ -79,8 +81,6 @@ public class ExecutorManager extends EventHandler implements
       "azkaban.activeexecutor.refresh.milisecinterval";
   private static final String AZKABAN_ACTIVE_EXECUTOR_REFRESH_IN_NUM_FLOW =
       "azkaban.activeexecutor.refresh.flowinterval";
-  private static final String AZKABAN_MAX_CONCURRENT_RUNS_ONEFLOW =
-      "azkaban.max.concurrent.runs.oneflow";
   private static final int DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW = 30;
   private static final String AZKABAN_EXECUTORINFO_REFRESH_MAX_THREADS =
       "azkaban.executorinfo.refresh.maxThreads";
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -79,6 +79,7 @@
       "azkaban.activeexecutor.refresh.milisecinterval";
   private static final String AZKABAN_ACTIVE_EXECUTOR_REFRESH_IN_NUM_FLOW =
       "azkaban.activeexecutor.refresh.flowinterval";
+  private static final int AZKABAN_MAX_NUM_CONCURRENT_FLOW = 30;
```
chengren311: can we make this configurable?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -978,6 +980,10 @@ public String submitExecutableFlow(final ExecutableFlow exflow, final String use
             throw new ExecutorManagerException("Flow " + flowId
                 + " is already running. Skipping execution.",
                 ExecutorManagerException.Reason.SkippedExecution);
+          } else if (running.size() > AZKABAN_MAX_NUM_CONCURRENT_FLOW) {
+            throw new ExecutorManagerException("Flow " + flowId
+                + " has more than 30 concurrent runs. Skipping",
```
chengren311: 30 in log message should be replace by AZKABAN_MAX_NUM_CONCURRENT_FLOW.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: is 30 too high?

kunkun-tang: Made it configurable. 30 might hight, but we may set a high number initially, and decrease it gradually, in case break something.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
kunkun-tang: Done.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -124,6 +127,10 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader loader,
     this.loadRunningFlows();
 
     this.queuedFlows = new QueuedExecutions(azkProps.getLong(AZKABAN_WEBSERVER_QUEUE_SIZE, 100000));
+
+    // The default threshold is set to 30 for now, in case some users are affected. We may
+    // decrease this number in future, to better prevent DDos attacks.
+    this.maxConcurrentRunsOneFlow = azkProps.getInt(AZKABAN_MAX_NUM_CONCURRENT_FLOW, 30);
```
chengren311: make 30 a variable

chengren311: if a negative number or 0 is set, should we throw exception?

kunkun-tang: Added a static variable to place the default number.
Checking number validity. we may bring up a util function to cover all similar cases. Should refactor in future.

azkaban-common/src/test/java/azkaban/executor/ExecutorManagerTest.java
```diff
@@ -365,6 +389,9 @@ private void testSetUpForRunningFlows() throws Exception {
     //so that flows will be dispatched to executors.
     this.props.put(ExecutorManager.AZKABAN_QUEUEPROCESSING_ENABLED, "true");
 
+    // allow two concurrent runs give one Flow
+    this.props.put("azkaban.max.concurrent.runs.oneflow", 2);
```
chengren311: should it be AZKABAN_MAX_CONCURRENT_RUNS_ONEFLOW instead of the hard coded string?

kunkun-tang: Fixed it!

PR #1594, opened 2018-01-01
template-based emails

mtrna:
current master: content and layout of Azkaban emails is hard-coded, emails contain only small subset of information relevant to flow failures
proposed change: expand the amount of information in the default setting, provide all information via configurable templates; keep current email creator as fallback solution in case of mis-configuration or failure; no extra dependencies added
expected benefits: email layouts are decoupled from code and configurable by Azkaban ~user~ admin, more information is provided by default (e.g. flow parameters)

your feedback is welcome!


jamiesjc:
@mtrna Thanks for contributing to Azkaban! I'm reviewing your code now.
Meanwhile, please take a look at the style guides in contributing page: https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md
We've been following some specific coding style. You can enable intellij's save actions plugin to auto-format your code.

mtrna:
Hi @jamiesjc ! Thanks for reviewing the pull request and link to the style guide!
If you don't mind I would wait until there is consensus about the functional part and update the formatting as the last step before the eventual merge. Would that work for you? :)

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -65,7 +66,7 @@
   private Integer pipelineExecId = null;
   private Integer queueLevel = 0;
   private String concurrentOption = CONCURRENT_OPTION_IGNORE;
-  private String mailCreator = DefaultMailCreator.DEFAULT_MAIL_CREATOR;
+  private String mailCreator = TemplateBasedMailCreator.NAME;
```
jamiesjc: I don't think it's a good idea to change default value of mailCreator to "template-based". If templateBasedMailCreator is not registered, then when you get mailCreator, you are actually getting the defaultMailCreator, but the name would be "template-based", right?

azkaban-common/src/main/java/azkaban/executor/mail/TemplateBasedMailCreator.java
```diff
@@ -0,0 +1,211 @@
...
+  public static TemplateBasedMailCreator fromPath(String emailTemplatesPath) throws Exception {
+    return new TemplateBasedMailCreator(
+      readFile(emailTemplatesPath + "/errorEmail.html"),
+      readFile(emailTemplatesPath + "/firstErrorEmail.html"),
+      readFile(emailTemplatesPath + "/successEmail.html")
+    );
+  }
+
+  public TemplateBasedMailCreator(String errorTemplate, String firstErrorTemplate, String successTemplate) throws Exception {
+    logger.info("TemplateBasedMailCreator(errorTemplate=" + errorTemplate + ", firstErrorTemplate=" + firstErrorTemplate + ", successTemplate=" + successTemplate + ")");
```
jamiesjc: Do you think it's necessary to print out all the templates in the server log? We can just view them in .html files, right?

azkaban-common/src/main/java/azkaban/flow/Flow.java
```diff
@@ -42,7 +44,7 @@
   private int numLevels = -1;
   private List<String> failureEmail = new ArrayList<>();
   private List<String> successEmail = new ArrayList<>();
-  private String mailCreator = DefaultMailCreator.DEFAULT_MAIL_CREATOR;
+  private String mailCreator = TemplateBasedMailCreator.NAME;
```
jamiesjc: Same comments as above.

azkaban-common/src/main/java/azkaban/executor/mail/TemplateBasedMailCreator.java
mtrna: that's a fair point, fixing...

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
mtrna: correct; will change it to obtain most recent creator from the `MailCreatorRegistry`, ok?

jamiesjc:
Hi @mtrna, I've tried out your new template-based email changes locally and was able to view the emails in new contents. 
As you mentioned, one of the main benefits of the template-based emails is that it is configurable by users. It seems to me that the new email templates are configured on the server side, not from the Azkaban user side. Can you elaborate more on how an Azkaban user can actually configure the email templates?
If a user can configure the email template as they want, will it cause security problems? Some companies have regulations on data privacy protection. If user misconfigures the email content which contains private user data, it might cause security issues.

mtrna:
fixed the tests for Travis

mtrna:
Hi @jamiesjc! You are right, only the users with access to the configuration of Azkaban can configure the templates. I have edited the original announcement, to avoid more confusion.
Regarding security: all the data accessible via the email template is already available via GUI/REST. `TemplateBasedMailCreator` in this PR provides only data strictly relevant to the success/failure of a job. Email templates are not intended to carry any private data. Notifications can still be sent to any email address.

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
```diff
@@ -71,11 +85,41 @@ public static Props loadProps(final String[] args, final OptionParser parser) {
...
       }
     } else {
-      logger
-          .info("Conf parameter not set, attempting to get value from AZKABAN_HOME env.");
+      logger.info("Conf parameter not set, attempting to get value from AZKABAN_HOME env.");
       azkabanSettings = loadConfigurationFromAzkabanHome();
     }
 
+    try {
+      if (options.has(emailDirectory)) {
```
jamiesjc: There are some code duplication in load conf and load email. Can we consolidate the code?

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
```diff
@@ -71,11 +85,41 @@ public static Props loadProps(final String[] args, final OptionParser parser) {
...
+        if (!dir.exists()) {
+          logger.error("email templates: path " + path + " does not exist.");
+        } else if (!dir.isDirectory()) {
+          logger.error("email templates: path " + path + " is not a directory.");
+        } else {
+          MailCreatorRegistry.registerCreator(TemplateBasedMailCreator.fromPath(path), true);
+        }
+      } else {
+        final String azkabanHome = System.getenv("AZKABAN_HOME");
+        final File emailPath = new File(azkabanHome, Constants.DEFAULT_EMAIL_TEMPLATE_PATH);
```
jamiesjc: I think it's better to check azkabanHome first. If it's null, then we don't create the new emailPath.

jamiesjc: I doubt whether AZKABAN_HOME will be used since conf and email should always be passed in the argument list. If that's the case, I don't even think we need to check AZKABAN_HOME.

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
jamiesjc: From my observation, if you register template-based mailCreator first and it is set in flow object and stored in DB. Next time when you restart the server and load the flow objects from DB, even if you don't register template_based mailCreator, the String mailCreator will still be "template-based".
I would prefer simpler approach which is to keep mailCreator as "default". Only set it to template-based when user explicitly sets it in executionOptions from web UI.

jamiesjc:
I like the idea of separating email content from the code. And it benefits Azkaban admins if user requests any change in the email content. But so far we haven't had any request in the email content change yet. It is nice to have, but the overhead if that we need to deploy and maintain the email configuration on all servers. I'd like to understand your use case better and evaluate the impact of this feature.

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
mtrna: fair point, adapting the code to what you are proposing

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
mtrna: we can; updating...

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
mtrna: hi @jamiesjc ! If I understand you correctly, we discuss the case when a user e.g. schedules a regular execution of a flow and the executor server somehow afterwards loses the emailCreator, is that correct? It seems UI option does not help in that case either (user can chose from a list of creators available at time T1, but the server alters its configuration at time T2>T1). What are our options?
From the other side: is there a good reason for having emailCreator in an `ExecutionOption`? Maybe it should be decided on the level of the executor server once for all executions during the initialization step (and at the time of load of the email creators). Was it considered?
My intention was not to let executing user chose the layouts ad hoc, but rather to provide the tool to evolve the content of emails outside the code-base.

mtrna:
hi @jamiesjc ! Mailbox, unlike Azkabans themselves, can be accessed with lower security barrier in the environment where I work. I would like the emails to become a helpful source of information about the status of the jobs (before I start a more involved investigation). The information I am frequently lacking is the flow parameters, concurrency option and possibly others. I do not want to impose this change on other adopters of Azkaban though. I would prefer to have a reasonable default template available in the installation, and give others the freedom to chose whatever they find useful for their use case, e.g. inline guidelines, add links to their internal operation books, change the layout.

jamiesjc:
@mtrna Sorry I didn't get a chance to review your PR this week. I think you're right, it would benefit users if we could provide them more detailed info about flow parameters and execution options in the email. We'd like to keep the current email content as default, but to provide the option of template-based emails. Users should have the freedom to choose their own.
I'll continue reviewing your code when I have time, thanks for the patience!

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
jamiesjc: @mtrna Your proposal sounds reasonable to me. 
Now we don't have the UI support for users to specify emailCreator like other execution options. We could provide them with the flexibility to choose from a list of emailCreators in the future but I don't think it's quite necessary for now.
You are right the execution options only apply to the following execution once, and next time when you submit the flow again it will disappear. But if you schedule the flow instead, the execution options will persist with every scheduled execution.  
I'm in favor of the logic to decide the emailCreator on the level of executor during initialization step. If email templates are configured, then we use template-based emailCreator, otherwise we use the default emailCreator. 
@HappyRay Does it sound like a good approach?

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
```diff
@@ -48,38 +54,71 @@ public static Props getAzkabanProperties() {
...
+
+    final Optional<Path> emailPath;
+    if (options.has(emailDirectory)) {
+      final String emailHome = options.valueOf(emailDirectory);
+      logger.info("Email parameter set, attempting to get value from " + emailHome);
+      emailPath = azkabanPath("email template", emailHome);
+    } else {
+      logger.info("Email parameter not set, attempting to get value from AZKABAN_HOME=" +
+          azkabanHome +", DEFAULT_EMAIL_TEMPLATE_PATH=" + Constants.DEFAULT_EMAIL_TEMPLATE_PATH + ".");
+      emailPath = azkabanPath("email template", azkabanHome, Constants.DEFAULT_EMAIL_TEMPLATE_PATH);
```
jamiesjc: Consider simplify some logging messages? Info like Constants.DEFAULT_CONF_PATH and Constants.DEFAULT_EMAIL_TEMPLATE_PATH  might not have to be printed out every time.

azkaban-common/src/main/java/azkaban/executor/mail/MailCreatorRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+
+  static {
+    recommendedCreator = defaultCreator = new DefaultMailCreator();
+    registerCreator(defaultCreator, true);
+  }
+
+  public static void registerCreator(final MailCreator creator, final boolean recommend) {
+    logger.info("registering mail creator " + creator.getName() + ", recommended = " + recommend);
+    registeredCreators.put(creator.getName(), creator);
+    if (recommend) {
```
jamiesjc: I don't fully understand why you need to introduce the recommendedCreator. 
It seems everywhere this registerCreator method is called, the recommend flag passed is always true. In what scenario would you set the recommend flag to false? 
If we could decide the creator during server initialization like you proposed, would it be possible to simplify the logic?

azkaban-common/src/main/java/azkaban/server/AzkabanServer.java
mtrna: This message is logged during the loading of properties, which happens during the initialization of the `AzkabanServer`. I am updating the PR.

azkaban-common/src/main/java/azkaban/executor/mail/MailCreatorRegistry.java
mtrna: The discussion I can see here is about the clarity of intent/design. The `AzkabanServer` class is currently the only one that registers a mail creator. It is hard for me to foresee whether there can be another such component and what would be the desired logic to satisfy their potential claims. The flag is one way to give registering side some control. A more straightforward approach would be that the last registered mail creator wins/is recommended for the rest of the system to use.
I am updating the PR to show this simpler approach would look.

mtrna:
rebased to the latest master

jamiesjc:
Hi @HappyRay,  @mtrna and I had some discussions about moving the emailCreator from execution level to server level.
I'm in favor of his proposal to decide the emailCreator on the level of executor during initialization step. If email templates are configured, we use template-based emailCreator, otherwise we use the default emailCreator. We don't have to support configuring emailCreator in ExecutionOptions since it's not supported anyway today on web UI. 
You can check our past discussions for more details. I'd like to hear your advise as well, thanks!

mtrna:
rebase to the latest master...

mtrna:
updated the PR: mail creator settings are read from MailCreatorRegistry, instead of ExecutionOption
@jamiesjc @HappyRay : please provide feedback ;)

jamiesjc:
@mtrna Sorry for not responding to your new change. I hope we can get your PR in asap. But now might not be the right timing since there is also some refactoring work going on recently on email part. We need time to fully understand the email part ourselves. I hope you are still interested in reopening this PR in the future. :)

mtrna:
@jamiesjc No problem. I am still interested. :)

PR #1596, opened 2018-01-02
Quick fix for failing cron tests due to new year. More thorough inves…

inramana:
…tigation needed for better fix.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/utils/UtilsTest.java
+++ b/azkaban-common/src/test/java/azkaban/utils/UtilsTest.java
@@ -60,12 +60,12 @@ public class UtilsTest {
 
     final DateTimeZone timezone = DateTimeZone.getDefault();
     Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * *", timezone));
-    Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * * 2017", timezone));
+    Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * * 2018", timezone));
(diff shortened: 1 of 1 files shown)


azkaban-common/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -60,12 +60,12 @@ public void testValidCronExpressionV() {
 
     final DateTimeZone timezone = DateTimeZone.getDefault();
     Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * *", timezone));
-    Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * * 2017", timezone));
+    Assert.assertTrue(Utils.isCronExpressionValid("0 0 3 ? * * 2018", timezone));
```
chengren311: can we use something like getCurrentYear instead of hard coded year?

HappyRay: I think it depends on the intention of this test.

The more likely culprit I think is the use of new Date() in 

https://github.com/HappyRay/azkaban/blob/b13ecc2854c0f28ba731fffb1586679329ed47ac/az-core/src/main/java/azkaban/utils/Utils.java#L617
cronExecutionTime.getNextValidTimeAfter(new Date()) == null)

This call doesn't allow the time to be controlled in the tests.

kunkun-tang: We have to fix it again at the same time next year. Should we use a high number replacement? like "0 0 3 ? * * 2022".

Issue #1597, opened 2018-01-03
Issue with deleting job property from UI


jamiesjc:
When I try to delete job property from web UI, the deleted property still shows up on the job properties page. The job override property is successfully updated in DB but It's not getting reflected correctly on UI page and this could be misleading.

PR #1598, opened 2018-01-03
Fix deleting job properties issue on web UI.

jamiesjc:
This is to fix issue: https://github.com/azkaban/azkaban/issues/1597

On job properties page, the properties are a combination of original job properties and jobOverride properties. This does not work when we delete the property from original ones. In project_properties DB table, the jobOverride properties (.jor) contains all the job properties instead of just the delta of changed properties.
It makes more sense to display the jobOverride properties if there is any. Otherwise, we just display the original job properties.


jamiesjc:
Thanks for the comment. I tested the change on localhost by manually deleting the properties from web UI.

HappyRay:
@kunkun-tang Thanks for the review and helping us improve the quality of our changes which include change descriptions.

PR #1600, opened 2018-01-09
adding job logger to credential instance

kunkun-tang:
In order to allow users to debug credential issues easily, this PR proposes adding job logger into the credential instance, so that related log messages will be shown in Azkaban web UI. 

Notice that this PR is backward compatible. In the credential client code side, we have both 2-parameter and 3-parameter constructor. If we need to roll back this commit, it will not fail.

Tested in staging cluster.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -346,17 +346,14 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
         logger.info("custom credential class name: " + credentialClassName);
         final Class credentialClass = Class.forName(credentialClassName);
 
-        final Constructor[] constructors = credentialClass.getConstructors();
-        for (final Constructor constructor : constructors) {
-          if (constructor.getParameterCount() == 3) {
-            final CredentialProvider customCredential = (CredentialProvider) constructors[1]
-                .newInstance(hadoopCred, props, jobLogger);
-            customCredential.register(userToProxy);
-            return;
-          }
-        }
-        logger.error("cannot find the appropriate credential constructor.");
-        throw new Exception("launching credential instance failed.");
+        // The credential class must have a constructor accepting 3 parameters, Credentials,
+        // Props, and Logger in order.
+        Constructor constructor = credentialClass.getConstructor (new Class[]
+            {Credentials.class, Props.class, Logger.class});
+        final CredentialProvider customCredential = (CredentialProvider) constructor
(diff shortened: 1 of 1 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -338,18 +338,25 @@ public boolean shouldProxy() {
...
-        final Constructor[] constructors = metricsClass.getConstructors();
-        final CredentialProvider customCredential = (CredentialProvider) constructors[0]
-            .newInstance(hadoopCred, props);
-        customCredential.register(userToProxy);
+        final Class credentialClass = Class.forName(credentialClassName);
+
+        final Constructor[] constructors = credentialClass.getConstructors();
+        for (final Constructor constructor : constructors) {
+          if (constructor.getParameterCount() == 3) {
+            final CredentialProvider customCredential = (CredentialProvider) constructors[1]
```
jamiesjc: Should it be (CredentialProvider) constructor instead of (CredentialProvider) constructors[1]? Otherwise there is no point in comparing the constructors in the loop.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -338,18 +338,25 @@ public boolean shouldProxy() {
...
-
-        final Constructor[] constructors = metricsClass.getConstructors();
-        final CredentialProvider customCredential = (CredentialProvider) constructors[0]
-            .newInstance(hadoopCred, props);
-        customCredential.register(userToProxy);
+        final Class credentialClass = Class.forName(credentialClassName);
+
+        final Constructor[] constructors = credentialClass.getConstructors();
+        for (final Constructor constructor : constructors) {
+          if (constructor.getParameterCount() == 3) {
```
jamiesjc: Is it possible to check the parameter class type and make sure it includes the Logger as well?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
kunkun-tang: Good catch! I removed the for loop and only look for the constructor accepting three parameters.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
kunkun-tang: Yes, it is feasible. See my second commit.

PR #1604, opened 2018-01-17
Fixing stdout/err redirect loop (Part I)

reallocf:
We ran into an issue on a test cluster (holdem4) where Azkaban was in a degraded state. Upon looking at the stack trace we found what looked like a loop occurring between the StdOutErrRedirect class and log4j. It appears log4j was throwing an error to stderr which was being redirected by this class back to log4j which then threw another error and so on.

The fix here is to look through the stack trace to see if there's evidence of looping (i.e. the same write method being in the call stack multiple times). If this is true, unbind stdout and stderr and write the message to stderr so Azkaban can move forward.

I tested this by creating a loop between the StdOutErrRedirect class and itself via calling System.err.println("Something.") from the write method. I'm not sure if there's a great way to unit test - but I'm open to suggestions! I'll think some more about it before committing.

Finally, there's a case to be made that Azkaban SHOULD fail in situations like this. But, given that Azkaban is a stateful system where failure would lead to data loss and/or missed jobs, it's probably better for there to be a way to recover the thread (allowing the OS to deal with figuring out where to output the message) and keep going.

---
The plan has changed, though I left the above message to retain the first attempt.

Now, this PR simply cleans up all old stdout/stderr messages and turns them into logger messages.

In subsequent PRs, I will:
~~1) Clean up old stdout/stderr messages~~
2) Handle errors so they are also output to logger
3) Remove StdOutErrRedirect class entirely

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -60,12 +60,13 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.lang.StringEscapeUtils;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ExecutorServlet extends LoginAbstractAzkabanServlet {
 
-  private static final Logger logger = Logger.getLogger(ExecutorServlet.class.getName());
+  private static final Logger logger = LoggerFactory.getLogger(ExecutorServlet.class.getName());
   private static final long serialVersionUID = 1L;
   private WebMetrics webMetrics;
   private ProjectManager projectManager;


azkaban-common/src/main/java/azkaban/utils/StdOutErrRedirect.java
```diff
@@ -52,7 +61,40 @@ public LogStream(final OutputStream out, final Level level) {
 
     // Underlying mechanism to log to log4j - all print methods will use this
     private void write(final String string) {
-      logger.log(this.level, string);
+      // If logs are looping, write message to system's stderr then rebind.
+      if (logIsLooping()) {
```
HappyRay: How big is the overhead to do this check on every log?

reallocf: Relatively large: https://stackoverflow.com/questions/2347828/how-expensive-is-thread-getstacktrace

I can't really think of a better way to go about this though. Do you have any ideas? I mean, we could run this check randomly with like a 5% chance. But that seems like it's starting to get ridiculous.

Honestly I don't know if there's a good way to recover from these sort of events short of restarting Azkaban. My dream would be able to hard fail Azkaban safely but it holds too much state in memory for that to be an option.

What do you think? Do you have any other ideas?

reallocf:
After talking with @HappyRay offline, I'm going about this in a different way. Instead of making sure that StdOutErrRedirect can detect and handle these kinds of loops (which adds both an expensive system call in getStackTrace and complexity in logging overall) I'm cleaning up everything that writes to stdout/stderr and removing the StdOutErrRedirect overall.

First (this PR) I'll fix all stdout/stderr messages so they write to a logger. Because this is already the best practice on the team there were only a few number that slipped through a few years ago.

Second (next PR) I'll work on how to make sure that error messages are sent to the logger. Once this is done, everything that I've seen being redirected through StdOutErrRedirect will be proceeding to the logger on its own.

Finally (third PR) I'll remove StdOutErrRedirect because it will no longer be needed.

This should be a simpler design and should avoid this bug.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowBase.java
```diff
@@ -35,7 +36,8 @@
   public static final String PROPERTIES_PARAM = "properties";
   public static final String SOURCE_PARAM = "source";
   public static final String INHERITED_PARAM = "inherited";
-
+  private static final Logger LOGGER =
+      Logger.getLogger(ExecutableFlowBase.class.getName());
```
jamiesjc: Consider using lower case for logger?

HappyRay:
Thanks for this work!

For new code, I would suggest slf4j as documented in

https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md
"Use slf4j instead of log4j.

Use the form:

private static final Logger logger = LoggerFactory.getLogger($CLASS_NAME$.class);
"
===
slf4j provides a more efficient API called  parameterized logging. 
Try to use that format.
see
https://stackoverflow.com/questions/30975026/log4j-implicit-string-formatting

I think we should update the guide to standardize on using parameterized logging. What do you think?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -65,8 +65,7 @@
 
 public class ExecutorServlet extends LoginAbstractAzkabanServlet {
 
-  private static final Logger LOGGER =
-      Logger.getLogger(ExecutorServlet.class.getName());
+  private static final Logger logger = Logger.getLogger(ExecutorServlet.class.getName());
```
chengren311: should we change log4j to slf4j?

reallocf: Do you think we should change them for all of the loggers in this PR? Do you know of any risks involved with changes like this?

chengren311: It's more like low hanging fruit given you are modifying this piece already. I can't think of any risk involved considering the usage of the logger in ExecutorServlet.

reallocf: ^ cool - will fix :+1:

HappyRay:
Thanks for the update.

It looks fine to me.

It would be great to convert to parameterized logging as mentioned earlier. But it can be done later or in a separate commit. Since this is legacy code, it's up to you if you would like to make the improvement.

PR #1606, opened 2018-01-19
Default user.to.proxy to submit user.

jamiesjc:
1. If user.to.proxy property is not set, default it to submit user.
2. Add more detailed logs to direct users to add Proxy User at project permissions page.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -17,6 +17,7 @@
 package azkaban.execapp;
 
 import azkaban.Constants;
+import azkaban.Constants.JobProperties;
 import azkaban.event.Event;
 import azkaban.event.EventData;
 import azkaban.event.EventHandler;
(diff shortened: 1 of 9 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -669,13 +670,22 @@ private Status prepareJob() throws RuntimeException {
...
+      if (this.props.containsKey(JobProperties.USER_TO_PROXY)) {
+        final String jobProxyUser = this.props.getString(JobProperties.USER_TO_PROXY);
         if (this.proxyUsers != null && !this.proxyUsers.contains(jobProxyUser)) {
+          final String permissionsPageURL = getProjectPermissionsURL();
           this.logger.error("User " + jobProxyUser
-              + " has no permission to execute this job " + this.jobId + "!");
+              + " has no permission to execute this job " + this.jobId + "!"
+              + " If you want to execute this flow as " + jobProxyUser
+              + ", please add it to Proxy Users under project permissions page: " +
+              permissionsPageURL);
```
HappyRay: Is it necessary to add the URL here?

jamiesjc: This is per user request in ticket LIHADOOP-25204. It provides better instructions to users in case they couldn't find where to add the proxy user.

PR #1608, opened 2018-01-21
session TTL returned as part of the successful login response

mtrna:
Hi! This PR adds session TTL to the response to the login request.
Expected benefit: Azkaban clients can better manage session lifecycle. E.g. avoid uploading large files for a session that has already expired.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
+++ b/azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
@@ -25,28 +25,32 @@ import java.util.concurrent.TimeUnit;
 /**
  * Cache for web session.
  *
- * The following global azkaban properties can be used: max.num.sessions - used to determine the
- * number of live sessions that azkaban will handle. Default is 10000 session.time.to.live -Number
- * of seconds before session expires. Default set to 10 hours.
+ * The following global Azkaban properties are used:
+ * <ul>
+ *   <li>{@code max.num.sessions} - number of live sessions that Azkaban handles, default is 10000
+ *   <li>{@code session.time.to.live} - number of milliseconds before the session expires,
+ *   default 36000000 ms, i.e. 10 hours.
+ * </ul>
  */
 public class SessionCache {
 
   private static final int MAX_NUM_SESSIONS = 10000;
-  private static final long SESSION_TIME_TO_LIVE = 10 * 60 * 60 * 1000L;
+  private static final long DEFAULT_SESSION_TIME_TO_LIVE = 10 * 60 * 60 * 1000L; // 10 hours
 
-  // private CacheManager manager = CacheManager.create();
   private final Cache<String, Session> cache;
+
+  private final long effectiveSessionTimeToLive;
 
   /**
    * Constructor taking global props.
    */
   @Inject
   public SessionCache(final Props props) {
+    this.effectiveSessionTimeToLive = props.getLong("session.time.to.live",
+        DEFAULT_SESSION_TIME_TO_LIVE);
     this.cache = CacheBuilder.newBuilder()
         .maximumSize(props.getInt("max.num.sessions", MAX_NUM_SESSIONS))
-        .expireAfterAccess(
-            props.getLong("session.time.to.live", SESSION_TIME_TO_LIVE),
-            TimeUnit.MILLISECONDS)
+        .expireAfterAccess(effectiveSessionTimeToLive, TimeUnit.MILLISECONDS)
         .build();
   }
 
@@ -56,6 +60,10 @@ public class SessionCache {
   public Session getSession(final String sessionId) {
     final Session elem = this.cache.getIfPresent(sessionId);
     return elem;
+  }
+
+  public long getEffectiveSessionTimeToLive() {
+    return effectiveSessionTimeToLive;
   }
 
   /**
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
@@ -20,6 +20,7 @@ import static azkaban.ServiceProvider.SERVICE_PROVIDER;
 
 import azkaban.project.Project;
 import azkaban.server.session.Session;
+import azkaban.server.session.SessionCache;
 import azkaban.user.Permission;
 import azkaban.user.Role;
 import azkaban.user.User;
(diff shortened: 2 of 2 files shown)


azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -25,28 +25,32 @@
 /**
  * Cache for web session.
  *
- * The following global azkaban properties can be used: max.num.sessions - used to determine the
- * number of live sessions that azkaban will handle. Default is 10000 session.time.to.live -Number
- * of seconds before session expires. Default set to 10 hours.
+ * The following global Azkaban properties are used:
+ * <ul>
```
HappyRay: Nice. Thanks for updating the comments.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -25,28 +25,32 @@
...
+ *   <li>{@code max.num.sessions} - number of live sessions that Azkaban handles, default is 10000
+ *   <li>{@code session.time.to.live} - number of milliseconds before the session expires,
+ *   default 36000000 ms, i.e. 10 hours.
+ * </ul>
  */
 public class SessionCache {
 
   private static final int MAX_NUM_SESSIONS = 10000;
-  private static final long SESSION_TIME_TO_LIVE = 10 * 60 * 60 * 1000L;
+  private static final long DEFAULT_SESSION_TIME_TO_LIVE = 10 * 60 * 60 * 1000L; // 10 hours
```
HappyRay: Nice. Thanks.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -412,9 +413,11 @@ protected void handleAjaxLoginAction(final HttpServletRequest req,
       final Cookie cookie = new Cookie(SESSION_ID_NAME, session.getSessionId());
       cookie.setPath("/");
       resp.addCookie(cookie);
-      getApplication().getSessionCache().addSession(session);
+      SessionCache cache = getApplication().getSessionCache();
+      cache.addSession(session);
       ret.put("status", "success");
       ret.put("session.id", session.getSessionId());
+      ret.put("session.ttl", String.valueOf(cache.getEffectiveSessionTimeToLive()));
```
HappyRay: This value is static per an AZ web server instance. Right?

Could you describe your use case in more details?

mtrna: Session TTL is a final/immutable value per AZ instance, it has a default value that can be overridden by a prop.
Use case for the TTL: once a client obtains a session id (token), it comes handy for the client to know how long the token is valid. E.g. not to to upload a large ZIP with an expired token (and be rejected, re-authenticate, and upload the ZIP again).
E.g. `linkedin-gradle-plugin-for-apache-hadoop` could benefit from having such information.

HappyRay: I haven't seen such a scenario even though some of our zips are very big. 

Isn't the session extended when there is activity? 

Do you have reference to other popular web applications that do this?

mtrna: It is extended. Client does not know what is the TTL though. As a client: do you login again or do you reuse a session token after an hour, five hours, twelve hours of inactivity? Think about session as a temporary token: if it is temporary, it is import to tell the client how long is it valid.

HappyRay: I see your point.

The client has the following choices that I can think of:
a. always login first
b. re-login if it gets an authentication error
Is it not enough?

Plenty of web applications use session tokens. I am curious how many of them use the patten you suggested.

mtrna: a. the benefit of using sessions is that the authentication is done when a session is not available/expires; what would be the benefit of having sessions when clients would authenticate before every action?
b. this leads to waste; from my experience, in many cases upload is the only action clients are performing via the API; on a slow VPN connections, this creates an unnecessary long round-trip

HappyRay: > what would be the benefit of having sessions when clients would authenticate before every action?

Not every action.
A session allows multiple actions to be performed without authentication for each one. But not long enough to reduce security risk. 
Did I miss something?

pranayhasan: Currently, the open-source Hadoop Plugin which has an azkaban client is performing **b. re-login if it gets an authentication error** in most of the Gradle tasks it supports. 
@mtrna As we've discussed in https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/pull/199 the Hadoop Plugin checks if authentication is valid by making an API call to createProject and then to upload. 
Ref: https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/blob/master/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanUploadTask.groovy#L66

Let us know if you have a different use-case which shall help clients with information about TTL.

mtrna: @HappyRay : I was reacting to "always login first": it gives some space for interpretation.

In summary: what I find is missing on the client side is the information about whether the session token is still valid or not. Apparently we disagree on whether this information should be available to the clients of Azkaban. I am closing the PR then.

HappyRay: Got it.

Actually, I like the changes you made except the line "ret.put("session.ttl", String.valueOf(cache.getEffectiveSessionTimeToLive()));" I understand it may have some value for some clients. But I am not convinced that the benefit is worth the lifetime cost of maintaining this additional contract. 

If you create a new PR with the line referenced above removed, I would be happy to review again and merge it. 

It would be even better if you can move the key name session.time.to.live to the common class that defines configuration keys. We would like to gradually migrate the existing constants there as the related code is changed.  

Thanks

HappyRay:
BTW, do you use Azkaban at work? If so, do you mind sharing your company name?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
mtrna: Hi @HappyRay ! Sure, the follow up PR with doc/config update only: https://github.com/azkaban/azkaban/pull/1626

mtrna:
@HappyRay : Answered via LinkedIn inbox. :)

PR #1611, opened 2018-01-23
flow trigger/dependency instance

chengren311:
PR added two core classes for flow trigger feature: TriggerInstance and DependencyInstance.

Conceptually trigger Instance represents an execution of a flow trigger. It holds execution context such as a historically unique trigger execution id, execution status, start/end time, and a list of dependency instances.

Similarly dependency Instance denotes an execution of a dependency performing availability check on a particular data dependency. It holds execution context such as execution status and start/end time, all dependency Instances within the same trigger instance share the same execution id as trigger instance.

A trigger instance and its dependency instances will be created on scheduled trigger start time. The trigger instance will wait until all dependency instances are available, at which point a new flow instance is created, or the maximum allowed wait time is exceeded. Once successfully created, trigger instance will equipped with an UUID to identify itself, initial status as running, and a list of dependency instances whose status are running as well.

Trigger instance can be cancelled manually by user, by timeout when max waiting time is exceeded, or by internal dependency failure. 

Trigger/Dependency instance has following status(Running, Cancelling, Cancelled, Succeeded). 
Since trigger instance is nothing but a collection of dependency instances, it doesn’t have variables to keep status, start or endtime, all of which can be quickly inferred from its belonging dependencies’ status, start/endtime, which unburdens us from maintaining extra variables.

Cancellation cause will be attached to dependency instance for better user experience when cancelled for whatever reason.

Valid cancellation cause includes:
 TIMEOUT, // cancellation is issued due to exceeding max wait time
 MANUAL, // cancellation is issued by user
 FAILURE, // cancellation is issued by internal dependency instance failure
 CASCADING // cancelled by cascading failure(peer dependency is cancelled)

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
@@ -1,0 +1,269 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.flowtrigger;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.List;
+import org.junit.Test;
+
+public class TriggerInstanceTest {
+
+  public static Date getDate(final int hour, final int minute, final int second) {
+    final Calendar cal = Calendar.getInstance();
+    cal.set(Calendar.YEAR, 2000);
+    cal.set(Calendar.MONTH, 1);
+    cal.set(Calendar.DAY_OF_MONTH, 1);
+    cal.set(Calendar.HOUR_OF_DAY, hour);
+    cal.set(Calendar.MINUTE, minute);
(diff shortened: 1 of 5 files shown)


azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    ti = new TriggerInstance("1", null, "1", 1,
+        "test", dependencyInstanceList, -1, null);
+    assertThat(ti.getStartTime()).isNull();
+    assertThat(ti.getEndTime()).isNull();
+    dependencyInstanceList.clear();
+  }
+
+  @Test
+  public void testTriggerInstanceStatus() throws Exception {
```
reallocf: It looks like this is multiple tests wrapped inside a single test method, is that right? If so, it'll be easier to parallelize if broken out into multiple smaller tests. That would also make it easier for us to identify quickly what is broken and why if a test fails.

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
```diff
@@ -0,0 +1,269 @@
...
+      final CancellationCause cancelCause, final Date startTime, final Date endTime) {
+    final DependencyInstance depInst = new DependencyInstance(null, startTime, endTime, null,
+        status, cancelCause);
+    depInst.setStatus(status);
+    depInst.setCancellationCause(cancelCause);
+    return depInst;
+  }
+
+  @Test
+  public void testTriggerInstanceStartTimeEndTime() throws Exception {
```
reallocf: Same as last comment.

reallocf:
Other than that it looks good! Check out comments then I'll give ship-it.

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
chengren311: thanks, updated.

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceTest.java
chengren311: thanks, updated.

PR #1613, opened 2018-01-23
Add debug logging for upload project issue.

jamiesjc:
User reported intermittent issues that they sometimes could not get job properties after successfully deploying their project. When they executed the flow, it still picked the old project version and run. Project info like lastModifiedTime was not updated on the web UI after CRT deployment.
This issue could not be reproduced locally and it only happened to certain projects occasionally. Suspect some issue with updating project cache. Add some log messages to help debug the issue.
Related issues: LIHADOOP-33098, LIHADOOP-34114, TOOLS-170605

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -213,6 +213,7 @@ public class ProjectManager {
       fetchedProject = this.projectsByName.get(name);
     } else {
       try {
+        logger.info("Project " + name + " doesn't exist in cache, fetching from DB now.");
         fetchedProject = this.projectLoader.fetchProjectByName(name);
       } catch (final ProjectManagerException e) {
         logger.error("Could not load project from store.", e);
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -213,6 +213,7 @@ public Project getProject(final String name) {
       fetchedProject = this.projectsByName.get(name);
     } else {
       try {
+        logger.info("Project " + name + " doesn't exist in cache, fetching from DB now.");
```
chengren311: should we make level debug? Also just FYI, with debug level, we can add more detailed logging for better debuggability without concerning overwhelming logging message

jamiesjc: Currently only logging with level INFO or above will be printed out in our server. We have to change log4j configuration and do a server deployment to enable debug level logging. It might worth considering using log4j2 to dynamically switch logging levels in the future.

HappyRay: For new logging code, I would suggest 
"slf4j provides a more efficient API called parameterized logging.
Try to use that format.
see
https://stackoverflow.com/questions/30975026/log4j-implicit-string-formatting

"
See my comment in #1604

jamiesjc: Agree. Thanks for the suggestion!

PR #1615, opened 2018-01-25
flow trigger JDBC loader

chengren311:
The PR added JDBC loader for flow trigger. The loader provides all DB related operations for flow trigger instance and dependency instance([#1611](https://github.com/azkaban/azkaban/pull/1611)) to interact with dependency instance table(https://github.com/azkaban/azkaban/pull/1612). 

The PR also added toString(), hashCode() and equals() to several existing classes for complementary purpose.


azkaban-web-server/src/main/java/azkaban/flowtrigger/db/JdbcFlowTriggerInstanceLoaderImpl.java
```diff
@@ -0,0 +1,466 @@
...
+      triggerInstance = !res.isEmpty() ? res.iterator().next() : null;
+    } catch (final SQLException ex) {
+      handleSQLException(ex);
+    }
+    if (triggerInstance != null) {
+      final int projectId = triggerInstance.getProject().getId();
+      final int projectVersion = triggerInstance.getProject().getVersion();
+      final String flowFileName = triggerInstance.getFlowId() + ".flow";
+      final int flowVersion = triggerInstance.getFlowVersion();
+      final File tempDir = Files.createTempDir();
```
reallocf: I don't understand, why are we creating a temp dir then deleting it?

azkaban-web-server/src/test/java/azkaban/flowtrigger/TestDependencyCheck.java
```diff
@@ -0,0 +1,49 @@
...
+public class TestDependencyCheck implements DependencyCheck {
+
+  private static final Logger logger = LoggerFactory.getLogger(TestDependencyCheck.class);
+
+  @Override
+  public DependencyInstanceContext run(final DependencyInstanceConfig config,
+      final DependencyInstanceRuntimeProps runtimeProps,
+      final DependencyInstanceCallback callback) {
+    logger.info("running TestDependencyCheck with config:" + config);
+//    try {
```
reallocf: Should this comment still be here?

azkaban-web-server/src/test/java/azkaban/flowtrigger/TestDependencyCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    return new TestDependencyInstanceContext(config, runtimeProps, callback);
+  }
+
+  @Override
+  public void shutdown() {
+  }
+
+  @Override
+  public void init(final DependencyPluginConfig config) {
+    //logger.info("initing TestDependencyCheck with config:" + config);
```
reallocf: This comment too

azkaban-web-server/src/test/java/azkaban/flowtrigger/TestDependencyCheck.java
chengren311: actually i found this class is no longer used. so just removed the whole class :)

azkaban-web-server/src/test/java/azkaban/flowtrigger/TestDependencyCheck.java
chengren311: same as above

azkaban-web-server/src/main/java/azkaban/flowtrigger/db/JdbcFlowTriggerInstanceLoaderImpl.java
chengren311: it's required by the API. Though I think the API can be improved, but the temp dir is used to store the downloaded file and it's caller's responsibility to delete the dir. cc @jamiesjc

PR #909, opened 2017-02-15
show execution options in schedule page

wyukawa:
Sometimes, I want to check execution options(for example, failureAction, concurrentOption)

But, it's difficult to check because Azkaban stores as blob data in mysql.

So, if there is "show execution option button" in schedule page, it would be helpful.

PR #1623, opened 2018-02-01
Fixed the UI display issue of same executionOptions for all schedules.

jamiesjc:
When users click the show button for executionOptions on /scheduling page, it shows the same info for all schedules, which is incorrect. This is related to a recent change in PR https://github.com/azkaban/azkaban/pull/909
The issue happens because the executionOptions id is the same for all schedules when loading the page. I made the change to specify the executionOptions id for each schedule. Also, I changed the display of executionOptions info and the show button style to make it more readable.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.codehaus.jettison.json.JSONObject;
 
 /**
  * Execution options for submitted flows and scheduled flows
@@ -263,6 +264,10 @@ public class ExecutionOptions {
     flowOptionObj.put(MAIL_CREATOR, this.mailCreator);
     flowOptionObj.put(MEMORY_CHECK, this.memoryCheck);
     return flowOptionObj;
+  }
+
+  public JSONObject toJSON() {
+    return new JSONObject(toObject());
   }
 
   public enum FailureAction {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -18,13 +18,14 @@ package azkaban.executor;
 
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.utils.TypedMapWrapper;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.codehaus.jettison.json.JSONObject;
 
 /**
  * Execution options for submitted flows and scheduled flows
@@ -266,8 +267,9 @@ public class ExecutionOptions {
     return flowOptionObj;
   }
 
-  public JSONObject toJSON() {
-    return new JSONObject(toObject());
+  public String toJSON() {
+    final Gson gson = new GsonBuilder().setPrettyPrinting().create();
+    return gson.toJson(toObject());
   }
 
   public enum FailureAction {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -18,7 +18,6 @@ package azkaban.executor;
 
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.utils.TypedMapWrapper;
-import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -268,8 +267,7 @@ public class ExecutionOptions {
   }
 
   public String toJSON() {
-    final Gson gson = new GsonBuilder().setPrettyPrinting().create();
-    return gson.toJson(toObject());
+    return new GsonBuilder().setPrettyPrinting().create().toJson(toObject());
   }
 
   public enum FailureAction {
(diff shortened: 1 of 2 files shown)


jamiesjc:
screen shot 2018-02-01 at 10 10 57 am

jamiesjc:
@wyukawa Let me know if you have any concern about the change.

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -24,6 +24,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.codehaus.jettison.json.JSONObject;
```
HappyRay: We also use the gson library. Is there any reason to use this json library over gson?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
```diff
@@ -114,17 +121,28 @@
...
-                <td><button type="button" data-toggle="modal" data-target="#executionOptions">show</button></td>
+                <td>
+                ## Changed the style of "Show" button to be consistent with other buttons.
+                  <button type="button" class="btn btn-sm btn-info" data-toggle="modal"
+                          data-target="#executionOptions-${velocityCount}">Show
+                  </button>
+                </td>
 
-                  <div class="modal fade" id="executionOptions" tabindex="-1" role="dialog" aria-labelledby="executionOptionsLabel">
+              ## Distinguished executionOptions of each schedule with unique count number to fix
```
HappyRay: Could you elaborate more on how the unique count works?

HappyRay:
Nice change description. I like it that you referenced the pr that introduced the bug.

How difficult would it be to have automated tests for this feature?

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
jamiesjc: Thanks for the suggestion! I didn't realize that gson was so convenient to use. I'll push a new commit with the simplified code.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
jamiesjc: The original issue was that all the schedules shared the same data-target id. I added "velocityCount" to each id because it is basically the index of each schedule and it's unique, so it can be used to identify each data-target.

jamiesjc:
@HappyRay I did some research but haven't figured out an easy way to do that. We probably need to add the framework for UI testing which is currently missing.

wyukawa:
@jamiesjc 
No concern.
Thank you for fixing my bug.

HappyRay:
I am glad you found a simpler way.

In your earlier attempt:

>  ## Converted the executionOptions to JSON strings for clearer readability purpose.
+      document.querySelectorAll('.executionOptions-json').forEach(
+          function (element) {
+            element.innerHTML = JSON.stringify(JSON.parse(element.innerText), null, 2);
+          });
+

Did you intend to inject an object ( instead of a string ) into the template? What was the behavior?
And you convert the html content to json using Javascript?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
```diff
@@ -114,17 +114,27 @@
...
```
HappyRay: Is this comment necessary?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
HappyRay: Where is the code that parses the executionOptions-<number>?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
```diff
@@ -114,17 +114,27 @@
...
                       <div class="modal-dialog" role="document">
                           <div class="modal-content">
                               <div class="modal-header">
                                   <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
                                   <h4 class="modal-title" id="executionOptionsLabel">Execution Options</h4>
                               </div>
                               <div class="modal-body">
-                                  ${sched.executionOptions.toObject()}
+                              ## Used <pre> to display text with code format
+                                <pre>${sched.executionOptions.toJSON()}</pre>
```
HappyRay: I suggest keeping the template code as simple as possible i.e. no function calls within ${}.

How about moving this logic into java code?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
jamiesjc: We first call handleGetAllSchedules to obtain all schedules. Then for the executionOptions in each schedule, we tag it with a unique id called "executionOptions-${velocityCount}", and then display it when clicking the show button. Not sure If I answered your question?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
jamiesjc: Since we are not very familiar with the UI part, I think it's necessary to fully document the details so that others can understand the intention of the code change.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/scheduledflowpage.vm
jamiesjc: The only way I can think of is to first convert the executionOptions to JSON for each schedule in the servlet call azkaban.webapp.servlet.ScheduleServlet#handleGetAllSchedules. Then return a new list of Objects which contain the converted results instead of the original schedules. I agree it's ideal to keep the template simple, but the additional complexity to the logic might not worth it.

Issue #1614, opened 2018-01-24
Unable to build


dave-r12:
Hey, I wanted to try out azkaban and am attempting to build from source. Unfortunately I'm seeing test failures. I haven't spent time debugging them and wanted to see if there was something I was missing.
 
I'm running: `./gradlew clean build installDist`

And here is the test failures that I see:
```
azkaban.executor.ExecutorDaoTest > classMethod FAILED
    java.sql.SQLException: Table "EXECUTION_LOGS" already exists; SQL statement:
    CREATE TABLE execution_logs (
      exec_id     INT NOT NULL,
      name        VARCHAR(128),
      attempt     INT,
      enc_type    TINYINT,
      start_byte  INT,
      end_byte    INT,
      log         LONGBLOB,
      upload_time BIGINT,
      PRIMARY KEY (exec_id, name, attempt, start_byte)
    ) [42101-193] Query: CREATE TABLE execution_logs (
      exec_id     INT NOT NULL,
      name        VARCHAR(128),
      attempt     INT,
      enc_type    TINYINT,
      start_byte  INT,
      end_byte    INT,
      log         LONGBLOB,
      upload_time BIGINT,
      PRIMARY KEY (exec_id, name, attempt, start_byte)
    ) Parameters: []
        at org.apache.commons.dbutils.AbstractQueryRunner.rethrow(AbstractQueryRunner.java:363)
        at org.apache.commons.dbutils.QueryRunner.update(QueryRunner.java:490)
        at org.apache.commons.dbutils.QueryRunner.update(QueryRunner.java:376)
        at azkaban.db.DatabaseSetup.runTableScripts(DatabaseSetup.java:102)
        at azkaban.db.DatabaseSetup.createTables(DatabaseSetup.java:82)
        at azkaban.db.DatabaseSetup.updateDatabase(DatabaseSetup.java:57)
        at azkaban.test.Utils.initTestDB(Utils.java:36)
        at azkaban.executor.ExecutorDaoTest.setUp(ExecutorDaoTest.java:40)

azkaban.executor.ExecutorDaoTest > classMethod FAILED
    java.lang.NullPointerException
        at azkaban.executor.ExecutorDaoTest.destroyDB(ExecutorDaoTest.java:46)

azkaban.executor.ExecutionFlowDaoTest > classMethod FAILED
    java.sql.SQLException: Table "EXECUTION_LOGS" already exists; SQL statement:
    CREATE TABLE execution_logs (
[...]

PR #1628, opened 2018-02-05
fix JDBC test failure reported in #1614

kunkun-tang:
This PR is mainly a refactor, in order to fix JDBC intermittent test failure reported in #1614. Looks like JdbcTriggerImplTest uses a deprecated class to construct h2 test db. The fix is to replace it with the one which all XXXDaoTest currently are depending on.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/trigger/JdbcTriggerImplTest.java
+++ b/azkaban-common/src/test/java/azkaban/trigger/JdbcTriggerImplTest.java
@@ -18,24 +18,19 @@ package azkaban.trigger;
 
 import static org.junit.Assert.assertTrue;
 
-import azkaban.database.AzkabanConnectionPoolTest;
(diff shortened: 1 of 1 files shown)


azkaban-common/src/test/java/azkaban/trigger/JdbcTriggerImplTest.java
```diff
@@ -18,45 +18,32 @@
...
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 
 public class JdbcTriggerImplTest {
 
-  public static AzkabanDataSource dataSource = new AzkabanConnectionPoolTest.EmbeddedH2BasicDataSource();
-  TriggerLoader loader;
-  DatabaseOperator dbOperator;
+  private static DatabaseOperator dbOperator;
```
HappyRay: Why does this field need to be static?

PR #1630, opened 2018-02-05
validate flow trigger definition when parsing the flow trigger section in flow yaml file

chengren311:
The following validation steps are performed when parsing flow yaml file:
1. if specified max wait min must >= 1, if not, allowed max wait min(10 days) is applied.
2. flow trigger schedule cannot be null
3. flow trigger schedule type must be cron
4. flow trigger schedule value must be a valid cron expression
5. flow trigger schedule section contains type and value only
6. dependency name must be unique at flow trigger level
7. dependency config(type + params) must be unique at flow trigger level.
8. if max wait min >= allowed max wait min, it will be automatically set to allowed one(10 days)
9. dependency name/type are required.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
@@ -97,11 +97,11 @@ public class NodeBeanLoader {
 
     Preconditions.checkArgument(
         scheduleMap.containsKey(Constants.SCHEDULE_TYPE) && scheduleMap.get(Constants.SCHEDULE_TYPE)
-            .equals(Constants.SCHEDULE_VALUE), "flow trigger schedule type must be cron");
+            .equals(Constants.CRON_SCHEDULE_TYPE), "flow trigger schedule type must be cron");
 
     Preconditions.checkArgument(scheduleMap.containsKey(Constants.SCHEDULE_VALUE) && CronExpression
-        .isValidExpression(Constants.SCHEDULE_VALUE), "flow trigger schedule value must be a "
-        + "valid cron expression");
+            .isValidExpression(scheduleMap.get(Constants.SCHEDULE_VALUE)),
+        "flow trigger schedule value must be a valid cron expression");
 
     Preconditions.checkArgument(scheduleMap.size() == 2, "flow trigger schedule must "
         + "contain type and value only");
--- a/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
@@ -102,10 +102,6 @@ public class NodeBeanLoader {
         .isValidExpression(scheduleMap.get("value")), "flow trigger schedule value must be a "
         + "valid cron expression");
 
-    Preconditions.checkArgument(scheduleMap.containsKey("value") && CronExpression
-        .isValidExpression(scheduleMap.get("value")), "flow trigger schedule value must be a "
-        + "valid cron expression");
-
     Preconditions.checkArgument(scheduleMap.size() == 2, "flow trigger schedule must "
         + "contain type and value only");
   }
--- a/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
@@ -95,11 +95,12 @@ public class NodeBeanLoader {
 
     Preconditions.checkNotNull(scheduleMap, "flow trigger schedule must not be null");
 
-    Preconditions.checkArgument(scheduleMap.containsKey("type") && scheduleMap.get("type")
-        .equals("cron"), "flow trigger schedule type must be cron");
+    Preconditions.checkArgument(
+        scheduleMap.containsKey(Constants.SCHEDULE_TYPE) && scheduleMap.get(Constants.SCHEDULE_TYPE)
+            .equals(Constants.SCHEDULE_VALUE), "flow trigger schedule type must be cron");
 
-    Preconditions.checkArgument(scheduleMap.containsKey("value") && CronExpression
-        .isValidExpression(scheduleMap.get("value")), "flow trigger schedule value must be a "
+    Preconditions.checkArgument(scheduleMap.containsKey(Constants.SCHEDULE_VALUE) && CronExpression
+        .isValidExpression(Constants.SCHEDULE_VALUE), "flow trigger schedule value must be a "
         + "valid cron expression");
 
     Preconditions.checkArgument(scheduleMap.size() == 2, "flow trigger schedule must "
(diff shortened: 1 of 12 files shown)


azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -86,15 +90,88 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
...
+    Preconditions.checkNotNull(scheduleMap, "flow trigger schedule must not be null");
+
+    Preconditions.checkArgument(scheduleMap.containsKey("type") && scheduleMap.get("type")
+        .equals("cron"), "flow trigger schedule type must be cron");
+
+    Preconditions.checkArgument(scheduleMap.containsKey("value") && CronExpression
+        .isValidExpression(scheduleMap.get("value")), "flow trigger schedule value must be a "
+        + "valid cron expression");
+
+    Preconditions.checkArgument(scheduleMap.containsKey("value") && CronExpression
```
jamiesjc: Is this check duplicate of line 101?

chengren311: thanks, fixed.

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -86,15 +90,88 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
...
+
+  /**
+   * check uniqueness of dependency type and params
+   */
+  private void validateDepDefinitionUniqueness(final List<TriggerDependencyBean> dependencies) {
+    final Set<String> seen = new HashSet<>();
+    for (final TriggerDependencyBean dep : dependencies) {
+      final Map<String, String> props = dep.getParams();
+      // set.add() returns false when there exists duplicate
+      Preconditions.checkArgument(seen.add(dep.getType() + ":" + props.toString()), String.format
```
jamiesjc: If params for two deps have the same content but in different order, will the check here fail?

chengren311: yes, since props is a hashmap which is order agnostic.

jamiesjc:
Consider adding more test cases for validating schedules and dependencies?

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -86,15 +90,84 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
     }
   }
 
+  private void validateSchedule(final FlowTriggerBean flowTriggerBean) {
+    final Map<String, String> scheduleMap = flowTriggerBean.getSchedule();
+
+    Preconditions.checkNotNull(scheduleMap, "flow trigger schedule must not be null");
+
+    Preconditions.checkArgument(scheduleMap.containsKey("type") && scheduleMap.get("type")
+        .equals("cron"), "flow trigger schedule type must be cron");
```
jamiesjc: "type", "cron" and "value" are already defined in Constants.java. You can use them directly.

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -86,15 +90,84 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
...
+        .isValidExpression(scheduleMap.get("value")), "flow trigger schedule value must be a "
+        + "valid cron expression");
+
+    Preconditions.checkArgument(scheduleMap.size() == 2, "flow trigger schedule must "
+        + "contain type and value only");
+  }
+
+  private void validateFlowTriggerBean(final FlowTriggerBean flowTriggerBean) {
+    // validate max wait mins
+    Preconditions.checkArgument(flowTriggerBean.getMaxWaitMins() >= 1, "max wait min must be "
```
jamiesjc: Define the numbers "2" and "1" in constants?

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
chengren311: sorry, i took it back, hashmap's printing order is not guaranteed, https://stackoverflow.com/questions/18778410/why-does-tostring-function-of-a-hashmap-prints-itself-with-a-different-order

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
chengren311: thanks, updated

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
chengren311: put 1 in constants, 2 is just the number of keys to be held, which is not a meaningful number, so i prefer to make it just a hard coded value here.

chengren311:
@jamiesjc thanks for reviewing, added more unit test.

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -131,14 +131,16 @@ private void validateDepNameUniqueness(final List<TriggerDependencyBean> depende
...
-      final Map<String, String> props = dep.getParams();
-      // set.add() returns false when there exists duplicate
-      Preconditions.checkArgument(seen.add(dep.getType() + ":" + props.toString()), String.format
-          ("duplicate "
-              + "dependency"
-              + "config %s found, dependency config should be unique", dep.getName()));
+    for (int i = 0; i < dependencies.size(); i++) {
+      for (int j = i + 1; j < dependencies.size(); j++) {
+        final boolean duplicateDepDefFound =
+            dependencies.get(i).getType().equals(dependencies.get(j)
```
jamiesjc: Do we need to check if type or params is null before comparing? Will it throw null pointer exception?

azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -86,15 +90,88 @@ public AzkabanNode toAzkabanNode(final NodeBean nodeBean) {
...
+  }
+
+  /**
+   * check uniqueness of dependency type and params
+   */
+  private void validateDepDefinitionUniqueness(final List<TriggerDependencyBean> dependencies) {
+    for (int i = 0; i < dependencies.size(); i++) {
+      for (int j = i + 1; j < dependencies.size(); j++) {
+        final boolean duplicateDepDefFound =
+            dependencies.get(i).getType().equals(dependencies.get(j)
```
jamiesjc: Do we need to check if type or params is null before comparing? Will it throw null pointer exception?

chengren311: added validation for that.

PR #1631, opened 2018-02-06
flow trigger scheduler

chengren311:
This PR created a wrapper class around QuartzScheduler - FlowTriggerScheduler.
The class provides
1. schedule all flows in a project with flow trigger defined with quarz scheduler.
2. unscheduleAll all flows in a project with flow trigger defined with quartz scheduler.
3. Retrieve the list of current flow schedules from quartz database.

FlowTriggerScheduler will be used to schedule/unschedule flows if schedule is part of flow configuration file(flow 2.0) when project is uploaded/deleted.

Also when flow trigger is defined, thus schedule is part of flow config, scheduling those flows from UI/API  is disallowed. 

Tested via manual uploading a project with flow trigger but unit test of this part is a bit tricky, will add unit test if possible or verify in integration test.


azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -0,0 +1,207 @@
...
+@Singleton
+public class FlowTriggerScheduler {
+
+  private static final Logger logger = LoggerFactory.getLogger(FlowTriggerScheduler.class);
+  private final ProjectLoader projectLoader;
+  private final QuartzScheduler scheduler;
+
+  @Inject
+  public FlowTriggerScheduler(final ProjectLoader projectLoader, final QuartzScheduler scheduler) {
+    this.projectLoader = requireNonNull(projectLoader);
```
kunkun-tang: Is it possible to check if the scheduler is running or not here in the constructor.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -0,0 +1,207 @@
...
+          + " schedule");
+      this.scheduler.unregisterJob(generateGroupName(flow));
+    }
+  }
+
+  private String generateGroupName(final Flow flow) {
+    return String.valueOf(flow.getProjectId()) + "." + flow.getId();
+  }
+
+  public void start() {
```
kunkun-tang: We should not need to start or shutdown here.

After sync-up, I still believe we should keep the original Quartz Scheduler a separate module. You don't need to handle start or shutdown API, which should be handled by AZ webserver, rather than plugin. Besides, you don't and shouldn't call all APIs like Pause or checkExists, which should be `QuartzScheduler`'s responsibility. This class should just call its API. It will be hard to evolve or enhance `QuartzScheduler` if we go this direction.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -82,6 +84,32 @@ public ProjectManager(final AzkabanProjectLoader azkabanProjectLoader,
...
+      throws IOException {
+    final String flowFileName = flow.getId() + ".flow";
+    final int latestFlowVersion = this.projectLoader.getLatestFlowVersion(project.getId(), flow
+        .getVersion(), flowFileName);
+    if (latestFlowVersion > 0) {
+      final File tempDir = com.google.common.io.Files.createTempDir();
+      final File flowFile;
+      try {
+        flowFile = this.projectLoader
+            .getUploadedFlowFile(project.getId(), project.getVersion(),
```
jamiesjc: It is possible that getUploadedFlowFile throws ProjectManagerException. Do you plan to catch it as well?

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -0,0 +1,207 @@
...
+                    FlowTriggerQuartzJob.FLOW_TRIGGER, flowTrigger,
+                    FlowTriggerQuartzJob.FLOW_ID, flow.getId(),
+                    FlowTriggerQuartzJob.FLOW_VERSION, latestFlowVersion,
+                    FlowTriggerQuartzJob.PROJECT, projectJson);
+            logger.info("scheduling flow " + flow.getProjectId() + "." + flow.getId());
+            this.scheduler
+                .registerJob(flowTrigger.getSchedule().getCronExpression(), new QuartzJobDescription
+                    (FlowTriggerQuartzJob.class, generateGroupName(flow), contextMap));
+          }
+        } catch (final IOException ex) {
```
jamiesjc: Same comment as above.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1701,9 +1720,19 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+        if (this.enableQuartz) {
+          this.scheduler.unscheduleAll(project);
+        }
         final Map<String, ValidationReport> reports =
             this.projectManager.uploadProject(project, archiveFile, type, user,
                 props);
+
+        if (this.enableQuartz) {
+          //schedule the new project
+          this.scheduler.scheduleAll(project, user.getUserId());
```
jamiesjc: What if there are errors in uploading the new project? Should we still schedule it?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
chengren311: will update, thanks

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
chengren311: good catch

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
chengren311: nope

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
chengren311: why do we need to check it in constructor?

azkaban-common/src/main/java/azkaban/project/CronSchedule.java
```diff
@@ -26,7 +27,7 @@
  * It couldn't be changed once gets constructed.
  * It will be used to schedule a trigger.
  */
-public class CronSchedule {
```
kunkun-tang: Why does this class need to be Serializable?

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerQuartzJob.java
```diff
@@ -0,0 +1,57 @@
...
+import azkaban.flowtrigger.FlowTriggerService;
+import azkaban.project.FlowTrigger;
+import azkaban.project.Project;
+import azkaban.scheduler.AbstractQuartzJob;
+import javax.inject.Inject;
+import org.quartz.JobDataMap;
+import org.quartz.JobExecutionContext;
+
+
+public class FlowTriggerQuartzJob extends AbstractQuartzJob {
```
kunkun-tang: I should have remembered. Does this QuartzJob run gobally only once or every flow trigger (called by user zip) will execute once?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -506,4 +534,4 @@ public boolean loadProjectWhiteList() {
     }
     return false;
   }
-}
+}
```
kunkun-tang: new line needed.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -628,6 +635,16 @@ private void handleRemoveProject(final HttpServletRequest req,
       this.setErrorMessageInCookie(resp, e.getMessage());
       resp.sendRedirect(req.getRequestURI() + "?project=" + projectName);
       return;
+    } finally {
+      try {
+        //unschedule the project regardless
+        if (this.enableQuartz) {
```
kunkun-tang: should this move to the removeProject try catch block?
Another question. Do we remove normal schedule today when removing a project?

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -0,0 +1,207 @@
...
+  @Inject
+  public FlowTriggerScheduler(final ProjectLoader projectLoader, final QuartzScheduler scheduler) {
+    this.projectLoader = requireNonNull(projectLoader);
+    this.scheduler = requireNonNull(scheduler);
+  }
+
+  /**
+   * Schedule flows containing flow triggers
+   */
+  public void scheduleAll(final Project project, final String submitUser)
```
kunkun-tang: why this method is called `scheduleAll` ?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -591,6 +593,24 @@ private void ajaxScheduleCronFlow(final HttpServletRequest req,
       return;
     }
 
+    final boolean hasFlowTrigger;
+    try {
```
kunkun-tang: Don't understand what does this part of code do?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -117,18 +121,20 @@ public void unregisterJob(final String groupName) throws SchedulerException {
...
    *
    * @param cronExpression the cron schedule for this job
    * @param jobDescription Regarding QuartzJobDescription#groupName, in order to guarantee no
    * duplicate quartz schedules, we design the naming convention depending on use cases: <ul>
-   * <li>User flow schedule: we use {@link org.quartz.JobKey#JobKey} to represent the identity of a
+   * <li>User flow schedule: we use {@link JobKey#JobKey} to represent the identity of a
    * flow's schedule. The format follows "$projectID_$flowName" to guarantee no duplicates.
-   * <li>Quartz schedule for AZ internal use: the groupName should start with letters,
-   * rather than
+   * <li>Quartz schedule for AZ internal use: the groupName should start with letters, rather
```
jamiesjc: Update the comments for groupName definition which should be "projectID.flowName" and start with number?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -108,7 +108,11 @@ public void shutdown() {
     }
   }
 
-  public void unregisterJob(final String groupName) throws SchedulerException {
+  /**
+   * Unregister a job given the groupname. Since unregister might be called when
+   * concurrently removing projects, so synchronized is added to ensure thread safety.
+   */
+  public synchronized void unregisterJob(final String groupName) throws SchedulerException {
```
kunkun-tang: Wondering if it is possible to add a parameter in this API, because jobname is also necessary to locate a quartz Job. Would suggest we should do the change ASAP before flow trigger is in production.

azkaban-common/src/main/java/azkaban/project/CronSchedule.java
chengren311: it's needs to persisted by quartz.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -45,7 +45,7 @@
 public class QuartzScheduler {
 
   //Unless specified, all Quartz jobs's identities comes with the default job name.
-  private static final String DEFAULT_JOB_NAME = "job1";
+  public static final String DEFAULT_JOB_NAME = "job1";
```
kunkun-tang: This should not be called Flow Trigger. FlowTrigger should has its own flowTrigger job name, for instance, "flowtrigger"

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerQuartzJob.java
chengren311: this job will be scheduled when project is uploaded and run based on cron schedule.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
chengren311: schedule all flows with flow trigger

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
chengren311: will do thanks.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
chengren311: - Do we remove normal schedule today when removing a project
  Yes.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
chengren311: this part is to disallow manual scheduling flows with flow trigger

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
chengren311: we can do it in next PR.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
chengren311: we can do it in next PR.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -0,0 +1,207 @@
...
+      throws SchedulerException, ProjectManagerException {
+    //todo chengren311: schedule on uploading via CRT
+
+    for (final Flow flow : project.getFlows()) {
```
jamiesjc: project.getFlows will return all the flows including embedded flows. If a user config flow trigger inside an embedded flow, do we need to support that?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
chengren311: - should this move to the removeProject try catch block?
Let me consolidate unschedule code together

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
chengren311: good point, i will add check here to skip embedded flow

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -591,6 +593,24 @@ private void ajaxScheduleCronFlow(final HttpServletRequest req,
...
+      hasFlowTrigger = this.projectManager.hasFlowTrigger(project, flow);
+    } catch (final Exception ex) {
+      logger.error(ex);
+      ret.put("status", "error");
+      ret.put("message", String.format("Flow %s.%s cannot be scheduled: %s", projectName,
```
kunkun-tang: Shouldn't the error message be `Error while looking for flow Trigger`?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -591,6 +593,24 @@ private void ajaxScheduleCronFlow(final HttpServletRequest req,
...
+          flowName, ex));
+      return;
+    }
+
+    if (hasFlowTrigger) {
+      ret.put("status", "error");
+      ret.put("message", String.format("Flow %s.%s is already associated with flow trigger, "
```
kunkun-tang: Is it possible that it is normal that this flow doesn't have flow trigger? Would it be an error given this case? Don't understand what the error message mean here?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
chengren311: updated

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
chengren311: updated

PR #1632, opened 2018-02-06
move dependency plugin interface code to a separate module

chengren311:
The motivation of moving dependency plugin interface code(https://github.com/azkaban/azkaban/pull/1609, https://github.com/azkaban/azkaban/pull/1610) out from azkaban web server module to a new module is to expose minimum set of code to external dependency plugin avoiding unnecessary introduction of other classes in the development of dependency plugin. If those interface sit in the azkaban web server module, dependency plugin developers will have to depend on jar of that module containing a lot of other classes more than needed. This might potentially cause dependency conflict. 

This PR also includes bettering renaming on test dependency type and consolidating different test dependency types into one package.


azkaban-web-server/build.gradle
```diff
@@ -1,191 +1,193 @@
 buildscript {
-  repositories {
```
chengren311: auto formatted by intellij plugin

az-flow-trigger-dependency-plugin/src/main/java/flowtrigger/DependencyCheck.java
```diff
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package azkaban.flowtrigger;
+package flowtrigger;
```
reallocf: I think all other azkaban package names start with `azkaban.`. Should we change it for this package? Is there an easy way to make this also `azkaban.flowtrigger` without losing the benefits of this PR? It would also make it easier for a trigger dependency developer to identify through which imports they depend on azkaban.

chengren311: good point, thanks

az-flow-trigger-dependency-plugin/build.gradle
```diff
@@ -0,0 +1,20 @@
+/*
```
kunkun-tang: Please add a readme as an instruction for the implementation development.

azkaban-web-server/build.gradle
```diff
@@ -1,191 +1,193 @@
...
 
-  generateRestli deps.restliGenerator
-  generateRestli deps.restliTools
+    generateRestli deps.restliGenerator
+    generateRestli deps.restliTools
 
-  testCompile project(':test')
-  testCompile project(path: ':azkaban-db', configuration: 'testOutput')
-  testRuntime deps.h2
+    //testcompile project(':az-flow-trigger-dependency-plugin')
```
kunkun-tang: I would believe az web server modules needs this test dependency if you write its implementations in the test.

azkaban-web-server/src/test/java/azkaban/flowtrigger/testplugin/FakeDependencyCheck1.java
```diff
@@ -0,0 +1,51 @@
...
+package azkaban.flowtrigger.testplugin;
+
+import azkaban.flowtrigger.DependencyCheck;
+import azkaban.flowtrigger.DependencyInstanceCallback;
+import azkaban.flowtrigger.DependencyInstanceConfig;
+import azkaban.flowtrigger.DependencyInstanceContext;
+import azkaban.flowtrigger.DependencyInstanceRuntimeProps;
+import azkaban.flowtrigger.DependencyPluginConfig;
+
+public class FakeDependencyCheck1 implements DependencyCheck {
```
kunkun-tang: What's the diff between Check1 and Check2? Should we use a more concrete name?

az-flow-trigger-dependency-plugin/build.gradle
chengren311: updated

azkaban-web-server/build.gradle
chengren311: added testcompile

azkaban-web-server/src/test/java/azkaban/flowtrigger/testplugin/FakeDependencyCheck1.java
chengren311: It's just for testing without any specific purpose. so i cannot think of any concrete name for it.

reallocf: What's the difference between the checks? Why would I use 2 instead of 1? Maybe a name can be made from that?

azkaban-web-server/src/test/java/azkaban/flowtrigger/testplugin/FakeDependencyInstanceContext2.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public FakeDependencyInstanceContext2(final DependencyInstanceConfig config,
+      final DependencyInstanceRuntimeProps runtimeProps,
+      final DependencyInstanceCallback callback) {
+  }
+
+  @Override
+  public void cancel() {
+  }
+}
```
kunkun-tang: * I don't think we need a jar here. You might want to investigate an approach to not check in jar this way.

az-flow-trigger-dependency-plugin/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2017 LinkedIn Corp.
```
kunkun-tang: It should be 2018, right?

azkaban-web-server/src/test/java/azkaban/flowtrigger/testplugin/FakeDependencyInstanceContext2.java
chengren311: sure, can i put it in todo?

chengren311: added it to TODO.

PR #1627, opened 2018-02-05
flow trigger service

chengren311:
This PR added flow trigger service,  a singleton class in the AZ web server to process all flow trigger-related operations. Externally it provides following operations -
1. Create a trigger instance based on trigger definition.
2. Cancel a trigger instance.
3. Query running and historic trigger instances.
4. Recover incomplete trigger instances.
     
Internally, it
1. maintains the list of running trigger instance in memory.
2. updates status, starttime/endtime of trigger instance.
3. persists trigger instance to DB.
     
FlowTriggerService will be leveraged by Quartz scheduler, our new AZ scheduler to schedule starting triggers.
 
Flow Trigger Service internally contains a single-threaded executor service to process all trigger instance related operation to avoid complicated cases of race condition.

PR #1633, opened 2018-02-07
integrate flow trigger service with az web server

chengren311:
This PR injects flow trigger service(https://github.com/azkaban/azkaban/pull/1627) into azkaban web server. Flow trigger service will start if enabled in the config file. Upon start, flow trigger service initializes all dependency plugins and recover running trigger instance from last run status.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
+++ b/azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
@@ -18,6 +18,7 @@ package azkaban.flowtrigger;
 
 import azkaban.Constants;
 import azkaban.flowtrigger.database.FlowTriggerInstanceLoader;
+import azkaban.flowtrigger.plugin.FlowTriggerDependencyPluginException;
 import azkaban.flowtrigger.plugin.FlowTriggerDependencyPluginManager;
 import azkaban.project.FlowTrigger;
 import azkaban.project.FlowTriggerDependency;
@@ -95,6 +96,11 @@ public class FlowTriggerService {
     this.triggerProcessor = triggerProcessor;
     this.dependencyProcessor = dependencyProcessor;
     this.flowTriggerInstanceLoader = flowTriggerInstanceLoader;
+  }
+
+  public void start() throws FlowTriggerDependencyPluginException {
+    this.triggerPluginManager.loadAllPlugins();
+    this.recoverIncompleteTriggerInstances();
   }
 
   private DependencyInstanceContext createDepContext(final FlowTriggerDependency dep, final long
@@ -241,7 +247,7 @@ public class FlowTriggerService {
   /**
    * Resume executions of all incomplete trigger instances by recovering the state from db.
    */
-  public void recoverIncompleteTriggerInstances() {
+  private void recoverIncompleteTriggerInstances() {
     final Collection<TriggerInstance> unfinishedTriggerInstances = this.flowTriggerInstanceLoader
         .getIncompleteTriggerInstances();
     //todo chengren311: what if flow trigger is not found?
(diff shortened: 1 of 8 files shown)


azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
```diff
@@ -97,6 +98,11 @@ public FlowTriggerService(final FlowTriggerDependencyPluginManager pluginManager
     this.flowTriggerInstanceLoader = flowTriggerInstanceLoader;
   }
 
+  public void start() throws FlowTriggerDependencyPluginException {
```
kunkun-tang: I'm seeing this variable `List<TriggerInstance> runningTriggers` exist in this class. In the long run, AZ web server will be stateless, so state should be removed eventually. For now, It might be ok to stay with this approach. Please come up with the solution to address it.

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
```diff
@@ -159,7 +158,14 @@ public void loadDependencyPlugin(final File pluginDir)
     }
   }
 
-  private void loadAllDependencyPlugins() throws FlowTriggerDependencyPluginException {
+  /**
+   * Initialize all dependency plugins.
+   * todo chengren311: Current design aborts loadAllPlugins if any of the plugin fails to be
```
kunkun-tang: That's good `Todo` list!

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
```diff
@@ -45,10 +50,26 @@
...
+  @Singleton
+  public FlowTriggerDependencyPluginManager getDependencyPluginManager(final Props props)
+      throws FlowTriggerDependencyPluginException {
+    //todo chengren311: disable requireNonNull for now in beta since dependency plugin dir is not
+    // required. Add it back when flow trigger feature is enabled in production
+    String dependencyPluginDir;
+    try {
+      dependencyPluginDir = props.getString(ConfigurationKeys.DEPENDENCY_PLUGIN_DIR);
+    } catch (final Exception ex) {
+      dependencyPluginDir = null;
```
kunkun-tang: Why don't we throw exception here

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -526,6 +541,14 @@ private void prepareAndStartServer()
       this.quartzScheduler.start();
     }
 
+    if (this.props.getBoolean(ConfigurationKeys.ENABLE_FLOW_TRIGGER, false)) {
+      // flow trigger service throws exception when any dependency plugin fails to be initialized
+      // (e.x if it's kafka dependency and kafka is down). In this case if azkaban admin still
+      // wishes to start azkaban web server, she can disable flow trigger in the az config file and
```
kunkun-tang: `she` ?

azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
chengren311: not sure what you mean here by stateless. let's discuss more in person.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
chengren311: she as the gender-neutral pronoun

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
chengren311: since it will fail the azkaban web server startup if DEPENDENCY_PLUGIN_DIR is not specified in the config file. But once the corresponding DEPENDENCY_PLUGIN_DIR is created on az machines and specified in the config file, i will add null check here.

azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
kunkun-tang: `stateless` means more than one AZ webservers will serve web requests, and each server should not contain states internally. Could you think of any solution?

chengren311: then how would quartz scheduler handle this case? E.x, how does quartz know which schedule is served by which web server?

PR #1643, opened 2018-02-14
exclude trigger instances in cancelling status from recently finished flows

chengren311:
This PR fixes the bug which includes trigger Instances in cancelling state into recently finished flows.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/flowtrigger/database/JdbcFlowTriggerInstanceLoaderImpl.java
+++ b/azkaban-web-server/src/main/java/azkaban/flowtrigger/database/JdbcFlowTriggerInstanceLoaderImpl.java
@@ -95,17 +95,18 @@ public class JdbcFlowTriggerInstanceLoaderImpl implements FlowTriggerInstanceLoa
               Status.SUCCEEDED.ordinal(),
               Constants.UNASSIGNED_EXEC_ID);
 
-  private static final String SELECT_RECENTLY_FINISHED =
+  private static final String SELECT_RECENTLY_FINISHED = String.format(
       "SELECT execution_dependencies.trigger_instance_id,dep_name,starttime,endtime,dep_status,"
           + "cancelleation_cause,project_id,project_version,flow_id,flow_version,"
           + "project_json, flow_exec_id \n"
           + "FROM execution_dependencies JOIN (\n"
           + "SELECT trigger_instance_id FROM execution_dependencies WHERE trigger_instance_id not in (\n"
-          + "SELECT distinct(trigger_instance_id)  FROM execution_dependencies WHERE dep_status ="
-          + "  0 or dep_status = 3)\n"
+          + "SELECT distinct(trigger_instance_id) FROM execution_dependencies WHERE dep_status = "
+          + "  %s or dep_status = %s)\n"
           + "GROUP BY trigger_instance_id\n"
-          + "ORDER BY  min(starttime) desc limit %s) temp on execution_dependencies"
-          + ".trigger_instance_id in (temp.trigger_instance_id);";
+          + "ORDER BY  min(starttime) desc limit %%s) temp on execution_dependencies"
+          + ".trigger_instance_id in (temp.trigger_instance_id);", Status.RUNNING.ordinal(),
+      Status.CANCELLING.ordinal());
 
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/flowtrigger/database/JdbcFlowTriggerInstanceLoaderImpl.java
```diff
@@ -101,7 +101,8 @@
           + "project_json, flow_exec_id \n"
           + "FROM execution_dependencies JOIN (\n"
           + "SELECT trigger_instance_id FROM execution_dependencies WHERE trigger_instance_id not in (\n"
-          + "SELECT distinct(trigger_instance_id)  FROM execution_dependencies WHERE dep_status =  0 or dep_status = 4)\n"
+          + "SELECT distinct(trigger_instance_id)  FROM execution_dependencies WHERE dep_status ="
+          + "  0 or dep_status = 3)\n"
```
jamiesjc: Would it be better to pass the dependency status as the parameter into the SQL string? In case the enum value is changed in the future, we don't need to change the code here.

chengren311: hmm, thanks for the suggestion.  I considered it initially, but the status is very unlikely to be changed so i would rather to keep the sql as a constant string.

HappyRay: It's ok to change code in the future. We do it all the time. 
The question is how easy it is to change. 
If it is an implementation detail, we can change them in the future relatively easily. In that case, I would prefer the simplest and most readable solution.

HappyRay: What's the significance of the number 0 and 3? Can it be made more readable?

chengren311: updated

azkaban-web-server/src/main/java/azkaban/flowtrigger/database/JdbcFlowTriggerInstanceLoaderImpl.java
```diff
@@ -95,16 +95,18 @@
               Status.SUCCEEDED.ordinal(),
               Constants.UNASSIGNED_EXEC_ID);
 
-  private static final String SELECT_RECENTLY_FINISHED =
+  private static final String SELECT_RECENTLY_FINISHED = String.format(
       "SELECT execution_dependencies.trigger_instance_id,dep_name,starttime,endtime,dep_status,"
           + "cancelleation_cause,project_id,project_version,flow_id,flow_version,"
           + "project_json, flow_exec_id \n"
           + "FROM execution_dependencies JOIN (\n"
           + "SELECT trigger_instance_id FROM execution_dependencies WHERE trigger_instance_id not in (\n"
```
jamiesjc: Why don't you just select trigger_instance_id where dep_status = SUCCEEDED or CANCELLED here? Will the sql query be simplified?

chengren311: yeah, good catch

azkaban-web-server/src/test/java/azkaban/flowtrigger/FlowTriggerInstanceLoaderTest.java
```diff
@@ -222,7 +228,8 @@ public void testGetIncompleteTriggerInstancesReturnsEmpty() {
       }
     }
 
-    all.forEach(triggerInst -> this.triggerInstLoader.uploadTriggerInstance(triggerInst));
+    final List<TriggerInstance> shuffled = new ArrayList<>(all);
+    shuffled.forEach(triggerInst -> this.triggerInstLoader.uploadTriggerInstance(triggerInst));
```
jamiesjc: Missing the part of shuffling...

PR #1644, opened 2018-02-15
zero dependency flow trigger

chengren311:
A flow trigger without dependencies is a regular schedule without waiting for data, so the trigger instance will be processed as succeed once created.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
+++ b/azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
@@ -117,9 +117,6 @@ public class FlowTriggerService {
   private TriggerInstance createTriggerInstance(final FlowTrigger flowTrigger, final String flowId,
       final int flowVersion, final String submitUser, final Project project) {
     final String triggerInstId = generateId();
-    logger.info(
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/flowtrigger/FlowTriggerService.java
```diff
@@ -344,25 +341,27 @@ public void startTrigger(final FlowTrigger flowTrigger, final String flowId,
       final TriggerInstance triggerInst = createTriggerInstance(flowTrigger, flowId, flowVersion,
           submitUser, project);
 
-      logger.info(String.format("Starting trigger instance[id: %s]", triggerInst.getId()));
+      logger.info(
+          String.format("Starting the flow trigger %s[execId %s] by %s", flowTrigger, triggerInst
```
jamiesjc: It seems triggerInst.getId() will return the trigger instance id, not the execution id?

chengren311: exec Id is the terminology i used initially to identify an execution of a flow trigger,  but was later changed to trigger instance id. updated the log.

PR #1649, opened 2018-02-17
change the button style of showing flow trigger properties in flow trigger instance UI

chengren311:
before:
image

after:
image

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/executingflowtriggerspage.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/executingflowtriggerspage.vm
@@ -175,9 +175,9 @@
               #end
 
               <td>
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/executingflowtriggerspage.vm
```diff
@@ -175,9 +175,9 @@
               #end
 
               <td>
-                <button type="button" id="showTriggerProperties" class="btn btn-danger btn-sm"
+                <button type="button" id="showTriggerProperties" class="btn btn-sm btn-info"
```
jamiesjc: Why do you want to change the button type from info to danger? 
`.btn-danger | Indicates a dangerous or potentially negative action`
It seems to me triggerProperties are just info not some dangerous action.

PR #1650, opened 2018-02-18
Job filter style fix

mtrna:
before:
before

after:
after

### DIFF
--- a/azkaban-web-server/src/main/less/base.less
+++ b/azkaban-web-server/src/main/less/base.less
@@ -25,7 +25,8 @@
 }
 
 div#graphView.container-fill {
-  position: static;
+  position: relative;
+  top: 0px;
   height: 750px;
 }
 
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/less/base.less
```diff
@@ -7,8 +7,8 @@
 }
 
 .container-fill {
-  position: absolute;
-  top: 230px;
+  position: relative;
```
kunkun-tang: This tag looks being used by many other pages, like flowpage, jobdetailpage. Have you checked that everywhere is fine?

mtrna: Thanks for pointing it out; I have checked the other tabs an there was impact.

Moved all the styling logic to `div#graphView.container-fill`, which is specific to the graph pane, and tested the updated version on chrome and firefox, it looks good to me now.

HappyRay:
@mtrna Are you no longer interested in finishing this PR? Is there anything we can do to help?

mtrna:
Hi @HappyRay ! I am interested, the MR is reopened here: https://github.com/azkaban/azkaban/pull/1685

Issue #1653, opened 2018-02-22
Azkaban web evicts running executions if there are too many


juhoautio:
If there are too many concurrent executions on any executor instance, the call from azkaban-web to azkaban-executor /update fails with `413 Payload Too Large`.

In practice this happens somewhere between 200-300 executions. Indeed this may easily happen because the execution ids to update are passed as URL query params:

https://github.com/azkaban/azkaban/blob/2add2405974e6adaa73670d95f3193d307807958/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java#L1476-L1478

As a result, azkaban-web removes the execution from its running executions list:

https://github.com/azkaban/azkaban/blob/2add2405974e6adaa73670d95f3193d307807958/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java#L1497-L1500

This in turn means that the executions are not shown on the Executions page of Azkaban UI any more. Evicting even finalizes those executions, even though they're still running on azkaban-executor.

PR #1654, opened 2018-02-22
Add boolean property for evicting executions

juhoautio:
New boolean property `azkaban.evictExecutions`. Default: `true` (to match existing behaviour).

This can be used to decide whether Azkaban should ever give up on getting a response from executor.

As before, an eventual successful response may still indicate that an execution is not running on that executor any more, which will cause cleaning it up as usual.

In practice the automatic evicting has caused more trouble than benefit in our case. It's really surprising to see that executions are disappearing from the _Executing_ list in UI while they are still running on azkaban-executor side. I would recommend setting this to `false` even by default, but of course evicting may still be needed in some use cases that I can't really imagine - so I wouldn't dare to change it without users' own consideration.

This is related to #1653 - which needs to be fixed though, before it makes sense for anyone to set `azkaban.evictExecutions=false`

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -66,29 +66,7 @@ import java.util.concurrent.TimeoutException;
 public class ExecutorManager extends EventHandler implements
     ExecutorManagerAdapter {
 
-  public static final String AZKABAN_USE_MULTIPLE_EXECUTORS =
-      "azkaban.use.multiple.executors";
-  public static final String AZKABAN_MAX_CONCURRENT_RUNS_ONEFLOW =
-      "azkaban.max.concurrent.runs.oneflow";
-  static final String AZKABAN_EXECUTOR_SELECTOR_FILTERS =
-      "azkaban.executorselector.filters";
-  static final String AZKABAN_EXECUTOR_SELECTOR_COMPARATOR_PREFIX =
-      "azkaban.executorselector.comparator.";
-  static final String AZKABAN_QUEUEPROCESSING_ENABLED =
-      "azkaban.queueprocessing.enabled";
-  private static final String AZKABAN_WEBSERVER_QUEUE_SIZE =
-      "azkaban.webserver.queue.size";
-  private static final String AZKABAN_ACTIVE_EXECUTOR_REFRESH_IN_MS =
-      "azkaban.activeexecutor.refresh.milisecinterval";
-  private static final String AZKABAN_ACTIVE_EXECUTOR_REFRESH_IN_NUM_FLOW =
-      "azkaban.activeexecutor.refresh.flowinterval";
   private static final int DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW = 30;
-  private static final String AZKABAN_EXECUTORINFO_REFRESH_MAX_THREADS =
-      "azkaban.executorinfo.refresh.maxThreads";
-  private static final String AZKABAN_MAX_DISPATCHING_ERRORS_PERMITTED =
-      "azkaban.maxDispatchingErrors";
-  private static final String AZKABAN_EVICT_EXECUTIONS =
-      "azkaban.evictExecutions";
   // 12 weeks
   private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
       * 24 * 60 * 60 * 1000L;
@@ -130,12 +108,14 @@ public class ExecutorManager extends EventHandler implements
     this.setupExecutors();
     this.loadRunningFlows();
 
-    this.queuedFlows = new QueuedExecutions(azkProps.getLong(AZKABAN_WEBSERVER_QUEUE_SIZE, 100000));
+    this.queuedFlows = new QueuedExecutions(
+        azkProps.getLong(Constants.ConfigurationKeys.WEBSERVER_QUEUE_SIZE, 100000));
 
     // The default threshold is set to 30 for now, in case some users are affected. We may
     // decrease this number in future, to better prevent DDos attacks.
-    this.maxConcurrentRunsOneFlow = azkProps.getInt(AZKABAN_MAX_CONCURRENT_RUNS_ONEFLOW,
-        DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW);
+    this.maxConcurrentRunsOneFlow = azkProps
+        .getInt(Constants.ConfigurationKeys.MAX_CONCURRENT_RUNS_ONEFLOW,
+            DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW);
     this.loadQueuedFlows();
 
     this.cacheDir = new File(azkProps.getString("cache.directory", "cache"));
@@ -147,9 +127,8 @@ public class ExecutorManager extends EventHandler implements
       setupMultiExecutorMode();
     }
 
-    final long executionLogsRetentionMs =
-        azkProps.getLong("execution.logs.retention.ms",
-            DEFAULT_EXECUTION_LOGS_RETENTION_MS);
+    final long executionLogsRetentionMs = azkProps
+        .getLong("execution.logs.retention.ms", DEFAULT_EXECUTION_LOGS_RETENTION_MS);
 
     this.cleanerThread = new CleanerThread(executionLogsRetentionMs);
     this.cleanerThread.start();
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -32,6 +32,12 @@
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import com.google.common.collect.Lists;
+import org.apache.commons.lang.StringUtils;
+import org.apache.log4j.Logger;
+import org.joda.time.DateTime;
```
HappyRay: Is it possible to gradually move away from joda time to java8's time apis?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -86,6 +87,8 @@
       "azkaban.executorinfo.refresh.maxThreads";
   private static final String AZKABAN_MAX_DISPATCHING_ERRORS_PERMITTED =
       "azkaban.maxDispatchingErrors";
+  private static final String AZKABAN_EVICT_EXECUTIONS =
```
HappyRay: Please move user-facing config keys to the Constant class.

HappyRay:
Thanks for the PR and reporting the issue!

> In practice the automatic evicting has caused more trouble than benefit in our case.
I wanted to simplify the automatic evicting logic and also the load balancing logic for a long while now. However, we haven't prioritized that work.
What if we remove the auto eviction logic entirely for now and re-think the strategy? I don't remember any incident that auto eviction helped in practice. 

We haven't seen this particular problem so far. I guess it is because we run with many executor hosts, set the max concurrency to a low number. Thus each executor doesn't run too many executions concurrently.

What's your setting? How often do you see this issue?

juhoautio:
> What's your setting? How often do you see this issue?

300 concurrent flow executions per executor currently. But would like to bump that up to 1000 at least.

This is not usually a problem for us either. But when we want to submit many executions (typically pipeline level 2) with different date params to backfill a year of data or more, this becomes a problem. We use a single executor and the executions themselves are usually light on the azkaban-executor, the heavy work is done by some external system, so it doesn't seem too sensible having to bump up number of azkaban executors just to have room for more executions.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Actually this was automatic reorganization by the save actions plugin, I didn't import new classes. Can we have this change though, so that next committer doesn't have to worry about these imports flying around?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Ok, will do, sorry.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: Yes, switching to new classes will take some manual work. 

>  Can we have this change though

What do you mean? Keeping joda time?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: No worry. 

Our strategy is to gradually move the keys to the commonplace as we make changes. Do you mind moving the other keys in this class too? 

Thanks

HappyRay:
I can see why this is a problem for you now. 

I would view this new config as a temporary workaround. Are you ok with NOT documenting it as a public API ( config setting ) and reserve the right to deprecate it after we fix the underlying issues? 

Someone from the team will review the details. We agreed internally to have secondary on-call to review new PRs.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Sure I can move them as well

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Ah sorry, I mean can we keep this reordering of imports (that was done automatically)? I don't intend to change the classes being imported.

juhoautio:
> I would view this new config as a temporary workaround. Are you ok with NOT documenting it as a public API ( config setting ) and reserve the right to deprecate it after we fix the underlying issues?

I don't view it as a temporary work-around though. Why should azkaban-web mark executions as failed any way just because it can't reach the executor to get the updates? As long as an executor is active, I would rather see the connectivity problem being highlighted in some other way so that it can be fixed, instead of marking executions as failed. For example e-mail alerts should work.

If it's not already the case, it should be checked if the executor has been removed from active executors or not. If it's been removed, then it's fine to evict executions bound to it. But this shouldn't be confused with problems in calling active executors.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: Oh, sorry, I didn't realize it was just a re-ordering. I thought you added new logic that uses joda-time. 

Please keep this change.

It still happens that auto formatting doesn't cover some cases. It can be because someone didn't set up the save action plugin correctly or the tools we used somehow missed some places.

HappyRay:
@juhoautio I would agree with you that this piece of code has a lot of room for improvement. That's one of the reasons I would like to refactor/simplify it. 
I am interested in your improvement ideas.
I would argue that it should be possible to improve this code so that users like you don't need to resort to configuration tweaks to make it work.

juhoautio:
> I wanted to simplify the automatic evicting logic and also the load balancing logic for a long while now. However, we haven't prioritized that work.
What if we remove the auto eviction logic entirely for now and re-think the strategy? I don't remember any incident that auto eviction helped in practice.

Sorry, I missed this comment before. Sure, I agree 100%. Shall I modify this PR to get rid of eviction-related code?

HappyRay:
I have to refresh my memory about the logic here before I can be of more help with details. But yes, I would like to see your more detailed proposal which may include removing the auto eviction logic.

@jamiesjc I remember you looked at this part of the code in depth in the past. Do you have any comment?

HappyRay:
If you decide to remove the logic instead, how about creating a separate PR? How would that decision affect the other PR #1655 ?

juhoautio:
> If you decide to remove the logic instead, how about creating a separate PR

Sure.

> How would that decision affect the other PR #1655 ?

Doesn't affect it. The 413 problem (when calling executor with many execution ids) should be fixed any way. Whichever route the eviction handling takes, #1655 is a mandatory pre-requisite.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: I created a separate PR for this https://github.com/azkaban/azkaban/pull/1660 (with the exception of not adding a new property for eviction)

juhoautio:
Closing, new take on this at https://github.com/azkaban/azkaban/pull/1661

ameyamk:
@juhoautio I am interested in understanding how we reach this state to begin with (ie. executors not responding correct state to the web server).

Web server not able to talk to executors when executors are actually up should be an extremely rare condition (network partition, or some sort of packet drop etc) - How did you manage to get into such a state?

When such a state is reached. What should be the right behaviour?
Options are:
1. Remove/ Evict execution, email owner of the failure - treat this is any other failure scenario and restart the flow execution

2. Remove/ Evict execution, email owner of the failure - treat this as special failure case - and do nothing. 

3. Only email owner of this special case situation

Anything else?

juhoautio:
@HappyRay what do you mean by "executors are actually up"? I think this happens also if an executor can't respond for any reason. This could be for example because the executor java process is down or because there's some network issue between web & executor. Right? Or if the executor node is too busy somehow so that it can't respond to the call by web node at all or in time.

If executor is running, this could happen due to some bug that causes executor to response with a non-ok response. But it shouldn't happen in any normal situation _if_ the executor _is_ running and web has network connectivity to it.

juhoautio:
@ameyamk question to you ^ (sorry pinged the wrong person at first)

ameyamk:
I am not saying this cannot happen - but this should be arguably corner case and should be a rare occurrence. So I was curious if you are facing this more often in your production env.

Second, I was not clear which of those 3 options you prefer (or if you have 4th one in mind)

juhoautio:
Ping @ameyamk my answer above.. ^

This PR is closed though. I had created a new one: https://github.com/azkaban/azkaban/pull/1661

ameyamk:
ok - I agree with most of the above. We may need to have super admin switch to evict executions (not as a default but having the ability to do as admin). In our case, we have thousands of flows running at the same time. And in case of catastrophic failure - going into each job and cleaning may not be possible. It's much easier to kill all and restart from a clean slate.

juhoautio:
> We may need to have super admin switch to evict executions

When would that be needed? Is there really a situation where you would not rather do either of these:
- Fix connectivity to the executor node so that web can successfully get updates again
- Remove the executor from active executors (if it's not supposed to be enabled any more)

?

Once the connectivity is re-established, executions get marked as failed if they're not running on the executor any more (executor answers that "I don't know this execution"). So there's no need to evict anything manually.

juhoautio:
@ameyamk ping ^

juhoautio:
@ameyamk, please?

PR #1655, opened 2018-02-22
WIP: call executor /update with POST

juhoautio:
This is a candidate fix for https://github.com/azkaban/azkaban/issues/1653.

The idea here is to use POST with form params instead of GET with URL params, so that the number of execution ids passed can be longer.

This implementation is basically ready, but not sure if this works yet, untested end to end.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
+++ b/azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
@@ -21,11 +21,13 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpMessage;
 import org.apache.http.HttpResponse;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.HttpResponseException;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
 import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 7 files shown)


juhoautio:
@HappyRay or someone, even though WIP, could you check if this is a valid strategy in your opinion? Too bad the code was structured so that quite big changes were needed in the infrastructure to support POST in addition to GET.

azkaban-common/src/main/java/azkaban/utils/RestfulApiClient.java
```diff
@@ -222,20 +226,20 @@ public T httpDelete(final URI uri, final List<NameValuePair> headerEntries) thro
    *
    * @param uri the URI of the request.
    * @param headerEntries extra entries to be added to request header.
-   * @param postingBody the content to be posted , optional.
+   * @param params the content to be posted , optional.
    * @return the response object type of which is specified by user.
    * @throws UnsupportedEncodingException, IOException
    */
   public T httpPut(final URI uri, final List<NameValuePair> headerEntries,
```
juhoautio: This method isn't used for anything though. Should it be deleted, or keep for "completeness"?

HappyRay: Please help to delete dead code!

Thanks!

HappyRay:
Your approach seems reasonable to me. 
Longer term, I would like to rethink this part of the code. If possible, retire the entire logic.

Have you considered switching to post only and remove the get support?

juhoautio:
> Have you considered switching to post only and remove the get support?

I have indeed. My change here is backward compatible as long as executor is upgraded first. But why not support all operations with post already now?

HappyRay:
> My change here is backward compatible as long as executor is upgraded first. 

I am not sure this level of backward compatibility is required. We typically upgrade both web server and executors together. 

>But why not support all operations with post already now?

Could you elaborate? I am not sure what your position is.

juhoautio:
>> But why not support all operations with post already now?

> Could you elaborate? I am not sure what your position is.

I just meant that in the ExecutorServlet it would be easy to have the same exact request handling regardless of the HTTP method, be it GET or POST. This would make the code change minimal on executor side, but on the other hand offer a seamless upgrade route if you upgrade the executor servers first. We could leave it like this, wouldn't do much harm.

And if I can switch everything from GET to POST that should simplify the changes on the client side (azkaban-web) quite a lot.

I can make another PR from these ideas.

juhoautio:
Closing in favour of https://github.com/azkaban/azkaban/pull/1659

PR #1656, opened 2018-02-22
Add an API to allow Props to load both current and parent properties 

kunkun-tang:
Today `toProperties` in Props API doesn't include parent properties. In this PR, we add an API to import both current and parent properties. 

Tested locally.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
+++ b/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
@@ -55,7 +55,7 @@ public class QuartzScheduler {
       return;
     }
     final StdSchedulerFactory schedulerFactory =
-        new StdSchedulerFactory(azProps.toProperties());
+        new StdSchedulerFactory(azProps.toAllProperties());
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -55,7 +55,7 @@ public QuartzScheduler(final Props azProps) throws SchedulerException {
       return;
     }
     final StdSchedulerFactory schedulerFactory =
-        new StdSchedulerFactory(azProps.toProperties());
+        new StdSchedulerFactory(azProps.toAllProperties());
```
chengren311: what do you think of reading quartz only properties with azProps.get() vs dumping every properties?
also i don't think we should specify mysql username/password twice in private property file. How about reading mysql username/password from azProps and convert them into quartz properties?

kunkun-tang: Synced up offline. I agree. We may convert quartz properties in Guice Module code.
Adding a todo here for reminder.

PR #1657, opened 2018-02-22
Change project cache to case insensitive.

jamiesjc:
The issue happened when a user uploaded a project through our internal deployment with project name mismatched in case (e.g. expected was my_project, actual was MY_PROJECT). From the UI page, it showed the project was deployed successfully but the lastModifiedTime of the project was not updated. Also, the job properties sometimes disappeared. When the user ran the flow, it actually picked up the old project version.
We figured out that the root cause of these issues is the inconsistency between project cache and Mysql database. The project name key in project cache is case sensitive whereas in Mysql DB it's case insensitive by default. When the case of project name does not match, the project couldn't be found in project cache, but it could be found in DB. As a result, the project object returned was a new instance, and the original cache was not updated.
To fix the inconsistency issue, either we can change Mysql DB to case sensitive by changing the COLLATION, e.g.:
mysql> ALTER TABLE projects MODIFY name VARCHAR(64) CHARACTER SET utf8 COLLATE utf8_bin;
Or we can change project cache to case insensitive. 
I've tried and verified that both can work properly. Since changing the COLLATION would involve creating index in DB, I would prefer changing the project cache instead. Besides, users wouldn't be allowed to create two projects with same letters but different cases, which remains the same behavior as of today.
I shared a document with more details in our internal ticket: LIHADOOP-34114
Also, I explicitly set Mysql DB to case insensitive and changed the H2 DB for unit tests to case insensitive as well.
For those who are using H2 DB for solo server, please note this change could cause potential problem if you already have two projects with different cases in DB. Please have additional check and update the case sensitivity for DB accordingly.

>>>>>>>>>>
Some updates:
I've evaluated both projectsByName cache and projectsById cache, and submitted the fix for the existing project cache synchronization issue together in this PR. 
1. When two users are creating new projects with the same name, only one of them will succeed and the other one will see "project already exists" error message.
2. When one user is removing the project while the other user is fetching the project, the correct project will be returned instead of null.


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -141,10 +142,6 @@ private void loadAllProjectFlows(final Project project) {
     }
   }
 
-  public List<String> getProjectNames() {
```
jamiesjc: This method was not used anywhere, so I removed it.

HappyRay: Thanks!

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -55,8 +56,8 @@
   private final boolean creatorDefaultPermissions;
   private final ConcurrentHashMap<Integer, Project> projectsById =
       new ConcurrentHashMap<>();
-  private final ConcurrentHashMap<String, Project> projectsByName =
-      new ConcurrentHashMap<>();
+  private final CaseInsensitiveConcurrentHashMap<Project> projectsByName =
```
HappyRay: Have you considered using an existing library https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/CaseInsensitiveMap.html ?

HappyRay:
I like the detailed description.

Could you remove the references to terminologies that are internal to Linkedin so that the content is more readable to people outside Linkedin? 

Your choice makes sense to me. 

As discussed off-line, could you consider updating the DB schema to make the sorting order explicit and not relying on the default? I assume this will not trigger an index rebuilt since it will match the current default setting. Also do you plan to make a change to the H2 DB which you found has a different default?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Yes, I've considered the existing CaseInsensitiveMap as well. But it is not synchronized and is not thread-safe. Since the project cache will be accessed by multiple threads, it is necessary to keep it synchronized.

jamiesjc:
@HappyRay Thanks for the comments!
I've submitted new PRs to explicitly set Mysql DB to case insensitive and change the H2 DB for unit tests to case insensitive. I've updated the change description as well.

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
```diff
@@ -0,0 +1,13 @@
+-- DB Migration from release 3.42.0 to 3.43.0
+-- PR #1657 changes project cache to case insensitive.
+--
+-- 1. When creating new database,
+-- use below query to explicitly set the COLLATION to case insensitive for the entire database:
+--
+ALTER DATABASE <database_name> CHARACTER SET utf8 COLLATE utf8_general_ci;
```
jamiesjc: https://docs.microsoft.com/en-us/sql/relational-databases/collations/set-or-change-the-database-collation:
`You can change the collation of any new objects that are created in a user database by using the COLLATE clause of the ALTER DATABASE statement. This statement does not change the collation of the columns in any existing user-defined tables. These can be changed by using the COLLATE clause of ALTER TABLE.`

kunkun-tang: So both mysql and h2 are verified that this query can work?

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
```diff
@@ -103,6 +103,21 @@ public void testCreateProject() throws Exception {
     Assert.assertEquals(project.getLastModifiedUser(), "testUser1");
   }
 
+  @Test
+  public void testCreateProjectsWithDifferentCases() {
+    final String projectName = "mytestproject";
+    final String projectDescription = "This is my new project with lower cases.";
+    final User user = new User("testUser1");
```
kunkun-tang: Is flowname also in-sensive today in mysql database?

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
jamiesjc: No, this query only works for Mysql. H2 and Mysql have different syntax in setting collation (case sensitivity).

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
jamiesjc: Yes. Mysql database is case insensitive by default.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: As discussed offline, please check if external synchronization is already required.

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
HappyRay: Thanks for adding new test cases

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -55,8 +56,8 @@
...
+      new CaseInsensitiveConcurrentHashMap<>();
```
HappyRay: Is it worth adding a new unit test to make sure that the cache stays consistent with the DB schemas?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Thanks for bringing up this discussion. 
ConcurrentHashMap provides concurrent access for multiple threads to the project cache. ConcurrentHashMap is thread safe itself but multiple operations on the map might not be thread safe. (https://stackoverflow.com/questions/14947723/is-concurrenthashmap-totally-safe).
I found two places in the code, in azkaban.project.ProjectManager#createProject, when two threads first check the project cache, the project might not exist yet. One thread puts the new project to the cache first and then might be overwritten by the other thread. 
Also, when one thread is trying to get the project, another thread might call azkaban.project.ProjectManager#removeProject at the same time. The project returned in the first thread could be null. 
To solve the above potential issue, we could add synchronized block on the operations. But there are some tradeoffs. For multiple threads, synchronization will have some effects on the performance. Also, we might have to reevaluate all other cache and see if they require additional synchronization. That might need a separate PR if we want to solve the cache synchronization issue completely.
For this particular PR, I would prefer still keeping the map as concurrent unless we have clear evidence that the performance won't be affected if we use non-concurrent map instead.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Added the test for project cache.

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
```diff
@@ -1,7 +1,13 @@
--- DB Migration from release 3.20.0 to 3.22.0
+-- DB Migration from release 3.42.0 to 3.43.0
+-- PR #1657 changes project cache to case insensitive.
 --
--- Release 3.21.0 is broken
```
HappyRay: Why are these comments removed? Where did they come from?

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
HappyRay: Please leave a comment that this is DB specific.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: Thanks for the summary, as discussed offline. I think it's premature to optimize for performance here. I don't think the performance impact will be big enough to offset the benefits of more reliable/easier to reason about code. 

A separate PR is fine. 

Could you have it reviewed by other team members as well?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: Thanks

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
```diff
@@ -0,0 +1,39 @@
...
+* distributed under the License is distributed on an “AS IS” BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations under
+* the License.
+*/
+package azkaban.utils;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+public class CaseInsensitiveConcurrentHashMap<V> extends ConcurrentHashMap<String, V> {
```
HappyRay: Add a class doc?

e.g.
The name doesn't say if the map key is case insensitive or the value will also be case insensitive.

azkaban-db/src/test/java/azkaban/db/AzDBTestUtility.java
```diff
@@ -36,7 +36,7 @@ public static DatabaseOperator initQuartzDB() throws Exception {
 
     public EmbeddedH2BasicDataSource() {
       super();
-      final String url = "jdbc:h2:mem:test";
+      final String url = "jdbc:h2:mem:test;IGNORECASE=TRUE";
```
HappyRay: Does H2 DB not use the DB schema definition files?

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public V put(final String key, final V value) {
+    return super.put(key.toLowerCase(), value);
+  }
+
+  public V get(final String key) {
```
cwsteinbach: Missing @Override annotations?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: There seems to be a bug in the existing code. Your change and investigation uncovered it. Thanks!

e.g. 
If two threads create projects with the same name at about the same time, it seems that they may both succeed and create two entries in the DB. What's in the cache is not deterministic, the projectsById cache may point to one and the projectsByName cache may point to another.
 
see
https://github.com/HappyRay/azkaban/blob/eb8450a55e03b7a5cbf3768373e5f091804ed880/azkaban-common/src/main/java/azkaban/project/ProjectManager.java#L270

The usage of synchronized keyword in this class is also inconsistent.

Granted that it should be very rare that such a race can happen and one may even argue that the resulting data corruption is acceptable given that it can be traced and corrected manually,  with a restart of the web server. You probably remember how costly and painful it is to debug timing related subtle bugs. I would rather spend time writing code that I have higher confidence that it is correct than debugging race conditions even if it is rare. And like other race condition bugs, it is hard to convince myself that I have covered all cases if we allow operations from multiple threads to interleave. That's what I mean by "easier to reason about" code.  

Given our access pattern and the chance of this race condition, I don't think I have a sufficiently strong argument to block your change. However, I would encourage you to seek other people's opinion and make your own decision. 

If you do decide to keep the current design, I'd suggest you add a comment to the code alerting readers of this code of this potential race condition and your reasoning to do it this way. 

As discussed, it should be possible to run a micro benchmark to evaluate how long the synchronization will add to the user-visible latency. It may be a beneficial exercise for you. It will help you build a better sense of how to estimate performance impact. Seek help from performance team if needed.

jamiesjc: Thanks @HappyRay! I totally agree with your point of fixing the current cache synchronization issue. Otherwise, it will ultimately bite us in the future. In this PR, I'll just focus on the consistency issue between project cache and DB. I'll submit another PR to address the cache synchronization issue and link to this one.

azkaban-db/src/test/java/azkaban/db/AzDBTestUtility.java
jamiesjc: H2 DB has the option of setting case insensitive in the url. It is similar to using a collation with strength PRIMARY.
http://www.h2database.com/html/grammar.html#set_ignorecase

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
jamiesjc: The get method in the parent class requires Object as the argument. Here in the child class, I want to specify String type as the argument because the case-insensitive key comparison (toLowerCase()) only applies to String.

kunkun-tang:
We can only have DB case sensitive or case insensitive, so we don't have an ideal solution for everyone. No matter which solution we accept, one portion of users are affected. Is it possible that we add a method `isDBCaseInsensitive` in DataSource interface, so that it enforces a check if DB is case in-sensitive when webserver or executor starts. AZ could display a strking warning in the log if DB is identified to be case sensitive. 

Regarding Synchronization issue in this PR, I don't have strong opinions on the choices. That would be great that we label/comment places where the race condition could happen. It will be greatly beneficial to the future development.

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
jamiesjc: Previously I committed the wrong file by mistake and then fixed it. Sorry for the confusion. :)

azkaban-db/src/main/sql/upgrade.3.42.0.to.3.43.0.sql
jamiesjc: Sure, added the comment in the new commit.

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
jamiesjc: Sure, added the class doc.

jamiesjc:
@kunkun-tang Thanks for the comments! It is possible that the entire DB is set to case insensitive but the specific columns or tables are case sensitive. I would suggest providing the info so that admins can check DB themselves and update according to their own case.

jamiesjc:
@HappyRay I've evaluated the project cache synchronization issue and submitted the changes in the same PR.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
HappyRay: Thanks. I am glad we are in agreement.

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
HappyRay: I see. Thanks for the explanation.

The concurrentHashMap is a complex class. I looked at the documentation. It doesn't seem to indicate that it is designed for inheritance. Are you sure it is safe to override these methods? Is it going to violate some invariance? 
It's one of the reasons I feel more comfortable with an existing library which has been tried in production. 

If you would like to keep this approach, I'd suggest you think about forwarding the calls to an internal private instance of the concurrentHashMap. This way you may have higher confidence that this code works by reducing the APIs your class exposes.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -254,10 +242,8 @@ public Project getProject(final String name) {
    * fetch active project from cache and inactive projects from db by project_id
    */
   public Project getProject(final int id) {
-    Project fetchedProject = null;
-    if (isActiveProject(id)) {
-      fetchedProject = this.projectsById.get(id);
-    } else {
+    Project fetchedProject = this.projectsById.get(id);
```
HappyRay: This is a clever change. 

I assume this is designed to address your goal
> When one user is removing the project while the other user is fetching the project, the correct project will be returned instead of null.

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
```diff
@@ -0,0 +1,39 @@
...
+    return super.get(key.toLowerCase());
+  }
+
+  public boolean containsKey(final String key) {
+    return super.get(key.toLowerCase()) != null;
```
chengren311: super.containsKey(key.toLowerCase())  ? given ConcurrentHashMap already provides containsKey

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
chengren311: +1 
- Favor composition over inheritance.
https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
jamiesjc: Updated, thanks!

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
jamiesjc: Thanks for the suggestions! I've updated the PR.

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
```diff
@@ -0,0 +1,45 @@
...
+package azkaban.utils;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A concurrent hash map with a case-insensitive string key.
+ *
+ * @param <V> the value type
+ */
+public class CaseInsensitiveConcurrentHashMap<V> {
```
HappyRay: Consider adding unit tests for this class?

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Yes, I combined "containsKey" and "get" operations together so that we can rely on the concurrent map itself for thread safety without adding additional external synchronization.

azkaban-common/src/main/java/azkaban/utils/CaseInsensitiveConcurrentHashMap.java
jamiesjc: Added unit tests in new commits, thanks!

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -89,6 +89,8 @@
   // The flow exec id for a flow trigger instance which hasn't started a flow yet
   public static final int UNASSIGNED_EXEC_ID = -1;
 
+  // The URL setting in H2 database to create tables with case insensitive texts.
+  public static final String H2_DB_CASE_INSENSITIVE = ";IGNORECASE=TRUE";
```
HappyRay: I don't think this constant should belong to this class.

This class is designed to be one place to capture user facing config keys, until we find a better way... 

This constant can stay in the class that uses it.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -53,10 +54,12 @@
   private final ProjectLoader projectLoader;
   private final Props props;
   private final boolean creatorDefaultPermissions;
+  // Both projectsById and projectsByName cache should use the concurrent data structure to
```
HappyRay: Thanks for adding the comment.

This is good. 

What do you think of an alternative?

"Both projectsById and projectsByName cache need to be thread safe since they are accessed from multiple threads concurrently without external synchronization for performance."

HappyRay:
Thanks @chengren311 for pointing out in his earlier comment that we do have

> Favor composition over inheritance.

in https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md

The earlier attempt of extending the ConcurrentHashMap class violates both conditions that are mentioned in the referenced article.
http://thefinestartist.com/effective-java/16

> Safe way to use inheritance
Where the subclass and the superclass implementations are under the control of the same programmers
When extending classes specifically designed and documented for extension

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
jamiesjc: Updated, thanks!

az-core/src/main/java/azkaban/Constants.java
jamiesjc: Sure, moved the constant back to original class.

jamiesjc:
@HappyRay Do you have any other comments on this PR? Is it good to check in?

PR #1661, opened 2018-02-23
No eviction, alert email instead

juhoautio:
Not evicting executions, sending an email instead.

This is to fix https://github.com/azkaban/azkaban/issues/1653 instead of https://github.com/azkaban/azkaban/pull/1654.

In practice it's always better to let the executions remain in running state if the assigned executor can't be reached. If azkaban-web is evicting executions without knowing the actual state, azkaban-web will start new scheduled executions and execution concurrency options (skip/pipeline) can be violated (if the same jobs are still running on an unreachable executor).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1536,8 +1536,8 @@ public class ExecutorManager extends EventHandler implements
     private final int waitTimeIdleMs = 2000;
     private final int waitTimeMs = 500;
     // When we have an http error, for that flow, we'll check every 10 secs, 6
-    // times (1 mins) before we evict.
-    private final int numErrors = 6;
+    // times (1 mins) before we send an email about unresponsive executor.
+    private final int numErrorsBeforeUnresponsiveEmail = 6;
     private final long errorThreshold = 10000;
     private boolean shutdown = false;
 
@@ -1603,29 +1603,26 @@ public class ExecutorManager extends EventHandler implements
                         executor.getPort(), ConnectorParams.UPDATE_ACTION,
                         null, null, executionIds, updateTimes);
               } catch (final ExecutorManagerException e) {
-                logger.error(e);
+                logger.error("Failed to get update from executor " + executor.getHost(), e);
+                boolean sendUnresponsiveEmail = false;
                 for (final ExecutableFlow flow : entry.getValue()) {
                   final Pair<ExecutionReference, ExecutableFlow> pair =
                       ExecutorManager.this.runningFlows.get(flow.getExecutionId());
 
                   ExecutorManager.this.updaterStage =
-                      "Failed to get update. Doing some clean up for flow "
-                          + pair.getSecond().getExecutionId();
+                      "Failed to get update for flow " + pair.getSecond().getExecutionId();
 
-                  if (pair != null) {
-                    final ExecutionReference ref = pair.getFirst();
-                    int numErrors = ref.getNumErrors();
-                    if (ref.getNumErrors() < this.numErrors) {
-                      ref.setNextCheckTime(System.currentTimeMillis()
-                          + this.errorThreshold);
-                      ref.setNumErrors(++numErrors);
-                    } else {
-                      logger.warn("Evicting execution " + flow.getExecutionId()
-                          + ". The executor is unresponsive.");
-                      // TODO should send out an unresponsive email here.
-                      finalizeFlows.add(pair.getSecond());
-                    }
+                  final ExecutionReference ref = pair.getFirst();
+                  ref.setNextCheckTime(System.currentTimeMillis() + this.errorThreshold);
+                  ref.setNumErrors(ref.getNumErrors() + 1);
+                  if (ref.getNumErrors() % this.numErrorsBeforeUnresponsiveEmail == 0) {
+                    // if any of the executions has failed many enough updates, alert
+                    sendUnresponsiveEmail = true;
                   }
+                }
+                if (sendUnresponsiveEmail) {
+                  final Alerter mailAlerter = ExecutorManager.this.alerterHolder.get("email");
+                  mailAlerter.alertOnFailedUpdate(executor, entry.getValue(), e);
                 }
               }
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1535,8 +1535,10 @@ public class ExecutorManager extends EventHandler implements
 
     private final int waitTimeIdleMs = 2000;
     private final int waitTimeMs = 500;
-    // When we have an http error, for that flow, we'll check every 10 secs, 6
-    // times (1 mins) before we send an email about unresponsive executor.
+    // When we have an http error, for that flow, we'll check every 10 secs, 360
+    // times (3600 seconds = 1 hour) before we send an email about unresponsive executor.
+    private final int numErrorsBetweenUnresponsiveEmail = 360;
+    // First email is sent after 1 minute of unresponsiveness
     private final int numErrorsBeforeUnresponsiveEmail = 6;
     private final long errorThreshold = 10000;
     private boolean shutdown = false;
@@ -1615,7 +1617,8 @@ public class ExecutorManager extends EventHandler implements
                   final ExecutionReference ref = pair.getFirst();
                   ref.setNextCheckTime(System.currentTimeMillis() + this.errorThreshold);
                   ref.setNumErrors(ref.getNumErrors() + 1);
-                  if (ref.getNumErrors() % this.numErrorsBeforeUnresponsiveEmail == 0) {
+                  if (ref.getNumErrors() == this.numErrorsBeforeUnresponsiveEmail
+                      || ref.getNumErrors() % this.numErrorsBetweenUnresponsiveEmail == 0) {
                     // if any of the executions has failed many enough updates, alert
                     sendUnresponsiveEmail = true;
                   }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1610,6 +1610,7 @@ public class ExecutorManager extends EventHandler implements
                 for (final ExecutableFlow flow : entry.getValue()) {
                   final Pair<ExecutionReference, ExecutableFlow> pair =
                       ExecutorManager.this.runningFlows.get(flow.getExecutionId());
+                  // TODO can runningFlows.get ever return null, causing NPE below?
 
                   ExecutorManager.this.updaterStage =
                       "Failed to get update for flow " + pair.getSecond().getExecutionId();
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -189,7 +189,7 @@ public class Emailer extends AbstractMailer implements Alerter {
   }
 
   /**
-   * Sends a single email.
+   * Sends a single email about failed updates.
    */
   private void sendFailedUpdateEmail(final Executor executor,
       final ExecutorManagerException exception, final MailCreator mailCreator,
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -22,18 +22,25 @@ import azkaban.Constants;
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.alert.Alerter;
 import azkaban.executor.ExecutableFlow;
+import azkaban.executor.Executor;
 import azkaban.executor.ExecutorLoader;
 import azkaban.executor.ExecutorManagerException;
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.executor.mail.MailCreator;
 import azkaban.metrics.CommonMetrics;
 import azkaban.sla.SlaOption;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Multimaps;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.mail.internet.AddressException;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.log4j.Logger;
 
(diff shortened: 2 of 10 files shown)


juhoautio:
Let's have these first (in order):
https://github.com/azkaban/azkaban/pull/1663
https://github.com/azkaban/azkaban/pull/1664
https://github.com/azkaban/azkaban/pull/1665

Then I can update this PR to build on top of those commits.

juhoautio:
@HappyRay @chengren311 I've rebased & updated this.

juhoautio:
Would you be able to spend a couple minutes on evaluating what you think about this concept of eviction? As I've written in the pr descr. I think it's a problematic behaviour..

juhoautio:
Rebased & resolved conflicts.

I'm hoping to get @ameyamk to accept this change on the conceptual level.

juhoautio:
@HappyRay I heard from Ameya that he moved away from Azkaban team. Can I get you or someone else to discuss this proposal?

HappyRay:
Let me ask the team to see who can take over the review.

jamiesjc:
Hi @juhoautio , I can see your point that the eviction could potentially break the concurrent execution options. However, if the executions remain in running status, should azkaban admins finalize the flows manually in the DB? If the execution status is not finalized, I don't think users will be able to execute the same flow again with the default skip execution option. 
Also when users try to fetch the flow/job log, it will call the executor to fetch it instead of fetching it from DB since the flow status is still running. But there is no response from the executor, so the log would always show empty. Even if we send emails to users, they still have no clue whether their job is actually still running or already failed.

juhoautio:
@jamiesjc thanks for stepping in.

Concurrent option is one issue that I'm trying to fix. Another one is just the reliability of scheduling, ie. not skipping scheduled executions just because executors are unreachable for any reason so that new executions can't be dispatched.

> However, if the executions remain in running status, should azkaban admins finalize the flows manually in the DB?

Not in any normal situation.

(I'm copying some text from https://github.com/azkaban/azkaban/pull/1654)

Is there really a situation where you would not rather do either of these:

- Fix connectivity to the executor node so that web can successfully get updates again
- Remove the executor from active executors (if it's not supposed to be enabled any more)

?

Once the connectivity is re-established, executions get marked as failed (ie. "evicted") if they're not running on the executor any more (executor answers that "I don't know this execution"). So there's no need to evict anything manually.

If an executor is removed from the DB entirely, also then azkaban-web should evict all running executions that are assigned to the non-existing executor.

> Also when users try to fetch the flow/job log, it will call the executor to fetch it instead of fetching it from DB since the flow status is still running. But there is no response from the executor, so the log would always show empty.

Did you know that the flow & execution logs are also written out as blank if azkaban-web evicts the execution? This is because the logs are only finalized into the DB once the execution ends. Until that point azkaban-executor keeps them on local disk. My suggestion doesn't really change this behaviour, except for that user may get some kind of error message in the UI if trying access logs when the corresponding executor is unresponsive. In my opinion that's not a bad thing, because this can be one more way for users to become aware of an on-going executor problem.

> Even if we send emails to users, they still have no clue whether their job is actually still running or already failed.

Sure, but I'm not suggesting to leave the situation like that. Azkaban is not healthy on system level if it can't connect to some of the active executors, and admins should fix such problem asap.

jamiesjc:
@juhoautio Thanks for the detailed explanation. It makes sense to me as I read more about the previous discussions. Just want to confirm with you,
`If an executor is removed from the DB entirely, also then azkaban-web should evict all running executions that are assigned to the non-existing executor.`
If I remember it correctly, this logic has been implemented in your PR #1833 right?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1599,30 +1599,27 @@ public void run() {
...
                         null, null, executionIds, updateTimes);
               } catch (final ExecutorManagerException e) {
-                logger.error(e);
+                logger.error("Failed to get update from executor " + executor.getHost(), e);
+                boolean sendUnresponsiveEmail = false;
                 for (final ExecutableFlow flow : entry.getValue()) {
                   final Pair<ExecutionReference, ExecutableFlow> pair =
                       ExecutorManager.this.runningFlows.get(flow.getExecutionId());
 
                   ExecutorManager.this.updaterStage =
```
jamiesjc: Do we need to check if pair is null?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -146,6 +153,55 @@ public void alertOnSuccess(final ExecutableFlow flow) {
...
+  public void alertOnFailedUpdate(final Executor executor, List<ExecutableFlow> flows,
+      final ExecutorManagerException updateException) {
+
+    flows = flows.stream()
+        .filter(flow -> flow.getExecutionOptions() != null)
+        .filter(flow -> CollectionUtils.isNotEmpty(flow.getExecutionOptions().getFailureEmails()))
+        .collect(Collectors.toList());
+
+    // group by mail creator in case some flows use different creators
+    final ImmutableListMultimap<String, ExecutableFlow> creatorsToFlows = Multimaps
```
jamiesjc: It seems we only have one mailCreator per flow today. Do you think it's necessary to do the grouping here?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -146,6 +153,55 @@ public void alertOnSuccess(final ExecutableFlow flow) {
...
+          .index(creatorFlows, flow -> flow.getExecutionOptions().getFailureEmails());
+
+      for (final List<String> emailList : emailsToFlows.keySet()) {
+        sendFailedUpdateEmail(executor, updateException, mailCreator, emailsToFlows.get(emailList));
+      }
+    }
+  }
+
+  /**
+   * Sends a single email.
```
jamiesjc: Maybe add more details in the java doc?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
```diff
@@ -115,6 +119,26 @@ public void testSendErrorEmail() throws Exception {
         .isEqualToIgnoringWhitespace(this.message.getBody());
   }
 
+  @Test
+  public void alertOnFailedUpdate() throws Exception {
```
jamiesjc: Consider the name `testAlertOnFailedUpdate()`?

azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java
```diff
@@ -145,4 +159,17 @@ public void createSuccessEmail() throws Exception {
         .isEqualToIgnoringWhitespace(this.message.getBody());
   }
 
+  @Test
+  public void createFailedUpdateMessage() throws Exception {
```
jamiesjc: Consider the name `testCreateFailedUpdateMessage()`?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1599,30 +1599,27 @@ public void run() {
...
+                  ref.setNextCheckTime(System.currentTimeMillis() + this.errorThreshold);
+                  ref.setNumErrors(ref.getNumErrors() + 1);
+                  if (ref.getNumErrors() % this.numErrorsBeforeUnresponsiveEmail == 0) {
+                    // if any of the executions has failed many enough updates, alert
+                    sendUnresponsiveEmail = true;
                   }
                 }
+                if (sendUnresponsiveEmail) {
+                  final Alerter mailAlerter = ExecutorManager.this.alerterHolder.get("email");
+                  mailAlerter.alertOnFailedUpdate(executor, entry.getValue(), e);
```
jamiesjc: Will users be flooded with alerting emails when the executor is unresponsive? UpdaterThread checks the executor status frequently and it might keep sending emails before we could take the action to remove the executor from DB.

juhoautio: Yes, it would send the emails ~every 1 minute.

```
// When we have an http error, for that flow, we'll check every 10 secs, 6
// times (1 mins) before we send an email about unresponsive executor.
// private final int numErrorsBeforeUnresponsiveEmail = 6;
```

The idea is to send e-mail again if the problem remains, but this is clearly too often.

 How about sending these emails once per hour, ie. change to `numErrorsBeforeUnresponsiveEmail = 60`?

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
juhoautio: I'd rather not change this. I believe prefixing with `test` doesn't add value. I've seen for example @HappyRay also omit such prefix in some new tests he's created.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: No. As you can see it's never null. Otherwise the existing code would've indeed failed on the next row.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
juhoautio: This is needed to be able to send a single e-mail for any executions that have the same mail creator & recipients. Flows that have different mail creators can not be combined into a single message.

Other option would be to send separate e-mail for every execution, but I think it's definitely nicer to group them.

juhoautio:
@jamiesjc thank you for comments! I made that javadoc update. Anything else?

Note that it's a separate commit, so if merging this it's best to use the squash-merge.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: The existing code actually checks null on line 1615. If you check other places in the existing code when getting the value from runningFlows, it does the check in most cases. I admit the existing code sometimes has some inconsistent behaviors. So I would prefer not making an assumption here that it would never fail simply based on what the existing code was doing.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
jamiesjc: Ok got it, thanks.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Ok, this part is complex enough that I haven't tried to figure out if the null check is actually needed or not.

However the existing placement of the null check didn't matter, because in case of pair=null an NPE would be thrown already before the check, so I removed it. **My change doesn't change existing behaviour, it only simplifies code**.

Btw this happens inside a while loop that is protected by try-catch, so if NPE is thrown, it shouldn't happen on the next loop cycle.

That code could be analyzed & improved, but it's not in the scope of this PR. If NPE is ever thrown here, that would be a bug clearly, but fixing it properly might not be as simple as adding a null check. I don't feel comfortable to blindly introduce a new null check at this point, because I'm not sure if that would change the behaviour in some wrong way, but on the other hand it would hide the actual bug so that it can't be noticed as easily as it would when the NPE gets logged as an error.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1603,30 +1605,29 @@ public void run() {
                         executor.getPort(), ConnectorParams.UPDATE_ACTION,
                         null, null, executionIds, updateTimes);
               } catch (final ExecutorManagerException e) {
-                logger.error(e);
+                logger.error("Failed to get update from executor " + executor.getHost(), e);
+                boolean sendUnresponsiveEmail = false;
                 for (final ExecutableFlow flow : entry.getValue()) {
                   final Pair<ExecutionReference, ExecutableFlow> pair =
                       ExecutorManager.this.runningFlows.get(flow.getExecutionId());
+                  // TODO can runningFlows.get ever return null, causing NPE below?
```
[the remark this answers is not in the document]
jamiesjc: Ok thanks.

azkaban-common/src/test/java/azkaban/utils/EmailerTest.java
HappyRay: 👍

HappyRay:
@juhoautio Thanks for making this change.

We deployed the version 3.59 which contains this change. We started noticing following issues:
Some flows are displayed as running while according to the DB, they have finished. This causes some subsequent flows to fail. It also causes users to get alerts which wouldn't stop. 

We think that there are some problems in the code that manefest themselves more clearly with this change. It may be related to how we deploy changes at Linkedin. 

The deployment to more clusters is currently blocked because of this issue.

I know @jamiesjc and you are discussing how to fix forward. 

I wonder if you are ok with making a change to 
a. rever back to the original behavior of finalizeFlows
b. stop sending alert emails
for now to unblock our next release.
You can keep other changes in.

This will allow you and code reviewers more time to prepare a better fix. I wouldn't want such a change to be rushed.

We should continue to discuss how a longer term fix should look like. 

What do you think?

Ray

HappyRay:
Thank you for your quick response and being cooperative! 

I will need to discuss with @chengren311 and @jamiesjc more on this. But my current understanding is as follows:

After a deployment, we keep the old executor running until all running flows finish. The deployment script will sed the shutdown command to the old executor instance. see https://github.com/azkaban/azkaban/blob/7cd51237ad9e9f1994931cbe64f6dc1a640c0ece/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java#L122

When flows finish, the executor instance is shut down and the DB entry removed. However, the web server keeps checking the running flows' status by asking the now missing executor instances.

@chengren311 @jamiesjc please correct me if I am wrong.

Does it make sense?

I will take a look at your change. I just thought of another workaround. I will discuss with the team first and get back to you.

juhoautio:
Yes, if az-web is not restarted after an executor has been stopped, az-web would keep trying to get updates and fail to remove those executions from the running list.

The plan is to improve this by making azkaban-web check from the db if the executor has been removed when an update call fails. Additionally it can check the flow’s status from the db to find out if it has finished without having to reach the executor.

However, if the executor shutdown is not graceful and the entry is left in the db, admin would still need to somehow make sure that it is removed, if any executions on that executor had been left in some unfinished status.

I think that this is the only way to reliably prevent double-scheduling. On the other hand this could of course be an overkill. It is after all quite rare that az-web evicts some executions that are actually running, which can only happen if it can’t reach an executor that’s still running. Of course even that shouldn’t really happen, especially considering that there have always been some retries before eviction.

Any way, in our case it was rather easy to add an external checker (actually run on azkaban itself) that compares executors in the db with running instances, and removes any missing executors from the db (+restarts az-web to reload).

But now that I think of it, I’m starting to realize that maybe this model doesn’t suit other users too well and automatic eviction should probably remain, at least as the default strategy if not disabled by setting a property.

PR #1662, opened 2018-02-23
Disable or optimize slow unit tests

juhoautio:
Cut ~40s off `azkaban-common` test runtime. Now they run in ~5s for me. Still leaves room for improvement..

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
@@ -88,7 +88,7 @@ public class SleepJavaJob {
     System.out.println("Sec " + sec);
     synchronized (this) {
       try {
-        this.wait(sec * 1000);
+        this.wait(sec * 1000 + 1);
       } catch (final InterruptedException e) {
         System.out.println("Interrupted " + this.fail);
       }
--- a/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
+++ b/azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
@@ -88,7 +88,9 @@ public class SleepJavaJob {
     System.out.println("Sec " + sec);
     synchronized (this) {
       try {
-        this.wait(sec * 1000 + 1);
+        if (sec > 0) {
+          this.wait(sec * 1000);
+        }
       } catch (final InterruptedException e) {
         System.out.println("Interrupted " + this.fail);
       }
--- a/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
@@ -11,6 +11,7 @@ import azkaban.utils.Props;
 import java.util.Date;
 import java.util.List;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
@@ -72,6 +73,7 @@ public class MetricManagerTest {
   /**
    * Test metric reporting methods, including InMemoryMetricEmitter methods
    */
+  @Ignore("TODO replace sleep with something better")
   @Test
   public void managerEmitterHandlingTest() throws Exception {
     this.emitter.purgeAllData();
--- a/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
@@ -10,8 +10,10 @@ import azkaban.metric.inmemoryemitter.InMemoryMetricEmitter;
 import azkaban.utils.Props;
 import java.util.Date;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.joda.time.DateTime;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 /**
@@ -22,6 +24,7 @@ public class MetricManagerTest {
   MetricReportManager manager;
   FakeMetric metric;
   InMemoryMetricEmitter emitter;
+  MetricEmitterWrapper emitterWrapper;
 
   @Before
   public void setUp() throws Exception {
@@ -29,7 +32,8 @@ public class MetricManagerTest {
     this.metric = new FakeMetric(this.manager);
     this.manager.addMetric(this.metric);
     this.emitter = new InMemoryMetricEmitter(new Props());
-    this.manager.addMetricEmitter(this.emitter);
+    this.emitterWrapper = new MetricEmitterWrapper();
+    this.manager.addMetricEmitter(this.emitterWrapper);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
```diff
@@ -88,7 +88,7 @@ public void run() throws Exception {
     System.out.println("Sec " + sec);
     synchronized (this) {
       try {
-        this.wait(sec * 1000);
+        this.wait(sec * 1000 + 1);
```
juhoautio: wait(1ms) when sec=0. wait(0ms) waits forever so that's why this is better.

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
```diff
@@ -72,6 +73,7 @@ public void managerEmitterMaintenanceTest() {
   /**
    * Test metric reporting methods, including InMemoryMetricEmitter methods
    */
+  @Ignore("TODO replace sleep with something better")
```
juhoautio: There's a 2000ms sleep in this method..

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
```diff
@@ -116,6 +117,7 @@ public void testHardlinkCopy() throws IOException {
     assertThat(areDirsEqual(this.baseDir, this.sourceDir, true)).isTrue();
   }
 
+  @Ignore("TODO too slow: over 30s - should this be deleted entirely?")
```
chengren311: thanks for reporting. I agree it's slow. But the purpose of this test is to verify the use case where hard linking a very long dir won't cause issue which happened before in our production. The old code was using linux shell command to do the hard link which will generate excessively long command exceeding the argument limit size.

juhoautio: Do you think that there's a substantial risk that the code would be changed to something else to cause regression?

Any way, I suggest:
- have this test `@Ignore`'d
- add a good comment in the actual code why it's done like it is now (maybe point out that there's an ignored unit test available)

This way it shouldn't be possible that anyone changes the hard linking code to something else, causing regression. But it allows unit tests serve their purpose of being quick to run while making any changes.

azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
HappyRay: Thanks for the explanation.

Is it to be expected that sometimes sec will be 0?

If so, would it be better to add a comment to the code since it is not obvious?

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
HappyRay: @kunkun-tang can you think of a way to make this test run faster?

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
HappyRay: @chengren311 thanks for the review.

It's a tradeoff. Ideally, we should invest in some mechanism to allow slower tests to run automatically to flag potential problem but without blocking commits. Linkedin internal testing infrastructure allows it. We should take advantage of it. e.g. have these tests run in our internal wrappers. I'd suggest that we research how other OSS projects handle this. @juhoautio I am interested in your suggestions. Before that, @juhoautio's suggestion makes sense to me. 

We should have a JIRA to track the long-term improvement plan for how to handle slower tests and set standards.

@ameyamk how do you propose we track the tasks? internal JIRA or Github JIRAs?

azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
juhoautio: > Is it to be expected that sometimes sec will be 0?

Yes

> If so, would it be better to add a comment to the code since it is not obvious?

In that case, what would you think about this instead:

```java
if (sec > 0) {
   this.wait(sec * 1000);
}
```
?

I wanted to avoid such verbosity, but if the +1 way is not clear enough, then I'd rather handle it with self-documenting code like this instead of explaining in a comment.

HappyRay:
Thanks for this PR!

We should regularly review the test runtime. I remember both intellij and gradle have nice reports regarding test run times.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
juhoautio: > @juhoautio I am interested in your suggestions

https://github.com/junit-team/junit4/wiki/categories

azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
HappyRay: The sec seems to control how many seconds to wait before failing the job. Am I right? 

Could you point me to a test that sets the wait time to 0? I can search for it. But since you are more familiar with this code, it's easier for me to ask. 

If 0 value is expected, then wait doesn't make logical sense, even with wait(1). Yes, your suggested alternative looks better to me. 

>  I'd rather handle it with self-documenting code like this instead of explaining in a comment.

In general, yes I agree with you.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
HappyRay: Thanks for the suggestion. This looks like something that can help. We will still need a way to automatically select the different category of tests to run. 

I will invite the team to participate in this discussion.

oliverhu: Not a big fan of adding "ignored", I don't think anyone would bother running the ignored tests. I'm curious what is the actual logic we're trying to test. From the code and conversation, we're not actually testing to logic in `createDeepHardlink(..)` but creating a scenario to make sure the `createDeepHardlink(..)` method works. I suggest:
1) figure out the part of logic that has the possibility of causing regression and guard only that part, proper mocking would make it much faster.
or
2) move these scenario based tests out and create another test suite of integration test to validate scenarios.

juhoautio:
Can I do something more to get this merged? It would be really nice to be able to run all tests quickly without some personal test run setup.

azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
HappyRay: Ping.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
HappyRay: @chengren311 do you have any objection if @juhoautio does what he suggested above?

@oliverhu 
Thanks for your comment. 

I suggest moving forward with this change with some minor modifications ( see this and other comments) and file issues to address the longer-term issues. An alternative is to wait until the improvements you suggested are implemented, which is currently unfunded. cc @ameyamk  IMHO, it is a better tradeoff to make this change as long as we don't delay the longer term investment for too long.

HappyRay:
@juhoautio left comments.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
chengren311: @HappyRay @juhoautio 's suggestion makes sense to me.

azkaban-common/src/test/java/azkaban/executor/SleepJavaJob.java
HappyRay: @juhoautio I am happy to continue to review and accept this PR if you can respond to the comments.

azkaban-common/src/test/java/azkaban/utils/FileIOUtilsTest.java
HappyRay: Thanks @chengren311 

@juhoautio do you mind making your suggested changes above?

juhoautio:
@HappyRay thanks.. I have resolved all slow tests properly, 3 commits added ^

HappyRay:
@juhoautio Thanks. It seems that you changed the metrics related tests implementation to make it faster instead of ignoring it. Shouldn't the PR description be updated to reflect the new changes?

@kunkun-tang could you also review since you are most familiar with this test?

juhoautio:
Updated PR title, original description works.

HappyRay:
@juhoautio Thanks.

@kunkun-tang Ping. 
The sooner we merge this PR, the sooner we can get the benefits.

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
```diff
@@ -75,15 +81,31 @@ public void managerEmitterMaintenanceTest() {
   @Test
   public void managerEmitterHandlingTest() throws Exception {
     this.emitter.purgeAllData();
-    final Date from = new Date();
+    final Date from = DateTime.now().minusMinutes(1).toDate();
     this.metric.notifyManager();
 
-    Thread.sleep(2000);
+    this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
```
kunkun-tang: Great Test strategy to fix long wait.

kunkun-tang: Looks like we can use the same strategy to fix the Thread.sleep issue in 
https://github.com/azkaban/azkaban/blob/d0d63361ca9def7190823a754b28b7dc5d9a980c/azkaban-web-server/src/test/java/azkaban/scheduler/QuartzSchedulerTest.java#L133

HappyRay:
@juhoautio, @kunkun-tang 

Thanks!

PR #1664, opened 2018-02-23
WIP: Implement unit tests for Emailer & EmailMessage

juhoautio:
These were previously ignored. Some refactoring was needed to inject dependencies.

Builds on https://github.com/azkaban/azkaban/pull/1663 (1st commit is the same) – for the actual scope of this PR, see the 2nd commit only: https://github.com/azkaban/azkaban/pull/1664/commits/972e43e2747cd2e0f95b4a4c34ea3a7d7c3beec3

PR #1665, opened 2018-02-23
Refactor Emailer class

juhoautio:
Refactor Emailer class: move duplicate code to methods.

Builds on https://github.com/azkaban/azkaban/pull/1664 (1st & 2nd commit are the same) – for the actual scope of this PR, see the 3rd commit only!


juhoautio:
@chengren311 thanks for merging https://github.com/azkaban/azkaban/pull/1691 – now I was able to do this refactoring as unit tests exist to verify the change.

I split these changes into separate commits so that it's hopefully easier to follow.

HappyRay:
@juhoautio 
Thanks for the refactoring!

Yes, multiple commits make it easier to review. 

I wonder if we should preserve these separate commits when we merge.

PR #1669, opened 2018-02-28
Make it clear which scripts are internal which are user facing

HappyRay:
**Issue:**

It's confusing which script to use.
e.g. here are scripts in the executors' bin directory
azkaban-executor-shutdown.sh
azkaban-executor-start.sh
start-exec.sh

To start executors start-exec needs to be used which calls azkaban-executor-shutdown.sh
However, to shut down the executor, the script azkaban-executor-start.sh
has to be used. The name is closer to the internal version of the start script than the user-facing start script. 

If the wrong script is used to start the service, AZ admins are confused as to why the
servers are not running. This type of mistakes has happened to multiple team members.

**Solution:**

Move the script that is designed to be called by the user-facing scripts to a new directory called internal.

Renamed scripts.

The console logs will now go to the current working directory instead of
 assuming that there is a logs directory under the current working
 directory.

Use the same file structure for the solo server for consistency.

Did some minor code clean up as well to improve readability and confirm
to the Google coding style.

Reduced code duplication in the shutdown scripts.

**Todo:** 
need to update docs to reflect the new script names.

Remove assumptions about where the AZ installation root directory
is and allow the startup and shutdown scripts to be invoked from a
directory other than the installation root. Also makes the conf file
configurations more portable i.e. independent of the working directory.

**Testing:**

Manually tested starting and stopping web, executor and solo servers.

e.g.
[~/oss/azkaban/azkaban-web-server/build/distributions/azkaban-web-server-3.43.0-9-g9545a072 (move-start-stop-scripts)]$ ./bin/start-web.sh

[~/oss/azkaban/azkaban-web-server/build/distributions/azkaban-web-server-3.43.0-9-g9545a072 (move-start-stop-scripts)]$ ./bin/shutdown-web.sh
Killing web-server. [pid: 32249], attempt: 1
shutdown succeeded

Testing executor:
start both an executor and a web server locally.
copy a test conf directory to the extracted binary directory.

Hit
http://localhost:8081/status
{
version: "3.43.0-9-g9545a072",
pid: "32950",
installationPath: "/Users/ruyang/oss/azkaban/azkaban-web-server/build/distributions/azkaban-web-server-3.43.0-9-g9545a072/lib/azkaban-web-server-3.43.0-9-g9545a072.jar",
usedMemory: 122591464,
xmx: 3817865216,
isDatabaseUp: true,
executorStatusMap: {
1: {
id: 1,
host: "localhost",
port: 12321,
isActive: true
}
}
}

[~/oss/azkaban/azkaban-exec-server/build/distributions/azkaban-exec-server-3.43.0-9-g9545a072 (move-start-stop-scripts)]$ ./bin/shutdown-executor.sh
Killing executor. [pid: 32739], attempt: 1
shutdown succeeded

Made sure the process is no longer running.

---
shell script unit tests passed

[~/oss/azkaban/azkaban-common/src/test/bash (move-start-stop-scripts)]$ ./test_util.shunit2
test_is_process_running
test_kill_process_with_retry

Ran 2 tests.

OK


HappyRay:
@chengren311 could you comment here with the sample program that demonstrates your alternative approach here?

HappyRay:
@li-afaris 
Thanks for the quick review!

Actually, I didn't add additional unit tests. I just modified the existing ones you added to make it work with the existing data structure. I did add more comments to the existing unit tests.

As a side note is there any way to run these tests automatically like the Java unit tests? I looked at the shunit project site and couldn't find a way. 
I'd suggest we move to Python as the scripts become more complex. Among other benefits, python has better unit test supports.

chengren311:
an alternative approach is to wrap what's in the internal scripts into a function in user facing script.

E.x:
#!/bin/bash
function internal_start_executor(){
 .....
}

internal_start_executor > out &



One advantage of this approach is less files to maintain making code navigation easier.

HappyRay:
Thanks @chengren311 
I chose the multiple files approach because:

a. It facilitates better code sharing. The same internal scripts can be shared.
b. Users can customize the simpler wrapper script more easily e.g. set different JAVA options. 
c. the readability is just as good IMHO

HappyRay:
Continue the discussion In #1703:

> I'd suggest you start that discussion in the PR that introduced this change. The PR description explains the rationale for that change. Happy to discuss if it is not clear or if you have concerns about the design.

> If you mean this PR #1669, I did read it. However I think I would probably just merge the two start scripts into one instead of keeping the current setup. In fact I think the start-exec.sh is redundant. User should just invoke the main script directly. If they want to redirect output to a file, they can easily wrap that with a one liner in upstart or systemd. As an OSS project, azkaban should let user choose what the launching behavior they want rather than choosing for them. In fact having all pieces in one file is generally what java based apps do when they provide launch scripts. For example: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script

> With that said I think this is out of the scope of this PR. I bring this up b/c otherwise the scripts can be simplified.

Users can still customize and use their own launch scripts today, can't they? 

How would you solve the stated problem in this PR?

xunnanxu:
> Users can still customize and use their own launch scripts today, can't they?

They can, but that means they would invoke `internal/XXX` which seems anti-pattern to me.
I think it would have been a good idea to just run azkaban in foreground by default and offer only one script `start-executor.sh`. In that script we use `exec` such that java process will use the same pid.

In that case if user wants to run it in background and dump pid they can just do

`(./start-executor.sh &); echo $! > currentpid`

And if they want to redirect output they can do
`(./start-executor.sh > output 2>&1 &); echo $! > currentpid`

In other words, azkaban scripts should not do too much IMHO. Users can always handle the output however they want but if azkaban launch scripts do too much, they can't reset the behavior.

xunnanxu:
> How would you solve the stated problem in this PR?

Err what stated problem?

PR #1672, opened 2018-03-02
Rename executor start and shutdown scripts

HappyRay:
Keep the name consistent with the old names to minimize changes in other
 places that still reference the old names.


HappyRay:
Related to #1669

PR #1682, opened 2018-03-09
Node, NPM, and moment version number updates

li-afaris:
This change is because github flagged the version of `moment`
used by Azkaban as having a CVE.  While updating moment, I also made the following changes.

1) In build.gradle, I updated Node & NPM versions.  
       a) The new version of NPM generates a package-lock.json file, which is similar to Ruby's Gemfile.lock file in listing all dependencies and should be checked into the repo to pin versions.
       b) Added a `force` flag to allow the NPM gradle plugin to clean the local NPM cache.  This is required by newer versions of NPM.
2) I used the NPM module `depcheck` to look for unused dependencies and updated package.json accordingly

### DIFF
--- a/azkaban-web-server/build.gradle
+++ b/azkaban-web-server/build.gradle
@@ -15,15 +15,15 @@ buildscript {
 // manage NodeJS distributions, use them from there.
 // ***npm install*** installs all dependencies in package.json. It will only run when changes are made to package.json
 plugins {
-    id "com.moowork.node" version "0.13"
+    id "com.moowork.node" version "1.2.0"
 }
 
 node {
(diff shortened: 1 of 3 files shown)


kunkun-tang:
I know that we don't have javascript integration test. How could we make sure the upgraded js binaries working with the latest version?

HappyRay:
see https://docs.npmjs.com/files/package-lock.json#description for why the file is checked in.

HappyRay:
@li-afaris I remember you said you saw an issue with the scheduling page with your first attempt. I don't see a description of the problem here and its resolution here. Could you provide some details?

Issue #1681, opened 2018-03-08
NPE thrown while shutting down


senecaso-sf:
Using build 3.43.0, I noticed that when I kill the web node, an NPE is logged while attempting to shutdown the Quartz scheduler, even though the Quartz scheduler has been disabled.

```
2018/03/08 00:04:49.599 +0000 INFO [AzkabanWebServer] [Azkaban] Shutting down flow trigger scheduler...
2018/03/08 00:04:49.599 +0000 INFO [QuartzScheduler] [Azkaban] Shutting down scheduler
2018/03/08 00:04:49.600 +0000 ERROR [StdOutErrRedirect] [Azkaban] Exception in thread "Thread-8" 
2018/03/08 00:04:49.600 +0000 ERROR [StdOutErrRedirect] [Azkaban] java.lang.NullPointerException
2018/03/08 00:04:49.600 +0000 ERROR [StdOutErrRedirect] [Azkaban]       at azkaban.scheduler.QuartzScheduler.shutdown(QuartzScheduler.java:108)
2018/03/08 00:04:49.601 +0000 ERROR [StdOutErrRedirect] [Azkaban]       at azkaban.flowtrigger.quartz.FlowTriggerScheduler.shutdown(FlowTriggerScheduler.java:174)
2018/03/08 00:04:49.601 +0000 ERROR [StdOutErrRedirect] [Azkaban]       at azkaban.webapp.AzkabanWebServer$1.run(AzkabanWebServer.java:240)
```

I believe this is happening because the QuartzScheduler class is leaving the `scheduler` variable as `null` if the scheduler is disabled, but during shutdown the `shutdown()` method is called and its not checking `scheduler` [for null first](https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java#L108).

chengren311:
@senecaso-sf  thanks for reporting this issue. The fix(https://github.com/azkaban/azkaban/pull/1684) was put in and we will do a release soon. BTW quartz feature is still under development so we don't expect this to be used in production.

senecaso-sf:
Perfect, thanks!

PR #1684, opened 2018-03-09
fix NPE when shutting flow trigger scheduler

chengren311:
Issue description: https://github.com/azkaban/azkaban/issues/1681

Since webServer.scheduler and webServer.flowTriggerService are injected and always non null, even if quartz is disabled, in old code webServer.scheduler.shutdown() will be still executed throwing NPE if quart is disabled.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
@@ -241,7 +241,7 @@ public class AzkabanWebServer extends AzkabanServer {
             webServer.scheduler.shutdown();
           }
         } catch (final Exception e) {
-          logger.error(("Exception while shutting down flow trigger service."), e);
+          logger.error("Exception while shutting down flow trigger service.", e);
         }
 
         try {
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -235,13 +235,18 @@ public static void launch(final AzkabanWebServer webServer) throws Exception {
...
-        if (webServer.scheduler != null) {
-          logger.info("Shutting down flow trigger scheduler...");
-          webServer.scheduler.shutdown();
+        try {
+          if (webServer.props.getBoolean(ConfigurationKeys.ENABLE_QUARTZ, false)) {
+            logger.info("Shutting down flow trigger scheduler...");
+            webServer.scheduler.shutdown();
+          }
+        } catch (final Exception e) {
+          logger.error(("Exception while shutting down flow trigger service."), e);
```
kunkun-tang: This place doesn't need a parenthesis.

chengren311: good catch

PR #1687, opened 2018-03-13
change blob to longblob to store large project object in quartz table

chengren311:
When a new project containing flow trigger-based flow is uploaded, associated project object will be converted to json and kept as part of quartz job detail and persisted into db. The db column to hold the job detail is of blob type which can only keep 65KB data which is not sufficient. Longblob's limit is 4GB.


chengren311:
closed this PR as a better solution is presented(https://github.com/azkaban/azkaban/pull/1704/).

PR #1690, opened 2018-03-14
Delete unused code in email-related classes

juhoautio:
The removed code is not used even by [ReportalMailCreator.java in azkaban-plugins repo](https://github.com/azkaban/azkaban-plugins/blob/master/plugins/reportal/src/azkaban/viewer/reportal/ReportalMailCreator.java)

(I wasn't able to build azkaban-plugins as a whole, so I copied the `EmailMessage.java` with these modifications on that side and tested that `ReportalMailCreator` still builds in IDEA.)


juhoautio:
@HappyRay could you have a look, please?

PR #1691, opened 2018-03-14
Implement unit tests for Emailer & EmailMessage

juhoautio:
These were previously ignored. Some refactoring was needed to inject dependencies.

- Builds on #1690 which should be merged first (1st commit is the same)
- For the actual scope of this PR, see the 2nd commit only: https://github.com/azkaban/azkaban/pull/1691/commits/2f5014b419353af68a3a5f208dbf9958c03a62cf

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
+++ b/azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
@@ -1,0 +1,42 @@
+package azkaban.utils;
+
+import java.util.Properties;
+import javax.inject.Inject;
+import javax.mail.NoSuchProviderException;
+
+public class EmailMessageCreator {
+
+  public static final int DEFAULT_SMTP_PORT = 25;
+
+  private final String mailHost;
+  private final int mailPort;
+  private final String mailUser;
+  private final String mailPassword;
+  private final String mailSender;
+  private final String tls;
+  private final boolean usesAuth;
+
+  @Inject
+  public EmailMessageCreator(final Props props) {
+    this.mailHost = props.getString("mail.host", "localhost");
+    this.mailPort = props.getInt("mail.port", DEFAULT_SMTP_PORT);
+    this.mailUser = props.getString("mail.user", "");
+    this.mailPassword = props.getString("mail.password", "");
+    this.mailSender = props.getString("mail.sender", "");
+    this.tls = props.getString("mail.tls", "false");
+    this.usesAuth = props.getBoolean("mail.useAuth", true);
+  }
+
+  public EmailMessage createMessage() {
+    final EmailMessage message = new EmailMessage(
+        this.mailHost, this.mailPort, this.mailUser, this.mailPassword, this);
+    message.setFromAddress(this.mailSender);
+    message.setTLS(this.tls);
+    message.setAuth(this.usesAuth);
+    return message;
+  }
+
+  public JavaxMailSender createSender(final Properties props) throws NoSuchProviderException {
+    return new JavaxMailSender(props);
+  }
+}
--- a/azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
+++ b/azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
@@ -1,8 +1,25 @@
+/*
+ * Copyright 2018 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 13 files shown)


juhoautio:
@HappyRay, anyone?

juhoautio:
Resolved conflicts.

juhoautio:
@ameyamk, could someone review this, please? Bumping up the test coverage!

chengren311:
@juhoautio thanks for your continuous contribution. i will take care of it.

juhoautio:
@chengren311 thanks! I can see that the diff is a bit messy.. Unfortunately had to move some code around to allow mocking.

chengren311:
Thanks, took an initial look.
IMO, we are reinventing the wheel by creating Emailer and EmailMessage. There're few well-tested and open source leverageable email client libs : 
e.x: https://commons.apache.org/proper/commons-email/userguide.html
http://www.simplejavamail.org/#/about
https://jodd.org/email/

this PR would help improve test coverage, but in the long term we should simplify email code and take advantage of reasonable 3rd party lib. i remember we have such plan, right @HappyRay? if so should we treat this piece as in maintenance mode and avoid significant change or still need to fix things here and there?

juhoautio:
Aren't these tests needed any way? They're mostly verifying the message formatting that needs to produce the same results regardless of what library is being used. If one were to change the library, isn't it even more important to have good test coverage _before_ the change to support that?

My reason for making this PR is that I would like to add a method in the e-mailer to alert about connectivity error from web node to executor node(s).

azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
```diff
@@ -0,0 +1,42 @@
+package azkaban.utils;
```
chengren311: copyright

azkaban-common/src/test/java/azkaban/utils/EmailMessageCreatorTest.java
```diff
@@ -0,0 +1,31 @@
...
+    final Props props = new Props();
+    props.put("mail.user", USER);
+    props.put("mail.password", PASSWORD);
+    props.put("mail.sender", SENDER);
+    props.put("mail.host", HOST);
+    props.put("mail.port", MAIL_PORT);
+    final EmailMessageCreator creator = new EmailMessageCreator(props);
+    final EmailMessage message = creator.createMessage();
+
+    assertEquals(MAIL_PORT, message.getMailPort());
```
chengren311: can we use assertj since it's our preferred assertion library?
see contributor guide:
```Use AssertJ as the assertion library in unit tests.```

File: azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java [hunk 3]
```diff
@@ -0,0 +1,42 @@
...
+    final EmailMessage message = new EmailMessage(
+        this.mailHost, this.mailPort, this.mailUser, this.mailPassword, this);
+    message.setFromAddress(this.mailSender);
+    message.setTLS(this.tls);
+    message.setAuth(this.usesAuth);
+    return message;
+  }
+
+  public JavaxMailSender createSender(final Properties props) throws NoSuchProviderException {
+    return new JavaxMailSender(props);
```
chengren311: can you help me understand better of the purpose of this method? why do we need to wrap a method around JavaxMailSender(props)?

azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
juhoautio: Added copyright

azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java
juhoautio: A new sender needs to be created every time a message is created (in `EmailMessage.java`):

```java
    final JavaxMailSender sender = this.creator.createSender(props);
    final Message message = sender.createMessage();
```

If you compare this code with the previous implementation, you can see why we need to wrap the code that touches the final / static classes of `javax.mail`, if we want to replace them with mocks in unit tests:

```java
final Session session = Session.getInstance(props, null);
final Message message = new MimeMessage(session);
```

The `JavaxMailSender` just wraps the operations that would use `javax.mail.Session`.

As the class level javadoc on `JavaxMailSender` says:

> Wraps javax.mail features, mostly because Session is a final class and can't be mocked

juhoautio:
@chengren311 sorry for missing the guidelines again. Could you have another look, please?

juhoautio:
ping @chengren311

juhoautio:
Thank you @chengren311 !

PR #1693, opened 2018-03-15
Add proxy check before execute flow

sthbig:
check weather the flow submitter can use a specific user as a proxy which is set via user.to.proxy property

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -884,10 +884,10 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
         final String flowId = getParam(req, "flow");
 
         // check proxy
-        final Map<String, String> props = getParamGroup(req, "flowOverride");
-        if (props != null && props.keySet().size() != 0) {
-            for (String key : props.keySet()) {
-                if ("user.to.proxy".equals(key) && !userManager.validateProxyUser(props.get(key), user)) {
+        final  Map<String,String> props = getParamGroup(req, "flowOverride");
+        if (props != null && props.keySet().size() != 0){
+            for (String key : props.keySet()){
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -883,6 +883,17 @@ private void ajaxAttemptExecuteFlow(final HttpServletRequest req,
     final String projectName = getParam(req, "project");
     final String flowId = getParam(req, "flow");
 
+    // check proxy
+    final  Map<String,String> props = getParamGroup(req, "flowOverride");
+    if (props != null){
+      for (String key : props.keySet()){
+         if ("user.to.proxy".equals(key) && !userManager.validateProxyUser(props.get(key), user)){
```
kunkun-tang: Hi @sthbig , I believe we have this check when job starts to run. What's the advantage having this check here?

sthbig: Hi @kunkun-tang , thanks for reply. Actually, there is no proxy check before executing flow. And the bug can be reproduced.

Following are my steps:
1.Before doing an experiment, I created a user who only has read permission. The conf/azkaban-users.xml looks like this.
```
<azkaban-users>
  <user groups="azkaban" password="azkaban" roles="admin" username="azkaban"/>
  <user password="metrics" roles="metrics" username="metrics"/>
  <user password="read" roles="read" username="read"/>

  <role name="admin" permissions="ADMIN"/>
  <role name="metrics" permissions="METRICS"/>
  <role name="read" permissions="read"/>
</azkaban-users>
```
2.start Azkaban-web-server and login in as read.

3.create a project and execute flow. Before executing flow, set user.to.proxy property to azkaban. 

4.Azkaban completes the job successfully.
azkaban web client 2018-03-19 15-57-50
azkaban web client 2018-03-19 15-58-55

PR #1694, opened 2018-03-15
increase job name varchar size in Mysql table to allow long job name to survive

kunkun-tang:
Users reported that log missed when composed job name was too long. The reason is that Mysql doesn't allow to insert a row when `name`'s length is bigger than 128. This PR increases the length to allow long job name to survive.

Since we don't know the Mysql prefix settings for other Azkaban users, we try to meet the requirement for defaults. In Default Mysql settings, the index key prefix length limit is 767 bytes for innoDB. We have to assume most users using innoDB.

Besides, users should gain difference regarding character seeting in Mysql. In `latin1`, each character is exactly one byte long. In `utf8`, a character can consist of more than one byte, usually 3 or 4. At our settings, we use `latin1`, so we can hold 512 character long name for a job.

Because varchar size enhancement is not a necessary upgrade, schema update file is not recommended here. If people run into the issues above, they should follow the below instructions:
* maximize the Mysql varchar size limit introduced in this PR.
* change the Mysql character setting to `latin1`, if your Azkaban only stores English or other latin characters.

For h2 database, there is no limit for the following entities, except the memory and storage capacity: maximum identifier length (table name, column name, and so on)
http://www.h2database.com/html/advanced.html#limits_limitations

### DIFF
--- a/azkaban-db/src/main/sql/create.execution_jobs.sql
+++ b/azkaban-db/src/main/sql/create.execution_jobs.sql
@@ -3,7 +3,7 @@ CREATE TABLE execution_jobs (
   project_id    INT          NOT NULL,
   version       INT          NOT NULL,
   flow_id       VARCHAR(128) NOT NULL,
-  job_id        VARCHAR(128) NOT NULL,
+  job_id        VARCHAR(512) NOT NULL,
   attempt       INT,
   start_time    BIGINT,
   end_time      BIGINT,
--- a/azkaban-db/src/main/sql/create.execution_logs.sql
+++ b/azkaban-db/src/main/sql/create.execution_logs.sql
@@ -1,6 +1,12 @@
+-- In table execution_logs, name is the combination of flow_id and job_id
+--
+-- prefix support and lengths of prefixes (where supported) are storage engine dependent.
+-- By default, the index key prefix length limit is 767 bytes for innoDB.
+-- from: https://dev.mysql.com/doc/refman/5.7/en/create-index.html
+
 CREATE TABLE execution_logs (
   exec_id     INT NOT NULL,
-  name        VARCHAR(128),
+  name        VARCHAR(640),
   attempt     INT,
   enc_type    TINYINT,
   start_byte  INT,


juhoautio:
A bit less than year ago we bumped `execution_jobs.flow_id` up to `VARCHAR(767)` (just used the max size possible) for similar reason in our Azkaban DBs. It was indeed the `flow_id` column, so different than this PR. The problem occurred for us when there were many embedded sub-flows, as the `flow_id` was composed from all ancestor flow names.

I wonder, has the DB structure been changed since then, or why is it that this PR is not changing `execution_jobs.flow_id` but `execution_jobs.job_id`? Or is it maybe some other kind of case, not one with multiple nested flows with rather long names?

kunkun-tang:
Hi @juhoautio , one of the major bottlenecks should be `name` in table `execution_jobs`, which is the combination and flow name and job name. I believe we should want to maximize `name` firstly.

Furthermore, the reason we don't change `flow_id` is AZ webserver being able to identity too-long-flow name when user uploads the project. Too-long `flow_id` exception is thrown immediately after we upload the project. I experimented embedded flow case, and same there. However, too long 'job_id' cannot be identified at the beginning. it is only thrown when job starts to run. Therefore, I believe we should pay more attention to en-long `job_id` firstly, to guarantee uploaded flows running smoothly. 

I understand you use case here, because you are trying to maximize the embeded_flow length. Unless we enforce AZ users modify mysql server defaults and [increase the innodb prefix length](https://dev.mysql.com/doc/refman/5.7/en/create-index.html), Could you think of more possible solution? @juhoautio

kunkun-tang:
This change has been in production for a while. We should merge this change.

chengren311:
@kunkun-tang 
- I experimented embedded flow case, and same there.
I don't think is true here. I have seen user reporting the issue for over long flow_id.

juhoautio:
Yes, @chengren311 the error does not occur when the flow is uploaded. It's because for embedded flows the concatenated id (parent:flows:embedded_flow:job_id) is only used as a key to store jobs logs. That doesn't happen on project upload. This model is quite limiting though. The index key prefix length limit of 767 is easily too low when you have any deeper nesting of embedded flows.

chengren311:
@juhoautio totally agree. 
So max allowed varchar length is 767? It seems surprisingly low to me. If this is true, how about ``TEXT``?

juhoautio:
@chengren311 it's not the limit of varchar type itself, it's because that column is used as an index key. The term is "index key prefix" and innodb (mysql) has a default limit of 767 for that.

I found this though:

> The index key prefix length limit is 767 bytes if you use the version of MySQL either 5.5 or 5.6. Or you use 5.7 the default is 3076 bytes.

^ https://blog.codingecho.com/2017/08/09/index-key-prefix-length-limitation-on-innodb/

We're using AWS RDS as our MySql DB. Looks like 5.7 is available so we might consider upgrading. Not a perfect fix, but at least it would offer 4x longer values than currently.

kunkun-tang:
@chengren311 We already simplified the flow-id composition format in flow 2.0 Project. The length of new flow-id has been optimized, if users us flow 2.0.

juhoautio:
What about https://github.com/azkaban/azkaban/issues/1809, did it occur in your end in a flow that uses the traditional flow format? Would using flow 2.0 somehow prevent that from happening?

chengren311:
@juhoautio it happened when using traditional flow format.

juhoautio:
Rewind my comment a bit, we recently got an error in writing to `execution_logs`, but that's no surprise as we still had `name        VARCHAR(128)` there. We had only bumped up `execution_jobs.flow_id` to `VARCHAR(767)`.

PR #1695, opened 2018-03-16
Dynamically generate job link URL.

jamiesjc:
For Hadoop/Spark jobs, the job log first stays in resource manager, after some time, it will be moved to job history server. The current web proxy job link in Azkaban user logs doesn't always work properly. When user click the link, it often expires. Users have to search for the application_id and then manually type the RM or JH link to find the actual job log.

I've added a new button on job log page to provide the dynamic job link. See attached images for the UI change. When user click the "Hadoop Job Log" or "Spark Job Log" button, it first logins to RM using authenticated keytab files, then parses the web content to determine whether the job log has expired on RM and been moved to JHS. It constructs the correct job link URL and returns to the user. If the job fails or is killed, the button color will turn red to attract users' attention to click the button and debug their jobs.

To enable this feature, we need configuration changes in azkaban.properties for all clusters. Here is an example:
```
# Hadoop/Spark Job link url
azkaban.server.external.resource_manager_job_url=http://***rm***:8088/cluster/app/application_${application.id}
azkaban.server.external.history_server_job_url=http://***jh***:19888/jobhistory/job/job_${application.id}
azkaban.server.external.spark_history_server_job_url=http://***sh***:18080/history/application_${application.id}/1/jobs
```
Note: ${application.id} will be replaced at run time.

I've enabled the configuration and tested on our testing cluster.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -195,6 +195,16 @@ public class Constants {
 
     // dir to keep dependency plugins
     public static final String DEPENDENCY_PLUGIN_DIR = "azkaban.dependency.plugin.dir";
+
+    /*
+     * Prefix used to construct Hadoop/Spark user job link.
+     * a) RM: resource manager
+     * b) JHS: Hadoop job history server
+     * c) SHS: spark job history server
+     * */
+    public static final String AZKABAN_RM_JOB_LINK = "azkaban.rm.job.link";
+    public static final String AZKABAN_JHS_JOB_LINK = "azkaban.jhs.job.link";
+    public static final String AZKABAN_SHS_JOB_LINK = "azkaban.shs.job.link";
   }
 
   public static class FlowProperties {
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -197,14 +197,15 @@ public class Constants {
     public static final String DEPENDENCY_PLUGIN_DIR = "azkaban.dependency.plugin.dir";
 
     /*
-     * Prefix used to construct Hadoop/Spark user job link.
-     * a) RM: resource manager
-     * b) JHS: Hadoop job history server
-     * c) SHS: spark job history server
+     * Hadoop/Spark user job link.
+     * Example:
+     * a) azkaban.server.external.resource_manager_job_url=http://***rm***:8088/cluster/app/application_${application.id}
+     * b) azkaban.server.external.history_server_job_url=http://***jh***:19888/jobhistory/job/job_${application.id}
+     * c) azkaban.server.external.spark_history_server_job_url=http://***sh***:18080/history/application_${application.id}/1/jobs
      * */
-    public static final String AZKABAN_RM_JOB_LINK = "azkaban.rm.job.link";
-    public static final String AZKABAN_JHS_JOB_LINK = "azkaban.jhs.job.link";
-    public static final String AZKABAN_SHS_JOB_LINK = "azkaban.shs.job.link";
+    public static final String RESOURCE_MANAGER_JOB_URL = "azkaban.server.external.resource_manager_job_url";
+    public static final String HISTORY_SERVER_JOB_URL = "azkaban.server.external.history_server_job_url";
+    public static final String SPARK_HISTORY_SERVER_JOB_URL = "azkaban.server.external.spark_history_server_job_url";
   }
 
   public static class FlowProperties {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -16,7 +16,10 @@
 
 package azkaban.executor;
 
+import static java.util.Objects.requireNonNull;
+
 import azkaban.Constants;
+import azkaban.Constants.ConfigurationKeys;
 import azkaban.alert.Alerter;
 import azkaban.event.EventHandler;
 import azkaban.executor.selector.ExecutorComparator;
@@ -32,9 +35,16 @@ import azkaban.utils.JSONUtils;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import com.google.common.collect.Lists;
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.lang.Thread.State;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.charset.StandardCharsets;
+import java.security.PrivilegedExceptionAction;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -52,9 +62,15 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.security.authentication.client.AuthenticatedURL;
+import org.apache.hadoop.security.authentication.client.AuthenticatedURL.Token;
 import org.apache.log4j.Logger;
 import org.joda.time.DateTime;
 
@@ -75,6 +91,15 @@ public class ExecutorManager extends EventHandler implements
       "azkaban.executorselector.comparator.";
   static final String AZKABAN_QUEUEPROCESSING_ENABLED =
       "azkaban.queueprocessing.enabled";
+  private static final String SPARK_JOB_TYPE = "spark";
+  // The regex to look for while fetching application ID from the Hadoop/Spark job log
+  private static final Pattern APPLICATION_ID_PATTERN = Pattern
+      .compile("application_\\d+_\\d+");
+  // The regex to look for while validating the content from RM job link
+  private static final Pattern FAILED_TO_READ_APPLICATION_PATTERN = Pattern
+      .compile("Failed to read the application");
+  private static final Pattern INVALID_APPLICATION_ID_PATTERN = Pattern
+      .compile("Invalid Application ID");
   private static final String AZKABAN_WEBSERVER_QUEUE_SIZE =
       "azkaban.webserver.queue.size";
   private static final String AZKABAN_ACTIVE_EXECUTOR_REFRESH_IN_MS =
@@ -151,6 +176,16 @@ public class ExecutorManager extends EventHandler implements
     this.cleanerThread = new CleanerThread(executionLogsRetentionMs);
     this.cleanerThread.start();
 
+  }
+
+  private String findApplicationIdFromLog(final String logData) {
+    final Matcher matcher = APPLICATION_ID_PATTERN.matcher(logData);
+    String appId = null;
+    if (matcher.find()) {
+      appId = matcher.group();
+    }
+    this.logger.info("Application ID is " + appId);
+    return appId;
   }
 
   private void setupMultiExecutorMode() {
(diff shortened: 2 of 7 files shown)


jamiesjc:
screen shot 2018-03-16 at 11 53 19 am
screen shot 2018-03-16 at 11 53 03 am

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -195,6 +195,16 @@
...
     // dir to keep dependency plugins
     public static final String DEPENDENCY_PLUGIN_DIR = "azkaban.dependency.plugin.dir";
+
+    /*
+     * Prefix used to construct Hadoop/Spark user job link.
+     * a) RM: resource manager
+     * b) JHS: Hadoop job history server
+     * c) SHS: spark job history server
+     * */
+    public static final String AZKABAN_RM_JOB_LINK = "azkaban.rm.job.link";
```
HappyRay: The '.' in the key names are used to separate namespaces, NOT to separate words. 
How about azkaban.external_resources.resource_manager? 

----
Why is it better than using the full name in the variable name e.g. RESOURCE_MANAGER_LINK ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -757,6 +792,117 @@ public LogData getExecutionJobLog(final ExecutableFlow exFlow, final String jobI
...
+
+    try {
+      url = new URL(this.azkProps.getString(ConfigurationKeys.AZKABAN_RM_JOB_LINK) + applicationId);
+      final HttpURLConnection connection = loginAuthenticatedURL(url);
+
+      try (final BufferedReader in = new BufferedReader(
+          new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8))) {
+        String inputLine;
+        while ((inputLine = in.readLine()) != null) {
+          if (FAILED_TO_READ_APPLICATION_PATTERN.matcher(inputLine).find()
```
HappyRay: Are you parsing the html content here? 
How reliable is this method? What if newer versions change the content?

az-core/src/main/java/azkaban/Constants.java
kunkun-tang: What could be the full value like given this key. Any examples? @jamiesjc 
Is this a resource manager link or a job log link?
Why not call it YARN at least? It should be more straightforward.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -757,6 +792,117 @@ public LogData getExecutionJobLog(final ExecutableFlow exFlow, final String jobI
...
+        }
+      }
+    } catch (final ExecutorManagerException e) {
+      this.logger.error("Failed to get application ID for execution " + exFlow.getExecutionId() +
+          ", job " + jobId + ", attempt " + attempt + ", data offset " + offset, e);
+    }
+    return null;
+  }
+
+  private HttpURLConnection loginAuthenticatedURL(final URL url) throws Exception {
```
kunkun-tang: Can this method move to a util place, and this API can be used by other features in future?

az-core/src/main/java/azkaban/Constants.java
jamiesjc: I'm not aware of the naming convention here. Thanks for pointing it out. 
This is more like a prefix for constructing Azkaban job logs. It's not the specific link for resource manager or job history server. So I think it's necessary to add "azkaban" and "job link" in the variable name. The reason I'm not using full name simply because it would look too long, something like AZKABAN_SPARK_JOB_HISTORY_SERVER_JOB_LINK. But I'm ok with using the full name if it's easier to understand.

jamiesjc: Since it's our internal configuration. I'm not able to give the details here. The value would look like: azkaban.rm.job.link=http://***rm***.linkedin.com:8088/cluster/app/
azkaban.jhs.job.link=http://***jh***.linkedin.com:19888/jobhistory/job/job_
azkaban.shs.job.link=http://***sh***.linkedin.com:18080/history/
As I said, it's a prefix for constructing the job log link. If naming it YARN, people still get confused whether the log is on resource manager or job history server.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: Yes, I'm parsing the html content. I've discussed it with Min and we couldn't think of any better way to do it. If anything changes in Hadoop side, then yes we have to change our code accordingly.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: Sure we can do that.

az-core/src/main/java/azkaban/Constants.java
jamiesjc: @HappyRay I notice we already use "azkaban.server.external" to name some external configurations, so probabaly no need for introducing the new name "external_resources"? In addition to my previous thoughts, I think it would be better to change the names to:
AZKABAN_SERVER_EXTERNAL_RM_JOB_URL = "azkaban.server.external.rm.job.url";
AZKABAN_SERVER_EXTERNAL_JHS_JOB_URL = "azkaban.server.external.jhs.job.url";
AZKABAN_SERVER_EXTERNAL_SHS_JOB_URL = "azkaban.server.external.shs.job.url";
I think using the full name would be a little too long. I've already added comments to clarify the names, so I think it should be easy to understand as well. 
Let me know if you have any concern.

HappyRay: >  AZKABAN_SPARK_JOB_HISTORY_SERVER_JOB_LINK
Why is it necessary to have the Azkaban prefix?

jamiesjc: For Spark jobs launched from Azkaban, the job link is:
http://***sh**.grid.linkedin.com:18080/history/application_***_***/1/jobs/
It's kind of specific to Azkaban. If launched from other places, the link might vary a little. So I added the Azkaban prefix.

jamiesjc: But since this is the only place in our code which references that link and shouldn't cause confusion, I think it's ok to remove Azkaban prefix if it's too long.

jamiesjc: Discussed offline and agreed on the naming convention. See updates in the new commit.

jamiesjc:
To improve this feature, I changed "Hadoop Job Log"/ "Spark Job Log" button color to red if the job fails or is killed. The button will be more noticeable and informative to users so that they are more likely to click the button to debug their jobs. 

screen shot 2018-03-21 at 11 58 05 am
screen shot 2018-03-21 at 12 03 28 pm

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobdetailspage.vm
```diff
@@ -37,6 +37,13 @@
     var jobId = "${jobid}";
     var attempt = ${attempt};
   </script>
+  <script type="text/javascript">
+    $(document).ready(function () {
+      #if ($jobFailed == "true")
```
kunkun-tang: The velocity embedding would be the only advantage of velocity that I could think of

jamiesjc: Yes, it's convenient to use javascript within velocity template.

jamiesjc:
Edited the job URL examples in the comment in a new PR: https://github.com/azkaban/azkaban/pull/1708

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: How do you plan to catch compatibility issues before users do when upgrading Hadoop or Spark?

jamiesjc: We can create a ticket to keep Hadoop team in the loop and notice us if anything changes related to Hadoop/Spark logs. Worst case would be the "Hadoop Job Log" button doesn't show up or the link is invalid. This is not affecting core Azkaban functionality and users still have other ways to find the job logs. This is a nice-to-have feature and has low risk. So I would say we do our best to communicate with Hadoop team to keep us updated with new changes. Let me know if you have any better suggestions.

HappyRay: > We can create a ticket to keep Hadoop team in the loop and notice us if anything changes related to Hadoop/Spark logs.

I don't think this is a realistic expectation. You can test by asking them.

> Worst case would be the "Hadoop Job Log" button doesn't show up or the link is invalid. This is not affecting core Azkaban functionality and users still have other ways to find the job logs. This is a nice-to-have feature and has low risk. So I would say we do our best to communicate with Hadoop team to keep us updated with new changes.

Understood. I would agree not to make the investment to make the code robust and have good test coverage now if this is an experimental feature and we are just testing if users will find it useful and this is not the final state. Is this the case? 
Otherwise, it is a slippery slope IMHO. How would you feel about a product that most of the "non-essential" features are not reliable? What are the "non-essential" features anyway? 

> Let me know if you have any better suggestions.
How difficult will it be to write an automated test to verify this dependency/assumption? Such a test only needs to run before any Hadoop/spark upgrade.

jamiesjc: Of course we would not want a product with unreliable non-essential feature. I might not conveying myself clearly so that there is a misunderstanding here. The difficulty I have to admit is that this feature highly depends on Hadoop side unlike other Azkaban features which we could test before new Azkaban release. Today Azkaban doesn't have UI test framework, otherwise it would be easy to add a test for this hadoop log button. Also the test case is not easy if you want to catch all possible scenarios. For now, I would prefer the most straightforward way which is to just click the button and see if it works when there is any related Hadoop upgrade.

HappyRay: I don't think testing this assumption would require testing UIs. The assumption here is not related to how the AZ button is presented, is it?

> Also the test case is not easy if you want to catch all possible scenarios. 
What are the scenarios you have in mind? 

Happy to discuss in person.

HappyRay:
Hadoop core team members pointed out that PIg/HIve jobs can run multiple MR jobs and the AZ logs may contain multiple links. 
e.g.
a pig job

Job DAG:
25-03-2018 13:06:54 PDT pymk-profile-views INFO - job_1521836707298_331557	->	job_1521836707298_334891,job_1521836707298_335536,
25-03-2018 13:06:54 PDT pymk-profile-views INFO - job_1521836707298_334891	->	job_1521836707298_335424,
25-03-2018 13:06:54 PDT pymk-profile-views INFO - job_1521836707298_335424	->	job_1521836707298_335536,
25-03-2018 13:06:54 PDT pymk-profile-views INFO - job_1521836707298_335536	->	job_1521836707298_337213,
25-03-2018 13:06:54 PDT pymk-profile-views INFO - job_1521836707298_337213

HappyRay:
Have you evaluated the option of just showing the additional job history server URLs after the existing log URLs as I suggested earlier offline?

By doing that the implementation is much simpler and we don't have to worry about the compatibility issues and its associated testing costs. 

I am not sure how much additional value users will perceive the buttons/dropdowns compared with the solution above. I don't expect them to look at the log URLs unless they are debugging an issue. In that situation, they would probably want to look at the complete logs for context anyway. At least they may not mind looking through the logs since they don't need to do it that often hopefully.

And users are already familiar with where to look already. As an example, when you tested it with Subu the other day. He was looking at the log and saw the old link and missed the new button until I told him. 

Have you got some feedback from users which design works better for them?

Why is it considered an experimental feature?

jamiesjc:
I agree with you. As I've mentioned in my previous comments:
`If needed, I can also add another jobhistory server link under the current web proxy link in the user log.`
I think there was some miscommunication about the new feature at the time when I showed to Subu. Anyway we can ask users if they want the additional button feature. One way is to provide the feature for them to try and then collect the feedback. That's part of the reason I want to make it experimental. Also, as we discussed, it is not easy to catch incompatibility issue if there is a Hadoop side change. So I prefer to make it experimental.

HappyRay:
Sounds good. Thanks

jamiesjc:
Originally I was thinking about adding the job history server link below the original web proxy link for each application. However, Azkaban doesn't seem to have the info about when an application starts or completes. It only knows when the entire job completes. So we decide to add the links at the end of the job log for all applications. Disable the button for now since there is some concern about the compatibility with Hadoop.

HappyRay:
Looks like most of the logic don't have unit test coverage. We really should enable code coverage reports.

PR #1704, opened 2018-03-20
move project json out of execution_dependencies table

chengren311:
The initial motivation of keeping project object as part of trigger instance and persist it into db is for consistency purpose - when a new project zip is uploaded, even if web server restarts after that, the running trigger instance would still pick up the old version of the flow to execute for consistency. 

However due to the huge size of project object, the cost of keeping project object for each dependency instance/flow trigger in db is extremely high(suppose a big project containing hundreds of flows, one of which are using flow trigger, then each row of execution_dependencies will contain a large blob field of project json, rendering deserialization costly when recovering incomplete trigger instances before web server restarts).  

So we decided to only keep the project id instead of the project object in database. And we will populate the project object of dependency instance by looking up the id from project database. The downside is we might compromise consistency as we mentioned above since looking up by project id will always fetch the latest-uploaded project.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
+++ b/azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
@@ -19,11 +19,11 @@ package azkaban.flowtrigger.quartz;
 import static java.util.Objects.requireNonNull;
 
 import azkaban.flow.Flow;
-import azkaban.flow.FlowUtils;
 import azkaban.project.FlowLoaderUtils;
 import azkaban.project.FlowTrigger;
 import azkaban.project.Project;
 import azkaban.project.ProjectLoader;
+import azkaban.project.ProjectManager;
 import azkaban.project.ProjectManagerException;
 import azkaban.scheduler.QuartzJobDescription;
 import azkaban.scheduler.QuartzScheduler;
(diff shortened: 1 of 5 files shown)


azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -124,15 +126,13 @@ public void scheduleAll(final Project project, final String submitUser)
...
           final String flowId = jobDataMap.getString(FlowTriggerQuartzJob.FLOW_ID);
-
-          final String projectJson = jobDataMap.getString(FlowTriggerQuartzJob.PROJECT);
-          final Project project = FlowUtils.toProject(projectJson);
+          final int projectId = jobDataMap.getInt(FlowTriggerQuartzJob.PROJECT_ID);
           final FlowTrigger flowTrigger = (FlowTrigger) jobDataMap
               .get(FlowTriggerQuartzJob.FLOW_TRIGGER);
           final String submitUser = jobDataMap.getString(FlowTriggerQuartzJob.SUBMIT_USER);
           final List<? extends Trigger> quartzTriggers = quartzScheduler.getTriggersOfJob(jobKey);
           scheduledFlowTrigger = new ScheduledFlowTrigger(
```
kunkun-tang: Interested to learn why does ScheduledFlowTrigger object need project name? If not, you don't need to inject project object, I believe.

chengren311: it's for UI display.

PR #1706, opened 2018-03-22
make flow trigger plugin manager search classes from plugin jar first

chengren311:
Flow Trigger plugin manager currently prefers class provided by azkaban platform even if there exists same class in plugin jar. This is designated behavior of URL class loader - when loading a class, it searches parent class loader first. (https://docs.oracle.com/javase/7/docs/api/java/net/URLClassLoader.html).  

But our Linkedin's internal dependency plugin jar(Dali: https://engineering.linkedin.com/blog/2017/11/dali-views--functions-as-a-service-for-big-data) has a class conflicting with a dependent class with downgraded version from hive library by Azkaban platform, causing dependency initialization failure.  As for now the quick workaround is to implement custom class loader prefering child classloader first. 

Also there could be some other alternatives:
1. make the conflicting class same by upgrading jar from AZ(short term resolution)
2. let plugin run in isolated process and manage its own classpath.


azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
```diff
@@ -47,19 +50,30 @@
...
-  public FlowTriggerDependencyPluginManager(final String pluginDir) throws
-      FlowTriggerDependencyPluginException {
+  public FlowTriggerDependencyPluginManager(final String pluginDir)
+      throws FlowTriggerDependencyPluginException {
     this.dependencyTypeMap = new ConcurrentHashMap<>();
     this.pluginDir = pluginDir;
+    this.prevClassLoader = Thread.currentThread().getContextClassLoader();
+  }
+
+  private File[] getFilesMatchingPath(final String path) {
```
reallocf: How will this method handle cases like:
`lib/gobblin-*`
?

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
```diff
@@ -108,35 +122,42 @@ private DependencyPluginConfig mergePluginConfig(final Map<String, String> publi
...
+    Thread.currentThread().setContextClassLoader(dependencyClassloader);
+
     Class<? extends DependencyCheck> clazz = null;
     try {
       clazz = (Class<? extends DependencyCheck>) dependencyClassloader.loadClass(pluginConfig.get
           (DEPENDENCY_CLASS));
-      final DependencyCheck dependencyCheck =
-          (DependencyCheck) Utils.callConstructor(clazz);
-      return dependencyCheck;
+      final Object obj = Utils.callConstructor(clazz);
```
reallocf: Is there a reason we're using `Object` here instead of a closer shared parent class? Does `DependencyCheck` fail to work now? I might not understand.

Also, even if we're using `Object` we can probably still use the same more descriptive name of `dependencyCheck` for the variable.

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
```diff
@@ -229,4 +253,79 @@ public void shutdown() {
       }
     }
   }
+
+  /**
+   * A parent-last classloader that will try the child classloader first and then the parent.
+   */
+  private static class ParentLastURLClassLoader extends ClassLoader {
```
reallocf: Do you think this should be pulled out into its own file? I'm not sure - what's your intuition?

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
chengren311: that's good question, actually it will only support dir/*.
Since user shouldn't upload the jars they don't want to import
the reason for supporting dir/* only is to provide another packaging option which let user upload a dir of all required jars instead of one fat jar.

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
chengren311: thanks, updated.

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
chengren311: yeah, i think this class is too big to treated as inner class, move it out. Thanks

azkaban-web-server/src/main/java/azkaban/flowtrigger/plugin/FlowTriggerDependencyPluginManager.java
reallocf: Can you leave a comment mentioning this? I could see somebody (either a dev or a user) getting confused in the future because there's certain expectations around the usage of `*` characters.

chengren311: thanks, added the comment

PR #1708, opened 2018-03-23
Edit the comments for job URL examples.

jamiesjc:
Followed the example in below document to clarify the job URL examples in comments.
https://hadoop.apache.org/docs/current/hadoop-mapreduce-client/hadoop-mapreduce-client-hs/HistoryServerRest.html

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -109,13 +109,13 @@ public class Constants {
     // Designates one of the external link topics to correspond to a job log viewer
     public static final String AZKABAN_SERVER_EXTERNAL_LOGVIEWER_TOPIC = "azkaban.server.external.logviewer.topic";
     public static final String AZKABAN_SERVER_EXTERNAL_LOGVIEWER_LABEL = "azkaban.server.external.logviewer.label";
-    
+
     /*
      * Hadoop/Spark user job link.
      * Example:
-     * a) azkaban.server.external.resource_manager_job_url=http://***rm***:8088/cluster/app/application_${application.id}
-     * b) azkaban.server.external.history_server_job_url=http://***jh***:19888/jobhistory/job/job_${application.id}
-     * c) azkaban.server.external.spark_history_server_job_url=http://***sh***:18080/history/application_${application.id}/1/jobs
+     * a) azkaban.server.external.resource_manager_job_url=http://resource-manager-http-address:port/cluster/app/application_${application.id}
+     * b) azkaban.server.external.history_server_job_url=http://history-server-http-address:port/jobhistory/job/job_${application.id}
+     * c) azkaban.server.external.spark_history_server_job_url=http://spark-history-server-http-address:port/history/application_${application.id}/1/jobs
      * */
     public static final String RESOURCE_MANAGER_JOB_URL = "azkaban.server.external.resource_manager_job_url";
     public static final String HISTORY_SERVER_JOB_URL = "azkaban.server.external.history_server_job_url";


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -109,13 +109,13 @@
...
     public static final String AZKABAN_SERVER_EXTERNAL_LOGVIEWER_LABEL = "azkaban.server.external.logviewer.label";
-    
+
     /*
      * Hadoop/Spark user job link.
      * Example:
-     * a) azkaban.server.external.resource_manager_job_url=http://***rm***:8088/cluster/app/application_${application.id}
-     * b) azkaban.server.external.history_server_job_url=http://***jh***:19888/jobhistory/job/job_${application.id}
-     * c) azkaban.server.external.spark_history_server_job_url=http://***sh***:18080/history/application_${application.id}/1/jobs
+     * a) azkaban.server.external.resource_manager_job_url=http://resource-manager-http-address:port/cluster/app/application_${application.id}
```
HappyRay: Why not use a more realistic server address e.g. 
https://rm.example.com:8443/....

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -109,13 +109,13 @@
...
+     * b) azkaban.server.external.history_server_job_url=http://history-server-http-address:port/jobhistory/job/job_${application.id}
```
HappyRay: Are the rest of the URL patterns expected to be same? i.e. which part of the URL needs to be customizable?

Is it sufficient to ask admins to only provide the hostname, protocol, and port only? 
e.g.
https://rm.example.com:8443 only?

az-core/src/main/java/azkaban/Constants.java
jamiesjc: I just followed the example from the apache website: https://hadoop.apache.org/docs/current/hadoop-mapreduce-client/hadoop-mapreduce-client-hs/HistoryServerRest.html
Also, "rm" is not clear enough to users maybe so I choose the full name.

az-core/src/main/java/azkaban/Constants.java
jamiesjc: It is not guaranteed that the rest of the URL patterns will be the same for external users.

HappyRay: Do you have references to docs that explains how the URL will look like in different configurations?

jamiesjc: I found some related documentation but not exactly explain how the URL will look like. At least some of the URL patterns are not the same as ours.
https://spark.apache.org/docs/2.1.2/monitoring.html
https://archive.cloudera.com/cdh5/cdh/5/hadoop/hadoop-yarn/hadoop-yarn-site/WebServicesIntro.html

az-core/src/main/java/azkaban/Constants.java
HappyRay: I see. The key name has the full name already. So I am not too worried about using the short name in the hostname. How often do you see the full names used in actual server names?

I've seen more example urls using more realistic host names though.

It's a small detail. I will defer to your judgement.

az-core/src/main/java/azkaban/Constants.java
HappyRay: Thanks.

I don't see configurations that would affect how the rest of the URL would look like in https://spark.apache.org/docs/2.1.2/monitoring.html, do you? 

> At least some of the URL patterns are not the same as ours.

Could you give an example?

az-core/src/main/java/azkaban/Constants.java
jamiesjc: I'm ok with your suggestions.

az-core/src/main/java/azkaban/Constants.java
jamiesjc: In the document it mentioned the endpoint for Spark job log is: /applications/[app-id]/jobs, in cluster mode, [app-id] will actually be [base-app-id]/[attempt-id], so the endpoint becomes /applications/[base-app-id]/[attempt-id]/jobs
In our server, the endpoint is /application_[base-app-id]/[attempt-id]/jobs
You can see there is slight different in the pattern.

HappyRay: Could you find out from the Spark team how they configured this different URL patterns and why?

jamiesjc: Sorry about the confusion. The previous link I shared was the Rest API end point for spark job logs, not the web UI end point. I just checked with hadoop team and the URL pattern for is {{uiroot}}/history/{{id}}/{{attemptId}}/jobs/">{{attemptId}} from the code:
https://github.com/apache/spark/blob/master/core/src/main/resources/org/apache/spark/ui/static/historypage-template.html#L79
In most cases, the "uiroot" is hostname:port, but in proxy mode, the pattern could be different. For hadoop jobs, I couldn't find clear documentation about the URL format. I would prefer providing the entire URL link in the configuration in case any external users are using different URL patterns or anything changes in the hadoop code.

PR #1659, opened 2018-02-23
Call executor servlet with POST

juhoautio:
Fix for #1653.

Switch to using POST with form params instead of GET with URL params, so that the number of execution ids passed can be longer.

Also deleting some unused code.

Tested manually that it works like this:
- run AzkabanSingleServer in IDEA with debugger
- start a flow via the UI (http://localhost:8081/)
- set breakpoint to check that requests come to `azkaban.execapp.ExecutorServlet` and are handled successfully
- set breakpoint to check that `ExecutorManager` can get the execution updates

(initial PR & discussion in https://github.com/azkaban/azkaban/pull/1655)

Builds on https://github.com/azkaban/azkaban/pull/1707 to validate the fix.

PR #1711, opened 2018-03-28
All executor servlets to use POST

juhoautio:
`ExecutorApiGateway` was changed to use `POST` for all requests, but I missed that azkaban-web also makes calls to `/serverStatistics`, `/jmx` & `/stats` via the gateway, not just `/executor`.

The problem wasn't seen in manual tests with AzkabanSingleServer because it doesn't use multi-executor mode.

Bug was reported by @chengren311 at https://github.com/azkaban/azkaban/pull/1659#issuecomment-376768014.


chengren311:
@juhoautio thanks for the fix. Would you mind adding the test methodology in the description?

chengren311:
verified the fix in our cluster. thanks!

Issue #1703, opened 2018-03-18
[Enhancement] allow azkaban to be run in foreground


xunnanxu:
Describing the issue
------------------------------
Right now azkaban processes are defaulted to running in background, which is fine for running in large datacenter but not great for dockerization.

Expected result
------------------------------
User can pass arguments to launch scripts to run azkaban in foreground/background.

PR #1715, opened 2018-04-01
#1703 allow azkaban to run in foreground (and possibly replacing the bash process)

xunnanxu:
Issue: #1703 

This is useful when running in docker or other containerized environments:
1. Running in background is only useful if service is supposed to run directly on a machine (either bare metal or cloud instance). If run in background inside a containerized environment, docker daemon would assume it’s already done and thus killing the container.
2. Simply running it in foreground might not be enough as docker shuts down containers through signals. If we fork the java process as usual, the signals will be received by shell process rather than java, thus the server itself is unable to perform graceful shutdown and will be forcefully killed at the end.

This PR introduces:
* allowing user to run azkaban in foreground (`-f` or `--foreground`), in which case we will fork the java process as usual but script logs will go to stdout rather than a file. No pid will be written.
* in addition to that, user can specify `--no-fork` such that the java process will replace the shell one, allowing the server to be directly responding to the control signals.
* `-v` will allow user to see the command that will be run by script for debugging purpose
* fix the missing _ in web server log file name to make it consistent with others

Discussion: I generally doubt why we need to split the original single script to external/internal. That doesn’t seem to provide much benefit but increases the complexity to modify the launch scripts. Users should be able to control the behavior of the scripts on their own (i.e. if they want to redirect output to somewhere rather than being forced to do so). With that said, that’s outside of the scope of this PR.

TODO: documentation?

### DIFF
--- a/azkaban-common/src/main/bash/internal/startup-shared.sh
+++ b/azkaban-common/src/main/bash/internal/startup-shared.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+
+#---
+# parse_args:              parses all given args and handles args that control the behavior of
+#                          launching. All other unrecognized args will be sent to the process
+#                          to be launched at the end.
+#                          Note this function is not supposed to be run in a sub shell, otherwise
+#                          the vars won't be correctly exported to the child process, which is why
+#                          it does not attempt to return the value via stdout.
+# args:                    the name of the var to set final command to, the original command
+#---
+function parse_args {
+  local target_arg="$1"
+  shift
+  local final_command=''
+  for arg in "$@"
+  do
+    case "$arg" in
+      -f | --foreground)
+      # azkaban will run as a foreground process
+      export RUN_IN_FOREGROUND='true'
+      ;;
+      --no-fork)
+      # in addition to running in foreground,
+      # this shell process would be replaced by the actual java service
+      # hence all signals would be directly handled by that
+      export NO_FORK='true'
+      export RUN_IN_FOREGROUND='true'
+      ;;
+      -v | --verbose)
+      export VERBOSE='true'
+      ;;
+      *)
+      final_command="${final_command} ${arg}"
+      ;;
+    esac
+  done
+  eval "${target_arg}='${final_command}'"
+}
+
+#---
+# launch:                  this will modify the command such that the output of the launch script
+#                          itself (not the process to be launched) will be written to the log file
+#                          specified
+# returns:                 the exit status of the command
+# args:                    command to be executed, log file path
+#---
+function launch {
+  local log_file="$2"
+  local cmd=''
+  # this will rip off args specific to launch script
+  # and pass others at the end to the callee
+  parse_args 'cmd' $1
+  if [[ ${RUN_IN_FOREGROUND} != 'true' ]]; then
+    if [[ ${VERBOSE} = 'true' ]]; then
+      echo "Launching process in background..."
+      echo "Execution script logs will be written to ${log_file}"
+    fi
+    cmd="${cmd} >${log_file} 2>&1 &"
+  else
+    if [[ ${VERBOSE} = 'true' ]]; then
+      echo "Launching process in foreground..."
+    fi
+    cmd="${cmd} 2>&1"
+  fi
+  run ${cmd}
+}
+
+#---
+# execute:                 this will modify the command such that if it's run in background,
+#                          it will write the pid to a file.
+# returns:                 the exit status of the command
+# args:                    all arguments will be treated as the command to be executed
+#---
+function execute {
+  local cmd="$@"
+  if [[ ${VERBOSE} = 'true' ]]; then
+    echo "Executing ${cmd}"
+  fi
+  if [[ ${RUN_IN_FOREGROUND} != 'true' ]]; then
+    pid_file="${azkaban_dir}/currentpid"
+    if [[ ${VERBOSE} = 'true' ]]; then
+      echo "Pid will be written to $(abs_path ${pid_file})"
+    fi
+    cmd="${cmd} & echo $! > ${pid_file} &"
+  fi
+  run ${cmd}
+}
+
+#---
+# run:                     runs the given command. If NO_FORK is specified,
+#                          the target process will replace the current shell.
+# returns:                 the exit status of the command
+# args:                    all arguments will be treated as the command to be executed
+#---
+function run {
+  local cmd="$@"
+  if [[ ${NO_FORK} != 'true' ]]; then
+    eval ${cmd}
+  else
+    exec ${cmd}
+  fi
+}
+
+function abs_path {
+  echo $(cd "$(dirname '$1')/.." &> /dev/null && printf "%s/%s" "$(pwd)" "${1##*/}")
+}
+
+# exporting util func specific to this file to the callee
+# this is bash specific so we might need to change this later
+# if we want to support general bourne shell
+export -f execute
+export -f abs_path
+export -f run
--- a/azkaban-exec-server/src/main/bash/internal/internal-start-executor.sh
+++ b/azkaban-exec-server/src/main/bash/internal/internal-start-executor.sh
@@ -57,7 +57,6 @@ else
 fi
 AZKABAN_OPTS="$AZKABAN_OPTS -server -Dcom.sun.management.jmxremote -Djava.io.tmpdir=$tmpdir -Dexecutorport=$executorport -Dserverpath=$serverpath"
 
-java $AZKABAN_OPTS $JAVA_LIB_PATH -cp $CLASSPATH azkaban.execapp.AzkabanExecutorServer -conf $conf $@ &
+java_cmd="java $AZKABAN_OPTS $JAVA_LIB_PATH -cp $CLASSPATH azkaban.execapp.AzkabanExecutorServer -conf $conf $@"
 
-echo $! > $azkaban_dir/currentpid
-
+execute ${java_cmd}
--- a/azkaban-exec-server/src/main/bash/start-exec.sh
+++ b/azkaban-exec-server/src/main/bash/start-exec.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
-
 script_dir=$(dirname $0)
 
-# pass along command line arguments to the internal launch script.
-${script_dir}/internal/internal-start-executor.sh "$@" >executorServerLog__`date +%F+%T`.out 2>&1 &
+source "${script_dir}/internal/startup-shared.sh"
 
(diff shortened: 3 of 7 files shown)


xunnanxu:
@HappyRay @chengren311 Let me know if this makes sense. 1st PR. Not sure whom should be reviewing this or how it should be labeled.

azkaban-exec-server/src/main/bash/internal/internal-start-executor.sh
```diff
@@ -57,7 +57,6 @@ else
 fi
 AZKABAN_OPTS="$AZKABAN_OPTS -server -Dcom.sun.management.jmxremote -Djava.io.tmpdir=$tmpdir -Dexecutorport=$executorport -Dserverpath=$serverpath"
 
-java $AZKABAN_OPTS $JAVA_LIB_PATH -cp $CLASSPATH azkaban.execapp.AzkabanExecutorServer -conf $conf $@ &
-
-echo $! > $azkaban_dir/currentpid
```
HappyRay: Why is it removed?

azkaban-exec-server/src/main/bash/start-exec.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
-
 script_dir=$(dirname $0)
 
+source "${script_dir}/internal/startup-shared.sh"
+
 # pass along command line arguments to the internal launch script.
-${script_dir}/internal/internal-start-executor.sh "$@" >executorServerLog__`date +%F+%T`.out 2>&1 &
+log_file="executorServerLog__$(date +%F+%T).out"
+startup_cmd="${script_dir}/internal/internal-start-executor.sh $@"
```
HappyRay: I like the way you break up long command lines into smaller more readable pieces.

azkaban-common/src/main/bash/internal/startup-shared.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
```
HappyRay: As the scripts become more complex, I'd prefer Python-based implementation over Bash.

HappyRay:
@xunnanxu Thanks for your contribution.

If you haven't already, please familiarize yourself with the contribution guide. See the project readme.

I will let secondary on-call or @li-afaris who is more familiar with operation to review this change.

HappyRay:
> Discussion: I generally doubt why we need to split the original single script to external/internal. 

I'd suggest you start that discussion in the PR that introduced this change. The PR description explains the rationale for that change. Happy to discuss if it is not clear or if you have objection to the design.

xunnanxu:
> I'd suggest you start that discussion in the PR that introduced this change. The PR description explains the rationale for that change. Happy to discuss if it is not clear or if you have concerns about the design.

If you mean this PR https://github.com/azkaban/azkaban/pull/1669, I did read it. However I think I would probably just merge the two start scripts into one instead of keeping the current setup. In fact I think the start-exec.sh is redundant. User should just invoke the main script directly. If they want to redirect output to a file, they can easily wrap that with a one liner in upstart or systemd. As an OSS project, azkaban should let user choose what the launching behavior they want rather than choosing for them. In fact having all pieces in one file is generally what java based apps do when they provide launch scripts. For example: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script

With that said I think this is out of the scope of this PR. I bring this up b/c otherwise the scripts can be simplified.

azkaban-common/src/main/bash/internal/startup-shared.sh
xunnanxu: I think you are generally right but I slightly disagree here b/c that would assume python is available. This is generally true if you run this in general full-fledged linux environment but if you run it inside docker then it's not always there (like popular bare bone alpine image). And it doesn't quite make sense to force people to install python just for this purpose. I even think we should support bourne shell only environment for that purpose to remove the requirement for bash.

xunnanxu: However since bash is typically available in most environments and installing bash is easier than python (as you don't need to worry about 2/3), degenerate to sh is less of a concern here.

azkaban-exec-server/src/main/bash/start-exec.sh
xunnanxu: This is not just for readability but to make it clear that these are the 2 arguments passed to that function. These two are quoted later for that reason.

azkaban-common/src/main/bash/internal/startup-shared.sh
HappyRay: I see your point.

Isn't one of the benefits of using docker to control the dependencies/environment? If the product requires Python (2 or 3), isn't it easier to make sure the requirement is met with a container?

azkaban-exec-server/src/main/bash/start-exec.sh
HappyRay: I see.

HappyRay:
I am glad you have read the contribution guide already. It's just a reminder since these are your first PRs to this project.

HappyRay:
Continued part of the discussion in #1669.

azkaban-common/src/main/bash/internal/startup-shared.sh
xunnanxu: I think it would make sense if user jobs require python to be present. I don't think python should be installed just for the purpose of launching the nodes.

li-afaris: I agree that for startup scripts, bash is sufficient. There was either a thread or Jira on the Apache Yetus mailing list that had a nice discussion of the pro's & cons of replacing bash with python. (Sorry I can't find it tonight.)

Essentially we can get 'good enough' testing by writing functions & using something like shunit2 or bats for testing. The other issue is portability. The differences between bash v3 & v4 are much less then python 2.6 or 2.7. Presuming every OS did ship with python 3.6, we would then need to build a pex file to package our python script and dependencies.

HappyRay: This is a server software, not a utility library or command line tool etc that are designed to run in as many different environments as possible. 

IMHO, it is a better tradeoff to use Python than Bash for developer productivity reasons than to optimize for the ease of deployment. I understand that developers who like Bash better than Python may disagree with the benefits. It's my personal opinion. 

Bash has its place. My evaluation criteria are not the function of the code but the complexity of the code. 

There are better ways to ease deployment. Docker is one of them.

xunnanxu: @HappyRay I disagree. This is an OSS project, not dogfood any more. We should be thinking how users should use our software, not "no you are using it wrong" unless it's made by Apple.

I think my concerns are:
1. Switching to python introduces unnecessary dependency issues. Yes you can say "why don't you set up your environment properly?". But be aware it's tricky to have python 2 and 3 both in the system without help of other software like virtualenv or pex. What if the script assumes 3 while user scripts want 2, or the other way around? This requires either extra build system (which increases complexity in terms of your concern) or requires users to separate the environments for their scripts and azkaban's launch script properly. Both are suboptimal IMHO.
2. It does not necessarily reduce complexity. Most of the complexity in this PR comes from the fact that the logic is unnecessarily split into 2 files (discussion in #1669), in which case to keep the pid both need to be executed with `exec` and vars need to be handled carefully. We would have similar issues if we just switch to python but keep the same design. Again I don't want to side track this PR as that's not our main concern.

xunnanxu: > it is a better tradeoff to use Python than Bash for developer productivity reasons than to optimize for the ease of deployment

does not seem like a good argument to me considering the fact that those scripts have only been touch so many times since the beginning of the 3.x.

azkaban-common/src/main/bash/internal/startup-shared.sh
```diff
@@ -0,0 +1,114 @@
...
+  shift
+  local final_command=''
+  for arg in "$@"
+  do
+    case "$arg" in
+      -f | --foreground)
+      # azkaban will run as a foreground process
+      export RUN_IN_FOREGROUND='true'
+      ;;
+      --no-fork)
```
ashahab: @xunnanxu `--no-fork` should be required if you are planning to run this in a containerized environment. 
The reason is sigterms from docker or other containers would go to the parent process, and it would be entirely up to the parent process to decide whether to propagate the signal to the children. 
For example, bash does not pass signal to children. If you send a sigterm to bash, it will simply exit, and send a sigkill to the children. This prevents the actual forked process(JVM or others) to get gracefully shutdown.

Coming up with a nice parent process that handles all the different signals and orphaned subprocesses is difficult. OS init(pid 1) does that, but it's not advisable to run init in a container. There are alternative implementations, such as dumb_init from Yelp and my_init from baseimage-docker, but both have some cons, and as you say in your comments, this would put additional burden on the users.
`--no-fork` on the other hand gives the user a very simple model. Container starts the main java process as the main process, and sends it all the signals, and container dies when the java process dies.
IMHO we should start with that simple model and let the user community decide if a more complex model is necessary.

azkaban-common/src/main/bash/internal/startup-shared.sh
ashahab: @xunnanxu I am not partial to either Python or Bash, but both scripts can get overly complicated unless you minimize their role(and let the called process do most of the work). Case in point is the Hadoop launch scripts upto 2.7, after which they had to be rearchitected to ensure users understand how arguments are going to be interpreted.

azkaban-common/src/main/bash/internal/startup-shared.sh
xunnanxu: Hi @ashahab, hmm I'm not quite sure if you are proposing something new here or just repeating my suggestion in this PR. In fact I think "no fork" should be the default option (see new discussion in  #1669). HOWEVER, the current default behavior is to run it in background. I don't want to change that because users might be relying on that behavior already so it's not something we should be changing in 3.x. This PR just gives other users (like me) an option to run stuff differently.

ashahab:
What I meant was, instead of relying on the users to know that no-fork is
advisable when running in a container, can you detect that the script is
running in a container and accordingly set no-fork as a default? That would
stop users from running without no-fork in a container and seeing
unpredictable shutdown behavior.

xunnanxu:
@ashahab I don't think there's a bullet proof way to detect that. On the other hand, user should know what they want. Containerization is just one of the techniques to get an isolated environment ready quickly. What if they just run it directly in a cloud instance using AMI for example?

Issue #1699, opened 2018-03-18
[Enhancement] executor min memory filter should be configurable


xunnanxu:
Describing the issue
------------------------------
Right now it's hard-coded to 6GB which doesn't make sense at all unless you run this on an enormous cluster machine.

https://github.com/azkaban/azkaban/blob/3.44.0/azkaban-common/src/main/java/azkaban/executor/selector/ExecutorFilter.java#L131

Expected result
------------------------------
User should be able to configure it in `azkaban.properties` through a config value

Design issues
------------------------------
* How should user specify that in config?
May be something like
```
azkaban.executorselector.filter.MinimumFreeMemory.minMemoryInMBs=300
```
And each corresponding filter would get a map of configs (do we enforce names cannot have whitespaces?).

or (to allow whitespaces)
```
azkaban.executorselector.filter.configs.0=MinimumFreeMemory:minMemoryInMBs:300
azkaban.executorselector.filter.configs.1= CpuStatus:maxCpuUsagePercent:60
```

PR #1716, opened 2018-04-02
#1699 make executor filters configurable

xunnanxu:
This basically implements what's suggested in #1699 (using proposal 1)

Please see individual commits to see what's changed.

Added unit tests.

### DIFF
--- a/az-core/src/main/java/azkaban/filter/Filter.java
+++ b/az-core/src/main/java/azkaban/filter/Filter.java
@@ -1,0 +1,24 @@
+package azkaban.filter;
+
+/**
+ * This lists the methods required to build a filter.
+ *
+ * @param <T>: the type of the objects to be compared.
+ * @param <V>: the type of the object to be used for filtering.
+ */
+public interface Filter<T, V> {
+
+  /**
+   * function to analyze the target item according to the reference object to decide whether the
+   * item should be filtered.
+   *
+   * @param filteringTarget   object to be checked.
+   * @param referencingObject object which contains statistics based on which a decision is made
+   *                          whether the object being checked need to be filtered or not.
+   * @return true if the check passed, false if check failed, which means the item need to be
+   * filtered.
+   */
+   boolean filterTarget(final T filteringTarget, final V referencingObject);
+
+
+}
--- a/az-core/src/main/java/azkaban/utils/Props.java
+++ b/az-core/src/main/java/azkaban/utils/Props.java
@@ -36,6 +36,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeMap;
+
+import com.google.common.base.Strings;
 import org.apache.log4j.Logger;
 
 /**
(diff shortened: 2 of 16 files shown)


HappyRay:
I would prefer simplifying the dispatching logic. I haven't seen evidence that the current design provides meaningful benefits in practice.

xunnanxu:
@HappyRay could you elaborate on what you mean by "meaningful benefits"? This is useful to prevent instances from being overloaded IMHO and it's part of dispatching logic today.

HappyRay:
Could you describe in more detail how the filter setup was broken for you?

There are configurations to control concurrency in executors. And recent changes made by @chengren311 will queue java jobs if memory is running low. Do you run java based jobs? Are these measures not enough for your use cases? Could you explain how the filters will work better for you?

HappyRay:
No, random assignment is not implemented today. It's a thought experiment. And we could implement it.

xunnanxu:
> Could you describe in more detail how the filter setup was broken for you?

As described in the issue, the hard-coded 6GB memory limit does not work if executor runs on an instance with 4/8GB only. 95% hard-coded CPU limit doesn't work for a similar reason. I have to not use the filters but implement other monitoring mechanisms to avoid nodes from being overloaded.

> recent changes made by @chengren311 will queue java jobs if memory is running low.

What changes are you specifically talking about?

> Do you run java based jobs?

We mainly run python jobs.

> Could you explain how the filters will work better for you?

The hard coded values should be replaced by configurable values and hence this PR.

HappyRay:
> recent changes made by @chengren311 will queue java jobs if memory is running low.
@chengren311 do you remember your PR link 

@xunnanxu I am not seeing some of the answers I am looking for. Maybe my questions are not clear enough. 

If it is not clear to you yet, I am interested in retiring this entire feature and replace it with random assignment and leverage flow controls at the individual executor level.

xunnanxu:
> If it is not clear to you yet, I am interested in retiring this entire feature and replace it with random assignment and leverage flow controls at the individual executor level.

I'm sry but I don't see the logical transition here. What's the answer you are expecting? What's the reason for "retiring this entire feature"? After all you have not yet proven to me that random assignment is better than current design after I typed this much. I guess I still don't understand why we keep talking about random assignment as that's more of an alternative to comparator in current design not filter.

Users should have the right to choose what they want. If you want to rip off all the code and replace it with something else that's ok but that should be on the 4.x road map.

xunnanxu:
I don't honestly care about what random assignment is. Maybe that's better. But my point is we should not be randomly taking features away in a public software project.

chengren311:
@HappyRay the PR to make job retry when memory is not enough: https://github.com/azkaban/azkaban/pull/937

xunnanxu:
Hmm after checking that PR if I understand your proposal correctly, you want to let scheduler send jobs to executor regardless, and let executor determine when to actually execute it based on certain conditions.

This might be fine in certain scenarios but I don't think that is right in general. Think of a scenario where executors run long-running jobs. If new jobs get blindly dispatched to busy executors, they will be stuck there forever until other jobs finish, which might take hours even days. That will severely delay the new jobs, whereas in the current model, they will be queued on the web node side, and I can easily mitigate the congestion by adding more executor nodes.

If you want to let executor do the job, the only right way I can think of is changing the push model to pull model. That means jobs will only get picked up by an executor node if it is capable of handling it (or it thinks so), otherwise it will stay in a queue. However that's a fundamental change and I don't think it can handle more sophisticated scheduling logic.

Again, I think users should have the right to choose what they want. That is one of the fundamental principles of building any successful OSS projects.

ameyamk:
I think it seems like we are mixing two separate issues:

Issue 1:  Unless there is at least 6 GB memory free on executor web server is not dispatching executions to this executor and 6 GB is not configurable - this creates issues in smaller VM instances.
(Which this PR attempts to solve)

Issue 2: Dispatching logic in itself may not be optimal, and perhaps random assignment is one possible solution
(Which this PR is not trying to solve)

I think #2 is a broad issue - and needs more discussion. I have created Issue #1739 to track this.

I do not see why we should not fix issue 1 right now unless we have an alternative easier fix available.

PR #1717, opened 2018-04-03
Move jobtype module to the main AZ repo

kunkun-tang:
This PR is a followup of #1712. We move jobtype module from azkaban-plugins to the main AZ repo, including their original tests.

Able to run test successfully and generate the corresponding jar. Verified the jar in the staging AZ.


HappyRay:
@kunkun-tang:

Thanks

Does it make sense to break up the plugins into their own subprojects? This will improve build/test parallelism and modularity in general. 

Why does this jar file need to be checked-in?

A az-jobtype-plugin/lib/teradata-connector-1.4.2.jar (0)

kunkun-tang:
1). This is a teradata jar, and I cannot find it in central maven.

2). That sounds a good direction to follow-up. Today we don't have HadoopPig.jar or gobblin.jar. Instead, we only have one jobtype.jar to represent all these plugins. We may consider break up all these code to submodule in future refactors.

kunkun-tang:
It might work, but the gradual migration solution should be taken care of carefully. If we move part of code to here, we may rename this package to az-hadoop-jobtype-plugin, and we should immediately remove the migrated code in azkaban-plugins repo, in order to avoid two jars having duplicate classes issue. How does it sound?

HappyRay:
@kunkun-tang your suggestion sounds good to me. Make sure you modify the deployment pipeline accordingly.

kunkun-tang:
I removed teradata related code, including cypto plugin. @HappyRay

kunkun-tang:
That's a good suggestion. Just created an Epic to track

PR #1721, opened 2018-04-05
Flow trigger service enhancement

chengren311:
The PR consists of following fix/enhancement around flow trigger service:

1. Cancel non-recoverable trigger instances after web server restarts:
 the following situation would cause trigger instances unrecoverable:
 1.1. project A with flow A associated with flow trigger A is uploaded
 1.2. flow trigger A starts to run
 1.3. project A with flow B without any flow trigger is uploaded
 1.4. web server restarts
 In this case, flow trigger instance of flow trigger A will be equipped with latest project, thus failing to find the flow trigger since new project doesn't contain flow trigger at all.

2. fix edge cases in getIncompleteTriggerInstances

3. shutdown executor service of TriggerInstanceProcessor when shutting down FlowTriggerService.


chengren311:
@kunkun-tang thanks!

PR #1722, opened 2018-04-05
Add user.to.proxy validation before reportal uploads the user zip to AZ

kunkun-tang:
A user is able to `user.to.proxy` any user in Reportal today. This has been a security vulnerability issue for long time. This PR proposes a fix to validate user.to.proxy configuration before the zip being uploaded to AZ. If the session user is not part of userToProxy user, error is returned to browser to prevent users saving the Reportal.


reallocf:
Change looks pretty good to me. Is there any way to write unit tests for these kinds of things?

kunkun-tang:
@reallocf Sorry about the late replies. Yes, it would be great that we can have unit tests for servlet methods. Probably, we should set up Servlet testing infrastructure to evaluate all rest endpoitns.

PR #1726, opened 2018-04-11
move example pig script back to jobtype plugin package

kunkun-tang:
That was a fault to keep the pig file in a non-module folder. So moving it back to az-hadoop-jobtype-plugin module.


HappyRay:
The file names/paths look the same to me.

What was the problem?

kunkun-tang:
I was trying to rename the whole package from az-jobtype-plugin to az-hadoop-jobtype-plugin.

I used Intellij's rename-refactor function. For some reason, it didn't work very well.

Issue #617, opened 2016-03-09
"alert.type" property


kddeisz:
Is there a way to set this property globally and have it apply to every flow? At the moment the only way I can figure out to set this property is through the override properties in the interface, and I'd much rather just set it in the job file or globally for all jobs.

kddeisz:
@logiclord Tagging you because it seems like you're the only one contributing to this anymore.

Is the above thing an issue or am I doing something wrong? Happy to work on it if it's an issue.

logiclord:
@kddeisz You are right. alert.type is part of flow properties section, which can only be specified overriide properties interface. We do have a backlog item to specify flow properties as part of zip project like .job properties file. Feel free to contribute on this

kddeisz:
Thanks @logiclord, is there already an existing issue? Otherwise I'll just reference this one in the future

logiclord:
I tried searching older ticket but couldn't find it. Feel free to propose something for this and refer this issue.

antey86:
https://github.com/azkaban/azkaban/pull/1727

Issue #480, opened 2015-09-03
Alerter doesn't recognise global.properties for alert.type


baank:
Custom alerters can not be set for all Azkaban Flows in global.properties or even uploaded flow properties files.

alert.type is specifically checked from the UI flow parameters only. It should check all possible configuration locations.

antey86:
Hi Azkaban Team,

I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. 

Besides we've experienced the same issue, so this fix would be helpful

Approach is straightforward - read **gloabal.properties** into some **final globalProps** at **ExecutorManager**, and use it in case if **options.getFlowParameters()** doesn't contain key **alert.type**

Thanks!

antey86:
The fix https://github.com/azkaban/azkaban/pull/1727

dmsolr:
What is different between 'flowParameters' and 'flow.props'.

in my mind, flowParameter should including the flow.props(global.properties).

but it's not.

PR #1727, opened 2018-04-12
Issues #480 #617. Alerter doesn't recognise alert.type from  global.properties (only from UI), refactoring of Alert notification code

antey86:
The fix for

- [Alerter doesn't recognise global.properties for alert.type #480](https://github.com/azkaban/azkaban/issues/480)

- ["alert.type" property](https://github.com/azkaban/azkaban/issues/617)

1. Custom alerters can not be set for all Azkaban Flows in global.properties or even uploaded flow properties files.
alert.type is specifically checked from the UI flow parameters only. It should check all possible configuration locations.

2. Refactoring of Email/Alert notification code has been performed


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -154,6 +159,28 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader loader,
...
+  }
+
+  private Props setupGlobalProps(Props azkProps) {
+    String globalPropsPath = azkProps.getString("executor.global.properties", null);
+    if (Objects.nonNull(globalPropsPath)) {
+      try {
+        if (globalPropsPath.startsWith("/")) {
+          return new Props(null, new File(globalPropsPath));
+        } else {
+          final URL globalPropsUrl = Resources.getResource(globalPropsPath);
```
reallocf: Do we need to introduce the new `Resources` dependency? Can we accomplish the same thing with the the Java `File` library?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -154,6 +159,28 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader loader,
     this.cleanerThread = new CleanerThread(executionLogsRetentionMs);
     this.cleanerThread.start();
 
+    this.globalProps = setupGlobalProps(azkProps);
+  }
+
+  private Props setupGlobalProps(Props azkProps) {
+    String globalPropsPath = azkProps.getString("executor.global.properties", null);
+    if (Objects.nonNull(globalPropsPath)) {
```
reallocf: `Objects.nonNull` isn't a typical way that we check if a variable is null.

Maybe we can just use `globalPropsPath != null`?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1262,60 +1289,52 @@ private void finalizeFlows(final ExecutableFlow flow) {
     // the reference.
 
     this.updaterStage = "finalizing flow " + execId + " alerting and emailing";
-    if (alertUser) {
-      final ExecutionOptions options = flow.getExecutionOptions();
-      // But we can definitely email them.
+    if (alertUser) sendAlertNotification(flow);
```
reallocf: Can we include braces here?

So something like:
```
if (alertUser) {
    sendAlertNotification(flow);
}
```

antey86:
fixed

antey86:
All notes were fixed, any reason to close this PR?

antey86:
@burgerkingeater please review

PR #1734, opened 2018-04-16
overrideSuccessEmails and overrideFailureEmails have mixed up its values

antey86:
**overrideSuccessEmails** and **overrideFailureEmails** have mixed up its values

`   var overrideSuccessEmails = this.model.get("failureEmailsOverride");
     var overrideFailureEmails = this.model.get("successEmailsOverride");
`


chengren311:
@antey86 thanks for your contribution! Would you add more context of this fix, like what's the behavior before and after. you can also post screenshot in change description for better illustration.

PR #1647, opened 2018-02-16
Solo Server can be started from any path

mtrna:
instead of `cd azkaban-solo-server/build/install/azkaban-solo-server; bin/azkaban-solo-start.sh` [1] you can directly run `azkaban-solo-server/build/install/azkaban-solo-server/bin/azkaban-solo-start.sh`

[1] http://azkaban.github.io/azkaban/docs/latest/#solo-setup

Issue #1725, opened 2018-04-10
3.45.2 build failed (1 test failed)


wilson-lauw:
Here is the output:

./gradlew build
Parallel execution with configuration on demand is an incubating feature.
Build cache is an incubating feature.

> Task :az-hadoop-jobtype-plugin:compileTestJava
Note: /home/traveloka/workspace/azkaban/az-hadoop-jobtype-plugin/src/test/java/azkaban/jobtype/TestHadoopJobUtilsFilterCommands.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

> Task :azkaban-common:test

azkaban.metric.MetricManagerTest > managerEmitterHandlingTest FAILED
    java.lang.AssertionError: Failed to report metric expected:<1> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at azkaban.metric.MetricManagerTest.managerEmitterHandlingTest(MetricManagerTest.java:92)

323 tests completed, 1 failed, 22 skipped

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':azkaban-common:test'.
> There were failing tests. See the report at: file:///home/traveloka/workspace/azkaban/azkaban-common/build/reports/tests/test/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 16s
66 actionable tasks: 43 executed, 23 from cache

juhoautio:
Seems to be my fault, caused by https://github.com/azkaban/azkaban/pull/1662. I must've missed synchronizing something.

juhoautio:
It seems that the actual issue was leaking DateTimeUtils tweaks from other tests, see fix PR at https://github.com/azkaban/azkaban/pull/1735

PR #1735, opened 2018-04-16
Tests should produce less false positives

mtrna:
Tests should produce less false positives *without intervention in the environment* by:
* ~~increasing time interval in a `MetricManagerTest`~~
* fixing the source for computing of the time interval in a `MetricManagerTest`
* closing file descriptors in a `DirectoryFlowLoaderTest`

Also, some typos were fixed.

This MR should not have any functional impact. Goal is to improve development experience.

Consider merging https://github.com/azkaban/azkaban/pull/1647, please. Thanks! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ temp/
 .shelf
 # Intellij default build output directory
 out/
+
+currentpid
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1734,7 +1734,7 @@ public class ExecutorManager extends EventHandler implements
 
     private static final long QUEUE_PROCESSOR_WAIT_IN_MS = 1000;
     private final int maxDispatchingErrors;
-    private final long activeExecutorRefreshWindowInMilisec;
+    private final long activeExecutorRefreshWindowInMillisec;
     private final int activeExecutorRefreshWindowInFlows;
 
     private volatile boolean shutdown = false;
@@ -1748,7 +1748,7 @@ public class ExecutorManager extends EventHandler implements
       this.maxDispatchingErrors = maxDispatchingErrors;
       this.activeExecutorRefreshWindowInFlows =
           activeExecutorRefreshWindowInFlows;
-      this.activeExecutorRefreshWindowInMilisec =
+      this.activeExecutorRefreshWindowInMillisec =
           activeExecutorRefreshWindowInTime;
       this.setName("AzkabanWebServer-QueueProcessor-Thread");
     }
@@ -1775,7 +1775,7 @@ public class ExecutorManager extends EventHandler implements
           try {
             // start processing queue if active, other wait for sometime
             if (this.isActive) {
-              processQueuedFlows(this.activeExecutorRefreshWindowInMilisec,
+              processQueuedFlows(this.activeExecutorRefreshWindowInMillisec,
                   this.activeExecutorRefreshWindowInFlows);
             }
             wait(QUEUE_PROCESSOR_WAIT_IN_MS);
--- a/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
@@ -170,9 +170,6 @@ public class ExecutionFlowDaoTest {
     final List<ExecutableFlow> flows = this.executionFlowDao
         .fetchRecentlyFinishedFlows(RECENTLY_FINISHED_LIFETIME);
     assertThat(flows.size()).isEqualTo(0);
-
-    //Restore the clock
-    DateTimeUtils.setCurrentMillisOffset(0);
   }
 
   @Test
@@ -358,6 +355,14 @@ public class ExecutionFlowDaoTest {
     assertThat(flow1.getExecutionOptions().getFailureAction())
         .isEqualTo(flow2.getExecutionOptions().getFailureAction());
     assertThat(new HashSet<>(flow1.getEndNodes())).isEqualTo(new HashSet<>(flow2.getEndNodes()));
+  }
+
+  /**
+   * restores the clock; see {@link #testFetchEmptyRecentlyFinishedFlows()}
+   */
+  @After
+  public void clockReset() {
+    DateTimeUtils.setCurrentMillisOffset(0);
   }
 
 }
--- a/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
@@ -170,6 +170,9 @@ public class ExecutionFlowDaoTest {
     final List<ExecutableFlow> flows = this.executionFlowDao
         .fetchRecentlyFinishedFlows(RECENTLY_FINISHED_LIFETIME);
     assertThat(flows.size()).isEqualTo(0);
+
+    //Restore the clock
+    DateTimeUtils.setCurrentMillisOffset(0);
   }
 
   @Test
--- a/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
@@ -80,13 +80,18 @@ public class MetricManagerTest {
    */
   @Test
   public void managerEmitterHandlingTest() throws Exception {
+
+    // metrics use System.currentTimeMillis, so that method should be the millis provider
+    final DateTime aboutNow = new DateTime(System.currentTimeMillis());
+
     this.emitter.purgeAllData();
-    final Date from = DateTime.now().minusMinutes(10).toDate();
+
+    final Date from = aboutNow.minusMinutes(1).toDate();
     this.metric.notifyManager();
 
     this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
 
-    final Date to = DateTime.now().plusMinutes(10).toDate();
+    final Date to = aboutNow.plusMinutes(1).toDate();
     final List<InMemoryHistoryNode> nodes = this.emitter.getMetrics("FakeMetric", from, to, false);
 
     assertEquals("Failed to report metric", 1, nodes.size());
--- a/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
@@ -81,12 +81,12 @@ public class MetricManagerTest {
   @Test
   public void managerEmitterHandlingTest() throws Exception {
     this.emitter.purgeAllData();
-    final Date from = DateTime.now().minusMinutes(1).toDate();
+    final Date from = DateTime.now().minusMinutes(10).toDate();
     this.metric.notifyManager();
 
     this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
 
-    final Date to = DateTime.now().plusMinutes(1).toDate();
+    final Date to = DateTime.now().plusMinutes(10).toDate();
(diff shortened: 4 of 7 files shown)


azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
```diff
@@ -81,12 +81,12 @@ public void managerEmitterMaintenanceTest() {
...
   public void managerEmitterHandlingTest() throws Exception {
     this.emitter.purgeAllData();
-    final Date from = DateTime.now().minusMinutes(1).toDate();
+    final Date from = DateTime.now().minusMinutes(10).toDate();
     this.metric.notifyManager();
 
     this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
 
-    final Date to = DateTime.now().plusMinutes(1).toDate();
+    final Date to = DateTime.now().plusMinutes(10).toDate();
```
juhoautio: Do you know somehow that this was the culprit? Even 1 minute is a lot, so I would expect the bug to lie somewhere else.

juhoautio:
See also https://github.com/azkaban/azkaban/issues/1725

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
mtrna: Sure, that is an interesting question. :) I posed that to myself and the result you can see in the updated PR. :)

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -170,6 +170,9 @@ public void testFetchEmptyRecentlyFinishedFlows() throws Exception {
     final List<ExecutableFlow> flows = this.executionFlowDao
         .fetchRecentlyFinishedFlows(RECENTLY_FINISHED_LIFETIME);
     assertThat(flows.size()).isEqualTo(0);
+
+    //Restore the clock
+    DateTimeUtils.setCurrentMillisOffset(0);
```
juhoautio: Oh, so this test leaks DateTimeUtils offset? It would be safer to place the reset in an `@After` method to not cause side effects even if the test case fails.

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
```diff
@@ -80,13 +80,18 @@ public void managerEmitterMaintenanceTest() {
...
   @Test
   public void managerEmitterHandlingTest() throws Exception {
+
+    // metrics use System.currentTimeMillis, so that method should be the millis provider
+    final DateTime aboutNow = new DateTime(System.currentTimeMillis());
+
     this.emitter.purgeAllData();
-    final Date from = DateTime.now().minusMinutes(1).toDate();
+
+    final Date from = aboutNow.minusMinutes(1).toDate();
```
juhoautio: It should be ok to use the "simple" `DateTime.now()` as long DateTimeUtils is not in some bad state?

azkaban-common/src/test/java/azkaban/project/DirectoryFlowLoaderTest.java
```diff
@@ -95,11 +96,14 @@ private static File decompressTarBZ2(InputStream is) throws IOException {
           parent.mkdirs();
         }
 
-        IOUtils.copy(tais, new FileOutputStream(outputFile));
+        FileOutputStream os = new FileOutputStream(outputFile);
+        IOUtils.copy(tais, os);
+        os.close();
```
juhoautio: Should be in a separate commit, maybe even PR of its own. If this change even matters..

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
juhoautio: I tried this – this makes the test fail consistently:

```java
  @Test
  public void managerEmitterHandlingTest() throws Exception {
    // ADDING THESE TWO LINES:
    final Duration FLOW_FINISHED_TIME = Duration.ofMinutes(2);
    DateTimeUtils.setCurrentMillisOffset(-FLOW_FINISHED_TIME.toMillis());

    this.emitter.purgeAllData();
    final Date from = DateTime.now().minusMinutes(1).toDate();
    this.metric.notifyManager();

    this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
    assertEquals(0, this.emitterWrapper.countDownLatch.getCount());

    final Date to = DateTime.now().plusMinutes(1).toDate();
    final List<InMemoryHistoryNode> nodes = this.emitter.getMetrics("FakeMetric", from, to, false);

    assertEquals("Failed to report metric", 1, nodes.size());
    assertEquals("Failed to report metric", nodes.get(0).getValue(), 4);
  }
```

If I add this after the two added lines it works again:

```java
    DateTimeUtils.setCurrentMillisSystem();
```

My suggestion for the eventual fix would be to only modify `ExecutionFlowDaoTest.java` by adding this into its existing `@After` method:

```java
    DateTimeUtils.setCurrentMillisSystem();
```

I found one more leaking test which is `BasicTimeCheckerTest` – it should also reset DateTimeUtils in an `@After` method.

juhoautio:
Great find @mtrna !

juhoautio:
Btw, how did you find this? Were you able to somehow run the test locally so that it failed consistently (so that you could debug what the actual timestamps are)?

azkaban-common/src/test/java/azkaban/project/DirectoryFlowLoaderTest.java
mtrna: Hi @juhoautio ! It matters, since it allows to release the file descriptors. `IOException` thrown upon the exhaustion of file descriptors is the reason why this test frequently fails.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
mtrna: fair point, I will update the PR

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
mtrna: `DateTime.now()` takes the milliseconds from the effective joda millis provider [1]. Since metrics use the system millis provider directly [citation needed :D], I suggest the `MetricManagerTest` use the system millis provider as well.

[1] https://github.com/JodaOrg/joda-time/blob/master/src/main/java/org/joda/time/DateTimeUtils.java#L112

azkaban-common/src/test/java/azkaban/project/DirectoryFlowLoaderTest.java
mtrna: I agree, will update the PR later today

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
mtrna: joda `DateTime` and `minus/plusMinutes` are used, though! :)

mtrna:
updated, rebased; PR is ready for another round of review

HappyRay:
I haven't had time to look at this PR in details yet. 
@mtrna I appreciate your contribution to make our product better, based on your description!

I am glad to see the community-driven reviews. 
Thanks @juhoautio !

juhoautio:
How about my suggestion on fixing this also in BasicTimeCheckerTest?

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
juhoautio: Yeah it just adds one line of code more. To me there’s no difference if using joda time directly because underneath it should also be calling System#currentTimeMillis. Also it’s safer to get the current time again _after_ the emitter has been called. This shouldn’t really happen, but if the metric call would take more that one minute this could fail because the to date is not after that.

mtrna:
added a fix of the `BasicTimeCheckerTest`

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -357,4 +357,12 @@ private void assertTwoFlowSame(final ExecutableFlow flow1, final ExecutableFlow
     assertThat(new HashSet<>(flow1.getEndNodes())).isEqualTo(new HashSet<>(flow2.getEndNodes()));
   }
 
+  /**
+   * restores the clock; see {@link #testFetchEmptyRecentlyFinishedFlows()}
+   */
+  @After
+  public void clockReset() {
+    DateTimeUtils.setCurrentMillisOffset(0);
```
juhoautio: I checked DateTimeUtils#setCurrentMillisSystem – calling setCurrentMillisOffset(0) is effectively the same. So why not use the simpler method that doesn't require an arg?

azkaban-common/src/test/java/azkaban/metric/MetricManagerTest.java
```diff
@@ -80,13 +80,18 @@ public void managerEmitterMaintenanceTest() {
...
     this.metric.notifyManager();
 
     this.emitterWrapper.countDownLatch.await(10L, TimeUnit.SECONDS);
 
-    final Date to = DateTime.now().plusMinutes(1).toDate();
+    final Date to = aboutNow.plusMinutes(1).toDate();
```
juhoautio: Should get the current time again here instead of using `aboutNow` from before the `notifyManager()` call.

juhoautio:
Left 2 more comments, otherwise 👍 from my perspective.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
mtrna: reuse of existing code; but sure we can replace it as well :)

.gitignore
```diff
@@ -26,3 +26,5 @@ temp/
 .shelf
 # Intellij default build output directory
 out/
+
+currentpid
```
HappyRay: A new line is needed at the end.

Did you follow the contributor guide and set up Save actions plugin?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1734,7 +1734,7 @@ private void cleanExecutionLogs() {
 
     private static final long QUEUE_PROCESSOR_WAIT_IN_MS = 1000;
     private final int maxDispatchingErrors;
-    private final long activeExecutorRefreshWindowInMilisec;
+    private final long activeExecutorRefreshWindowInMillisec;
```
HappyRay: Thanks for fixing typos. 
In the future, I would suggest separate PRs for independent changes.

juhoautio:
> @juhoautio do you think this change is going to fix the recent unit test in-stability reported by multiple users?

Yes.

> I trust @juhoautio's judgment. If he is ok with it, chances are the change is fine. :)

👍 from me for this PR in its current state.

mtrna:
resolved conflicts with master

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
kunkun-tang: In the class `DefaultMailCreatorTest`,

https://github.com/azkaban/azkaban/blob/c75ea9beaea61ef3d92a2948b0d16834e078fb1f/azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java#L53

Should we also change DataTimeUtils back when that test finishes?

juhoautio: No need to change it, it's already done properly:
https://github.com/azkaban/azkaban/blob/c75ea9beaea61ef3d92a2948b0d16834e078fb1f/azkaban-common/src/test/java/azkaban/executor/mail/DefaultMailCreatorTest.java#L84

kunkun-tang: Thanks for checking!

Issue #1618, opened 2018-01-26
Issue with downloading zip from project


ggalves:
Any logged user can download zip from projects with a request to `/manager?project=${project.name}&download=true`, even if they don't have any permission to the project

HappyRay:
Thanks for reporting this issue!
Someone will follow up. 
Are you interested in contributing a fix?

ggalves:
Sorry for the late to respond. I created a PR to fix it, can you review?

Thanks!

HappyRay:
@ggalves 
Thanks!

@reallocf  you are secondary on call this week. Could you take a look?

PR #1737, opened 2018-04-17
Download permission fix and NullPointerException

ggalves:
Bug fix to #1618 

Any logged user could download a zip from project by a request to `host/manager?project=projectName&download=true`, so this PR creates a check for READ permission to download.

Additionally, create a fix to a NullPointerException in `hasPermission` method of `LoginAbstractAzkabanServlet` class: role could be null and that creates a HTTP500 response.

#### NullPointerException
Having `azkaban-users.xml` like this one 
```
<azkaban-users>

    <role name="admin" permissions="ADMIN" />
    <role name="metrics" permissions="METRICS"/>

	<group name="administrators" roles="admin" />
	<group name="guests" roles="metrics" />

    <user username="adm" password="adm" groups="administrators" />
    <user username="guest" password="guest" roles="fakeRole"/>

</azkaban-users>
```

`guest` user has a role that is not defined. This cause `final Role role = userManager.getRole(roleName);` to be null, and accessing a project that does exist but `guest` does not have permission creates a NullPointerException.

`host/manager?project=projectName`

screen shot 2018-04-17 at 11 36 37 am

#### Downloading a project

Accessing `host/manager?project=projectName&download=true`, any logged user could download the project with no permission

#### NullPointerException

With the same `azkaban-users.xml`, the result of accessing a project that `guest` does not have permission:

`host/manager?project=projectName`

screen shot 2018-04-17 at 11 20 45 am

#### Downloading a project

`guest` has no permission to `READ` project `blabla`.

`host/manager?project=blabla&download=true`

screen shot 2018-04-17 at 11 21 08 am

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
@@ -388,6 +388,11 @@ public abstract class LoginAbstractAzkabanServlet extends
 
     for (final String roleName : user.getRoles()) {
       final Role role = userManager.getRole(roleName);
+
+      if (role == null) {
+          continue;
+      }
+
       if (role.getPermission().isPermissionSet(type)
           || role.getPermission().isPermissionSet(Permission.Type.ADMIN)) {
         return true;
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -465,6 +465,13 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       return;
     }
 
+    if (!hasPermission(project, user, Type.READ)) {
+      this.setErrorMessageInCookie(resp, "No permission to download project " + projectName
+          + ".");
+      resp.sendRedirect(req.getContextPath());
+      return;
+    }
+
     int version = -1;
     if (hasParam(req, "version")) {
       version = getIntParam(req, "version");


azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -388,6 +388,11 @@ protected boolean hasPermission(final Project project, final User user,
 
     for (final String roleName : user.getRoles()) {
       final Role role = userManager.getRole(roleName);
+
+      if (role == null) {
+          continue;
```
reallocf: What are the consequences of this? Are users then able to access the resource or do they not have permission to access any resources?

ggalves: You mean the role being null? 
This check just avoids a NullPointerException that will be throwed if code gets to the next if condition, as it tries to call role.getPermission(), so i'm skipping to the next iteration of for loop if role == null
In pratice, the user would not have access to the page, but would receive an HTTP500 response, instead of a treated error

reallocf: Should we throw a 403 Forbidden error instead?

Hmm.. I'm wondering if there's any way to bubble up this error sooner. Like, when the user tries to login, it fails because of an improper role. I'd like to see this fail more quickly instead of when a user tries to access a resource. Maybe we can make a similar request at login time to confirm user role and, if it fails, stop login and throw an error?

Have any thoughts along those lines? What do you think @HappyRay @kunkun-tang ?

reallocf:
Hey @ggalves I've tested your changes locally. Could you possibly split this into two PRs? The `Download permission fix` seems really important and I've tested it locally and it looks good, so we'd like to merge that.

The second fix involving the NullPointerException I think requires more thought. I opened a new issue to track: #1740 . Once we have resolution on how to handle these improperly defined users, maybe you can submit a second PR?

Thanks!

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
ggalves: I agree, this should fail before. I just did this way as `getRole` method was documented "Returns the user role. This may return null.", so i thought that would be nice to make this check.

ggalves:
@reallocf Sure, I'm splitting into two PRs. Thanks!

ggalves:
@reallocf I've opened another PR #1741 with the download fix. I'm leaving the other fix for the issue you have opened

I'm closing this PR. Thanks again!

PR #1742, opened 2018-04-18
Load global.properties even if common.properties doesn't exist

antey86:
Use **global.properties** even if **common.properties** doesn't exist

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -123,8 +123,8 @@ public class JobTypeManager {
       }
     } else {
       logger.info("Common plugin job props file " + commonJobPropsFile
-          + " not found. Using empty props.");
-      commonPluginJobProps = new Props();
+          + " not found. Using only globals props");
+      commonPluginJobProps = new Props(this.globalProperties);
     }
 
     // Loads the common properties used by all plugins when loading


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -124,7 +124,7 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
     } else {
       logger.info("Common plugin job props file " + commonJobPropsFile
           + " not found. Using empty props.");
-      commonPluginJobProps = new Props();
+      commonPluginJobProps = new Props(globalProperties);
```
reallocf: Makes sense to me! Can you call it `this.globalProperties` though so it's in line with our style guide? After that I will ship it + merge 👍 

Thanks for the contribution!

antey86:
Changed! )

Thanks for the review!

Could you please check one more PR and write your notes?
https://github.com/azkaban/azkaban/pull/1727

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -124,7 +124,7 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
     } else {
       logger.info("Common plugin job props file " + commonJobPropsFile
           + " not found. Using empty props.");
```
reallocf: Actually, one last thing. Can you have this comment say `Using only globals props.` instead of `Using empty props.`?

That will make this logger message line up with the new logic. Thanks! :)

PR #1745, opened 2018-04-21
restrict second level schedule for flow trigger

chengren311:
Azkaban won't support schedule like:
run every second
run every 2 second
run every minute at 2nd second.

The check is done on project uploading.


azkaban-common/src/main/java/azkaban/project/NodeBeanLoader.java
```diff
@@ -107,6 +107,13 @@ private void validateSchedule(final FlowTriggerBean flowTriggerBean) {
                 .isValidExpression(scheduleMap.get(FlowTriggerProps.SCHEDULE_VALUE)),
             "flow trigger schedule value must be a valid cron expression");
 
+    final String cronExpression = scheduleMap.get(FlowTriggerProps.SCHEDULE_VALUE).trim();
+    final String[] cronParts = cronExpression.split("\\s+");
+
+    Preconditions
```
kunkun-tang: Do you have any sort of integration test suite? How would it affect the evaluation time length after this change?

chengren311: the unit tests covered the illegal case. Time length is trivial since it's just regular expression matching cost for a string.

PR #1746, opened 2018-04-27
clean up old trigger executions from db table

chengren311:
This PR adds purging mechanism of old flow trigger execution records from the db table. Otherwise the table will keep growing indefinitely as more triggers are executed, leading to excessive query time on the table - e.x recovering incomplete trigger instances or querying historic trigger instances

The cleanup policy is removing trigger instances finishing older than 20 days back, and will be practiced every 10 mins.


kunkun-tang:
Why does it need to clean up data triggers short, like every 10 mins? Any specific reasons? Is the database operation expensive?

chengren311:
@kunkun-tang yes, since it requires the table scan, so it's always good to keep the table size smaller to limit the scan query time.

PR #1749, opened 2018-05-01
more informative email alerting on flow trigger cancellation

chengren311:
This PR added more information to alerting email sent on flow trigger cancellation.

Email content includes - flow trigger instance start/end time, flow trigger instance page link and cancelled dependency instances and the corresponding cancel reason.


juhoautio:
How about doing refactoring separately first:
https://github.com/azkaban/azkaban/pull/1665

HappyRay:
+1 to @juhoautio's suggestion.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -158,5 +170,4 @@ private String getJobOrFlowName(final SlaOption slaOption) {
       return flowName;
     }
   }
-
-}
+}
```
juhoautio: Why?

azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java
```diff
@@ -88,22 +91,58 @@ private void executeFlowAndUpdateExecID(final TriggerInstance triggerInst) {
...
+    message.println("<tr><td>" + sdf.format(new Date(triggerInst.getEndTime())) + "</td><td>");
+    message.println("<tr><td>Duration</td><td>"
+        + Utils.formatDuration(triggerInst.getStartTime(), triggerInst.getEndTime())
+        + "</td></tr>");
+    message.println("<tr><td>Status</td><td>" + triggerInst.getStatus() + "</td></tr>");
+    message.println("</table>");
+    message.println("");
+    final String executionUrl =
+        this.emailer.getScheme() + "://" + this.emailer.getClientHostname() + ":" +
+            this.emailer.getClientPortNumber() + "/" + "executor?" + "triggerinstanceid="
```
juhoautio: Would it be better to refactor this to `this.emailer.getAzkabanUrl() + "/executor?triggerinstanceid="`? `getAzkabanUrl()` could then be used also in Emailer itself, and wouldn't need to expose all those fields.

juhoautio:
> @juhoautio rebased

Thanks for letting my PR be merged first.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
chengren311: why for which part?

azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java
chengren311: good point.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
juhoautio: I mean why removing the newline at the end of the file? Github even adds a warning about it.

juhoautio:
Aha, one refactoring idea after this one's been merged: change all [MailCreator.java](https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/executor/mail/MailCreator.java#L24-L34) interface methods to pass a ready-baked `String azkabanUrl` instead of all three: `String scheme, String clientHostname, String clientPortNumber`. The MailCreator implementations always just build the URL from them any way.

azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java
```diff
@@ -88,22 +91,56 @@ private void executeFlowAndUpdateExecID(final TriggerInstance triggerInst) {
...
+
+    message.println("<tr><td>End Time</td><td>");
+    message.println("<tr><td>" + sdf.format(new Date(triggerInst.getEndTime())) + "</td><td>");
+    message.println("<tr><td>Duration</td><td>"
+        + Utils.formatDuration(triggerInst.getStartTime(), triggerInst.getEndTime())
+        + "</td></tr>");
+    message.println("<tr><td>Status</td><td>" + triggerInst.getStatus() + "</td></tr>");
+    message.println("</table>");
+    message.println("");
+    final String executionUrl = this.emailer.getAzkabanURL()+ "/executor?triggerinstanceid="
```
juhoautio: Missing space between `getAzkabanURL()` and `+`. How could that be? Save Actions plugin not in use?

azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java
```diff
@@ -88,22 +91,56 @@ private void executeFlowAndUpdateExecID(final TriggerInstance triggerInst) {
...
+    return message;
   }
 
   private void sendFailureEmailIfConfigured(final TriggerInstance triggerInstance) {
     final List<String> failureEmails = triggerInstance.getFailureEmails();
     if (!failureEmails.isEmpty()) {
-      this.emailer.sendEmail(failureEmails, generateFailureEmailSubject(triggerInstance),
-          generateFailureEmailBody(triggerInstance));
+      final EmailMessage message = this.createFlowTriggerEmailMessage(triggerInstance);
+      this.emailer.sendEmail(message, true, "email message " + message.getBody());
```
juhoautio: Isn't `operation = "email message " + message.getBody()` a bit too much to be included in info-level logging (see Emailer.java)? I'd suggest for example `"flow trigger failure email"` instead (result => `logger.info("Sent " + "flow trigger failure email")`). Any way if you're including the entire email body content on purpose, fine by me..

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -86,7 +91,7 @@ public void testProcessTermination() throws ExecutorManagerException, Interrupte
     final TriggerInstance triggerInstance = createTriggerInstance();
     processor.processTermination(triggerInstance);
     sleep(1000);
-    verify(emailer).sendEmail(any(), any(), any());
+    verify(emailer).sendEmail(any(), anyBoolean(), anyString());
```
juhoautio: How about using this strategy to check the output:
https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/test/java/azkaban/utils/EmailerTest.java#L97-L98
?

When you have the expected output saved as a .html file, you can easily render it in browser to check that it looks as expected, like this one:
https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/test/resources/azkaban/executor/mail/errorEmail2.html

azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java
```diff
@@ -88,22 +91,56 @@ private void executeFlowAndUpdateExecID(final TriggerInstance triggerInst) {
...
-    return String.format(FAILURE_EMAIL_SUBJECT, flowFullName, this.emailer.getAzkabanName());
+    return String.format(FAILURE_EMAIL_SUBJECT, triggerInstance.getFlowId(), triggerInstance
+        .getProjectName(), this.emailer.getAzkabanName());
   }
 
-  private String generateFailureEmailBody(final TriggerInstance triggerInstance) {
-    final String triggerInstFullName =
-        triggerInstance.getProjectName() + "." + triggerInstance.getFlowId();
-    return String.format(FAILURE_EMAIL_BODY, triggerInstFullName);
+  private EmailMessage createFlowTriggerEmailMessage(final TriggerInstance triggerInst) {
```
juhoautio: Rename to `createFailureEmailMessage`? To me it was confusing to find out that this method is actually calling `generateFailureEmailSubject`, so it's not just a "generic notification message" but specifically about a failure case.

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -82,11 +106,22 @@ public void testProcessSucceed() throws ExecutorManagerException {
...
     final TriggerInstance triggerInstance = createTriggerInstance();
     processor.processTermination(triggerInstance);
     sleep(1000);
-    verify(emailer).sendEmail(any(), any(), any());
+    InputStream is = null;
+    try {
+      is = TriggerInstanceProcessorTest.class.getClassLoader()
+          .getResourceAsStream("emailTemplate/flowtriggerfailureemail.html");
+      final String emailHtml = IOUtils.toString(is, Charsets.UTF_8).trim();
+      assertThat(emailHtml).isEqualToIgnoringWhitespace(message.getBody());
```
juhoautio: There'a already this in DefaultMailCreatorTest:

```java
  public static String read(final String file) throws Exception {
    final InputStream is = DefaultMailCreatorTest.class.getResourceAsStream(file);
    return IOUtils.toString(is, Charsets.UTF_8).trim();
  }
```

Hmm, `IOUtils.toString` doesn't close the stream? Then it should be changed to use try-with-resources to take care of closing the InputStream:

```java
  public static String read(final String file) throws Exception {
    try (final InputStream is = TriggerInstanceProcessorTest.class.getResourceAsStream(file)) {
      return IOUtils.toString(is, Charsets.UTF_8).trim();
    }
  }
```

Why not add a similar method in TriggerInstanceProcessorTest? Or better yet, combine them into some generic utility method that takes the parent class as an argument?

Any way, then the 10 lines of assertion code here would reduce to 2:

```java
    assertThat(TriggerInstanceProcessorTest.read("emailTemplate/flowtriggerfailureemail.html"))
        .isEqualToIgnoringWhitespace(this.message.getBody());
```

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -82,11 +106,22 @@ public void testProcessSucceed() throws ExecutorManagerException {
   }
 
   @Test
-  public void testProcessTermination() throws ExecutorManagerException, InterruptedException {
+  public void testProcessTermination()
+      throws ExecutorManagerException, InterruptedException, IOException {
     final TriggerInstance triggerInstance = createTriggerInstance();
     processor.processTermination(triggerInstance);
     sleep(1000);
```
juhoautio: Not in the scope of this PR, but sleep should be replaced with some dynamic wait to make unit tests run fast.

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -47,28 +58,41 @@
...
         new ArrayList<>(),
         Duration.ofMinutes(1)
     );
     final Project proj = new Project(1, "proj");
-    final Flow flow = new Flow("flowId");
+    final Flow flow = new Flow("123");
     flow.addFailureEmails(Lists.newArrayList(EMAIL));
     proj.setFlows(Maps.newHashMap("flowId", flow));
     final List<DependencyInstance> depInstList = new ArrayList<>();
+    depInstList.add(
```
juhoautio: Could use Arrays.asList(varargs) instead of new + add? Just slightly cleaner..

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
chengren311: thanks

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -47,28 +58,41 @@
...
     executorManager = mock(ExecutorManager.class);
     when(executorManager.submitExecutableFlow(any(), anyString())).thenReturn("return");
-    emailer = mock(Emailer.class);
-    doNothing().when(emailer).sendEmail(any(), any(), any());
+    final CommonMetrics commonMetrics = new CommonMetrics(new MetricsManager(new MetricRegistry()));
+    emailer = new Emailer(new Props(), commonMetrics, messageCreator);
     processor = new TriggerInstanceProcessor(executorManager, triggerInstLoader, emailer);
   }
 
+
```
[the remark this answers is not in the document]
chengren311: updated

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
chengren311: updated, thanks

azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
chengren311: updated, thanks

chengren311:
@juhoautio let me know if you have more suggestions.  thanks!

PR #1752, opened 2018-05-05
Test cleanup: Close InputStream & refactor

juhoautio:
- Close InputStream by using try-with-resources
- Better isolation as EmailerTest depends on TestUtils instead of DefaultMailCreatorTest


azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -77,4 +80,10 @@ public static UserManager createTestXmlUserManager() {
     return manager;
   }
 
+  public static String read(final String file, final Object parent) throws Exception {
```
chengren311: readResouceFile to be more descriptive?

juhoautio: It would be more descriptive.. But, I wanted to keep this short to not bloat too much the test code that use this.

IMHO "read" is clear enough when you look at the tests that are calling this. I'd tend to think that a more descriptive name would be worth the added verbosity if this would be more like a public interface (where you for example let your IDE's auto-complete show what methods there are to pick on this class).

But it's your call, obviously. Do you want it renamed?

chengren311: Thanks for sharing your thoughts. but it would be best to rename and letting intellij's refactor to do renaming is pretty much effortless.

juhoautio:
Renamed TestUtils#read to readResource. It's not necessarily only for regular files, so I didn't add "file" to the method name.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -77,4 +80,10 @@ public static UserManager createTestXmlUserManager() {
     return manager;
   }
 
+  public static String readResource(final String name, final Object parent) throws Exception {
+    try (final InputStream is = parent.getClass().getResourceAsStream(name)) {
```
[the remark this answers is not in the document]
juhoautio: Using the parent arg is there so that you don't need to prefix the resource name with the test class' package name. In my opinion it would be a bit messy to offer the same with Guava utils.

If there's a case where the resource path doesn't begin with the test class' package name, then I would use Guava. Personally I kinda like matching the package structure in resource paths though, so if it's up to me that wouldn't be too common..

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -77,4 +80,10 @@ public static UserManager createTestXmlUserManager() {
     return manager;
   }
 
+  public static String readResource(final String name, final Object parent) throws Exception {
```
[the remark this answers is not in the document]
juhoautio: Will do if this approach is otherwise ok'd?

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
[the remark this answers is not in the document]
HappyRay: I see.

I can see some advantage of organizing the resources based on the test class's package name. However, this strategy doesn't work if a resource is used by multiple test classes in different packages. Thus, I don't think this organization strategy can be applied consistently. IMHO, allowing the resources to be organized independently seems to offer better flexibility without sacrificing readability if named appropriately. 

That said I acknowledge that this is a departure from some of the existing code. I will leave the decision to you as the author of this code. 

-----

If you choose the package name based organization, have you considered

public static URL getResource(Class<?> contextClass,
                              String resourceName)
Given a resourceName that is relative to contextClass, returns a URL pointing to the named resource.

https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/Resources.html#getResource-java.lang.Class-java.lang.String-

The parent parameter's type is weaker than I would like to see. The API above makes a different choice. I am interested in your opinion comparing these two different designs.

Using the Gauva APIs referenced above seems to simplify the implementation slightly IMHO.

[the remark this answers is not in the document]
juhoautio: Ok, I have been personally a bit disappointed with the Guava Resources utils, because they pretty much always have failed to give me a one-liner solution. If the solution requires calling multiple Guava methods, I rather pick calling the same amount of JDK's own methods.

The reason for choosing `Object` arg over `Class<?>` was to have less verbosity in the test cases. This is also about personal taste: I like to keep unit tests readable in a way that you can easily see what the different test cases are _about_, not how different things about the test are implemented in practical detail. You can always look it up by jumping into methods, but I think it's valuable to have the test cases structured so that the _code_ reveals at a glance what each test method is essentially about. This way there's no need to add comments/javadocs (that are prone to get outdated) to explain what each test is trying to check.

> The parent parameter's type is weaker than I would like to see. The API above makes a different choice.

Yes. On general level you're absolutely right and I'm hesitating on this one a bit. I just tend to think that on the test side avoiding additional verbosity is really nice, and that it's enough if the code seems unambiguous when looking at it on the test case. It doesn't need to be that strongly typed or even easily understandable as a general-purpose API (although that's of course always a plus).

In this case I think there is no problem in understanding what for example this does: 

```java
TestUtils.readResource("firstErrorMessage.html", this)
```

So, to have ~stronger type, it could be:

```java
TestUtils.readResource("firstErrorMessage.html", DefaultMailCreatorTest.class)
```

This is more verbose, and requires one more context-switch for the reader to check if `DefaultMailCreatorTest` is indeed this or some other class.

Because of that I would rather do this:

```java
TestUtils.readResource("firstErrorMessage.html", this.getClass())
```
..but that also adds complexity by one more method call.

When you just see `this` you pretty much automatically ignore it and move on, making it faster to read the code.

I of course realize that my proposed utility method won't work if multiple tests are using the same input files. But why would they? If tests are pure unit tests this should be at least rare. Any way if that ever happens, adding another utility method for that purpose would be ok.

Bottom line, I haven't formed any strong opinion on this, so if you do, just let me know and I'll make those edits. I feel kinda bad for writing such a long post about such a marginal issue.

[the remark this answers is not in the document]
HappyRay: I found this discussion useful.

---

>  "I just tend to think that on the test side avoiding additional verbosity is really nice"

I agree. I would create wrapper methods in tests for that purpose.
In this case, what do you think about having a private method  getTestString("firstErrorMessage.html")? 

The shared utility method can be more flexible. 

I would let you decide.

azkaban-common/src/test/java/azkaban/utils/TestUtils.java
```diff
@@ -77,4 +81,19 @@ public static UserManager createTestXmlUserManager() {
...
 
+  /**
+   * Reads a resource into a String
+   *
+   * @param name Relative path to the resource (relative to the parent object's package)
+   * @param parent Instance of the class to use in finding the resource. The resource is looked up in the same package
+   * where the class of the parent object is in.
+   * @return Resource content as a String
+   * @throws IOException if an I/O error occurs
+   */
```
juhoautio: Added this javadoc ^

Is it ok?

chengren311: LGTM. @HappyRay WDYT?

HappyRay: It's not my preferred choice. But I am ok with it.

juhoautio:
@chengren311 anything else?

juhoautio:
Thank you @chengren311 & @HappyRay!

HappyRay:
@juhoautio Thank you for your continued support.

PR #1753, opened 2018-05-05
Remove sleep in TriggerInstanceProcessorTest

juhoautio:
- Don't use static fields (to have isolated mocks for each `@Test` method)
- Apply save actions
- Replace static sleep method with async callback to avoid extra wait


chengren311:
@HappyRay i think my intellij setup was somehow messed. I reenabled the save plugin.
@juhoautio i will review it after merging https://github.com/azkaban/azkaban/pull/1749

PR #1755, opened 2018-05-08
New DAG engine initial checkin

HappyRay:
Initial DAG engine itself

#1486

The code is not used yet.

See earlier prototype and change history at #1487

More unit tests and refactoring will follow.

Node state machine diagram

azkaban dag state machine diagram - page 1


azkaban-exec-server/src/main/java/azkaban/dag/DagService.java
```diff
@@ -0,0 +1,108 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Thread safe and non blocking service for DAG processing.
+ *
+ * <p>Allow external inputs to be given to a dag or node to allow the dag to transition states
+ * . Since only one thread is used to progress the DAG, thread synchronization is avoided.
+ */
+@SuppressWarnings("FutureReturnValueIgnored")
```
reallocf: Can you either resolve this warning or leave a comment with why it deserves to be suppressed? (I believe this is still the best practice, right?)

HappyRay: I think this is self-explanatory. The return value is not used.

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
```diff
@@ -0,0 +1,55 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.dag;
+
+import com.google.common.collect.ImmutableSet;
+
+enum Status {
```
chengren311: should we have two separated status enum for dag and node each?

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
```diff
@@ -0,0 +1,55 @@
...
+  READY, // ready to run
+  DISABLED, // disabled by users. Treat as the node has the status of success
+  BLOCKED, // temporarily blocked. Need to be unblocked by another external event
```
chengren311: any example of node in blocked status?

azkaban-exec-server/src/test/java/azkaban/dag/DagTest.java
```diff
@@ -0,0 +1,164 @@
...
+ * Tests the dag state ( including its nodes' states) transitions.
+ *
+ * Focuses on how the dag state changes in response to one external request.
+ */
+public class DagTest {
+
+  private final Dag testFlow = new Dag("fa", mock(DagProcessor.class));
+
+  @Test
+  public void dag_finish_with_only_disabled_nodes() {
```
chengren311: why is public test method name separated by _? should we follow a consistent standard and document in the contributor guide?

azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
```diff
@@ -0,0 +1,139 @@
...
+ * A DAG (Directed acyclic graph) consists of {@link Node}s.
+ */
+class Dag {
+
+  private final String name;
+  private final DagProcessor dagProcessor;
+  private final List<Node> nodes = new ArrayList<>();
+  private Status status = Status.READY;
+
+  Dag(final String name, final DagProcessor dagProcessor) {
```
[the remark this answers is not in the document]
HappyRay: I plan to refactor this part later: separate construction of DAG from running a DAG.

I am interested in what you think of the different API choices. We can chat more off line.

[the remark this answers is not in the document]
HappyRay: Good question.

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: Good question.

I thought about it.

The main factor that made me decide to use the same enum is to make testing easier. The same type can be put in the StatusChangeRecorder. 

That said, I can use string in the StatusChangeRecorder. Let's discuss more if it is worth refactoring.

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: Not yet. It's not implemented.

azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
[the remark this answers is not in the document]
HappyRay: Discussed offline, we agree it is ideal to make the DAG as immutable as possible.

azkaban-exec-server/src/main/java/azkaban/dag/Status.java
HappyRay: I will try it.

azkaban-exec-server/src/test/java/azkaban/dag/DagTest.java
HappyRay: Discussed offline, Cheng is fine with this style.

azkaban-exec-server/src/test/java/azkaban/dag/TestDagProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+
+  TestDagProcessor(final CountDownLatch dagFinishedLatch,
+      final StatusChangeRecorder statusChangeRecorder) {
+    this.dagFinishedLatch = dagFinishedLatch;
+    this.statusChangeRecorder = statusChangeRecorder;
+  }
+
+  @Override
+  public void changeStatus(final Dag dag, final Status status) {
+    System.out.println(dag);
```
chengren311: system out

HappyRay: This is intentional. 
Any downside?

HappyRay: Discussed with Cheng offline. 
He agreed that it is ok to use in test code.

PR #1759, opened 2018-05-15
Separate dag creation from dag execution

HappyRay:
Special care is given so that the node and dag instances returned to code outside of this package can't be used to manipulate the internal states of the dags directly. This eliminates a class of program errors caused by changing the dag states while it is running.

The dag creation classes can be used separately in the future. To
detect circular dependencies for example.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
+++ b/azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
@@ -20,27 +20,54 @@ import static java.util.Objects.requireNonNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * A DAG (Directed acyclic graph) consists of {@link Node}s.
+ *
+ * <p>Most of the methods in this class should remain package private. Code outside of this
+ * package should mainly interact with the {@link DagService}.
  */
 class Dag {
 
   private final String name;
   private final DagProcessor dagProcessor;
   private final List<Node> nodes = new ArrayList<>();
+  private final Map<String, Node> nameToNodeMap = new HashMap<>();
   private Status status = Status.READY;
 
   Dag(final String name, final DagProcessor dagProcessor) {
+    requireNonNull(name, "The name of the Dag can't be null");
     this.name = name;
     requireNonNull(dagProcessor, "The dagProcessor parameter can't be null.");
     this.dagProcessor = dagProcessor;
   }
 
+  /**
+   * Adds a node to the current dag.
+   *
+   * <p>It's important NOT to expose this method as public. The design relies on this to ensure
+   * correctness. The DAG's structure shouldn't change after it is created.
+   *
+   * @param node a node to add
+   */
   void addNode(final Node node) {
-    node.setDag(this);
+    assert (node.getDag() == this);
     this.nodes.add(node);
+    assert (!this.nameToNodeMap.containsKey(node.getName()));
+    this.nameToNodeMap.put(node.getName(), node);
+  }
+
+  /**
+   * Gets the node associated with the name.
+   *
+   * @param name node name
+   * @return node. null if the node with this name doesn't exist
+   */
+  Node getNodeByName(final String name) {
+    return this.nameToNodeMap.get(name);
   }
 
   void start() {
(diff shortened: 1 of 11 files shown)


azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
```diff
@@ -20,27 +20,54 @@
...
+    requireNonNull(name, "The name of the Dag can't be null");
     this.name = name;
     requireNonNull(dagProcessor, "The dagProcessor parameter can't be null.");
     this.dagProcessor = dagProcessor;
   }
 
+  /**
+   * Adds a node to the current dag.
+   *
+   * <p>It's important NOT to expose this method as public. The design relies on this to ensure
```
reallocf: What's the consequence of this node having default scope instead of it being specified as `private`? Is it used elsewhere in the package? Is that intended? Because you mention it NOT being public it might be useful to have a comment saying why it IS default scope.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -39,35 +39,34 @@
...
+  private final Dag dag;
 
-  Node(final String name, final NodeProcessor nodeProcessor) {
-    this.name = name;
+  Node(final String name, final NodeProcessor nodeProcessor, final Dag dag) {
     requireNonNull(nodeProcessor, "The nodeProcessor parameter can't be null.");
     this.nodeProcessor = nodeProcessor;
+    requireNonNull(name, "The name of the node can't be null");
+    this.name = name;
+    requireNonNull(name, "The dag of the node can't be null");
```
reallocf: small type - I think `name` is supposed to be `dag`, right?

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
```diff
@@ -39,35 +39,34 @@
...
 
-  public void setDag(final Dag dag) {
-    this.dag = dag;
-  }
-
-  private void addParent(final Node node) {
+  /**
+   * Adds the node as the current node's parent i.e. the current node depends on the given node.
+   *
+   * <p>It's important NOT to expose this method as public. The design relies on this to ensure
```
reallocf: Same question regarding the consequences of this scoping.

azkaban-exec-server/src/main/java/azkaban/dag/Dag.java
HappyRay: This is how package private access level is specified.

https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

The method is needed by other classes in the same package.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: See the answer to the other one.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
HappyRay: Thanks. Will fix.

HappyRay: Fixed in the next patch.

azkaban-exec-server/src/main/java/azkaban/dag/Node.java
reallocf: Sounds good!

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+    requireNonNull(name, "The name of the DagBuilder can't be null");
+    this.name = name;
+    requireNonNull(name, "The dagProcessor of the DagBuilder can't be null");
+    this.dagProcessor = dagProcessor;
+  }
+
+  public NodeBuilder createNode(final String name, final NodeProcessor nodeProcessor) {
+    final NodeBuilder builder = new NodeBuilder(name, nodeProcessor, this);
+    this.builders.add(builder);
+    if (this.nodeNamesSet.contains(name)) {
```
chengren311: should we do existence check before builders.add(builder)?

HappyRay: I don't expect the users of this method to recover from this exception by dynamically changing the node name and re-use the existing builder. Thus, it doesn't make much difference.

That said, I agree it is better to do the check before modifying the builders. 
I will make a change. 

Thanks

HappyRay: see #1808

PR #1760, opened 2018-05-16
Check circular dependency

HappyRay:
Throws an exception when circular dependency is detected when building a
dag.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
+++ b/azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
@@ -21,6 +21,7 @@ import static java.util.Objects.requireNonNull;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -69,13 +70,100 @@ public NodeBuilder createNode(final String name, final NodeProcessor nodeProcess
...
+      // One sample of nodes that form a circular dependency
+      private final List<NodeBuilder> sampleCircularNodes = new ArrayList<>();
+
+      /**
+       * Checks if the builder contains nodes that form a circular dependency ring.
+       *
+       * @throws DagException if true
+       */
+      private void check() {
+        // The algorithm is described in
```
kunkun-tang: Is the comment not finished?

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -69,13 +70,100 @@ public NodeBuilder createNode(final String name, final NodeProcessor nodeProcess
...
+      private final Set<NodeBuilder> toVisit = new HashSet<>(DagBuilder.this.builders);
+
+      // The nodes that have finished traversing all their parent nodes
+      private final Set<NodeBuilder> finished = new HashSet<>();
+
+      // The nodes that are waiting for their parent nodes to finish visit.
+      private final Set<NodeBuilder> ongoing = new HashSet<>();
+
+      // One sample of nodes that form a circular dependency
+      private final List<NodeBuilder> sampleCircularNodes = new ArrayList<>();
```
kunkun-tang: Is sampleCircularNodes referring to circular results given a DAG? Why is it called `sample` ?

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
```diff
@@ -79,6 +79,27 @@ private NodeBuilder createNodeBuilder(final String name) {
...
+    nodeBuilder2.addParents(nodeBuilder1);
+    nodeBuilder3.addParents(nodeBuilder2);
+    nodeBuilder1.addParents(nodeBuilder3);
+
+    // when
+    final Throwable thrown = catchThrowable(() -> {
+      this.dagBuilder.build();
+    });
+
+    System.out.println("Expect exception: " + thrown);
```
kunkun-tang: should we use System.err ?

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
HappyRay: Because there may be others.

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
HappyRay: If true means " if the check returns true" 
I can make it a complete sentence if it is not clear enough.

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
HappyRay: No. This is for debugging purposes. It's not an error.

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
kunkun-tang: What I want to point out is 
>The algorithm is described in

The "where" is not defined.

HappyRay: I see. I forgot to remove this comment. I moved it to the class javadoc above. Will delete.

HappyRay: Fixed.

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
juhoautio: IMHO there's no use leaving test debug output like this in place. If one needs similar thing at some point it's easy to add. Getting rid of extra output keeps the test logs cleaner. Just something for you to consider..

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
```diff
@@ -79,6 +79,27 @@ private NodeBuilder createNodeBuilder(final String name) {
     return this.dagBuilder.createNode(name, mock(NodeProcessor.class));
   }
 
+  @Test
+  public void build_should_throw_exception_when_circular_dependency_is_detected() {
+    // given
```
juhoautio: Maybe add, to help reader:
`// circular dependency like this: nb1 <- nb2 <- nb3 <- nb1`

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -69,13 +70,99 @@ public NodeBuilder createNode(final String name, final NodeProcessor nodeProcess
...
+        if (this.finished.contains(node)) {
+          return false;
+        }
+        if (this.ongoing.contains(node)) {
+          this.sampleCircularNodes.add(node);
+          return true;
+        }
+        this.toVisit.remove(node);
+        this.ongoing.add(node);
+        for (final NodeBuilder parent : node.getParents()) {
```
juhoautio: Is this only the direct dependencies? What about sub-flows, if one creates for example:

x.job:
```
type=flow
flow.name=z
```

z.job:
```
type=flow
flow.name=x
```

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
HappyRay: Gradle writes such output to a log file instead of the console by default. I agree that writing large verbose debug outputs can make searching for failures in the log more difficult. However, most of the test failures can be reproduced locally and it is easy to see which tests fail from console or IDE. 

I left this log in the code to make it easier to see the exception message when running in an IDE.

On balance, I think it is OK to leave some reasonable amount of debugging information in the test output log after considering the tradeoffs.

azkaban-exec-server/src/test/java/azkaban/dag/DagBuilderTest.java
HappyRay: Done.

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
HappyRay: In this example, x and z in the parent dag would form a circle and this code will detect it.

a node can't be added to two different dags. see checkBuildersBelongToSameDag in NodeBuilder.

HappyRay:
@juhoautio Thanks for your review.

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
juhoautio: I couldn't tell.. Maybe because I'm not familiar enough with how the DAG is exploded. But maybe it wouldn't hurt to have explicit unit test for sub-flows, like the two that I mentioned in my comment?

HappyRay: Subflow is a higher level concept and will be implemented at a higher layer. 

The included unit tests can illustrate how this package should be used. see simple_sub_dag_success_case in DagServiceTest for example.

HappyRay: >  "what if there's a circular dependency inside a sub-flow"

Could you give a specific example?

PR #1762, opened 2018-05-17
Add new tests to Dag

HappyRay:
Test that killing a dag in certain state such as killing won't have
any effect.

With this change, the DAG class has 100% code coverage.

Also make ImmutableSet in the Status enum more strongly typed.


HappyRay:
Forgot to create a branch first. Will create a new one.

PR #1763, opened 2018-05-17
Add new tests to Dag

HappyRay:
Test that killing a dag in certain state such as killing won't have
any effect.

With this change, the DAG class has 100% code coverage.

Also make ImmutableSet in the Status enum more strongly typed.


balajinima:
/src/test/java/azkaban/dag/DagTest.java
 -----------------------------------------------
final Node node = TestUtil.createNodeWithNullProcessor(name, this.testFlow);
this.testFlow.addNode(node);
-------------------------------------
add node to testFlow twice。

HappyRay:
Could you elaborate? 
I don't see the code adding the same node twice in this file.

PR #1764, opened 2018-05-17
Test multiple ready nodes are canceled when the parent node failed.

HappyRay:
With this test, the Node class has 100% code coverage.


li-afaris:
The new test looks good. +1

PR #1765, opened 2018-05-21
Refactor DagService shutdown method and add tests

HappyRay:
Part of #1486

Create a new utility class to encapsulate the generic shutdown logic to
make it easier to test and allow better code reuse.

Also reduced the shutdown timeout from 60 seconds to 10 seconds.

With this change and #1763, the dag package will have 100% line
coverage.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/ExecutorServiceUtils.java
+++ b/az-core/src/main/java/azkaban/utils/ExecutorServiceUtils.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2018 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.utils;
+
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Executor service related utilities.
+ */
+public class ExecutorServiceUtils {
+
+  private static final Logger logger = LoggerFactory.getLogger(ExecutorServiceUtils.class);
+  private static final TimeUnit MILLI_SECONDS_TIME_UNIT = TimeUnit.MILLISECONDS;
+
+  /**
+   * Gracefully shuts down the given executor service.
+   *
+   * <p>Adopted from
+   * <a href="https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html">
+   * the Oracle JAVA Documentation.
+   * </a>
+   *
+   * @param service the service to shutdown
+   * @param timeout max wait time for the tasks to shutdown. Note that the max wait time is 2
+   *     times this value due to the two stages shutdown strategy.
+   * @throws InterruptedException if the current thread is interrupted
+   */
+  public void gracefulShutdown(final ExecutorService service, final Duration timeout)
+      throws InterruptedException {
+    service.shutdown(); // Disable new tasks from being submitted
+    final long timeout_in_unit_of_miliseconds = timeout.toMillis();
+    // Wait a while for existing tasks to terminate
+    if (!service.awaitTermination(timeout_in_unit_of_miliseconds, MILLI_SECONDS_TIME_UNIT)) {
+      service.shutdownNow(); // Cancel currently executing tasks
+      // Wait a while for tasks to respond to being cancelled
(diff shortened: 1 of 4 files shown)


az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertThat(service.isShutdown()).isTrue();
+  }
+
+  @Test
+  public void force_shutdown_after_timeout() throws InterruptedException {
+    // given
+    final ExecutorService service = Executors.newSingleThreadExecutor();
+
+    // when
+    service.submit(this::sleep);
```
kunkun-tang: When service submits the sleep job, it will continue running the rest of the code?

az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
```diff
@@ -0,0 +1,85 @@
...
+    final long beginShutdownTime = System.currentTimeMillis();
+    this.executorServiceUtils.gracefulShutdown(service, Duration.ofMillis(1));
+    final long endShutdownTime = System.currentTimeMillis();
+
+    // then
+    assertThat(service.isShutdown()).isTrue();
+    final long shutdownDuration = endShutdownTime - beginShutdownTime;
+    // Give some buffer for overhead to reduce false positives.
+    assertThat(shutdownDuration).isLessThan(10);
```
kunkun-tang: Will there be any random errors if we write this way

az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
HappyRay: Yes, it is a non blocking call.

az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
HappyRay: No. Unless the test machine is super busy. We can adjust the buffer if it happens.

azkaban-exec-server/src/main/java/azkaban/dag/DagService.java
```diff
@@ -81,28 +87,14 @@ void killDag(final Dag dag) {
...
-      // Preserve interrupt status
-      Thread.currentThread().interrupt();
-    }
+  void shutdownAndAwaitTermination() throws InterruptedException {
+    logger.info("DagService is shutting down.");
+    this.executorServiceUtils.gracefulShutdown(this.executorService, SHUTDOWN_WAIT_TIMEOUT);
+  }
+
+  @VisibleForTesting
+  ExecutorService getExecutorService() {
```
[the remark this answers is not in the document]
HappyRay: It's tested indirectly by the tests.

The annotation is just for documentation.

see
/**
 * Annotates a program element that exists, or is more widely visible than otherwise necessary, only
 * for use in test code.
 *
 * @author Johannes Henkel
 */
@GwtCompatible
public @interface VisibleForTesting {
}

azkaban-exec-server/src/main/java/azkaban/dag/DagService.java
```diff
@@ -81,28 +87,14 @@ void killDag(final Dag dag) {
...
-          logger.error("The DagService did not terminate.");
-        }
-      }
-    } catch (final InterruptedException ie) {
-      // (Re-)Cancel if current thread also interrupted
-      this.executorService.shutdownNow();
-      // Preserve interrupt status
-      Thread.currentThread().interrupt();
-    }
+  void shutdownAndAwaitTermination() throws InterruptedException {
```
kunkun-tang: Just one more question. Where will you call the method? I don't see any callers.

HappyRay: Thanks.

The class is not used in production code yet. But you can see example usages in the unit tests.

az-core/src/main/java/azkaban/utils/ExecutorServiceUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Executor service related utilities.
+ */
+public class ExecutorServiceUtils {
```
chengren311: Since it's an util class, why not make gracefulShutdown static? What's the advantage of making this class injectable?

HappyRay: I am glad you asked.

How would you test the code that uses this method if this method is made static?

chengren311: Some mocking lib allows mocking a static method.
https://stackoverflow.com/questions/21105403/mocking-static-methods-with-mockito

HappyRay: I haven't used these libs. Do you think using these mocking libraries is better? If so, why?

chengren311: Discussed offline, static method needs a non-standard way of injecting and test. With dependency injection, we can declare an util class as singleton, which serves the same as an util class containing only static methods.

PR #1766, opened 2018-05-22
Improve tests for Status

HappyRay:
Even though we have 100% line coverage, the logical coverage is more
limited. This change improves that.


HappyRay:
@reallocf 
Thanks for such a quick response!

PR #1767, opened 2018-05-22
Fix flaky test ExecutorServiceUtilsTest

HappyRay:
Failure on Travis:

azkaban.utils.ExecutorServiceUtilsTest > force_shutdown_after_timeout FAILED
    java.lang.AssertionError:
    Expecting:
     <11L>
    to be less than:
     <10L>
        at azkaban.utils.ExecutorServiceUtilsTest.force_shutdown_after_timeout(ExecutorServiceUtilsTest.java:60)
azkaban.utils.ExecutorServiceUtilsTest

Cause:

When the test host is busy, the execution time may be greater.

Fix:
Add more buffer and a comment to explain the test.

### DIFF
--- a/az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
+++ b/az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
@@ -57,8 +57,9 @@ public class ExecutorServiceUtilsTest {
     assertThat(service.isShutdown()).isTrue();
     final long shutdownDuration = endShutdownTime - beginShutdownTime;
     // Give some buffer for overhead to reduce false positives.
-    // 100 is still much smaller than 1000 if the task were not forcefully terminated.
-    assertThat(shutdownDuration).isLessThan(100);
+    // 5 seconds is still much smaller than 30 seconds sleep time if the task were not forcefully
+    // terminated.
+    assertThat(shutdownDuration).isLessThan(5000);
   }
 
   @Test
@@ -80,7 +81,7 @@ public class ExecutorServiceUtilsTest {
 
   private void sleep() {
     try {
-      Thread.sleep(1000);
+      Thread.sleep(30_000);
     } catch (final InterruptedException ex) {
     }
   }
--- a/az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
+++ b/az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
@@ -30,6 +30,8 @@ public class ExecutorServiceUtilsTest {
 
   private final ExecutorServiceUtils executorServiceUtils = new ExecutorServiceUtils();
 
+  private boolean isTestThreadInterrupted = false;
+
   @Test
   public void gracefulShutdown() throws InterruptedException {
     // given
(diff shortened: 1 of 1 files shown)


az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
```diff
@@ -57,7 +57,8 @@ public void force_shutdown_after_timeout() throws InterruptedException {
     assertThat(service.isShutdown()).isTrue();
     final long shutdownDuration = endShutdownTime - beginShutdownTime;
     // Give some buffer for overhead to reduce false positives.
-    assertThat(shutdownDuration).isLessThan(10);
+    // 100 is still much smaller than 1000 if the task were not forcefully terminated.
+    assertThat(shutdownDuration).isLessThan(100);
```
juhoautio: That's milliseconds? 100 would've still been way too low as sleep time to stabilize some flaky unit tests before as Travis builds seem to suffer from random slowdowns. Even 1 second wasn't enough sometimes. If it was up to me, I'd remove this assertion if it can't be set to something that's guaranteed to be high enough, like 10 seconds.

HappyRay: Yes, that's milliseconds. 

I'd like to see how often this operation takes longer than 100ms. The worst I have seen so far took 11ms.

juhoautio: Maybe this operation is then more immune to random slowdowns.. I wonder if Travis can randomly have some longer GC pauses or such that cause unexpected extra delay between operations. But this is just speculation, I didn't ever measure the times on Travis and see how much they vary.

HappyRay: @juhoautio 
I thought more about your suggestion. I think you are right. I increased the threshold.

@kunkun-tang 
The time is how long it took to interrupt the thread that is sleeping. The 11ms was seen on travis.

az-core/src/test/java/azkaban/utils/ExecutorServiceUtilsTest.java
```diff
@@ -80,7 +81,7 @@ public void can_not_submit_tasks_after_shutdown() throws
 
   private void sleep() {
     try {
-      Thread.sleep(1000);
+      Thread.sleep(30_000);
```
kunkun-tang: 30_000 sounds too long here. What if the shutdown method fail? Will it take 30 seconds to fail this particular test? I'd like this combination: Thread.sleep(5000) and islessThan(1000).

If JVM can not interrupt the sleep thread within 1s, there must be something seriously wrong.

HappyRay: I don't think the change would make any difference. If the shutdown method does nothing, the test will still run as quickly as the host allows. 

Thanks for your comment, I realized that the test is not testing the right thing. Made a change.

HappyRay: It appears that the junit runner main method will call system.exist and force the process to terminate even when a non daemon thread is still running.

    public static void main(String... args) {
        Result result = new JUnitCore().runMain(new RealSystem(), args);
        System.exit(result.wasSuccessful() ? 0 : 1);
    }

PR #1771, opened 2018-05-25
Remove incorrect throws from ExecutorServlet

HappyRay:
These methods never throw ServletException.


HappyRay:
Thanks. Mine does.

PR #1772, opened 2018-05-25
Remove e.printStackTrace

HappyRay:
Replace them with the logger.error call

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorServlet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012 LinkedIn Corp.
+ * Copyright 2018 LinkedIn Corp.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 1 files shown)


li-afaris:
It looks like the intent to move the stack trace from the HTTP reply into the host logs?  Why not have the stack trace in both places?  While not frequently used, our documentation does reference `WEBSERVER_URL/executor?ajax=...`

HappyRay:
@li-afaris 
The error out goes to a different log file. Is there any reason not to use the standard log4j config to capture such error logs?

What are the benefits to log the same in both places?

> While not frequently used, our documentation does reference WEBSERVER_URL/executor?ajax=...

What do you mean?

li-afaris:
> What are the benefits to log the same in both places?

My understanding is that e.printStackTrace would be printed to STDOUT and this could be a bad premise on my part.  But if the stack trace is printed to STDOUT, then the stack trace would be displayed to the client or browser accessing the API.  

> What do you mean?

I mentioned the documentation because people occasionally have the need to access the API through a web browser.   As a web browser can be another api client, printing a stack trace to the client could be useful. 

If I'm wrong about e.printStackTrace output going to the client accessing the API, then your change is good.

HappyRay:
I see. No,  the error output won't be returned to clients.

PR #1779, opened 2018-05-30
Fix compression vulnerability.

jamiesjc:
This is to check whether the unzipped file will be outside of the target directory. If the file has a name which traverses its parent directory, e.g. `../../../`, it might end up overwriting other files outside the target directory, causing security issues.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/Utils.java
+++ b/az-core/src/main/java/azkaban/utils/Utils.java
@@ -27,6 +27,8 @@ import java.io.OutputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.ParseException;
 import java.util.Collection;
 import java.util.Date;
@@ -198,6 +200,14 @@ public class Utils {
     final Enumeration<?> entries = source.entries();
     while (entries.hasMoreElements()) {
       final ZipEntry entry = (ZipEntry) entries.nextElement();
+      final Path fileDestinationPath = Paths.get(dest.getAbsolutePath(),
+          entry.getName()).normalize();
+      if (!fileDestinationPath.startsWith(dest.getAbsolutePath())) {
+        throw new IOException(
+            "Extracting zip entry would have resulted in a file outside the specified destination"
+                + " directory.");
+      }
+
       final File newFile = new File(dest, entry.getName());
       if (entry.isDirectory()) {
         newFile.mkdirs();
--- a/az-core/src/main/java/azkaban/utils/Utils.java
+++ b/az-core/src/main/java/azkaban/utils/Utils.java
@@ -27,8 +27,6 @@ import java.io.OutputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.text.ParseException;
 import java.util.Collection;
 import java.util.Date;
@@ -200,15 +198,13 @@ public class Utils {
     final Enumeration<?> entries = source.entries();
     while (entries.hasMoreElements()) {
       final ZipEntry entry = (ZipEntry) entries.nextElement();
-      final Path fileDestinationPath = Paths.get(dest.getAbsolutePath(),
-          entry.getName()).normalize();
-      if (!fileDestinationPath.startsWith(dest.getAbsolutePath())) {
+      final File newFile = new File(dest, entry.getName());
+      if (!newFile.getCanonicalPath().startsWith(dest.getCanonicalPath())) {
         throw new IOException(
             "Extracting zip entry would have resulted in a file outside the specified destination"
                 + " directory.");
       }
 
-      final File newFile = new File(dest, entry.getName());
       if (entry.isDirectory()) {
         newFile.mkdirs();
       } else {
--- a/az-core/src/test/java/azkaban/utils/UtilsTest.java
+++ b/az-core/src/test/java/azkaban/utils/UtilsTest.java
@@ -36,7 +36,7 @@ public class UtilsTest {
     final ZipFile source = new ZipFile(zipFile);
     final File dest = Utils.createTempDir();
     assertThatThrownBy(() -> Utils.unzip(source, dest)).isInstanceOf(IOException.class)
-        .hasMessageContaining("Extracting Zip entry would have resulted in a file outside the "
+        .hasMessageContaining("Extracting zip entry would have resulted in a file outside the "
             + "specified destination directory.");
   }
 }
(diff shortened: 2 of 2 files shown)


az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.util.zip.ZipFile;
+import org.junit.Test;
+
+/**
+ * Test class for azkaban.utils.Utils
+ */
+public class UtilsTest {
+
+  private static final String INVALID_ZIP_FILE = "zip-slip.zip";
```
reallocf: Does this resource need to be included in this PR as well?

az-core/src/main/java/azkaban/utils/Utils.java
```diff
@@ -198,6 +200,14 @@ public static void unzip(final ZipFile source, final File dest) throws IOExcepti
     final Enumeration<?> entries = source.entries();
     while (entries.hasMoreElements()) {
       final ZipEntry entry = (ZipEntry) entries.nextElement();
+      final Path fileDestinationPath = Paths.get(dest.getAbsolutePath(),
+          entry.getName()).normalize();
```
reallocf: How thorough is the `normalize()` method?

Looking at the documentation, it looks like it handles `.` and `..`, but does it handle `~`? How does it handle symlinks? (i.e. what if there's a symlink in the zip file that points to root, then the file leverages that symlink?)

az-core/src/test/java/azkaban/utils/UtilsTest.java
jamiesjc: My bad, I forgot to include in the commit, thanks!

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Test
+  public void testUnzipInvalidFile() throws IOException {
```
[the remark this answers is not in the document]
jamiesjc: Insecure sounds better to me as well, thanks!

az-core/src/test/java/azkaban/utils/UtilsTest.java
HappyRay: Good question.

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.zip.ZipFile;
+import org.junit.Test;
+
+/**
+ * Test class for azkaban.utils.Utils
+ */
+public class UtilsTest {
```
HappyRay: Thanks for adding this test!

az-core/src/test/java/azkaban/utils/UtilsTest.java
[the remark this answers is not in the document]
HappyRay: Consider explaining why this zip is insecure.

az-core/src/main/java/azkaban/utils/Utils.java
jamiesjc: Good question. As discussed offline, "normalize()" does not handle "~". If there is "~" in the file name, it will just directly interpret as the folder name "~". 
E.g. if dest path is "/user/foo" and entry name is "~/bar.txt", after calling normalize(), fileDestinationPath would be "/user/foo/~/bar.txt". It will still create the unzipped file in the current directory, which should not be a problem. 
"normalize()" does not resolve symlinks. Our current unzipping method only handles files and directories. So even if there are evil symlinks, we will not be trapped.

az-core/src/test/java/azkaban/utils/UtilsTest.java
jamiesjc: I created the zip in the code using ZipOutputStream instead of uploading the actual zip. This would be clearer to the users what the content would look in the zip.

az-core/src/test/java/azkaban/utils/UtilsTest.java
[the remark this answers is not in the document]
jamiesjc: Updated in the comments, thanks!

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
+import org.junit.Test;
+
+/**
+ * Test class for azkaban.utils.Utils
+ */
+public class UtilsTest {
+
+  /* An insecure zip file may hold path traversal filenames. During unzipping, the filename gets
```
HappyRay: Please use the standard class doc format.

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,50 @@
...
+  concatenated to the target directory. The final path may end up outside the target directory,
+  causing security issues. */
+  @Test
+  public void testUnzipInsecureFile() throws IOException {
+    final File zipFile = new File("myTest.zip");
+    final ZipOutputStream out = new ZipOutputStream(new FileOutputStream(zipFile));
+    final ZipEntry entry = new ZipEntry("../../../../../evil.txt");
```
HappyRay: Nice. The intent is more clear than using a file in the previous version.

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,50 @@
...
+    out.putNextEntry(entry);
+    out.close();
```
HappyRay: Consider https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

az-core/src/test/java/azkaban/utils/UtilsTest.java
jamiesjc: Sure, will update, thanks!

az-core/src/test/java/azkaban/utils/UtilsTest.java
jamiesjc: Sure, thanks!

az-core/src/test/java/azkaban/utils/UtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * An insecure zip file may hold path traversal filenames. During unzipping, the filename gets
+   * concatenated to the target directory. The final path may end up outside the target directory,
+   * causing security issues.
+   *
+   * @throws IOException the io exception
+   */
+  @Test
+  public void testUnzipInsecureFile() throws IOException {
+    final File zipFile = new File("myTest.zip");
```
HappyRay: The file is not cleaned up after this test. Please fix.

jamiesjc: Sure, fixed it in a new PR https://github.com/azkaban/azkaban/pull/1785, thanks!

PR #1781, opened 2018-06-01
Hdfsviewer move from azkaban-plugins

kunkun-tang:
This moduleThis PR is follow-up of #1712 .
We move hdfs viewer module from Azkaban-plugins to the main AZ repo.
Note: We decprecated JsonSequenceFileViewer out in this change, because voldemort jars cannot be found through maven. Also, we cannot find any users who still use this functionality.


az-hdfs-viewer/src/main/java/azkaban/viewer/hdfs/AvroFileViewer.java
```diff
@@ -0,0 +1,196 @@
+/*
+ * Copyright 2012 LinkedIn Corp.
```
chengren311: update copy right

kunkun-tang: This is old code. Let's change the copyright when we do refactors or format changes to these legacy codes.

PR #1782, opened 2018-06-01
code refactor on flow trigger

chengren311:
1. change retention period of historic trigger execution rows to 30 days.
2.  some code refactoring:
2.1 separate runnable code from methods which submit those runnable.
2.2 rearrange method/member variable order to make it more organized
2.3 use constant to keep the hard coded thread pool size.


HappyRay:
I would suggest separate commits for independent changes.

e.g.
> change retention period of historic trigger execution rows to 30 days.
is a functional/config change, not a refactoring.

chengren311:
@HappyRay thanks.

PR #1791, opened 2018-06-05
Handling recursive directories for Flow 2.0 projects.

jamiesjc:
This PR is to enhance the current way of loading project flows in Flow 2.0 design.
When users create the project zip, sometimes they create a new directory first, put their project/flow yaml files inside and then zip the directory. This is different from zipping those files directly. Sometimes users want to create multiple subdirectories inside the project zip.
In order to handle different use cases, I extended the current logic of loading project/flow yaml files to subdirectories as well.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
@@ -22,6 +22,7 @@ import static java.util.Objects.requireNonNull;
 import azkaban.Constants;
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.flow.Flow;
+import azkaban.project.FlowLoaderUtils.DirFilter;
 import azkaban.project.FlowLoaderUtils.SuffixFilter;
 import azkaban.project.ProjectLogEvent.EventType;
 import azkaban.project.validator.ValidationReport;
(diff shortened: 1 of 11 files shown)


azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -227,6 +222,19 @@ private void persistProject(final Project project, final FlowLoader loader, fina
     }
   }
 
+  private void uploadFlowFilesRecursively(final File projectDir, final Project project, final int
+      newProjectVersion) {
+    for (final File file : projectDir.listFiles(new SuffixFilter(Constants.FLOW_FILE_SUFFIX))) {
```
kunkun-tang: Is two for-loop a duplicate? if me, I would use one for loop to walk through all the potential files.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -112,14 +113,22 @@ private void convertYamlFiles(final File projectDir) {
               + ". Duplicate nodes found or dependency undefined.");
         } else {
           final AzkabanFlow azkabanFlow = (AzkabanFlow) loader.toAzkabanNode(nodeBean);
-          final Flow flow = convertAzkabanFlowToFlow(azkabanFlow, azkabanFlow.getName(), file);
-          this.flowMap.put(flow.getId(), flow);
+          if (this.flowMap.containsKey(azkabanFlow.getName())) {
+            this.errors.add("Duplicate flows found in the project with name " + azkabanFlow
```
kunkun-tang: Where do you check duplicate project file?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: One for-loop is to traverse .flow files in the current directory, the other for loop is to traverse the subdirectories.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Inside azkaban.project.FlowLoaderFactory#checkForValidProjectYamlFile. Need to check .project file first when deciding whether to create directoryFlowLoader (Flow 1.0) or directoryYamlFlowLoader (Flow 2.0). The check happens before loading project flows.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
jamiesjc: The filter can only be applied to directories, not individual files. If we want to use the existing filter api, then we cannot avoid two for-loops. Otherwise, we need to list all the files first, and then implement our own filtering logic on each individual file inside the directory, which I think is not quite necessary.

PR #1792, opened 2018-06-05
Configurable access error message

weidious:
What we had for error message was hard-coded static string. But now, we created a configurable error message.

### DIFF
--- a/az-hdfs-viewer/src/conf/plugin.properties
+++ b/az-hdfs-viewer/src/conf/plugin.properties
@@ -1,0 +1,2 @@
+#viewer.access_denied_message=The folder you are trying to access is protected.
+#Need a filed of "viewer.access_denied_message" specifying the error message we want user to get when they don't have a permission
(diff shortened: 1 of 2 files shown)


az-hdfs-viewer/src/conf/plugin.properties
```diff
@@ -0,0 +1,14 @@
...
+viewer.external.classpaths=extlib/*
+viewer.servlet.class=azkaban.viewer.hdfs.HdfsBrowserServlet
+hadoop.security.manager.class=azkaban.security.HadoopSecurityManager_H_1_0
+azkaban.should.proxy=false
+proxy.user=azkaban
+proxy.keytab.location=
+allow.group.proxy=false
+file.max.lines=1000
+#viewer.access_denied_message=The folder you are trying to access is protected.
+#Need a filed of "viewer.access_denied_message" specifying the error message we want user to get when they don't have a permission
```
kunkun-tang: The comment should on the top of real configuration.

PR #1794, opened 2018-06-06
fix node level computation of a flow

### LINKED ISSUE #1793: Level of a Node in a flow is always 0
224d278ddb7bacf482ff2264e807ff186e2302eb refactored the way a level of a node is calculated. The new code contains a minor bug that causes the level to be 0 for all nodes. This breaks the job list dropdown on the project page as the jobs are not ordered / indented properly (level is used for order and indentation).

The bug is at https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/flow/Flow.java#L198 where the code calls `setLevelsAndEdgeNodes(nextLevelNodes, level++);` to set levels for the next level of nodes. However, the way postfix operators work in Java is that the value of the variable is first used and only then incremented. This means the current value of level is passed to the recursive setLevelsAndEdgeNodes call instead of the incremented value.

That line should be modified to either do `++level` which first increments the value and then returns it or `level + 1` which achieves the same result.

sjakthol:
level++ returns the level variable BEFORE it is incremented. This means
the level of a node is never incremented and is always zero.

Hence, we need to increment the value by one before passing it to the
function.

Fixes #1793.

The issue was introduced in #1680

### DIFF
--- a/azkaban-common/src/test/java/azkaban/flow/FlowTest.java
+++ b/azkaban-common/src/test/java/azkaban/flow/FlowTest.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.flow;
+
+import azkaban.project.DirectoryFlowLoader;
+import azkaban.project.Project;
+import azkaban.test.executions.ExecutionsTestUtil;
+import azkaban.utils.Props;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+
+public class FlowTest {
+
+
+  private Project project;
+  private Props props;
+
+
+  @Before
+  public void setUp() throws Exception {
+    this.project = new Project(11, "myTestProject");
+    this.props = new Props();
+    final DirectoryFlowLoader loader = new DirectoryFlowLoader(this.props);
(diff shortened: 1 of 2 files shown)


HappyRay:
Thanks for the PR!

I added the following to the description to help with navigation and alert the original author and reviewers.
> The issue was introduced in #1680

Could you or @senecaso-sf think of a unit test to test this condition?

sjakthol:
I created a new test case to ensure this feature works as intended in the future.

azkaban-common/src/test/java/azkaban/flow/FlowTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class FlowTest {
+
+
+  private Project project;
+  private Props props;
+
+
+  @Before
+  public void setUp() throws Exception {
```
HappyRay: Thanks for adding tests!

Is there a good reason to make this a setup method instead of a regular private method?

azkaban-common/src/test/java/azkaban/flow/FlowTest.java
```diff
@@ -0,0 +1,56 @@
...
+    this.project = new Project(11, "myTestProject");
+    this.props = new Props();
+    final DirectoryFlowLoader loader = new DirectoryFlowLoader(this.props);
+    loader.loadProjectFlow(this.project, ExecutionsTestUtil.getFlowDir("embedded"));
+    Assert.assertEquals(0, loader.getErrors().size());
```
HappyRay: This assert seems to suggest that this test is designed to test loading of the project too. That logic should have been covered in a separate test, I assume.

Maybe a regular assert is more appropriate?

azkaban-common/src/test/java/azkaban/flow/FlowTest.java
sjakthol: No particular reason. I just copied the boilerplate over from FlowUtilsTest.java and that happened to use this kind of approach.

I have now changed this to be a private method instead.

azkaban-common/src/test/java/azkaban/flow/FlowTest.java
sjakthol: Good comment. I changed this to a regular assert.

HappyRay:
Thanks.

LGTM although I haven't verified the examples in the test. 

@jamiesjc could you do a more detailed review?

Issue #1778, opened 2018-05-30
Add a text to indicate that the table rows are sortable


fwantastic:
It's not super clear that the table rows can be sorted. Having an icon, such as up and down arrows, would be nice but the rows won't be formatted correctly when resizing the window, so at least having some kind of text at the top of the table to indicate that the rows are sortable will give a better user experience.

PR #1795, opened 2018-06-06
adding text to indicate that columns are sortable

jh3155:
As mentioned in issue #1778, having a tooltip on the table would let the user know that it can be sorted when the header is clicked. This pull request will add the tooltip on scheduling page and history page which are being mostly used.

Sample screenshot:
image

This is a duplicate PR (https://github.com/azkaban/azkaban/pull/1780) but from a different branch, I moved these changes from master->sort-text in my repo.


chengren311:
@jh3155 thanks, would you mind adding header to flow execution page as well?

jh3155:
@chengren311 it has been added on the execution page as well

chengren311:
@jh3155 thanks.

PR #1798, opened 2018-06-06
fix flaky test TriggerInstanceProcessorTest#testProcessSucceed 

chengren311:
Since processSucceed is an asyn call, test needs to wait for updateAssociatedFlowExecId to be called before verification of its invocation. So when updateAssociatedFlowExecId is called, the associated countdownlatch is decremented to zero and unit test will wait on the countdown latch before verifying invocation of updateAssociatedFlowExecId.

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
@@ -101,7 +101,6 @@ public class TriggerInstanceProcessorTest {
   public void setUp() throws Exception {
     this.message = EmailerTest.mockEmailMessage();
     this.messageCreator = EmailerTest.mockMessageCreator(this.message);
-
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
```diff
@@ -107,6 +108,7 @@ public void setUp() throws Exception {
     final CommonMetrics commonMetrics = new CommonMetrics(new MetricsManager(new MetricRegistry()));
     this.emailer = Mockito.spy(new Emailer(EmailerTest.createMailProperties(), commonMetrics,
         this.messageCreator));
+
```
shuzhang1989: extra line

chengren311: thanks, this is intentional left for separation.

PR #1803, opened 2018-06-13
Clean shared project directory when disk usage is too high

chengren311:
This PR implements LRU project purging to prevent project files eating up disk space on executor.

We will encapsulate following logic inside the background cleanup thread:
if(disk space consumed by projects >= predefined threshold) 
   List all project files and sort them by creation time in ascending order
   Iterate over the project file list, delete the file if the project is not running until disk size of projects drops down below a predefined lower threshold or no more projects to delete.

Why using creation time not last access time:
Directory last access time is not maintained by most file systems so we cannot rely on file system API to get last access time. An alternative considered is to modify the project dir every time the code reads it, then last modification time would be equivalent of last access time. The associate cons are code complexity and overhead of disk IO. We will use last creation time as the indicator of oldness of the project. Although creation time might not be as indicative as last access time to determine the hotness of a project, we are still ok with using it based on the assumption that the older the project is, more likely it’s not being used.

This PR also changes execution dir retention from 1 day to 2 hours. Since execution dir is hard linked to project dir, so disk space will be released only when no reference to project dir exists. That's why we want to shorten the execution dir retention time so that disk pressure can be alleviated sooner.

Project cleanup is executed by active executor only to prevent ensure only one thread will be performing deletion. Currently it relies on FlowRunnerManager#isExecutorActive to determine whether itself is active. But isExecutorActive is reliable only when azkaban admin uses activate API to active the executor. If admin manually updates the executors table in database and call reloadExecutors API, then the flag won't be set which inactive executor to perform deletion. So improvement need to be done here to make sure isExecutorActive is authentic.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -235,7 +235,7 @@ public class Constants {
         + ".start_cleanup_threshold";
 
     // the disk usage threshold to stop project dir cleanup.
-    // E.g, if set to 0.6, cleanup will stop when project dir size <= 0.6 of
+    // E.g, if set to 0.6, cleanup will stop when project dir size < 0.6 of
     // {@value#PROJECT_DIR_MAX_SIZE}.
     public static final String PROJECT_DIR_CLEANUP_STOP_THRESHOLD = "azkaban.project"
         + ".stop_cleanup_threshold";
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -224,6 +224,21 @@ public class Constants {
     public static final String QUEUEPROCESSING_ENABLED = "azkaban.queueprocessing.enabled";
 
     public static final String SESSION_TIME_TO_LIVE = "session.time.to.live";
+
+    // allowed max size of project dir in mb
+    public static final String PROJECT_DIR_MAX_SIZE = "azkaban.project.dir_max_size";
+
+    // the disk usage threshold to trigger project dir cleanup.
+    // E.g, if set to 0.8, cleanup will trigger when project dir size >= 0.8 of
+    // {@value#PROJECT_DIR_MAX_SIZE}.
+    public static final String PROJECT_DIR_CLEANUP_START_THRESHOLD = "azkaban.project"
+        + ".start_cleanup_threshold";
+
+    // the disk usage threshold to stop project dir cleanup.
+    // E.g, if set to 0.6, cleanup will stop when project dir size <= 0.6 of
+    // {@value#PROJECT_DIR_MAX_SIZE}.
+    public static final String PROJECT_DIR_CLEANUP_STOP_THRESHOLD = "azkaban.project"
+        + ".stop_cleanup_threshold";
   }
 
   public static class FlowProperties {
--- a/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
@@ -16,6 +16,7 @@
 
 package azkaban.utils;
 
+import com.google.common.base.Preconditions;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
@@ -28,6 +29,8 @@ import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.FileTime;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -67,6 +70,32 @@ public class FileIOUtils {
       }
     }
     return true;
+  }
+
+
+  /**
+   * Return creation time of file.
+   */
+  public static long creationTime(final File file) throws IOException {
+    final BasicFileAttributes attrs = Files
+        .readAttributes(file.toPath(), BasicFileAttributes.class);
+    final FileTime time = attrs.creationTime();
+    return time.toMillis();
+  }
+
+  /**
+   * Return the size of dir in KB.
+   */
+  public static long sizeInKB(final File dir) throws IOException {
+    Preconditions.checkArgument(dir.isDirectory(), dir + " is not a directory");
+    Preconditions.checkArgument(dir.exists(), dir + " doesn't exist");
+
+    final java.util.Scanner s = new java.util.Scanner(
+        Runtime.getRuntime().exec("du -sh -k " + dir.getAbsolutePath()).getInputStream())
+        .useDelimiter("\\A");
+    final String str = s.hasNext() ? s.next() : "";
+    final String[] res = str.split("\\s+");
+    return Long.valueOf(res[0]);
   }
 
   public static String getSourcePathFromClass(final Class<?> containedClass) {
--- a/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
@@ -78,6 +78,7 @@ public class FileIOUtils {
    * Return creation time of file.
    */
   public static long getCreationTime(final File file) throws IOException {
+    Preconditions.checkArgument(file.exists(), file + " doesn't exist");
     final BasicFileAttributes attrs = Files
         .readAttributes(file.toPath(), BasicFileAttributes.class);
     final FileTime time = attrs.creationTime();
--- a/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
@@ -77,7 +77,7 @@ public class FileIOUtils {
   /**
    * Return creation time of file.
    */
-  public static long creationTime(final File file) throws IOException {
+  public static long getCreationTime(final File file) throws IOException {
     final BasicFileAttributes attrs = Files
         .readAttributes(file.toPath(), BasicFileAttributes.class);
     final FileTime time = attrs.creationTime();
--- a/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -91,8 +92,8 @@ public class FileIOUtils {
     Preconditions.checkArgument(dir.exists(), dir + " doesn't exist");
 
     final java.util.Scanner s = new java.util.Scanner(
-        Runtime.getRuntime().exec("du -sh -k " + dir.getAbsolutePath()).getInputStream())
-        .useDelimiter("\\A");
+        Runtime.getRuntime().exec("du -sh -k " + dir.getAbsolutePath()).getInputStream(),
+        Charset.defaultCharset().name()).useDelimiter("\\A");
     final String str = s.hasNext() ? s.next() : "";
     final String[] res = str.split("\\s+");
     return Long.valueOf(res[0]);
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -70,14 +70,16 @@ public class FlowPreparer {
       // First get the ProjectVersion
       final ProjectVersion projectVersion = getProjectVersion(flow);
 
-      // Setup the project
-      setupProject(projectVersion);
+      synchronized (projectVersion) {
+        // Setup the project
+        setupProject(projectVersion);
 
-      // Create the execution directory
-      execDir = createExecDir(flow);
+        // Create the execution directory
+        execDir = createExecDir(flow);
 
-      // Create the symlinks from the project
-      copyCreateHardlinkDirectory(projectVersion.getInstalledDir(), execDir);
+        // Create the symlinks from the project
+        copyCreateHardlinkDirectory(projectVersion.getInstalledDir(), execDir);
+      }
 
       log.info(String.format("Flow Preparation complete. [execid: %d, path: %s]",
           flow.getExecutionId(), execDir.getPath()));
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -70,15 +70,21 @@ public class FlowPreparer {
       // First get the ProjectVersion
       final ProjectVersion projectVersion = getProjectVersion(flow);
 
+      // Synchronized on {@code projectVersion} to prevent
+      // cleaning thread in {@link azkaban.execapp.FlowRunnerManager}
+      // cleaning up the same project version when creating hardlink
       synchronized (projectVersion) {
+        log.info("start to setup project" + projectVersion);
+        Thread.sleep(5000);
         // Setup the project
         setupProject(projectVersion);
 
         // Create the execution directory
         execDir = createExecDir(flow);
 
-        // Create the symlinks from the project
+        // Create the links from the project
         copyCreateHardlinkDirectory(projectVersion.getInstalledDir(), execDir);
+        log.info("completes setuping project " + projectVersion);
       }
 
       log.info(String.format("Flow Preparation complete. [execid: %d, path: %s]",
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -74,8 +74,6 @@ public class FlowPreparer {
       // cleaning thread in {@link azkaban.execapp.FlowRunnerManager}
       // cleaning up the same project version when creating hardlink
       synchronized (projectVersion) {
-        log.info("start to setup project" + projectVersion);
-        Thread.sleep(5000);
         // Setup the project
         setupProject(projectVersion);
 
@@ -84,11 +82,10 @@ public class FlowPreparer {
 
         // Create the links from the project
         copyCreateHardlinkDirectory(projectVersion.getInstalledDir(), execDir);
-        log.info("completes setuping project " + projectVersion);
+        log.info(String.format("Flow Preparation complete. [execid: %d, path: %s]",
+            flow.getExecutionId(), execDir.getPath()));
       }
 
-      log.info(String.format("Flow Preparation complete. [execid: %d, path: %s]",
-          flow.getExecutionId(), execDir.getPath()));
     } catch (final Exception e) {
       log.error("Error in setting up project directory: " + this.projectsDir + ", Exception: " + e);
       cleanup(execDir);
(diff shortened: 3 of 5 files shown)


azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -69,6 +73,32 @@ public static boolean isDirWritable(final File dir) {
...
+    final BasicFileAttributes attrs = Files
+        .readAttributes(file.toPath(), BasicFileAttributes.class);
+    final FileTime time = attrs.creationTime();
+    return time.toMillis();
+  }
+
+  /**
+   * Return the size of dir in KB.
+   */
+  public static long sizeInKB(final File dir) throws IOException {
```
jamiesjc: Can we use `FileUtils.sizeOfDirectory` to get the size directly?

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -69,6 +73,32 @@ public static boolean isDirWritable(final File dir) {
     return true;
   }
 
+
+  /**
+   * Return creation time of file.
+   */
+  public static long creationTime(final File file) throws IOException {
```
jamiesjc: Does `getCreationTime` sound better?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -70,14 +70,16 @@ void setup(final ExecutableFlow flow) {
       // First get the ProjectVersion
       final ProjectVersion projectVersion = getProjectVersion(flow);
 
-      // Setup the project
-      setupProject(projectVersion);
+      synchronized (projectVersion) {
```
jamiesjc: What's the intention of synchronizing on projectVersion here?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
       this.projectDirectory.mkdirs();
     }
 
+    this.projectDirMaxSizeInMB = props.getLong(ConfigurationKeys.PROJECT_DIR_MAX_SIZE,
+        2000000); // default value as 2TB
+    this.projectDirStartDeletionThreshold = props
+        .getDouble(ConfigurationKeys.PROJECT_DIR_CLEANUP_START_THRESHOLD, 0.9);
+    this.projectDirStopDeletionThreshold = props
+        .getDouble(ConfigurationKeys.PROJECT_DIR_CLEANUP_STOP_THRESHOLD, 0.8);
```
jamiesjc: Is there any reason why we set the threshold between 0.8-0.9?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -970,6 +992,102 @@ private void cleanOlderProjects() {
...
+          new HashSet<>();
+      for (final FlowRunner runner : FlowRunnerManager.this.runningFlows.values()) {
+        final ExecutableFlow flow = runner.getExecutableFlow();
+        activeProjectVersions.add(new Pair<>(flow
+            .getProjectId(), flow.getVersion()));
+      }
+      return activeProjectVersions;
+    }
+
+    private boolean isActiveProject(final ProjectVersion version) {
```
jamiesjc: Do we define active projects as the projects that have currently running flows? What if they have scheduled flows? Should it also be considered as active?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -970,6 +992,102 @@ private void cleanOlderProjects() {
...
+            } else {
+              return 0;
+            }
+          } catch (final IOException ex) {
+            logger.error("error reading project dir", ex);
+            return 0;
+          }
+        });
+
+        for (final ProjectVersion version : projectVersionList) {
```
jamiesjc: Will it take a long time to calculate the project cache directory size since it's very big? And it will be called inside `shouldKeepCleanup` method which gets executed for every project version in the for loop. Would it be more efficient to do some batch cleaning?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: it's a reasonable number.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: Active projects are the one containing currently running flows.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: sizeOfDirectory is not precise to get the read disk usage. E.g directory itself has disk usage but it doesn't count in sizeOfDirectory.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
chengren311: updated in comment

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: yes, but this the frequency of doing it is once/30 mins. 
What do you mean by batch cleaning?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -70,17 +70,22 @@ void setup(final ExecutableFlow flow) {
...
-
-      // Create the execution directory
-      execDir = createExecDir(flow);
-
-      // Create the symlinks from the project
-      copyCreateHardlinkDirectory(projectVersion.getInstalledDir(), execDir);
+      // Synchronized on {@code projectVersion} to prevent
+      // cleaning thread in {@link azkaban.execapp.FlowRunnerManager}
+      // cleaning up the same project version when creating hardlink
+      synchronized (projectVersion) {
```
jamiesjc: In `getProjectVersion()`, it seems the projectVersion returned is a new object. Will the synchronization really work if they are synchronized on different objects here?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jamiesjc: I was thinking if it might be possible to clean up like 10 projects at a time and then do the check in `shouldKeepCleanup`. But I'm not sure how much improvement this will have depending on how long it will take in calculating the directory size. If you think efficiency should not be a problem, then I'm ok with this implementation.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -134,7 +142,7 @@
   private Props globalProps;
 
   private long lastCleanerThreadCheckTime = -1;
-  private long executionDirRetention = 1 * 24 * 60 * 60 * 1000; // 1 Day
+  private long executionDirRetention = 1 * 6 * 60 * 60 * 1000; // 6 hours
```
jamiesjc: Why change it to 6 hours?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
chengren311: it will only return new object only when projectId, version don't present in the hash map.
```   
 final ProjectVersion projectVersion;
    synchronized (this.installedProjects) {
      projectVersion = this.installedProjects
          .computeIfAbsent(new Pair<>(flow.getProjectId(), flow.getVersion()),
              k -> new ProjectVersion(flow.getProjectId(), flow.getVersion()));
    }
    return projectVersion;
 ```

File: azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java (hunk 8 above)
Comment (chengren311) [reply azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java]: reduce it to 2 hours.
```
This PR also changes execution dir retention from 1 day to 2 hours. Since execution dir is hard linked to project dir, so disk space will be released only when no reference to project dir exists. That's why we want to shorten the execution dir retention time so that disk pressure can be alleviated sooner.

```

File: az-core/src/main/java/azkaban/Constants.java [hunk 9]
```diff
@@ -224,6 +224,21 @@
     public static final String QUEUEPROCESSING_ENABLED = "azkaban.queueprocessing.enabled";
 
     public static final String SESSION_TIME_TO_LIVE = "session.time.to.live";
+
+    // allowed max size of project dir in mb
+    public static final String PROJECT_DIR_MAX_SIZE = "azkaban.project.dir_max_size";
```
HappyRay: Consider adding the unit in the key name? e.g.
dir_max_size_in_mb?

Do you foresee other keys in the project namespace? Project-dir can be interpreted as an individual project directory. 
Would it be better to name it 
azkaban.project_cache_max_size_in_mb?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -224,6 +224,21 @@
...
+
+    // the disk usage threshold to trigger project dir cleanup.
+    // E.g, if set to 0.8, cleanup will trigger when project dir size >= 0.8 of
+    // {@value#PROJECT_DIR_MAX_SIZE}.
+    public static final String PROJECT_DIR_CLEANUP_START_THRESHOLD = "azkaban.project"
+        + ".start_cleanup_threshold";
+
+    // the disk usage threshold to stop project dir cleanup.
+    // E.g, if set to 0.6, cleanup will stop when project dir size < 0.6 of
```
HappyRay: What do you think about making this an integer?
e.g.
60 means 60%?

HappyRay:
Where are the tests for this change?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -134,7 +142,7 @@
   private Props globalProps;
 
   private long lastCleanerThreadCheckTime = -1;
-  private long executionDirRetention = 1 * 24 * 60 * 60 * 1000; // 1 Day
+  private long executionDirRetention = 1 * 2 * 60 * 60 * 1000; // 2 hours
```
HappyRay: Is this change necessary to be included in this change?

This is not related to the project cache directory.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
       this.projectDirectory.mkdirs();
     }
 
+    this.projectDirMaxSizeInMB = props.getLong(ConfigurationKeys.PROJECT_DIR_MAX_SIZE,
```
HappyRay: "projectDirMaxSizeInMb" see coding standard naming.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
...
+        2000000); // default value as 2TB
```
HappyRay: Why is 2 TB a good default? I don't think most of the deployment will require such a big cache.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
...
+    this.projectDirStartDeletionThreshold = props
+        .getDouble(ConfigurationKeys.PROJECT_DIR_CLEANUP_START_THRESHOLD, 0.9);
```
HappyRay: Is 10% buffer big enough when the default max size is smaller?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
...
+    this.projectDirStopDeletionThreshold = props
```
HappyRay: I would suggest a lower number so that each clean up can clean up at least 20% or more of the max to reduce the number of times the clean up needs to run and thus reducing the amortized cost of each clean up.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -173,6 +181,16 @@ public FlowRunnerManager(final Props props,
...
+        .getDouble(ConfigurationKeys.PROJECT_DIR_CLEANUP_STOP_THRESHOLD, 0.8);
+    Preconditions.checkArgument(this.projectDirStartDeletionThreshold >= 0 && this
```
HappyRay: These are not arguments. It seems a bit odd.

HappyRay: Consider adding methods to get config values and add validation logic there.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -781,11 +799,14 @@ public void deleteExecutionDirectory() {
     private static final long OLD_PROJECT_DIR_INTERVAL_MS = 5 * 60 * 1000;
     // Every 2 mins clean the recently finished list
     private static final long RECENTLY_FINISHED_INTERVAL_MS = 2 * 60 * 1000;
+    // Every 30 mins clean least recently used project dir
```
HappyRay: Should this value be configurable?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -840,6 +861,13 @@ public void run() {
               this.lastExecutionDirCleanTime = currentTime;
             }
 
+            if (currentTime - LRU_PROJECT_DIR_INTERVAL_MS > this.lastLRUProjectCleanTime
```
HappyRay: Make this block a private method?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -970,6 +992,105 @@ private void cleanOlderProjects() {
         }
       }
     }
+
+    private HashSet<Pair<Integer, Integer>> getActiveProjectVersions() {
+      final HashSet<Pair<Integer, Integer>> activeProjectVersions =
+          new HashSet<>();
+      for (final FlowRunner runner : FlowRunnerManager.this.runningFlows.values()) {
+        final ExecutableFlow flow = runner.getExecutableFlow();
```
HappyRay: What happens if the runningFlows changes between this statement and the call to .values()?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
HappyRay: Have you measured the performance of the current design?

In case performance is proven to be an issue, have you considered the following alternative design? 

Delete older versions of the project immediately. This will free up disk space more quickly.

Create a file in each project directory and write the size of the project to the file when the project is created. The project files are not supposed to change after creation. Touch this file each time the project is used. This way, we can have a more efficient LRU algorithm based on last access time, not creation time. 

Maintain the total size of the project cache in memory to avoid the overhead of re-calculating it. The size shouldn't change too often. This way we can afford to run the check more frequently.

HappyRay:
How do you plan to monitor this feature?

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: high accuracy doesn't hurt, also FileUtils.sizeOfDirectory doesn't really indicate the real disk usage of the directory as size of a file is not equivalent of its associated disk cost. https://superuser.com/questions/94217/why-ls-and-du-show-different-size

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -224,6 +224,21 @@
...
+    // {@value#PROJECT_DIR_MAX_SIZE}.
+    public static final String PROJECT_DIR_CLEANUP_STOP_THRESHOLD = "azkaban.project"
+        + ".stop_cleanup_threshold";
```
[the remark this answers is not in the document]
chengren311: it's auto formatted by save plugin.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
HappyRay: The downside is unnecessary complexity. 

The referenced article seems to refer to a sparse file. The project files are not sparse files, are they?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
HappyRay: Was it subject to race conditions then? Could you describe the problem you tried to fix in more details?

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
HappyRay: public static final String PROJECT_DIR_CLEANUP_START_THRESHOLD =
        "azkaban.project.start_cleanup_threshold";
seems to work for me.

az-core/src/main/java/azkaban/Constants.java
chengren311: updated  in https://github.com/azkaban/azkaban/pull/1812

az-core/src/main/java/azkaban/Constants.java
chengren311: updated in https://github.com/azkaban/azkaban/pull/1812

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
chengren311: updated in #1812

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated in #1812

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated in #1812

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated in #1812

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
chengren311: updated in #1812

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: FileUtils.sizeOfDirectory is a lot slower than running du -sh based on my experiment of reading 10000+ project dir.

PR #1804, opened 2018-06-15
ExecutorManager refactor

chengren311:
1. Simplify logic
2. Change activeExecutors from HashSet to ImmutableSet to guarantee thread safety.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -235,6 +235,7 @@ public class ExecutorManager extends EventHandler implements
       newExecutors.add(new Executor(executor.getId(), executorHost,
           executorPort, true));
     } else {
+      // single executor mode and no executor port specified in azkaban properties
       //todo chengren311: convert to slf4j and parameterized logging
       final String error = "Missing" + ConfigurationKeys.EXECUTOR_PORT + " in azkaban properties.";
       logger.error(error);
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -258,41 +258,39 @@ public class ExecutorManager extends EventHandler implements
    * Refresh Executor stats for all the actie executors in this executorManager
    */
   private void refreshExecutors() {
-    synchronized (this.activeExecutors) {
 
-      final List<Pair<Executor, Future<ExecutorInfo>>> futures =
-          new ArrayList<>();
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -232,19 +234,19 @@ public void setupExecutors() throws ExecutorManagerException {
       }
       newExecutors.add(new Executor(executor.getId(), executorHost,
           executorPort, true));
+    } else {
+      //todo chengren311: convert to slf4j and parameterized logging
```
kunkun-tang: May we add a comment here to explain what cases should go into the else block?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -105,6 +105,8 @@
   private final int maxConcurrentRunsOneFlow;
   QueuedExecutions queuedFlows;
   File cacheDir;
+  //make it immutable to ensure threadsafety
+  private volatile ImmutableSet<Executor> activeExecutors = null;
```
kunkun-tang: How is the thread-safe guaranteed compared to HashSet?

chengren311: since immutable data structure is always thread safe.

kunkun-tang: I'd like to learn which part of code need the thread-safe data structure, activeExecutors? Would you mind point it out?

chengren311: yes, so the class contains multiple pieces iterating over activeExecutor, and code piece populating activeExecutors from database. activeExecutors should be immune to race condition when iterating and populating activeExecutors are in different thread context.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -256,41 +260,39 @@ private boolean isMultiExecutorMode() {
...
-        } catch (final TimeoutException e) {
-          wasSuccess = false;
-          logger.error("Timed out while waiting for ExecutorInfo refresh"
-              + executor, e);
-        } catch (final Exception e) {
-          wasSuccess = false;
-          logger.error("Failed to update ExecutorInfo for executor : "
-              + executor, e);
-        }
+    final List<Pair<Executor, Future<ExecutorInfo>>> futures =
```
chengren311: this just removes synchronized (this.activeExecutors)

PR #1806, opened 2018-06-19
add proxy user in log when fetching token from hive metastore

chengren311:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -340,26 +340,26 @@ public boolean shouldProxy() {
...
-        final Class credentialClass = Class.forName(credentialClassName);
-
-        // The credential class must have a constructor accepting 3 parameters, Credentials,
-        // Props, and Logger in order.
-        Constructor constructor = credentialClass.getConstructor (new Class[]
-            {Credentials.class, Props.class, Logger.class});
-        final CredentialProvider customCredential = (CredentialProvider) constructor
-              .newInstance(hadoopCred, props, jobLogger);
-        customCredential.register(userToProxy);
+    try {
```
chengren311: auto format change

HappyRay: Consider making the format change in a separate PR first.

HappyRay:
Why is this change needed?

Consider adding an example before and after log.

PR #1812, opened 2018-06-20
address feedback from #1803

chengren311:
see comments from #1803 

1. change configurable key name
2. change default value for max cache size to 128GB which is a more standard number.
3. change default value for stop cleanup threshold to 60 to allow more space to be freed.
4. change data type of stop cleanup threshold from double to int.
5. rename the parameter name(projectDirMaxSizeInMB - > projectDirMaxSizeInMb) to align with coding standard.


HappyRay:
Is it safer to turn off this feature by default? Otherwise, users including us may get surprised with the current default behavior.

PR #1813, opened 2018-06-20
Enable project uploading from restli endpoint schedule flow trigger

chengren311:
Upon project uploading, azkaban should be able to schedule the associated flow trigger. This is already achieved when project upload is done from ajax API call(#1631). This PR enables scheduling flow trigger when uploading project from restli endpoint.

### DIFF
--- a/azkaban-web-server/src/restli/java/azkaban/restli/ProjectManagerResource.java
+++ b/azkaban-web-server/src/restli/java/azkaban/restli/ProjectManagerResource.java
@@ -68,8 +68,7 @@ public class ProjectManagerResource extends ResourceContextHolder {
 
     final FlowTriggerScheduler scheduler = getAzkaban().getScheduler();
     final boolean enableQuartz = getAzkaban().getServerProps().getBoolean(ConfigurationKeys
-            .ENABLE_QUARTZ,
-        false);
+        .ENABLE_QUARTZ, false);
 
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/restli/java/azkaban/restli/ProjectManagerResource.java
```diff
@@ -111,16 +120,24 @@ public String deploy(@ActionParam("sessionId") final String sessionId,
...
     try {
+      if (enableQuartz) {
+        scheduler.unscheduleAll(project);
+      }
       // Check if project upload runs into any errors, such as the file
       // having blacklisted jars
       final Props props = new Props();
       final Map<String, ValidationReport> reports = projectManager
           .uploadProject(project, archiveFile, "zip", user, props);
+
```
jamiesjc: If it throws exception at uploadProject step, the existing schedules will still be removed but not added back?

chengren311: it's a good point. Actually it could happen when HDFS as project zip storage layer is down but mysql is up. Schedule is not the only thing that worth concerning if this is the case. Maintaining atomicity is complicated enough to fit into another PR. added todo here for record.

VemberZhang:
how to setup a project depend another project in job config file?

PR #1814, opened 2018-06-21
when trigger is still running, endtime should show "-"

chengren311:
Before endtime was showing current time when trigger instance is running,
after:
image

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/flow-trigger-list.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/flow-trigger-list.js
@@ -229,11 +229,15 @@ azkaban.FlowTriggerInstanceListView = Backbone.View.extend({
     var startTime = node.dependencyStartTime == 0 ? (new Date()).getTime()
         : node.dependencyStartTime;
 
-    var endTime = node.dependencyEndTime == 0 ? (new Date()).getTime()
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/web/js/azkaban/view/flow-trigger-list.js
```diff
@@ -229,11 +232,15 @@ azkaban.FlowTriggerInstanceListView = Backbone.View.extend({
     var startTime = node.dependencyStartTime == 0 ? (new Date()).getTime()
         : node.dependencyStartTime;
 
-    var endTime = node.dependencyEndTime == 0 ? (new Date()).getTime()
-        : node.dependencyEndTime;
+    var endTime = node.dependencyEndTime;
 
     $(tdStart).text(getDateFormat(new Date(startTime)));
-    $(tdEnd).text(getDateFormat(new Date(endTime)));
+    if (node.dependencyEndTime == 0) {
```
jamiesjc: Maybe combine the if-else statement of `tdEnd` and `tdElapse`? Also do you need that `var endTime = node.dependencyEndTime;`?

chengren311: updated, thanks

PR #1816, opened 2018-06-21
MR / YARN application is not killed when HadoopJavaJob is killed if azkaban.flow.execid is overridden (#1815)

### LINKED ISSUE #1815: MR / YARN application is not killed when HadoopJavaJob is killed if azkaban.flow.execid is overridden
Azkaban writes the job logs to a file `_job.<execId>.<jobId>.log`. If a job is killed, Azkaban scans this log file for YARN applications it should kill. However, if `azkaban.flow.execid` is overridden (either set as a job property or overridden as a flow parameter), then if the Azkaban job is killed, when `HadoopJavaJob.cancel()` is called, it encounters the error

```
java.lang.IllegalArgumentException: the logFilePath does not exist
```

and the launched YARN applications are not killed.

To fix this, when constructing the log file name in the `cancel()` method, it should use the original Azkaban-generated execution id, which can be found in the JVM parameters (`JobRunner.insertJVMAargs()` adds the original execution id as a system property).

Internal JIRA ref: LIHADOOP-37904

erwa:
Fixes #1815.

With this change, even if a user overrides `azkaban.flow.execid`, the original execution id Azkaban generated will still be used to find the log file, which is searched for YARN applications to kill when `HadoopJavaJob.cancel()` is called.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJob.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJob.java
@@ -19,13 +19,11 @@ package azkaban.jobtype;
 import static org.apache.hadoop.security.UserGroupInformation.HADOOP_TOKEN_FILE_LOCATION;
 
 import java.io.File;
-import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Properties;
 import java.util.StringTokenizer;
-
-import org.apache.hadoop.security.UserGroupInformation;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.log4j.Logger;
 
 import azkaban.flow.CommonJobProperties;
@@ -43,6 +41,8 @@ public class HadoopJavaJob extends JavaProcessJob {
   public static final String DEFAULT_CANCEL_METHOD = "cancel";
   public static final String DEFAULT_RUN_METHOD = "run";
   public static final String DEFAULT_PROGRESS_METHOD = "getProgress";
+
+  private static final Pattern AZKABAN_EXEC_ID_PATTERN = Pattern.compile("-Dazkaban.execid=(\\d+)");
 
   private String _runMethod;
   private String _cancelMethod;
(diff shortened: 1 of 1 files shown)


oliverhu:
What's the point of providing the capability to override _azkaban.flow.execid_ ?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJob.java
```diff
@@ -249,6 +249,22 @@ public String toString() {
         + getJobProps() + '}';
   }
 
+  /**
+   * Returns the {@code azkaban.execid} system property, if set, and otherwise returns the value of
+   * {@value CommonJobProperties#EXEC_ID}. {@value CommonJobProperties#EXEC_ID} may be overridden by a flow param or
+   * job property, whereas the {@code azkaban.execid} system property will be the unique execution id Azkaban generates
+   * for each flow.
+   */
+  private String getExecId() {
```
oliverhu: I'm not sure if we want to piggyback on this tho, what if user intentionally wants to kill the exec_id he overrides with?

erwa:
Hey Oliver,

The use case for overriding the execution id was as follows:

* A flow writes data to HDFS at a location configured via a parameter `output_location=/tmp/${azkaban.flow.execid}`.
* In one run (execution id `123`), the flow fails on job 3.
* In the next run (execution id `124`), instead of re-running all the jobs, the user just wants to run job 3 and read the data already generated, but it is located at `/tmp/123` rather than `/tmp/124`. So he overrides `${azkaban.flow.execid}` and sets it to `123`.

A simple workaround that does not require overriding Azkaban runtime properties is to define another property:

```
guid=${azkaban.flow.execid}
output_location=/tmp/${guid}
```

Then the user can just override `guid` and leave `azkaban.flow.execid` untouched.

I've also filed #1818 for discussing whether Azkaban should allow Azkaban-injected runtime properties to be overwritten or not.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJob.java
erwa: User does not control the `cancel()` method that kills the job. The `cancel()` method should be able to find the log file that Azkaban created -- the log file is then searched for YARN application that need to be killed.

chengren311:
@erwa thanks for the PR.
 In this case, user job should have a parameter for input data location. Overriding flow execution id itself seems a bit odd to me. azkaban runtime properties by concept are generated by azkaban platform and immutable to users.

erwa:
Thanks for the discussion, Cheng. Upon further reflection, I think using the workaround is a cleaner solution. Will discard this pull request.

PR #1819, opened 2018-06-22
Azkaban configuration validator

chengren311:
Currently azkaban configuration validation logic is scattered and validation is performed when using the config. This PR adds a config validator serving as a central place to perform validation for executor and web server. This validator will be used when starting azkaban server, and startup will fail if validation fails.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -311,4 +311,13 @@ public class Constants {
     public static final String START_TIME = "startTime";
     public static final String TRIGGER_INSTANCE_ID = "triggerInstanceId";
   }
+
+
+  public static class DefaultValue {
+
+    public static final int PROJECT_DIR_CLEANUP_START_THRESHOLD = 90;
+    public static final int PROJECT_DIR_CLEANUP_STOP_THRESHOLD = 60;
+    //128G
+    public static final long PROJECT_DIR_MAX_SIZE = 12800;
+  }
 }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorConfigValidator.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorConfigValidator.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2018 LinkedIn Corp.
+ *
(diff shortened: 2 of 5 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -311,4 +311,13 @@
...
     public static final String TRIGGER_INSTANCE_ID = "triggerInstanceId";
   }
+
+
+  public static class DefaultValue {
+
+    public static final int PROJECT_DIR_CLEANUP_START_THRESHOLD = 90;
+    public static final int PROJECT_DIR_CLEANUP_STOP_THRESHOLD = 60;
+    //128G
+    public static final long PROJECT_DIR_MAX_SIZE = 12800;
```
HappyRay: Is it in MB? 
Should it be 128_000 then?

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorConfigValidator.java
```diff
@@ -0,0 +1,65 @@
...
+import azkaban.utils.Props;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Validator of Azkaban executor's configuration
+ */
+
+@Singleton
+public class ExecutorConfigValidator extends ConfigValidator {
```
HappyRay: Is it better to use composition instead of inheritance here? Using inheritance for code reuse only is an anti-pattern IMHO.

az-core/src/main/java/azkaban/Constants.java
chengren311: thanks.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecutorConfigValidator.java
shuzhang1989: i think here is not just for code reuse? Semantically inheriting an abstract class makes sense, users are free to pass in customized logic to build their own validator right?

 Another example is rocksdb's CompactionFilter, which the db provides an interface, user can write their own compaction logic: https://github.com/facebook/rocksdb/blob/c364eb42b5fde3bf501e04c5a23f32c99ec4465c/utilities/blob_db/blob_compaction_filter.cc#L17

PR #1823, opened 2018-06-26
remove in-memory cache of project list from FlowRunnerMananger

chengren311:
Previously FlowRunnerManager has a variable named "installedProjects" to keep the projects currently residing in the shared project dir. It gets initialized when FlowRunnerManager is constructed and used to look up the project directory by project ID and version. It's maintained by a cleaner thread inside FlowRunnerManager and FlowPreparer -  when new project is downloaded or old project is deleted, its corresponding in-memory mapping is kicked out from  installedProjects. 

In addition to the complexity of maintaining this data structure, it has the risk of getting out of sync - e.g: When FlowRunnerManager populates installedProjects upon construction, sometimes executor is still in inactive mode, bearing some verification work like running integration test. However, another executor at same host which is active is performing project cleaning up. Some projects might be deleted by the active executor but not reflected by installedProjects in the inactive executor. 

FlowRunnerManager reads from installedProjects when deleting project of old versions and cleaning up project cache dir, these two operations happen at low frequency(every 5 mins and every 30 mins), loading project dirs into the installedProjects takes trivial amount of time in our experiments (< 1 sec loading 16000 dirs).

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -238,6 +238,9 @@ public class Constants {
     // the disk usage threshold to stop project dir cleanup.
     // E.g, if set to 60, cleanup will stop when project dir size < 60% of {@value#PROJECT_DIR_MAX_SIZE}.
     public static final String PROJECT_DIR_CLEANUP_STOP_THRESHOLD = "azkaban.project_cache.stop_cleanup_threshold";
+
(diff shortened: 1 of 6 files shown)


chengren311:
Closing the PR for now. Another solution could be refresh the installedProjects when executor is set to active, and we could leverage the installProjects to keep disk usage data for each project.

PR #1825, opened 2018-06-27
POC for conditional workflow design. Condition on runtime parameters.

jamiesjc:
This PR is to demonstrate how we can achieve condition on runtime parameters.
Users can specify the condition on some common job/flow/cluster related properties or runtime variables generated in previous jobs.
In order to achieve that, users need to add a `condition` section for the target job using logical expressions. Below operators will be supported:

- Equality and Relational Operators: `==, !=,  >, >=, <, <= `
- Conditional Operators: `&&, ||`

Use `${}` for variable substitution. 
Use `:` for namespacing. It will separate a job name and the parameter. 

Here is one example of conditional worflow with the condition on runtime parameters.
First JobA will generate param1 and write to $JOB_OUTPUT_PROP_FILE. Similar for JobB. After both jobs finish,  we will check whether the condition for JobC: `${JobA:param1} == 1 && ${JobB:param2} > 5` is satisfied. If yes, JobC will start to run, otherwise it will be cancelled.

conditional_workflow.flow
```
nodes:
 - name: JobA
   type: command
   config:
     command: echo {"param1":"1"} > $JOB_OUTPUT_PROP_FILE 

 - name: JobB
   type: command
   config:
     command: echo {"param2":"6"} > $JOB_OUTPUT_PROP_FILE 

 - name: JobC
   type: command
   dependsOn:
     - JobA
     - JobB
   config:
     command: echo “This is JobC.”
   condition: ${JobA:param1} == 1 && ${JobB:param2} > 5
```

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -55,6 +55,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.io.Files;
 import java.io.File;
+import java.io.FilenameFilter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -68,6 +69,11 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
+import javax.script.ScriptException;
 import org.apache.commons.io.FileUtils;
 import org.apache.log4j.Appender;
 import org.apache.log4j.FileAppender;
@@ -83,6 +89,9 @@ public class FlowRunner extends EventHandler implements Runnable {
 
   private static final Layout DEFAULT_LAYOUT = new PatternLayout(
       "%d{dd-MM-yyyy HH:mm:ss z} %c{1} %p - %m\n");
+  // Pattern to match job variables in condition expressions: ${jobName:variable}
+  private static final Pattern VARIABLE_REPLACEMENT_PATTERN = Pattern
+      .compile("\\$\\{([^:{}]+):([^:{}]+)\\}");
   // We check update every 5 minutes, just in case things get stuck. But for the
   // most part, we'll be idling.
   private static final long CHECK_WAIT_MS = 5 * 60 * 1000;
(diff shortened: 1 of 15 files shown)


jamiesjc:
#1826

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -865,6 +893,64 @@ public Status getImpliedStatus(final ExecutableNode node) {
...
+      this.logger.info("Condition is " + replaced);
+    }
+
+    // Evaluate string expression using script engine
+    return evaluateExpression(replaced);
+  }
+
+  private String findValueForJobVariable(final String jobName, final String variable) {
+    // Get props from job props tmp file
+    final File jobPropsFile = findPropsFileForJob(jobName + "_props");
```
kunkun-tang: Do we only have Prop_In and Prop_Out files given a job? Is jobName_Props_tmp is the naming template we currently follow. why is called "_tmp"? Is it possible that we can change its naming template?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalFlowTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    assertStatus(flow, "jobA", Status.SUCCEEDED);
+    assertStatus(flow, "jobB", Status.SUCCEEDED);
+    assertStatus(flow, "jobC", Status.CANCELLED);
+    assertStatus(flow, "jobD", Status.CANCELLED);
+    assertFlowStatus(flow, Status.KILLED);
+  }
+
+  @Test
+  public void runFlowOnJobOutputCondition() throws Exception {
```
kunkun-tang: Should we include a successful-run test?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
jamiesjc: We only have <jobname>_output_***_ tmp and <jobname>_props_***_tmp files. See azkaban.jobExecutor.AbstractProcessJob#createFlattenedPropsFile and azkaban.jobExecutor.AbstractProcessJob#createOutputPropsFile for the naming. These are basically temp files. Why do you think we need to change the naming?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalFlowTest.java
jamiesjc: Each test case has already included both jobs that satisfy the condition and those don't. So I think it should be sufficient to demonstrate.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
kunkun-tang: Why the name followed up with `_tmp`?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalFlowTest.java
kunkun-tang: I see. Thanks for the explanation.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
jamiesjc: Because the files are created using File.createTempFile, it makes sense to me having _tmp in the name, e.g., jobA_output_1416309822424015744_tmp. This can also be easily distinguished from user's files in the execution directory.

jamiesjc:
@kunkun-tang Also String comparison is supported. E.g. ` condition: ${jobA:azkaban.server.name} == 'foo'`

PR #1831, opened 2018-06-29
Solo server to start executor first

juhoautio:
This is the correct dependency order. Otherwise it can happen that web server tries to launch an execution, but it fails as executor server hasn't started yet.

Also start ExecutorManager's threads (that periodically call executors to get updates) not sooner than when the web server is started. Those threads used to be started already as soon as Guice injection creates the ExecutorManager instance (which happens before executor server is started – even now that the order is fixed).

----

This is one step towards having everything run in "multi-executor" mode.

As you may know, currently there are two possible modes, controlled by `azkaban.use.multiple.executors=true|false`:
1. multi-executor mode, based on the `executors` DB table
2. single executor mode, based on fixed props `executor.host` & `executor.port`

That adds extra complexity and different branches of code are executed depending on the mode. However, everything can just as well be run as "multi-executor" (but just one executor if that's what you want, and naturally so in case of solo-server). It makes it also easier to reach a better test coverage when testing is always done in the same "mode" that's used in production configurations.

I also noticed that ExecutorManager loads executors from the DB at the time of Guice injection (before executor server is started). This may need to be changed as well to properly support multi-executor mode on a single server.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -158,6 +158,9 @@ public class ExecutorManager extends EventHandler implements
   public void start() {
     this.executingManager.start();
     this.cleanerThread.start();
+    if (isMultiExecutorMode()) {
+      this.queueProcessor.start();
+    }
   }
 
   private String findApplicationIdFromLog(final String logData) {
@@ -203,8 +206,6 @@ public class ExecutorManager extends EventHandler implements
             this.azkProps.getInt(
                 Constants.ConfigurationKeys.MAX_DISPATCHING_ERRORS_PERMITTED,
                 this.activeExecutors.size()));
-
-    this.queueProcessor.start();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


juhoautio:
@kunkun-tang thanks! Anything else required for this to be merged?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -154,8 +153,11 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader loader,
             DEFAULT_EXECUTION_LOGS_RETENTION_MS);
 
     this.cleanerThread = new CleanerThread(executionLogsRetentionMs);
-    this.cleanerThread.start();
+  }
 
+  public void start() {
```
chengren311: how about queueProcessor thread?

juhoautio: Yeah, why not. Added commit.

juhoautio:
I don't understand this Travis build failure.

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':azkaban-common:compileTestJava'.
> java.lang.reflect.InvocationTargetException
```

Locally the build works. Am I missing the actual reason?

PR #1832, opened 2018-06-29
Remove unused method fetchActiveFlowByExecId

juhoautio:

PR #1833, opened 2018-06-29
Finalize running flows without matching executor

juhoautio:
A more complete cleanup: don't leave executions whose executor doesn't exist any more in some interim state like RUNNING - finalize them as FAILED.

----

To explain the problem scenario, it goes like this:
- if an executor is stopped gracefully (so that it has time to finish at least some shutdown hooks), it removes itself from the `executors` table 
- however, it's still possible that not all executions on that executor were properly finalized ie. status of each interrupted execution is set to one of FAILED/KILLED/SUCCESS
- when an executor is started, it adds itself into the `executors` table with an executor id that is the _next running number_

As a result, there can be executions with status RUNNING (for example) that would never be marked as FAILED, because the executor id of those executions is not found in `executors` table any more. This PR fixes that: the status of such executions are finalized.

This is only a concern in multi-executor mode. To be noted, it would be good to support _only multi-exec mode_ in the future. See also description of https://github.com/azkaban/azkaban/pull/1831.

----

Somewhat related: https://github.com/azkaban/azkaban/pull/1832 (merge order doesn't matter).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -54,6 +54,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
@@ -393,10 +394,6 @@ public class ExecutorManager extends EventHandler implements
     return this.lastThreadCheckTime;
   }
 
-  public long getLastCleanerThreadCheckTime() {
-    return this.lastCleanerThreadCheckTime;
-  }
-
   @Override
   public Collection<Executor> getAllActiveExecutors() {
     return Collections.unmodifiableCollection(this.activeExecutors);
@@ -438,7 +435,10 @@ public class ExecutorManager extends EventHandler implements
     for (final Pair<ExecutionReference, ExecutableFlow> running : this.runningFlows
         .values()) {
       final ExecutionReference ref = running.getFirst();
-      ports.add(ref.getHost() + ":" + ref.getPort());
+      if (ref.getExecutor().isPresent()) {
+        final Executor executor = ref.getExecutor().get();
+        ports.add(executor.getHost() + ":" + executor.getPort());
+      }
     }
     return ports;
   }
@@ -505,9 +505,9 @@ public class ExecutorManager extends EventHandler implements
    * @see azkaban.executor.ExecutorManagerAdapter#getActiveFlowsWithExecutor()
    */
   @Override
-  public List<Pair<ExecutableFlow, Executor>> getActiveFlowsWithExecutor()
+  public List<Pair<ExecutableFlow, Optional<Executor>>> getActiveFlowsWithExecutor()
       throws IOException {
-    final List<Pair<ExecutableFlow, Executor>> flows =
+    final List<Pair<ExecutableFlow, Optional<Executor>>> flows =
         new ArrayList<>();
     getActiveFlowsWithExecutorHelper(flows, this.queuedFlows.getAllEntries());
     getActiveFlowsWithExecutorHelper(flows, this.runningFlows.values());
@@ -516,7 +516,7 @@ public class ExecutorManager extends EventHandler implements
 
   /* Helper method for getActiveFlowsWithExecutor */
   private void getActiveFlowsWithExecutorHelper(
-      final List<Pair<ExecutableFlow, Executor>> flows,
+      final List<Pair<ExecutableFlow, Optional<Executor>>> flows,
       final Collection<Pair<ExecutionReference, ExecutableFlow>> collection) {
     for (final Pair<ExecutionReference, ExecutableFlow> ref : collection) {
       flows.add(new Pair<>(ref.getSecond(), ref
@@ -1442,15 +1442,15 @@ public class ExecutorManager extends EventHandler implements
   }
 
   /* Group Executable flow by Executors to reduce number of REST calls */
-  private Map<Executor, List<ExecutableFlow>> getFlowToExecutorMap() {
-    final HashMap<Executor, List<ExecutableFlow>> exFlowMap =
+  private Map<Optional<Executor>, List<ExecutableFlow>> getFlowToExecutorMap() {
+    final HashMap<Optional<Executor>, List<ExecutableFlow>> exFlowMap =
         new HashMap<>();
 
     for (final Pair<ExecutionReference, ExecutableFlow> runningFlow : this.runningFlows
         .values()) {
       final ExecutionReference ref = runningFlow.getFirst();
       final ExecutableFlow flow = runningFlow.getSecond();
-      final Executor executor = ref.getExecutor();
+      final Optional<Executor> executor = ref.getExecutor();
 
       // We can set the next check time to prevent the checking of certain
       // flows.
@@ -1550,20 +1550,27 @@ public class ExecutorManager extends EventHandler implements
           ExecutorManager.this.lastThreadCheckTime = System.currentTimeMillis();
           ExecutorManager.this.updaterStage = "Starting update all flows.";
 
-          final Map<Executor, List<ExecutableFlow>> exFlowMap =
+          final Map<Optional<Executor>, List<ExecutableFlow>> exFlowMap =
               getFlowToExecutorMap();
-          final ArrayList<ExecutableFlow> finishedFlows =
-              new ArrayList<>();
           final ArrayList<ExecutableFlow> finalizeFlows =
               new ArrayList<>();
 
           if (exFlowMap.size() > 0) {
-            for (final Map.Entry<Executor, List<ExecutableFlow>> entry : exFlowMap
+            for (final Map.Entry<Optional<Executor>, List<ExecutableFlow>> entry : exFlowMap
                 .entrySet()) {
               final List<Long> updateTimesList = new ArrayList<>();
               final List<Integer> executionIdsList = new ArrayList<>();
 
-              final Executor executor = entry.getKey();
+              final Optional<Executor> executorOption = entry.getKey();
+              if (!executorOption.isPresent()) {
+                for (final ExecutableFlow flow : entry.getValue()) {
+                  logger.error("Finalizing execution " + flow.getExecutionId()
+                      + ". Executor id of this execution doesn't exist");
+                  finalizeFlows.add(flow);
+                }
+                continue;
+              }
+              final Executor executor = executorOption.get();
 
               ExecutorManager.this.updaterStage =
                   "Starting update flows on " + executor.getHost() + ":"
@@ -1627,7 +1634,6 @@ public class ExecutorManager extends EventHandler implements
                     ExecutorManager.this.updaterStage = "Updated flow " + flow.getExecutionId();
 
                     if (isFinished(flow)) {
-                      finishedFlows.add(flow);
                       finalizeFlows.add(flow);
                     }
                   } catch (final ExecutorManagerException e) {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1564,7 +1564,7 @@ public class ExecutorManager extends EventHandler implements
               final Optional<Executor> executorOption = entry.getKey();
               if (!executorOption.isPresent()) {
                 for (final ExecutableFlow flow : entry.getValue()) {
-                  logger.error("Finalizing execution " + flow.getExecutionId()
+                  logger.warn("Finalizing execution " + flow.getExecutionId()
                       + ". Executor id of this execution doesn't exist");
                   finalizeFlows.add(flow);
                 }
@@ -1613,7 +1613,7 @@ public class ExecutorManager extends EventHandler implements
                           + this.errorThreshold);
                       ref.setNumErrors(++numErrors);
                     } else {
-                      logger.error("Evicting flow " + flow.getExecutionId()
+                      logger.warn("Evicting execution " + flow.getExecutionId()
                           + ". The executor is unresponsive.");
                       // TODO should send out an unresponsive email here.
                       finalizeFlows.add(pair.getSecond());
@@ -1641,7 +1641,7 @@ public class ExecutorManager extends EventHandler implements
                     logger.error(e);
 
                     if (flow != null) {
-                      logger.error("Finalizing flow " + flow.getExecutionId());
+                      logger.warn("Finalizing execution " + flow.getExecutionId());
                       finalizeFlows.add(flow);
                     }
                   }
--- a/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
@@ -16,8 +16,8 @@
 
 package azkaban.executor;
 
-import azkaban.db.EncodingType;
 import azkaban.db.DatabaseOperator;
+import azkaban.db.EncodingType;
 import azkaban.utils.GZIPUtils;
 import azkaban.utils.Pair;
 import java.io.IOException;
@@ -29,9 +29,12 @@ import java.util.Map;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.dbutils.ResultSetHandler;
+import org.apache.log4j.Logger;
 
 @Singleton
 public class FetchActiveFlowDao {
+
+  private static final Logger logger = Logger.getLogger(FetchActiveFlowDao.class);
 
   private final DatabaseOperator dbOperator;
 
@@ -56,9 +59,9 @@ public class FetchActiveFlowDao {
     // Select running and executor assigned flows
     private static final String FETCH_ACTIVE_EXECUTABLE_FLOW =
         "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-            + "et.port port, et.id executorId, et.active executorStatus"
+            + "et.port port, ex.executor_id executorId, et.active executorStatus"
             + " FROM execution_flows ex"
-            + " INNER JOIN "
+            + " LEFT JOIN "
             + " executors et ON ex.executor_id = et.id"
             + " Where ex.status NOT IN ("
             + Status.SUCCEEDED.getNumVal() + ", "
@@ -91,8 +94,13 @@ public class FetchActiveFlowDao {
             final ExecutableFlow exFlow =
                 ExecutableFlow.createExecutableFlowFromObject(
                     GZIPUtils.transformBytesToObject(data, encType));
-
-            final Executor executor = new Executor(executorId, host, port, executorStatus);
+            final Executor executor;
+            if (host == null) {
+              logger.warn("Executor id " + executorId + " (on execution " + id + ") wasn't found");
+              executor = null;
+            } else {
+              executor = new Executor(executorId, host, port, executorStatus);
+            }
             final ExecutionReference ref = new ExecutionReference(id, executor);
             execFlows.put(id, new Pair<>(ref, exFlow));
           } catch (final IOException e) {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionReference.java
@@ -16,6 +16,9 @@
 
 package azkaban.executor;
 
+import java.util.Optional;
+import javax.annotation.Nullable;
+
 public class ExecutionReference {
 
   private final int execId;
@@ -30,11 +33,7 @@ public class ExecutionReference {
     this.execId = execId;
   }
 
-  public ExecutionReference(final int execId, final Executor executor) {
-    if (executor == null) {
-      throw new IllegalArgumentException(String.format(
-          "Executor cannot be null for exec id: %d ExecutionReference", execId));
-    }
+  public ExecutionReference(final int execId, @Nullable final Executor executor) {
     this.execId = execId;
     this.executor = executor;
   }
@@ -59,14 +58,6 @@ public class ExecutionReference {
     return this.execId;
   }
 
-  public String getHost() {
-    return this.executor.getHost();
-  }
-
-  public int getPort() {
-    return this.executor.getPort();
-  }
-
   public int getNumErrors() {
     return this.numErrors;
   }
@@ -75,11 +66,11 @@ public class ExecutionReference {
     this.numErrors = numErrors;
   }
 
-  public Executor getExecutor() {
-    return this.executor;
+  public Optional<Executor> getExecutor() {
+    return Optional.ofNullable(this.executor);
   }
 
-  public void setExecutor(final Executor executor) {
+  public void setExecutor(final @Nullable Executor executor) {
     this.executor = executor;
   }
 }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
@@ -46,14 +46,16 @@ public class ExecutorApiGateway {
 
   Map<String, Object> callWithReference(final ExecutionReference ref, final String action,
       final Pair<String, String>... params) throws ExecutorManagerException {
-    return callWithExecutionId(ref.getHost(), ref.getPort(), action, ref.getExecId(),
+    final Executor executor = ref.getExecutor().get();
+    return callWithExecutionId(executor.getHost(), executor.getPort(), action, ref.getExecId(),
         null, params);
   }
 
   Map<String, Object> callWithReferenceByUser(final ExecutionReference ref,
       final String action, final String user, final Pair<String, String>... params)
       throws ExecutorManagerException {
-    return callWithExecutionId(ref.getHost(), ref.getPort(), action,
+    final Executor executor = ref.getExecutor().get();
+    return callWithExecutionId(executor.getHost(), executor.getPort(), action,
         ref.getExecId(), user, params);
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManagerAdapter.java
@@ -25,6 +25,7 @@ import java.lang.Thread.State;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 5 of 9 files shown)


kunkun-tang:
I must say this is a very important fix!

juhoautio:
@kunkun-tang Executor info is not needed at the moment of finalizing a flow. But the initial loading also loads running executions that are periodically checked by calling the assigned executors. If an executor answers that such execution is not running any more, then the execution is finalized. This change also immediately finalizes executions whose assigned executor doesn’t exist.

chengren311:
@juhoautio i'm curious how did you encounter this issue since executor is not supposed to be stopped manually? The recommended way is to call shutdown api of the executor then it will go into the shutting down mode where it's no longer active to accept new flow execution,  all running flows will be killed in 10 days, and executor will shutdown itself when no flow is running on it.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1550,20 +1550,27 @@ public void run() {
...
+            for (final Map.Entry<Optional<Executor>, List<ExecutableFlow>> entry : exFlowMap
                 .entrySet()) {
               final List<Long> updateTimesList = new ArrayList<>();
               final List<Integer> executionIdsList = new ArrayList<>();
 
-              final Executor executor = entry.getKey();
+              final Optional<Executor> executorOption = entry.getKey();
+              if (!executorOption.isPresent()) {
+                for (final ExecutableFlow flow : entry.getValue()) {
+                  logger.error("Finalizing execution " + flow.getExecutionId()
```
chengren311: is this an error? should we consider info?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -56,9 +59,9 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
     // Select running and executor assigned flows
     private static final String FETCH_ACTIVE_EXECUTABLE_FLOW =
         "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-            + "et.port port, et.id executorId, et.active executorStatus"
+            + "et.port port, ex.executor_id executorId, et.active executorStatus"
             + " FROM execution_flows ex"
-            + " INNER JOIN "
+            + " LEFT JOIN "
```
chengren311: why is this needed?  can you give an example of wrong case which might be caused by original sql

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: yeah, let's use warn.

juhoautio:
@chengren311 I've made those changes to use warn level.

chengren311:
@juhoautio 
looks we have below error when running SOLO mode:
```
Exception in thread "main" com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) Error injecting constructor, java.lang.NullPointerException
  at azkaban.executor.ExecutorManager.<init>(ExecutorManager.java:125)
  at azkaban.executor.ExecutorManager.class(ExecutorManager.java:78)
  while locating azkaban.executor.ExecutorManager
    for the 3rd parameter of azkaban.webapp.AzkabanWebServer.<init>(AzkabanWebServer.java:165)
  at azkaban.webapp.AzkabanWebServer.class(AzkabanWebServer.java:122)
  while locating azkaban.webapp.AzkabanWebServer
    for the 1st parameter of azkaban.soloserver.AzkabanSingleServer.<init>(AzkabanSingleServer.java:48)
  while locating azkaban.soloserver.AzkabanSingleServer
Caused by: java.lang.NullPointerException
	at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1011)
	at java.util.concurrent.ConcurrentHashMap.putAll(ConcurrentHashMap.java:1084)
	at azkaban.executor.ExecutorManager.loadRunningFlows(ExecutorManager.java:450)
```

The NPE is thrown from 
```
  private void loadRunningFlows() throws ExecutorManagerException {
    this.runningFlows.putAll(this.executorLoader.fetchActiveFlows());
  }
```
when this.executorLoader.fetchActiveFlows returns a hashmap which contains an entry with null value. Do you think it's related to this PR? I tried to reverted this commit and the error won't appear.

juhoautio:
@chengren311 Ok, I didn't have such data in the DB and didn't expect it either. It's probably some old left-over data. This is the sequence that's setting the null value in the returned map:

```java
        final byte[] data = rs.getBytes(3);
```
```java
        if (data == null) {
          execFlows.put(id, null);
        } else {
```

So it seems like you have an unfinished execution row in your DB that has a `null` value for the `data` field, but its executor id is not found in the `executors` table any more.

I will fix that NPE.

You would get the same NPE also _without_ my changes, if you would "add back" the executor id of that execution into the `executors` table.

chengren311:
@juhoautio , seems something is still off. This is what we observed in solo mode. 
```
2018/07/23 12:58:52.359 -0700 ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException
2018/07/23 12:58:52.359 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2624. Executor id of this execution doesn't exist
2018/07/23 12:58:52.359 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 1724. Executor id of this execution doesn't exist
2018/07/23 12:58:52.360 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2047. Executor id of this execution doesn't exist
2018/07/23 12:58:52.362 -0700 ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException
2018/07/23 12:58:52.363 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2624. Executor id of this execution doesn't exist
2018/07/23 12:58:52.363 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 1724. Executor id of this execution doesn't exist
2018/07/23 12:58:52.363 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2047. Executor id of this execution doesn't exist
2018/07/23 12:58:52.365 -0700 ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException
2018/07/23 12:58:52.365 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2624. Executor id of this execution doesn't exist
2018/07/23 12:58:52.365 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 1724. Executor id of this execution doesn't exist
2018/07/23 12:58:52.366 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2047. Executor id of this execution doesn't exist
2018/07/23 12:58:52.368 -0700 ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException
2018/07/23 12:58:52.368 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2624. Executor id of this execution doesn't exist
2018/07/23 12:58:52.368 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 1724. Executor id of this execution doesn't exist
2018/07/23 12:58:52.368 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2047. Executor id of this execution doesn't exist
2018/07/23 12:58:52.371 -0700 ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException
```

juhoautio:
@chengren311 would you be able to get some additional details? For example, the results of running the sql `FetchActiveExecutableFlows.FETCH_ACTIVE_EXECUTABLE_FLOW` against that DB.

chengren311:
@juhoautio 
```mysql> SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, et.port port, ex.executor_id executorId, et.active executorStatus FROM execution_flows ex LEFT JOIN  executors et ON ex.executor_id = et.id Where ex.status NOT IN (50, 60, 70);
+---------+----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+------+------------+----------------+
| exec_id | enc_type | flow_data                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   | host | port | executorId | executorStatus |
+---------+----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+------+------------+----------------+
|   12820 |     NULL | NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | NULL | NULL |       NULL |           NULL |
|    1724 |        2 |        ?TMs?0?/{vS?\?o
                                             ??3??k݈ؒ??4??W??1?䐋ǫ]?{ow?{H???M???vo?t??^?k?? ?16?	>??WA?????H????Qx?c=Z???h1?_ϗQF???
                                          3!w??????<?Y?)?8?@!???4? <6R"?%?Ea??hsr?Y???	?JOJ.$$UxrNɵ??:??H??1*5?l??=???????~?<?L?M????T??%?d?Q??.R?a?w$?h!??	1?????6??)?X7?K??????x?w9V?֘庨?͡???ͻ???ֶ???wzЈp?wg???=?R?w???َd??7?
                                ?f6ă????ᯐ?*}cO?Oi"u??S?\??m?qL8=yKq????\?ё????"?,=?)?
dml???o?p?>????ZT?????:??$????6??TՏK}]
?&cWL?~?r-?I$/?dC????                 X\
8?v??5V??3)*%??p?o???XG?T?zK2?^?{D??@????????s??ފA??j??Z?v??W?x??.h?                                                                                                                                                                                                                                                                                  | NULL | NULL |       NULL |           NULL |
|    2047 |        2 |        ?TKs?0?/{vSpy????2CM??!?aa?%??ꑆ2??J??W??ﱒ???6?$M?9??f??
??^?&?0)???j???ƭ??į@?q<~??,f?a<?<?f?8?? ?
                                         3?6?L? 2?b ?,Z??;r?@?r?L???)D?r@"Eb?Ba
\??T?z
iX???Sڌ
.??k<&'?2WB?
            $m??f?XO?Q)F?[R?,D??P?ɚ#-????D?
*???eQ?
?[??s?hu?V??k?A???r%?GiU??D?L?;???0W??\W??0????5Gm?\?UhC?)?j???1??;=+A?5?PGK?p*???????ʡa"(,?,?uu0}?]??j?/e??????&xM|??u?*[?⨰?zҚ[??r?]???Ywz?n??ej?џV?獬??)???g??w?b??? ?I???;??????????\/?ߴ???r???f???c/?$R@??\?> ??Sw~??0                                                                                                                                                                                                                                                                      | NULL | NULL |       NULL |           NULL |
|    2624 |        2 |        ?TKs?0?/{vSp\????z??
???+al޽?ⱴ????p??I
??q??.?;??(?҇???|:f???r:ϲ4??XJ?<c???&Ť?_
?Q?e??ߐC? *V!g?(??y ??U
??q
{R??)?a?z"FLi3??@R?q??X?????s?zX?d=yC???I????\@?&+??^\??'??hPyw6'%?4?;PH?TN
łXn??	??d+J
             ????q'??N?|3?
,+?????[?????p۵??a?z????Y@?d? ????HYJq??P?Rx??6?
?O??Z3????\?6D??????f???g%?>??h?
??2?????X:?#L?Ŗ??.?û???	?O?A?s?m?????K??_?!o??V??WOZs?Xl?+vU?Sַq?ׄ??????dE?U?]74?g??w??k??? ?I???;????d?'4????$^??i??y???h7?????????!????`??`??/????0                                                                                                                                                                                                                                                                                       | NULL | NULL |       NULL |           NULL |
+---------+----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+------+------------+----------------+
4 rows in set (0.00 sec)```

**Comment by juhoautio** [comment]
Thanks @chengren311, unfortunately it doesn't really help. The flow_data is in bytes, so I would need it in a different format to be able to parse it into an ExecutableFlow object.

But let's debug this way instead: could you merge or apply this patch https://github.com/azkaban/azkaban/pull/1867, test again, and report the stacktrace, please?

**Comment by jamiesjc** [comment]
Hi @juhoautio, here is the stacktrace after applying your new patch. Hopefully, it would help.

```
2018/07/25 12:56:52.853 -0700 ERROR [ExecutorManager] [Azkaban] Unexpected exception in updating executions
	at azkaban.executor.ExecutableNode.fillMapFromExecutable(ExecutableNode.java:297)
	at azkaban.executor.ExecutableFlowBase.fillMapFromExecutable(ExecutableFlowBase.java:219)
	at azkaban.executor.ExecutableFlow.toObject(ExecutableFlow.java:226)
	at azkaban.executor.ExecutionFlowDao.updateExecutableFlow(ExecutionFlowDao.java:230)
	at azkaban.executor.ExecutionFlowDao.updateExecutableFlow(ExecutionFlowDao.java:220)
	at azkaban.executor.JdbcExecutorLoader.updateExecutableFlow(JdbcExecutorLoader.java:73)
	at azkaban.executor.ExecutorManager.finalizeFlows(ExecutorManager.java:1245)
	at azkaban.executor.ExecutorManager.access$900(ExecutorManager.java:78)
	at azkaban.executor.ExecutorManager$ExecutingManagerUpdaterThread.run(ExecutorManager.java:1660)
2018/07/25 12:56:52.856 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2624. Executor id of this execution doesn't exist
2018/07/25 12:56:52.856 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 1724. Executor id of this execution doesn't exist
2018/07/25 12:56:52.856 -0700 WARN [ExecutorManager] [Azkaban] Finalizing execution 2047. Executor id of this execution doesn't exist
```

juhoautio:
@chengren311 great, fix candidate here: #1868.

Could you test again with both that and #1867, please?

jamiesjc:
Hi @juhoautio, I have a question regarding this PR. 
When an executor is shut down and removed from the DB table, the running flow cache could still remain the old executor info. It would skip the `!executorOption.isPresent()` check and not be added to `finalizeFlows`. The web server would still call that executor to fetch the flow status but it would fail to get the update from the executor since it's already shut down. As a result, an email will be sent instead of flow being finalized. 
We recently observed such behavior on our server. I think it might require a web server restart so that the running flows cache could be refreshed and then this PR would take effect. 
Can you share your thoughts and let me know if I miss anything?

jamiesjc:
@juhoautio Ok thanks. It would be more convenient if we don't have to restart the web server for it to take effect. Currently we have shutdown api which would shut down an executor when all the flows have finished on that executor. However it seems there is some timing issue that the executor would be shut down early even before the web server could get the chance to update the flow status. I think we need a fix either for shutdown api @chengren311  or finalizing flows without matching executor.

juhoautio:
@jamiesjc would you have more details on that timing issue? I thought that azkaban-web wouldn't keep asking for updates from executor as long as executor finalizes the status for all its executions before shutting down. If this is not the case, azkaban-web should be changed to act like that.

jamiesjc:
@juhoautio From what we observed in the log, the executor was shut down first and then the web server still tried to connect to that executor to get an update and it failed. They happened almost at the same time, but the shutdown happened a little early before the flow status was updated, thus causing the issue.

Server logs:
executor shutdown: 
ltx1-***exec14.grid.linkedin.com:
`azkaban-execserver.log:2018/09/29 08:22:54.130 +0000 WARN [FlowRunnerManager] [Azkaban] Shutting down FlowRunnerManager now...`

web server get update from executor:
`2018/09/29 08:22:54.565 +0000 ERROR [ExecutorManager] [Azkaban] Failed to get update from executor ltx1-***exec14.grid.linkedin.com`

jamiesjc:
@juhoautio Yes, the error log is irrelevant to your change. The concern is about the timing issue between shut down api action and finalizing flow status from web server. I would prefer fixing this issue since this is the root cause. One possible solution I can think of is to shut down the executor when there is no flow in the recently finished flow cache. All the finished flows will be moved from running flows cache to recently finished flow cache on executor and the recently finished flow cache will be cleaned up after 1~2 min. The web server should still have sufficient time to get flow status update from executor before the executor is shut down. Could this prevent the timing issue mentioned above? @chengren311 Let us know if you have any better suggestion on the shut down api behavior.
@juhoautio For your proposed solution, does it mean the running flows cache on web server will be updated periodically as well? That could be a solution, but I'm not sure how much complexity it would add to the existing updater thread on web server side which is already complicated.

juhoautio:
> Yes, the error log is irrelevant to your change. The concern is about the timing issue between shut down api action and finalizing flow status from web server. I would prefer fixing this issue since this is the root cause.

To clarify, is the problem just that an error is logged? You wouldn't want any error logged on azkaban-web's side when executor is shut down gracefully? Indeed that can be avoided, but note that error _should_ be logged if the executor is still found in the DB but is not responding.

> One possible solution I can think of is to shut down the executor when there is no flow in the recently finished flow cache. All the finished flows will be moved from running flows cache to recently finished flow cache on executor and the recently finished flow cache will be cleaned up after 1~2 min. The web server should still have sufficient time to get flow status update from executor before the executor is shut down. Could this prevent the timing issue mentioned above?

In my opinion that's not so nice way to handle it. Because it adds extra delay in executor shutdown and depends on timing to match on executor & web.
 
I think a better option would be to always reload executors from the DB before the update loop. Also, if an executor still cannot be reached, check the list of executors immediately again. If the executor was removed, finalize the execution as usual. If the executor has not been removed, then it's worth an error to be logged etc.

> @juhoautio For your proposed solution, does it mean the running flows cache on web server will be updated periodically as well?

Not needed. The current way should work.

You can check `ExecutorManager#finalizeFlows` – it takes care of removing executions from `ExecutorManager.runningFlows`. Especially if a flow is finalized in a situation where the flow's locally cached status is RUNNING, `finalizeFlows` loads the actual flow status from the DB to check the actual status, and only marks it as FAILED if `!isFinished(dsFlow)` (where `dsFlow` is the up-to-date execution info from the DB).

jamiesjc:
@juhoautio 

> To clarify, is the problem just that an error is logged? 

No, the main problem is that the flow status is inconsistent now. They are showing up on running flows page but are actually finished already. Users won't be able to rerun the same flow if they choose "skip" execution option. 

> Indeed that can be avoided, but note that error should be logged if the executor is still found in the DB but is not responding.

In this case, the executor is not found in the DB because they are shut down by the shutdown api call and we are still seeing the error before we restart the web server.

> I think a better option would be to always reload executors from the DB before the update loop.
By reloading, you mean reloading all the executors, not just the active executors, right? Why is it necessary to alwasys reload executors? It seems you only need to fetch the executor info from DB when the executor cannot be reached?

> Also, if an executor still cannot be reached, check the list of executors immediately again. If the executor was removed, finalize the execution as usual. 

I understand that `ExecutorManager#finalizeFlows ` will remove executions from running flows cache. If the new change could make sure that the flows are properly finalized, then that should be good. 

One more thing I want to mention, we also have the plan to totally remove running flows cache from web server, and to make the updater thread much cleaner and easier to maintain in the future.

juhoautio:
> No, the main problem is that the flow status is inconsistent now. They are showing up on running flows page but are actually finished already.

Actually finished in what sense? Do you mean that they have some finished status in the DB, but azkaban-web still thinks they're running? If so, that's a corner case that is more about dispatching than about getting updates, and should be handled separately. I mean, when checking for the concurrent option, the status of any blocking executions should be checked from the DB, unless that's already the case.

> In this case, the executor is not found in the DB because they are shut down by the shutdown api call and we are still seeing the error before we restart the web server.

Ok, this won't be an issue after the fix I proposed (refresh active executors from the DB before treating as an error).

> By reloading, you mean reloading all the executors, not just the active executors, right?

Yes.

> Why is it necessary to always reload executors? It seems you only need to fetch the executor info from DB when the executor cannot be reached?

It's not a must, but:

1. This way new executors are automatically taken into use without having to restart azkaban-web
   - But is it possible that people would not want this to happen in some cases? If so, it would be better to change executors so that they don't automatically activate themselves, because there could be other reasons to restart azkaban-web while not being aware that there are new executors in the DB that shouldn't be taken into use yet.
2. I thought it would be better to avoid making a failed / timed out call on a removed executor.
   - This is a small optimization, but basically redundant (as you pointed out) – the code has to handle the failure case any way.

The executors table is about as small as the number of executor nodes though, so it shouldn't be a problem to read it often.

----

How about I make a PR with this plan:

- Refresh executors from the DB before each update round
- If calling an executor fails (that is, no JSON can be received at all, not even JSON with the "error" field in it):
   - Refresh executors from the DB
   - If the executor had been removed, finalize the execution
   - If the executor is still in the DB, do as before ie. increment failed update count & eventually notify about it with alerter according to the alerting threshold

jamiesjc:
@juhoautio 

> Actually finished in what sense? Do you mean that they have some finished status in the DB, but azkaban-web still thinks they're running? If so, that's a corner case that is more about dispatching than about getting updates, and should be handled separately.

Yes, they have finished status in DB but on the "Currently Running" executing flows page, they show up as Running status. And when you click the execution, it shows as finished status. The root cause of this inconsistency is the getting updates. Even if users do not submit the flows again, the inconsistency issue itself still exists, visible from the UI page.

> This way new executors are automatically taken into use without having to restart azkaban-web
> But is it possible that people would not want this to happen in some cases? If so, it would be better to change executors so that they don't automatically activate themselves, because there could be other reasons to restart azkaban-web while not being aware that there are new executors in the DB that shouldn't be taken into use yet.

Right, for example, when we do deployment, we have the last step to reload executors from DB (update `activeExecutors` cache), but only after the deployment is successful and integration tests run successfully. I think refreshing executors should not happen automatically.

juhoautio:
@jamiesjc thanks for your reply.

> Yes, they have finished status in DB but on the "Currently Running" executing flows page, they show up as Running status. And when you click the execution, it shows as finished status. The root cause of this inconsistency is the getting updates. Even if users do not submit the flows again, the inconsistency issue itself still exists, visible from the UI page.

It's natural though, that sometimes an execution finishes just after you clicked it, even if it was actually running. But if there's too much delay currently (by the updater thread) maybe the page should be optimized to get the running executions directly from the DB somehow. Still, I see this as a separate issue.

> when we do deployment, we have the last step to reload executors from DB (update activeExecutors cache), but only after the deployment is successful and integration tests run successfully. I think refreshing executors should not happen automatically.

Ok but wouldn't it be safer then that executors wouldn't automatically add themselves to the DB? Or they could add but as inactive only. Otherwise it could happen that azkaban-web is restarted before a new executor has been validated and it gets added to active executors for dispatching. Obviously if you control azkaban-web it most likely won't be restarted during executor deployment :) Any way I think it would be a cleaner concept to require enabling executors explicitly if you want to do some checks first.

In your case, instead of restarting azkaban-web your last step would be to activate the new executor. There's an API for that already, to do it in a clean way.

The executor behavior should be configurable so that azkaban would also allow adding executors just by creating them without having to restart azkaban-web or call the `activate` API, if simple is what the user wants.

One thing in favor of live reloading executors is that azkaban-web doesn't support HA. This change would remove azkaban-web downtime during executor deployments.

jamiesjc:
Hi @juhoautio 

> But if there's too much delay currently (by the updater thread) maybe the page should be optimized to get the running executions directly from the DB somehow. 

This is not just a matter of delay, the inconsistency of flow status remains all the time until the web server is restarted. This is also part of the reason why we prefer to totally remove the running flows cache and instead, getting it from DB directly. Anyway, this would require some changes in the updater thread.

`Ok but wouldn't it be safer then that executors wouldn't automatically add themselves to the DB? Or they could add but as inactive only.`

True. I might not have explained it clearly. Actually when a new executor is deployed, we do the validation first and then set the executor to active, and finally reload the executors. It doesn't require restarting of web server.

We'll let you decide how to design and implement the changes since you already have much knowledge about it. It would be great if you could figure out a proper way to test the changes as well.  Thanks!

juhoautio:
> This is not just a matter of delay, the inconsistency of flow status remains all the time until the web server is restarted

Ok, I thought you were talking about the normal situation, but you actually meant the situation where  an executor has been removed but azkaban-web doesn't yet know it. That will be fixed by the live-updating of executors by azkaban-web.

> Actually when a new executor is deployed, we do the validation first and then set the executor to active

Wow, we had missed this change:
https://github.com/azkaban/azkaban/commit/309a53e87eb5536531b5bce0b08a49c70fc9c5d9#diff-6bba907bedf62d34e7421866572b7f4d

Thus, in our DB the default is still active=TRUE, and thus any new executor automatically sets itself as active. We must change our internal deployment model to be similar to yours, ie.
- change the default of executors.active to FASLE
- validate the executor first
- set executor as active
- bonus: don't restart azkaban-web, call `reloadExecutors` instead (there seems to be an action `reloadExecutors` on azkaban-web's ExecutorServlet. I haven't tried it yet)

So that part indeed doesn't need any changes in Azkaban.

Will proceed with the plan outlined on the bottom of: https://github.com/azkaban/azkaban/pull/1833#issuecomment-427935427.

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
jamiesjc: Hi @juhoautio, I have two questions regarding this change:
1. `LEFT JOIN` will also include executions that are in `PREPARING` status and haven't been dispatched and assigned an executor yet. When a web server is restarted, it will call `loadRunningExecutions` and `loadQueuedFlows`. The flows in `PREPARING` status will be included in both `runningExecutions` and `queuedFlows`. They will be finalized in `updateExecutions` because the executor is null. They will again be dispatched by `queueProcessor` as well. We've noticed one incident from our current web server log. Although it doesn't impact users too much, I think it's still an issue since the `runningExecutions` definition doesn't quite match what is fetched from DB.
2. For those executions that have been assigned an executor id but the `executor_id` is removed from executor table, this query will return the `executor_id` from `execution_flows` table, which is not null. Can the execution still be finalized in `updateExecutions`?

jamiesjc: @juhoautio Thanks, the solution sounds good to me.

juhoautio: @jamiesjc were you expecting me to make a PR for this? If yes, let me know and I'll do it.

jamiesjc: Sure, please go ahead with the fix. I'll review and merge it.

PR #1839, opened 2018-07-09
Add a script to create a new release tag

HappyRay:
This change will pave the way for adding more python based tools.

Future work:
* Run the python tests automatically as other unit tests.
* Separate test code and production code into separate directories.


kunkun-tang:
Don't fully understand
> Separate test code and production code into separate directories.

What's that for?

HappyRay:
As an example, the current Java test code is in a separate directory. Similar patterns should help with python too.

PR #1840, opened 2018-07-09
Add a test to run a v2 flow file with the new DAG engine

HappyRay:
This is a step towards building an integration test for the new DAG engine.

The jobs runs are skipped in this test.

Refactored the DagBuilder API to make it easier to use.
Instead of using a DagBuilder class to link nodes, use the name of the
nodes directly.

This facility can also be used in the future to build the tools to run
flows locally for testing purposes.


azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -35,39 +35,83 @@
...
+  /**
+   * A builder for building a DAG.
+   *
+   * @param name name of the DAG
+   * @param dagProcessor the associated DagProcessor
+   */
   public DagBuilder(final String name, final DagProcessor dagProcessor) {
     requireNonNull(name, "The name of the DagBuilder can't be null");
-    this.name = name;
     requireNonNull(name, "The dagProcessor of the DagBuilder can't be null");
```
jamiesjc: Should this be `requireNonNull(dagProcessor, ` ?

HappyRay: Fixed. Thanks

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
```diff
@@ -98,16 +141,16 @@ private void checkCircularDependencies() {
     class CircularDependencyChecker {
```
jamiesjc: The dependency check has already been done when loading the flow files: azkaban.project.DirectoryYamlFlowLoader#buildFlowEdges
Is there any particular reason why this needs to be implemented again in building the DAG?

jamiesjc:
The test cases look good, but I don't see it's using Flow 2.0 yaml files? How would it be integrated with flow 2.0?

azkaban-exec-server/src/main/java/azkaban/dag/DagBuilder.java
HappyRay: The DAG component is designed with minimal assumptions including the validations done to its inputs prior to passing these inputs to the DAG engine.

This allows this component to be used in more scenarios. e.g. the test included in this PR doesn't have a need to convert the v2 flow into Flow object which is where the azkaban.project.DirectoryYamlFlowLoader#buildFlowEdges is called. 

In the future, we can consider getting rid of one of the validations in the AZ loading scenario.  For the time being, the cost of double validation seems small.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunner2Test.java
```diff
@@ -0,0 +1,161 @@
...
+    private final NodeBean flowNode;
+    private final DagBuilder dagBuilder;
+
+    DagCreator(final NodeBean flowNode) {
+      final String flowName = flowNode.getName();
+      this.flowNode = flowNode;
+      this.dagBuilder = new DagBuilder(flowName, new SimpleDagProcessor());
+    }
+
+    Dag create() {
```
jamiesjc: It seems to me that logic of creating Dag is similar to what has been done in DirectoryYamlFlowLoader. Would it be easier to just reuse the existing part? Do you plan to eventually integrate them together or keep them as separate paths?

HappyRay: I haven't looked at the DirectoryYamlFlowLoader code in detail.

Eventually, yes, we will need to handle running both v1 and v2 flows. Will look into reducing logic duplication then.

PR #1848, opened 2018-07-11
project LRU cache part 1

chengren311:
The problem this PR targets to solve is detailed in https://github.com/azkaban/azkaban/pull/1803 

Drawback of previous design is detailed in https://github.com/azkaban/azkaban/pull/1841

New design:
1. Create a file in each project directory and write the size of the project to the file when the project is created. The project files are not supposed to change after creation. Touch this file each time the project is used. This way, we can have a more efficient LRU algorithm based on last access time, not creation time.
2. Maintain the total size of the project cache in memory to avoid the overhead of re-calculating it. The size shouldn't change too often. This way we can afford to run the check more frequently.

This PR implements part 1.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -89,6 +89,8 @@ public class Constants {
   // The flow exec id for a flow trigger instance unable to trigger a flow yet
   public static final int FAILED_EXEC_ID = -2;
 
+  // Name of the file which keeps project directory size
+  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_project_dir_size_";
 
   public static class ConfigurationKeys {
 
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -90,7 +90,7 @@ public class Constants {
   public static final int FAILED_EXEC_ID = -2;
 
   // Name of the file which keeps project directory size
-  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_azkaban_project_dir_size_in_bytes_";
+  public static final String PROJECT_DIR_SIZE_FILE_NAME = "___azkaban_project_dir_size_in_bytes___";
 
   public static class ConfigurationKeys {
 
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -90,7 +90,7 @@ public class Constants {
   public static final int FAILED_EXEC_ID = -2;
 
   // Name of the file which keeps project directory size
-  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_azkaban_project_dir_size_";
+  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_azkaban_project_dir_size_in_bytes_";
 
   public static class ConfigurationKeys {
 
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -89,9 +89,6 @@ public class Constants {
   // The flow exec id for a flow trigger instance unable to trigger a flow yet
   public static final int FAILED_EXEC_ID = -2;
 
-  // Name of the file which keeps project directory size
-  public static final String PROJECT_DIR_SIZE_FILE_NAME = "___azkaban_project_dir_size_in_bytes___";
-
   public static class ConfigurationKeys {
 
     // Configures Azkaban Flow Version in project YAML file
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -90,7 +90,7 @@ public class Constants {
   public static final int FAILED_EXEC_ID = -2;
 
   // Name of the file which keeps project directory size
-  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_project_dir_size_";
+  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_azkaban_project_dir_size_";
 
   public static class ConfigurationKeys {
 
--- a/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
@@ -32,6 +32,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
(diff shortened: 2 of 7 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -89,6 +89,8 @@
   // The flow exec id for a flow trigger instance unable to trigger a flow yet
   public static final int FAILED_EXEC_ID = -2;
 
+  // Name of the file which keeps project directory size
+  public static final String PROJECT_DIR_SIZE_FILE_NAME = "_azkaban_project_dir_size_in_bytes_";
```
HappyRay: Make it "___*" three underscores to make it even less likely to have a collision?  I assume it will be hard to change the file name once this change is deployed...

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -69,6 +71,51 @@ public static boolean isDirWritable(final File dir) {
...
+  }
+
+  /**
+   * Dumps a number into a new file.
+   *
+   * @param filePath the target file
+   * @param num the number to dump
+   * @throws IOException if file already exists
+   */
+  public static void dumpNumberToFile(final String filePath, final long num) throws IOException {
```
HappyRay: Have you considered accepting a https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html instead? You can benefit from stronger type checking.

HappyRay: And it will be easier to unit test.

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -69,6 +70,43 @@ public static boolean isDirWritable(final File dir) {
...
+    try (BufferedWriter writer = Files
+        .newBufferedWriter(Paths.get(filePath), StandardCharsets.UTF_8)) {
+      writer.write(String.valueOf(num));
+      writer.write("\n");
```
HappyRay: Is this necessary?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -107,6 +121,9 @@ void setupProject(final ProjectVersion pv)
     // If directory exists. Assume its prepared and skip.
     if (pv.getInstalledDir().exists()) {
       log.info("Project already cached. Skipping download. " + pv);
+      touchIfExists(
+          new File(Paths.get(pv.getInstalledDir().getPath(), Constants.PROJECT_DIR_SIZE_FILE_NAME)
```
HappyRay: Why not use Path as the argument to touchIfExists? File is a legacy class. Path is preferred now if I remember correctly.

HappyRay: https://gquintana.github.io/2017/09/02/Java-File-vs-Path.html

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
```diff
@@ -77,10 +81,32 @@ public void testSetupProject() throws Exception {
...
 
+  @Test
+  public void testSetupProjectTouchesTheDirSizeFile() throws Exception {
+    //verifies setup project updates last modified time of project dir size file.
+    final ProjectVersion pv = new ProjectVersion(12, 34,
+        new File(this.projectsDir, "sample_project_01"));
+    this.instance.setupProject(pv);
+    final long lastModifiedTime1 = new File(Paths.get(pv.getInstalledDir().getPath(), Constants
+        .PROJECT_DIR_SIZE_FILE_NAME).toString()).lastModified();
+    Thread.sleep(1000);
```
HappyRay: Is there a better way?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
chengren311: good point, updated. thanks

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: removed

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: updated, thanks

az-core/src/main/java/azkaban/Constants.java
chengren311: updated, thanks

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -89,6 +89,8 @@
   // The flow exec id for a flow trigger instance unable to trigger a flow yet
   public static final int FAILED_EXEC_ID = -2;
 
+  // Name of the file which keeps project directory size
```
HappyRay: Since this constant is an implementation detail rather than a user-facing API, is it better to define it in a place where it is used?

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
```diff
@@ -69,6 +71,42 @@ public static boolean isDirWritable(final File dir) {
...
+  }
+
+  /**
+   * Reads a number from a file.
+   *
+   * @param filePath the target file
+   */
+  public static long readNumberFromFile(final Path filePath)
+      throws IOException, NumberFormatException {
+    try (BufferedReader reader = Files
```
HappyRay: Have you considered https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#readAllLines(java.nio.file.Path,%20java.nio.charset.Charset)

HappyRay:
How do you plan to generate the files for existing cache directories?

az-core/src/main/java/azkaban/Constants.java
chengren311: according to the doc of this class, ```Constants used in configuration files or shared among classes``` the variable it could keep doesn't necessarily need to be a user-facing API.  
Even if the variable is defined in the class where it's used, associated test class  still needs to access it.

chengren311:
@HappyRay plan to run a util program to generate the files before enabling this feature.

az-core/src/main/java/azkaban/Constants.java
HappyRay: It's generally a good practice to reduce the scope of a concept.  For example, if a constant is only used in a class, define it for that scope. The test classes can access it via FooClass.BAR if needed. 

We don't have a good way to organize our user-facing key names etc yet, the current arrangement for this central class is a compromise until we find a better solution. 

Make sense?

azkaban-common/src/main/java/azkaban/utils/FileIOUtils.java
chengren311: updated

az-core/src/main/java/azkaban/Constants.java
chengren311: updated, thanks.

chengren311:
@HappyRay updated the PR, any more feedback?

HappyRay:
I saw "Thread.sleep(1000);" is still there. I don't see a response to my earlier comment regarding this. 
Is the sleep necessary? Unit tests should be as fast as possible.

Please ask another member for a review too.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
chengren311: updated, thanks.

chengren311:
@HappyRay updated.

Issue #1826, opened 2018-06-27
Conditional workflow design


jamiesjc:
Problem Statement:
Currently, Azkaban does not support job execution based on conditions. 
1. Some users want to run or disable certain jobs at runtime based on the current date or the current cluster name. It’s not feasible with current Azkaban execution logic. 
2. Also, the current dependency logic only supports “AND” operation, meaning the children job will only run when all its parent jobs are successful. In many use cases, “OR” operation is also desirable. Users may want to run the children job when any of its parent jobs is successful. 
3. Branching is not supported currently. Sometimes users want to proceed with one branch when the parent job succeeds and the other branch when its parent job fails. See user request for branching a workflow.

Proposed Design:
We mainly want to support two types of condition for jobs:
1. Condition on runtime parameters
    Users can decide whether to run or cancel current job based on the runtime variable values generated in the previous jobs and also some common job/flow/cluster related properties. 
2. Condition on job status
    Users can use predefined macros to decide whether to run or cancel current job based on the status of parent jobs. For example, they will be able to run the current job if one of the parent jobs has succeeded simply by applying the macro: one_success in the condition.

Note: This new feature will be developed on top of Flow 2.0 design. Users need to switch to YAML format to try this new feature.

PR #1854, opened 2018-07-16
Condition on job status - Step 1

jamiesjc:
Users can use predefined macros to specify the condition on job status. For example, they can write `condition: one_success` for a job in the flow yaml file. And this job will run as soon as one of its parent jobs has succeeded. 
Below is the list of condition macros we will support:

- **all_success** (default)
- **all_failed**
- **all_done**
- **one_failed**: fire as soon as at least one parent has failed.
- **one_success**: fire as soon as at least one parent has succeeded.
- **one_failed_all_done**: fire when at least one parent has failed and all parents have finished.
- **one_success_all_done**: fire when at least one parent has succeeded and all parents have finished.

By default, the condition on job status is all_success.
The execution logic will change under different conditions. For example, the `flowFailed `flag might not always be set to true when a job fails for a conditional workflow. 
Users can define both condition on runtime parameters and job status. E.g., `condition: one_success && ${jobA: param1} == 1  `

Please see the design doc for more details: #1826

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
@@ -16,6 +16,7 @@
 
 package azkaban.executor;
 
+import azkaban.flow.ConditionOnJobStatus;
 import azkaban.flow.Node;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
@@ -44,6 +45,7 @@ public class ExecutableNode {
   public static final String OUTNODES_PARAM = "outNodes";
   public static final String TYPE_PARAM = "type";
   public static final String CONDITION_PARAM = "condition";
+  public static final String CONDITION_ON_JOB_STATUS_PARAM = "conditionOnJobStatus";
   public static final String PROPS_SOURCE_PARAM = "propSource";
   public static final String JOB_SOURCE_PARAM = "jobSource";
   public static final String OUTPUT_PROPS_PARAM = "outputProps";
@@ -68,6 +70,7 @@ public class ExecutableNode {
   private long delayExecution = 0;
   private ArrayList<ExecutionAttempt> pastAttempts = null;
   private String condition;
+  private ConditionOnJobStatus conditionOnJobStatus = ConditionOnJobStatus.ALL_SUCCESS;
 
   // Transient. These values aren't saved, but rediscovered.
   private ExecutableFlowBase parentFlow;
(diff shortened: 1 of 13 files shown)


azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -217,4 +222,22 @@ private void addEdges(final AzkabanNode node, final AzkabanFlow azkabanFlow,
     }
   }
 
+  private void setConditionOnJobStatus(final Node node) {
+    String condition = node.getCondition();
+    if (condition != null) {
+      final String patternString =
```
kunkun-tang: Maybe add a comment to show what pattern should be correct, and what are wrong.

jamiesjc:
I added some Todo notes in the code about the remaining steps:
1. Cancel all pending jobs before the flow is finalized and execturorService is shutdown. This is mainly to support ONE_SUCCESS and ONE_FAILURE conditions because the flow could potentially finish before all running jobs have finished. This would cause jobs to run forever if not canceled properly.  
2.  Support conditions on embedded flows.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Sure, will add the comment.

azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
```diff
@@ -324,6 +330,8 @@ public void fillExecutableFromMapObject(
     this.id = wrappedMap.getString(ID_PARAM);
     this.type = wrappedMap.getString(TYPE_PARAM);
     this.condition = wrappedMap.getString(CONDITION_PARAM);
+    this.conditionOnJobStatus = ConditionOnJobStatus.fromString(wrappedMap.getString
```
kunkun-tang: What if the condition is null? Should we handle the edge case?

jamiesjc:
Yes, UI part will be added later when all the main logic has been implemented. 
Users will be able to click the job and see the condition for that job like the current job properties.
You can see more details in the design doc: 
https://docs.google.com/document/d/1op_RPD1G65NpKsunAv04Rww5klU7AhbqnPMA0QunP0g/edit?usp=sharing

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -937,8 +963,16 @@ private String findValueForJobVariable(final String jobName, final String variab
       }
     }
 
+    // Todo jamiesjc: need to handle condition for embedded flows
     // Get job output props
-    final Props outputProps = this.getExecutableFlow().getExecutableNode(jobName).getOutputProps();
+    final ExecutableNode target = node.getParentFlow().getExecutableNode(jobName);
```
kunkun-tang: Assume every node have a parent flow? What does the code do here?

azkaban-common/src/main/java/azkaban/executor/ExecutableNode.java
jamiesjc: The condition can be null. If it's null, it just follows our current flow execution logic. 
However the conditionOnJobStatus is `all_success` by default, so it won't be null. If no condition is specified, by default we just follow the current flow execution logic which is `all_success`.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
jamiesjc: Yes, when an executableNode is created, azkaban.executor.ExecutableNode#setParentFlow will be called.
Instead of calling `this.getExecutableFlow()`, we should call `node.getParentFlow()` because for jobs inside an embedded flow, they should only have conditions on jobs that are also inside the same embedded flow. `this.getExecutableFlow()` would fetch the flow at root level whereas `node.getParentFlow()` would fetch the embedded flow.  Also I added a Null check for the target node. This part will be modified later to fully support condition for embedded flows.

PR #1855, opened 2018-07-17
make max wait min non-required when flow trigger's data dependency set is empty

chengren311:
When no dependency is defined for a flow trigger, it just works like a regular schedule - when the scheduled time comes, flow will be kicked off without waiting for any dependency. In this case, max wait min is not required.


azkaban-web-server/src/main/java/azkaban/webapp/servlet/FlowTriggerServlet.java
```diff
@@ -104,19 +110,17 @@ private void handleAJAXAction(final HttpServletRequest req,
...
-        }
-        else if (!hasPermission(project, session.getUser(), Type.ADMIN)) {
+        } else if (!hasPermission(project, session.getUser(), Type.ADMIN)) {
           ret.put("error", "Permission denied. Need ADMIN access.");
-        }
-        else {
+        } else {
           try {
-            if(ajaxName.equals("pauseTrigger")) {
+            if (ajaxName.equals("pauseTrigger")) {
```
chengren311: this is automatically done by save plugin.

PR #1856, opened 2018-07-17
Kafka Event trigger

weidious:
This PR is to demonstrate how we can achieve a new feature to trigger workflows by events.
A trigger instance will wait for all dependency instances are available so as to kick-off until reaching the maximum waiting time. This feature enables users to specify all the dependencies <Topic, Event> pair. By the Trigger.flow example provided below, the user scheduled a flow depends on both event with regular expression match of rules ".*" and  "hadoop?.*". All we can do is matching a Kafka event with Regex, however, one can implement an interface so as to extend this feature.
```
# Flow Trigger Example
trigger:
  maxWaitMins: 2880
  schedule:
    type: cron
    value: 0 0/2 * * * ?

  triggerDependencies:
    - name: dep1 # an unique name to identify the dependency
      type: kafka
      params:
        match: .*
        topic: AzEvent_Topic4
    - name: dep2 # an unique name to identify the dependency
      type: kafka
      params:
        match: hadoop?.*
        topic: AzEvent_Topic4


weidious:
Check out Event trigger PR #1858 for more information

PR #1861, opened 2018-07-18
Set up new Documentation toolset and publish it to readTheDocs

kunkun-tang:
1. This PR proposes a new toolset to do Azkaban documentation. We use Sphinx and ReStructuredText to write down docs, and publish it to readTheDocs. README is included to educate users how to develop the documentation.

2. Getting started section is rewritten and make sure it is up to date. 

See https://kunkun-azkaban.readthedocs.io/en/latest/getStarted.html for demo.


HappyRay:
Thanks for improving our documentation infrastructure! The demo doc looks good to me.
As discussed, consider having one shared python virtual environment i.e. requirements.txt at the root level?

PR #1862, opened 2018-07-19
Fix NPE in FetchActiveFlowDao

juhoautio:
Ignore active executions with flow_data=null.

For error details, see https://github.com/azkaban/azkaban/pull/1833#issuecomment-406108760.

Cleaning these up manually would require too many changes, because `ExecutableFlow` object is used in many places.

Note that it's still possible to end up having this kind of rows in the DB if the application fails between creation of an execution id in `execution_flows` and uploading the data. See here:

https://github.com/azkaban/azkaban/blob/2be81ebe5d0a931eee828ca968090c728735ae71/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java#L71-L74

I'm not sure if doing both inside a transaction is even possible, so I rather just avoid the NPE and log a warning about this corner case that should be rare.

This weakness was in place already before my change https://github.com/azkaban/azkaban/pull/1833. But before my change it could have only caused an NPE if the same executor id is still used after missing to update the flow data (meaning, using a fixed executor id, or starting up an executor after a sudden crash where the executor didn't get to remove itself from the DB as a part of the shutdown hook).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
@@ -20,6 +20,7 @@ import azkaban.db.DatabaseOperator;
 import azkaban.db.EncodingType;
 import azkaban.utils.GZIPUtils;
 import azkaban.utils.Pair;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -53,6 +54,7 @@ public class FetchActiveFlowDao {
     }
   }
 
+  @VisibleForTesting
   static class FetchActiveExecutableFlows implements
       ResultSetHandler<Map<Integer, Pair<ExecutionReference, ExecutableFlow>>> {
 
--- a/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
@@ -53,7 +53,7 @@ public class FetchActiveFlowDao {
     }
   }
 
-  private static class FetchActiveExecutableFlows implements
+  static class FetchActiveExecutableFlows implements
       ResultSetHandler<Map<Integer, Pair<ExecutionReference, ExecutableFlow>>> {
 
     // Select running and executor assigned flows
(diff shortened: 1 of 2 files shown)


juhoautio:
@chengren311 could you review, please?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -53,7 +53,7 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
     }
   }
 
-  private static class FetchActiveExecutableFlows implements
+  static class FetchActiveExecutableFlows implements
```
chengren311: consider adding @VisibleForTesting ?

juhoautio: Do you have some grounds for using `@VisibleForTesting`? Is it for example checked automatically in the build that it's respected ie. not accessed by other than test code?

Otherwise it seems like redundant documentation that could easily get outdated and thus isn't too useful. If the wider visibility of the method is later desired also for some usage in non-test code, one might easily miss removing the `@VisibleForTesting` annotation and just use the method.

It's easy enough to find references if you want to check where the method is actually used. And if the annotation can't be trusted to be up-to-date, you'd need to check the actual refs any way.

As long as there isn't static code analysis that makes the build fail if `@VisibleForTesting` is not respected, I wouldn't use the annotation at all.

Just my thoughts. If you want that annotation added to get this merged, I'd be happy to do it. Please let me know.

chengren311: Thanks for sharing your thought. 
No we don't have any build tool respecting the annotation, 
```The @VisibleForTesting annotation advertises to other developers (or reminds the code's author) when a decision was made for relaxed visibility to make testing possible or easier```
https://dzone.com/articles/two-generally-useful-guava

This annotation is for reminder purpose only, just like a line of comment, and the only reason here to relax the access level is for testing purpose, so I would imagine that the annotation would be helpful for other people who read this piece of code to understand why the inner class is not private, which worths explanation. I agree there's a maintenance cost as you said, but there's no other simpler alternative than it to let other developers know why we made such decision.

juhoautio: Ok, I will add `@VisibleForTesting`.

> there's no other simpler alternative than it to let other developers know why we made such decision

You can search for actual current references and deduce that the wider visibility is for tests (if it's still the case). Additionally, git history can be used to find out the exact (original) intention. These "built-in" ways are _not as fast_ as reading an annotation, but they can be _trusted_ because they never get outdated like a manually maintained explicit annotation might do. Which one is _simpler_, that's a matter of taste I guess :)

PR #1865, opened 2018-07-24
project LRU cache part 2

chengren311:
The problem this PR targets to solve is detailed in #1803

Drawback of previous design is detailed in #1841

New design:

1. Create a file in each project directory and write the size of the project to the file when the project is created. The project files are not supposed to change after creation. Touch this file each time the project is used. This way, we can have a more efficient LRU algorithm based on last access time, not creation time.

2. Maintain the total size of the project cache in memory to avoid the overhead of re-calculating it. The size shouldn't change too often. This way we can afford to run the check more frequently. Project dir size check and corresponding deletion will be performed when a new project is downloaded. 

This PR implements part 2.


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -776,6 +802,26 @@ public void deleteExecutionDirectory() {
...
+   * Checks if the project version contains any running flow
+   */
+  private boolean isActiveProject(final ProjectVersion version) {
+    final Pair<Integer, Integer> versionKey = new Pair<>(version.getProjectId(),
+        version.getVersion());
+    return getActiveProjectVersions().contains(versionKey);
+  }
+
+
   private class CleanerThread extends Thread {
```
kunkun-tang: Today LRU cache cannot work immediately on delete, because there's always a hardlink from execution to project directory.

Please add a todo comment to describe the current the situation.

chengren311: thanks, mentioned it in PR.

PR #1867, opened 2018-07-25
Include error stacktrace in updater thread logging

juhoautio:
It was calling `log.error(Object message)`, so we didn't get a stack trace at all. For example this was logged, which is not too helpful: `ERROR [ExecutorManager] [Azkaban] java.lang.NullPointerException` (the value of `e.toString()`, as you can see).


juhoautio:
@chengren311 to me it would seem sensible to merge this in. What do you say?

chengren311:
@juhoautio i think so, thanks!

PR #1870, opened 2018-07-26
Conditional workflow - Cancel remaining jobs before flow is finalized.

jamiesjc:
For conditional workflows, if conditionOnJobStatus is ONE_SUCCESS or ONE_FAILED, some jobs might still be running or not started yet when the end nodes have finished. In this case, we need to cancel all the remaining jobs before finalizing the flow.


jamiesjc:
Thanks. One thing to clarify, the end node doesn't have to define condition on job status. As long as there is node in the flow which has defined one_success or one_failed condition, this PR could apply.

jamiesjc:
This is part of the conditional workflow design - #1826

PR #1871, opened 2018-07-27
add dismiss button to banner

chengren311:
A dismiss button is added to banner when it's showing up. After the dismiss button is clicked, the banner will disappear until the browner restarts. 

image

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/nav.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/nav.vm
@@ -47,15 +47,15 @@
   }
 
   function dismissBanner() {
-    var dismissBanner = sessionStorage.getItem("dismissBanner");
+    var dismissBanner = localStorage.getItem("dismissBanner");
     if (dismissBanner != "true") {
-      sessionStorage.setItem("dismissBanner", "true");
+      localStorage.setItem("dismissBanner", "true");
     }
     $("#banner").slideUp();
   }
 
   $(function () {
-    var dismissBanner = sessionStorage.getItem("dismissBanner");
+    var dismissBanner = localStorage.getItem("dismissBanner");
     if (dismissBanner == "true") {
       return;
     }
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/nav.vm
```diff
@@ -32,13 +32,33 @@
...
+    color: #fff;
+  }
 </style>
 <script type="text/javascript">
   function navMenuClick(url) {
     window.location.href = url;
   }
 
+  function dismissBanner() {
+    var dismissBanner = sessionStorage.getItem("dismissBanner");
```
kunkun-tang: Have we considerred using `localStorage`? What's disadvantages of it over `sessionStorage`?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/notepage.vm
```diff
@@ -36,6 +36,12 @@
 </head>
 <body>
 
+<script type="text/javascript">
+  function enableBanner() {
+    sessionStorage.setItem("dismissBanner", null);
```
kunkun-tang: Should we use "false" rather than `null`, for better expression?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/nav.vm
chengren311: key stored in localStorage never expires until explicitly cleaning it. SessionStorage will last as long as the browser tab persists.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/notepage.vm
chengren311: updated, i think removeItem would be better to avoid conflict of same key.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/nav.vm
kunkun-tang: I see clearnning is enforced when button is submitted or clearred. So what's disadvantages of persisting the data across all tabs?

chengren311: i see your point. After thinking, localstorage could be a better option in some sense.

PR #1875, opened 2018-07-31
Conditional workflow - condition on job props for embedded flows.

jamiesjc:
Jobs in an embedded flow can only define conditions on jobs within the same embedded flow. To distinguish the job props files with the same job name prefix between different embedded flows and the parent flows, I added a new field in the ExecutableNode to store that job props file path so that each ExecutableNode can be associated with a unique job props file. 
To achieve that, I also added a default method `getJopProps()` in the `Job` interface. Since it's default, it will not break other classes which implement this `Job` interface.
See conditional workflow design #1826 for more details.


jamiesjc:
In phase one, we will roll out changes for condition on output props. Will see if there is any need for condition on job props to roll it out in the next step.

PR #1881, opened 2018-08-03
migrate user manager documentation

kunkun-tang:


docs/userManager.rst
```diff
@@ -0,0 +1,193 @@
...
+
+UserManager
+==================================
+
+When you start Azkaban, you may notice the login page. Azkaban makes you authenticate before you can use it.
+This is prevent seeing or executing workflows you shoudn’t see or touch.
+
+We also used authenticated users for auditing purposes. Whenever project files change, is modified, scheduled, etc.
+we often want to know which user performed that action.
+
```
weidious: A figure is missing.

docs/index.rst
```diff
@@ -35,6 +35,7 @@ Features
    configuration
    userGuide
    useAzkaban
+   userManager
    eventTrigger
```
weidious: Shouldn't this userManager exist above useAzkaban?

kunkun-tang: Good catch! WIll fix

docs/userManager.rst
kunkun-tang: Figure is just added.

PR #1883, opened 2018-08-03
Fix flow log CSS

juhoautio:
The Flow Log positioning was broken, as you can see here:

nayttokuva 2018-8-4 kello 10 20 51

This happens if the bar above (the one that has `Submit User`, `Duration` etc.) spans to multiple rows. Even one of them is enough to cause overflowing positioning, but the screenshot above demonstrates an extreme case where the tab is barely visible any more.

As a quick fix, I copied the same rules that are set for `div#graphView.container-fill` (you can find it right above the changed lines).

Screenshot after fix:

nayttokuva 2018-8-4 kello 10 23 11

As you can see, the fix sets a fixed height to the log textbox, and as the screenshot shows it gets cut off if the window size is not as high. There must be a way to make the position relative properly.. If that is found, why not apply that better rule also for `div#graphView.container-fill`?

### DIFF
--- a/azkaban-web-server/src/main/less/base.less
+++ b/azkaban-web-server/src/main/less/base.less
@@ -30,6 +30,12 @@ div#graphView.container-fill {
   height: 750px;
 }
 
+div#flowLogView.container-fill {
+  position: relative;
+  top: 0px;
+  height: 750px;
+}
+
 .alert-default {
   color: #a0a0a0;
   background-color: #f5f5f5;


HappyRay:
Could you include screenshots of before and after the fix?

juhoautio:
@HappyRay I included a screenshot _after_ fix. Would you be able to check what the Flow Log currently looks like at your end using the current master, and maybe add a screenshot?

I don't know how to build the less to css so that I could see the UI changes when running AzkabanSingleServer from IDEA.

juhoautio:
Actually I suspect that this may have been caused by some merge conflict in our fork, and is not an issue with the public master(?). I'll have to check more closely.

juhoautio:
@HappyRay I figured out the cause: for example a long value for `Submit User` causes this. Updated PR description accordingly.

juhoautio:
I checked build failure on Travis, the failed test is:

```
azkaban.execapp.FlowPreparerTest > testProjectCacheDirCleaner FAILED
    java.io.IOException: java.lang.RuntimeException: java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
        at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteLeastRecentlyUsedProjects(FlowPreparer.java:268)
        at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteProjectDirsIfNecessary(FlowPreparer.java:292)
        at azkaban.execapp.FlowPreparer.setupProject(FlowPreparer.java:170)
        at azkaban.execapp.FlowPreparerTest.testProjectCacheDirCleaner(FlowPreparerTest.java:176)

        Caused by:
        java.lang.RuntimeException: java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
            at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.lambda$deleteLeastRecentlyUsedProjects$0(FlowPreparer.java:264)
            at java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
            at java.util.TimSort.sort(TimSort.java:220)
            at java.util.Arrays.sort(Arrays.java:1512)
            at java.util.ArrayList.sort(ArrayList.java:1460)
            at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteLeastRecentlyUsedProjects(FlowPreparer.java:258)
            ... 3 more

            Caused by:
            java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
                at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
                at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
                at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
                at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
                at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
                at sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
                at java.nio.file.Files.readAttributes(Files.java:1737)
                at java.nio.file.Files.getLastModifiedTime(Files.java:2266)
                at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.getLastReferenceTime(FlowPreparer.java:250)
                at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.lambda$deleteLeastRecentlyUsedProjects$0(FlowPreparer.java:260)
                ... 8 more
```

So it's the same as the one reported at https://github.com/azkaban/azkaban/issues/1901

HappyRay:
> As you can see, the fix sets a fixed height to the log textbox, and as the screenshot shows it gets cut off if the window size is not as high.

I am not sure I follow.

Otherwise LGTM

juhoautio:
I mean that the existing CSS rule makes the flow log box grow according to the window size. My change makes it fixed height. So if screen is small, it gets cut and you need to scroll the screen to see the rest of the box. Then the box itself also has its own scrollbar :) And if you have a big screen then you would have some empty space below the flow log text box.

HappyRay:
Thanks for the explanation.

Can you think of a way to use the available space more efficiently? If this is a blocking issue, I am fine with checking in this change first.

juhoautio:
Personally I'm not an expert in CSS, but it seemed like it might require rather extensive refactoring to allow dynamically sizing the flow log textbox.

I understood that the existing rule with `position: absolute` would be the way to go, but the structure of divs would need to be changed to make the tab bar be the "parent" of the textbox.

It would be better to get someone that's proficient with CSS to look at this. The bug itself is easy to reproduce as I showed with the screenshot.

HappyRay:
@kunkun-tang do you know a better way?

kunkun-tang:
We had similar issues when the warning banner is pulled down. Thanks for the fix. It looks good to me

juhoautio:
@HappyRay ping

HappyRay:
I am not sure this is the best tradeoff. Ideally, we should not have to choose. We should try to find a better solution.

I will defer to @kunkun-tang.

PR #1885, opened 2018-08-03
Converted ajax API documentation to restructured text format

inramana:
Migrated the ajax API docs to restructured text format. 
Used the tool pandoc to convert existing html docs to rst format. Had to manually edit some text after conversion.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,6 +36,7 @@ Features
    userGuide
    useAzkaban
    eventTrigger
+   ajaxApi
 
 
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -36,6 +36,7 @@ Features
    userGuide
    useAzkaban
    eventTrigger
+   ajaxApi
```
weidious: Is this the right order? I'm not sure about where should we put the event trigger doc.

PR #1894, opened 2018-08-06
Pig Job Auto Tuning Integration

mkumar1984:
This change is for integrating Pig job with auto tuning framework (https://github.com/linkedin/dr-elephant/wiki/Auto-Tuning). Auto Tuning framework provides a way to automatically tune hadoop jobs. Currently it supports Pig Job for resource usage optimization. Integration is pretty easy. It requires calling getCurrentRunParameters API before running any job. This API responds with a parameter suggestion. These parameters should be used for the configuration of the job. 

**Design**

- Currently configurations are injected using HadoopConfigurationInjector. We have written a new HadoopTuningConfigurationInjector. This new injector will inject default configuration first and then call getCurrentRunParameter API to get current run of parameters from AutoTuning framework and inject those parameters in configuration of the job. 

- Failure Handling
-- API down: In case API is down, fallback method would be to let job use default configuration. 
-- Other failed reason: In case there is other failures, we will identify if this is because of auto tuning by analyzing log and then retry getCurrentRunParameter API with isRetry flag enabled. In this case getCurrentRunParameter returns best parameters from already tried parameters. 

- We will create a new wrapper HadoopTuneInSecurePigWrapper, similar to HadoopSecurePigWrapper which will have this new flow of injecting parameter, handling failures and calling status API. And based on flag auto_tuning_enabled (default false), we can choose which wrapper to use. (HadoopPigJob) Another approach could be to use auto_tuning_enabled flag to decide on flow and don’t create HadoopTuneInSecurePigWrapper. 

- We will avoid doing configuration override in the script and will ask users to not keep any configuration inside the script for enabling AutoTuning. 

**Manual Test Cases**
[AzkabanPigJobTypeTestCases.xlsx](https://github.com/azkaban/azkaban-plugins/files/1992019/AzkabanPigJobTypeTestCases.xlsx)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+  private static Props props;
+
+  private static final Logger logger;
+
+  /**
+   * In case if job is failed because of auto tuning parameters, it will be retried with the best parameters
+   * we have seen so far. Maximum number of try by default is 2, which can be configured using parameter
+   * tuning.job.retry.count
+   */
+
```
xkrogen: Extra space

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+  private static final String TUNING_JOB_RETRY_COUNT = "tuning.job.retry.count";
+
+  static {
+    logger = Logger.getRootLogger();
```
xkrogen: Why a static initialization block instead of doing this on line 53? Also if it is a static final constant, shouldn't it be `LOGGER`?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * This function runs the pig job with tuning enabled. In case tuning job fails in first try, it asks for
+   * last best parameter from tuning and re-run the job. In case tuning end point is not responding this job
+   * will run with default parameters.
+   * @param args
```
xkrogen: Should have an empty line before `@param`

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+import azkaban.jobtype.tuning.TuningParameterUtils;
+import azkaban.utils.Props;
+
+
+/**
+ * This class represent wrapper for running pig job with tuning enabled.
+ */
+public class HadoopTuningSecurePigWrapper {
+
+  public static final String WORKING_DIR = "working.dir";
```
xkrogen: I wonder if there is a global working dir config defined somewhere else that can be used here instead of a custom one just for HadoopTuningSecurePigWrapper?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+      tuningParameterUtils.updateAutoTuningParameters(props);
+
+      HadoopTuningConfigurationInjector.prepareResourcesToInject(props,
+          HadoopTuningSecurePigWrapper.getWorkingDirectory(props));
+
+      HadoopTuningConfigurationInjector.injectResources(props);
+
+      // special feature of secure pig wrapper: we will append the pig error file
+      // onto system out
+      pigLogFile = new File(System.getenv("PIG_LOG_FILE"));
```
xkrogen: constant for this?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static final Logger logger;
+
+  /**
+   * In case if job is failed because of auto tuning parameters, it will be retried with the best parameters
+   * we have seen so far. Maximum number of try by default is 2, which can be configured using parameter
+   * tuning.job.retry.count
+   */
+
+  private static final String TUNING_JOB_RETRY_COUNT = "tuning.job.retry.count";
```
xkrogen: Should this be in `TuningCommonConstants`?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
mkumar1984: Removed.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
mkumar1984: Removed static block.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopTuningSecurePigWrapper.java
mkumar1984: Changed.

PR #1896, opened 2018-08-06
Write documents for Flow 2.0.

jamiesjc:


docs/createFlows.rst
```diff
@@ -0,0 +1,134 @@
+.. _CreatingFlows:
```
kunkun-tang: When do you plan to release conditional workflow docs?

jamiesjc: It's not yet ready. Still need to add some changes to it, so will release it later.

PR #1898, opened 2018-08-07
Move python related gitignore configs to the root level

HappyRay:
This is done to set up the python virtual environment at the root level
instead of the tools/ level.

This virtual environment will be shared by multiple submodules e.g.
tools and docs.


HappyRay:
see https://github.com/direnv/direnv

I can show you.

Issue #1901, opened 2018-08-07
Flaky test azkaban.execapp.FlowPreparerTest > testProjectCacheDirCleaner


HappyRay:
Saw this failure in unrelated change #1898 

azkaban.execapp.FlowPreparerTest > testProjectCacheDirCleaner FAILED
    java.io.IOException: java.lang.RuntimeException: java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
        at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteLeastRecentlyUsedProjects(FlowPreparer.java:268)
        at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteProjectDirsIfNecessary(FlowPreparer.java:292)
        at azkaban.execapp.FlowPreparer.setupProject(FlowPreparer.java:170)
        at azkaban.execapp.FlowPreparerTest.testProjectCacheDirCleaner(FlowPreparerTest.java:176)

        Caused by:
        java.lang.RuntimeException: java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
            at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.lambda$deleteLeastRecentlyUsedProjects$0(FlowPreparer.java:264)
            at java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
            at java.util.TimSort.sort(TimSort.java:220)
            at java.util.Arrays.sort(Arrays.java:1512)
            at java.util.ArrayList.sort(ArrayList.java:1460)
            at azkaban.execapp.FlowPreparer$ProjectCacheDirCleaner.deleteLeastRecentlyUsedProjects(FlowPreparer.java:258)
            ... 3 more

            Caused by:
            java.nio.file.NoSuchFileException: projects/1.1/___azkaban_project_dir_size_in_bytes___
                at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
                at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
                at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
                at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
                at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
[...]

PR #1903, opened 2018-08-07
FlowPreparerTest to use @Rule TemporaryFolder

juhoautio:
Obviously cleaner this way.

Candidate fix for https://github.com/azkaban/azkaban/issues/1901 (just a candidate, because hard to reproduce the random failure).


juhoautio:
@HappyRay @chengren311 please have a look, I just noticed that @chengren311 was possibly working on that flaky test..

PR #1904, opened 2018-08-07
Inject azkaban configurations into hadoop application tags

hungj:
Hadoop/YARN has support for application tags, which MR/Spark can include in its job submission. This PR will add some azkaban metadata into MR config so MR can include them in the YARN application tags when it submits to YARN.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -16,6 +16,8 @@
 
 package azkaban.jobtype;
 
+import com.google.common.base.Joiner;
+
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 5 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -593,4 +598,21 @@ public static String javaOptStringFromHadoopConfiguration(Configuration conf, St
...
     return String.format("-D%s=%s", key, value);
   }
+
+  /**
+   * Construct a CSV of tags for the Hadoop application.
+   *
+   * @param List of keys to construct tags from.
+   * @return a CSV of tags
+   */
+  public static String constructHadoopTags(Props props, String[] keys) {
```
kunkun-tang: A quick question. will this method be used by Spark job? If not, we may not use public for that? Also, that would be great if we can have some basic tests for this function.

hungj: Thanks for the review @kunkun-tang, yes it will be used by spark. I added some unit tests for this function

az-hadoop-jobtype-plugin/src/test/java/azkaban/jobtype/TestHadoopJobUtilsConstructHadoopTags.java
```diff
@@ -0,0 +1,48 @@
+package azkaban.jobtype;
+
+import azkaban.utils.Props;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import junit.framework.Assert;
```
HappyRay: Thanks for improving tests!

Please use assertj for new tests. See the contributor's guide.

hungj: Thanks for the pointer @HappyRay , updated the PR

PR #1907, opened 2018-08-07
replace documentation link by the readthedocs one

kunkun-tang:


kunkun-tang:
Not really, but we need to invest time to build a new one.

PR #1911, opened 2018-08-08
when loading project dir metadata into memory, create the project size file if it doesn't exist

chengren311:
the method updateDirSize is already verified by FlowPreparerTest#testSetupProject, so unit test is not needed.


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -300,16 +301,20 @@ public boolean accept(final File dir, final String name) {
           final int versionNum = Integer.parseInt(fileName.split("\\.")[1]);
           final ProjectVersion version =
```
kunkun-tang: the naming should be called like projVersion, not version

PR #1912, opened 2018-08-08
Conditional workflow - Apply security restriction on arbitrary code in the condition

jamiesjc:
In conditional workflow design, we use scriptEngine to evaluate the condition provided by users. To prevent users from injecting malicious code inside the condition, we need to restrict the permissions like creating a new file or overwriting an existing one, or calling any system method.  
`AccessController.doPrivileged` can be used to reduce the permissions. Also `SecurityManager` needs to be enabled so that the access control will take effect.
Apart from this PR, we will also work on another PR to whitelist the syntax allowed for a valid condition.

Reference: https://wiki.sei.cmu.edu/confluence/display/java/IDS52-J.+Prevent+code+injection


jamiesjc:
Verified the changes on staging cluster and approved by the security team.

PR #1915, opened 2018-08-10
reorganize the crypto module's directory structure to match with maven convention.

chengren311:
move all java files into 'src/main/java'.


az-crypto/src/main/java/azkaban/crypto/Crypto.java
```diff
@@ -38,11 +39,11 @@ public Crypto() {
   }
 
   public static String encode(final String s) {
-    return Base64.getEncoder().encodeToString(s.getBytes());
+    return Base64.getEncoder().encodeToString(s.getBytes(Charset.defaultCharset()));
```
kunkun-tang: Should we use UTF-8 instead of default character setting?

chengren311: thanks. updated.

PR #1916, opened 2018-08-10
check if project is removed while uploading

kunkun-tang:
We found that upload API doesn't return an error when a deleted project zip is uploaded. 

When a project is deleted, Azkaban just marks it to be inactive today. In this PR, we fix this issue by adding the check. 410 Gone client error response code indicates that access to the target resource is no longer available at the origin server and that this condition is likely to be permanent.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -1671,6 +1671,11 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
     final User user = session.getUser();
     final String projectName = (String) multipart.get("project");
     final Project project = this.projectManager.getProject(projectName);
+    if (project.isActive() == false) {
+      registerError(ret, "Installation Failed. Project '" + project.getName()
+          + "' was alreaady removed.", resp, 410);
+      return;
+    }
     logger.info(
         "Upload: reference of project " + projectName + " is " + System.identityHashCode(project));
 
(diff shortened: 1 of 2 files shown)


HappyRay:
Thanks for fixing this.

Can you think of a reason why the project is not deleted?

kunkun-tang:
Do you mean the original design choice why projects are not deleted. Actually, I cannot think of valid points for that. Projects should be entirely deleted IMO. @HappyRay

HappyRay:
@jamiesjc I remember we had a discussion about this behavior before. This behavior causes some other issues too if I remember correctly.

Do you remember the outcome of that discussion?

jamiesjc:
@HappyRay I remember last time we were talking about the upload project issue caused by inconsistency of project cache and DB table. 
I agree the project should be deleted instead of setting it to inactive. I couldn't think of the case where inactive projects will be needed.

HappyRay:
@jamiesjc Thanks.

Shall we try removing the inactive project logic then? This will simplify the code a bit.

HappyRay:
We can still make this change for now.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1671,6 +1671,11 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
     final User user = session.getUser();
     final String projectName = (String) multipart.get("project");
     final Project project = this.projectManager.getProject(projectName);
+    if (project.isActive() == false) {
```
jamiesjc: In CRT deployment, `uploadProject` will be called inside `azkaban.restli.ProjectManagerResource#deploy`. Do we also need to add the same fix there?

kunkun-tang: Good found! Let me fix it.

PR #1918, opened 2018-08-14
Conditional workflow - Validate conditions

jamiesjc:
For security concerns, we want to whitelist only valid conditions for a project during uploading project zip step.
A valid condition expression should only contain the operators &&, ||, ==, !=, >, >=, <, <= and below operands:
1. conditionOnJobStatus macros. e.g., one_success, all_done
2. variable substitution. e.g., ${jobA:param}. The job name should be valid. A job is not allowed to define the condition on its descendant jobs.
3. number or a string.e.g., 1234, "hello", 'foo'

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
@@ -45,7 +45,17 @@ import org.slf4j.LoggerFactory;
  */
 public class DirectoryYamlFlowLoader implements FlowLoader {
 
+  // Pattern to match job variables in condition expressions: ${jobName:variable}
+  public static final Pattern CONDITION_VARIABLE_REPLACEMENT_PATTERN = Pattern
+      .compile("\\$\\{([^:{}]+):([^:{}]+)\\}");
   private static final Logger logger = LoggerFactory.getLogger(DirectoryYamlFlowLoader.class);
+  // Pattern to match conditionOnJobStatus macros, e.g. one_success, all_done
+  private static final Pattern CONDITION_ON_JOB_STATUS_PATTERN =
+      Pattern.compile("(?i)\\b(" + StringUtils.join(ConditionOnJobStatus.values(), "|") + ")\\b");
+  // Pattern to match a number or a string, e.g. 1234, "hello", 'foo'
+  private static final Pattern DIGIT_STRING_PATTERN = Pattern.compile("\\d+|'.*'|\".*\"");
+  // Valid operators in condition expressions: &&, ||, ==, !=, >, >=, <, <=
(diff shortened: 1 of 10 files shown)


jamiesjc:
@kunkun-tang I've thought about this and not sure if it has any use case now. If needed, we can always enable it later.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -45,7 +45,17 @@
...
 
+  // Pattern to match job variables in condition expressions: ${jobName:variable}
+  public static final Pattern CONDITION_VARIABLE_REPLACEMENT_PATTERN = Pattern
+      .compile("\\$\\{([^:{}]+):([^:{}]+)\\}");
   private static final Logger logger = LoggerFactory.getLogger(DirectoryYamlFlowLoader.class);
+  // Pattern to match conditionOnJobStatus macros, e.g. one_success, all_done
+  private static final Pattern CONDITION_ON_JOB_STATUS_PATTERN =
+      Pattern.compile("(?i)\\b(" + StringUtils.join(ConditionOnJobStatus.values(), "|") + ")\\b");
+  // Pattern to match a number or a string, e.g. 1234, "hello", 'foo'
+  private static final Pattern DIGIT_STRING_PATTERN = Pattern.compile("\\d+|'.*'|\".*\"");
```
chengren311: is this private or public? CONDITION_VARIABLE_REPLACEMENT_PATTERN is public

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
```diff
@@ -222,25 +231,97 @@ private void addEdges(final AzkabanNode node, final AzkabanFlow azkabanFlow,
...
+    if (valid) {
+      node.setCondition(condition);
+    } else {
+      this.logger.info("Condition of " + node.getId() + ": " + azkabanNode.getCondition()
+          + " is invalid, set it to false");
+      node.setCondition("false");
+    }
+  }
+
+  private boolean validateVariableSubstitution(final String operand, final AzkabanNode azkabanNode,
```
chengren311: isValidVariableSubstitution as it returns a boolean?

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: CONDITION_VARIABLE_REPLACEMENT_PATTERN is public because it's also used in another file as well. DIGIT_STRING_PATTERN is private and only used by this file.

azkaban-common/src/main/java/azkaban/project/DirectoryYamlFlowLoader.java
jamiesjc: Updated, thanks.

PR #1922, opened 2018-08-14
 No recursion in dispatch error handling

juhoautio:
The code had a recursion loop like this:

```
selectExecutorAndDispatchFlow
  -> handleDispatchExceptionCase
    -> selectExecutorAndDispatchFlow
      -> ..
```

..with max depth of `activeExecutors.size()`

Iterative approach is easier to reason about, and is totally safe against `StackOverflowError`.

This is also preliminary refactoring to eventually allow configuring higher max for dispatching errors. Currently max dispatch attempts are capped by the total number of active executors (because each executor is only tried once at most). In the future I'd like to remove that limit and rely solely on the `MAX_DISPATCHING_ERRORS_PERMITTED` property.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -1853,8 +1853,7 @@ public class ExecutorManager extends EventHandler implements
         } else {
           exflow.setUpdateTime(currentTime);
           // process flow with current snapshot of activeExecutors
-          selectExecutorAndDispatchFlow(reference, exflow, new HashSet<>(
-              ExecutorManager.this.activeExecutors));
+          selectExecutorAndDispatchFlow(reference, exflow);
           ExecutorManager.this.runningCandidate = null;
         }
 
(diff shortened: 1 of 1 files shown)


juhoautio:
@chengren311 thanks for checking in the unit tests. How about this one now?

juhoautio:
Or @jamiesjc ?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1866,29 +1865,67 @@ private void processQueuedFlows(final long activeExecutorsRefreshWindow,
...
-                selectedExecutor, exflow.getExecutionId()), e);
-            handleDispatchExceptionCase(reference, exflow, selectedExecutor,
-                availableExecutors);
+        while (true) {
+          String giveUpReason = checkGiveUpDispatching(reference, remainingExecutors);
+          if (giveUpReason != null) {
+            logger.error("Failed to dispatch queued execution " + exflow.getId() + " because "
+                + giveUpReason);
+            finalizeFlows(exflow);
+            // FAILED TOO MANY TIMES - exit
```
chengren311: does the comment match the code?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1866,29 +1865,67 @@ private void processQueuedFlows(final long activeExecutorsRefreshWindow,
...
-          ExecutorManager.this.commonMetrics.markDispatchFail();
-          handleNoExecutorSelectedCase(reference, exflow);
         }
       }
     }
 
+    private String checkGiveUpDispatching(ExecutionReference reference,
+        Set<Executor> remainingExecutors) {
+      if (reference.getNumErrors() >= this.maxDispatchingErrors) {
+        return "reached " + ConfigurationKeys.MAX_DISPATCHING_ERRORS_PERMITTED
```
chengren311: should we also log the actual max permitted dispatch limit as well?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1866,29 +1865,67 @@ private void processQueuedFlows(final long activeExecutorsRefreshWindow,
...
+            + " (tried " + reference.getNumErrors() + " executors)";
```
chengren311: tried ** executions?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: True that. I thought this comment kind of covers both.. but not as clearly as it could. I'd like to change the comment to: "GIVE UP DISPATCHING - exit". Ok?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: "Executors" is correct currently, because each available executor is only tried once at most – if all have been tried, give up. Also, this loop is about dispatching a single execution.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1867,29 +1866,67 @@ private void processQueuedFlows(final long activeExecutorsRefreshWindow,
...
-                selectedExecutor, exflow.getExecutionId()), e);
-            handleDispatchExceptionCase(reference, exflow, selectedExecutor,
-                availableExecutors);
+        while (true) {
+          String giveUpReason = checkGiveUpDispatching(reference, remainingExecutors);
+          if (giveUpReason != null) {
+            logger.error("Failed to dispatch queued execution " + exflow.getId() + " because "
+                + giveUpReason);
+            finalizeFlows(exflow);
+            // GIVE UP DISPATCHING - exit
```
juhoautio: @chengren311 comment better now?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1867,29 +1866,67 @@ private void processQueuedFlows(final long activeExecutorsRefreshWindow,
...
-            dispatch(reference, exflow, selectedExecutor);
-            ExecutorManager.this.commonMetrics.markDispatchSuccess();
-          } catch (final ExecutorManagerException e) {
-            ExecutorManager.this.commonMetrics.markDispatchFail();
-            logger.warn(String.format(
-                "Executor %s responded with exception for exec: %d",
-                selectedExecutor, exflow.getExecutionId()), e);
-            handleDispatchExceptionCase(reference, exflow, selectedExecutor,
-                availableExecutors);
+        while (true) {
```
chengren311: can we use for loop here? my concern for while(true) is it's exposed to the risk of infinite loop if some exit condition is not omitted somehow.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: thanks.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: @chengren311 I changed while loop to for.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: thanks

chengren311:
thanks! @juhoautio

juhoautio:
Thank _you_!!

Issue #1921, opened 2018-08-14
Flaky test FlowRunnerTestYaml


juhoautio:
This PR merge build had 1 test failure: https://travis-ci.org/azkaban/azkaban/builds/415808148?utm_source=github_status&utm_medium=notification

I searched full Travis log with regex ` > \w* FAILED` and found this:

```
azkaban.execapp.FlowRunnerTestYaml > testKillBasicFlowWithoutEndNode STANDARD_ERROR
    testJob: jobB SUCCEEDED
    testJob: jobA SUCCEEDED
    testJob: jobC SUCCEEDED
    ExecutableFlow: basic_flow KILLED
    ExecutableNode: jobB SUCCEEDED
    ExecutableNode: jobA SUCCEEDED
    ExecutableNode: jobC SUCCEEDED

azkaban.execapp.FlowRunnerTestYaml > testKillBasicFlowWithoutEndNode FAILED
    java.lang.AssertionError: Wrong status for [jobC] expected:<KILLED> but was:<SUCCEEDED>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:133)
        at azkaban.execapp.FlowRunnerTestBase.assertStatus(FlowRunnerTestBase.java:138)
        at azkaban.execapp.FlowRunnerTestYaml.testKillBasicFlowWithoutEndNode(FlowRunnerTestYaml.java:69)
```

`FlowRunnerTestYaml` should be stabilized.

Also, it would be nice to rename the test file/class to `FlowRunnerYamlTest`? To have the usual `*Test` suffix, right?

HappyRay:
> Also, it would be nice to rename the test file/class to FlowRunnerYamlTest? To have the usual *Test suffix, right?

Make sense to me. @jamiesjc is there any reason not to follow the standard test class naming convention?

jamiesjc:
@HappyRay @juhoautio 
Yes, I agree. It should be renamed to FlowRunnerYamlTest.
For this flaky test, I remembered fixing it before by extending the job runtime of jobC because the kill sometimes takes a long time. https://github.com/azkaban/azkaban/pull/1847
[...]

PR #1923, opened 2018-08-14
Fix flaky test and remove sleep time in FlowRunnerYamlTest.

jamiesjc:
1. Fix the flaky test #testKillBasicFlowWithoutEndNode.
2. Reduce the sleep time to 0 in the test cases.
3. Rename FlowRunnerTestYaml to FlowRunnerYamlTest.

See more discussions in https://github.com/azkaban/azkaban/issues/1921

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
@@ -20,6 +20,7 @@ import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.when;
 
 import azkaban.executor.ExecutableFlow;
+import azkaban.executor.InteractiveTestJob;
 import azkaban.executor.Status;
 import azkaban.project.Project;
 import azkaban.test.executions.ExecutionsTestUtil;
@@ -27,7 +28,7 @@ import java.io.File;
 import java.util.HashMap;
 import org.junit.Test;
 
-public class FlowRunnerTestYaml extends FlowRunnerTestBase {
+public class FlowRunnerYamlTest extends FlowRunnerTestBase {
 
   private static final String BASIC_FLOW_YAML_DIR = "basicflowwithoutendnode";
   private static final String FAIL_BASIC_FLOW_YAML_DIR = "failbasicflowwithoutendnode";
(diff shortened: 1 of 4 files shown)


kunkun-tang:
Great thanks for your review. @juhoautio

jamiesjc:
@juhoautio  Actually, I rethought about this and want to make some changes to this PR:
1. If we want to use `failJob()` or `succeedJob()` in the `InteractiveTestJob` class, we need to wait for `failJob()` or `succeedJob()` to finish and call `notify()`. If we set the sleep to 0, it means no wait and the job could finish early even before calling `failJob()` or `succeedJob()`. The default sleep time is 10 seconds and it should be sufficient enough for the method call to finish. We don't need to worry about the sleep time here because it won't actually take that long. 
2. If we don't want to use `failJob()` or `succeedJob()`, and just want to wait for the job itself to complete, then we need to explicitly set the sleep time and the `fail` key in the config. E.g.:
```
    config:
      fail: false
      seconds: 0
```
It means the job will succeed in 0 seconds.

juhoautio:
> If we set the sleep to 0, it means no wait and the job could finish early even before calling failJob() or succeedJob()

Not true if `fail` property is not set at all. Note that there's a while loop in `InteractiveTestJob`, that keeps looping unless the `fail` property exists. If sleep=0, then `wait()` is not called at all, just because `wait(0)` would wait forever. The `InteractiveTestJob` class is a bit messy and this particular behaviour is a bit unintuitive.

My understanding is that it works like this:
- fail=true -> fail immediately
- fail=false -> sleep (if `seconds` > 0), then succeed (no further looping). Additionally the sleep can be interrupted from outside, for example if the flow is killed.
- fail missing -> wait until `succeedJob()` or `failJob()` is explicitly called (note that killing ie. `InteractiveTestJob.cancel()` calls `failJob()`)

Finally, setting `sleep=0` when `fail` is missing makes the while loop "busy". I think such scenario should be handled separately in `InteractiveTestJob`. Simple way to handle that would be:

```
if (!this.jobProps.containsKey("fail") && waitMillis <= 0) {
    waitMillis = 10_000L;
}
```

But that makes the code more and more messy. I'm sure there's a cleaner way to incorporate this behaviour..

I could make another PR to make this improvement + code structure cleanup for InteractiveTestJob.

To comment your actual change, the last commit you made is a good one, because it avoids the busy while loop. The test would've been stable also without that change though.

juhoautio:
👍 by me for the whole changeset

HappyRay:
A test that is hard to understand is a sign that there is room for improvement. Maybe the test is done at a suboptimal level or the production code needs to be improved... A good test needs to be readable just like production code. 

I suggested to @jamiesjc to take a look at the unit tests I wrote for the new engine as a comparison.

jamiesjc:
Thanks @HappyRay for the suggestion. I can definitely see the efficiency and convenience of the test cases for dag engine.
I think the reason why the original `InteractiveTestJob` seems a little complicated is that it's trying to provide multiple options:
1. Explicitly fail or succeed the job like what dag engine test does in `markFailed` or `markSuccess`.
2. Also, it wants to allow the job itself to fail or succeed after a certain duration of time by providing the `fail` and `seconds` parameter in the job configuration.
Because of the second part, it adds the wait loop which complicates the logic a little bit. If we don't need the second part, I think it can be simplified. However we cannot get rid of it easily becasue many existing test cases are still using it. 
I agree for new tests it's preferable to apply some new test framework.

PR #1928, opened 2018-08-16
Improve InteractiveTestJob

juhoautio:
- Avoid busy while loop by using a non-zero wait time when the while loop won't be exited immediately after first wait.
- Reorganize code to improve readability

Was discussed at https://github.com/azkaban/azkaban/pull/1923#issuecomment-413662560.


juhoautio:
@jamiesjc this was also related to timing problems that you faced when writing some tests. How about having this "fix"?

PR #1797, opened 2018-06-06
Make pageSize configurable

fwantastic:
This will address the issue #1796. A new property called azkaban.display.numOfExecutions will be added in the azkaban.properties and be used to determine the number of rows to be displayed.

As it will default to 16 when the property does not exist, it will not have any impact on the existing page size.

PR #1938, opened 2018-08-30
fix PR 1797, a UI bug

kunkun-tang:
Found a UI bug in #1797  , when we deploy latest master to our staging cluster. The observation of this bug is all execution history table is missing, like:
￼
image

The cause looks like a javascript or backbone grammar issue.

The fix quite simple.


azkaban-web-server/src/web/js/azkaban/view/flow.js
```diff
@@ -302,6 +302,7 @@ azkaban.ExecutionsView = Backbone.View.extend({
   handlePageChange: function (evt) {
     var page = this.model.get("page") - 1;
     var pageSize = this.model.get("pageSize");
+    console.log("pageSize = " + pageSize)
```
juhoautio: @kunkun-tang this seems like an accident

PR #1310, opened 2017-08-07
Fix for failing to kill a job due to a race condition

HappyRay:
Problem:

After a flow starts, kill the flow quickly.
The flow and the job will show as killed. However, the job actually runs
to completion.

Sample job log:
Logs:
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Kill has been called.
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Process has not yet started.
25-07-2017 17:14:53 PDT IntegrationTest_jobCommand ERROR - Failed trying to cancel job. Maybe it hasn't started running yet or just finished.
25-07-2017 17:15:03 PDT IntegrationTest_jobCommand INFO - Process completed successfully in 10 seconds.
25-07-2017 17:15:03 PDT IntegrationTest_jobCommand INFO - Finishing job IntegrationTest_jobCommand attempt: 0 at 1501028103180 with status KILLED

Analysis:

Jobrunner thread runs: azkaban.jobExecutor.utils.process
.AzkabanProcess#run

A jetty thread processes the kill command:

azkaban.jobExecutor.ProcessJob#cancel
azkaban.jobExecutor.utils.process.AzkabanProcess#softKill
azkaban.jobExecutor.utils.process.AzkabanProcess#checkStarted
here it throws exception, because the job process has not been created
yet at this point.

This exception is caught and ignored.
Users are informed that the kill action is completed and reflected in
the job and flow pages.

Fix:

Synchronize the killing thread and the jobrunner thread.
The jetty thread will wait for the job process to be created if needed
before killing the job process.

Scope of the fix:

This fix only addresses the synchronization issue between the killing Jetty thread and the jobrunner thread. It doesn't address the potential synchronization issues between the killing thread and the flowrunner thread e.g. There may be still some other race condition that causes a kill action to not kill a running job. 

====

This is a fix built on the idea proposed in #1289
The previous fix #1253 for the same race condition is not complete.

Issue #1311, opened 2017-08-07
Potential race condition between flowRunner thread and jetty killing thread


jamiesjc:
Issue:
When killing the flow immediately after the flow starts, there could be race condition between 
azkaban.execapp.FlowRunner#kill() and azkaban.execapp.FlowRunner#runFlow. Users will see the job is still running into completion although the flow status is KILLED.

Analysis:
FlowRunner thread runs:
azkaban.execapp.FlowRunner#runFlow
azkaban.execapp.FlowRunner#runReadyJob
azkaban.execapp.FlowRunner#runExecutableNode
This will modify activeJobRunners list, and it is not synchronized. 
Jetty killing thread runs:
azkaban.execapp.FlowRunner#kill() 
Kill could happen during runReadyJob and cause the race condition.

Proposed fix:
Add synchronized block for runReadyJob. 

See original discussion in https://github.com/azkaban/azkaban/pull/1289

PR #1943, opened 2018-09-06
Disable flaky test - testKillBasicFlowWithoutEndNode

jamiesjc:
Disable the flaky test for now to unblock the travis build.
See the detailed discussion in #1921

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
@@ -26,6 +26,7 @@ import azkaban.project.Project;
 import azkaban.test.executions.ExecutionsTestUtil;
 import java.io.File;
 import java.util.HashMap;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class FlowRunnerYamlTest extends FlowRunnerTestBase {
@@ -56,6 +57,7 @@ public class FlowRunnerYamlTest extends FlowRunnerTestBase {
     assertFlowStatus(flow, Status.SUCCEEDED);
   }
 
+  @Ignore
   @Test
   public void testKillBasicFlowWithoutEndNode() throws Exception {
     setUp(BASIC_FLOW_YAML_DIR, BASIC_FLOW_YAML_FILE);
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
@@ -57,6 +57,12 @@ public class FlowRunnerYamlTest extends FlowRunnerTestBase {
     assertFlowStatus(flow, Status.SUCCEEDED);
   }
 
+  /**
+   * There seems to be an actual race condition bug in the runtime code. See: issue #1921: Flaky
+   * test FlowRunnerTestYaml & issue #1311: Potential race condition between flowRunner thread and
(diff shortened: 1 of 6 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerYamlTest.java
```diff
@@ -56,6 +57,7 @@ public void testBasicFlowWithoutEndNode() throws Exception {
     assertFlowStatus(flow, Status.SUCCEEDED);
   }
 
+  @Ignore
```
juhoautio: Could include a comment, something like this:

There seems to be an actual race condition bug in the runtime code. See: issue #1921: Flaky test FlowRunnerTestYaml & issue #1311: Potential race condition between flowRunner thread and jetty killing thread

jamiesjc: Good suggestion! I will update it.

HappyRay:
@jamiesjc 
What's the plan for this test next? Is there a ticket to track the remaining work?
If this test has no value, it should be removed.
If it is flaky because of test implementation, could you try fixing it? 
If the test failure is signaling us that there is a bug in the production code, you know what to do :)

jamiesjc:
@HappyRay, see my comments in #1921:

> When killing happens, jobC was not even added to activeJobRunners even though it was already submitted. We've seen this race condition in killing flows before and it would take much effort to completely fix it. I would disable this test for now to unblock the travis build. With the new DAG engine design, we should not see this kind of issue again. 
> 
> ```
> 2018/09/06 14:02:37.544 +0000 INFO [FlowRunner-exec-116] [basic_flow] Submitting job 'jobC' to run.
>     2018/09/06 14:02:37.540 +0000 INFO [Test worker] [basic_flow] Kill has been called on flow 116
>     2018/09/06 14:02:37.544 +0000 INFO [Test worker] [basic_flow] Killing 0 jobs.
>     2018/09/06 14:02:37.545 +0000 INFO [JobRunner-jobC-116] [basic_flow] Created file appender for job jobC
>     2018/09/06 14:02:37.545 +0000 INFO [JobRunner-jobC-116] [basic_flow] Attached file appender for job jobC
>     2018/09/06 14:02:37.545 +0000 INFO [JobRunner-jobC-116] [basic_flow] Job Started: jobC
> ```

You might still remember the original race condition issue we discussed in #1311. I proposed a fix before but it could not completely resolve this issue so I aborted it. I think you might have some fix for that? Anyway, this would not be a problem once we change to the new DAG engine model. 
I can create a ticket to track this issue internally.

HappyRay:
@jamiesjc 
Thanks for the explanation. I saw your comment in the test now. It is great that you are documenting why a test is disabled.

I saw that #1311 is still open. I don't remember the exact details now.  That references my previous fix #1310. Is that fix what you are referring to?

I am too busy on other projects to continue the work on the new DAG engine. There is still a lot of work to do. With the added complexity of conditional workflow, the integration work will take even longer. 

If you know a fix already for the existing code, I wouldn't wait for the DAG engine to replace this code.

jamiesjc:
There were several potential race condition issues we discovered at that time.  #1310 addressed one of them. I think there is still one issue remaining as mentioned in the change description:
```
Scope of the fix:

This fix only addresses the synchronization issue between the killing Jetty thread and the jobrunner thread. It doesn't address the potential synchronization issues between the killing thread and the flowrunner thread e.g. There may be still some other race condition that causes a kill action to not kill a running job.
```
I don't remember all the details. But I can take a look at it and try to refresh my memory.

PR #1944, opened 2018-09-07
Improve alerting on executions finalized by azkaban-web

juhoautio:
- Include reasons for finalizing an execution
- Improve status check: alert with error email if the flow status is anything else but SUCCEEDED
- Get rid of `e.printStackTrace()`, use logger instead
- Fix `MailCreator` argument name 'reasons' & remove unused args in interface

----

Note: looks like save actions plugin hasn't been applied on `ReportalMailCreator` yet. I'm following the current style of that class for now. It should be re-formatted separately.


juhoautio:
ok done

PR #1946, opened 2018-09-07
add log before prefetching hadoop token

chengren311:
Since hadoopSecurityManager.prefetchToken is synchronized, one thread might block all others from entering into this method when it's stuck fetching token. So it's important to prompt users about which step the job is stuck with.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -528,16 +528,17 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
    * Gets hadoop tokens for a user to run mapred/hive jobs on a secured cluster
    */
   @Override
-  public synchronized void prefetchToken(final File tokenFile,
-      final Props props, final Logger logger)
+  public void prefetchToken(final File tokenFile, final Props props, final Logger logger)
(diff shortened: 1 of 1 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -209,6 +200,7 @@ public static File getHadoopTokens(HadoopSecurityManager hadoopSecurityManager,
       throw new HadoopSecurityManagerException("Failed to create the token file.", e);
     }
 
+    log.info("prefetching hadoop token...");
```
li-ygerchikov: It's a fine idea. I just wonder if this is the best spot to log the message. If another client calls `prefetchToken` from somewhere else, wouldn't it be nice if we logged that as well? I think you are trying to log *before* we enter the synchronized method, in case that blocks. That's understood.  Would it be too much trouble to make `prefetchToken` an unsynchronized wrapper around the (private) `do_prefetch...`? 
Sorry, I think this was meant to be a quick change -- please feel free to ignore.

chengren311: @li-ygerchikov thanks and updated. Discussed with @HappyRay offline , the code is over complicated to reason around, but one side benefit of being synchronized i can think of is to prevent other threads hitting the job history server/metastore when the server is down or busy, which otherwise would make server even busier. Although it might not be the original motivation of being synchronized here.

PR #1953, opened 2018-09-14
Allow more dispatch retries

juhoautio:
Allow more dispatch attempts than number of active executors.

The configuration key `azkaban.maxDispatchingErrors` is thus respected without an upper cap.

Normally azkaban-web shouldn't ever fail dispatching executions to executors as long as there are active executors available. This change allows configuring a limit that is in practice high enough to keep retrying forever, until at least one responsive executor appears.

----

NOTE: The handling of dispatch errors should be improved to distinguish between retriable & non-retriable errors.

The dispatch call itself is simple: it contains the execution id and username. I can't imagine a case where the request would be syntactically invalid/incompatible with the executor.

However, If executor returns a response with "error" in it, the dispatch on ExecutorManager currently fails the dispatch, and keeps retrying until the give up condition is met. It shouldn't be like this in all cases.

For example:
- if the error is about "already running" (on that executor), ExecutorManager should treat that as a successful dispatch (even though it may be that this can't ever happen in practice).
   - Actually for this case I think executor shouldn't even return an error response.
- If the error reason is that the execution is not found in the DB when executor tries to load it (how could that happen though..?), ExecutorManager should just give up dispatching.
- And so on.

Any way, cleaning up Azkaban DB manually from problematic executions like mentioned above can be left for the responsibility of the admin, if the admin chooses to configure a non-default value for maxDispatchingErrors. This PR doesn't have to deal with handling different dispatch error cases. It can be handled later.

But this is how I'd plan to do it:
- If response is connection error -> keep retrying
- If response is received but status is not HTTP 200 OK -> keep retrying
- If response is "already running on this executor" -> treat as a success (implement this so that executor doesn't return an error in the first place)
- If response is received with any other error -> give up after receiving this kind of error from all active executors

----

Additional note on retrying after dispatch failure with "is already running":

I'm afraid that this can happen with the current azkaban code (remains to be verified though):
- azkaban-web tries to dispatch execution 123 to executor 1
- azkaban-web crashes / is killed
- executor 1 has started the execution 123 and it's running
- azkaban-web starts again
- azkaban-web fetches the queued executions from the DB
- azkaban-web tries to dispatch execution 123 to the assigned executor 1
- executor 1 returns an error
- azkaban-web rolls back executor assignment of execution 123
- azkaban-web dispaches execution 123 to executor 2
- execution 123 is running on both executors 1 & 2 at the same time

Also fix is simple: don't return an error if execution is already running on the executor.

However even that fix is not bullet-proof. Azkaban-web could also fail to receive the response of a  dispatch call because of a connection error for example. It would also then automatically try the next executor.. One option would be to have some cooldown period after a failed dispatch attempt and checking from DB if the execution is running before trying to dispatch on another executor? Seems hard to get this right without adding some proper locking though. Whew, I'm happy to realize that in our setup we typically have only 1 active executor at a time.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -109,7 +109,7 @@ public class ExecutorManager extends EventHandler implements
   private Map<String, Integer> comparatorWeightsMap;
   private long lastSuccessfulExecutorInfoRefresh;
   private ExecutorService executorInforRefresherService;
-  private long sleepAfterDispatchFailureMillis = 1000L;
+  private Duration sleepAfterDispatchFailure = Duration.ofSeconds(1L);
 
   @Inject
   public ExecutorManager(final Props azkProps, final ExecutorLoader executorLoader,
@@ -222,7 +222,7 @@ public class ExecutorManager extends EventHandler implements
             this.azkProps.getInt(
                 Constants.ConfigurationKeys.MAX_DISPATCHING_ERRORS_PERMITTED,
                 this.activeExecutors.getAll().size()),
-            this.sleepAfterDispatchFailureMillis);
+            this.sleepAfterDispatchFailure);
   }
 
   /**
@@ -1327,7 +1327,7 @@ public class ExecutorManager extends EventHandler implements
     private final int maxDispatchingErrors;
     private final long activeExecutorRefreshWindowInMillisec;
     private final int activeExecutorRefreshWindowInFlows;
-    private final long sleepAfterDispatchFailureMillis;
+    private final Duration sleepAfterDispatchFailure;
 
     private volatile boolean shutdown = false;
     private volatile boolean isActive = true;
@@ -1336,14 +1336,14 @@ public class ExecutorManager extends EventHandler implements
         final long activeExecutorRefreshWindowInTime,
         final int activeExecutorRefreshWindowInFlows,
         final int maxDispatchingErrors,
-        final long sleepAfterDispatchFailureMillis) {
+        final Duration sleepAfterDispatchFailure) {
       setActive(isActive);
       this.maxDispatchingErrors = maxDispatchingErrors;
       this.activeExecutorRefreshWindowInFlows =
           activeExecutorRefreshWindowInFlows;
       this.activeExecutorRefreshWindowInMillisec =
           activeExecutorRefreshWindowInTime;
-      this.sleepAfterDispatchFailureMillis = sleepAfterDispatchFailureMillis;
+      this.sleepAfterDispatchFailure = sleepAfterDispatchFailure;
       this.setName("AzkabanWebServer-QueueProcessor-Thread");
     }
 
@@ -1493,7 +1493,7 @@ public class ExecutorManager extends EventHandler implements
 
     private void sleepAfterDispatchFailure() {
       try {
-        Thread.sleep(this.sleepAfterDispatchFailureMillis);
+        Thread.sleep(this.sleepAfterDispatchFailure.toMillis());
       } catch (final InterruptedException e1) {
         ExecutorManager.logger.warn("Sleep after dispatch failure was interrupted - ignoring");
       }
@@ -1579,7 +1579,7 @@ public class ExecutorManager extends EventHandler implements
   }
 
   @VisibleForTesting
-  void setSleepAfterDispatchFailureMillis(long sleepAfterDispatchFailureMillis) {
-    this.sleepAfterDispatchFailureMillis = sleepAfterDispatchFailureMillis;
+  void setSleepAfterDispatchFailure(Duration sleepAfterDispatchFailure) {
+    this.sleepAfterDispatchFailure = sleepAfterDispatchFailure;
   }
 }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -33,6 +33,7 @@ import azkaban.utils.FileIOUtils.JobMetaData;
 import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import java.io.BufferedReader;
 import java.io.File;
@@ -108,6 +109,7 @@ public class ExecutorManager extends EventHandler implements
   private Map<String, Integer> comparatorWeightsMap;
   private long lastSuccessfulExecutorInfoRefresh;
   private ExecutorService executorInforRefresherService;
+  private long sleepAfterDispatchFailureMillis = 1000L;
 
   @Inject
   public ExecutorManager(final Props azkProps, final ExecutorLoader executorLoader,
@@ -219,7 +221,8 @@ public class ExecutorManager extends EventHandler implements
                 Constants.ConfigurationKeys.ACTIVE_EXECUTOR_REFRESH_IN_NUM_FLOW, 5),
             this.azkProps.getInt(
                 Constants.ConfigurationKeys.MAX_DISPATCHING_ERRORS_PERMITTED,
-                this.activeExecutors.getAll().size()));
+                this.activeExecutors.getAll().size()),
+            this.sleepAfterDispatchFailureMillis);
   }
 
   /**
@@ -1324,6 +1327,7 @@ public class ExecutorManager extends EventHandler implements
     private final int maxDispatchingErrors;
     private final long activeExecutorRefreshWindowInMillisec;
     private final int activeExecutorRefreshWindowInFlows;
+    private final long sleepAfterDispatchFailureMillis;
 
     private volatile boolean shutdown = false;
     private volatile boolean isActive = true;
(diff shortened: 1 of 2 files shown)


juhoautio:
@chengren311 because you reviewed https://github.com/azkaban/azkaban/pull/1922 you're quite familiar with this part of the code. Would you have time to have a look?

juhoautio:
Bump. If you have time, please review..

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -220,6 +220,7 @@
     public static final String EXECUTORINFO_REFRESH_MAX_THREADS =
         "azkaban.executorinfo.refresh.maxThreads";
     public static final String MAX_DISPATCHING_ERRORS_PERMITTED = "azkaban.maxDispatchingErrors";
+    public static final String SLEEP_AFTER_DISPATCH_FAILURE = "azkaban.sleepAfterDispatchFailure";
```
chengren311: i feel making this parameter as configurable exposes complexity to azkaban users. Can you think any situation where the parameter needs to be changed?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -1870,57 +1870,58 @@ private void selectExecutorAndDispatchFlow(final ExecutionReference reference,
...
-              } catch (final ExecutorManagerException e) {
-                logFailedDispatchAttempt(reference, exflow, selectedExecutor, e);
-                ExecutorManager.this.commonMetrics.markDispatchFail();
-                reference.setNumErrors(reference.getNumErrors() + 1);
-                // FAILED ATTEMPT - try other executors except selectedExecutor
-                remainingExecutors.remove(selectedExecutor);
-              }
+            } catch (final ExecutorManagerException e) {
+              logFailedDispatchAttempt(reference, exflow, selectedExecutor, e);
+              ExecutorManager.this.commonMetrics.markDispatchFail();
```
chengren311: should this case count as dispatch failure?

chengren311:
@juhoautio thanks for this PR. IMO we have to distinguish between retriable & non-retriable errors in this PR. Otherwise this would be dangerous when dispatch fails with "is already running" and all other flows will hang there due to retrying on an already running flow.

az-core/src/main/java/azkaban/Constants.java
juhoautio: For quicker unit tests. Users shouldn’t have any reason to customize it AFAIK. But it’s not totally absurd to think that in some setups it could even be useful to be able to change this.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Sorry, what so you mean?

juhoautio:
@chengren311 you’re absolutely right.

The history is that prior to making this PR I didn’t realize that there should be a distinction between retriable @ non-retriable. But then I did, and wrote about those concerns in the PR description.

I would still like to have this available as an experimental/use-at-your-own-risk feature before adding checks to detect retriable cases. After all, non-retriable dispatch errors seem very rare.

Finally, I think the bug with ”already running” is severe enough to be worth a fix asap: if executor responds with that message, stop trying to dispatch that execution again.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: i mean since the flow will still be dispatched in this case, still marking this as dispatch fail is a bit confusing. IMO we should distinguish retryable & non-retryable failure, and only mark dispatch failure for non-retryable cases.

chengren311:
@juhoautio thanks. 
```
I would still like to have this available as an experimental/use-at-your-own-risk feature before adding checks to detect retriable cases. After all, non-retriable dispatch errors seem very rare.```

Since once merged, we cannot turn on/off this feature, so i would not deem it as an experimental-able. 

What do you think of the complexity of implementing distinction between retriable & non-retriable errors? Since we are actually considering a new dispatching paradigm - 
1. web server pushes flow to dispatch to standalone dispatching queue(e.g mysql db as queue)
2. executor pulls the flow from the queue when its resource is available. 

The advantages would be
1. dispatching logic is much simpler,e.g web server doesn't need to retry on failure.
2. paves way for decoupling web server and executor

look forward to your opinion.

juhoautio:
Quick question on the matter of experimental feature: what do you mean it can’t be turned off? There’s no change in behaviour at all compared to previous version unless user specifically sets a high max retry in configuration.

juhoautio:
@chengren311 here goes,

> What do you think of the complexity of implementing distinction between retriable & non-retriable errors?

It shouldn't be too complicated I think. I would do it so that clearly non-retriable errors are detected based on what the executor response says. I would add a new field in the response specifically to communicate this. Any other error would be assumed to be retriable.

> Since we are actually considering a new dispatching paradigm -

That sounds like a major effort. I feel like there are more important improvements to be done first. Allowing more dispatch errors (infinite in practice) would work for me.

> 1. web server pushes flow to dispatch to standalone dispatching queue(e.g mysql db as queue)
> 2. executor pulls the flow from the queue when its resource is available.
>
> The advantages would be 
> 1.dispatching logic is much simpler, eg. web server doesn't need to retry on failure.

Not sure if it's actually going to be much simpler though? Somehow an executor would need to make in initial reservation to an execution in the queue, then only if it actually manages to start it, remove it from the queue. But you would need to achieve this kind of transactionality in a multi-node setup. The point is to avoid a case where an execution is removed from queue but executor doesn't even start running it because of an error.

Current model makes an initial executor assignment and removes it if dispatching fails. It has some weakness in dispatching currently, but it could be rather easily be fixed to guarantee exactly once dispatching, thanks to azkaban-web's ability to synchronize on itself, being a single node. I'm sure same things can be done well enough with a queue-based solution, but I wouldn't expect it to be simpler to implement – rather more complex.

> 2. paves way for decoupling web server and executor

The cleanup of executions left behind is also still needed. It can be done obviously, but just saying.

What actual problem does the "decoupling" solve?

Could you check also my previous comment, please?

chengren311:
@juhoautio 
```There’s no change in behaviour at all compared to previous version unless user specifically sets a high max retry in configuration.```
if i understand correctly about this change, in this version failed dispatch will keep retrying until it gets successfully dispatched, but previous version will retry to a upper cap, am i right?

juhoautio:
@chengren311, this PR only _allows configuring more dispatch attempts_ than number of active executors, which is the upper cap currently. Default value remains the same as before: the number of executors.

From the PR description:

> This change allows configuring a limit that is in practice high enough to keep retrying forever, until at least one responsive executor

By this I don't mean an actual infinite loop. I just mean that if you configure a really high value for max attempts, in practice it's basically the same as retrying forever.

juhoautio:
The existing unit test testDispatchFailed verifies the behavior of giving up dispatching when max attempt is reached.

juhoautio:
@chengren311 did my answers make sense? If yes, I can resolve the conflicts.

az-core/src/main/java/azkaban/Constants.java
chengren311: do you have some examples of such setups?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
chengren311: Someone might set up alert on dispatching failure, like us. So we want to make sure alert indicates real dispatching failures.

chengren311:
@juhoautio thanks, left some replies.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: Yes. But as I said, this matches the existing behaviour, I'm not changing it. For example my proposal of splitting to different counters could be done later if desired.

az-core/src/main/java/azkaban/Constants.java
juhoautio: I guess there are two branches:

1. Delay minimization
   - This is the one that makes more sense probably
   - User is scheduling a lot of executions all the time and wants to minimize the dispatching delay due to any temporary dispatch failures (caused by executor replacement, for example).
   - For this to be used effectively in executor replacement, assuming that azkaban-web has been modified to automatically refresh executors without having to restart it (discussion is going on about that with @jamiesjc).
2. Call throttling
   - User wants to use a longer sleep to avoid a high call rate on executors if dispatching keeps failing.
   - This seems more artificial because the dispatch call alone won't probably be too expensive – it would make more sense to increase the max number of dispatch attempts than to increase the sleep time between attempts.

But as I wrote _it’s not totally absurd_, I mean that most likely users won't ever need to touch these. But I would like to have them as properties like this so that they can be customized for unit tests. Other option would be to add setters with package visibility, but that's not as clean IMHO.

juhoautio: Nope, it makes the tests slower if there's some sleeping involved. Tests should never sleep.

chengren311: thanks, i don't have very strong opinion, so up to your preference.
One thing I want to add is we should add time unit in property name to avoid confusion.

juhoautio:
@chengren311 I've rebased + resolved conflicts. To make the test quick, I used a package-visible setter `setSleepAfterDispatchFailureMillis(0L)` instead of introducing a new property name. I hope you like that.

HappyRay:
Consider using Duration? 
And in the configuration, I now think it is better to standardize on what @suvodeep-pyne suggested earlier in #955

"
Regarding configuration parameters, Duration can be there as well. There are methods like java.time.Duration#parse which uses the ISO-8601 duration format PnDTnHnMn.nS. Example:

# Set max flow running time to 15 min
azkaban.server.max.flow.running=PT15M
"

juhoautio:
@HappyRay ok I changed to use Duration. Code complexity grew slightly, but at least the arg name is shorter now.

HappyRay:
@juhoautio Thanks.

PR #1958, opened 2018-09-19
Conditional workflow UI - add condition in job info and flow summary

jamiesjc:
Added the UI support for conditional workflow:
1. Added condition in job info page.
2. Added condition in flow summary page.
Previously, the condition was resolved and updated during the validation step. However, in order to show the original condition on UI page, the condition should remain unchanged until the execution time. Also for the condition of an embedded flow to show up on UI, I explicitly added the condition for flow object.
I also improved some logging messages during the evaluation of the condition.

Example-1 Condition defined on job info page: 
screen shot 2018-09-18 at 3 53 57 pm

Example-2 No condition defined on job info page: 
screen shot 2018-09-18 at 3 45 35 pm

Example-3 Condition defined on flow summary page: 
screen shot 2018-09-18 at 3 43 48 pm

Example-4 No condition defined on flow summary page: 
screen shot 2018-09-18 at 3 45 51 pm

Users might need to clear the browser cache to see the new UI change.
Conditional workflow design #1826

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -923,17 +923,19 @@ public class FlowRunner extends EventHandler implements Runnable {
     String replaced = condition;
     // Replace the condition on job status macro with "true" to skip the evaluation by Script
     // Engine since it has already been evaluated.
-    Matcher matcher = CONDITION_ON_JOB_STATUS_PATTERN.matcher(condition);
-    if (matcher.find()) {
-      replaced = condition.replace(matcher.group(1), "true");
+    final Matcher jobStatusMatcher = CONDITION_ON_JOB_STATUS_PATTERN.matcher
+        (condition);
+    if (jobStatusMatcher.find()) {
+      replaced = condition.replace(jobStatusMatcher.group(1), "true");
     }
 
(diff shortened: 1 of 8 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -917,15 +920,22 @@ private Boolean isConditionOnRuntimeVariableMet(final ExecutableNode node) {
...
-    final Matcher matcher = CONDITION_VARIABLE_REPLACEMENT_PATTERN.matcher(condition);
     String replaced = condition;
+    // Replace the condition on job status macro with "true" to skip the evaluation by Script
+    // Engine since it has already been evaluated.
+    Matcher matcher = CONDITION_ON_JOB_STATUS_PATTERN.matcher(condition);
+    if (matcher.find()) {
+      replaced = condition.replace(matcher.group(1), "true");
+    }
+
+    matcher = CONDITION_VARIABLE_REPLACEMENT_PATTERN.matcher(replaced);
```
juhoautio: Minor thing but maybe it'd be clearer to use a different name for the different matchers here, and thus allowing both variables to be final.

jamiesjc: Sure, I'll update the matcher name, thanks!

Issue #1920, opened 2018-08-14
Update Save Actions settings in CONTRIBUTING.md > Style guides


juhoautio:
https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md#style-guides

It currently says that:

> Please turn on all the options except
> - Remove unused suppress warning annotation

But looks like Save Actions plugin has changed a bit. The azkaban dev instructions need an update.

I was trying this:

![nayttokuva 2018-8-14 kello 11 19 03](https://user-images.githubusercontent.com/4446608/44080162-e1e492f0-9fb3-11e8-8e55-087d1afce89d.png)

..but still didn't get it right? For example applying save actions on `ExecutorManager.java` causes quite many changes, because it turns `logger.*` calls inside sub-classes into `ExecutorManager.logger.*`. What are the correct settings currently and has all code in the repo already been reformatted by save actions?

HappyRay:
I agree. 
I opened an internal issue to get it fixed. But it still hasn't been prioritized. :(

PR #1962, opened 2018-09-21
update save action guide

chengren311:
See https://github.com/azkaban/azkaban/issues/1920

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,7 +53,7 @@ Install and enable the intellij's
 [save actions plugin](https://github.com/dubreuia/intellij-plugin-save-actions) 
 to reformat/refactor code automatically:
 
-Please turn on the following options only(all of them are available in release 1.1.0 and above, see https://github.com/dubreuia/intellij-plugin-save-actions/releases):
+Please turn on the following options only (all of them are available in release 1.1.0 and above, see https://github.com/dubreuia/intellij-plugin-save-actions/releases):
   * Optimize imports
   * Reformat file
   * Add final modifier to field	
@@ -67,7 +67,7 @@ Please turn on the following options only(all of them are available in release 1
   * Remove explicit generic type for diamond	
   * Remove unnecessary semicolon	
 
-Please refer to save actions plugin(https://github.com/dubreuia/intellij-plugin-save-actions) for implication of each option.
+Please refer to save actions plugin (https://github.com/dubreuia/intellij-plugin-save-actions) for implication of each option.
 
 ### New line at the end of a file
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -53,20 +53,21 @@ Install and enable the intellij's
...
-* Reformat code based on the code style
-* Rearrange code (reorder methods, fields, etc.)
-* Add final to local variable
-* Add final to field
-* Remove explicit generic type for diamond
-* Qualify field access with this
-* Remove final from private method
-* Remove unnecessary semicolon
-* Add missing @Override annotations
+Please turn on the following options only:
```
HappyRay: Specify the minimum version of the plugin that has these options?

chengren311: updated, thanks

CONTRIBUTING.md
```diff
@@ -53,20 +53,21 @@ Install and enable the intellij's
...
-* Reformat code based on the code style
-* Rearrange code (reorder methods, fields, etc.)
-* Add final to local variable
-* Add final to field
-* Remove explicit generic type for diamond
-* Qualify field access with this
-* Remove final from private method
-* Remove unnecessary semicolon
-* Add missing @Override annotations
+Please turn on the following options only(all of them are available in release 1.1.0 and above, see https://github.com/dubreuia/intellij-plugin-save-actions/releases):
```
kunkun-tang: Just a quick typo fix: parentheses need a space ahead.

HappyRay:
@juhoautio what do you think of the proposed conventions?

juhoautio:
👍 This looks really good.

Any plans to apply these rules on all classes in the repo?

For example, if I apply save actions with this configuration on ExecutorManager.java, there are many references to the `logger` in the inner classes that then become `ExecutorManager.logger`.

chengren311:
@juhoautio yes, once merged, reformatting the whole repo will be needed to be done.

HappyRay:
> For example, if I apply save actions with this configuration on ExecutorManager.java, there are many references to the logger in the inner classes that then become ExecutorManager.logger.

I believe it won't happen with the proposed rule above:
Add class qualifier to static member access **outside declaring class only**

chengren311:
@HappyRay actually class qualifier will be added if it's inside the inner class of declaring class.

HappyRay:
I see. Do these inner classes need to share the parent class's logger?

chengren311:
@HappyRay some inner class are used just for encapsulation purpose, but for the inner classes in ExecutorManager, i think it makes more sense to separate them out as regular classes.

HappyRay:
I need to review the code to remember what they do. But a good thought. Care to refractor it?

chengren311:
sure, @juhoautio FYI.

juhoautio:
@chengren311 @HappyRay see my other pr, I already started doing that.

PR #1971, opened 2018-10-10
Refactor ExecutorManager

juhoautio:
After this it'll be possible to create for example a pure unit test for the RunningExecutionsUpdaterThread alone.

TODO: once reviewed, check once more that save actions have been applied on all changed classes

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ActiveExecutors.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ActiveExecutors.java
@@ -1,0 +1,70 @@
+package azkaban.executor;
+
+import azkaban.Constants;
+import azkaban.Constants.ConfigurationKeys;
+import azkaban.utils.Props;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.log4j.Logger;
+
+public class ActiveExecutors {
+
+  private static final Logger logger = Logger.getLogger(ExecutorManager.class);
+
+  private volatile ImmutableSet<Executor> activeExecutors;
+  private final Props azkProps;
+  private final ExecutorLoader executorLoader;
+
+  public ActiveExecutors(final Props azkProps, final ExecutorLoader executorLoader) {
+    this.azkProps = azkProps;
+    this.executorLoader = executorLoader;
+  }
+
+  public void setupExecutors() throws ExecutorManagerException {
+    final Set<Executor> newExecutors = new HashSet<>();
+
+    if (ExecutorManager.isMultiExecutorMode(this.azkProps)) {
+      logger.info("Initializing multi executors from database.");
+      newExecutors.addAll(this.executorLoader.fetchActiveExecutors());
+    } else if (this.azkProps.containsKey(ConfigurationKeys.EXECUTOR_PORT)) {
+      // add local executor, if specified as per properties
+      final String executorHost = this.azkProps
+          .getString(Constants.ConfigurationKeys.EXECUTOR_HOST, "localhost");
+      final int executorPort = this.azkProps.getInt(ConfigurationKeys.EXECUTOR_PORT);
+      logger.info(String.format("Initializing local executor %s:%d",
+          executorHost, executorPort));
+      Executor executor =
+          this.executorLoader.fetchExecutor(executorHost, executorPort);
+      if (executor == null) {
+        executor = this.executorLoader.addExecutor(executorHost, executorPort);
+      } else if (!executor.isActive()) {
+        executor.setActive(true);
+        this.executorLoader.updateExecutor(executor);
+      }
+      newExecutors.add(new Executor(executor.getId(), executorHost,
+          executorPort, true));
+    } else {
+      // throw exception when in single executor mode and no executor port specified in azkaban
+      // properties
+      //todo chengren311: convert to slf4j and parameterized logging
+      final String error = "Missing" + ConfigurationKeys.EXECUTOR_PORT + " in azkaban properties.";
+      logger.error(error);
+      throw new ExecutorManagerException(error);
+    }
+
+    if (newExecutors.isEmpty()) {
+      final String error = "No active executor found";
+      logger.error(error);
+      throw new ExecutorManagerException(error);
+    } else {
+      this.activeExecutors = ImmutableSet.copyOf(newExecutors);
+    }
+  }
+
+  public Collection<Executor> getAll() {
+    return activeExecutors;
+  }
+
+}
--- a/azkaban-common/src/main/java/azkaban/executor/ActiveExecutors.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ActiveExecutors.java
@@ -7,6 +7,7 @@ import com.google.common.collect.ImmutableSet;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
+import javax.inject.Inject;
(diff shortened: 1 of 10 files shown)


juhoautio:
@jamiesjc this is some groundwork for https://github.com/azkaban/azkaban/pull/1833#issuecomment-428481883.

And especially as you wrote in https://github.com/azkaban/azkaban/pull/1833#issuecomment-428276401:

> It would be great if you could figure out a proper way to test the changes as well.

azkaban-common/src/main/java/azkaban/executor/ActiveExecutors.java
```diff
@@ -0,0 +1,72 @@
...
+import azkaban.Constants.ConfigurationKeys;
+import azkaban.utils.Props;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import javax.inject.Inject;
+import org.apache.log4j.Logger;
+
+public class ActiveExecutors {
```
jamiesjc: Please add the javadoc for all the new public class and methods.

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdaterThread.java
```diff
@@ -0,0 +1,293 @@
...
+  public void run() {
+    while (!this.shutdown) {
+      try {
+        this.lastThreadCheckTime = System.currentTimeMillis();
+        this.updaterStage.set("Starting update all flows.");
+        // TODO refresh active executors
+        // TODO BUT this needs to know the up-to-date list of active AND inactive executors. OR:
+        // TODO OR then this would need to call getFlowToExecutorMap() again (which returns all
+        // executors regardless of active=true|false - or null optional if it doesn't exist)
+        // ..so would that make "refresh active executors" just a side-effect?
```
jamiesjc: I don't quite follow the TODO here. Do you intend to load all executors or just active executors?

kunkun-tang:
That would be great if we could rely on this opportunity to construct the new Executor Manager and new Unite Test solution to make our lives easy.

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdaterThread.java
juhoautio: The thing is I'm not so sure what would be the best way, but I'll think about it. The main point for having this TODO here is to remind that checking for _active_ executors only is not enough.

What I'm trying to say there is that one option would be call `getFlowToExecutorMap()` if executor can't be reached. Because the result of that can be used to check if the executor doesn't exist any more.

On the other hand reloading _active_ executors won't be enough for this check, because active executors don't include inactive executors. This class needs to know if the executor has been removed entirely.

Outside of this class there's a need for updating _active_ executors: for the dispatching side to take new executors into use without manually asking to refresh via API or restarting azkaban-web.

I'm wondering what would be the cleanest way to achieve both things. The ways mentioned above are nice in a way that they're pretty much additive in nature. Other approaches might involve more refactoring, but result in a cleaner outcome. For example, maybe it would make sense to load both inactive and active executors at the same time, and then filter them to only use the active ones in dispatching side.

juhoautio:
@jamiesjc javadocs added

azkaban-common/src/main/java/azkaban/executor/RunningExecutions.java
```diff
@@ -1,44 +1,41 @@
...
-import java.util.Collection;
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.inject.Singleton;
 
+/**
+ * Provides access to running executions.
+ */
 @Singleton
 public class RunningExecutions {
```
jamiesjc: Is it necessary to wrap runningExecutions inside a new class?

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdaterThread.java
jamiesjc: > What I'm trying to say there is that one option would be call getFlowToExecutorMap() if executor can't be reached.

Currently, `getFlowToExecutorMap()` will only check the runningExecutions cache to get the result. However in the previous case we discussed where an executor was shut down first and the runningExecutions cache didn't get a chance to update, this would not solve the problem. 
How about checking the executor info from DB directly if that executor cannot be reached? Then you wouldn't need to reload active executors.

> Outside of this class there's a need for updating active executors: for the dispatching side to take new executors into use without manually asking to refresh via API or restarting azkaban-web.

This would be an improvement but not necessarily the fix for the current issue. Also in our case, updating new executors is only required after new executor deployment, so I don't see much efficiency in updating frequently in the updater thread. 
Correct me if I understand it wrong.

azkaban-common/src/main/java/azkaban/executor/RunningExecutions.java
juhoautio: > Is it necessary to wrap runningExecutions inside a new class?

Not really, but this makes injection clearer. I could bind an instance of ConcurrentHashMap as a named singleton, but that would require using the `@Named` annotation also in the classes that require it in their `@Inject` constructor. So I would prefer having a class. Please let me know.

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdaterThread.java
juhoautio: > Currently, getFlowToExecutorMap() will only check the runningExecutions cache to get the result.

I'm sorry, I somehow confused it with `ExecutorManager#loadRunningExecutions()`, which is only called when the ExecutorManager is created.

> How about checking the executor info from DB directly if that executor cannot be reached? Then you wouldn't need to reload active executors.

Yeah, why not. I was just thinking of some way to possibly combine shared needs. At least I would implement that check so that I would call the existing `executorLoader.fetchActiveExecutors()` and check if the result includes the executor or not, instead of adding new DB-level code to check for the existence of a specific executor id.

> Then you wouldn't need to reload active executors.
> This would be an improvement but not necessarily the fix for the current issue.

Yes, exactly. I also came to the conclusion that these are separate things, and it may not even make sense to combine them.

> Also in our case, updating new executors is only required after new executor deployment, so I don't see much efficiency in updating frequently in the updater thread.

That is true, but then it would require user to call the refresh executors API, which complicates the operating side. My point was that listing executors periodically is not expensive at all because it's such a small table, and doing that often enough would make adding and removing executors more convenient.

juhoautio:
@jamiesjc I removed that TODO comment about refreshing executors. Would this PR be acceptable now?

azkaban-common/src/main/java/azkaban/executor/RunningExecutions.java
```diff
@@ -0,0 +1,41 @@
...
+
+import azkaban.utils.Pair;
+import java.util.concurrent.ConcurrentHashMap;
+import javax.inject.Singleton;
+
+/**
+ * Provides access to running executions.
+ */
+@Singleton
+public class RunningExecutions {
```
kunkun-tang: Looks like we don't need to bind this class to an object in a guice module?

juhoautio: Yeah, Guice creates the instance if an instance is required for injection. It will only create one, because this class is annotated with `@Singleton`.

PR #1975, opened 2018-10-15
Create a unit test for RunningExecutionsUpdater + minor refactoring

juhoautio:
1. Create a unit test for RunningExecutionsUpdater
2. Move update request code to ExecutorApiGateway
   - Better scope for RunningExecutionsUpdater & cleaner unit tests

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
+++ b/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
@@ -18,7 +18,6 @@ package azkaban.executor;
 
 import azkaban.alert.Alerter;
 import azkaban.metrics.CommonMetrics;
-import azkaban.utils.JSONUtils;
 import azkaban.utils.Pair;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -75,8 +74,6 @@ public class RunningExecutionsUpdater {
 
     for (final Map.Entry<Optional<Executor>, List<ExecutableFlow>> entry : exFlowMap
         .entrySet()) {
-      final List<Long> updateTimesList = new ArrayList<>();
-      final List<Integer> executionIdsList = new ArrayList<>();
 
       final Optional<Executor> executorOption = entry.getKey();
       if (!executorOption.isPresent()) {
@@ -92,24 +89,9 @@ public class RunningExecutionsUpdater {
       this.updaterStage.set("Starting update flows on " + executor.getHost() + ":"
           + executor.getPort());
 
-      // We pack the parameters of the same host together before we
-      // query.
-      fillUpdateTimeAndExecId(entry.getValue(), executionIdsList,
-          updateTimesList);
-
-      final Pair<String, String> updateTimes =
-          new Pair<>(
-              ConnectorParams.UPDATE_TIME_LIST_PARAM,
-              JSONUtils.toJSON(updateTimesList));
-      final Pair<String, String> executionIds =
-          new Pair<>(ConnectorParams.EXEC_ID_LIST_PARAM,
-              JSONUtils.toJSON(executionIdsList));
-
       Map<String, Object> results = null;
       try {
-        results = this.apiGateway.callWithExecutionId(executor.getHost(),
-            executor.getPort(), ConnectorParams.UPDATE_ACTION,
-            null, null, executionIds, updateTimes);
+        results = this.apiGateway.updateExecutions(executor, entry.getValue());
       } catch (final ExecutorManagerException e) {
         handleException(entry, executor, e);
       }
@@ -204,14 +186,6 @@ public class RunningExecutionsUpdater {
     }
 
     return exFlowMap;
-  }
-
-  private void fillUpdateTimeAndExecId(final List<ExecutableFlow> flows,
-      final List<Integer> executionIds, final List<Long> updateTimes) {
-    for (final ExecutableFlow flow : flows) {
-      executionIds.add(flow.getExecutionId());
-      updateTimes.add(flow.getUpdateTime());
-    }
   }
 
   private ExecutableFlow updateExecution(final Map<String, Object> updateData)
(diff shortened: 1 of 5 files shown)


azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
```diff
@@ -0,0 +1,158 @@
...
+      this.updater.updateExecutions();
+      DateTimeUtils.setCurrentMillisFixed(
+          DateTimeUtils.currentTimeMillis() + this.updater.errorThreshold + 1L);
+    }
+    verify(this.mailAlerter).alertOnFailedUpdate(
+        this.activeExecutor, Collections.singletonList(this.execution), API_CALL_EXCEPTION);
+
+    // TODO change to checking if executor exist in the DB any more
+    verifyZeroInteractions(this.executionFinalizer);
+    // verify(this.executionFinalizer).finalizeFlow(this.execution, "TODO", null);
```
juhoautio: This is to eventually (also) test the improvement where executor is checked again from the DB & it's not found any more. But this test should be also kept for the case where executor exists in the DB but is still not responding.

juhoautio:
@jamiesjc could you review maybe?

I rebased & resolved conflicts. Possibly still getting unrelated test failures.. Would be good to merge at least this fix: https://github.com/azkaban/azkaban/pull/1985.

juhoautio:
> The job exceeded the maximum log length, and has been terminated.

XD

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
```diff
@@ -0,0 +1,153 @@
...
+  private void mockUpdateCallFails() throws ExecutorManagerException {
+    doThrow(API_CALL_EXCEPTION).when(this.apiGateway).updateExecutions(any(), any());
+  }
+
+  private void verifyCallUpdateApi() throws ExecutorManagerException {
+    verify(this.apiGateway).updateExecutions(
+        this.activeExecutor, Collections.singletonList(this.execution));
+  }
+
+  private void mockFlowDoesNotExist() throws Exception {
```
jamiesjc: Do we need to add `("error", "Flow does not exist")` in the executionMap first before mocking the update response for this scenario?
Also, better to use `ConnectorParams.RESPONSE_ERROR` instead of `error`.

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
```diff
@@ -0,0 +1,153 @@
...
+  private void mockExecutorDoesNotExist() {
+    this.runningExecutions.get().put(EXECUTION_ID_77, new Pair<>(
+        new ExecutionReference(EXECUTION_ID_77, null), this.execution));
+  }
+
+  private void mockUpdateCallFails() throws ExecutorManagerException {
+    doThrow(API_CALL_EXCEPTION).when(this.apiGateway).updateExecutions(any(), any());
+  }
+
+  private void verifyCallUpdateApi() throws ExecutorManagerException {
```
jamiesjc: Do we need to verify the returned value from `updateExecutions`? It should be different for `updateExecutionsStillRunning` and `updateExecutionsFlowDoesNotExist` right?

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
```diff
@@ -0,0 +1,153 @@
...
+    when(this.alerterHolder.get("email")).thenReturn(this.mailAlerter);
+  }
+
+  @After
+  public void tearDown() {
+    DateTimeUtils.setCurrentMillisSystem();
+  }
+
+  @Test
+  public void updateExecutionsStillRunning() throws Exception {
```
jamiesjc: Consider adding the test for finished executions as well?

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
```diff
@@ -40,8 +40,8 @@
   // times (3600 seconds = 1 hour) before we send an email about unresponsive executor.
   private final int numErrorsBetweenUnresponsiveEmail = 360;
   // First email is sent after 1 minute of unresponsiveness
-  private final int numErrorsBeforeUnresponsiveEmail = 6;
-  private final long errorThreshold = 10000;
+  final int numErrorsBeforeUnresponsiveEmail = 6;
```
kunkun-tang: Is there a reason why we wipe out private declaration?

juhoautio: This field is read in the unit test to know how many times to run in order to simulate a failure email. Other option would be to hardcode 6 in the test, but I prefer this.

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
juhoautio: Yes, somehow didn't occur to me.

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
juhoautio: Not sure what you mean? The line below verifies that the mocked apiGateway's updateExecutions method was called with the expected args. There's no reason to assert that a mock has returned what it was mocked to return.

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
juhoautio: > Do we need to add ("error", "Flow does not exist") in the executionMap first before mocking the update response for this scenario?

No, the order doesn't matter because it's the same map instance. I mean, the mock won't be called before the entry has been added to the map.

> Also, better to use ConnectorParams.RESPONSE_ERROR instead of error.

That's true.

juhoautio:
@jamiesjc I added a commit

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
jamiesjc: ok makes sense.

PR #1979, opened 2018-10-16
Demo: Fix email alerts resolve props

juhoautio:
I found a problem in the way how failure.emails etc. are handled. Unfortunately variable placeholders are not resolved.

I had conf like this (simplified):

```properties
failure_emails=receive2@domain.com
failure.emails=${failure_emails}
```

This caused flow-level email alerts to fail, because they were tried to be sent to `${failure_emails}` instead of `receive2@domain.com`. Turns out this is because props are not resolved at all in this case, even though properties are inherited from parent nodes etc.

----

To demonstrate the problem in an easy to reproduce form I did this:

1. Test flow level email options
2. Demonstrate email fix with props resolving

NOTE: this fix alone can't be applied in practice, because job type plugin props & common props added by JobRunner are missing, and which can make the job properties resolving fail in many cases . The full set of props should be resolved in the same way on both sides (web:DirectoryFlowLoader & executor:JobRunner & co.).

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
+++ b/azkaban-common/src/main/java/azkaban/project/DirectoryFlowLoader.java
@@ -26,6 +26,7 @@ import azkaban.project.FlowLoaderUtils.DirFilter;
 import azkaban.project.FlowLoaderUtils.SuffixFilter;
 import azkaban.project.validator.ValidationReport;
 import azkaban.utils.Props;
+import azkaban.utils.PropsUtils;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


juhoautio:
Anyone to comment on this, please?

juhoautio:
@burgerkingeater could you have a look at this, please? What do you think; is it clearly a bug & should be fixed?

nianglao:
I met the same issue.

`failure.emails=${MAIL_TO}`

I want to set MAIL_TO when I trigger the flow to execute manually.

Can anyone help on this?

burgerkingeater:
@juhoautio thanks for reporting this issue. Yes this is unexpected behavior and should be fixed.

juhoautio:
Closing this.. Created an issue: https://github.com/azkaban/azkaban/issues/2187

Issue #1981, opened 2018-10-16
Flaky test FlowTriggerServiceTest


juhoautio:
From https://travis-ci.org/azkaban/azkaban/builds/442211810?utm_source=github_status&utm_medium=notification

Two failed tests:
- azkaban.flowtrigger.FlowTriggerServiceTest > testStartZeroDependencyTrigger FAILED
- azkaban.flowtrigger.FlowTriggerServiceTest > testRecovery FAILED

More details below.

```
azkaban.flowtrigger.FlowTriggerServiceTest > testRecovery FAILED
    java.lang.AssertionError: 
    Expecting empty but was:<[TriggerInstance{depInstances=[DependencyInstance{startTime=1539701245696, depName='2secs', context=azkaban.flowtrigger.testplugin.TestDependencyInstanceContext@1063335c, endTime=1539701248807, status=SUCCEEDED, cause=NONE}, DependencyInstance{startTime=1539701245696, depName='3secs', context=azkaban.flowtrigger.testplugin.TestDependencyInstanceContext@3a9e564b, endTime=1539701251763, status=SUCCEEDED, cause=NONE}, DependencyInstance{startTime=1539701245696, depName='4secs', context=azkaban.flowtrigger.testplugin.TestDependencyInstanceContext@7b46f3ec, endTime=1539701251883, status=SUCCEEDED, cause=NONE}], id='1a9d0afd-3c90-4c54-89a2-d9441b118b58', submitUser='test', project=azkaban.project.Project@75a3f9c6, flowId='testflow', flowVersion=1, flowTrigger=FlowTrigger{schedule=azkaban.project.CronSchedule@2cd37026, maxWaitDurationInMins=PT10S
     FlowTriggerDependency{name='2secs', type='TestDependencyCheck', props={failure=false, runtime=2}}
    FlowTriggerDependency{name='3secs', type='TestDependencyCheck', props={failure=false, runtime=3}}
    FlowTriggerDependency{name='4secs', type='TestDependencyCheck', props={failure=false, runtime=4}}}, flowExecId=-1},
[...]

PR #1982, opened 2018-10-16
Ignore Flaky test FlowTriggerServiceTest & slow tests

juhoautio:
- Temporary fix for https://github.com/azkaban/azkaban/issues/1981
- Those slow tests took 5 & 6 seconds. There shouldn't be any sleeping in unit tests..


juhoautio:
@chengren311 @jamiesjc could this be merged, please? Almost all PR merge builds are failing and it makes it quite slow to check if a PR is legit.

juhoautio:
When making PRs it's a bit counterproductive that the "codecov/project" check is almost always failed. Do you think it could be disabled, or then set some higher threshold for that check? For example a limit of 0.5% would allow this PRs check to pass.

jamiesjc:
I think it makes sense to set some higher threshold for codecov.
cc @kunkun-tang

PR #1986, opened 2018-10-17
Always use multi-executor mode

juhoautio:
This to simplify the code.

This requires users to migrate to `azkaban.use.multiple.executors=true` if not already using it.
After this change Azkaban will refuse to start if the property is missing or if it's set to false.

----

Note: a related pattern on executor side is the RemoteFlowWatcher vs. LocalFlowWatcher, which is used to track pipelined jobs & flows. It would be better to have just a single implementation (remote). However, that doesn't have anything to do with the multi-executor mode per se. Executor picks the FlowWatcher type based on whether the tracked execution is running on the same executor or a different one.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -58,8 +58,7 @@ public class Constants {
   public static final String AZKABAN_PROPERTIES_FILE = "azkaban.properties";
   public static final String AZKABAN_PRIVATE_PROPERTIES_FILE = "azkaban.private.properties";
   public static final String DEFAULT_CONF_PATH = "conf";
-  public static final String AZKABAN_EXECUTOR_PORT_FILENAME = "executor.port";
-  public static final String AZKABAN_EXECUTOR_PORT_FILE = "executor.portfile";
+  public static final String DEFAULT_EXECUTOR_PORT_FILE = "executor.port";
 
   public static final String AZKABAN_SERVLET_CONTEXT_KEY = "azkaban_app";
 
@@ -156,7 +155,8 @@ public class Constants {
 
     // Legacy configs section, new configs should follow the naming convention of azkaban.server.<rest of the name> for server configs.
 
-    // The property is used for the web server to get the port of the executor when running in SOLO mode.
+    public static final String EXECUTOR_PORT_FILE = "executor.portfile";
+    // To set a fixed port for executor-server. Otherwise some available port is used.
     public static final String EXECUTOR_PORT = "executor.port";
 
     // Max flow running time in mins, server will kill flows running longer than this setting.
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -156,9 +156,6 @@ public class Constants {
 
     // Legacy configs section, new configs should follow the naming convention of azkaban.server.<rest of the name> for server configs.
 
-    // The property is used for the web server to get the host name of the executor when running in SOLO mode.
-    public static final String EXECUTOR_HOST = "executor.host";
-
     // The property is used for the web server to get the port of the executor when running in SOLO mode.
     public static final String EXECUTOR_PORT = "executor.port";
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -108,8 +108,9 @@ public class ExecutorManager extends EventHandler implements
   private List<String> filterList;
   private Map<String, Integer> comparatorWeightsMap;
   private long lastSuccessfulExecutorInfoRefresh;
-  private ExecutorService executorInforRefresherService;
+  private final ExecutorService executorInfoRefresherService;
   private Duration sleepAfterDispatchFailure = Duration.ofSeconds(1L);
+  private boolean initialized = false;
 
   @Inject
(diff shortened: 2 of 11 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -129,30 +130,33 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader executorLoader
...
     this.cleanerThread = new CleanerThread(executionLogsRetentionMs);
+    this.executorInfoRefresherService = createExecutorInfoRefresherService();
+  }
+
+  void initialize() throws ExecutorManagerException {
+    if (initialized) {
+      return;
+    }
+    initialized = true;
+    this.setupExecutors();
```
juhoautio: Moved this away from the constructor to fix AzkabanSingleServer (azkaban-solo). Because in AzkabanSingleServer Guice provisioning first creates instances of ExecutorManager etc., then starts the executor servlet, and finally web servlet. Executors can't be loaded (in ExecutorManager of azkaban-web) before the executor servlet has been started.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -232,11 +233,16 @@ private void setupMultiExecutorMode() {
    */
   @Override
   public void setupExecutors() throws ExecutorManagerException {
+    checkMultiExecutorMode();
     this.activeExecutors.setupExecutors();
   }
 
-  private boolean isMultiExecutorMode() {
-    return isMultiExecutorMode(this.azkProps);
+  private void checkMultiExecutorMode() {
```
HappyRay: Have you considered retiring this config? What are the downsides?

HappyRay:
Thanks for the pull request! I want to simplify this logic for a long time now.

HappyRay:
@jamiesjc since you are familiar with this part of the code, do you mind reviewing this change?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -129,30 +130,33 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader executorLoader
...
+  void initialize() throws ExecutorManagerException {
+    if (initialized) {
+      return;
+    }
+    initialized = true;
+    this.setupExecutors();
+    this.loadRunningExecutions();
+    this.queuedFlows = new QueuedExecutions(
+        this.azkProps.getLong(ConfigurationKeys.WEBSERVER_QUEUE_SIZE, 100000));
+    // The default threshold is set to 30 for now, in case some users are affected. We may
```
jamiesjc: This comment is for `maxConcurrentRunsOneFlow`, please move it to the right place.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jamiesjc: +1  retiring this config.

juhoautio: I wanted to make sure that users don't accidentally start azkaban with a new version thinking that their configured `executor.host` would be used. This will make it clear for the users that local mode won't work any more and they have to update their conf to start using the only supported mode (where the executor hostnames are automatically resolved on executor side & read from the db table on azkaban-web).

Eventually this code can be removed once confident enough that all active users have already updated to some version new enough to have this change, so that they must've acknowledgingly switched to multi-executor mode.

juhoautio:
@jamiesjc added commit ^

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -156,9 +156,6 @@
 
     // Legacy configs section, new configs should follow the naming convention of azkaban.server.<rest of the name> for server configs.
 
-    // The property is used for the web server to get the host name of the executor when running in SOLO mode.
-    public static final String EXECUTOR_HOST = "executor.host";
-
     // The property is used for the web server to get the port of the executor when running in SOLO mode.
     public static final String EXECUTOR_PORT = "executor.port";
```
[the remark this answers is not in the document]
jamiesjc: I did a code search but didn't see where `EXECUTOR_PORT` is used on executor side. Can you point to me the location? If it is still needed, consider updating the comment for this property?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
HappyRay: LGTM.

Could you add a TODO to retire this required config?

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
juhoautio: @jamiesjc thanks for noticing. I've added a commit to clean up all usages of executor port related properties and constants.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
juhoautio: > Could you add a TODO to retire this required config?

@HappyRay, done.

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
jamiesjc: Thanks for cleaning up the executor port.

PR #1990, opened 2018-10-18
Disable noisy logger "property x.jor is empty"

juhoautio:
It's totally normal that `.jor` doesn't exist in the DB if there's no override, not worth logging on WARN level.

https://github.com/azkaban/azkaban/blob/cefb48346c21c3163e4fed0ebd8e5070c3f2c227/az-core/src/main/java/azkaban/Constants.java#L54-L55

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -905,7 +905,7 @@ public class JdbcProjectImpl implements ProjectLoader {
                   propsName);
 
       if (properties == null || properties.isEmpty()) {
-        logger.warn("Project " + projectId + " version " + projectVer + " property " + propsName
+        logger.debug("Project " + projectId + " version " + projectVer + " property " + propsName
             + " is empty.");
         return null;
       }


juhoautio:
Anyone? Tiny one line change..

juhoautio:
Please?

juhoautio:
@chengren311 maybe? :)

juhoautio:
@HappyRay @jamiesjc ?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -905,7 +905,7 @@ public Props fetchProjectProperty(final int projectId, final int projectVer,
                   propsName);
 
       if (properties == null || properties.isEmpty()) {
-        logger.warn("Project " + projectId + " version " + projectVer + " property " + propsName
+        logger.debug("Project " + projectId + " version " + projectVer + " property " + propsName
```
chengren311: it's not just fetching override properties here, right?

juhoautio: Correct

juhoautio:
Hi, any chance to merge this one-liner in to avoid the flooded logs?

juhoautio:
@HappyRay @jamiesjc how about? I wonder, aren't your logs also full of this warning..

jamiesjc:
I agree it is not worth logging on WARN level for `.jor is empty` since not all jobs have the override property. However, this also logs for normal job/properties file if they do not exist. Do you think it might be helpful for debugging or can be totally removed together?

juhoautio:
@jamiesjc in _practice_ this has been only spamming a lot with `.jor`. I would keep the log line though, so that it can be enabled in logger conf for debugging if needed in some case.

By the way, I now tried to analyze a bit how the method is used otherwise. It doesn't seem like all corner cases are handled perfectly. For example, if `ProjectManagerServlet#handleJobPage` is called and job is concurrently deleted between the calls of `projectManager.getProject` and `projectManager.getProperties`, `handleJobPage` can fail with an NPE. Doesn't seem like a major issue to me but.. For example in this case there's no real need to have the warn about missing properties though, it's not really needed to find out what went wrong.

PR #1992, opened 2018-10-18
Make ActiveExecutors a singleton

juhoautio:
It was only used by one class that's singleton itself, so no harm done for now. But this is how it should be, in case it will be injected to more than one instances later.


HappyRay:
Thanks

PR #1993, opened 2018-10-18
Improve error message for a cleaned project version

juhoautio:
1st commit shows the problem:
When trying to execute a cleaned project version, an exception is thrown that says that hash code comparison failed.

2nd commit improves the error message in case of trying to read a cleaned (deleted) version. It also fails _faster_ (don't try to generate a hashcode from 0 chunks).

\+ Move `project.version.retention` to `Constants.java`.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -230,6 +230,9 @@ public class Constants {
     // allowed max size of shared project dir in MB
     public static final String PROJECT_DIR_MAX_SIZE_IN_MB = "azkaban.project_cache_max_size_in_mb";
 
+    // how many older versions of project files are kept in DB before deleting them
+    public static final String PROJECT_VERSION_RETENTION = "project.version.retention";
+
     // number of rows to be displayed on the executions page.
     public static final String DISPLAY_EXECUTION_PAGE_SIZE = "azkaban.display.execution_page_size";
   }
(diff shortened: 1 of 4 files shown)


juhoautio:
Could someone review, please? Created 11 days ago.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -230,6 +230,9 @@
     // allowed max size of shared project dir in MB
     public static final String PROJECT_DIR_MAX_SIZE_IN_MB = "azkaban.project_cache_max_size_in_mb";
 
+    // how many older versions of project files are kept in DB before deleting them
+    public static final String PROJECT_VERSION_RETENTION = "project.version.retention";
```
chengren311: is this equivalent to AZKABAN_STORAGE_ARTIFACT_MAX_RETENTION?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -524,7 +532,10 @@ public ProjectFileHandler getUploadedFile(final int projectId, final int version
     if (Arrays.equals(projHandler.getMd5Hash(), md5)) {
       logger.info("Md5 Hash is valid");
     } else {
-      throw new ProjectManagerException("Md5 Hash failed on retrieval of file");
+      throw new ProjectManagerException(
```
chengren311: Consider string.format?

az-core/src/main/java/azkaban/Constants.java
chengren311: thanks for the refactor!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
juhoautio: Done (added commit)

chengren311:
1st commit shows the problem:
When trying to execute a cleaned project version, an exception is thrown that says that hash code comparison failed.

2nd commit improves the error message in case of trying to read a cleaned (deleted) version. It also fails faster (don't try to generate a hashcode from 0 chunks).

+ Move project.version.retention to Constants.java.

PR #1994, opened 2018-10-19
Don't delete (clean) project versions with running executions

juhoautio:
Builds on https://github.com/azkaban/azkaban/pull/1993 which should be merged first.

----

When a new version of a project is uploaded, the old versions are cleaned up (zip files are deleted from the DB).

This way the project versions with running executions are skipped in the old versions cleanup.

Even those versions will be eventually cleaned up once those executions have finished, because the basic filter is `version < ?`


HappyRay:
By running executions do you mean the executions that have started consuming the old version of the project? 

Could you make it more clear what the symptoms are in this PR description? I understand they are described in the referenced PR. IMHO, it is better to make it more explicit here as well.

juhoautio:
@HappyRay I wrote a brand new PR description. Is that clear enough?

juhoautio:
@HappyRay, in case you missed this ^

juhoautio:
Please, could someone review? Would be nice to have this fix.

juhoautio:
@chengren311 thanks for merging 
https://github.com/azkaban/azkaban/pull/1993. I have now rebased this one to have a single commit.

juhoautio:
@HappyRay, if you have time, could you revisit your comment https://github.com/azkaban/azkaban/pull/1994#issuecomment-431398770 and see if I've addressed your concern?

juhoautio:
Please, anyone?

burgerkingeater:
@juhoautio, sorry for the delay. We currently don’t have enough bandwidth as some of us are on vacation. I will try to take a look when I have a chance.

juhoautio:
> I remember the executor will hard link the project directory when it starts execution. I wonder how that design doesn't prevent the issue you observed.

@HappyRay that doesn't play a part in this. This is about a timing issue where the project version has been cleaned up from the DB before an executor has handled the dispatch request.

HappyRay:
I see. I will need to read the code again to refresh my memory first.  It may take a while before I can get to it.

@burgerkingeater could you also take a look?

juhoautio:
@burgerkingeater?

juhoautio:
@burgerkingeater did you see this? ^

juhoautio:
@HappyRay @burgerkingeater @jamiesjc what's up? Could you say something at least?

jamiesjc:
@juhoautio To be more specific, does the timing issue happens when `azkaban.project.ProjectLoader#fetchProjectMetaData` is called to fetch the project version but that specific version has already been deleted due to new project uploads? 
We haven't seen this kind of issue before, but would like to know how often this would happen.

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -59,16 +62,19 @@
   private final FlowLoaderFactory flowLoaderFactory;
   private final File tempDir;
   private final int projectVersionRetention;
+  private final RunningExecutions runningExecutions;
 
   @Inject
   AzkabanProjectLoader(final Props props, final ProjectLoader projectLoader,
-      final StorageManager storageManager, final FlowLoaderFactory flowLoaderFactory) {
+      final StorageManager storageManager, final FlowLoaderFactory flowLoaderFactory,
+      final RunningExecutions runningExecutions) {
```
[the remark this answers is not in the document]
juhoautio: Will do. I suppose that even though `fetchActiveFlows` visits the DB, calling it on every project upload is not too much.

juhoautio:
> does the timing issue happens when azkaban.project.ProjectLoader#fetchProjectMetaData is called to fetch the project version but that specific version has already been deleted due to new project uploads?

I'm not entirely sure what you mean by that, but in practice I've hit this in `JdbcProjectImpl#getUploadedFile`. For reference, maybe check this earlier PR: https://github.com/azkaban/azkaban/pull/1993/files

> We haven't seen this kind of issue before, but would like to know how often this would happen.

Yes, that's not surprising because in normal operation this would be rare to happen.

It only happens if there's a significant delay in dispatching (and on the other hand many project uploads in the meanwhile).

What makes the issue more prominent in our setup is that we have implemented _moving existing executions from an inactive executor to an active executor_. For any long running executions (that are moved) it can happen much more easily that there have been enough uploads for the project to clean up the version that a running execution needs.

jamiesjc:
Hi @juhoautio, I have a few questions: 

> It only happens if there's a significant delay in dispatching (and on the other hand many project uploads in the meanwhile).

What would cause the significant delay? Is it possible to solve the delay problem itself instead of fixing it in deleting project versions?

> What makes the issue more prominent in our setup is that we have implemented moving existing executions from an inactive executor to an active executor. 
By inactive executor, do you mean the executor is still running but the active flag is set to false in DB? Why do you need to move executions from inactive executor to an active executor?

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -236,11 +246,19 @@ private void uploadFlowFilesRecursively(final File projectDir, final Project pro
   }
 
   private void cleanUpProjectOldInstallations(final Project project)
-      throws ProjectManagerException{
+      throws ProjectManagerException, ExecutorManagerException {
     log.info("Cleaning up old install files older than "
         + (project.getVersion() - this.projectVersionRetention));
+    // TODO fetch high level metadata without DAG object
+    Map<Integer, Pair<ExecutionReference, ExecutableFlow>> activeFlows = this.executorLoader
```
jamiesjc: What about queued flows? As you mentioned, there might be a delay in dispatching to executors and the executions are queued on the web server. Should we exclude the project versions from being deleted for those queued flows as well? If that's the case, you can use `this.executorLoader.fetchUnfinishedFlows` which will include the queued flows.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -42,6 +42,9 @@ ExecutableFlow fetchExecutableFlow(int execId)
   Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedFlows()
       throws ExecutorManagerException;
 
+  Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedExecutions()
```
jamiesjc: The method names `fetchUnfinishedExecutions` and `fetchUnfinishedFlows` are too similar to each other. Would it be better to use the name `fetchUnfinishedFlowsMetadata` since you are only fetching metadata info about the flow?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -46,41 +48,60 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
...
+      executor = null;
+    } else {
+      final boolean executorStatus = rs.getBoolean("executorStatus");
+      executor = new Executor(executorId, host, port, executorStatus);
+    }
+    final ExecutionReference ref = new ExecutionReference(exFlow.getExecutionId(), executor);
+    return new Pair<>(ref, exFlow);
+  }
+
+  private static Pair<ExecutionReference, ExecutableFlow> getExecutionHelper(
```
jamiesjc: Consider the name `getExecutableFlowMetadataHelper`?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -177,7 +214,46 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
...
+    }
+  }
+
+  // TODO create unit test
+  @VisibleForTesting
+  static class FetchActiveExecutions implements
+      ResultSetHandler<Map<Integer, Pair<ExecutionReference, ExecutableFlow>>> {
+
+    // Select flows that are not in finished status
+    private static final String FETCH_UNFINISHED_EXECUTIONS =
```
jamiesjc: Same comment about naming.

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -46,41 +48,60 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
...
+  }
+
+  private static Pair<ExecutionReference, ExecutableFlow> getExecutionHelper(
+      final ResultSet rs) throws SQLException {
+    final Flow flow = new Flow(rs.getString("flow_id"));
+    final Project project = new Project(rs.getInt("project_id"), null);
+    project.setVersion(rs.getInt("version"));
+    final ExecutableFlow exFlow = new ExecutableFlow(project, flow);
+    exFlow.setExecutionId(rs.getInt("exec_id"));
+    exFlow.setStatus(Status.fromInteger(rs.getInt("status")));
```
jamiesjc: What about "submit_user"?

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -186,9 +186,9 @@ Props fetchProjectProperty(Project project, String propsName)
       throws ProjectManagerException;
 
   /**
-   * Cleans all project versions less tha
+   * Cleans all project versions less than the provided version.
```
jamiesjc: Maybe mention about excluding the versions of unfinished flows in the comment?

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -336,6 +336,24 @@ public void testFetchUnfinishedFlows() throws Exception {
     assertTwoFlowSame(unfinishedFlows.get(flows.get(0).getExecutionId()).getSecond(), flows.get(0));
   }
 
+  @Test
+  public void testFetchUnfinishedExecutions() throws Exception {
```
jamiesjc: `testFetchUnfinishedFlowsMetadata`?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -78,7 +78,13 @@ public ExecutableFlow fetchExecutableFlow(final int execId)
...
   public Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedFlows()
       throws ExecutorManagerException {
-    return new ConcurrentHashMap<>();
+    return this.activeFlows;
+  }
+
+  @Override
+  public Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedExecutions()
+      throws ExecutorManagerException {
+    return this.activeFlows;
```
jamiesjc: unfinishedFlows might not be the same as activeFlows. Will it cause confusion here?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
juhoautio: > What about "submit_user"?

Sure, added it for completeness

juhoautio:
Great reviewing once again. Sorry for being sloppy with the naming so that you had to point those out.

juhoautio:
@jamiesjc are you keeping this open until the next release or something, or just missed merging?

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -98,6 +120,22 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
     }
   }
 
+  /**
+   * Fetch unfinished flows similar to {@link #fetchUnfinishedFlows}, excluding flow data.
```
juhoautio: @jamiesjc, do you have some practical experience on the performance impact on azkaban-web after switching to get unfinished executions from DB instead of caching? I mean situation when there are at least some, say 100 concurrent executions and those are for considerable big flows (with many nodes & properties) so that the flow data is big.

I have a vague memory that it can take long to load even a couple of flows, when the data is included.. I think I saw that in azkaban-web startup when it's initializing the cache of unfinished flows from DB, it could be something like 10 seconds. For example, if the flow updater thread is now fetching the list in a loop and it takes that long, it may become unmanageable with any higher number of parallel executions. Just suspicions so far, I hope you have some knowledge based on manual testing?

Not sure how big difference it makes to exclude the flow data like here, but maybe this should be used as much as possible and only fetch the full flow data when it's required?

jamiesjc: @juhoautio Good point. I've tested the performance on local mysql DB with hundreds of running flows and didn't see significant degradation. I'm about to test it in our staging environment before enabling it in production. 
I agree we should fetch full flow data only when it's needed for better DB performance. Thanks for reminding me of this. I'll share any test result I get later.

PR #1995, opened 2018-10-19
Evict executions on unresponsive executors (by default)

juhoautio:
See discussion in https://github.com/azkaban/azkaban/pull/1661#issuecomment-431197754.

However, I don't think the problem is in the application design. If an executor instance is removed, the deployment strategy should make sure that the corresponding executor entry is deleted from the DB. Or then there's some actual bug that I haven't seen before.

Any way, enabling evicting again for now. Users (like me) can set `azkaban.evictUnresponsiveExecutions=false` to keep the previous behaviour.

There is actually one nice-to-have improvement to be done before stopping evictions by default again: automatic refreshing of removed executors from the DB. It was blocked by some preliminary work before, but now this is available to be built upon:
https://github.com/azkaban/azkaban/pull/1975#discussion_r225306559.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
+++ b/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
@@ -19,6 +19,7 @@ package azkaban.executor;
 import azkaban.alert.Alerter;
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.Pair;
+import azkaban.utils.Props;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


juhoautio:
@HappyRay here it is

juhoautio:
@HappyRay is this still urgent?

juhoautio:
@HappyRay ref 
https://github.com/azkaban/azkaban/pull/1661#issuecomment-431197754, I thought you were in a need of this change. Did you find a way to live with the current code maybe?

HappyRay:
We decided to try an alternative fix first #2007 
The deployment is taking place. So far so good. But it takes a while for the symptoms to show up. 

This change will be our backup plan.

juhoautio:
@HappyRay this should provide an even better alternative: https://github.com/azkaban/azkaban/pull/2016. I wouldn't want to see additional delay in executor shutdown. Indeed https://github.com/azkaban/azkaban/pull/2016 makes it totally redundant.

HappyRay:
@chengren311 does the #2007 slow down the normal shutdown process too?

PR #2002, opened 2018-10-24
Shut down single server if startup fails

juhoautio:
The single server first starts executor server, then web server.

If web server startup failed, the single server process was left running, making it harder to realize that there was an issue in startup.

This PR makes it easier to spot a startup error by stopping JVM in case of a startup error.


HappyRay:
Thanks. I added your explanation to the merged commit message.

PR #2006, opened 2018-10-25
Only incorporate user pig additional jar setting or system's but not both

kunkun-tang:
We ran into a bug in the production environment. Users' jars cannot override default pig libraries. It was found that AZ always adds pig additional jar if pig job type defines it under plugin.properties. The fix is straightforward. If a user defines the property, we use it directly by not including system default settings.

To be tested in the staging environment.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopPigJob.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopPigJob.java
@@ -319,29 +319,30 @@ public class HadoopPigJob extends JavaProcessJob {
     return udfImports;
   }
 
-  protected List<String> getAdditionalJarsList() {
-    List<String> additionalJars = new ArrayList<String>();
-    mergeAdditionalJars(additionalJars, PIG_ADDITIONAL_JARS);
-    mergeAdditionalJars(additionalJars, DEFAULT_PIG_ADDITIONAL_JARS);
-    return additionalJars;
-  }
-
   /**
    * Merging all additional jars first from user specified/plugin.properties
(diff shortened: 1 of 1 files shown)


xkrogen:
Hey @kunkun-tang , I'm not sure if this is the right approach. Previously, if a user specified `pig.additional.jars`, they would still get all of the default JARs. I think it's probably common for people to use this setting to add only a few JARs, e.g. their custom UDFs. Now, if you specify additional JARs, you have to be sure to include the defaults as well.

My original thought was just to allow the user to _additionally_ specify `default.pig.additional.jars`, which would allow them to override the system-provided defaults. I think this change is much safer, as users shouldn't really have been specifying `default.pig.additional.jars` in the past, whereas they certainly are using the `pig.additional.jars` property.

kunkun-tang:
Ah, Got you now. That makes sense to me. Let me update the RB. @xkrogen

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopPigJob.java
```diff
@@ -319,29 +319,39 @@ protected String getScript() {
...
    */
-  private void mergeAdditionalJars(List<String> additionalJars,
-      String additionalJarProperty) {
-    List<String> jobJars =
-        getJobProps().getStringList(additionalJarProperty, null, ",");
-    List<String> typeJars =
-        getSysProps().getStringList(additionalJarProperty, null, ",");
-    if (jobJars != null) {
-      additionalJars.addAll(jobJars);
+  protected List<String> getAdditionalJarsList() {
```
kunkun-tang: Do you want to review this again? @xkrogen

PR #2007, opened 2018-10-25
delaying shutting executor down

chengren311:
Executor will be in shutting down mode when its shutdown API is called. Executor will wait for all the ongoing flows to complete then shut itself down. Web server talks to executor periodically to fetch the status of each flow in its running flow cache. This PR accommodates a grace period between completion of all flows and killing the executor process, in order to allow web server to finalize flow status updating.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
@@ -62,6 +62,7 @@ import java.nio.file.Paths;
 import java.security.Permission;
 import java.security.Policy;
 import java.security.ProtectionDomain;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.TimeZone;
@@ -505,6 +506,14 @@ public class AzkabanExecutorServer {
     return getHost() + ":" + getPort();
   }
 
+  private void sleep(final Duration duration) {
+    try {
+      Thread.sleep(duration.toMillis());
+    } catch (final InterruptedException e) {
+      logger.error(e);
+    }
+  }
+
   /**
    * Shutdown the server. - performs a safe shutdown. Waits for completion of current tasks - spawns
    * a shutdown thread and returns immediately.
@@ -512,12 +521,8 @@ public class AzkabanExecutorServer {
   public void shutdown() {
     logger.warn("Shutting down AzkabanExecutorServer...");
     new Thread(() -> {
-      try {
-        // Hack: Sleep for a little time to allow API calls to complete
-        Thread.sleep(2000);
-      } catch (InterruptedException e) {
-        logger.error(e);
-      }
+      // Hack: Sleep for a little time to allow API calls to complete
+      sleep(Duration.ofSeconds(2));
       shutdownInternal();
     }, "shutdown").start();
   }
@@ -529,6 +534,9 @@ public class AzkabanExecutorServer {
    */
   private void shutdownInternal() {
     getFlowRunnerManager().shutdown();
+    // Sleep for an hour to wait for web server updater thread
+    // {@link azkaban.executor.RunningExecutionsUpdaterThread#updateExecutions} to finalize updating
+    sleep(Duration.ofHours(1));
     // trigger shutdown hook
     System.exit(0);
   }


juhoautio:
What about refreshing executors automatically on az-web if the executor can't be reached? If the executor was shut down gracefully, it should have removed itself from the DB any way.

HappyRay:
The executor was shut down gracefully and the entry removed.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -505,19 +506,23 @@ public String getExecutorHostPort() {
...
   public void shutdown() {
     logger.warn("Shutting down AzkabanExecutorServer...");
     new Thread(() -> {
-      try {
-        // Hack: Sleep for a little time to allow API calls to complete
-        Thread.sleep(2000);
-      } catch (InterruptedException e) {
-        logger.error(e);
-      }
+      // Hack: Sleep for a little time to allow API calls to complete
```
HappyRay: Make it a Todo and add more details? Copying the description from the PR is fine.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -505,19 +506,23 @@ public String getExecutorHostPort() {
     return getHost() + ":" + getPort();
   }
 
+  private void sleep(final Duration duration) {
```
HappyRay: I like the usage of Duration here.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -505,19 +506,23 @@ public String getExecutorHostPort() {
...
     logger.warn("Shutting down AzkabanExecutorServer...");
     new Thread(() -> {
-      try {
-        // Hack: Sleep for a little time to allow API calls to complete
-        Thread.sleep(2000);
-      } catch (InterruptedException e) {
-        logger.error(e);
-      }
+      // Hack: Sleep for a little time to allow API calls to complete
+      sleep(Duration.ofSeconds(2));
```
HappyRay: Is it just a refactoring? 
Nice.

HappyRay: I like shorter methods.

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -529,6 +534,9 @@ public void shutdown() {
    */
   private void shutdownInternal() {
     getFlowRunnerManager().shutdown();
+    // Sleep for an hour to wait for web server updater thread
```
HappyRay: Make it a Todo and add more details? Copying the description from the PR is fine.

HappyRay:
How did you test this change?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
chengren311: It might allow the shutdown api to return before invoking system.exit(), it's just speculation since the original author has left.

chengren311:
@HappyRay tested on staging cluster, 
1. launch a flow
2. shutdown executor
3. watch web server log to see if there's exception when updating the flow status.

chengren311:
@juhoautio web server keeps a list of running flow and its corresponding executor. Even if the executor list is fresh, web server will still try to reach to inactive executor when iterating the running flow list.

juhoautio:
>> What about refreshing executors automatically on az-web if the executor can't be reached? If the executor was shut down gracefully, it should have removed itself from the DB any way.

> The executor was shut down gracefully and the entry removed.

Yes, my point was that a better fix would be to make the updater check from DB does the executor exist any more if it can't be reached. And if the executor has been removed, just finalize the execution instead of trying to call executor to get the same information.

Why would this be better?
- executor can shut down as soon as it has finished its work
- updater can't miss the update depending on if the grace period was long enough

juhoautio:
@chengren311 no, if a flow's executor has been removed from the DB, the execution gets finalized (and no trying to call executor for updates any more). If updater would also update executor list on the fly, the result would be that it would finalize any executions left behind by removed executors.

HappyRay:
> Yes, my point was that a better fix would be to make the updater check from DB does the executor exist any more if it can't be reached. And if the executor has been removed, just finalize the execution instead of trying to call executor to get the same information.

I am not sure this solution may subject to a race condition:
the web server checks a running flow status for an executor instance that is shutting down. It gets an error because the executor is shutting down. It checks the DB and sess that the instance is still there because the executor instance hasn't removed itself from the DB yet. According to the code below
https://github.com/azkaban/azkaban/blob/c564f24837cf4ea1cf0c1d6469eef735ba81d5f0/azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java#L185 

This race condition may not be possible. But we need to be sure. What do you think?

@jamiesjc told me that you and she were discussing other alternative fixes. She mentioned the above too. Please continue to explore better alternatives including your suggestion above.

I suggested this "simpler" approach for now to
a. unblock the release
b. allow us to test if this issue indeed disappears because your earlier change makes the issue more apparent. 

I am still not sure if emailing users, in this case, is a better choice in notifying administrators though.

@jamiesjc is writing a new design proposal that would hopefully simplify the dispatching logic. She will share it with you and others here to get feedback.

juhoautio:
In the updater a couple of failed attempts are tolerated before taking action. I think checking executor's existence from the DB should work fine.

True that it would be better to notify admins if an executor is deemed unresponsive. While there's an on-going issue like that, users should only be notified with SLA alerts if anything. However, after admins have resolved the issue (typically remove executor from DB), then the executions would be finalized as failed and users would get the alerts about it if configured.

PR #2016, opened 2018-11-05
Finalize execution if executor doesn't exist

juhoautio:
Problem scenario:
1. An executor is stopped [1]
2. Executor is removed from DB table: `executors` [2]
3. Updater thread tries to get an update of an execution that was running on the removed executor
  - Request fails as executor is not running any more
  - Updater sends alert emails about unresponsive executor

Fix (this PR):
- **If update call fails against an executor, check from the DB directly if the executor still exists or not. If not, finalize the execution.** 
- If executor still exists (or couldn't be checked), do as before, ie. periodically alert about unresponsive executor.

----

[1] This can be a graceful shutdown or a sudden crash, doesn't matter

[2] Executor can be removed from the DB in the following ways
- Graceful shutdown: executor removes itself
- Unexpected shutdown – admin removes executor from DB manually, or external monitoring software does it after seeing that the instance has been stopped

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
+++ b/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
@@ -81,7 +81,7 @@ public class RunningExecutionsUpdater {
       if (!executorOption.isPresent()) {
         for (final ExecutableFlow flow : entry.getValue()) {
           logger.warn("Finalizing execution " + flow.getExecutionId()
-              + ". Executor id of this execution doesn't exist");
+              + ". Executor is removed");
           finalizeFlows.add(flow);
         }
         continue;
@@ -139,7 +139,7 @@ public class RunningExecutionsUpdater {
       final ArrayList<ExecutableFlow> finalizeFlows) {
     logger.error("Failed to get update from executor " + executor.getHost(), e);
     boolean sendUnresponsiveEmail = false;
-    final boolean executorRemoved = executorRemoved(executor.getId());
+    final boolean executorRemoved = isExecutorRemoved(executor.getId());
     for (final ExecutableFlow flow : entry.getValue()) {
       final Pair<ExecutionReference, ExecutableFlow> pair =
           this.runningExecutions.get().get(flow.getExecutionId());
@@ -168,7 +168,7 @@ public class RunningExecutionsUpdater {
     }
   }
 
-  private boolean executorRemoved(final int id) {
+  private boolean isExecutorRemoved(final int id) {
     Executor fetchedExecutor;
     try {
       fetchedExecutor = this.executorLoader.fetchExecutor(id);
--- a/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
+++ b/azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
@@ -81,7 +81,7 @@ public class RunningExecutionsUpdater {
       if (!executorOption.isPresent()) {
         for (final ExecutableFlow flow : entry.getValue()) {
           logger.warn("Finalizing execution " + flow.getExecutionId()
-              + ". Executor is removed");
+              + ". Executor id of this execution doesn't exist");
           finalizeFlows.add(flow);
         }
         continue;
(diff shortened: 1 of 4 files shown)


juhoautio:
@jamiesjc, as discussed.

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
```diff
@@ -133,24 +135,31 @@ public void updateExecutions() {
...
-      final ExecutionReference ref = pair.getFirst();
-      ref.setNextCheckTime(DateTime.now().getMillis() + this.errorThreshold);
-      ref.setNumErrors(ref.getNumErrors() + 1);
-      if (ref.getNumErrors() == this.numErrorsBeforeUnresponsiveEmail
-          || ref.getNumErrors() % this.numErrorsBetweenUnresponsiveEmail == 0) {
-        // if any of the executions has failed many enough updates, alert
-        sendUnresponsiveEmail = true;
+      if (executorRemoved) {
+        logger.warn("Finalizing execution " + flow.getExecutionId()
+            + ". Executor id of this execution doesn't exist");
```
jamiesjc: The logging message here is not accurate, please update it to reflect "executor is removed".

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
```diff
@@ -159,6 +168,18 @@ private void handleException(final Entry<Optional<Executor>, List<ExecutableFlow
     }
   }
 
+  private boolean executorRemoved(final int id) {
```
jamiesjc: Is it better to name it `isExecutorRemoved`?

azkaban-common/src/test/java/azkaban/executor/RunningExecutionsUpdaterTest.java
```diff
@@ -108,10 +113,33 @@ public void updateExecutionsUpdateCallFails() throws Exception {
     }
     verify(this.mailAlerter).alertOnFailedUpdate(
         this.activeExecutor, Collections.singletonList(this.execution), API_CALL_EXCEPTION);
+    verifyZeroInteractions(this.executionFinalizer);
+  }
 
-    // TODO change to checking if executor exist in the DB any more
+  /**
+   * Shuold finalize execution if executor doesn't exist in the DB.
```
jamiesjc: Type -> Should

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
juhoautio: > The logging message here is not accurate, please update it to reflect "executor is removed".

I changed that as requested.

Although I don't see how the message `Executor id of this execution doesn't exist` is not accurate. To me it seems to say the same thing + a bit more context. To be specific, this part of code _doesn't necessarily know that executor has been removed_. All it knows is that the executor id that the execution is assigned to _doesn't exist_. From this code's point of view it could also be that such executor id never existed :) Although with current Azkaban code that shouldn't ever happen of course – it's not possible that an execution is assigned an executor id that _never existed_.

juhoautio:
@jamiesjc, updated ^

azkaban-common/src/main/java/azkaban/executor/RunningExecutionsUpdater.java
jamiesjc: I think we are talking about different pieces of the code. Because originally the logger messages were the same for line 84 and line 152, I suggest changing the one on line 152 to reflect the reason "executor is removed" not line 84.

juhoautio: Oh, sorry, I already forgot that there are two similar cases. I think they are about the same thing, could you check please? I believe the warn message could be unified for these. Should I extract a method? What would you want the message to be?

jamiesjc: I think they are still slightly different. The previous one checks whether the executor is null in the running executions cache. And this one checks if the executor exists in DB. For better debugging purpose, I would suggest keeping two different logging messages.

juhoautio: > For better debugging purpose, I would suggest keeping two different logging messages.

Great thinking 👍 I'll just flip them around then to match with your original request. Sorry for my mistake with the wrong line. I would still say that "doesn't exist" is more exact than "is removed" (which could be a stretch) but both work for me..

> The previous one checks whether the executor is null in the running executions cache

That can only happen immediately after startup btw, because running executions are not updated from the DB afterwards.

PR #2019, opened 2018-11-09
New 'expand/collapse all flows' menu options on flow views

ypadron-in:
Improvement:
-The new menu options on Flow, Flow Execution and Schedule/Execute pages prevent users from having to unfold deeply nested flows by clicking one by one.
-Users will be able to browse flows and enable or disable job executions more easily.
-Options are shown and work on projects using both flow 1.0 and flow 2.0

screen shot 2018-11-09 at 5 42 21 pm
screen shot 2018-11-09 at 5 41 11 pm
screen shot 2018-11-09 at 5 44 58 pm

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
+++ b/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
@@ -81,18 +81,34 @@ var nodeClickCallback = function (event, model, node) {
     var flowRequestURL = contextURL + "/manager?project=" + projectName
         + "&flow=" + node.flowId;
     if (node.expanded) {
-      menu = [{
-        title: "Collapse Flow...", callback: function () {
-          model.trigger("collapseFlow", node);
+      menu = [
+        {
+          title: "Collapse Flow...", callback: function () {
+            model.trigger("collapseFlow", node);
+          }
+        },
+        {
+          title: "Collapse all Flows...", callback: function () {
+            model.trigger("collapseAllFlows", node);
+            model.trigger("resetPanZoom");
+          }
         }
-      }];
+      ];
     }
     else {
-      menu = [{
-        title: "Expand Flow...", callback: function () {
-          model.trigger("expandFlow", node);
+      menu = [
+        {
+          title: "Expand Flow...", callback: function () {
+            model.trigger("expandFlow", node);
+          }
+        },
+        {
+          title: "Expand all Flows...", callback: function () {
+            model.trigger("expandAllFlows", node);
+            model.trigger("resetPanZoom");
+          }
         }
-      }];
+      ];
     }
 
     $.merge(menu, [
@@ -100,30 +116,30 @@ var nodeClickCallback = function (event, model, node) {
       {break: 1},
       {
         title: "Open Flow...", callback: function () {
-        window.location.href = flowRequestURL;
-      }
+          window.location.href = flowRequestURL;
+        }
       },
       {
         title: "Open Flow in New Window...", callback: function () {
-        window.open(flowRequestURL);
-      }
+          window.open(flowRequestURL);
+        }
       },
       {break: 1},
       {
         title: "Open Properties...", callback: function () {
-        window.location.href = requestURL;
-      }
+          window.location.href = requestURL;
+        }
       },
       {
         title: "Open Properties in New Window...", callback: function () {
-        window.open(requestURL);
-      }
+          window.open(requestURL);
+        }
       },
       {break: 1},
       {
         title: "Center Flow", callback: function () {
-        model.trigger("centerNode", node);
-      }
+          model.trigger("centerNode", node);
+        }
       }
     ]);
   }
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
```diff
@@ -81,49 +81,65 @@ var nodeClickCallback = function (event, model, node) {
...
-        title: "Collapse Flow...", callback: function () {
-          model.trigger("collapseFlow", node);
+      menu = [
+        {
+          title: "Collapse Flow...", callback: function () {
+            model.trigger("collapseFlow", node);
+          }
+        },
+        {
+          title: "Collapse all Flows...", callback: function () {
```
jamiesjc: Consider changing from lower case to upper case: "Collapse All Flows..."?

jamiesjc:
You can attach some screenshots to demonstrate what the UI looks like with the new change.
Also mention the new change will apply to both flow 1.0 and flow 2.0.

azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
```diff
@@ -235,21 +251,34 @@ var graphClickCallback = function (event, model) {
       + flowId;
 
   var menu = [
+    {
+      title: "Expand all Flows...", callback: function () {
```
jamiesjc: Same: consider all upper cases.

azkaban-web-server/src/web/js/azkaban/view/svg-graph.js
```diff
@@ -439,6 +441,50 @@ azkaban.SvgGraphView = Backbone.View.extend({
...
+      }
+    }
+  },
+
+  collapseAllFlows: function (node) {
+    if (node) {
+      // collapse all embedded flows inside given node
+
+      // collapse already rendered nodes of type flow
+      if (node.type == 'flow' && node.gNode) {
```
jamiesjc: Why `node.gNode `is checked in `collapseAllFlows` but not `expandAllFlows`?

azkaban-web-server/src/web/js/azkaban/view/svg-graph.js
```diff
@@ -439,6 +441,50 @@ azkaban.SvgGraphView = Backbone.View.extend({
...
+
+      if (node.nodes && node.nodes.length > 0) {
+        for (var i = 0; i < node.nodes.length; ++i) {
+          this.collapseAllFlows(node.nodes[i]);
+        }
+      }
+    } else {
+      // collapse all embedded flows in the graph
+      var nodes = this.model.get("data").nodes;
+      for (var i = 0; i < nodes.length; ++i) {
```
jamiesjc: Why is necessary to check `if (node.nodes && node.nodes.length > 0) ` in line 474 but not before line 482?

azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
ypadron-in: sure, I will change it.

azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
ypadron-in: ok

azkaban-web-server/src/web/js/azkaban/view/svg-graph.js
ypadron-in: variable 'nodes' in line 481 refers to all nodes (or flows) in the graph. Uploading an empty workflow to Azkaban will generate an error so I guess is guaranteed variable 'nodes' will exist but adding an extra check will not hurt. Let me know what you prefer.

ypadron-in:
> You can attach some screenshots to demonstrate what the UI looks like with the new change.
> Also mention the new change will apply to both flow 1.0 and flow 2.0.

PR description updated

azkaban-web-server/src/web/js/azkaban/view/svg-graph.js
ypadron-in: In expandAllFlows is not needed because this method actually triggers the creation of gNode, which is a JavaScript object containing info about the html tag that represents the node visually. This gNode is accessed in collapseAllFlows to hide the element so it needs to exist at this point. 

If we remove this check we get an error whenever we try to collapse before ever expanded because we iterate over all nodes when doing collapsing but some of them, the embedded ones, don’t have a gNode yet.

azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
jamiesjc: Please change all the places to keep consistency.

azkaban-web-server/src/web/js/azkaban/view/svg-graph.js
```diff
@@ -439,6 +441,50 @@ azkaban.SvgGraphView = Backbone.View.extend({
...
+    if (node) {
+      // collapse all embedded flows inside given node
+
+      // collapse already rendered nodes of type flow
+      if (node.type == 'flow' && node.gNode) {
+        this.collapseFlow(node);
+      }
+
+      if (node.nodes && node.nodes.length > 0) {
+        for (var i = 0; i < node.nodes.length; ++i) {
```
jamiesjc: Consider also moving this for-loop inside the `if (node.type == 'flow' && node.gNode)` check?

PR #2023, opened 2018-11-09
Improve dispatch request handling of a previously submitted execution

juhoautio:
Improve dispatch request handling of a previously submitted execution
- If the execution is indeed running, return OK so that dispatcher knows that it was successfully dispatched
- If the execution was left in some intermediate state, return an error so that dispatcher knows to retry or finalize the execution as failed

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -424,6 +424,10 @@ public class FlowRunnerManager implements EventListener,
     // Finally, queue the sucker.
     this.runningFlows.put(execId, runner);
 
+    submitFlowRunner(runner);
+  }
+
+  private void submitFlowRunner(FlowRunner runner) throws ExecutorManagerException {
     try {
       // The executorService already has a queue.
       // The submit method below actually returns an instance of FutureTask,
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -354,28 +354,35 @@ public class FlowRunnerManager implements EventListener,
   }
 
   public void submitFlow(final int execId) throws ExecutorManagerException {
-    // Load file and submit
-    if (this.runningFlows.containsKey(execId)) {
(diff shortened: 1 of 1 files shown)


jamiesjc:
I think your first point is to address the case where an execution has been submitted successfully and is indeed running, but web server thinks it fails to dispatch due to some network error. In this case, I doubt it really makes any difference whether it throws an exception or just return OK because anyway web server cannot receive the response because of the network issue. And it will finally give up dispatching when it reaches max dispatching error.
The second point addresses the case where an execution is added to running flows cache but failed 
 to be submitted to executorService. It doesn't make much difference to me that whether it throws an exception with 

> Execution is already running

 or 

> Execution is in runningFlows but not in submittedFlows

One thing to let you know beforehand, we are working on a new design of executor polling model instead of web server pushing model. I'll share with you the design doc once it's done.

juhoautio:
> the case where an execution has been submitted successfully and is indeed running, but web server thinks it fails to dispatch due to some network error. In this case, I doubt it really makes any difference whether it throws an exception or just return OK because anyway web server cannot receive the response because of the network issue.

Network error could happen randomly. In that case the first submit attempt would fail to read the response but already second time would succeed. I'm just trying to handle this corner case a bit better. (Although in reality I think it's very rare to end up in this state.) The current logic is flawed: dispatcher would finalize the newly submitted execution (mark flow status = FAILED) while executor is still running it.

> the case where an execution is added to running flows cache but failed
to be submitted to executorService. It doesn't make much difference to me that whether it throws an exception with..

As long as it's an "error", dispatcher shall treat it the same way, ie. basically give up & finalize. But maybe it can give a wrong idea if the message is just "Execution is already running". That sounds like nothing's wrong actually, even though there's this unexpected state (execution is in runningFlows but not in submittedFlows).

Thanks for sharing info on plans to switch to a pull model (push vs. pull, right?).

juhoautio:
@jamiesjc I'd like to highlight this:

> The current logic is flawed: dispatcher would finalize the newly submitted execution (mark flow status = FAILED) while executor is still running it.

This PR just fixes that, so I don't see why this wouldn't be an improvement worth doing?

seayoun:
@juhoautio

seayoun:
@juhoautio 
The web server will redispatch the flow again if the flow has been submitted from executorService and the second check has finished on the executor however it still response failed.

PR #2026, opened 2018-11-13
Npm tests for version updates

li-afaris:
This PR is to add additional testing for Node.js library updates which happen infrequently.  Testing will be done by cd'ing to the azkaban-web-server directory and running `npm test`.  This will avoid having to wait for azkaban-solo to start during `./gradlew test`.  

Additional tests should be added to validate all pages render correctly, but this should get things started.  I'm also open to recommendations on how to document this feature for future changes.


azkaban-web-server/build.gradle
```diff
@@ -117,7 +117,7 @@ task jsTest(dependsOn: ['npm_install'], type: NodeTask) {
     script = file('node_modules/mocha/bin/mocha')
 
     // args keep the test files' location
-    args = ['src/web/js/azkaban/test/']
+    args = ['src/web/js/azkaban/test/test.js']
```
li-ygerchikov: Does this change mean "run only the old test under this task"?

azkaban-web-server/package.json
```diff
@@ -3,18 +3,26 @@
...
   },
+  "devDependencies": {
+    "chai": "3.5.0",
+    "chai-http": "4.0.0",
+    "mocha": "5.0.4"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "pretest": "npm run startSoloServer",
+    "test": "mocha --recursive 'src/web/js/azkaban/test/' || npm run stop_n_exit",
```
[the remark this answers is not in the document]
li-afaris: test_running_server.js and package.json are used by `npm test` to start a azkaban-solo instance and confirms the web page renders.  

test.js is used by `gradle test` and only tests JavaScript functions return expected results.  A azkaban-solo instance is not started by `gradle test`.

azkaban-web-server/build.gradle
li-afaris: Correct.  This is to prevent `gradle test` from running everything in the test directory.   Maybe it's better to reorg the directory tree?  

```
src/web/js/azkaban/test/
                                       ├── test_running
                                       └── test_static
```

This would allow us to keep the current `args` behavior of running all tests in the test directory, but still have separation between gradle & NPM tests.

azkaban-web-server/package.json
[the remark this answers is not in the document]
li-ygerchikov: I guess I am still not sure how `npm test` knows to run `test_running_server.js` but not `test.js` when its parameter only specifies the (entire?) directory. Will it not run both? Or maybe it will, and it's ok?

azkaban-web-server/build.gradle
li-ygerchikov: I like it. Or maybe just `src/web/js/azkaban/test_running` and `src/web/js/azkaban/test_static`, no need for (empty?) `src/web/js/azkaban/test`?

azkaban-web-server/package.json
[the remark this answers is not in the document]
li-afaris: By default Mocha by default looks for "test/*js", [documentation here](https://mochajs.org/#the-test-directory).  We are overriding default behavior by passing a directory path as an argument to mocha. 

To answer your question, `NPM test` will run tests from both files.  I think this is okay because the few seconds spent testing components is much smaller then waiting for running tests to complete.

azkaban-web-server/build.gradle
li-afaris: Sounds good.  I'll submit a updated PR for the directory restructuring and remove the `test` dir.

azkaban-web-server/src/web/js/azkaban/test/test_running_server.js
```diff
@@ -0,0 +1,42 @@
...
+
+var chai   = require('chai');
+var chaiHttp = require('chai-http');
+var server = 'http://localhost:8081';
+var should = chai.should();
+
+chai.use(chaiHttp);
+
+// Test that Azkaban web interface is running
+describe("Will Azaban Login Work?", function() {
```
kunkun-tang: Have limited knowledge of chai testing. Looks like this script can only test against HTTP status code? what does `version change` mean here? Is it able to catch any js or css library break?

PR #2030, opened 2018-11-16
Remove unused method from HadoopSecurityManager

chengren311:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -367,69 +368,6 @@ public boolean isHadoopSecurityEnabled() {
     return this.securityEnabled;
   }
 
-  /*
-   * Gets hadoop tokens for a user to run mapred/pig jobs on a secured cluster
-   */
-  @Override
-  public synchronized void prefetchToken(final File tokenFile,
```
jamiesjc: Have you checked if this method is currently used by any external user?

PR #2035, opened 2018-11-20
Expose hit ratio of project dir cache as metrics

burgerkingeater:
This PR exposes hit ratio of a project dir cache(https://github.com/azkaban/azkaban/pull/1848) as metrics. Cache miss counts as a flow being setup without downloading its project zip, otherwise it counts as a cache hit.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
@@ -43,5 +43,7 @@ public class ExecMetrics {
         .addGauge("EXEC-NumRunningFlows", flowRunnerManager::getNumRunningFlows);
     this.metricsManager
         .addGauge("EXEC-NumQueuedFlows", flowRunnerManager::getNumQueuedFlows);
+    this.metricsManager
+        .addGauge("EXEC-ProjectDirCacheHitRatio", flowRunnerManager::getProjectDirHitRatio);
   }
 }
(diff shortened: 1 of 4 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
```diff
@@ -43,5 +43,7 @@ public void addFlowRunnerManagerMetrics(final FlowRunnerManager flowRunnerManage
         .addGauge("EXEC-NumRunningFlows", flowRunnerManager::getNumRunningFlows);
     this.metricsManager
         .addGauge("EXEC-NumQueuedFlows", flowRunnerManager::getNumQueuedFlows);
+    this.metricsManager
+        .addGauge("EXEC-ProjectDirCacheHitRatio", flowRunnerManager::getProjectDirHitRatio);
```
jamiesjc: Is it better to use the same name `getProjectDirCacheHitRatio`?

azkaban-exec-server/src/test/java/azkaban/execapp/FlowPreparerTest.java
```diff
@@ -183,4 +184,37 @@ public void testProjectCacheDirCleaner() throws IOException, InterruptedExceptio
...
+    //when
+    cacheMetrics.incrementCacheMiss();
+    //then
+    assertThat(cacheMetrics.getHitRatio()).isEqualTo(0);
+
+  }
+
+  @Test
+  public void testProjectsCacheMetricsHit() {
+    //given
```
jamiesjc: The comments in the test cases are not quite informative. Consider either adding more details or removing redundant ones?

burgerkingeater: updated thanks

Issue #2038, opened 2018-11-28
AZ New Dispatching Design


jamiesjc:
Advantages of the new dispatching design:

- Simplifies the current dispatching logic.
- Web server doesn’t need to constantly fetch executor info or update flow status in DB. The complicated logic in updater thread will be totally removed. The model is simpler and easier to understand and maintain. 
- Resolves current inconsistency issue of flow status caused by running flows cache.
- There is no cache involved. DB is the only source of truth.
- Paves the way for web server HA.
- Decouples web server from executors then web server doesn't need to maintain executor list in memory. 

New changes in the new dispatching design:
Web server will simply insert new executions to DB table when a user submits/schedules the flow.
All the existing cache will be removed from web server. Fetching running flows cache and queued flows cache will be replaced by direct DB queries. Queue processor thread will be removed. No need for active executors cache since web server doesn’t need to monitor or control executors now. Updater thread will be totally removed so that web server doesn’t need to ping executors for the flow status or check the executor health. Web server doesn’t need to refresh executors to get the executor info since it doesn’t need to push new executions to executors. 
Executors need to poll the executions from DB table. Only executors will be responsible for updating flow status. Also, executor will send notification emails instead of web server.

jamiesjc:
Detailed design doc: https://docs.google.com/document/d/1sSH70fPBB1mnwIBAmJrGgT_PCOl1UEdJMJe9m_ouw80/edit?usp=sharing

seayoun:
in the poll mode(ExecutionController), how to choose the executor, I looked the code , however I didnot saw some choose factor, what I saw is an exclusive lock on mysql, does it grab the clock？

jamiesjc:
Hi @seayoun, good question. In the poll model, executors poll from DB in a round-robin fashion. They don't calculate complicated factors before polling. The logic is simplified and proves to be working properly in our staging environment.  
We have a section discussing this part in the design doc. Feel free to check it out: https://docs.google.com/document/d/1sSH70fPBB1mnwIBAmJrGgT_PCOl1UEdJMJe9m_ouw80/edit#heading=h.l6p0a36sbeth

PR #2039, opened 2018-11-28
AZNewDispatchingLogic - Create ExecutionController module

jamiesjc:
This is the initial PR for AZ New Dispatching Logic #2038

A new module called `ExecutionController` is created to implement the new polling model on the web server side. When `azkaban.poll.model `is enabled, `ExecutionController` will be injected into `AzkabanWebServer`, otherwise `ExecutorManager` will be injected. Both of the two classes implement `ExecutorManagerAdapter`. In this way, we can develop the new model independently without affecting the current logic and switch between the two models easily. The new `ExecutionController` will ultimately replace `ExecutorManager` in the future.

Todo: 
implement the methods in `ExecutionController`. 
Enable the jmx/stats for `ExecutionController`.

Please see more details in the design description.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -1,0 +1,331 @@
+/*
+* Copyright 2018 LinkedIn Corp.
+*
+* Licensed under the Apache License, Version 2.0 (the “License”); you may not
+* use this file except in compliance with the License. You may obtain a copy of
+* the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an “AS IS” BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations under
+* the License.
+*/
+package azkaban.executor;
+
+import azkaban.event.EventHandler;
+import azkaban.project.Project;
+import azkaban.utils.FileIOUtils.LogData;
+import azkaban.utils.Pair;
+import java.io.IOException;
+import java.lang.Thread.State;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
+
+/**
+ * Controls flow executions on web server. This module implements the polling model
+ * in the new AZ dispatching design. It's injected only when azkaban.poll.model is configured to
+ * true. It will ultimately replace ExecutorManager in the future.
+ */
+@Singleton
+public class ExecutionController extends EventHandler implements ExecutorManagerAdapter {
+
+  private static final Logger logger = Logger.getLogger(ExecutionController.class);
+  private static final Duration RECENTLY_FINISHED_LIFETIME = Duration.ofMinutes(10);
+  private final ExecutorLoader executorLoader;
+  private final ExecutorApiGateway apiGateway;
+
+  @Inject
+  ExecutionController(final ExecutorLoader executorLoader, final ExecutorApiGateway apiGateway) {
+    this.executorLoader = executorLoader;
+    this.apiGateway = apiGateway;
+  }
+
+  @Override
+  public void setupExecutors() throws ExecutorManagerException {
+    // Todo: deprecate this method
+  }
+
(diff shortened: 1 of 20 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -228,7 +230,10 @@ public static void launch(final AzkabanWebServer webServer) throws Exception {
     /* This creates the Web Server instance */
     app = webServer;
 
-    webServer.executorManager.start();
+    // Todo jamiesjc: also start the threads in ExecutionController if needed.
+    if (webServer.executorManagerAdapter instanceof ExecutorManager) {
```
li-ygerchikov: Often `instanceof` check is considered anti-pattern. What if we had yet another, third implementation of `ExecutorManagerAdapter` that is not ExecutorManager but behaves more like it than like the new ExecutionController?
Could we delegate and trust the implementation to do the right thing? In this case, could we call `start` unconditionally here, but make `start` implementation a noop in the new class?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -544,7 +549,9 @@ private void prepareAndStartServer()
     createThreadPool();
     configureRoutes();
 
-    if (this.props.getBoolean(Constants.ConfigurationKeys.IS_METRICS_ENABLED, false)) {
+    // Todo jamiesjc: enable web metrics for azkaban poll model later
```
li-ygerchikov: Aside: some people like to add date (along with the author name) to their todos, so that readers can identify which era the todo comes from.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -589,7 +596,9 @@ private void addThreadPoolGauges(final QueuedThreadPool threadPool) {
 
 
   private void startWebMetrics() throws Exception {
-    this.metricsManager.addGauge("WEB-NumQueuedFlows", this.executorManager::getQueuedFlowSize);
+    this.metricsManager
+        .addGauge("WEB-NumQueuedFlows", ((ExecutorManager) this
```
li-ygerchikov: I hope this is temporary, predicated on the above `todo`. Obviously, this cast will fail when `executorManagerAdapter` is an `ExecutionController`. Is this just to postpone `addGauge` refactoring to accept ExecutorManagerAdapter instead of ExecutorManager?

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -0,0 +1,331 @@
+/*
```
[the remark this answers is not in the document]
jamiesjc: I've thought about creating a separate utility class to share the common methods between the new ExecutionController and old ExecutorManager. But since our plan is to deprecate ExecutorManager in the near future, I think this temporary duplication is ok for now.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
jamiesjc: Yes, as I mentioned in the "todo", after we add the implementation for `start` in the new class, we can remove this `instanceof` check. This would also require adding the `start` in the `ExecutorManagerAdapter` interface.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
jamiesjc: That's a good suggestion. Currently, I'm just following our existing style guide in CONTRIBUTING.md:  
 
>  todo username: todo details

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
jamiesjc: Yes, you are right. I have left the metrics/jmx part for later implementation since there are quite some differences between the new module and the old one, which would require a separate PR change.

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -0,0 +1,331 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
```
burgerkingeater: please use slf4j

jamiesjc: Updated, thanks!

Issue #1950, opened 2018-09-12
HTTP job callbacks can't parse params from global.properties


vladislav-sidorovich:
Hi guys.
There is a job file
```type=command
command=echo Hello World! env property is ${env}.
        
job.notification.started.1.url=https://...
job.notification.started.1.method=POST
job.notification.started.1.body={"text": "The flow: ?{flow} with execution id: ?{executionId} started"}
job.notification.started.1.headers=Content-Type:application/json
```

the `env` property is defined in global.properties file and successfully resolved in the job itself.  But during callback processing, I got an exception.

> 10-09-2018 20:25:14 UTC extraction ERROR - Encountered error while hanlding job callback event
azkaban.utils.UndefinedPropertyException: Could not find variable substitution for variable(s) [command->env]
    at azkaban.utils.PropsUtils.resolveVariableReplacement(PropsUtils.java:206)
    at azkaban.utils.PropsUtils.resolveProps(PropsUtils.java:167)
    at azkaban.execapp.event.JobCallbackManager.processJobCallOnStart(JobCallbackManager.java:213)
    at azkaban.execapp.event.JobCallbackManager.handleEvent(JobCallbackManager.java:113)
    at azkaban.event.EventHandler.fireEventListeners(EventHandler.java:37)
    at azkaban.execapp.JobRunner.fireEvent(JobRunner.java:801)
    at azkaban.execapp.JobRunner.fireEvent(JobRunner.java:794)
    at azkaban.execapp.JobRunner.doRun(JobRunner.java:583)
    at azkaban.execapp.JobRunner.run(JobRunner.java:552)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)

But job finished successfully and I see `Hello World! env property is dev.` in job's logs

I tried to change `command=echo Hello World! env property is ${env}.` to `command=echo Hello World! ` and job and callback work well in this case. That means the issue is in different property sources for job and callbacks.

[...]

PR #2041, opened 2018-11-28
Adding the ability to read global.properties in HTTP job callbacks 

vladislav-sidorovich:
It's the fix of the issue #1950

Technically I just save azkabanProps passed in JobCallbackManager and use these properties during HTTP job callbacks params resolving.


vladislav-sidorovich:
@HappyRay could you please take a look at this pool request.

kyrill007:
Would it be possible to prioritize merging of this pull request? I verified it locally and it works. This is pretty important for our project. Thank you so much!!!

PR #2042, opened 2018-11-28
Fix mismatch in name of job log files

ypadron-in:
Issue: when a user tries to kill a Hadoop/Spark job within an embedded flow the application looks for a log file using the following name pattern "_job.<exec_id>.<job_id>.log" no matter whether the job was embedded or not. But actually if a job is inside an embedded flow the name format used to create its log file is:  "_job.<exec_id>.<parent_flow>._.<job_id>.log"

So trying to kill a job inside an embedded flow before would fail because of the log file name mismatch.

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -36,10 +36,10 @@ import azkaban.spi.EventType;
 import azkaban.test.TestUtils;
 import azkaban.utils.Props;
 import java.io.BufferedReader;
-import java.io.InputStreamReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
+import java.io.InputStreamReader;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.nio.charset.Charset;
@@ -98,7 +98,7 @@ public class JobRunnerTest {
 
     eventCollector.handleEvent(Event.create(null, EventType.JOB_STARTED, new EventData(node)));
     Assert.assertTrue(runner.getStatus() != Status.SUCCEEDED
-        || runner.getStatus() != Status.FAILED);
+        && runner.getStatus() != Status.FAILED);
 
     runner.run();
     eventCollector.handleEvent(Event.create(null, EventType.JOB_FINISHED, new EventData(node)));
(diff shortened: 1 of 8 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -124,6 +129,30 @@ public void testBasicRun() throws IOException {
         .assertEvents(EventType.JOB_STARTED, EventType.JOB_STATUS_CHANGED, EventType.JOB_FINISHED);
   }
 
+  private void checkRequiredJobProperties(JobRunner runner, File logFile) {
+    Field jobField = null;
+    try {
+      jobField = runner.getClass().getDeclaredField("job");
```
jamiesjc: What is the benefit of using reflection here?

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
```diff
@@ -124,6 +129,30 @@ public void testBasicRun() throws IOException {
...
+    jobField.setAccessible(true);
+    InteractiveTestJob job = null;
+    try {
+      job = (InteractiveTestJob) jobField.get(runner);
+    } catch (IllegalAccessException e) {
+      Assert.fail("'job' field not accessible");
+      e.printStackTrace();
+    }
+    Props jobProps = job.getJobProps();
+    Assert.assertNotNull("Log file path not found in properties",
```
jamiesjc: You might not need this `assertNotNull` check since it will be covered in the `assertEquals` check later.

jamiesjc:
Please also mention in the change description that this issue happens to embedded flows only. Also you can give some examples to explain the issue so that others could understand it better.

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
ypadron-in: correct

azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
ypadron-in: It prevents us from having to create a get method in JobRunner class to access its private property 'job'. It prevents us from breaking encapsulation for testing purposes.

ypadron-in:
I updated the PR description

PR #2043, opened 2018-12-01
Fix bug: queued flows included in active flows

juhoautio:
Bug was introduced in #1833.

Fix as discussed in https://github.com/azkaban/azkaban/pull/1833#discussion_r237985252.

Improved the test coverage of this SQL query.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
@@ -18,6 +18,7 @@ package azkaban.executor;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.Assert.assertEquals;
 
 import azkaban.db.DatabaseOperator;
 import azkaban.project.JdbcProjectImpl;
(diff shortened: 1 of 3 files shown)


azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -307,34 +309,67 @@ public void testAssignExecutorInvalidExecution() throws Exception {
...
         this.fetchActiveFlowDao.fetchActiveFlows();
 
-    assertThat(activeFlows1.containsKey(flow1.getExecutionId())).isTrue();
-    assertThat(activeFlows1.get(flow1.getExecutionId()).getFirst().getExecutor().isPresent())
-        .isTrue();
-    assertThat(activeFlows1.containsKey(flow2.getExecutionId())).isTrue();
-    assertThat(activeFlows1.get(flow2.getExecutionId()).getFirst().getExecutor().isPresent())
-        .isFalse();
+    assertFound(activeFlows, flow1, true);
+    assertNotFound(activeFlows, flow2, "Returned a queued execution");
```
juhoautio: This assert reproduces the bug. It fails if run without the fix in FetchActiveFlowDao.

PR #2044, opened 2018-12-01
Feature to disable some http methods eg. TRACE

zulk666:
Possibility to disable TRACE.
Before when
curl -v -X TRACE localhost:8081 we have got 200
After we can add TRACE to jetty.disable.http-methods and curl will return 403

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
@@ -23,11 +23,15 @@ import azkaban.Constants;
 import azkaban.utils.Props;
 import javax.inject.Inject;
 import com.google.inject.Provider;
+import java.util.ArrayList;
 import java.util.List;
 import org.apache.log4j.Logger;
 import org.mortbay.jetty.Connector;
 import org.mortbay.jetty.Server;
 import org.mortbay.jetty.bio.SocketConnector;
+import org.mortbay.jetty.security.Constraint;
+import org.mortbay.jetty.security.ConstraintMapping;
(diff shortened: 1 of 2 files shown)


kamilduda:
@HappyRay Could you take a look at this?

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -61,12 +65,42 @@ public Server get() {
...
     // setting stats configuration for connectors
     setStatsOnConnectors(server);
+    disableHttpMethods(server);
 
     logger.info(String.format(
         "Starting %sserver on port: %d # Max threads: %d", useSsl ? "SSL " : "", port, maxThreads));
     return server;
   }
 
+    /**
```
burgerkingeater: please format the change appropriately with save plugin https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md

azkaban-web-server/src/main/java/azkaban/webapp/WebServerProvider.java
```diff
@@ -61,12 +65,42 @@ public Server get() {
...
+     * Disable HTTP methods defined in jetty.disable.http-methods property
+     *
+     * Multiple methods can be separated by coma eg. TRACE,OPTION
+     * @param server - jetty server instance
+     */
+  private void disableHttpMethods(Server server) {
+    String toDisable = props.getString("jetty.disable.http-methods","");
```
burgerkingeater: consider adding this property to azkaban.Constants class?

burgerkingeater:
what are the other http methods that prefer to be disabled?

PR #2046, opened 2018-12-03
Switch two checks from runningExecutions cache to DB.

jamiesjc:
There is a critical issue that `RunningExecutionsUpdaterThread` runs out of memory occasionally. The UpdaterThread stops updating `runningExecutions` cache, causing new flows not able to be submitted and flow/job logs not showing up. 
This fix is to switch two checks in submitting flows and fetching flow/job logs from cache to DB. This would mitigate the current problem while we are figuring out the root cause of the OOM issue.
This fix is also part of new AZ dispatching design which aims to remove all the cache from web server. All the places where `runningExecutions` cache is fetched will ultimately be replaced by direct DB calls. See #2038
To test the performance of DB queries, I launched ~ 1000 flows in solo-server at the same time and captured the time of calling `fetchUnfinishedFlows` and `fetchActiveFlowByExecId`. When the number of running flows is small, they are similar around 1-3 ms. When the number of running flows grows to ~ 1000, time of `fetchActiveFlowByExecId` remains the same, and `fetchUnfinishedFlows` is ~ 50 ms, sometimes reach ~ 100 ms. It will slightly affect the time to submit a flow, but still within an acceptable range.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
@@ -39,6 +39,12 @@ public interface ExecutorLoader {
   Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchActiveFlows()
       throws ExecutorManagerException;
 
+  Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedFlows()
+      throws ExecutorManagerException;
+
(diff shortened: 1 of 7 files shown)


jamiesjc:
Since the current OOM issue can be temporarily relieved by restarting web server, I'll move the code in this PR to the `ExecutionController` module in the new AZ dispatching design without touching the current `ExecutorManager` code.

PR #2051, opened 2018-12-05
Added 'fetchProjectLogs' call description to docs

antey86:


antey86:
DevTeam, 
please review this non code change

PR #2057, opened 2018-12-07
AZNewDispatchingLogic - Switch fetching running flows from cache to DB.

jamiesjc:
As part of the new design in #2038, we want to remove the cache from web server so that it can be stateless. This is also an important step towards the ultimate goal of web server HA.
In this PR, we replaced fetching from cache to direct DB calls in the new module `ExecutionController`. Also created new DB queries for performance improvement. 
Other than that, most of the functionalities remain the same as those in `ExecutorManager`.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -25,13 +25,14 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -89,8 +90,13 @@ public long getLastExecutorManagerThreadCheckTime() {
 
   @Override
   public Collection<Executor> getAllActiveExecutors() {
-    // Todo: get the executor info from DB
-    return Collections.emptyList();
+    List<Executor> executors = new ArrayList<>();
+    try {
```
kunkun-tang: I'm trying to understand why the exception should be cached in this block. Any particular reason?

jamiesjc: Added an additional check for empty executor list in the new commit.

PR #2058, opened 2018-12-08
adding plugin.dir to plugin job property set

kunkun-tang:
We recently received a new requirement that "plugin.dir" must be exposed to user code to navigate the classpath folder. This change make it happen.


HappyRay:
Where does this requirement come from i.e. why is it necessary?

kunkun-tang:
This jobtype needs to launch gobblin worker during job running phase, and that requires `jobtype.classpath`, that is a necessary setting for Gobblin.

PR #2067, opened 2018-12-14
AZNewDispatchingLogic - Submit flow in Polling Model

jamiesjc:
In the new AZ dispatching design #2038, one important change is to switch from pushing model to polling(pulling) model. Instead of web server selecting an executor and pushing new execution to that executor, now web server will just insert the new executions into DB queue and then executors will take turns to poll new executions from DB.
The PR implements the polling model for submitting new executions. Tested on solo server. Plan to test it on multi-executor mode later.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -92,6 +92,7 @@ public class Constants {
 
     // Configures Azkaban to use new polling model for dispatching
     public static final String AZKABAN_POLL_MODEL = "azkaban.poll.model";
+    public static final String AZKABAN_POLLING_INTERVAL_MS = "azkaban.polling.interval.ms";
 
     // Configures Azkaban Flow Version in project YAML file
     public static final String AZKABAN_FLOW_VERSION = "azkaban-flow-version";
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -15,16 +15,21 @@
 */
 package azkaban.executor;
 
+import azkaban.Constants.ConfigurationKeys;
 import azkaban.event.EventHandler;
+import azkaban.flow.FlowUtils;
 import azkaban.project.Project;
+import azkaban.project.ProjectWhitelist;
 import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.Pair;
+import azkaban.utils.Props;
 import java.io.IOException;
 import java.lang.Thread.State;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -46,13 +51,24 @@ public class ExecutionController extends EventHandler implements ExecutorManager
 
   private static final Logger logger = LoggerFactory.getLogger(ExecutionController.class);
   private static final Duration RECENTLY_FINISHED_LIFETIME = Duration.ofMinutes(10);
+  private static final int DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW = 30;
   private final ExecutorLoader executorLoader;
   private final ExecutorApiGateway apiGateway;
+  private final int maxConcurrentRunsOneFlow;
 
   @Inject
-  ExecutionController(final ExecutorLoader executorLoader, final ExecutorApiGateway apiGateway) {
+  ExecutionController(final Props azkProps, final ExecutorLoader executorLoader, final
+  ExecutorApiGateway apiGateway) {
     this.executorLoader = executorLoader;
     this.apiGateway = apiGateway;
+    this.maxConcurrentRunsOneFlow = getMaxConcurrentRunsOneFlow(azkProps);
+  }
+
+  private int getMaxConcurrentRunsOneFlow(final Props azkProps) {
+    // The default threshold is set to 30 for now, in case some users are affected. We may
+    // decrease this number in future, to better prevent DDos attacks.
+    return azkProps.getInt(ConfigurationKeys.MAX_CONCURRENT_RUNS_ONEFLOW,
+        DEFAULT_MAX_ONCURRENT_RUNS_ONEFLOW);
   }
 
   @Override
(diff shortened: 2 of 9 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -489,8 +505,71 @@ public void retryFailures(final ExecutableFlow exFlow, final String userId)
...
+      }
+
+      final boolean memoryCheck =
+          !ProjectWhitelist.isProjectWhitelisted(exflow.getProjectId(),
+              ProjectWhitelist.WhitelistType.MemoryCheck);
+      options.setMemoryCheck(memoryCheck);
+
+      // The exflow id is set by the loader. So it's unavailable until after
+      // this call.
+      this.executorLoader.uploadExecutableFlow(exflow);
```
jamiesjc: Most of the implementations remain the same as those in `ExecutorManager`. The only difference is that we are not enqueueing to `queueFlows` cache now and not adding new executions to `active_executing_flows` DB table which has been deprecated already.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -92,6 +92,7 @@
 
     // Configures Azkaban to use new polling model for dispatching
     public static final String AZKABAN_POLL_MODEL = "azkaban.poll.model";
+    public static final String AZKABAN_POLLING_INTERVAL_MS = "azkaban.polling.interval.ms";
```
burgerkingeater: is millisecond over fine-grained?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +922,54 @@ private void cleanRecentlyFinished() {
     }
   }
 
+  private class PollingThread extends Thread {
+
+    private final long pollingIntervalMs;
```
burgerkingeater: consider Duration?

burgerkingeater: consider Duration?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +922,54 @@ private void cleanRecentlyFinished() {
...
+    @Override
+    public void run() {
+      while (!this.shutdown) {
+        synchronized (this) {
+          try {
+            if (AzkabanExecutorServer.getApp() == null) {
+              continue;
+            }
+
+            final Executor executor = requireNonNull(FlowRunnerManager.this.executorLoader
```
burgerkingeater: technically we won't need executor table any more with change. Are you considering eliminating this dependency?

az-core/src/main/java/azkaban/Constants.java
jamiesjc: The polling frequency will be adjusted based on how many flows are scheduled/executed at the same time. For example, if 100 flows are scheduled to run at the same time and we only have one executor, we might want to configure small polling intervals like 100 ms to reduce execution delays. In the future, it's very likely that we need to support more flows scheduled at the same time, so I think millisecond is a proper time unit for the polling interval.

jamiesjc:
@burgerkingeater 

> what if same flow's two executions get polled into two different executors and the flow is disallowing any concurrent run?

If the flow disallows concurrent run, web server won't be able to insert two executions into the DB as the check is done in `submitExecutableFlow` on web server side.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +922,54 @@ private void cleanRecentlyFinished() {
     }
   }
 
+  private class PollingThread extends Thread {
```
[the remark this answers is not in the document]
jamiesjc: I can add a comment to explain the functionality of this new class. 
However I think the name `PollingThread` is relatively clear about what the thread mainly does. To me, `ScheduledThreadPoolExecutor` seems a little confusing and not easy to understand the actual functionality of this class.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +922,54 @@ private void cleanRecentlyFinished() {
...
+                        AzkabanExecutorServer.getApp().getPort()),
+                "The executor can not be null");
+
+            // Todo jamiesjc: check executor capacity before polling from DB
+            try {
+              final int execId = FlowRunnerManager.this.executorLoader
+                  .selectAndUpdateExecution(executor.getId());
+
+              if (execId != -1) {
+                logger.info("Submit flow " + execId);
```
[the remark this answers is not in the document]
jamiesjc: will fix, thanks

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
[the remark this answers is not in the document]
burgerkingeater: sorry for the confusion. My intention is https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html#schedule(java.lang.Runnable,%20long,%20java.util.concurrent.TimeUnit)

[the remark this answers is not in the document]
jamiesjc: Thanks for the suggestion. I'll consider using the `ScheduledThreadPoolExecutor` to do the polling instead.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +924,56 @@ private void cleanRecentlyFinished() {
     }
   }
 
+  /**
+   * Polls new executions from DB periodically and submits the executions to run on the executor.
+   */
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private class PollingThread {
```
burgerkingeater: it's not a thread anymore,  might give it another name?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -906,4 +924,56 @@ private void cleanRecentlyFinished() {
...
+      this.pollingIntervalMs = pollingIntervalMs;
+      this.scheduler = Executors.newSingleThreadScheduledExecutor();
+    }
+
+    public void start() {
+      this.scheduler.scheduleAtFixedRate(() -> pollingExecution(), 0L, this.pollingIntervalMs,
+          TimeUnit.MILLISECONDS);
+    }
+
+    private void pollingExecution() {
```
burgerkingeater: poll ?

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
burgerkingeater: would this introduce code duplication?

PR #2075, opened 2018-12-22
Logs kill command failures and removes unused code

ypadron-in:
There have been times when killing an execution has taken over an hour to complete without an apparent reason to do so. This PR helps to debug this type of issues by adding logs to detect kill command failures.

Also we are removing a piece of code that was not in use and was potentially adding a bit of delay to the killing process. The functionality of this piece of code was verified on a test server as well as in solo server. It was repeating logs that are already in the job logs.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
@@ -16,7 +16,6 @@
 
 package azkaban.jobExecutor.utils.process;
 
-import azkaban.utils.LogGobbler;
 import com.google.common.base.Joiner;
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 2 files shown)


jamiesjc:
Overall LGTM. It'll be better to add some details in the change description, i.e., mentioning the original issue, any testing done, or action items after checking in this change.

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
```diff
@@ -204,24 +175,38 @@ public void hardKill() {
...
   }
 
+  private void executeAndWaitForKillCmd(final String cmd) {
+    try {
+      final Process p = Runtime.getRuntime().exec(cmd);
+      // kill should return immediately since it only sends a signal
+      final int exitCode = p.waitFor();
+      if (exitCode != 0) {
+        this.logger.error("Kill terminated with exit code " + exitCode);
+        final BufferedReader b = new BufferedReader(
```
burgerkingeater: should we close the reader at final?

PR #2079, opened 2019-01-02
Fix cron test failure due to new year.

jamiesjc:
The cron test failed because of the hard-coded calendar year 2018. Changed it to get the current calendar year using `Calendar.getInstance().get(Calendar.YEAR)` so that it won't fail when the new year comes.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/utils/UtilsTest.java
+++ b/azkaban-common/src/test/java/azkaban/utils/UtilsTest.java
@@ -16,6 +16,7 @@
 
 package azkaban.utils;
 
+import java.util.Calendar;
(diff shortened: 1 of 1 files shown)


HappyRay:
Thanks for fixing this test failure.

Is there a way to remove the hardcoded year assumption in this test?

jamiesjc:
@HappyRay I've removed the hardcoded year and replaced it with the current year. Can you explain a little more about your question?

HappyRay:
Never mind.
I thought you replaced 2018 with 2019 based on your description.

jamiesjc:
Thanks, updated the description with more details.

PR #2088, opened 2019-01-07
AZNewDispatchingLogic - Enable JMX for ExecutionController

jamiesjc:
This PR enables JMX for `ExecutionController`. The new /jmx page will display running/queued flows info but remove some other info like updater thread and queued processor thread which are deprecated in the new dispatching design. (See design #2038)
Also remove unused class `JmxExecutorManagerAdapter` and `JmxExecutorManagerAdapterMBean`.

screen shot 2019-01-07 at 2 37 04 pm


azkaban-common/src/main/java/azkaban/jmx/JmxExecutionController.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2014 LinkedIn Corp.
```
burgerkingeater: year

jamiesjc: Fixed, thanks!

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -249,28 +249,74 @@ public ExecutableFlow getExecutableFlow(final int execId)
...
-    for (final Pair<ExecutionReference, ExecutableFlow> ref : collection) {
-      flows.add(ref.getSecond());
+    collection.stream().forEach(ref -> flows.add(ref.getSecond()));
+  }
+
+  /**
+   * Get execution ids of all running (unfinished) flows from database. {@inheritDoc}
+   */
+  @Override
+  public String getRunningFlowIds() {
```
burgerkingeater: why returning a string instead of a list?

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -249,28 +249,74 @@ public ExecutableFlow getExecutableFlow(final int execId)
...
+      this.logger.error("Failed to get running flow ids.", e);
+    }
+    return allIds.toString();
+  }
+
+  /**
+   * Get execution ids of all non-dispatched flows from database. {@inheritDoc}
+   */
+  @Override
+  public String getQueuedFlowIds() {
```
burgerkingeater: same question as above.

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
shuzhang1989: +1, we can do list->string serialization on the caller side. the function should provide behavior consistent with its name.

jamiesjc: Changed from string to list, thanks.

PR #2094, opened 2019-01-15
Rollback changes on PR #2075:

ypadron-in:
Reverting changes made on PR (https://github.com/azkaban/azkaban/pull/2075)  because they are causing jobs to hang.  
Commits to revert:
-Remove unused code (https://github.com/azkaban/azkaban/pull/2075/commits/661351f1f82eaf1d2a63dffc50946ac4113fd49c). This is the commit that's causing the issue. It eliminated code that read from the standard output of a subprocess but did nothing with it. That caused the subprocess’ standard output to not be consumed, so when the pipe was full the process blocked trying to write to standard output.
-Log kill command failures when killing jobs (https://github.com/azkaban/azkaban/pull/2075/commits/734f01b049e75701e97c0b3fe97f16676f00c831)

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
@@ -16,8 +16,8 @@
 
 package azkaban.jobExecutor.utils.process;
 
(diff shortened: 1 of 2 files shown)


burgerkingeater:
thanks @ypadron-in can you elaborate in the description which commit causes the issue and why?

PR #2105, opened 2019-01-22
AZNewDispatchingLogic - Enable alertOnFirstError

jamiesjc:
In AZ new dispatching design, executors will be responsible for alerting on Flow status change instead of the web server. 
When a user chooses `First failure` in the `Notification` option when executing the flow, it is expected that a notification email will be sent immediately after any job fails and the flow status changes to FAILED_FINISHING. This PR consolidates the code and enables this `alertOnFirstError` option in the new AZ dispatching design.

See more details in design #2038

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -74,7 +74,7 @@ public class ExecutionControllerUtils {
     }
 
     if (alertUser) {
-      alertUser(flow, alerterHolder, getFinalizeFlowReasons(reason, originalError));
+      alertUserOnFlowFinished(flow, alerterHolder, getFinalizeFlowReasons(reason, originalError));
     }
   }
 
@@ -85,8 +85,8 @@ public class ExecutionControllerUtils {
    * @param alerterHolder the alerter holder
    * @param extraReasons the extra reasons for alerting
    */
-  public static void alertUser(final ExecutableFlow flow, final AlerterHolder alerterHolder,
-      final String[] extraReasons) {
+  public static void alertUserOnFlowFinished(final ExecutableFlow flow, final AlerterHolder
+      alerterHolder, final String[] extraReasons) {
     final ExecutionOptions options = flow.getExecutionOptions();
     final Alerter mailAlerter = alerterHolder.get("email");
     if (flow.getStatus() != Status.SUCCEEDED) {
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -136,6 +136,40 @@ public static void alertUser(final ExecutableFlow flow, final AlerterHolder aler
     }
   }
 
+  /**
+   * Alert the user when the flow has encountered the first error.
+   *
+   * @param flow the execution
+   * @param alerterHolder the alerter holder
+   */
+  public static void alertUserOnFirstError(final ExecutableFlow flow,
```
jamiesjc: This method is moved from RunningExecutionsUpdater.java with slight modification.

burgerkingeater:
can you reference the PR that making executor send alerting email?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -623,12 +629,28 @@ private boolean retryJobIfPossible(final ExecutableNode node) {
...
+   * dispatching design.
+   *
+   * @param base the base flow
+   * @param status the status to be propagated
+   */
   private void propagateStatus(final ExecutableFlowBase base, final Status status) {
     if (!Status.isStatusFinished(base.getStatus()) && base.getStatus() != Status.KILLING) {
       this.logger.info("Setting " + base.getNestedId() + " to " + status);
-      base.setStatus(status);
+      boolean shouldAlert = false;
```
burgerkingeater: Have you considered separating the alerting part out into another method? method with one single responsibility brings more simplicity and composability.

jamiesjc: Thanks for the suggestion. I've also considered separating the alerting part. However, alerting happens when the flow status is propagated to `FAILED_FINISHING` for the first time. So it anyway needs to recursively call `getParentFlow` to access the root flow status, which is the same as what `propagateStatus` is doing. Also to make sure we only alert once, a global variable might be needed to keep the info whether alerting has happened or not. Considering the efficiency and simplicity of the code, I would prefer combining them together, maybe changing the method name to better describe the purpose of this method.

jamiesjc: Discussed offline and agreed on adding a new FLOW_STATUS_CHANGED event type and switching to the event-driven model in the future.

PR #2112, opened 2019-01-25
Fix local conf template to work out of the box

juhoautio:
`azkaban.use.multiple.executors=true` is now the only supported mode, so add it to the default template that is copied & used when one runs AzkabanSingleServer for the first time (in IDE, typically).

This property is required to be explicitly set to true in properties, so that users don't accidentally miss that the mode is switched, when they upgrade Azkaban. That's why it's better to have it like this for now instead of defaulting to true.


juhoautio:
@HappyRay ping

PR #2113, opened 2019-01-25
Optimize gradle build time

juhoautio:
https://github.com/azkaban/azkaban/pull/1231 configured gradle to parallelize test runs. However, in practice enabling parallel tests actually makes the full azkaban test run slower on some machines (at least mine & @burgerkingeater's machine). There hasn't been any noticeable speed-up on Travis. Hence, changing back to no parallel tests.

Instead of removing entirely, keeping the setting with value 1 & adding a comment to warn about it, so that it will be seen instead of enabling it again in the future without proper validation.

----

`./gradlew cleanTest test` took 3m 42s on my machine. That seemed way too much. I was running tests of each module in IDEA and for sure the total time wouldn't be that long. It lead to think that the command-line gradle runner must be doing something that's not optimal.

I knew that idea runs all tests sequentially, so I tried disabling parallelism entirely also for the command-line test runner. It was ~4 times faster:

maxParallelForks = 1
-> `./gradlew cleanTest test` 49s

I have 4 cores on my machine so I also gave this a try, but not much help:
maxParallelForks = 4
-> `./gradlew cleanTest test` 2m 50s

To make sure it's not just random fluctuation I tested the "fix" once more:
maxParallelForks = 1
-> `./gradlew cleanTest test` 1m 4s

And one more go with the original setting:
maxParallelForks = 12
-> `./gradlew cleanTest test` 2m 0s

It could be that the forked execution was even slower when my machine had other heavy workloads going on (hence the big variation from 2m to almost 4m).

Maybe the tests interfere with each other.. Maybe the additional forking is just expensive because every test needs to load some heavy classes / static members again.. Whatever the reason is, there's no point in having parallelism on if makes the total build time longer.

It would be best if someone else could also give this a try on a differen machine & measure the build times to compare that it's not just my machine where this happens. I'm excited to see how this affects the build time on Travis..

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -179,7 +179,9 @@ subprojects {
         test {
             // Set maxParallelForks to a large number and let gradle to force it to a the
             // max-workers number when needed.
-            maxParallelForks = 12
+            // NOTE! For some reason anything more than 1 doesn't work well for azkaban build:
+            // any maxParallelForks > 1 seems to make `./gradlew cleanTest test` ~4 times slower
+            maxParallelForks = 1
         }
     }
 


juhoautio:
The quickest build in master build history's first page is `4 min 45 sec`.

This PR built on Travis `4 min 57 sec`.

On the other hand azkaban builds on Tarvis may some times take even 7 minutes.

Need more data to say anything sure about does this have an impact on build times on Travis on not.

If it's at least not making things worse, I'd like to have this change to optimize builds on my personal machine – which hopefully helps others too.

juhoautio:
@HappyRay @jamiesjc anyone to test on your machine if you can see similar speed-up with this change?

burgerkingeater:
according to https://github.com/azkaban/azkaban/pull/1231, enabling maxParallelForks should reduce the test running time, @juhoautio  have you compared setup without maxParallelForks specified?

juhoautio:
@burgerkingeater yeah that's the theory, but the reality in practice seems different.

These are just guesses of course:

> Maybe the tests interfere with each other.. Maybe the additional forking is just expensive because every test needs to load some heavy classes / static members again..

But:

> Whatever the reason is, there's no point in having parallelism on if makes the total build time longer.

If I entirely remove `maxParallelForks` the execution time is on the same level as with `maxParallelForks = 1`.

Could you also test it on your machine locally?

juhoautio:
@burgerkingeater bump ^

juhoautio:
@burgerkingeater have you been able to check what your local build time is and does this setting affect it?

burgerkingeater:
@juhoautio  i did the experiment, it's strange with `maxParallelForks = 1`, `./gradlew cleanTest test` failed for some tests and it failed in 2m 37s,  and `maxParallelForks = 12` succeeded in  2m 10s.

juhoautio:
@burgerkingeater tests shouldn't fail because of the maxParallelForks setting, at least when the value is 1. So maybe it was just some random error. Did you try only once? Maybe test once more with different settings?

If the conclusion is that at least maxParallelForks isn't making tests faster on any setup, would you be open to setting it to 1 in the repo?

juhoautio:
@burgerkingeater are we able to conclude that this change can only make test run faster? In some environments it doesn’t, but it also doesn’t slow it down. Right? Please, this would be helpful for me at least, probably also for at least some other developers also. Now about a month old pr, so would be nice to reach a conclusion.

burgerkingeater:
@juhoautio sorry for the delayed response. It's strange that the test run time is increased with https://github.com/azkaban/azkaban/pull/1231. I also observed the same result https://github.com/azkaban/azkaban/pull/1231#issuecomment-469506796. Can you summarize our experiment and refer this PR to https://github.com/azkaban/azkaban/pull/1231, then I will merge it, thanks.

juhoautio:
@burgerkingeater I've updated the PR description, is it what you were looking for?

burgerkingeater:
thanks!

ypadron-in:
> @HappyRay @jamiesjc anyone to test on your machine if you can see similar speed-up with this change?

Ran several times `./gradlew cleanTest test` with `maxParallelForks = 1` in a MacOS' 10.14 terminal. This change reduced running time from 1m 15s (with maxParallelForks=12) to 28s overall.

Should we consider set `maxParallelForks = 1` to the TRAVIS configuration inside the build.gradle file as well? Currently is set `maxParallelForks = 2`

Issue #2114, opened 2019-01-25
Optimize the slow unit tests


juhoautio:
The slowest test duration was in these modules (as far as I could quickly find):

- azkaban-common: 10 s
- azkaban-exec-server: 12 s
   - `FlowPreparerTest#testProjectCacheDirCleaner`: 3 s (does some Thread.sleep)
- azkaban-web-server: 20 s
   - `TriggerInstanceProcessorTest#testProcessTermination`: 10 s (must be something wrong)
   - `FlowTriggerInstanceLoaderTest#testDeleteOldTriggerInstances`: 3 s (too much sleeping)

Could it be a goal in Azkaban project to keep the tests quick & especially avoid Thread.sleep?

HappyRay:
> Could it be a goal in Azkaban project to keep the tests quick & especially avoid Thread.sleep?
Yes. Thanks for the investigation.

PR #2111, opened 2019-01-25
Unify jvm memory settings validation & improve upload error message

juhoautio:
Users shouldn't need to go view Azkaban source to find out why project upload failed in this kind of a case (from azkaban mailing list):

> I am trying to upload a zip project file and getting this error, could not figure out the reason.

> Users were getting java heap space error very often and by mistake while trying to solve it I have introduced the "job.max.Xmx=" twice in azkaban.properties, one of them empty.

```
Installation Failed.
For input string: "" 

2019/01/24 13:25:02.852 -0300 INFO [ProjectManagerServlet] [Azkaban] Upload: reference of project test_project_for_delete is 101056896
2019/01/24 13:25:02.853 -0300 INFO [ProjectManagerServlet] [Azkaban] Uploading file lte.zip
2019/01/24 13:25:02.853 -0300 INFO [AzkabanProjectLoader] [Azkaban] Uploading files to test_project_for_delete
2019/01/24 13:25:02.864 -0300 WARN [XmlValidatorManager] [Azkaban] Validator directory validators does not exist or is not a directory.
2019/01/24 13:25:02.865 -0300 WARN [XmlValidatorManager] [Azkaban] Azkaban properties file does not contain the key project.validators.xml.file
2019/01/24 13:25:02.865 -0300 INFO [AzkabanProjectLoader] [Azkaban] Validating project lte.zip using the registered validators []
2019/01/24 13:25:02.875 -0300 INFO [ProjectManagerServlet] [Azkaban] Installation Failed.
java.lang.NumberFormatException: For input string: ""
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.base/java.lang.Long.parseLong(Long.java:702)
        at java.base/java.lang.Long.parseLong(Long.java:817)
        at azkaban.utils.Utils.parseMemString(Utils.java:453)
        at azkaban.project.FlowLoaderUtils.checkJobProperties(FlowLoaderUtils.java:260)
        at azkaban.project.DirectoryFlowLoader.loadProjectFlow(DirectoryFlowLoader.java:142)
        at azkaban.project.AzkabanProjectLoader.uploadProject(AzkabanProjectLoader.java:105)
        at azkaban.project.ProjectManager.uploadProject(ProjectManager.java:506)
        at azkaban.webapp.servlet.ProjectManagerServlet.ajaxHandleUpload(ProjectManagerServlet.java:1753)
[...]

PR #2116, opened 2019-01-25
Optimize slow test TriggerInstanceProcessorTest#testProcessTermination

juhoautio:
Squashing the lowest hanging fruit from https://github.com/azkaban/azkaban/issues/2114.

1st commit: Prove that sendEmailLatch is not incremented in testProcessTermination
- Adding this assertion makes the test fail

2nd commit: Fix mocking of overloaded method -> test passes quickly now

It's actually best to leave the new assertion on CountdownLatch in place even though the problem was fixed. Because it will make the test fail instead of just being slow & succeeding, if the bug with skipping CountdownLatch is ever introduced again.

### DIFF
--- a/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/flowtrigger/TriggerInstanceProcessorTest.java
@@ -18,6 +18,7 @@ package azkaban.flowtrigger;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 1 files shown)


HappyRay:
@abti is the new manager. Please work with him directly. Thanks

juhoautio:
@abti hi, could I get someone to review this one, please?

Same question also for:
https://github.com/azkaban/azkaban/pull/2111
https://github.com/azkaban/azkaban/pull/2112
https://github.com/azkaban/azkaban/pull/2113

All are quite small.

burgerkingeater:
@juhoautio why would this fix optimize the test slowness by adding the assert??

juhoautio:
@burgerkingeater assert doesn't optimize the test, but:

> It's actually best to leave the new assertion on CountdownLatch in place even though the problem was fixed. Because it will make the test fail instead of just being slow & succeeding, if the bug with skipping CountdownLatch is ever introduced again.

juhoautio:
@burgerkingeater anything else?

burgerkingeater:
@juhoautio `2nd commit: Fix mocking of overloaded method -> test passes quickly now` do you know how much time save this fix brings and why?

juhoautio:
@burgerkingeater usually it saves 9-10 seconds. Without this fix the test was always blocking the full 10 seconds on the countdown latch. Now it's only blocking until the latch triggers, which usually takes less than a second.

PR #2120, opened 2019-02-04
Fix pagination bugs on Flow page

ypadron-in:
Fix pagination bugs on Flow page:
 -Next and Previous buttons don’t work when clicked
 -Invalid pagination is generated in some cases. For example:
   * if the number of executions yields 3 pages in total, when the last page(3) is clicked the pagination displayed will be -1 0 1 2 3.

It also separates the implementation of Flow Trigger tab from that of Executions tab by creating a Backbone Model and View for the former.


burgerkingeater:
@ypadron-in  please check  build failure

PR #2122, opened 2019-02-06
Refactor and bug fix on Job History page pagination

ypadron-in:
Fixes a pagination issue on Job History page where it would show an additional empty page if the last page was full. For example:
 -with page size 10 and 30 elements it would show 4 pages with the last one being empty.

Simplifies implementation by delegating common pagination functionality to [https://github.com/josecebe/twbs-pagination](https://github.com/josecebe/twbs-pagination) jQuery pagination plugin instead of doing it manually.

With this change we are still loading the entire page every time a user interacts with the pagination controls. Ultimately we want to create an API endpoint that returns pages as data and that we only need to update the view with the new data, but this will not be done now because we are planning to redesign existing APIs soon.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -1151,6 +1151,22 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
     final Page page =
         newPage(req, resp, session,
             "azkaban/webapp/servlet/velocity/jobhistorypage.vm");
+
+    final String jobId = getParam(req, "job");
+    page.add("jobId", jobId);
+
+    int pageNum = Math.max(1, getIntParam(req, "page", 1));
+    page.add("page", pageNum);
+
+    final int pageSize = Math.max(1, getIntParam(req, "size", 25));
+    page.add("pageSize", pageSize);
+
+    page.add("recordCount", 0);
+    page.add("projectId", "");
+    page.add("projectName", "");
+    page.add("dataSeries", "[]");
+    page.add("history", null);
+
     final String projectName = getParam(req, "project");
     final User user = session.getUser();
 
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1166,88 +1182,36 @@ private void handleJobHistoryPage(final HttpServletRequest req,
...
+        page.add("page", pageNum);
       }
-      page.add("history", jobInfo);
+      final int elementsToSkip = (pageNum - 1) * pageSize;
+      final List<ExecutableJobInfo> jobInfo =
+          this.executorManagerAdapter.getExecutableJobs(project, jobId, elementsToSkip, pageSize);
 
-      page.add("previous", new PageSelection("Previous", pageSize, true, false,
-          Math.max(pageNum - 1, 1)));
+      if (jobInfo != null && !jobInfo.isEmpty()) {
```
burgerkingeater: consider this?: https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isNotEmpty-java.util.Collection-

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobhistorypage.vm
```diff
@@ -35,8 +36,22 @@
 
     var projectId = "$projectId";
     var projectName = "$projectName";
-    var jobName = "$jobid";
+    var jobName = "$jobId";
     var dataSeries = $dataSeries;
+
+    var jobHistoryPageSettings = {
```
burgerkingeater: nice refactoring

azkaban-web-server/src/web/js/azkaban/view/job-history.js
```diff
@@ -21,18 +21,44 @@ var jobHistoryView;
...
-  var selected;
-  var series = dataSeries;
-  dataModel = new azkaban.DataModel();
-  dataModel.set({
-    "data": series
+var initJobHistoryPage = function (settings) {
+  dataModel = new azkaban.DataModel({
+    page: settings.page,
+    pageSize: settings.pageSize,
+    visiblePages: 5,
```
burgerkingeater: why is this not part of settings?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobhistorypage.vm
```diff
@@ -35,8 +36,22 @@
 
     var projectId = "$projectId";
     var projectName = "$projectName";
-    var jobName = "$jobid";
+    var jobName = "$jobId";
```
burgerkingeater: thanks for doing that

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ypadron-in: I changed the if condition to use the utility method isNotEmpty(). Thanks

azkaban-web-server/src/web/js/azkaban/view/job-history.js
ypadron-in: Added "visiblePages" to settings object to keep consistency

Issue #2020, opened 2018-11-09
one azkaban executor process


burgerkingeater:
Currently when deploying new executor, we deactivate the old one and start a new one, allowing multiple executor process running. This creates a lot of complexity and operation overhead. 
We should only allow one executor process up and running, who is able to pick up the state of job/flow status of old executor.

juhoautio:
This can't mean stopping to support multiple executors per se.. Do you run multiple executor processes on the same instance during deployment or what is this about? Even if so, isn't that just deployment specifics that azkaban itself doesn't need to be know about?

chengren311:
@juhoautio sorry for any confusion. This is to allow Azkaban executors to be upgraded without disruption to running flows and without keeping multiple instances running.
 we run multiple executor processes on the same instance. When starting a new azkaban executor process, we deactivate old one to prevent web server dispatching new flow execution to this executor process. Old executor processes will be solely responsible for finishing ongoing flows. Allowing multiple executor processes running adds complexity. E.g: when duplicate metrics could be produced, or race condition could happen when two executor processes are cleaning up the shared project dir: https://github.com/azkaban/azkaban/pull/2017

juhoautio:
Ok, so we're in sync. We haven't tried running multiple executor processes on the same instance. I think the reasons are quite obvious. We create a new executor instance to deploy code changes, and keep the old executor running in inactive state until executions on it have finished.

I don't see the need to _add complexity to Azkaban as a product_ to prevent users from running multiple executor processes on the same instance. To me it seems quite obvious that it's not so easy to pull off without unwanted side effects and rather should not be attempted at all. But maybe it could be documented.

Just my two cents. Maybe I'm missing something.

PR #2123, opened 2019-02-09
Project shared cache race condition fix

burgerkingeater:
This PR 
1. handles race condition where two threads are downloading the same project, and renaming temp directory at the same time, the latter renaming operation would throw exception since destination directory is already created by the first renaming operation. If this is the case, renaming should ignore the exception and move on with next steps.

2. adds the existence check of project dir for updateDirSize/updateFileCount to avoid project dir is deleted in the middle by another thread that is being setup.

Long term fix: https://github.com/azkaban/azkaban/issues/2020

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -30,6 +30,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -271,8 +275,13 @@ void setupProject(final ProjectVersion pv)
       if (this.projectDirCleaner != null) {
         this.projectDirCleaner.deleteProjectDirsIfNecessary(pv.getDirSizeInBytes());
       }
-      Files.move(tempDir.toPath(), pv.getInstalledDir().toPath(), StandardCopyOption.ATOMIC_MOVE);
+      try {
+        Files.move(tempDir.toPath(), pv.getInstalledDir().toPath(), StandardCopyOption.ATOMIC_MOVE);
```
edwinalu: Is it possible that the other thread, which downloaded the same project directory, could delete the directory before this thread has a chance to hard link?

edwinalu:
Could a unit test be added? It's a race condition so harder, but an option would be to create 2 threads which move to the same dir a few times.

burgerkingeater:
closing this PR as it's duplicate of https://github.com/azkaban/azkaban/pull/2130

PR #2129, opened 2019-02-20
 Implement "useExecutor" feature for new dispatching logic

ypadron-in:
This is one the features pending after the implementation of the new dispatching logic started on #2038.

When launching a new execution, an Azkaban admin can choose the executor for it by specifying a useExecutor parameter in the request.

New executions are inserted into execution_flows table in the database. A newly added column to this table, "use_executor", will hold the executor id passed as a parameter in the request.

Active executors will poll executions with `(use_executor == null or use_executor == pollingExecutorId)`
Inactive executors will only poll executions with their ids  `(use_executor == pollingExecutorId)`


azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -307,10 +315,16 @@ public int selectAndUpdateExecution(final int executorId) throws ExecutorManager
   public static class SelectFromExecutionFlows implements
       ResultSetHandler<List<Integer>> {
 
-    static String SELECT_EXECUTION_FOR_UPDATE = "SELECT exec_id from execution_flows where "
-        + "status = " + Status.PREPARING.getNumVal()
-        + " and executor_id is NULL and flow_data is NOT NULL ORDER BY submit_time ASC LIMIT 1 FOR "
-        + "UPDATE";
+    static String selectExecutionForUpdateQueryString(final int executorId,
+        final boolean isActive) {
+      final String useExecutorCondition = isActive ?
```
burgerkingeater: consider string.format?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -47,9 +48,14 @@ public ExecutionFlowDao(final DatabaseOperator dbOperator) {
 
   public synchronized void uploadExecutableFlow(final ExecutableFlow flow)
```
edwinalu: Not related to this change, but does this function need to be synchronized?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -47,9 +48,14 @@ public ExecutionFlowDao(final DatabaseOperator dbOperator) {
...
       throws ExecutorManagerException {
+
+    final String useExecutorParam =
+        flow.getExecutionOptions().getFlowParameters().get(ExecutionOptions.USE_EXECUTOR);
+    final String executorId = StringUtils.isNotEmpty(useExecutorParam) ? useExecutorParam : "-1";
```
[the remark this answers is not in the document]
ypadron-in: I'll change default value to NULL, thanks! @li-ygerchikov also pointed out a few performance and space benefits of defaulting to NULL.

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE execution_flows (
   enc_type    TINYINT,
   flow_data   LONGBLOB,
   executor_id INT                   DEFAULT NULL,
+  use_executor INT                  DEFAULT -1,
```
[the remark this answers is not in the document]
ypadron-in: will change it

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: using it now.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: I don't see why it would be necessary. It's the only synchronized method in the class so it doesn't protect any private attributes, it's only preventing concurrent calls to this method but there is nothing in there that would be a problem in those cases. I'll remove it in a separate PR, thanks.

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE execution_flows (
   enc_type    TINYINT,
   flow_data   LONGBLOB,
   executor_id INT                   DEFAULT NULL,
+  use_executor INT,
```
burgerkingeater: do we need to set the default value here?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -307,10 +315,16 @@ public int selectAndUpdateExecution(final int executorId) throws ExecutorManager
   public static class SelectFromExecutionFlows implements
       ResultSetHandler<List<Integer>> {
 
-    static String SELECT_EXECUTION_FOR_UPDATE = "SELECT exec_id from execution_flows where "
-        + "status = " + Status.PREPARING.getNumVal()
-        + " and executor_id is NULL and flow_data is NOT NULL ORDER BY submit_time ASC LIMIT 1 FOR "
-        + "UPDATE";
+    static String selectExecutionForUpdateQueryString(final int executorId,
```
edwinalu: This is creating a new String each time the function is called. Can this return a constant String, either one which would handle use_executor, or one of 2 depending on use_executor?

ypadron-in: I don’t see how we could avoid creating a string every time because the parameters to the query are not known at compile time. What I can do is to replace the string concatenations with String.format. 
BTW, this query string will likely get even more parameters when we implement the flow priority feature, which is why I still prefer to use an auxiliary method instead of constant strings.

azkaban-db/src/main/sql/create.execution_flows.sql
ypadron-in: no need to set it explicitly, default is NULL

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
edwinalu: You can use "?" for the parameters. Some information at https://docs.oracle.com/javase/tutorial/jdbc/basics/prepared.html. Also examples in uploadFlowFile() and updateChunksInProjectVersions().

Allocating new String each time means that they will need to get garbage collected. Also, It will need to be re-parsed for the query. The earlier link also mentions reusing a PreparedStatement, although the current code doesn't use this.

Some databases do query caching, in which case having the same String can help find a cache hit, but it looks like this was removed in MySQL 8.0: https://mysqlserverteam.com/mysql-8-0-retiring-support-for-the-query-cache/

azkaban-db/src/main/sql/create.execution_flows.sql
burgerkingeater: it's still better to set it explicitly 1. different DBMS might have different rules 2. being consistent with executor_id statement as above 3. easy for reader to know the default value of this column.

burgerkingeater:
please resolve conflicts.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: I understand that creating a new string every time means they have to be garbage collected, but that's not worse than using a string constant with "?" in it I believe: a new string has to be created anyways to substitute those ? with the actual values because strings are immutable so because some elements of the query are not known at compile time there's no way to avoid creating a new string every time. The only difference is whether that new string is created by us or by the JDBC driver, but I don't think that's significant difference.

azkaban-db/src/main/sql/create.execution_flows.sql
ypadron-in: ok, but we might want to specify it for the rest of the columns as well.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
edwinalu: If the text of the String does not need to change, then we can create a constant (final static), and return this constant String. This would avoid allocating memory for each call. It's highly unlikely that a String substituting the "?" would be created. The query with the "?" can be executed on the database directly, with the bind variables specified.

ypadron-in: I changed the query strings to be constant strings using "?" to define parameters. It looks Java delegates parameter resolution to the Mysql connector. Thanks.

juhoautio:
@ypadron-in, questions about this:

> Inactive executors will only poll executions with their ids (use_executor == pollingExecutorId)

So are inactive executors still polling & dispatching executions from the queue after they have been inactivated?

Is it even allowed to submit new executions (ie. add to the queue) with use_executor == someInactiveExecutorId, after the executor has been inactivated?

ypadron-in:
> @ypadron-in, questions about this:
> 
> > Inactive executors will only poll executions with their ids (use_executor == pollingExecutorId)
> 
> So are inactive executors still polling & dispatching executions from the queue after they have been inactivated?
> 
> Is it even allowed to submit new executions (ie. add to the queue) with use_executor == someInactiveExecutorId, after the executor has been inactivated?

@juhoautio Yes to both questions. 
Our use case in this scenario of inactive executor polling and running executions is that following deployment of binaries we run test executions before activating the executor so that production flows are not affected if there is something wrong with the binaries. If tests pass then we activate the executor. 
Also useExecutor parameter will only be honored if set by an Azkaban admin, we assume is going to be used wisely.

PR #2133, opened 2019-02-26
block deactivating executor until all flow preparation work completes

burgerkingeater:
This PR makes deactivation process block until all flow preparation work finishes.

When deploying new executor, old running executor will be deactivated before new one is activated and only one executor is allowed to delete/hard-linking project directories to avoid race condition(see https://github.com/azkaban/azkaban/pull/2130). So to make deactivation process block until flow preparation work finishes guarantees the old executor won't access FlowPreparer#setup after deactivation.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -280,6 +280,9 @@ public class ExecutionFlowDao {
     }
   }
 
+  /**
+   * set executor id to null for the execution id
+   */
   public void unsetExecutorIdForExecution(final int executionId) throws ExecutorManagerException {
     final String UNSET_EXECUTOR = "UPDATE execution_flows SET executor_id = null where exec_id = ?";
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -280,6 +280,20 @@ public class ExecutionFlowDao {
     }
   }
 
+  public void unsetExecutorIdForExecution(final int executionId) throws ExecutorManagerException {
+    final String UNSET_EXECUTOR = "UPDATE execution_flows SET executor_id = null where exec_id = ?";
+
+    final SQLTransaction<Integer> unsetExecutor =
+        transOperator -> transOperator.update(UNSET_EXECUTOR, executionId);
+
+    try {
+      this.dbOperator.transaction(unsetExecutor);
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error unsetting executor id for execution " + executionId,
+          e);
+    }
+  }
+
   public int selectAndUpdateExecution(final int executorId) throws ExecutorManagerException {
     final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ? where exec_id = ?";
 
@@ -406,7 +420,7 @@ public class ExecutionFlowDao {
         final byte[] data = rs.getBytes(3);
 
         if (data == null) {
-          logger.error("Found a flow with empty data blob exec_id: " + id);
+          ExecutionFlowDao.logger.error("Found a flow with empty data blob exec_id: " + id);
         } else {
           final EncodingType encType = EncodingType.fromInteger(encodingType);
           try {
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -356,6 +356,11 @@ public class FlowRunnerManager implements EventListener,
     return false;
   }
 
+  private boolean isExecutorSpecified(final ExecutableFlow flow) {
+    return flow.getExecutionOptions().getFlowParameters()
+        .containsKey(ExecutionOptions.USE_EXECUTOR);
+  }
+
   private FlowRunner createFlowRunner(final int execId) throws ExecutorManagerException {
     final ExecutableFlow flow;
     flow = this.executorLoader.fetchExecutableFlow(execId);
@@ -367,8 +372,7 @@ public class FlowRunnerManager implements EventListener,
     // Sets up the project files and execution directory.
     this.preparingFlowCount.incrementAndGet();
     try {
-      if (this.active || flow.getExecutionOptions().getFlowParameters()
-          .containsKey(ExecutionOptions.USE_EXECUTOR)) {
+      if (this.active || isExecutorSpecified(flow)) {
         this.flowPreparer.setup(flow);
       } else {
         // Unset the executor.
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -303,8 +303,10 @@ public class FlowRunnerManager implements EventListener,
    * Wait until ongoing flow preparation work finishes.
    */
   private void waitUntilFlowPreparationFinish() {
-    while (this.preparingFlowCount.intValue() == 0) {
+    while (this.preparingFlowCount.intValue() != 0) {
       try {
+        logger.info(this.preparingFlowCount + " flow(s) is/are still being setup before complete "
+            + "deactivation.");
         Thread.sleep(Duration.ofSeconds(5).toMillis());
       } catch (final InterruptedException ex) {
         logger.debug(ex);
@@ -336,6 +338,11 @@ public class FlowRunnerManager implements EventListener,
     try {
       this.preparingFlowCount.getAndIncrement();
       runner = createFlowRunner(execId);
+      try {
+        Thread.sleep(1000 * 20);
+      } catch (final InterruptedException e) {
+        e.printStackTrace();
+      }
     } finally {
       this.preparingFlowCount.decrementAndGet();
     }
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -53,10 +53,10 @@ import azkaban.utils.ThreadPoolExecutingListener;
 import azkaban.utils.TrackingThreadPool;
 import azkaban.utils.UndefinedPropertyException;
 import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.Uninterruptibles;
 import java.io.File;
 import java.io.IOException;
 import java.lang.Thread.State;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -304,13 +304,9 @@ public class FlowRunnerManager implements EventListener,
    */
   private void waitUntilFlowPreparationFinish() {
     while (this.preparingFlowCount.intValue() != 0) {
-      try {
-        logger.info(this.preparingFlowCount + " flow(s) is/are still being setup before complete "
-            + "deactivation.");
-        Thread.sleep(Duration.ofSeconds(5).toMillis());
-      } catch (final InterruptedException ex) {
-        logger.debug(ex);
-      }
+      logger.info(this.preparingFlowCount + " flow(s) is/are still being setup before complete "
+          + "deactivation.");
+      Uninterruptibles.sleepUninterruptibly(5, TimeUnit.SECONDS);
     }
   }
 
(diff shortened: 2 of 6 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -283,6 +287,31 @@ public void setExecutorActive(final boolean isActive, final String host, final i
           "Set active action ignored. Executor is already " + (isActive ? "active" : "inactive"));
     }
     this.active = isActive;
+    if (!this.active) {
+      // When deactivating this executor, this call will return until every thread in {@link
```
edwinalu: Change "will return until every thread in {@link#createFlowRunner} to finish" to "will wait to return until every thread in {@link#createFlowRunner}  has finished".

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -283,6 +287,31 @@ public void setExecutorActive(final boolean isActive, final String host, final i
...
+   * Wait until ongoing flow preparation work finishes.
+   */
+  private void waitUntilFlowPreparationFinish() {
+    while (this.preparingFlowCount.intValue() != 0) {
+      try {
+        logger.info(this.preparingFlowCount + " flow(s) is/are still being setup before complete "
+            + "deactivation.");
+        Thread.sleep(Duration.ofSeconds(5).toMillis());
+      } catch (final InterruptedException ex) {
+        logger.debug(ex);
```
edwinalu: Interrupts should be re-raised.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -283,6 +287,31 @@ public void setExecutorActive(final boolean isActive, final String host, final i
           "Set active action ignored. Executor is already " + (isActive ? "active" : "inactive"));
     }
     this.active = isActive;
+    if (!this.active) {
```
edwinalu: It looks like active state is also tracked by Executor. Is it possible for the state to be stored in one place, instead of both Executor and FlowRunnerManager? There would be less chance of the states getting out of sync (if Executor is set but not FlowRunnerManager for example, or just before FlowRunnerManager is set) if there is just one state.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -283,6 +287,31 @@ public void setExecutorActive(final boolean isActive, final String host, final i
...
+      // deactivation.
+      waitUntilFlowPreparationFinish();
+    }
+  }
+
+  /**
+   * Wait until ongoing flow preparation work finishes.
+   */
+  private void waitUntilFlowPreparationFinish() {
+    while (this.preparingFlowCount.intValue() != 0) {
```
edwinalu: With the new dispatch logic, in pollExecution(), active is checked before selectAndUpdateExecution() and then submitFlow() is called. submitFlow() increments preparingFlowCount, so there could be some lag, and in this amount of time, it's possible that there could be a call to inactivate the executor, and preparingFlowCount would be zero.

With the original dispatch logic, I wasn't able to find a check for if the executor is active before the flow is submitted from ExecutorServlet.handleAjaxExecute().

edwinalu: One option is to add an additional check to see if it is active after incrementing preparingFlowCount. If inactive, then skip execution and decrement preparingFlowCount. Might need to update the row in the db, so that it can be picked up by another executor for the new dispatch logic, or return an error, or some status saying the flow was not executed, with the original dispatch logic.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
edwinalu: More info: https://www.ibm.com/developerworks/library/j-jtp05236/index.html

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
edwinalu: Discussed offline, and Executor is a representation of the db row, and only used in this method, so would not be available in other code looking at isActive.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -283,6 +287,27 @@ public void setExecutorActive(final boolean isActive, final String host, final i
...
+      logger.info(this.preparingFlowCount + " flow(s) is/are still being setup before complete "
+          + "deactivation.");
+      Uninterruptibles.sleepUninterruptibly(5, TimeUnit.SECONDS);
```
edwinalu: Why isn't it OK for the sleep to be interrupted?

Please add a constant for the 5.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -966,16 +994,24 @@ private void pollExecution() {
...
+          } catch (final Exception e) {
+            FlowRunnerManager.logger.error("Failed to submit flow ", e);
+          } finally {
+            FlowRunnerManager.this.preparingFlowCount.getAndDecrement();
           }
-        } catch (final Exception e) {
-          FlowRunnerManager.logger.error("Failed to submit flow ", e);
+        } else {
+          FlowRunnerManager.this.preparingFlowCount.getAndDecrement();
         }
```
edwinalu: Could a unit test be added.

burgerkingeater: will add todo given poor testability of this class.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
burgerkingeater: thanks updated.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
burgerkingeater: https://github.com/azkaban/azkaban/issues/2134

edwinalu: Could you please also file a JIRA ticket for tracking?

burgerkingeater: created internally.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -982,18 +1011,25 @@ private void pollExecution() {
...
       } else if (FlowRunnerManager.this.active) {
-        try {
-          // Todo jamiesjc: check executor capacity before polling from DB
-          final int execId = FlowRunnerManager.this.executorLoader
-              .selectAndUpdateExecution(this.executorId);
-          if (execId != -1) {
-            FlowRunnerManager.logger.info("Submitting flow " + execId);
-            submitFlow(execId);
-            commonMetrics.markDispatchSuccess();
+        FlowRunnerManager.this.preparingFlowCount.getAndIncrement();
```
edwinalu: Having the increment for preparingFlowCount here means that it is blocking for the select from the database as well, which is not needed. In general it's better to keep synchronized/blocking code as minimal and as fast as possible, so that more work can be done in parallel, and to minimize the amount of waiting.

Since this code is not called by the original dispatch logic, the original dispatch logic is still not safe (could still end up preparing flows when inactive).

burgerkingeater: updated, thanks

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -280,6 +280,20 @@ public ExecutableFlow fetchExecutableFlow(final int execId) throws ExecutorManag
     }
   }
 
+  public void unsetExecutorIdForExecution(final int executionId) throws ExecutorManagerException {
```
edwinalu: Please add some comments for the method.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -309,14 +335,18 @@ public void submitFlow(final int execId) throws ExecutorManagerException {
     if (isAlreadyRunning(execId)) {
       return;
     }
+
     final FlowRunner runner = createFlowRunner(execId);
+
+
```
edwinalu: nit: extra newline.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -309,14 +335,18 @@ public void submitFlow(final int execId) throws ExecutorManagerException {
...
     // Check again.
     if (isAlreadyRunning(execId)) {
       return;
     }
     submitFlowRunner(runner);
   }
 
+
```
edwinalu: nit: extra newline.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -341,18 +371,28 @@ private FlowRunner createFlowRunner(final int execId) throws ExecutorManagerExce
...
-        flow.getExecutableFlow().getSubmitTime());
-
-    final Timer.Context flowPrepTimerContext = commonMetrics.getFlowSetupTimerContext();
-    try {
-      // Sets up the project files and execution directory.
-      this.flowPreparer.setup(flow);
-    } finally {
-      flowPrepTimerContext.stop();
+    // Sets up the project files and execution directory.
+    if (this.active) {
```
edwinalu: This would also need an else case to call "this.executorLoader.unsetExecutorIdForExecution(execId);"

burgerkingeater: thanks, added

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
burgerkingeater: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
burgerkingeater: updated

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
burgerkingeater: added

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -341,18 +369,30 @@ private FlowRunner createFlowRunner(final int execId) throws ExecutorManagerExce
...
```
edwinalu: It would be possible to skip the outer "if (this.active)" check, since this is checked again after the increment, and would save the additional else clause, so would be more concise.

edwinalu: The if check also needs to check if use executor has been set for the flow, and still execute if true.

burgerkingeater: updated

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -333,25 +361,40 @@ private boolean isAlreadyRunning(final int execId) throws ExecutorManagerExcepti
...
         flow.getExecutableFlow().getSubmitTime());
 
     final Timer.Context flowPrepTimerContext = commonMetrics.getFlowSetupTimerContext();
+
+    // Sets up the project files and execution directory.
+    this.preparingFlowCount.incrementAndGet();
     try {
-      // Sets up the project files and execution directory.
-      this.flowPreparer.setup(flow);
+      if (this.active || isExecutorSpecified(flow)) {
```
edwinalu: Please add some comments for isExecutorSpecified(flow).

burgerkingeater: added, thanks

PR #2137, opened 2019-03-04
Project cache hit ratio metrics enhancement

burgerkingeater:
This PR replaced old project cache hit ratio metrics impl. with a new one which only keep tracks of last 100 cache accesses' hit ratio instead of all past access since Azkaban executor starts. In this way, this metrics is more sensitive to project cache setting change (e.g, cache size is changed on the fly) since hit ratio won't be offset by too much old historical data.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/metric/ProjectCacheHitRatio.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/metric/ProjectCacheHitRatio.java
@@ -18,42 +18,36 @@ package azkaban.execapp.metric;
 
 import com.codahale.metrics.RatioGauge;
 import com.codahale.metrics.SlidingWindowReservoir;
+import java.util.Arrays;
 
 /**
  * Project cache hit ratio of last 100 cache accesses.
+ *
+ * <p>The advantage of sampling last 100 caches accesses over time-based sampling like last hour's
+ * cache accesses is the former is more deterministic. Suppose there's only few execution in last
+ * hour, then hit ratio might not be truly informative, which doesn't necessarily reflect
+ * performance of the cache.</p>
  */
 public class ProjectCacheHitRatio extends RatioGauge {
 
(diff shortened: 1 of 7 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/metric/ProjectCacheHitRatio.java
```diff
@@ -0,0 +1,59 @@
...
+import com.codahale.metrics.SlidingWindowReservoir;
+
+/**
+ * Project cache hit ratio of last 100 cache accesses.
+ */
+public class ProjectCacheHitRatio extends RatioGauge {
+
+  private final SlidingWindowReservoir hits;
+  private final SlidingWindowReservoir calls;
+  public static final int WINDOW_SIZE = 100;
```
edwinalu: 100 seems small, for cases where there are a few thousand executions per day. It would be good if this is a configurable parameter (at some point, it would be good to be able to deploy configurations without a complete Azkaban deployment), or if it is time-based (last hour, etc.).

burgerkingeater: explained in class comment.

azkaban-exec-server/src/main/java/azkaban/execapp/metric/ProjectCacheHitRatio.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.codahale.metrics.RatioGauge;
+import com.codahale.metrics.SlidingWindowReservoir;
+
+/**
+ * Project cache hit ratio of last 100 cache accesses.
+ */
+public class ProjectCacheHitRatio extends RatioGauge {
+
+  private final SlidingWindowReservoir hits;
```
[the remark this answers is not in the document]
burgerkingeater: updated, thanks

PR #2139, opened 2019-03-06
Remove unnecessary code

ypadron-in:
Reason for deletion is provided in comments below.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -46,7 +46,7 @@ public class ExecutionFlowDao {
     this.dbOperator = dbOperator;
   }
 
-  public synchronized void uploadExecutableFlow(final ExecutableFlow flow)
+  public void uploadExecutableFlow(final ExecutableFlow flow)
       throws ExecutorManagerException {
 
     final String useExecutorParam =
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
@@ -122,12 +122,7 @@ public class AzkabanProcess {
       errorGobbler.awaitCompletion(5000);
 
       if (exitCode != 0) {
-        final String output =
-            new StringBuilder().append("Stdout:\n")
-                .append(outputGobbler.getRecentLog()).append("\n\n")
-                .append("Stderr:\n").append(errorGobbler.getRecentLog())
-                .append("\n").toString();
-        throw new ProcessFailureException(exitCode, output);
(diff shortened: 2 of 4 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -46,7 +46,7 @@ public ExecutionFlowDao(final DatabaseOperator dbOperator) {
     this.dbOperator = dbOperator;
   }
 
-  public synchronized void uploadExecutableFlow(final ExecutableFlow flow)
```
ypadron-in: Marked for deletion here: https://github.com/azkaban/azkaban/pull/2129#discussion_r259039490

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -943,30 +943,6 @@ private void ajaxFetchExecutableFlow(final HttpServletRequest req,
     ret.putAll(flowObj);
   }
 
-  private void ajaxAttemptExecuteFlow(final HttpServletRequest req,
```
ypadron-in: This method duplicated work done in ajaxExecuteFlow(..) and is not used anywhere else.

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
```diff
@@ -122,12 +122,7 @@ public void run() throws IOException {
       errorGobbler.awaitCompletion(5000);
 
       if (exitCode != 0) {
-        final String output =
```
ypadron-in: This info should be in the logs already. Also, it wasn't used before because toString/getMessage methods of ProcessFailureException were not overwritten.

azkaban-common/src/main/java/azkaban/jobExecutor/utils/process/AzkabanProcess.java
```diff
@@ -122,12 +122,7 @@ public void run() throws IOException {
...
-            new StringBuilder().append("Stdout:\n")
-                .append(outputGobbler.getRecentLog()).append("\n\n")
-                .append("Stderr:\n").append(errorGobbler.getRecentLog())
-                .append("\n").toString();
```
ypadron-in: This info should be in the logs already. Also, it wasn't used before because toString/getMessage methods of ProcessFailureException were not overwritten.

ypadron-in:
> Thanks for cleaning this up! Is there more information about AzkabanProcess and LogGobbler? There's some comments about "Output is read by separate threads to avoid deadlock and logged to log4j loggers." I know this is very old code, so most likely the information is lost, but am curious about the issues.

No problem :). 
Sorry, I don't know if there is more info about AzkabanProcess and LogGobble.

PR #2140, opened 2019-03-08
Implement "flowPriority" (at the time of polling) feature for new Dispatching Logic

ypadron-in:
"flowPriority" is one of the execution options of a flow, which only takes effect if set by an Azkaban admin. By default all flows have a priority of 5 but if a bigger number is specified from the UI that flow will be dispatched for execution first.

With the new Dispatching Logic(#2038) setting a high "flowPriority" means that flow will be polled first by executors.
Note that it has been said "polled first" not "executed first". On the executor server there is a queue to hold executions when the maximum number of threads availables/setup to run flows is reached. This queue is NOT a priority queue.

To implement this feature for the new Dispatching Logic, a new column "flow_priority" needs to be added to table "execution_flows" to speed up access to this info (currently flowPriority is stored in "flow_data" longblob column). When new executions are inserted into "execution_flows" table, user-specified or default flow priority will be inserted as well. Later, executors will poll executions sorting by
`flow_priority DESC, submit_time ASC, exec_id ASC`

Adding an index on flow_priority column or a multicolumn index on (flow_priority, submit_time) for example, is not necessary because "status" column is already indexed and this column is more selective than "flow_priority" so MySQL should pick this index over the rest whenever polling executions.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -61,9 +61,7 @@ public class ExecutionFlowDao {
     final String INSERT_EXECUTABLE_FLOW = "INSERT INTO execution_flows "
         + "(project_id, flow_id, version, status, submit_time, submit_user, update_time, "
         + "use_executor, flow_priority) values (?,?,?,?,?,?,?,?,?)";
-    final long submitTime = System.currentTimeMillis();
-    flow.setStatus(Status.PREPARING);
-    flow.setSubmitTime(submitTime);
+    final long submitTime = flow.getSubmitTime();
 
     /**
      * Why we need a transaction to get last insert ID?
@@ -73,7 +71,7 @@ public class ExecutionFlowDao {
      */
     final SQLTransaction<Long> insertAndGetLastID = transOperator -> {
       transOperator.update(INSERT_EXECUTABLE_FLOW, flow.getProjectId(),
-          flow.getFlowId(), flow.getVersion(), Status.PREPARING.getNumVal(),
+          flow.getFlowId(), flow.getVersion(), flow.getStatus().getNumVal(),
           submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority);
       transOperator.getConnection().commit();
       return transOperator.getLastInsertId();
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -53,9 +53,14 @@ public class ExecutionFlowDao {
         flow.getExecutionOptions().getFlowParameters().get(ExecutionOptions.USE_EXECUTOR);
     final String executorId = StringUtils.isNotEmpty(useExecutorParam) ? useExecutorParam : null;
 
+    final String flowPriorityParam =
+        flow.getExecutionOptions().getFlowParameters().get(ExecutionOptions.FLOW_PRIORITY);
+    final int flowPriority = StringUtils.isNotEmpty(flowPriorityParam) ?
+        Integer.parseInt(flowPriorityParam) : ExecutionOptions.DEFAULT_FLOW_PRIORITY;
+
     final String INSERT_EXECUTABLE_FLOW = "INSERT INTO execution_flows "
         + "(project_id, flow_id, version, status, submit_time, submit_user, update_time, "
-        + "use_executor) values (?,?,?,?,?,?,?,?)";
+        + "use_executor, flow_priority) values (?,?,?,?,?,?,?,?,?)";
     final long submitTime = System.currentTimeMillis();
     flow.setStatus(Status.PREPARING);
     flow.setSubmitTime(submitTime);
@@ -69,7 +74,7 @@ public class ExecutionFlowDao {
     final SQLTransaction<Long> insertAndGetLastID = transOperator -> {
       transOperator.update(INSERT_EXECUTABLE_FLOW, flow.getProjectId(),
           flow.getFlowId(), flow.getVersion(), Status.PREPARING.getNumVal(),
-          submitTime, flow.getSubmitUser(), submitTime, executorId);
+          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority);
       transOperator.getConnection().commit();
       return transOperator.getLastInsertId();
     };
@@ -337,7 +342,7 @@ public class ExecutionFlowDao {
     private static final String SELECT_EXECUTION_FOR_UPDATE_FORMAT =
         "SELECT exec_id from execution_flows WHERE status = " + Status.PREPARING.getNumVal()
             + " and executor_id is NULL and flow_data is NOT NULL and %s"
-            + " ORDER BY submit_time ASC LIMIT 1 FOR UPDATE";
+            + " ORDER BY flow_priority DESC, submit_time ASC, exec_id ASC LIMIT 1 FOR UPDATE";
 
     public static final String SELECT_EXECUTION_FOR_UPDATE_ACTIVE =
         String.format(SELECT_EXECUTION_FOR_UPDATE_FORMAT,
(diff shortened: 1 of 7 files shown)


azkaban-db/src/main/sql/upgrade.3.69.0.to.3.70.0.sql
```diff
@@ -0,0 +1,8 @@
+-- DB Migration from release 3.69.0 to 3.70.0
+-- PR #xxxx Implement “flowPriority” (at the time of polling) feature for new dispatching logic.
```
burgerkingeater: please fill in the PR id.

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -13,6 +13,7 @@ CREATE TABLE execution_flows (
   flow_data   LONGBLOB,
   executor_id INT                   DEFAULT NULL,
   use_executor INT                  DEFAULT NULL,
+  flow_priority INT                 DEFAULT 5,
```
burgerkingeater: consider enum type? Then priority would be limited instead of any integer.

edwinalu:
Could you please check the query plan, for the new select with flow_priority, if not done already?

azkaban-db/src/main/sql/upgrade.3.69.0.to.3.70.0.sql
ypadron-in: yes, will do

ypadron-in:
> Could you please check the query plan, for the new select with flow_priority, if not done already?

Do you mean this:
```
mysql> EXPLAIN EXTENDED SELECT exec_id from execution_flows where status = 20 and executor_id is NULL and flow_data is NOT NULL and use_executor is NULL ORDER BY flow_priority DESC, submit_time ASC, exec_id ASC LIMIT 1 FOR UPDATE;
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
| id | select_type | table           | type | possible_keys              | key            | key_len | ref   | rows | filtered | Extra                       |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
|  1 | SIMPLE      | execution_flows | ref  | executor_id,ex_flows_staus | ex_flows_staus | 2       | const |    1 |   100.00 | Using where; Using filesort |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
1 row in set, 1 warning (0.01 sec)
```
This DB has over 10 million rows in execution_flows table

edwinalu:
Yes, thanks. How does the explain plan change if an index is added on (flow_priority, submit_time), or (flow_priority, submit_time, exec_id)? Each executor is polling with this query, so it's getting executed very frequently.
https://dev.mysql.com/doc/refman/8.0/en/order-by-optimization.html

azkaban-db/src/main/sql/create.execution_flows.sql
ypadron-in: MySQL ENUM type? I think it will artificially limit what we can do with priorities in the future or make it hard to maintain at least. There have been comments around setting priorities based on groups. For example: groupA could set priorities from 1-10, groupB from 11-20, etc

ypadron-in:
> Yes, thanks. How does the explain plan change if an index is added on (flow_priority, submit_time), or (flow_priority, submit_time, exec_id)? Each executor is polling with this query, so it's getting executed very frequently.
> https://dev.mysql.com/doc/refman/8.0/en/order-by-optimization.html

Here is the EXPLAIN output after adding (flow_priority, submit_time) index:
`mysql> ALTER TABLE execution_flows ADD INDEX priority_submit (flow_priority DESC, submit_time ASC);`
```
mysql> EXPLAIN EXTENDED SELECT exec_id from execution_flows WHERE status = 20 and executor_id is NULL and flow_data is NOT NULL and use_executor is NULL ORDER BY flow_priority DESC, submit_time ASC, exec_id ASC LIMIT 1 FOR UPDATE;
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
| id | select_type | table           | type | possible_keys              | key            | key_len | ref   | rows | filtered | Extra                       |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
|  1 | SIMPLE      | execution_flows | ref  | executor_id,ex_flows_staus | ex_flows_staus | 2       | const |    7 |   100.00 | Using where; Using filesort |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
1 row in set, 1 warning (0.00 sec)
```

Here is the output if we dropped index (flow_priority, submit_time) and create instead an index on (flow_priority, submit_time, exec_id)
`mysql> ALTER TABLE execution_flows ADD INDEX priority_submit_execid (flow_priority DESC, submit_time ASC, exec_id ASC);`
```
mysql> EXPLAIN EXTENDED SELECT exec_id from execution_flows WHERE status = 20 and executor_id is NULL and flow_data is NOT NULL and use_executor is NULL ORDER BY flow_priority DESC, submit_time ASC, exec_id ASC LIMIT 1 FOR UPDATE;
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
| id | select_type | table           | type | possible_keys              | key            | key_len | ref   | rows | filtered | Extra                       |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
|  1 | SIMPLE      | execution_flows | ref  | executor_id,ex_flows_staus | ex_flows_staus | 2       | const |    7 |   100.00 | Using where; Using filesort |
+----+-------------+-----------------+------+----------------------------+----------------+---------+-------+------+----------+-----------------------------+
1 row in set, 1 warning (0.00 sec)
```

edwinalu:
Thanks for checking the query plan for the different index variations. As you had mentioned, it is choosing to use the more selective index on status, and the additional index is not being used.

azkaban-db/src/main/sql/upgrade.3.69.0.to.3.70.0.sql
```diff
@@ -0,0 +1,8 @@
+-- DB Migration from release 3.69.0 to 3.70.0
+-- PR #2140 Implement “flowPriority” (at the time of polling) feature for new dispatching logic.
```
burgerkingeater: nit: *implements

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -337,7 +342,7 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
     private static final String SELECT_EXECUTION_FOR_UPDATE_FORMAT =
         "SELECT exec_id from execution_flows WHERE status = " + Status.PREPARING.getNumVal()
             + " and executor_id is NULL and flow_data is NOT NULL and %s"
-            + " ORDER BY submit_time ASC LIMIT 1 FOR UPDATE";
+            + " ORDER BY flow_priority DESC, submit_time ASC, exec_id ASC LIMIT 1 FOR UPDATE";
```
burgerkingeater: is there any chance that priority-based dispatching cause starvation? Should we also take waiting time into consideration?

azkaban-db/src/main/sql/create.execution_flows.sql
burgerkingeater: discussed offline, we agree to it's better to have a fixed shorter range for priority and 
enforce flow priority range limitation on application side.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: Thanks! It is possible right now. I'll modify the query to prevent starvation.

ypadron-in: Talked about this topic offline. We decided to not rollout a fix to prevent starvation of low priority executions right now. "flowPriority" feature is only available to Azkaban admins and we consider it's not very used. 
We will monitor the polling of executions to detect the moment when starvation becomes a real issue to take into consideration.
If anyone in the open source community needs a fix sooner, please refer to this commit
https://github.com/ypadron-in/azkaban/commit/0c0fe3aa2f03ad5e96b3291a20ea8937776b1f77

burgerkingeater: thanks!

PR #2144, opened 2019-03-11
Replace Counter with AtomicLong for counter-based metrics

burgerkingeater:
We should leverage metrics data type available in https://metrics.dropwizard.io/3.1.0/getting-started/ as much as possible.


HappyRay:
Could you elaborate the benefit of doing so in the PR description?

It seems that the counter is backed by AtomicLong too.

Thanks.

burgerkingeater:
@HappyRay yes it is, given that Counter should be preferred since it is a native type in metrics lib.

burgerkingeater:
@HappyRay let me know if you have more questions with this change.

PR #2150, opened 2019-03-20
Propagate select properties from flow or job to event metadata

abti:
In certain scenarios, event reporter framework can benefit by enriching its reporting metadata by including custom flow or job properties. 
An example use-case where this can be useful is when a user wants to tag all their flows (of a specific type through flow properties) for better processing of their flow execution metadata downstream. 

This PR enables the user to specify the property keys (in a comma separated list) that they want to propagate, and values for those properties is included in the event metadata. 
Eg. 
If the flow Yaml looked like: 
```
config:
  user.to.proxy: abti
  tags: ai,ml
  category: training
  azkaban.event.reporting.propagateProperties: tags,category

nodes:
  - name: Job1
    type: java
    config:
      key1: val1
      key2: val2
      azkaban.event.reporting.propagateProperties: key1,key2

  - name: Job2
    type: hive
    config:
      key3: val3
      key4: val4
      azkaban.event.reporting.propagateProperties: key3,key4
    dependsOn:
      - Job1
```
Then, flow event reporter will include values for tags and category in its metadata; job event reporter for Job1 will include values for key1 and key2 in its metadata; and job event reporter for Job2 will include values for key3 and key4 in its metadata. 

Future improvement can include regex support for azkaban.event.reporting.propagateProperties


abti:
@burgerkingeater @li-ygerchikov please review.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1457,4 +1470,42 @@ public synchronized void handleEvent(final Event event) {
...
+  /***
+   * Propagate properties (specified in {@code AZKABAN_EVENT_REPORTING_PROPERTIES_TO_PROPAGATE})
+   * to metadata for event reporting.
+   * @param metaData Metadata map to update with properties.
+   * @param inputProps Input properties for flow or job.
+   * @param nodeType Flow or job.
+   * @param nodeName Flow or job name.
+   * @param logger Logger from invoking class for log sanity.
+   */
+  public static void propagateMetadataFromProps(Map<String, String> metaData, Props inputProps, String nodeType,
```
burgerkingeater: can we make the method package-private and annotate with @VisibleForTesting?

abti: Done

abti:
> Looks good to me!
> If I were picky, and cared about test coverage, I might perhaps add some negative tests as well, but that's just to be pedantic.
> Also, should we at some point document the new property somewhere (besides code)?

@li-ygerchikov Thanks. I added a negative and backward compatibility test case. I will create a separate PR with doc changes may be? Need to explore if we can open source KafkaEventReporter to paint a full story. 

@li-ygerchikov @burgerkingeater I have addressed comments, and tested with Flow 1.0, Flow 2.0, their jobs and manual property overrides with KafkaEventReporter. So, this is now ready for review and merge (if all looks good?).

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1366,6 +1371,24 @@ public FlowRunnerEventListener() {
...
+
+      // Propagate flow properties to Event Reporter
+      if (FlowLoaderUtils.isAzkabanFlowVersion20(flow.getAzkabanFlowVersion())) {
+        // In Flow 2.0, flow has designated properties (defined at its own level in Yaml)
+        propagateMetadataFromProps(metaData, flow.getInputProps(), "flow", flow.getId(), logger);
+      } else {
+        // In Flow 1.0, flow properties are combination of shared properties in individual files (order not defined,
+        // .. because it's loaded by fs list order and put in a HashMap).
+        Props combinedProps = null;
+        for (Props sharedProp : flowRunner.sharedProps.values()) {
```
edwinalu: It looks like Props is meant to be hierarchical, with a parent Props, and the current props as a Map. sharedProps is a Map with values as Props, but there doesn't seem to be any order to these, hierarchical or otherwise. Why is a hierarchical graph of Props being created? Would creating the props at the same level make sense?

Would the different Props have the duplicate properties? If so, what is the precedence, if any?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1457,4 +1484,49 @@ public synchronized void handleEvent(final Event event) {
...
+    if (null == metaData || null == inputProps || null == logger ||
+        Strings.isNullOrEmpty(nodeType) || Strings.isNullOrEmpty(nodeName)) {
+      throw new IllegalArgumentException("Input params should not be null or empty.");
+    }
+
+    final String propsToPropagate = inputProps.getString(AZKABAN_EVENT_REPORTING_PROPERTIES_TO_PROPAGATE);
+    if (Strings.isNullOrEmpty(propsToPropagate)) {
+      // Nothing to propagate
+      logger.info(String.format("No properties to propagate to metadata for %s: %s", nodeType, nodeName));
+
```
edwinalu: nit: extra newline.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
abti: In Flow 1.0, flow specifically does not has dedicated properties (except shared props that it maintains for sharing with all jobs). 
The shared props itself is generated by fs ls on property files and puts in HashMap thereby losing any ordering guarantees. 
So, I have simply 'attached' the share props together in one shared props for simplicity (marking one as parent of other). So that when we do a lookup, it looks in one shared prop and if it does not finds the property, then it goes on to look into another (parent). 
Alternatively, I could have iterated over all the shared props one by one .. and stopped when I would have found a property key. But that is still affected by no natural ordering of HashMap key value pairs. 

So, to answer your question, there is no presumed order or precedence within shared props.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
abti: Will remove

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
edwinalu: Could the properties be created at one level then? Right now it is creating a hierarchical order, which indicates an ordering that we're not really guaranteeing. There is the following constructor:

  /**
   * Create properties from maps of properties
   */
  public Props(final Props parent, final Map<String, String>... props) {
    this(parent);
    for (int i = props.length - 1; i >= 0; i--) {
      this.putAll(props[i]);
    }
  }

abti: As discussed offline, changing to iterative combination of map to make code reflective of intent (though behavior remains unchanged). 
While at it, figured out that props.putAll takes care of resolving parent configs. so didn't have to write recursive resolution.

abti:
Thanks @edwinalu @burgerkingeater @li-ygerchikov

PR #2151, opened 2019-03-20
Cache cleanup chunking MVP

li-ygerchikov:
Currently, for regularly scheduled jobs that are scheduled/dispatched via the web server dispatch thread, dispatch is single threaded through web server and executor. The web server makes an ajax call to the executor (ExecutorManager.dispatch()), to execute the flow. On the executor (ExecutorServlet.handleAjaxExecute()), before it can return from the call, it does the work for flow preparation. There is no timeout on the web server, and it will wait until the call to the executor returns before continuing. Thus no other flows are dispatched to any executors while it is waiting.

In order to reduce blocking and make executors more responsive, this change limits the time executor spends cleaning up the project cache in each invocation of FlowPreparer.setup().

This is a concept demonstration only. A more robust implementation would probably allow max # of dirs cleaned up per invocation to be parameterised, with some reasonable default provided as well. Tests, too, would need to be added.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
@@ -155,7 +155,7 @@ class ProjectCacheCleaner {
     final Set<File> projectDirsToDelete = new HashSet<>();
 
     for (final ProjectDirectoryMetadata proj : projectDirMetadataList) {
-      if (sizeToFreeInBytes > 0) {
+      if (sizeToFreeInBytes > 0 || projectDirsToDelete.size() < 5) {
         // Delete the project directory even if flow within is running. It's OK to
         // delete the directory since execution dir is HARD linked to project dir. Note that even
         // if project is deleted, disk space will be freed up only when all associated execution


azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -155,7 +155,7 @@ private void deleteLeastRecentlyUsedProjects(long sizeToFreeInBytes,
     final Set<File> projectDirsToDelete = new HashSet<>();
 
     for (final ProjectDirectoryMetadata proj : projectDirMetadataList) {
-      if (sizeToFreeInBytes > 0) {
+      if (sizeToFreeInBytes > 0 || projectDirsToDelete.size() < 5) {
```
abti: Can we make 5 configurable with default value of 5? And why 5? :)

abti: When you graduate this from MVP, please make this configurable (with default value of 5? but why 5 would still be a question :) )

li-ygerchikov: ...or "if". 
And yes, of course, as mentioned above:
> This is a concept demonstration only. A more robust implementation would probably allow max # of dirs cleaned up per invocation to be parameterised, with some reasonable default provided as well. (Between two and five would be a reasonable default to guarantee that we are deleting more than we’re adding, on average.) Tests, too, would need to be added.

burgerkingeater:
how would this ensure that the disk usage is below a certain threshold. If not, alert will still fire.

li-ygerchikov:
> how would this ensure that the disk usage is below the threshold. If not, alert will still fire.

This does not ensure disk usage below threshold. 

If disk is filled up by uses *other* than project cache (e.g. executions storing local data), and if it is filled faster than we can clean it up, then neither this nor the original unchanged algorithm is going to prevent the alert.

This change ensures that project cache is being cleaned up *faster than it is being filled*, because for every new flow that is being prepared we would remove up to five (or other configurable number, potentially) projects from cache. It also ensures that executor remains more responsive if a large number of projects need to be deleted from cache. The projects that are not deleted immediately on the first pass of `FlowPreparer.setup()` will be deleted eventually on subsequent passes. In this sense the behavior and outcome is not different from the original unchanged implementation. The difference is that that outcome may take longer to achieve, which is the (low) price for responsive executor.

Does this address your concern, @burgerkingeater? If not, what do you propose to change?

PR #2154, opened 2019-03-25
Executor Tags

toby-allsopp:
NOT COMPLETE - missing tests and documentation

Here's a draft of a feature that we would like in Azkaban - executor tags.

The problem that we want to solve is that not all of our executors are the same - there might be some running on Linux and some running on Windows (a separate PR with improvements to Windows support will come later) and there might be different external tools installed on different executors. Our flows require different things from their execution environments, for example some might involve running PowerShell scripts and require a Windows machine while others involve bash scripts and require a Linux machine.

The solution that I'm proposing is to associate with each executor a set of "executor tags" (where a tag is just a string) that represent the capabilities of that executor and to associate with each flow a set of executor tags that represent the requirements of that flow. Then, when an executor looks for flows to execute (I've only considered the polling model) it only considers those flows whose requirements it can satisfy.

I'd like to contribute this change to Azkaban when it's ready, so please let me know if the approach I'm taking makes sense and stands a chance of being merged. Any suggestions on how to go about this differently would also be appreciated.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -94,6 +94,9 @@ public class Constants {
     public static final String AZKABAN_POLL_MODEL = "azkaban.poll.model";
     public static final String AZKABAN_POLLING_INTERVAL_MS = "azkaban.polling.interval.ms";
 
+    // Configures comma-separated executor tags provided by a particular executor
+    public static final String AZKABAN_EXECUTOR_TAGS = "azkaban.executor.tags";
+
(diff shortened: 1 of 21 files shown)


toby-allsopp:
I'm wondering if it might be a preferable solution to evaluate the tag criteria in Java code rather than in SQL. `selectAndUpdateExecution` could loop over the selected rows checking suitability. This would allow logging of _why_ certain executions aren't being picked up which would make it easier to see what's going on. Doing it in Java code would also allow for more-complicated criteria.

Thoughts?

PR #2155, opened 2019-03-25
Improvements to Windows support

toby-allsopp:
This is by no means complete Windows support but it is sufficient to get it building, tests passing and basic functionality working on Windows.


burgerkingeater:
@toby-allsopp  thanks for contributing this PR. There're several outstanding items when it comes to windows support. Few I can think of 1. execute as user support , currently azkaban uses execute-as-users.c to run a job process as user process to achieve isolation. Not sure if it will work with windows. 2. OsMemoryUtil is designed for linux only.

toby-allsopp:
Yes, this PR doesn't address nearly all of the issues on Windows. For our purposes the things you mention are not important because we don't require different flows to execute as different users and we don't really care about memory usage. I understand that others will have different requirements though.

I'm not sure if something like the execute-as-user functionality can be done on Windows without running the exec server itself with Administrator permissions, which doesn't seem like a great idea. I could be wrong as I'm not an expert on Windows security.

The memory information can probably be implemented for Windows fairly easily but as I said it's not a priority for us.

PR #2156, opened 2019-03-25
JobRunner: Quote JVM args

toby-allsopp:
This allows job names to have spaces in them (but not single quotes).

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -49,7 +49,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
-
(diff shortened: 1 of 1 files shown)


burgerkingeater:
@toby-allsopp thanks, can you elaborate what the benefit is for this PR?

toby-allsopp:
The benefit for me is so that I can have job ids containing spaces. If there was a way to specify a display name for a job that could contain any characters and that was shown in the web UI instead of the id then that would work for me as well.

burgerkingeater:
i see. Can you fix the build failure? thanks

toby-allsopp:
The failing test (`testAlertOnFlowFinished`) appears to fail intermittently. I don't think it;s related to these changes but I'll look into it.

toby-allsopp:
I've created #2160 to address the flakey unit tests. If that gets merged then hopefully this PR will pass.

burgerkingeater:
thanks!

PR #2161, opened 2019-03-26
Quartz scheduler refactor

burgerkingeater:
This PR refactors quartz scheduler:
1. Remove unused methods.
2. Reduce visibility of methods that are only used in unit tests.
3. When pause/resume/unschedule a non-existing job, make them return false instead of throwing exception to distinguish this case from exceptional cases. Same apply for schedule an existing job.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
+++ b/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
@@ -149,7 +149,7 @@ public class QuartzScheduler {
    * @return true if job is found and unscheduled.
    * @throws SchedulerException
    */
-  public synchronized boolean unschedule(final String jobName, final String groupName) throws
+  public synchronized boolean unscheduleJob(final String jobName, final String groupName) throws
       SchedulerException {
     return this.scheduler.deleteJob(new JobKey(jobName, groupName));
   }
@@ -169,7 +169,7 @@ public class QuartzScheduler {
    *
    * @return true if job has been scheduled, false if the same job exists already.
    */
-  public synchronized boolean scheduleIfAbsent(final String cronExpression, final QuartzJobDescription
+  public synchronized boolean scheduleJobIfAbsent(final String cronExpression, final QuartzJobDescription
       jobDescription) throws SchedulerException {
 
     requireNonNull(jobDescription, "jobDescription is null");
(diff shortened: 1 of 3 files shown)


azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -148,23 +126,32 @@ public synchronized boolean isJobPaused(final String jobName, final String group
...
-   * Unregister a job given the groupname. Since unregister might be called when
-   * concurrently removing projects, so synchronized is added to ensure thread safety.
+   * Unschedule a job.
+   * @param jobName
+   * @param groupName
+   * @return true if job is found and unscheduled.
+   * @throws SchedulerException
    */
-  public synchronized void unregisterJob(final String jobName, final String groupName) throws
+  public synchronized boolean unschedule(final String jobName, final String groupName) throws
```
kunkun-tang: How about unscheduleJob. The method name better to be consistent.

azkaban-web-server/src/test/java/azkaban/scheduler/QuartzSchedulerTest.java
```diff
@@ -87,74 +85,60 @@ public void init() {
...
-    scheduler.pause();
-    final int count = SampleQuartzJob.COUNT_EXECUTION;
-    Thread.sleep(1500);
-    assertThat(SampleQuartzJob.COUNT_EXECUTION).isEqualTo(count);
-    scheduler.resume();
-    Thread.sleep(1200);
-    assertThat(SampleQuartzJob.COUNT_EXECUTION).isGreaterThan(count);
-  }
-
   private QuartzJobDescription createJobDescription() {
```
kunkun-tang: I'm feeling the naming <JobDescription> might be not appropriate. We may think of a better name.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
burgerkingeater: udpated

burgerkingeater:
@kunkun-tang when a project containing flows with flow trigger is being uploaded, az will schedule those flow trigger bind flows. But when scheduling a job fails due to exceptional conditions e.g  db connection issue, Azkaban should be able to properly handle those error handling - like retry. So quartz scheduler should be able to distinguish those cases from case where same schedule already exists. Currently both cases will throw same exception.

PR #2163, opened 2019-03-27
Revert "Optimize gradle build time"

edwinalu:
Reverts azkaban/azkaban#2113.

Quote from @ypadron-in 
This change exposed unit tests failures when tests are run using MacOS's console with ./gradlew cleanTest test command. We will revert it temporarily until we fix the tests.

Failing tests: createErrorEmail, createFirstErrorMessage, createSuccessEmail and testDispatchMultipleRetries

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -179,9 +179,7 @@ subprojects {
         test {
             // Set maxParallelForks to a large number and let gradle to force it to a the
             // max-workers number when needed.
-            // NOTE! For some reason anything more than 1 doesn't work well for azkaban build:
-            // any maxParallelForks > 1 seems to make `./gradlew cleanTest test` ~4 times slower
-            maxParallelForks = 1
+            maxParallelForks = 12
         }
     }
 


burgerkingeater:
@juhoautio #2113 caused unit test failure when building locally.

juhoautio:
@edwinalu @ypadron-in @burgerkingeater How did you verify that local test failures happen because of no forking? I can't get any failures even if I flip it back to 1.

That 12 forks setting is making the build much slower for me at least, so would be nice to get it back to 1.

juhoautio:
@ypadron-in did https://github.com/azkaban/azkaban/pull/2164 actually fix those errors, so that maxParallelForks = 1 could be restored?

ypadron-in:
> @edwinalu @ypadron-in @burgerkingeater How did you verify that local test failures happen because of no forking? I can't get any failures even if I flip it back to 1.
> 
> That 12 forks setting is making the build much slower for me at least, so would be nice to get it back to 1.

@juhoautio Because the build failed everytime on same tests when using maxParallelForks = 1.
(Using this command in a Mac console to build the project: ./gradlew build)

ypadron-in:
> @ypadron-in did #2164 actually fix those errors, so that maxParallelForks = 1 could be restored?

#2164 fixed the test failures happening all the time. Particularly in my case I still get failures on
testDispatchMultipleRetries() and testDispatchException() with some frequency but I guess we could set maxParallelForks = 1 again.

juhoautio:
Ok, here goes: https://github.com/azkaban/azkaban/pull/2242

PR #2165, opened 2019-03-27
Closes #1066. Fix HadoopSecurityManager_H_2_0 to use addDelegationTokens API instead of deprecated/private getDelegationToken

### LINKED ISSUE #1066: Remove usage of Hadoop private internal API FileSystem#getDelegationToken
`HadoopSecurityManager_H_2_0` currently uses `FileSystem#getDelegationToken()` to retrieve delegation tokens for the default `FileSystem`. See the Javadoc of this method:

```java
  /**
   * Get a new delegation token for this file system.
   * This is an internal method that should have been declared protected
   * but wasn't historically.
   * Callers should use {@link #addDelegationTokens(String, Credentials)}
   * 
   * @param renewer the account name that is allowed to renew the token.
   * @return a new delegation token
   * @throws IOException
   */
  @InterfaceAudience.Private()
  public Token<?> getDelegationToken(String renewer) throws IOException {
```
The current usage will break if e.g. the default FS is `ViewFileSystem` and federation is being used, in which case `addDelegationTokens` will properly fetch all necessary DTs, but `getDelegationToken` will not.

xkrogen:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -549,20 +549,19 @@ private void fetchNameNodeToken(final String userToProxy, final Props props, fin
...
         logger.error("Failed to fetch DFS token for ");
         throw new HadoopSecurityManagerException(
             "Failed to fetch DFS token for " + userToProxy);
       }
 
-      logger.info(String
-          .format("DFS token from namenode pre-fetched, token kind: %s, token service: %s",
-              fsToken.getKind(), fsToken.getService()));
-
-      cred.addToken(fsToken.getService(), fsToken);
```
burgerkingeater: why is `cred.addToken(fsToken.getService(), fsToken)` deleted?

xkrogen:
The `addDelegationTokens` method already adds the tokens to the provided
Credential.

PR #2168, opened 2019-03-28
Session monitoring enhancement

burgerkingeater:
1. The metrics being added shows the number of user sessions in the system, which can be leveraged to determine the session cache size appropriately.
2. Logging showing the number of sessions with same IP is added to help detect DDoS.


azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -409,6 +410,10 @@ protected void handleAjaxLoginAction(final HttpServletRequest req,
       cookie.setPath("/");
       resp.addCookie(cookie);
       getApplication().getSessionCache().addSession(session);
+      final Set<Session> sessionsOfSameIP =
+          getApplication().getSessionCache().findSessionsByIP(session.getIp());
+      // Check potential DDoS attack by bad hosts.
+      logger.info(sessionsOfSameIP.size() + " session(s) found from this IP " + session.getIp());
```
ypadron-in: Can we add the log message when there are more than one sessions per ip?
BTW, not all sessions are registered in SessionCache. Here `LoginAbstractAzkabanServlet.doPost()` sessions are temporarily created but are not registered. Do you know the reason behind it?

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -71,6 +73,10 @@ private void setupAllMetrics() {
     this.queueWaitMeter = this.metricsManager.addHistogram(QUEUE_WAIT_HISTOGRAM_NAME);
   }
 
+  public <T> void addSessionCount(final Supplier<T> getSessionCount) {
+    this.metricsManager.addGauge(SESSION_COUNT, getSessionCount);
```
ypadron-in: Have you thought about registering this metric after SessionCache's initialization, outside its constructor? Maybe SessionCache doesn't need to know about metrics existence.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -46,13 +52,16 @@
...
+  public SessionCache(final Props props, @Nullable final CommonMetrics metrics) {
     this.effectiveSessionTimeToLive = props.getLong(ConfigurationKeys.SESSION_TIME_TO_LIVE,
         DEFAULT_SESSION_TIME_TO_LIVE);
     this.cache = CacheBuilder.newBuilder()
         .maximumSize(props.getInt("max.num.sessions", MAX_NUM_SESSIONS))
-        .expireAfterAccess(effectiveSessionTimeToLive, TimeUnit.MILLISECONDS)
+        .expireAfterAccess(this.effectiveSessionTimeToLive, TimeUnit.MILLISECONDS)
         .build();
+    if (metrics != null) {
+      metrics.addSessionCount(this::getSessionCount);
```
ypadron-in: Have you thought about registering this metric after SessionCache's initialization, outside its constructor? Maybe SessionCache doesn't need to know about metrics existence.

burgerkingeater: good point, updated.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
burgerkingeater: enrich the logging to have more info.

Temp session is used when user specify username/password in their non-login request. But I do think this is an anti-pattern in terms of security. Let's see if we can deprecate this behavior.

azkaban-web-server/src/test/java/azkaban/fixture/MockLoginAzkabanServlet.java
```diff
@@ -107,7 +108,7 @@ public static MockLoginAzkabanServlet getServletWithSession(final String session
 
     // Create a concrete SessionCache so a session will get persisted
     // and can get looked up
-    final SessionCache cache = new SessionCache(props);
+    final SessionCache cache = new SessionCache(props, null);
```
ypadron-in: Can you update constructor parameters here?

burgerkingeater: thanks, updated

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -181,6 +181,8 @@ public AzkabanWebServer(final Props props,
     this.scheduler = requireNonNull(scheduler, "scheduler is null.");
     this.flowTriggerService = requireNonNull(flowTriggerService, "flow trigger service is null");
 
+    this.metricsManager.addGauge("session-count", this.sessionCache::getSessionCount);
```
ypadron-in: Should we move this line to startWebMetrics()?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -409,6 +412,14 @@ protected void handleAjaxLoginAction(final HttpServletRequest req,
...
       resp.addCookie(cookie);
       getApplication().getSessionCache().addSession(session);
+      final Set<Session> sessionsOfSameIP =
+          getApplication().getSessionCache().findSessionsByIP(session.getIp());
+
+      // Check potential DDoS attack by bad hosts.
+      logger.info(
+          "Session id created for user '" + session.getUser().getUserId() + "' and ip " + session
+              .getIp() + ", " + sessionsOfSameIP.size() + " session(s) found from this IP "
+              + session.getIp());
```
ypadron-in: Do we need to repeat the ip information?

burgerkingeater: good catch

PR #2171, opened 2019-04-01
Change ExecutorManager to ExecutorManagerAdapter in Flow Trigger

burgerkingeater:
Given ExecutorManager is no longer used in new dispatch logic and replaced with ExecutionController(https://github.com/azkaban/azkaban/pull/2039), replace it with ExecutorManagerAdapter  in flow trigger-related code.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
@@ -68,6 +68,7 @@ import org.joda.time.DateTime;
 
 /**
  * Executor manager used to manage the client side job.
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -70,6 +70,7 @@
  * Executor manager used to manage the client side job.
  */
 @Singleton
+@Deprecated
```
ypadron-in: Can you add a comment to specify the class that's replacing this one?

burgerkingeater: added, thanks

PR #2172, opened 2019-04-01
Auto reload azkaban-users.xml if there is any change in it.

djaiswal83:
The logic could be extended to reload any configuration file.
It uses WatchService to detect changes in the files.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
+++ b/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
@@ -4,5 +4,5 @@ package azkaban.user;
  * Lambda interface for parsing user config file.
  */
 public interface ParseConfigFile {
-  public void parseConfigFile();
+  void parseConfigFile();
 }
--- a/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
+++ b/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
@@ -1,0 +1,8 @@
+package azkaban.user;
+
+/**
+ * Lambda interface for parsing user config file.
+ */
+public interface ParseConfigFile {
+  public void parseConfigFile();
+}
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -1,5 +1,6 @@
 package azkaban.user;
 
+import com.google.common.base.Preconditions;
 import com.sun.nio.file.SensitivityWatchEventModifier;
 import java.io.File;
 import java.io.IOException;
@@ -10,6 +11,7 @@ import java.nio.file.StandardWatchEventKinds;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
+import java.util.Map;
 import org.apache.log4j.Logger;
 
 public final class UserUtils {
@@ -40,58 +42,62 @@ public final class UserUtils {
    * Creates a watch thread which listens to specified files' modification and reloads
    * configurations
    */
-  static void setupWatch(final String fileName, ParseConfigFile parser) {
-    if (fileName == null || parser == null) {
-      throw new IllegalArgumentException("filename or parser is NULL.");
-    }
-
-    final File file = new File(fileName);
-    if (!file.exists()) {
-      throw new IllegalArgumentException("User provided file " + fileName + " does not exist.");
-    }
+  static void setupWatch(final Map<String, ParseConfigFile> configFileMap) {
+    Preconditions.checkArgument(configFileMap != null);
 
     Runnable runnable = () -> {
       WatchService watchService;
       Path path;
-      try {
-        watchService = FileSystems.getDefault().newWatchService();
-        path = Paths.get(fileName).getParent();
-        path.register(watchService, new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
-            SensitivityWatchEventModifier.HIGH);
-      } catch (IOException e) {
-        // Ignore the IOException
-        log.warn("IOException while setting up watch on conf"
-            + e.getMessage());
-        return;
-      }
-      for (;;) {
-        // Watch for modifications
-        WatchKey watchKey;
-        try {
-          watchKey = watchService.take();
-        } catch (InterruptedException ie) {
-          log.warn(ie.getMessage());
-          return;
+      for (Map.Entry<String, ParseConfigFile> entry : configFileMap.entrySet()) {
+        String fileName = entry.getKey();
+        ParseConfigFile parser = entry.getValue();
+        Preconditions.checkArgument(fileName != null && parser != null);
+
+        final File file = new File(fileName);
+        if (!file.exists()) {
+          throw new IllegalArgumentException("User provided file " + fileName + " does not exist.");
         }
 
-        for (WatchEvent<?> event : watchKey.pollEvents()) {
-          // Make sure the modification happened to user config file
-          @SuppressWarnings("unchecked")
-          final Path name = ((WatchEvent<Path>)event).context();
-          final Path child = path.resolve(name);
-          if (!child.toString().equals(fileName)) {
-            continue; // not user xml
-          }
-          // reparse the config file
-          log.info("Modification detected, reloading user config");
-          parser.parseConfigFile();
+        try {
+          watchService = FileSystems.getDefault().newWatchService();
+          path = Paths.get(fileName).getParent();
+          path.register(watchService, new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
+              SensitivityWatchEventModifier.HIGH);
+        } catch (IOException e) {
+          // Ignore the IOException
+          log.warn("IOException while setting up watch on conf"
+              + e.getMessage());
+          return;
         }
-        watchKey.reset();
+        for (; ; ) {
+          // Watch for modifications
+          WatchKey watchKey;
+          try {
+            watchKey = watchService.take();
+          } catch (InterruptedException ie) {
+            log.warn(ie.getMessage());
+            return;
+          }
+
+          for (WatchEvent<?> event : watchKey.pollEvents()) {
+            // Make sure the modification happened to user config file
+            @SuppressWarnings("unchecked")
+            final Path name = ((WatchEvent<Path>) event).context();
+            final Path child = path.resolve(name);
+            if (!child.toString().equals(fileName)) {
+              continue; // not user xml
+            }
+            // reparse the config file
+            log.info("Modification detected, reloading user config");
+            parser.parseConfigFile();
+          }
+          watchKey.reset();
+        }
       }
     };
 
     final Thread thread = new Thread(runnable);
-    log.info("Starting configuration watching thread on config file " + fileName);
+    log.info("Starting configuration watching thread.");
     thread.start();
   }
 }
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -13,6 +13,7 @@ import org.apache.log4j.Logger;
 
 public final class UserUtils {
 
+  private static final Logger log = Logger.getLogger(UserUtils.class);
   private UserUtils() {
 
   }
@@ -38,15 +39,22 @@ public final class UserUtils {
    * Creates a watch thread which listens to specified files' modification and reloads
    * configurations
    */
-  public static void setupWatch(final String fileName, Logger log, ParseConfigFile parser) {
+  static void setupWatch(final String fileName, ParseConfigFile parser) {
+    if (fileName == null || parser == null) {
+      throw new IllegalArgumentException("filename or parser is NULL.");
+    }
+
+    final File file = new File(fileName);
+    if (!file.exists()) {
+      throw new IllegalArgumentException("User provided file " + fileName + " does not exist.");
+    }
+
     Runnable runnable = () -> {
       WatchService watchService;
       Path path;
       try {
         watchService = FileSystems.getDefault().newWatchService();
-        final File file = new File(fileName);
-        final String dirPath = file.getParent();
-        path = Paths.get(dirPath);
+        path = Paths.get(fileName).getParent();
         path.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY,
             StandardWatchEventKinds.ENTRY_CREATE);
       } catch (IOException e) {
@@ -66,14 +74,14 @@ public final class UserUtils {
         }
 
         for (WatchEvent<?> event : watchKey.pollEvents()) {
-          // Make sure the modification happened to user xml
+          // Make sure the modification happened to user config file
           @SuppressWarnings("unchecked")
           final Path name = ((WatchEvent<Path>)event).context();
           final Path child = path.resolve(name);
           if (!child.toString().equals(fileName)) {
             continue; // not user xml
           }
-          // reparse the XML
+          // reparse the config file
           log.info("Modification detected, reloading user config");
           parser.parseConfigFile();
         }
@@ -82,7 +90,7 @@ public final class UserUtils {
     };
 
     final Thread thread = new Thread(runnable);
-    System.out.println("Starting thread");
+    log.info("Starting configuration watching thread on config file " + fileName);
     thread.start();
   }
 }
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -85,7 +85,6 @@ public final class UserUtils {
           // Ignore the IOException
           log.warn("IOException while setting up watch on conf " + fileName + ". "
               + e.getMessage());
-          Thread.currentThread().interrupt();
           return;
         }
       }
@@ -93,7 +92,6 @@ public final class UserUtils {
       // Return if WatchService is not initialized
       if (watchService == null) {
         log.warn("Watchservice not setup for any config file(s).");
-        Thread.currentThread().interrupt();
         return;
       }
 
@@ -114,17 +112,13 @@ public final class UserUtils {
           // Make sure the modification happened to user config file
           @SuppressWarnings("unchecked")
           final Path name = ((WatchEvent<Path>) event).context();
-          final String child = dir.resolve(name).toString();
+          final String filename = dir.resolve(name).toString();
           // Lookup the file in dirToFilesMap
-          for (String fileName : dirToFilesMap.get(dir)) {
-            if (!child.equals(fileName)) {
-              continue;
-            }
-
+          if (dirToFilesMap.containsEntry(dir, filename)) {
             // Match!
             // reparse the config file
-            log.info("Modification detected, reloading user config");
-            configFileMap.get(fileName).parseConfigFile();
+            log.info("Modification detected, reloading config file " + filename);
+            configFileMap.get(filename).parseConfigFile();
             break;
           }
         }
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -15,11 +15,12 @@ import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
 import java.util.HashMap;
 import java.util.Map;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class UserUtils {
 
-  private static final Logger log = Logger.getLogger(UserUtils.class);
+  private static final Logger log = LoggerFactory.getLogger(UserUtils.class);
   private UserUtils() {
 
   }
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -1,6 +1,8 @@
 package azkaban.user;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.sun.nio.file.SensitivityWatchEventModifier;
 import java.io.File;
 import java.io.IOException;
@@ -11,6 +13,7 @@ import java.nio.file.StandardWatchEventKinds;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
+import java.util.HashMap;
 import java.util.Map;
 import org.apache.log4j.Logger;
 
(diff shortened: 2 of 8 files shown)


azkaban-common/src/main/java/azkaban/user/XmlUserManager.java
```diff
@@ -73,6 +80,54 @@ public XmlUserManager(final Props props) {
...
+      try {
+        watchService = FileSystems.getDefault().newWatchService();
+        final File file = new File(this.xmlPath);
+        final String dirPath = file.getParent();
+        path = Paths.get(dirPath);
+        path.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY,
+            StandardWatchEventKinds.ENTRY_CREATE);
+      } catch (IOException e) {
+        // Ignore the IOException
+        logger.warn("IOException while setting up watch on conf"
```
HappyRay: Is there any reason not to alert administrators right away in a way that is harder to miss?

djaiswal83: This exception occurs when a new watch service cannot be created or the directory could not be registered with the watch service. If such a thing occurs, the thread would log the warning and return silently. The idea is to keep the system up and running even if this new logic fails to start and avoid any regression.

HappyRay:
Thanks for working on this.
Could you elaborate on how you intend to extend it to reload any configs?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
 
     return false;
   }
+
+  /**
+   * Creates a watch thread which listens to specified files' modification and reloads
+   * configurations
+   */
+  public static void setupWatch(final String fileName, Logger log, ParseConfigFile parser) {
```
burgerkingeater: should we expose log as an argument given this method will be exposed publicly? Consider cases when caller is using slf4j as its logging framework.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+    Runnable runnable = () -> {
+      WatchService watchService;
+      Path path;
+      try {
+        watchService = FileSystems.getDefault().newWatchService();
+        final File file = new File(fileName);
+        final String dirPath = file.getParent();
```
burgerkingeater: oneliner: Paths.get(fileName).getParent()

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+        WatchKey watchKey;
+        try {
+          watchKey = watchService.take();
+        } catch (InterruptedException ie) {
+          log.warn(ie.getMessage());
+          return;
+        }
+
+        for (WatchEvent<?> event : watchKey.pollEvents()) {
+          // Make sure the modification happened to user xml
```
burgerkingeater: the file doesn't necessarily to be user xml, so in the comments should we just call it "config file"?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+          @SuppressWarnings("unchecked")
+          final Path name = ((WatchEvent<Path>)event).context();
+          final Path child = path.resolve(name);
+          if (!child.toString().equals(fileName)) {
+            continue; // not user xml
+          }
+          // reparse the XML
+          log.info("Modification detected, reloading user config");
+          parser.parseConfigFile();
```
burgerkingeater: should we notify admin in case of parsing failure or file missing/deleted?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
burgerkingeater: consider changing String to Path to stronger-type?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
 
     return false;
   }
+
+  /**
+   * Creates a watch thread which listens to specified files' modification and reloads
+   * configurations
+   */
+  public static void setupWatch(final String fileName, Logger log, ParseConfigFile parser) {
+    Runnable runnable = () -> {
```
burgerkingeater: since it's a public API might be used by other UserManager, should we have stricter argument validation? Like file existence check and null pointer check.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+        for (WatchEvent<?> event : watchKey.pollEvents()) {
+          // Make sure the modification happened to user xml
+          @SuppressWarnings("unchecked")
+          final Path name = ((WatchEvent<Path>)event).context();
+          final Path child = path.resolve(name);
+          if (!child.toString().equals(fileName)) {
+            continue; // not user xml
+          }
+          // reparse the XML
+          log.info("Modification detected, reloading user config");
```
burgerkingeater: "reloading config" + file path? to be applicable to other types of config.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+          parser.parseConfigFile();
+        }
+        watchKey.reset();
+      }
+    };
+
+    final Thread thread = new Thread(runnable);
+    System.out.println("Starting thread");
```
burgerkingeater: "starting configuration watching thread" and log the config file name to be more debuggable?

azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
```diff
@@ -0,0 +1,8 @@
+package azkaban.user;
+
+/**
+ * Lambda interface for parsing user config file.
+ */
+public interface ParseConfigFile {
+  public void parseConfigFile();
```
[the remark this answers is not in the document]
djaiswal83: Thanks for pointing this out.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
     return false;
   }
+
+  /**
+   * Creates a watch thread which listens to specified files' modification and reloads
+   * configurations
+   */
+  public static void setupWatch(final String fileName, Logger log, ParseConfigFile parser) {
+    Runnable runnable = () -> {
+      WatchService watchService;
```
burgerkingeater: should we close the WatchService in web server's shutdown hook?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: I will add logger in Utils.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: Thanks.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: Thanks for pointing this out. Remnant of last patch.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: The logic for that should be in parseConfigFile. Incase of XmlUserManager, it is already in place.
Anywhere else, it must be handled.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: String is already available. Path will have to be created before it can be passed in.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: That is a good point.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: Thanks for pointing out. This should be a log line instead of print statement as well.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: We may, although when the server goes down, so will the WatchService.

azkaban-common/src/main/java/azkaban/user/XmlUserManager.java
edwinalu: One option would be to send an email to a configurable list of email addresses. The Azkaban email code should be refactored, but that may be out of the scope of this change.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
HappyRay: How did you test this behavior?
Will the watch thread prevent the process from existing?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +33,56 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+          return;
+        }
+
+        for (WatchEvent<?> event : watchKey.pollEvents()) {
+          // Make sure the modification happened to user xml
+          @SuppressWarnings("unchecked")
+          final Path name = ((WatchEvent<Path>)event).context();
+          final Path child = path.resolve(name);
+          if (!child.toString().equals(fileName)) {
+            continue; // not user xml
```
HappyRay: When will this happen?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
HappyRay: What's the intended behavior of parsing a config file? Any side effects? 
Are synchronization required?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +78,59 @@ public void testDoNotExist() throws Exception {
...
+      } else {
+        lines.add(line);
+      }
+    }
+    // Update the file
+    Files.write(Paths.get(path), lines);
+
+    // This should reload the XML file. Check for user8 again
+    try {
+      user8 = manager.getUser("user8", "password8");
```
HappyRay: It's not clear to me what the test is asserting. Could you elaborate?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +78,59 @@ public void testDoNotExist() throws Exception {
...
+    } catch (final UserManagerException e) {
+      System.out.println("USer8 has updated password" + e.getMessage());
+    }
+
+    try {
+      user8 = manager.getUser("user8", "passwordModified");
```
HappyRay: How does the test make sure the watcher thread has processed the event before this point?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +78,59 @@ public void testDoNotExist() throws Exception {
...
+    } catch (final UserManagerException e) {
+      System.out.println("Test failed " + e.getMessage());
+    } finally {
+      // Revert the file back
```
HappyRay: What if the tests are interrupted, canceled etc? Will it cause the inconsistent state to persist?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +78,59 @@ public void testDoNotExist() throws Exception {
...
+        lines.add(line);
+      }
+    }
+    // Update the file
+    Files.write(Paths.get(path), lines);
+
+    // This should reload the XML file. Check for user8 again
+    try {
+      user8 = manager.getUser("user8", "password8");
+    } catch (final UserManagerException e) {
```
edwinalu: Would the test still pass if the exception isn't thrown? Azkaban is using AssertJ(http://joel-costigliola.github.io/assertj/), which supports assertThatThrownBy.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +78,59 @@ public void testDoNotExist() throws Exception {
...
+      System.out.println("USer8 has updated password" + e.getMessage());
+    }
+
+    try {
```
edwinalu: Please assert that the value is as expected.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: I ran solo server manually and it came down without any issue.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: The watch works on the whole directory, however, it is intended for a file in this case. The check skips other files. Moreover, the watch event is triggered even a .swp file is created which needs to be ignored as well.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: WatchService picks all the events. Even if a config file is modified multiple times, it will be reloaded every time.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: The config file originally contains user/password "user8" and "password8". The password is modified in "files.write() call which results in a watch event which reloads the config.
This try catch block fetches a user with the old password but should fail with "does not exist error" which is caught in the catch block.
In the subsequent try catch block, if an exception occurs then the test failed as it uses the updated password to fetch the user.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: This is a good point. I will think of a way to do it.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: If it does not reach "finally" yes.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: Thanks for brining this up. The test uses fail from Assert.java but I will look at this one too.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: Sure.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
HappyRay: I see. Would it be more straightforward to test the value?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
HappyRay: In our production environment, do we expect other files in this directory?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
HappyRay: I am curious about how it worked. This is not a daemon thread, is it?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: Edwina also suggested something similar. Will update the patch.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +34,63 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
 
     return false;
   }
+
+  /**
+   * Creates a watch thread which listens to specified files' modification and reloads
+   * configurations
+   */
+  static void setupWatch(final String fileName, ParseConfigFile parser) {
+    if (fileName == null || parser == null) {
```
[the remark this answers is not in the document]
djaiswal83: Thanks for the tip.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +37,67 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
 
     return false;
   }
+
+  /**
+   * Creates a watch thread which listens to specified files' modification and reloads
+   * configurations
+   */
+  static void setupWatch(final Map<String, ParseConfigFile> configFileMap) {
+    Preconditions.checkArgument(configFileMap != null);
```
edwinalu: There is Preconditions.checkNotNull().

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +37,67 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+      WatchService watchService;
+      Path path;
+      for (Map.Entry<String, ParseConfigFile> entry : configFileMap.entrySet()) {
+        String fileName = entry.getKey();
+        ParseConfigFile parser = entry.getValue();
+        Preconditions.checkArgument(fileName != null && parser != null);
+
+        final File file = new File(fileName);
+        if (!file.exists()) {
+          throw new IllegalArgumentException("User provided file " + fileName + " does not exist.");
```
edwinalu: It would be good to still process the other files, even if one if missing or misconfigured.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +37,67 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+              + e.getMessage());
+          return;
+        }
+        for (; ; ) {
+          // Watch for modifications
+          WatchKey watchKey;
+          try {
+            watchKey = watchService.take();
+          } catch (InterruptedException ie) {
+            log.warn(ie.getMessage());
```
edwinalu: This link should work: https://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java

Please call Thread.currentThread().interrupt(); to set the interrupt flag.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +37,67 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+            // Make sure the modification happened to user config file
+            @SuppressWarnings("unchecked")
+            final Path name = ((WatchEvent<Path>) event).context();
+            final Path child = path.resolve(name);
+            if (!child.toString().equals(fileName)) {
+              continue; // not user xml
+            }
+            // reparse the config file
+            log.info("Modification detected, reloading user config");
+            parser.parseConfigFile();
```
edwinalu: Would the existing configuration be kept if there are any errors while processing the changed file?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +79,85 @@ public void testDoNotExist() throws Exception {
...
+
+    // Get the user8 from existing XML with password == password8
+    User user8 = manager.getUser("user8", "password8");
+
+    // Modify the password for user8
+    // TODO : djaiswal : Find a better way to modify XML
+    List<String> lines = new ArrayList<>();
+    List<String> origLines = new ArrayList<>();
+    Path filePath = Paths.get(path);
+    System.out.println("File modification time = " + Files.getLastModifiedTime(filePath).toString());
```
edwinalu: Should the println be removed?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +79,85 @@ public void testDoNotExist() throws Exception {
...
+      } else {
+        lines.add(line);
+      }
+    }
+
+    // Make sure the file gets reverted back.
+    try {
+      // Update the file
+      Files.write(filePath, lines);
+      System.out.println("File modification time after write = " + Files.getLastModifiedTime(filePath).toString());
```
edwinalu: println here and also in the remainder of the function.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: My bad.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: If there are errors in parsing, the logs would show it. It is upto user to revert back the config file. When user does that, it reloads the config again.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
djaiswal83: It goes to std out, not on the console. It helps in debugging if the test fails.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -22,4 +40,100 @@ public static boolean hasPermissionforAction(final UserManager userManager, fina
...
+
+        // Get the directory for which watch service event triggered.
+        Path dir = keys.get(watchKey);
+        for (WatchEvent<?> event : watchKey.pollEvents()) {
+          // Make sure the modification happened to user config file
+          @SuppressWarnings("unchecked")
+          final Path name = ((WatchEvent<Path>) event).context();
+          final String child = dir.resolve(name).toString();
+          // Lookup the file in dirToFilesMap
+          for (String fileName : dirToFilesMap.get(dir)) {
```
edwinalu: How about using a SetMultimap, and then checking for existence with contains?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
edwinalu: Can this be an assert instead?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -1,7 +1,25 @@
...
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.log4j.Logger;
```
burgerkingeater: sl4fj is preferred. see: https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -73,6 +80,90 @@ public void testDoNotExist() throws Exception {
...
+      // Update the file
+      Files.write(filePath, lines);
+      FileTime lastModifiedTime = Files.getLastModifiedTime(filePath);
+      System.out.println("File modification time after write = " + lastModifiedTime.toString());
+      if (origModifiedTime.equals(lastModifiedTime)) {
+        // File did not update
+        fail("File did not update.");
+      }
+      // Try for a minute polling every 2 seconds if the config reloaded
+      try {
```
burgerkingeater: have you considered https://github.com/awaitility/awaitility/wiki/Usage ?

djaiswal83: Thanks for the tip. I got it working with this library. Patch coming in.

PR #2173, opened 2019-04-01
Add schedule option to skip execution if previous execution failed

toby-allsopp:
In some use cases it is not appropriate to continue to execute a scheduled job if an execution fails. This PR adds a boolean option, defaulting to false, that can be specified when scheduling a flow that causes scheduled executions of the flow to be skipped if the most recent previous execution of the flow did not succeed.

The new option is available in the web UI (on the Schedule panel) and via the AJAX API (both the deprecated `scheduleFlow` endpoint and `scheduleCronFlow`). The name of the option in the AJAX API is `skipIfPreviousExecutionFailed`.


burgerkingeater:
`In some use cases it is not appropriate to continue to execute a scheduled job if an execution fails`

Do you have some specific examples? Thanks

toby-allsopp:
> Do you have some specific examples? Thanks

Sure, first some background:

We make a product that allows users to define and schedule their own jobs to do various tasks related to their data warehouse such as loading data from source systems and transforming it into a form suitable for reporting tools to access. At the moment we ship our own custom scheduler but we are evaluating replacing it with Azkaban for a few reasons that are not that interesting. The pull requests I've been creating (and there will be more) are filling in gaps where Azkaban is missing some functionality that we currently have.

Now, onto an example of a use-case for the skipping-when-failed functionality:

A common pattern for our customers is to load source data into a set of "load tables" that are usually just mirrors of a subset the source data, normally the records that have been added or updated since the last time it was processed. From there various transformations are applied by inserting into other tables based on the contents of the load tables. If one of these transformations fails for some reason (e.g. transient network issues, bugs in the transformation process itself, errors in the source data, etc.) then some manual action might be needed to correct the error and restart the process from the appropriate point.

The problem with starting a new process while the previous one is in a failed state comes about because the old, failed process and the new process use the same database tables as working space and the new execution would truncate the tables, resulting in possibly-unrecoverable data loss. A lot of the time the source systems could be re-queried to recover the missing data but some source systems are more transient than that and the original data might not be available.

Now, it is probably possible to design the processing such that it can be cancelled at any point and restarted from the beginning without an loss of data, but that is up to the design that our customers come up with and is not directly under our control.

I hope this helps :)

toby-allsopp:
@burgerkingeater Any thoughts on this PR?

toby-allsopp:
So, turns out I misunderstood our requirement and this change is not necessary. Apologies if I wasted anyone's time!

PR #2174, opened 2019-04-01
When flow trigger fails to be scheduled/unscheduled, throw exception to terminate the subsequent steps

burgerkingeater:
When scheduling/unscheduling fails(E.g quartz database connection issue), fail fast to avoid inconsistent situation where unscheduling last version of flow trigger fails, but project with new version of flow trigger gets uploaded successfully.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
+++ b/azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
@@ -118,13 +118,13 @@ public class FlowTriggerScheduler {
     }
   }
 
-  public boolean pauseFlowTrigger(final int projectId, final String flowId)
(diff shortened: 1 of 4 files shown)


azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -112,16 +119,12 @@ public void scheduleAll(final Project project, final String submitUser)
   }
 
   public void pauseFlowTrigger(final int projectId, final String flowId) throws SchedulerException {
-    logger.info(String.format("pausing flow trigger for [projectId:%s, flowId:%s]", projectId,
-        flowId));
```
ypadron-in: Don't we need this anymore?

burgerkingeater: logging is moved out to function caller.

PR #2175, opened 2019-04-02
Add web server host to event metadata

abti:
Add web server host to event metadata, leave the existing azkabanHost metadata to represent executor for backward compatibility.


burgerkingeater:
why is web server host name needed to be included in event metadata?

abti:
> why is web server host name needed to be included in event metadata?

It is needed by OI for associating Azkaban cluster with flows, since they will consume from all clusters. 
Executor hostname alone makes it harder for them (especially if we go to say execs on Yarn or K8s, it will be quite a bit of mapping to maintain)

abti:
Thanks for the review.

Issue #1757, opened 2018-05-11
email with ssl send fail


magicsilence:
How to use the email with ssl?  the version is "azkaban-3.43.0".

I check the source code  and found the  protocol is **smtp** , not **smtps**.

`
//file: azkaban-common / EmailMessage.java

public class EmailMessage {
    private static final String protocol = "smtp";
    
    //....
    
    public void sendEmail() throws MessagingException {
        //....
        final SMTPTransport t = (SMTPTransport) session.getTransport(protocol);
        //....
  }

    //...
}
`

Is there have a better way to send email with ssl ?

juhoautio:
Have you tried setting `mail.tls=true` in Azkaban properties? AFAIK the javax mail library always uses protocol "smtp" but it can be encrypted (see also: https://stackoverflow.com/a/18333887/1068385).

PR #2176, opened 2019-04-03
allow user to specify the security protocol of ssl/tls in azkaban.properties manually

abcfyk:
allow user to specify the security protocol of ssl/tls in azkaban.properties manually。
refers  to: [#1757](https://github.com/azkaban/azkaban/issues/1757)


abcfyk:
by setting: "mail.security=ssl"  or  "mail.security=tls" to switch

PR #2178, opened 2019-04-03
allow user to specify the security protocol of ssl/tls in azkaban.properties  

ummmme:
allow user to specify the security protocol of ssl/tls in azkaban.properties manually,
by setting up "mail.security=tls" or "mail.security=ssl"


abcfyk:
refers to: #1757

PR #2180, opened 2019-04-08
getProjectId api no longer returns IDs of deleted/inactive projects.

ypadron-in:
Before, users couldn't know for sure if a certain project existed or not.
With this changes getProjectId api will either return:
-if project “ABC” exists
```
{
  "project" : "ABC",
  "projectId" : 29
}
```
-if project “ABC” doesn't exist (either never existed or was deleted)
```
{
  "project" : "ABC",
  "error" : "Project ABC doesn't exist."
}
```
Follow up work: Verify permissions to access project id information.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -106,8 +106,8 @@ public class JdbcProjectImpl implements ProjectLoader {
         }
       });
     } catch (final SQLException ex) {
-      logger.error(ProjectResultHandler.SELECT_PROJECT_BY_ID + " failed.", ex);
-      throw new ProjectManagerException("Error retrieving all projects", ex);
+      logger.error(ProjectResultHandler.SELECT_ALL_ACTIVE_PROJECTS + " failed.", ex);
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -155,17 +155,12 @@ public Project fetchProjectByName(final String name) throws ProjectManagerExcept
...
-      List<Project> projects = this.dbOperator
+      final List<Project> projects = this.dbOperator
           .query(ProjectResultHandler.SELECT_ACTIVE_PROJECT_BY_NAME, handler, name);
       if (projects.isEmpty()) {
-        projects = this.dbOperator
-            .query(ProjectResultHandler.SELECT_PROJECT_BY_NAME, handler, name);
-        if (projects.isEmpty()) {
-          throw new ProjectManagerException("No project with name " + name + " exists in db.");
-        }
+        throw new ProjectManagerException("No active project with name " + name + " exists in db.");
```
burgerkingeater: should we just return null instead of throwing exception?

ypadron-in: changed, thanks.

PR #2181, opened 2019-04-08
[REFACTOR] Dedupe common codes

kxu1026:
There are no logic changes besides 2 tiny bugfixes, which I will list as below,

1. Create AbstractHadoopJavaProcessJob class to share common codes which are used by
        HadoopHiveJob, HadoopJavaJob, HadoopPigJob & HadoopSparkJob.
2. Create AbstractMBeanRegistrableServer class to share common codes which are used by AzkabanExecServer, AzkabanWebServer.
3. Clean some non-referenced import files to improve the code quality
4. Remove duplicate code if it can be replaced by FileIOUtils.getSourcePathFromClass() function
5. Introduce PropsUtils.loadPluginProps() Utility function for removing duplicate code in the following classes (AlerterHolder, AzkabanWebServer, PluginCheckerAndActionsLoader, TriggerPluginLoader)
6. Add private constructor for PropsUtils class to match the singleton pattern for Utility Class.
7. Create new PluginUtils class includes getUrls, getURLClassLoader, getPluginClass Utils which are used by AzkabanWebServer, PluginCheckerAndActionsLoader, TriggerPluginLoader.
8. Add default parameter for Utils.createPeriodString() function, which make it be reusable in Schedule.class, BasicTimeChecker.class, ScheduleServlet.calss
9. Introduce Props.getMapByPrefix(final String prefix, boolean ignoreCase) function to simplify the AbstractProcessJob.getEnvironmentVariables() function
10. Introduce SlaOption.convertToObjects static function which can be shared by ExecuteFlowAction.class and Schedule.class
11. Move check and delete tokenFile logic into the HadoopJobUtils.cancelHadoopTokens Util since it is duplicated in all callers
12. Create Utils.runProcess Utils to reduce duplicate code in ServerStatisticsServlet class.
13. Create FilterProjectByPermission function in LoginAbstractAzkabanServlet for making it be reusable in all extended classes.

BUG:
1. Remove Logger log in AbstractProcessJob since log object has been existed in the parent AbstractJob class. 
2 In both PluginCheckerAndActionsLoader.load() and TriggerPluginLoader.load() function., error log was fired regardless if(pluginClass == null) is true/false. Fix this bug. log.error when (plugClass == null) and skip the following unnecessary logic. log.info when (plugClass != null) and run the following expected logic

Will Introduce more test cases to cover the logic soon.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/PluginUtils.java
+++ b/az-core/src/main/java/azkaban/utils/PluginUtils.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2019 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.utils;
+
+import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.log4j.Logger;
+
+
+public class PluginUtils {
+  private static final Logger logger = Logger.getLogger(PluginUtils.class);
+
+  /**
+   * Private constructor.
+   */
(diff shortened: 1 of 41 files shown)


burgerkingeater:
@kxu1026  thanks for taking the time refactoring our code base! 
1. Please resolve the conflicts 
2. Please follow contributor guide https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md, in particular 1. use save plugin to auto format the code to follow azkaban coding style convention. 2. please use slf4j instead of log4j as our logging framework.

kxu1026:
@burgerkingeater 
1. Resolved
2. Followed (Since the existing code is using log4j, I will change log4j to slf4j in a seperate review)

kxu1026:
create a new one. close this.

PR #2184, opened 2019-04-11
Auto Reload Config : Fix race condition to make sure WatchService is started before its watcher thread starts.

djaiswal83:
AUto Reload Config : Fix race condition to make sure WatchService is started before its watcher thread starts.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -46,7 +46,7 @@ public final class UserUtils {
    * Creates a watch thread which listens to specified files' modification and reloads
    * configurations
    */
-  static void setupWatch(final Map<String, ParseConfigFile> configFileMap) {
+  static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws IOException {
     Preconditions.checkNotNull(configFileMap);
     Preconditions.checkArgument(configFileMap.size() > 0);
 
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -50,52 +50,63 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) {
...
       }
+    }
 
-      // Return if WatchService is not initialized
-      if (watchService == null) {
-        log.warn("Watchservice not setup for any config file(s).");
-        return;
+    // Return if WatchService is not initialized
+    if (keys.size() == 0) {
+      log.warn("Watchservice wad not setup for any config file(s).");
```
burgerkingeater: typo: *was

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -50,52 +50,63 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) {
...
-        final File file = new File(fileName);
-        if (!file.exists()) {
-          log.warn("Failed to reload config, user provided file " + fileName + " does not exist.");
-          continue;
-        }
+    final WatchService watchService;
+    try {
+      watchService = FileSystems.getDefault().newWatchService();
+    } catch (IOException e) {
+      log.warn(" Failed to create WatchService " + e.getMessage());
```
burgerkingeater: should we rethrow the exception to notify caller the watchService is not setup correctly.

djaiswal83: okay.

PR #2188, opened 2019-04-14
fixed SessionfindByIp TestCase

rso1904:
There are Session Time Issue. So, It'll be failed sometime while buliding. 
The issue is generated by short session time. I checked disappearing cache variable of SessionCache variable in SessionfindByIP.
So, I changed sessing time and Thread.sleep time in SessionCacheMiss to pass tests.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/server/session/SessionCacheTest.java
+++ b/azkaban-common/src/test/java/azkaban/server/session/SessionCacheTest.java
@@ -23,7 +23,6 @@ import azkaban.user.User;
 import azkaban.utils.Props;
 import org.junit.Test;
 
-
 public final class SessionCacheTest {
 
   final long shortTTL = 100L;
@@ -37,7 +36,7 @@ public final class SessionCacheTest {
 
   @Test
   public void SessionCacheHit() throws Exception {
-    final SessionCache sessionCache = createSessionCacheWithTTL(shortTTL);
+    final SessionCache sessionCache = createSessionCacheWithTTL(longTTL);
     final Session session = new Session("TEST_SESSION_ID", new User("TEST_USER_HIT"),
         "123.12.12.123");
     sessionCache.addSession(session);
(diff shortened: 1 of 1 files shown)


burgerkingeater:
thanks for your fix, how about testing SessionCacheMiss with a small timeout to make sure expiration of session, and a large timeout for the rest making sure no session will expire.

rso1904:
okay, I changed my source code. My source code test sessioncache with short time and long time

azkaban-common/src/test/java/azkaban/server/session/SessionCacheTest.java
```diff
@@ -32,6 +32,10 @@ SessionCache freshSessionCache() throws Exception {
     return new SessionCache(props);
   }
 
+  SessionCache makeSessionCacheWithLongTime() throws Exception {
```
burgerkingeater: how about making session.time.to.live as an argument? E.g:
`createSessionCacheWithTTL(long ttl)`,
 sessionCacheFindByIP will pass a longer ttl
for rest test cases, passing a shorter ttl like 200 ms?

It's less confusing if ttl is explicitly set.

rso1904: I do refactoring makeSessionCacheWithLongTimeto createSessionCacheWithTTl. Please check it.

burgerkingeater: Thanks, sorry for back & forth,  should SessionCacheHit, SessionCacheCount also use long TTL to avoid being flaky?

rso1904: I think it's more safe than before that use shortTTL. Would I do refactoring from this code to use longTTL above two methods?

rso1904: I changed from shortTTL to longTTL in SessionCacheHit and SessionCacheCount methods.
Please check it. Thanks.

burgerkingeater: thanks!

PR #2189, opened 2019-04-15
Fix dynamic job link url

jamiesjc:
When dynamically generating job link url, we need to fetch application id from the log.
If the log data length is 0, meaning there is no new data, we don't need to search for application id again from the log data. 
This is related to PR #1695


jamiesjc:
Todo: Enable this feature for ExecutionController as well. Add new tests.

PR #2192, opened 2019-04-17
Change data type of project_flows.json from BLOB to LONGBLOB

toby-allsopp:
This allows JSON greater than 64kB to be stored for a flow.


toby-allsopp:
Closing as we will be using PostgreSQL for the back-end database and it doesn't have differently-sized blob types.

PR #2195, opened 2019-04-17
Implement ability to check executor's capacity before polling for new Dispatching Logic

ypadron-in:
Continues the work started in PR #2038 by providing executors the ability to check their capacity before taking new executions. Capacity is defined as: 
-availability of flow threads at the time of polling (don't enqueue executions in the executor)
-free physical memory available (in Gigabtyes)
-cpu usage (in percentage)

Each criteria can be enabled or disabled through properties in azkaban.properties file. Executors will only poll a new execution when all the configured criteria are satisfied.

**Configurable properties**:
`azkaban.polling_criteria.flow_threads_available` - possible values true or false. if true executors can poll a new execution only when there are flow threads available.

`azkaban.polling_criteria.min_free_memory_gb` - any value > 0. At least this amount of free physical memory needs to be available to allow polling
`azkaban.polling_criteria.max_cpu_utilization_pct` - 0 < value < 100. Cpu load needs to be less than this value to allow polling
`azkaban.polling_criteria.cpu_load_period_sec` - Period of time over which cpu usage is calculated when cpu utilization criteria is enabled.

In all cases criteria will be considered satisfied and polling allowed if properties are not configured or are configured with invalid values.

Using the new dispatch logic without enabling the "flow threads available" criteria can potentially cause issues, so it should probably be required. The rest of the parameters can be enabled as needed. 

**Potential issues**:
-Executors will poll new executions every time if no criteria is configured. When the maximum number of flow threads configured is reached executions are going to be queued in the executor server. Then, if a queued execution is killed from the UI, that execution will remain in "KILLING" state until it gets a thread to run. This can go from minutes to hours depending on the execution's place in the queue and how long it will take for currently running executions to finish. Meanwhile users can only start new executions of the same flow by specifying the "Run concurrently" option which is not always possible.

-Not all flows take the same amount of time to execute, if executions are allowed to be enqueued in the executor server, long running executions may prevent following executions from starting and cause them to remain enqueued for long periods of time even when other executor servers have the capacity (available flow threads) to start them immediately causing unnecessary delays and an increased chance of missing SLAs.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
+++ b/azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
@@ -76,8 +76,11 @@ public class OsCpuUtil {
     if (this.collectedCpuStats.isEmpty()) {
       return -1;
     }
-    final CpuStats oldestCpuStats = this.collectedCpuStats.pollLast();
     final CpuStats newestCpuStats = getCpuStats();
+    if (newestCpuStats == null) {
+      return -1;
+    }
+    final CpuStats oldestCpuStats = this.collectedCpuStats.pollLast();
     this.collectedCpuStats.push(newestCpuStats);
 
     return calcCpuLoad(oldestCpuStats, newestCpuStats);
--- a/azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
+++ b/azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
@@ -1,0 +1,179 @@
+package azkaban.utils;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Utility class for getting CPU usage (in percentage)
+ *
+ * CPU information is obtained from /proc/stat, so only Linux systems will support this class
+ *
+ * Assumes frequent calls at regular intervals to {@link #getCpuLoad() getCpuLoad}. The length of
+ * time over which cpu load is calculated can be adjusted with parameter
+ * {@code numCpuStatsToCollect} in the class constructor and how often {@link #getCpuLoad()
+ * getCpuLoad} is called.
+ * Example: if {@link #getCpuLoad() getCpuLoad} is called every second and {@code
+ * numCpuStatsToCollect} is set to 60 then each call to {@link #getCpuLoad() getCpuLoad} returns
+ * the cpu load over the last minute.
+ */
+public class OsCpuUtil {
+
+  private static final Logger logger = LoggerFactory.getLogger(OsCpuUtil.class);
+
+  private static final String CPU_STAT_FILE = "/proc/stat";
+
+  private final Deque<CpuStats> collectedCpuStats;
+
+  @Inject
+  public OsCpuUtil(int numCpuStatsToCollect) {
+    if (numCpuStatsToCollect <= 0) {
+      numCpuStatsToCollect = 1;
+    }
+    this.collectedCpuStats = new ArrayDeque<>(numCpuStatsToCollect);
+
+    final CpuStats cpuStats = getCpuStats();
+    if (cpuStats != null) {
+      for (int i = 0; i < numCpuStatsToCollect; i++) {
+        this.collectedCpuStats.push(cpuStats);
+      }
+    }
+  }
+
+  /**
+   * Collects a new cpu stat data point and calculates cpu load with it and the oldest one
+   * collected which is then deleted.
+   *
+   * @return percentage of CPU usage. -1 if there are no cpu stats.
+   */
+  public double getCpuLoad() {
+    if (this.collectedCpuStats.isEmpty()) {
+      return -1;
+    }
+    final CpuStats oldestCpuStats = this.collectedCpuStats.pollLast();
+    final CpuStats newestCpuStats = getCpuStats();
+    this.collectedCpuStats.push(newestCpuStats);
+
+    return calcCpuLoad(oldestCpuStats, newestCpuStats);
+  }
+
+  private double calcCpuLoad(final CpuStats startCpuStats, final CpuStats endCpuStats) {
+    final long startSysUptime = startCpuStats.getSysUptime();
+    final long startTimeCpuIdle = startCpuStats.getTimeCpuIdle();
+    final long endSysUptime = endCpuStats.getSysUptime();
+    final long endTimeCpuIdle = endCpuStats.getTimeCpuIdle();
+
+    if (endSysUptime == startSysUptime) {
+      logger.error("Failed to calculate cpu load: division by zero");
+      return -1.0;
+    }
+    final double percentageCpuIdle =
+        (100.0 * (endTimeCpuIdle - startTimeCpuIdle)) / (endSysUptime - startSysUptime);
+    return 100.0 - percentageCpuIdle;
+  }
+
+  private CpuStats getCpuStats() {
+    if (!Files.isRegularFile(Paths.get(CPU_STAT_FILE))) {
+      // Mac doesn't use proc pseudo files for example.
+      return null;
+    }
+
+    final String cpuLine = getCpuLineFromStatFile();
+    if (cpuLine == null) {
+      return null;
+    }
+
+    return getCpuStatsFromLine(cpuLine);
+  }
+
+  private String getCpuLineFromStatFile() {
+    BufferedReader br = null;
+    try {
+      br = Files.newBufferedReader(Paths.get(CPU_STAT_FILE), StandardCharsets.UTF_8);
+      String line;
+      while ((line = br.readLine()) != null) {
+        // looking for a line starting with "cpu<space>" which aggregates the values in all of
+        // the other "cpuN" lines.
+        if (line.startsWith("cpu ")) {
+          return line;
+        }
+      }
+    } catch (final IOException e) {
+      final String errMsg = "Failed to read cpu stat file: " + CPU_STAT_FILE;
+      logger.error(errMsg, e);
+    } finally {
+      if (br != null) {
+        try {
+          br.close();
+        } catch (final IOException e) {
+          final String errMsg = "Failed to close cpu stat file: " + CPU_STAT_FILE;
+          logger.error(errMsg, e);
+        }
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Parses cpu usage information from /proc/stat file.
+   * Example of line expected with the meanings of the values below:
+   * cpu  4705  356  584    3699   23    23     0       0     0          0
+   * ---- user nice system idle iowait  irq  softirq steal guest guest_nice
+   *
+   * Method visible within the package for testing purposes.
+   *
+   * @param line the text containing cpu usage statistics
+   * @return CpuStats object. null if there is an error.
+   */
+  CpuStats getCpuStatsFromLine(final String line) {
+    try {
+      final String[] cpuInfo = line.split("\\s+");
+      final long user = Long.parseLong(cpuInfo[1]);
+      final long nice = Long.parseLong(cpuInfo[2]);
+      final long system = Long.parseLong(cpuInfo[3]);
+      final long idle = Long.parseLong(cpuInfo[4]);
+      final long iowait = Long.parseLong(cpuInfo[5]);
+      final long irq = Long.parseLong(cpuInfo[6]);
+      final long softirq = Long.parseLong(cpuInfo[7]);
+      final long steal = Long.parseLong(cpuInfo[8]);
+
+      // guest and guest_nice are counted on user and nice respectively, so don't add them
+      final long totalCpuTime = user + nice + system + idle + iowait + irq + softirq + steal;
+
+      final long idleCpuTime = idle + iowait;
+
+      return new CpuStats(totalCpuTime, idleCpuTime);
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,179 @@
...
+   * Example of line expected with the meanings of the values below:
+   * cpu  4705  356  584    3699   23    23     0       0     0          0
+   * ---- user nice system idle iowait  irq  softirq steal guest guest_nice
+   *
+   * Method visible within the package for testing purposes.
+   *
+   * @param line the text containing cpu usage statistics
+   * @return CpuStats object. null if there is an error.
+   */
+  CpuStats getCpuStatsFromLine(final String line) {
```
jamiesjc: Is it a private method?

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,179 @@
...
+  private static final String CPU_STAT_FILE = "/proc/stat";
+
+  private final Deque<CpuStats> collectedCpuStats;
+
+  @Inject
+  public OsCpuUtil(int numCpuStatsToCollect) {
+    if (numCpuStatsToCollect <= 0) {
+      numCpuStatsToCollect = 1;
+    }
+    this.collectedCpuStats = new ArrayDeque<>(numCpuStatsToCollect);
```
jamiesjc: Why do we need a double-ended queue here? Wil a queue be sufficient?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1053,4 +1056,63 @@ public void shutdown() {
...
+      final double maxCpuUtilizationConfig = this.azkabanProps.
+          getDouble(ConfigurationKeys.AZKABAN_POLLING_CRITERIA_MAX_CPU_UTILIZATION_PCT, 100);
+
+      // allow polling if criteria not present or configured with invalid value
+      if (maxCpuUtilizationConfig <= 0 || maxCpuUtilizationConfig >= 100) {
+        return true;
+      }
+
+      final double cpuLoad = this.cpuUtil.getCpuLoad();
+      return (cpuLoad == -1) ? true : cpuLoad < maxCpuUtilizationConfig;
```
jamiesjc: Consider using `Double.compare()` instead of `==` to compare double values?

jamiesjc:
Thanks for the detailed PR description.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: We don’t really need a double-ended queue, we need a FIFO data struct which java.util.Queue doesn't guarantee according to oracle docs "Queues typically, but do not necessarily, order elements in a FIFO (first-in-first-out) manner." ArrayDeque was the most suitable FIFO implementation I could find.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: it has package visibility for testing purposes.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Assumes frequent calls at regular intervals to {@link #getCpuLoad() getCpuLoad}. The length of
+ * time over which cpu load is calculated can be adjusted with parameter
+ * {@code numCpuStatsToCollect} in the class constructor and how often {@link #getCpuLoad()
+ * getCpuLoad} is called.
+ * Example: if {@link #getCpuLoad() getCpuLoad} is called every second and {@code
+ * numCpuStatsToCollect} is set to 60 then each call to {@link #getCpuLoad() getCpuLoad} returns
+ * the cpu load over the last minute.
+ */
+public class OsCpuUtil {
```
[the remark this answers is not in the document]
ypadron-in: /proc/loadavg refers to system load, it measures both CPU and I/O utilization. We only care about CPU usage in this case.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,179 @@
+package azkaban.utils;
```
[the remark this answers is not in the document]
ypadron-in: sure, thanks

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
ypadron-in: Here -1 is an expected exceptional value to indicate a calculation failure. I think it's safe to use == here.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1053,4 +1056,63 @@ public void shutdown() {
       this.scheduler.shutdownNow();
     }
   }
+
+  private class PollingCriteria {
+
+    private final Props azkabanProps;
+    private final SystemMemoryInfo memInfo = SERVICE_PROVIDER.getInstance(SystemMemoryInfo.class);
+    private final OsCpuUtil cpuUtil = SERVICE_PROVIDER.getInstance(OsCpuUtil.class);
```
jamiesjc: Why do you use provider instead of normal injection?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1053,4 +1056,63 @@ public void shutdown() {
...
+    private boolean satisfiesFlowThreadsAvailableCriteria() {
+      final int flowThreadsAvailableConfig = this.azkabanProps.
+          getInt(ConfigurationKeys.AZKABAN_POLLING_CRITERIA_FLOW_THREADS_AVAILABLE, 0);
+
+      // allow polling if not present or configured with invalid value
+      if (flowThreadsAvailableConfig <= 0) {
+        return true;
+      }
+      final int remainingFlowThreads = FlowRunnerManager.this.getMaxNumRunningFlows() -
+          FlowRunnerManager.this.getNumRunningFlows();
```
jamiesjc: Do you think we can add a log message here when `remainingFlowThreads` <= 0 so that we know this check kicks in?

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
```diff
@@ -22,16 +23,30 @@
...
    */
   long getOsTotalFreeMemorySize() {
+    return getAggregatedFreeMemorySize(MEM_KEYS);
+  }
+
+  /**
+   * @return the free physical memory size of the OS. 0 if there is an error or the OS doesn't
+   * support this memory check.
+   */
+  long getOsFreePhysicalMemorySize() {
```
jamiesjc: What is the reason for checking physical memory instead of total memory?

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,195 @@
...
+      final long nice = Long.parseLong(cpuInfo[2]);
+      final long system = Long.parseLong(cpuInfo[3]);
+      final long idle = Long.parseLong(cpuInfo[4]);
+      final long iowait = Long.parseLong(cpuInfo[5]);
+      final long irq = Long.parseLong(cpuInfo[6]);
+      final long softirq = Long.parseLong(cpuInfo[7]);
+      final long steal = Long.parseLong(cpuInfo[8]);
+
+      // guest and guest_nice are counted on user and nice respectively, so don't add them
+      final long totalCpuTime = user + nice + system + idle + iowait + irq + softirq + steal;
```
jamiesjc: Consider adding the link to reference the way of calculating cpu time?

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+    final CpuStats oldestCpuStats = this.collectedCpuStats.pollLast();
+    final CpuStats newestCpuStats = getCpuStats();
+    this.collectedCpuStats.push(newestCpuStats);
+
+    return calcCpuLoad(oldestCpuStats, newestCpuStats);
+  }
+
+  private double calcCpuLoad(final CpuStats startCpuStats, final CpuStats endCpuStats) {
+    final long startSysUptime = startCpuStats.getSysUptime();
```
jamiesjc: Will it throw NPE when `startCputStats` or `endCpuStats` is null?

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+  /**
+   * Collects a new cpu stat data point and calculates cpu load with it and the oldest one
+   * collected which is then deleted.
+   *
+   * @return percentage of CPU usage. -1 if there are no cpu stats.
+   */
+  public double getCpuLoad() {
```
jamiesjc: Is calculating the newest and oldest data points the right way to get CPU load? Let's discuss more about it.

burgerkingeater:
@ypadron-in could you elaborate the motivation behind memory check given we already have memory check on job-level?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1053,4 +1056,63 @@ public void shutdown() {
...
+      }
+      return false;
+    }
+
+    public PollingCriteria(final Props azkabanProps) {
+      this.azkabanProps = azkabanProps;
+    }
+
+    private boolean satisfiesFlowThreadsAvailableCriteria() {
+      final int flowThreadsAvailableConfig = this.azkabanProps.
```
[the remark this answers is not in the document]
ypadron-in: yes true/false is more readable, thanks. I'll make the change.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
ypadron-in: yes, I'll do it.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
ypadron-in: This is a non-static inner class of FlowRunnerManager. The injector cannot (as far as I know) inject this type of classes, it would have to be a static inner class but it uses FlowRunnerManager’s logger, getMaxNumRunningFlows() and getNumRunningFlows() which are non-static members. PollingCriteria behaves similarly to PollingService class, I just followed the latter’s pattern.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: will add it.

ypadron-in:
> @ypadron-in could you elaborate the motivation behind memory check given we already have memory check on job-level?

The reason behind this check is to prevent executors from taking executions they cannot run immediately because there are no resources available.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: The data points are the pairs (system uptime and cpu idle time since last boot). By taking the difference between a data point from now and another from 5 minutes ago, for example, we are calculating what happened during that time in terms of cpu usage, basically we are saying over the last 5 minutes how much of this time was the cpu idle as a percentage, the remaining percentage would then indicate the cpu usage.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
ypadron-in: Currently, for `azkaban.polling_criteria.min_free_memory_gb` I'm using a rough approximation of the memory available without pushing the system into swap(it's actually an under estimation), total memory would include swap which I'm trying to avoid.

Based on this [https://unix.stackexchange.com/questions/499649/is-cached-memory-de-facto-free](https://unix.stackexchange.com/questions/499649/is-cached-memory-de-facto-free) I could improve the estimation of free memory by defaulting to field "MemAvailable" whenever is supported. Let me know what you think.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: NPE shouldn't be thrown. In the constructor we verify whether /proc/stat file is supported or not by the OS where the app is running. If it's supported we push a first data point if not collectedCpuStats remains empty indicating an immediate return(stats are not available) to all future method calls.

azkaban-common/src/main/java/azkaban/utils/OsMemoryUtil.java
jamiesjc: Based on your investigation, it seems to me MemAvailable is a better choice. You can do some experiment to verify and improve it later, thanks!

ypadron-in: sounds good.

azkaban-common/src/main/java/azkaban/utils/OsCpuUtil.java
ypadron-in: Talked offline. I will add a check for null values returned by getCpuStats() in case of some random errors while reading a data point. Thanks!

PR #2196, opened 2019-04-18
Add a third configuration file, azkaban.local.properties, that can override things in azkaban.properties

toby-allsopp:
It is sometimes convenient to make modifications to Azkaban's configuration without modifying azkaban.properties. For example, when upgrading Azkaban it is easy to lose configuration changes because azkaban.properties is overwritten when extracting the distribution zip/tarball.

This PR introduces a new file, azkaban.local.properties, that, if present, is consulted before azkaban.properties.


juhoautio:
> when upgrading Azkaban it is easy to lose configuration changes because azkaban.properties is overwritten when extracting the distribution zip/tarball

This makes me wonder, aren't you checking the diff between your current `azkaban.properties` & the one in the new Azkaban package before overwriting with extracted package? Do you keep _your_ version of `azkaban.properties` in version control?

toby-allsopp:
> This makes me wonder, aren't you checking the diff between your current `azkaban.properties` & the one in the new Azkaban package before overwriting with extracted package? Do you keep _your_ version of `azkaban.properties` in version control?

Hi, @juhoautio. We are intending to bundle Azkaban with a larger suite of software in such a way that our users don't need to know about how to configure Azkaban. This means that we will have a GUI tool to install, configure and upgrade Azkaban. This tool will be responsible for maintaining the `azkaban.properties` file during upgrades and it would be much simpler if we didn't have to deal with merging.

Having said that, it might be sufficient for our purposes to just put the automatically-generated settings at the end of `azkaban.properties` as I think if something is set more than once in a properties file then the last occurrence will take precedence.

PR #2198, opened 2019-04-18
 [DEDUPE] Refactor and Deduplicate the common codes NEW REVISION

kxu1026:
There are no logic changes besides 2 tiny bugfixes, which I will list as below,

    Create AbstractHadoopJavaProcessJob class to share common codes which are used by
    HadoopHiveJob, HadoopJavaJob, HadoopPigJob & HadoopSparkJob.
    Create AbstractMBeanRegistrableServer class to share common codes which are used by AzkabanExecServer, AzkabanWebServer.
    Clean some non-referenced import files to improve the code quality
    Remove duplicate code if it can be replaced by FileIOUtils.getSourcePathFromClass() function
    Introduce PropsUtils.loadPluginProps() Utility function for removing duplicate code in the following classes (AlerterHolder, AzkabanWebServer, PluginCheckerAndActionsLoader, TriggerPluginLoader)
    Add private constructor for PropsUtils class to match the singleton pattern for Utility Class.
    Create new PluginUtils class includes getUrls, getURLClassLoader, getPluginClass Utils which are used by AzkabanWebServer, PluginCheckerAndActionsLoader, TriggerPluginLoader.
    Add default parameter for Utils.createPeriodString() function, which make it be reusable in Schedule.class, BasicTimeChecker.class, ScheduleServlet.calss
    Remove unnecessary ENV. prefix checking in AbstractProcessJob.getEnvironmentVariables() function
    Introduce SlaOption.convertToObjects static function which can be shared by ExecuteFlowAction.class and Schedule.class
    Move check and delete tokenFile logic into the HadoopJobUtils.cancelHadoopTokens Util since it is duplicated in all callers
    Create Utils.runProcess Utils to reduce duplicate code in ServerStatisticsServlet class.
    Create FilterProjectByPermission function in LoginAbstractAzkabanServlet for making it be reusable in all extended classes.
    BUG:

    Remove Logger log in AbstractProcessJob since log object has been existed in the parent AbstractJob class.
    2 In both PluginCheckerAndActionsLoader.load() and TriggerPluginLoader.load() function., error log was fired regardless if(pluginClass == null) is true/false. Fix this bug. log.error when (plugClass == null) and skip the following unnecessary logic. log.info when (plugClass != null) and run the following expected logic
    Will Introduce more test cases to cover the logic soon.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/PluginUtils.java
+++ b/az-core/src/main/java/azkaban/utils/PluginUtils.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2019 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.utils;
+
+import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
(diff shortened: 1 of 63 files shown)


kxu1026:
duplicated one

PR #2199, opened 2019-04-18
Preserve JSON encoding in generated output properties

mactaggart:
Currently, if an output property is JSON, the generated property doesn't preserve the JSON encoding. For example, the following:

`echo '{"param1”:[“a”,”b”]}’ > $JOB_OUTPUT_PROP_FILE`

results in param1 having a value of “[a,b]”, which is not valid JSON; the result of calling toString() on the java List containing the “a” and “b” strings. This prevents subsequent jobs from deserializing JSON.

This pull request preserves JSON output properties without affecting non-JSON string values.


mactaggart:
To add some context to the motivation of this PR, the use case in mind is to allow command type jobs to receive (and parse) JSON output from the previous job. Preserving JSON in job output is a convenient way of passing (possibly complex) state to the next job. So, in the below example, 'myutil' could parse 'param1' as JSON: 

` - name: run_myutil
   type: command
   config:
     command: myutil '${param1}'
`

Thoughts?

PR #2206, opened 2019-04-23
Fix failing autoReload test

djaiswal83:
-Lower the sensitivity of watch service to give it ample time to ensure the file is closed before it is parsed.

This maybe causing -1:-1 Premature end of file error
- Added header for ParseConfigFile Interface
- Some minor typo fixes and indentations.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
+++ b/azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2012 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
 package azkaban.user;
 
 /**
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -89,7 +89,7 @@ public final class UserUtils {
           // There is no entry for this directory, create a watchkey
           WatchKey watchKey = dir.register(watchService,
               new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
-              SensitivityWatchEventModifier.LOW);
+              SensitivityWatchEventModifier.MEDIUM);
           keys.put(watchKey, dir);
         }
         // Add the config file to dir map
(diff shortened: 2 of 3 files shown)


azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -78,12 +78,12 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
...
 
       try {
-        Path dir = Paths.get(fileName).getParent();
+        final Path dir = Paths.get(fileName).getParent();
         if (!dirToFilesMap.containsKey(dir)) {
-          // There is not entry for this directory, create a watchkey
+          // There is no entry for this directory, create a watchkey
           WatchKey watchKey = dir.register(watchService,
               new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
-              SensitivityWatchEventModifier.HIGH);
```
edwinalu: It seems like reducing the sensitivity would make the test failure less likely to occur, but not fix the issue?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -78,16 +82,18 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
...
+              SensitivityWatchEventModifier.LOW);
           keys.put(watchKey, dir);
         }
         // Add the config file to dir map
         dirToFilesMap.put(dir, fileName);
+        fileTimeMap.put(fileName, Files.getLastModifiedTime(filePath));
```
edwinalu: Is this the full path name? It will be safer to use the full path, in case there are some configs that are in different directories that have the same file name.

djaiswal83: Yes, the file name is fully qualified.

edwinalu: How about Path.getCanonicalPath(), or adding a comment saying that the key for the map should be the fully qualified path.

azkaban-common/src/main/java/azkaban/user/ParseConfigFile.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2012 LinkedIn Corp.
```
burgerkingeater: please update the year. it can be setup in intellij template.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -128,10 +135,22 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
           // Lookup the file in dirToFilesMap
           if (dirToFilesMap.containsEntry(dir, filename)) {
             // Match!
+            // Verify if the file has been modified
+            FileTime modifiedTime;
+            try {
+              modifiedTime = Files.getLastModifiedTime(Paths.get(filename));
+            } catch (IOException e) {
```
burgerkingeater: can you make sure autoformat is turned on for this PR? it seems not so...

burgerkingeater:
looks like the build is still failing for the same test case even for this PR :)

djaiswal83:
Recreating new PR to avoid messy commit history

PR #2218, opened 2019-04-25
WatchService : Make sure to wait for 1 second before acting on event to avoid multiple reloads

djaiswal83:
WatchService : Make sure to wait for 1 second before acting on event to avoid multiple reloads

Use style formatter.
Add copyright info to ParseConfigFile interface

### DIFF
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -21,6 +21,7 @@ import org.slf4j.LoggerFactory;
 public final class UserUtils {
 
   private static final Logger log = LoggerFactory.getLogger(UserUtils.class);
+
   private UserUtils() {
 
   }
@@ -53,7 +54,7 @@ public final class UserUtils {
     final WatchService watchService;
     try {
       watchService = FileSystems.getDefault().newWatchService();
-    } catch (IOException e) {
+    } catch (final IOException e) {
       log.warn(" Failed to create WatchService " + e.getMessage());
       throw e;
     }
@@ -64,9 +65,9 @@ public final class UserUtils {
     final Multimap<Path, String> dirToFilesMap = HashMultimap.create();
 
     // Iterate over each file.
-    for (Map.Entry<String, ParseConfigFile> entry : configFileMap.entrySet()) {
-      String fileName = entry.getKey();
-      ParseConfigFile parser = entry.getValue();
+    for (final Map.Entry<String, ParseConfigFile> entry : configFileMap.entrySet()) {
+      final String fileName = entry.getKey();
+      final ParseConfigFile parser = entry.getValue();
       Preconditions.checkNotNull(fileName);
       Preconditions.checkNotNull(parser);
 
@@ -78,17 +79,17 @@ public final class UserUtils {
       }
 
       try {
-        Path dir = Paths.get(fileName).getParent();
+        final Path dir = Paths.get(fileName).getParent();
         if (!dirToFilesMap.containsKey(dir)) {
-          // There is not entry for this directory, create a watchkey
-          WatchKey watchKey = dir.register(watchService,
+          // There is no entry for this directory, create a watchkey
+          final WatchKey watchKey = dir.register(watchService,
               new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
-              SensitivityWatchEventModifier.HIGH);
+              SensitivityWatchEventModifier.MEDIUM);
           keys.put(watchKey, dir);
         }
         // Add the config file to dir map
         dirToFilesMap.put(dir, fileName);
-      } catch (IOException e) {
+      } catch (final IOException e) {
         // Ignore the IOException
         log.warn("IOException while setting up watch on conf " + fileName + ". "
             + e.getMessage());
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -100,38 +101,39 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
...
+    final Runnable runnable = () -> {
       // Watchservice is established, now listen for the events till eternity!
-      for (;; ) {
-        WatchKey watchKey;
+      for (; ; ) {
+        final WatchKey watchKey;
         try {
           watchKey = watchService.take();
-        } catch (InterruptedException ie) {
+          // Wait for a second to ensure there is only one event for a modification.
```
edwinalu: 1 second seems like a long time to wait, although it doesn't matter much in this context (file change).

Can you explain more about how having the wait suppresses an event?

djaiswal83: This question puts some light on it. The author of best answer suggests upto 3 seconds but uses 50 milli seconds in his example code.
https://stackoverflow.com/questions/16777869/java-7-watchservice-ignoring-multiple-occurrences-of-the-same-event

edwinalu: Thanks for the link. It  looks like one of the options is checking the last modified time -- you were considering this as well? Can the sleep be shorter, and can you update the comments to say that the 2 ENTRY_MODIFY events would be consolidated into one ENTRY_MODIFY event would have count of 2.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -100,38 +101,39 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
...
+          // WatchService otherwise creates two events 1 for content and 1 for modification time.
+          Thread.sleep(1000L);
```
juhoautio: Really? Does it matter how "often" `watchService.take()` is called, or are all events returned by `watchKey.pollEvents()` any way?

juhoautio:
> Make sure to wait for 1 second before acting on event to avoid multiple reloads

So is the idea is to just make it less probable that read happens while file is being written, although this doesn't really guarantee the consistency if file is being written to really slowly (once), or multiple times within short intervals (less than 1s between different writes)?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: All events in the queue are returned by take(). Please take a look at the latest patch. Hopefully it clarifies.
https://stackoverflow.com/questions/16777869/java-7-watchservice-ignoring-multiple-occurrences-of-the-same-event

juhoautio: Ok. Sorry I didn't expect the WatchService  API to work like that. As far as I understood, this kind of scenario is still possible:

> user: 0.0 s: start writing, file is empty
watcher thread: 0.01 s: `watchService.take()` returns, go to `Thread.sleep`
user: 0.1 s: write rest of the stuff to file & close it
watcher thread: 1.01 s: `Thread.sleep` has ended, `watchKey.pollEvents()` is called
user: 1.02s: start writing again, file is empty again
watcher thread: 1.03 s: **try to read the file -> fail because it's empty**, as long as exception is caught, **loop starts from beginning** -> go to `watchService.take()` again
user: 1.1 s: write rest of the stuff to file & close it
watcher thread: 1.11 s: `watchService.take()` returns, go to `Thread.sleep`
watcher thread: 2.11 s: Thread.sleep has ended, `watchKey.pollEvents()` is called
watcher thread: 2.13 s: **reading file succeeds this time**

I guess it's still good to try to avoid the failure when possible, but this kind of corner cases still exist. Should be fine I guess, but makes me wonder if it would be possible to lock the file to make it 100% robust :)

djaiswal83: @juhoautio thanks for reviewing the code. As you mentioned, incase the sleep is not enough, the exception would occur and eventually another event will catch the write, thus guaranteeing that reload occurs. Did you get chance to look at the latest patch? I took some cues from your work.

djaiswal83: Regarding locking, we don't do it now. Maybe we can create an issue to track it separately?

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -100,38 +101,52 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
...
+          }
+          // Match!
+          // Reparse the config file
+          log.info("Modification detected, reloading config file " + filename + ".");
+          try {
             configFileMap.get(filename).parseConfigFile();
-            break;
+          } catch (final Exception e) {
+            // If there is any exception while parsing the config file, log it and move on
+            log.warn("Reload failed for config file " + filename + " due to " + e.getMessage());
```
juhoautio: I suppose just `+ e` ie. toString() may be usually nicer than `e.getMessage()` because message might be null.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -159,12 +162,78 @@ public void testAutoReload() throws Exception {
         System.out.println("Test failed " + e.getMessage());
         fail("Test failed " + e.getMessage());
       }
-    } catch (ConditionTimeoutException te) {
-      System.out.println("The config did not reload in 60 seconds");
-      fail("The config did not reload in 60 seconds");
+    } catch (final ConditionTimeoutException te) {
+      System.out.println("The config did not reload in 30 seconds");
```
juhoautio: println seems redundant, `fail` message below should be enoguh?

djaiswal83: I added them for debugger as fail message does not show up on it's console. I will remove them.
Thanks

azkaban-common/src/main/java/azkaban/user/UserUtils.java
djaiswal83: Good idea. Updating it everywhere in the patch.

juhoautio:
Seems like a working fix for the failures 👍

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -108,13 +114,10 @@ public void testAutoReload() throws Exception {
 
     // Modify the password for user8
     // TODO : djaiswal : Find a better way to modify XML
-    List<String> lines = new ArrayList<>();
-    List<String> origLines = new ArrayList<>();
-    Path filePath = Paths.get(path);
-    FileTime origModifiedTime = Files.getLastModifiedTime(filePath);
+    final FileTime origModifiedTime = Files.getLastModifiedTime(filePath);
     System.out.println("File modification time = " + origModifiedTime.toString());
```
edwinalu: There's a lot of use of System.out.println -- please remove or convert to log.info/debug

djaiswal83: Thanks. I just updated them to use log.

azkaban-common/src/main/java/azkaban/user/UserUtils.java
```diff
@@ -53,8 +54,8 @@ static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws
     final WatchService watchService;
     try {
       watchService = FileSystems.getDefault().newWatchService();
-    } catch (IOException e) {
-      log.warn(" Failed to create WatchService " + e.getMessage());
+    } catch (final IOException e) {
+      log.warn(" Failed to create WatchService " + e.toString());
```
edwinalu: log.warn can also take Exception as an argument. Can you please change to pass in the Exception instead of concatenating.

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -150,21 +156,82 @@ public void testAutoReload() throws Exception {
...
+      // Update the file to make it empty.
+      Files.write(filePath, new byte[0], StandardOpenOption.TRUNCATE_EXISTING);
+      final FileTime lastModifiedTime = Files.getLastModifiedTime(filePath);
+      log.info("File modification time after write = " + lastModifiedTime.toString());
+      if (origModifiedTime.equals(lastModifiedTime)) {
+        // File did not update
+        fail("File did not update.");
+      }
+      // Try for 30 seconds polling every 10 seconds if the config is reloaded
+      Awaitility.await().atMost(30L, TimeUnit.SECONDS).
```
juhoautio: Whoa, so this is adding a fixed +30s to the test suite runtime? Maybe this should be disabled or then optimize to run without any actual sleeping.

edwinalu: Awaitility can check for a condition -- @djaiswal83 , can you add this, so that it will continue as soon as the condition is satisfied.

PR #2225, opened 2019-04-30
[CLEANUP] Replace all log4j to slf4j

kxu1026:
Review all code base and change all log4j dependencies to slf4j when it is possible. However, the core Job Management part is heavily depending on the log4j. If we want to totally solve the problem, we should review and discuss that offline first. It may introduce back-ward incompatible problem.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/PropsUtils.java
+++ b/az-core/src/main/java/azkaban/utils/PropsUtils.java
@@ -15,7 +15,6 @@
  */
 package azkaban.utils;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.MapDifference;
 import com.google.common.collect.Maps;
 import java.io.File;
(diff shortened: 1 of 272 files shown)


az-core/src/main/java/azkaban/utils/PropsUtils.java
```diff
@@ -33,15 +33,15 @@
...
+import org.slf4j.LoggerFactory;
 
 
 /**
  * Utility Functions related to Prop Operations
  */
 public class PropsUtils {
 
-  private static final Logger logger = Logger.getLogger(PropsUtils.class);
+  private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(PropsUtils.class);
```
edwinalu: import org.slf4j.Logger?

kaixu1026: addressed

PR #2226, opened 2019-04-30
Delete schedule of renamed/deleted flows on project upload.

ypadron-in:
When a new project version is uploaded existing schedules of renamed/deleted flows must be removed.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -1771,7 +1771,8 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       type = "zip";
     } else {
       item.delete();
-      registerError(ret, "File type " + contentType + " unrecognized.", resp, 400);
+      registerError(ret, "File type " + contentType + " unrecognized.", resp,
+          HttpServletResponse.SC_BAD_REQUEST);
       return;
     }
 
@@ -1830,7 +1831,8 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       if (error.length() > 512) {
         error = error.substring(0, 512) + "<br>Too many errors to display.<br>";
       }
-      registerError(ret, "Installation Failed.<br>" + error, resp, 500);
+      registerError(ret, "Installation Failed.<br>" + error, resp,
+          HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
     } finally {
       if (out != null) {
         out.close();
@@ -1851,14 +1853,14 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
   private Project validateUploadAndGetProject(final HttpServletResponse resp,
       final Map<String, String> ret, final User user, final String projectName) {
     if (projectName == null || projectName.isEmpty()) {
-      registerError(ret, "No project name found.", resp, 400);
+      registerError(ret, "No project name found.", resp, HttpServletResponse.SC_BAD_REQUEST);
       return null;
     }
     final Project project = this.projectManager.getProject(projectName);
     if (project == null || !project.isActive()) {
       final String failureCause = (project == null) ? "doesn't exist." : "was already removed.";
       registerError(ret, "Installation Failed. Project '" + projectName + " "
-          + failureCause, resp, 410);
+          + failureCause, resp, HttpServletResponse.SC_GONE);
       return null;
     }
 
@@ -1871,13 +1873,13 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
           "Project uploading is locked out. Only admin users and users with special permissions can upload projects. "
               + "User " + user.getUserId() + " doesn't have permission to upload project.";
       logger.info(message);
-      registerError(ret, message, resp, 403);
+      registerError(ret, message, resp, HttpServletResponse.SC_FORBIDDEN);
       return null;
     }
     if (!hasPermission(project, user, Type.WRITE)) {
       registerError(ret,
           "Installation Failed. User '" + user.getUserId() + "' does not have write access.",
-          resp, 400);
+          resp, HttpServletResponse.SC_BAD_REQUEST);
       return null;
     }
     return project;
@@ -1943,7 +1945,7 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       // get discarded.
       registerError(ret,
           errorMsgs.length() > 4000 ? errorMsgs.substring(0, 4000) : errorMsgs.toString(), resp,
-          500);
+          HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
     }
     if (warnMsgs.length() > 0) {
       ret.put("warn", warnMsgs.length() > 4000 ? warnMsgs.substring(0, 4000) : warnMsgs.toString());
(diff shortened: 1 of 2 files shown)


djaiswal83:
Thanks for fixing this. Is it possible to add a test for it?

djaiswal83:
There is conflict in the code. Please resolve.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1728,6 +1729,28 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
         }
 
+        // remove schedule of renamed/deleted flows
+        final Set<String> flowNameList = new HashSet<>();
+        for (final Flow flow : project.getFlows()) {
+          flowNameList.add(flow.getId());
+        }
+
+        try {
+          for (final Schedule schedule : this.scheduleManager.getSchedules()) {
```
jamiesjc: In some clusters, we have thousands of schedules. Will this check impact the uploading time for users?

jamiesjc:
One minor suggestion: the `ajaxHandleUpload` method is already huge. Would it be nice to move some functionalities out into smaller pieces?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
djaiswal83: Maybe we can measure it and test it out on a dev cluster? If this comes out to be a real problem, then we can maintain a set of scheduleIds in Project class. Then we only look up schedules for the given project.
ScheduleManager already maintain a map of scheduleID to Schedule which can be used to remove the schedule.

ypadron-in: I don't think the check will add considerable latency. I tested locally with a List of 1 million objects containing an Integer and the time it took to iterate and compare values is under 15 milliseconds.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1774,144 +1784,177 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+  static void removeScheduleOfDeletedFlows(final Project project,
+      final ScheduleManager scheduleManager, final Consumer<Schedule> onDeletedSchedule)
+      throws ScheduleManagerException {
+    final Set<String> flowNameList = project.getFlows().stream().map(f -> f.getId()).collect(
+        Collectors.toSet());
 
-        if (this.enableQuartz) {
-          this.scheduler.schedule(project, user.getUserId());
-        }
+    for (final Schedule schedule : scheduleManager.getSchedules()) {
```
edwinalu: Would it be possible to get schedules by project, to limit the number of flows fetched?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1774,144 +1784,177 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+            schedule.getScheduleId(), schedule.getFlowName(), schedule.getProjectName());
+        this.projectManager.postProjectEvent(project, EventType.SCHEDULE, "azkaban",
+            "Schedule " + schedule.toString() + " has been removed.");
+      });
+
+      registerErrorsAndWarningsFromValidationReport(resp, ret, reports);
+    } catch (final Exception e) {
+      logger.info("Installation Failed.", e);
+      String error = e.getMessage();
+      if (error.length() > 512) {
```
edwinalu: Please add some constants for 512 and 500.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1774,144 +1784,177 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
-
-      logger.info("Upload: project " + projectName + " version is " + project.getVersion()
-          + ", reference is " + System.identityHashCode(project));
-      ret.put("version", String.valueOf(project.getVersion()));
+    }
+    if (errorMsgs.length() > 0) {
+      // If putting more than 4000 characters in the cookie, the entire message will somehow
+      // get discarded.
+      registerError(ret,
+          errorMsgs.length() > 4000 ? errorMsgs.substring(0, 4000) : errorMsgs.toString(), resp,
```
edwinalu: Is there a reason why this is 4000 and not 512? Can this be consistent?

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ProjectManagerServletTest.java
```diff
@@ -0,0 +1,60 @@
+package azkaban.webapp.servlet;
```
edwinalu: Please add the copyright info.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ypadron-in: It looks like there is no direct relationship between Project and Schedule. The list of schedules is coming from memory and Im filtering out by project first to reduce work. Creating a getSchedulesByProject() method in ScheduleManager class will do essentially the same thing I'm doing.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ProjectManagerServletTest.java
ypadron-in: done, thanks.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ypadron-in: Using constants for HTTP status code now, thanks.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ypadron-in: I'm only moving the code around to reduce the size of ajaxHandleUpload(). I don't know what these numbers: 512 and 4000 represent.

djaiswal83: I think this is a good chance to understand why the numbers are different. Can you please find out?

ypadron-in: 512 limits the size of an unexpected error message while processing the request and seems a pretty arbitrary choice.
4000 limits the size of project validation messages, apparently it has something to do with browser cookies, but if that's the case the right place to do the trimming is when it's actually stored in a cookie, not here.
This is the only place where these numbers are used and we wouldn't be gaining much by making this change. Given our current priorities I would suggest to leave it as it is and revisit this when we have a better motivation.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1774,144 +1785,178 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+  private Project validateUploadAndGetProject(final HttpServletResponse resp,
+      final Map<String, String> ret, final User user, final String projectName) {
+    if (projectName == null || projectName.isEmpty()) {
+      registerError(ret, "No project name found.", resp, HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    final Project project = this.projectManager.getProject(projectName);
+    if (project == null || !project.isActive()) {
+      final String failureCause = (project == null) ? "doesn't exist." : "was already removed.";
+      registerError(ret, "Installation Failed. Project '" + projectName + " "
```
li-ygerchikov, line 1862: Missing the closing single-quota. Not worth its own PR, but if we are ever changing this, or anything close to it...

PR #2227, opened 2019-04-30
Enable dynamic job link for ExecutionController.

jamiesjc:
This is related to PR #1695
1. Added the same logic to dynamically generating job link URL in ExecutionController.
2. Consolidate the code in ExecutorManger and ExecutionController.
3. Added unit tests to verify the functionality.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -16,9 +16,21 @@
 
 package azkaban.executor;
 
+import static java.util.Objects.requireNonNull;
+
+import azkaban.Constants.ConfigurationKeys;
 import azkaban.alert.Alerter;
+import azkaban.utils.AuthenticationUtils;
+import azkaban.utils.Props;
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -242,4 +264,90 @@ public static boolean isFinished(final ExecutableFlow flow) {
...
+    return jobLinkUrl;
+  }
+
+  /**
+   * Find the application id from job log by matching "application_<id>" pattern.
+   *
+   * @param logData the log data
+   * @return the application id
+   */
+  public static String findApplicationIdFromLog(final String logData) {
```
ypadron-in: In HadoopJobUtils class there is a findApplicationIdFromLog() method that returns a collection of Strings. Why is not the same case here?

jamiesjc: Some historical reason. At the time when this feature was first created (1 year ago), Hadoop plugin code was in a separate repo. So they are developed separately. 
One difference between the two methods, one is getting app_id from log file path, the other is getting from log data directly. Another difference is that one is fetching all the app_ids, the other is fetching only one. For now we are only providing one button, so fetching one would be enough, but as an improvement in the future, we might consider using a drop-down button to show all the job links. Let me know if you have any better ideas about how the job links should look like in the web UI.

PR #2228, opened 2019-05-02
[BUGFIX] Backward Incompatible to some JobTypes

kxu1026:
1. resolve the backward incompatible problem in 3.72.0 release.
2. fix an issue in new introduced interface. The object in Interface will be a static object which should be a new introduced bug in v3.72.0.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
@@ -45,10 +45,13 @@ public abstract class AbstractProcessJob extends AbstractJob {
   private static final String SENSITIVE_JOB_PROP_VALUE_PLACEHOLDER = "[MASKED]";
   private static final String JOB_DUMP_PROPERTIES_IN_LOG = "job.dump.properties";
 
-  private final String jobPath;
-  protected String cwd;
-  private volatile Props jobProps;
-  private volatile Props sysProps;
+  //Notes: These variables will be accessed directly throw inherited classes,
+  //       which are out of this package. Please remain them to be protected
+  protected final String _jobPath;
+  protected String _cwd;
+  protected volatile Props jobProps;
+  protected volatile Props sysProps;
+
   private volatile Props generatedProperties;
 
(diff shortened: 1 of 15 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -45,10 +45,13 @@
   private static final String SENSITIVE_JOB_PROP_VALUE_PLACEHOLDER = "[MASKED]";
   private static final String JOB_DUMP_PROPERTIES_IN_LOG = "job.dump.properties";
 
-  private final String jobPath;
+  //Notes: These variables will be accessed directly throw inherited classes,
+  //       which are out of this package. Please remain them to be protected
+  protected final String _jobPath;
   protected String cwd;
```
edwinalu: Please change to _cwd, in case any separate jobtype MPs are using it.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -58,7 +61,7 @@ protected AbstractProcessJob(final String jobId, final Props sysProps, final Pro
     this.jobProps = jobProps;
     this.sysProps = sysProps;
     this.cwd = getWorkingDirectory();
-    this.jobPath = this.cwd;
+    this._jobPath = this.cwd;
   }
 
   private File createOutputPropsFile(final String id, final String workingDir) {
```
edwinalu: Please change back to public for now, until we can verify no one is calling directly.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
kxu1026: good catch. copied.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
kxu1026: copied.

kxu1026: sorry. it was protected instead of public.

PR #2231, opened 2019-05-05
Add links to job logs in job context menu in execution graph

juhoautio:
Job log links have been only available in the `Job List` tab of an `Execution`, not on the `Graph` tab.

However, `Graph` is the landing tab for an `Execution`, and the easiest place to see the overall situation. It would be convenient to access job logs directly from the Graph by right-clicking a job.

This PR implements that.

Now the context-menu (with right-click) includes links to job logs 🎉:

Näyttökuva 2019-5-5 kello 20 13 31

----

## Tests & verifications

Näyttökuva 2019-5-5 kello 20 13 44

^ It works the same inside sub-flows (using nested ID so the URL is correct, too)

----

Näyttökuva 2019-5-5 kello 20 13 56

^ Jobs that haven't been run yet don't have log links

----

Näyttökuva 2019-5-5 kello 20 24 01

^ Disabled jobs don't have log links

----

Näyttökuva 2019-5-5 kello 20 14 21

^ In Flow view (not an Execution) jobs don't have log links

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
+++ b/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
@@ -75,6 +75,8 @@ var nodeClickCallback = function (event, model, node) {
 
   var requestURL = contextURL + "/manager?project=" + projectName + "&flow="
       + flowId + "&job=" + jobId;
+  var logURL = contextURL + "/executor?execid=" + execId + "&job="
+      + node.nestedId;
   var menu = [];
 
   if (type == "flow") {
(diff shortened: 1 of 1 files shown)


azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
```diff
@@ -161,13 +184,13 @@ var nodeClickCallback = function (event, model, node) {
...
       }
-    ];
+    ]);
   }
   contextMenuView.show(event, menu);
 }
 
 var jobClickCallback = function (event, model, node) {
-  console.log("Node clicked callback");
+  console.log("Job clicked callback");
```
juhoautio: Bonus fix. As you can know, it doesn't affect the actual behaviour.

juhoautio: By the way, the content of `jobClickCallback` seems pretty much the same as `nodeClickCallback`'s. But I don't even know what it's for? All cases that I tested about this PR worked correctly even though I didn't "echo" the changes into `jobClickCallback`.

burgerkingeater: if it's not being used anywhere, we should delete it.

juhoautio:
@burgerkingeater Thanks for the fast reviews!

jamiesjc:
Thanks @juhoautio for adding the link for job logs. Users appreciate and like this feature as this brings much convenience to them.
One suggestion from the user,
```
One suggestion -- to add new links in the middle of the menu.  This way it's less disruptive for long time users: we are used to the first menu item being Open Job, and continue right-clicking and clicking it expecting to open job properties, and being taken to the log instead.  
I would keep the first two entries as Open Job and Open Job in new windows, keep the last item as Center, and add the two new entries in the middle.
```
I'll let you decide whether to make the modification. Thanks!

juhoautio:
@jamiesjc I would mostly ignore how it _used_ to be and consider purely what's the best way to organize this context menu.

I chose to put log links on top because they are most relevant on nodes that have a start time.

Another way to order the items would be that most common menu items would be always on top, and then additional ones (if the node has a start time) would be added to the bottom of the menu. That would be least confusing, but also require moving mouse more to access the log links.

How do you feel, considering these viewpoints – order by relevance or by ubiquitousness?

jamiesjc:
I think it makes more sense to show `Open Job` at the top as this is common to every job and `Open Log` option below that since it only applies to jobs that have a start time.
Either way is fine though.

PR #2235, opened 2019-05-07
[update] use JAVA_HOME to exec java

ilibx:


burgerkingeater:
@ilibx can you share more details in your PR description?

ilibx:
@burgerkingeater for what?

ilibx:
@burgerkingeater merge one commit need PR ?

djaiswal83:
@ilibx can you provide the reason why you need those changes? Changing product code to accommodate testing is not a good idea to begin with.

ilibx:
@djaiswal83 i want manage it with ambari, but it can not be started by default as the java env, so i change the start cmd

Issue #2220, opened 2019-04-28
Flaky test XmlUserManagerTest > testAutoReload


juhoautio:
From https://travis-ci.org/azkaban/azkaban/builds/524621558?utm_source=github_status&utm_medium=notification:
```
azkaban.user.XmlUserManagerTest > testAutoReload STANDARD_OUT
    The config did not reload in 60 seconds

azkaban.user.XmlUserManagerTest > testAutoReload FAILED
    java.lang.AssertionError: The config did not reload in 60 seconds
        at org.junit.Assert.fail(Assert.java:88)
        at azkaban.user.XmlUserManagerTest.testAutoReload(XmlUserManagerTest.java:164)

azkaban.user.XmlUserManagerTest > testAutoReload STANDARD_OUT
    2019/04/25 19:05:01.439 +0000 INFO [XmlUserManager] Loading user user0
    2019/04/25 19:05:01.440 +0000 INFO [XmlUserManager] Loading user user1

azkaban.user.XmlUserManagerTest > testAutoReload STANDARD_OUT
    2019/04/25 19:05:01.447 +0000 INFO [XmlUserManager] Loading user user2
    2019/04/25 19:05:01.448 +0000 INFO [XmlUserManager] Loading user user3
    2019/04/25 19:05:01.448 +0000 INFO [XmlUserManager] Loading user user4
    2019/04/25 19:05:01.449 +0000 INFO [XmlUserManager] Loading user user5
    2019/04/25 19:05:01.450 +0000 INFO [XmlUserManager] Loading user user6
    2019/04/25 19:05:01.451 +0000 INFO [XmlUserManager] Loading user user7
    2019/04/25 19:05:01.451 +0000 INFO [XmlUserManager] Loading user user8
    2019/04/25 19:05:01.451 +0000 INFO [XmlUserManager] Loading user user9
    2019/04/25 19:05:01.486 +0000 INFO [UserUtils] Starting configuration watching thread.
    File modification time = 2019-04-25T19:04:37Z
    2019/04/25 19:05:01.508 +0000 INFO [UserUtils] Modification detected, reloading config file /home/travis/build/azkaban/azkaban/azkaban-common/build/resources/test/test-conf/azkaban-users-test1.xml

azkaban.user.XmlUserManagerTest > testAutoReload STANDARD_ERROR
    [Fatal Error] :-1:-1: Premature end of file.

azkaban.user.XmlUserManagerTest > testAutoReload STANDARD_OUT
[...]

PR #2240, opened 2019-05-11
Make XmlUserManagerTest run quicker

juhoautio:
Optimize XmlUserManagerTest:
- Before: 53 s
- After: 10 s

How:
- Configured for test scope to watch file every 2s instead of every 10s, which allows using shorter polling times & getting the expected update faster.
- Also reduced the sleep time after new notification(s) 1000ms->10ms.

See also https://github.com/azkaban/azkaban/issues/2220#issuecomment-491445541.

----

This is still too much of a system test I think. For example there's no need to have `testAutoReloadFail` as a full multithreaded case. It could be done simpler & without any extra sleeps by calling the reload method directly:
- create empty file
- call reload
- check that reload didn't change the returned value

And the behaviour of UserUtils in case of a parseConfigFile throws exception could be tested as a unit test that mocks both XmlUserManager & watchService (instead of using a real watchService).

----

Also this issue is still there:

> It looks like the test is also modifying a build-generated file (azkaban-common/build/resources/test/test-conf/azkaban-users-test1.xml). Even though test has a finally block to "revert the contents", this shouldn't be done IMHO. Instead, a copy of the file should be made to a temporary folder or something.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/user/UserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/user/UserUtils.java
@@ -45,11 +45,16 @@ public final class UserUtils {
 
   /**
    * Creates a watch thread which listens to specified files' modification and reloads
-   * configurations
+   * configurations.
+   *
+   * @param sensitivity a string value of the enum {@link SensitivityWatchEventModifier}, or null
    */
-  static void setupWatch(final Map<String, ParseConfigFile> configFileMap) throws IOException {
+  static void setupWatch(final Map<String, ParseConfigFile> configFileMap,
+      final String sensitivity, final long watchGracePeriodMillis) throws IOException {
     Preconditions.checkNotNull(configFileMap);
     Preconditions.checkArgument(configFileMap.size() > 0);
+
+    final long watchGracePeriod = resolveWatchGracePeriod(watchGracePeriodMillis);
 
     final WatchService watchService;
     try {
@@ -84,7 +89,7 @@ public final class UserUtils {
           // There is no entry for this directory, create a watchkey
           final WatchKey watchKey = dir.register(watchService,
               new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_MODIFY},
-              SensitivityWatchEventModifier.MEDIUM);
+              resolveSensitivity(sensitivity));
           keys.put(watchKey, dir);
         }
         // Add the config file to dir map
@@ -121,7 +126,7 @@ public final class UserUtils {
           // more important that the config reloads successfully than immediately.
           // If there is any modification happening to file(s) in the meantime, it is all queued up
           // in the watch service.
-          Thread.sleep(1000L);
+          Thread.sleep(watchGracePeriod);
         } catch (final InterruptedException ie) {
           log.warn(ie.toString());
           Thread.currentThread().interrupt();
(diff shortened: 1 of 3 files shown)


djaiswal83:
This is a classic case of modifying the product to make test work. I don't think it is ideal. From product perspective, we don't need to pass sensitivity as a parameter.
May I ask, why the urgency to run the tests quickly? It is not like it is adding significant time to the build process.

juhoautio:
> This is a classic case of modifying the product to make test work. I don't think it is ideal. From product perspective, we don't need to pass sensitivity as a parameter.

Fair enough. But I still don't think this is too much. _If_ it's needed to test on system level like this, this is still better than letting the test suite runtime grow like that. Personally I would rather use mocks & just test against the contracts without any need for Thread.sleep.

djaiswal83:
Ok, lets make the sensitivity high and see the test runtime and take out all the other changes. Using medium sensitivity was to stay conservative, but I guess we have enough fail safe now to work with high sensitivity as well. I tested it out locally and it works fine.

djaiswal83:
Let me know if I should push a patch for it.

juhoautio:
If `user.manager.xml.watchGracePeriodMillis` is removed then there will be +1s sleep inside the loop. Maybe `testAutoReloadFail` polling time must be bumped up from 5 seconds to avoid random failures on Travis. It could actually be too low even now. Maybe it would need at least 10s to guarantee stability on Travis.

But another problem with testAutoReloadFail is it doesn't even know if reload was called or not (ie. was there an event from the watch service or not).

I would be happy to replace the whole test class with a mock-based test if you're fine with it.

Otherwise, may I ask is there a place where we could discuss principles/best practices for the Azkaban project in general? For example on gitter maybe? Are you currently in charge, or is it more like some collaborative effort among LinkedIn devs? As a contributor, to me it seems like a weird change that test stability & build speed are not cared about as much as before, which is adding overhead to running the checks on PRs.

djaiswal83:
I have never done a mock based test before. If that is something which can be done easily, please go ahead with it.
As far as build speed is concerned, yes we like to do it ASAP, however, this test is something of an outlier from what we have done before. Since, it requires another thread to finish, polling needs to be done which can cause it to take longer than usual.
As I mentioned earlier, if the sensitivity is brought up to HIGH, the polling happens every 2 seconds instead of 30 seconds which makes the biggest difference as it happens twice in the overall test suite.
The 1 second sleep in the watch service is well documented and should remain like that.

juhoautio:
Ok, then I'll be happy to convert to mocks. The most important part will be using a mock object instead of a real `WatchService` obtained from `FileSystems.getDefault().newWatchService();`. The test will setup the mock to immediately return the expected sequence of results, for example `watchService.take()` would immediately return the WatchEvent without any blocking. As you can imagine, there would be no need to sleep because in the mocked context the file is won't be written to in parallel. One option is to create an interface that wraps WatchService and includes the additional sleep in scope of take(). That interface would be the one that can be mocked, and passed as an argument in the unit test. Otherwise, real impl provided by Guice. Does that sound ok to you?

djaiswal83:
Sounds okay. However, can you set the sensitivity to HIGH and see if it is satisfactory for you before going down this lane?
I would prefer to test the real code instead of just mocking the logic.

juhoautio:
Thanks. I tried it. It's 15s (using 10s poll time to ensure stability). So not a major difference.

15s is still too much, it won't work in general if tests like this are getting added. 10s was also still too much in my opinion, but I wanted to submit what I was able to reach with the obvious optimizations.

Any way, there's also that aspect of not actually verifying that there was a WatchEvent. Both issues can be squashed by separating the concerns with mock-based tests.

> I would prefer to test the real code instead of just mocking the logic

The actual WatchService can also be tested separately (after extracting that part to a separate class). But then it doesn't need to be a multithreaded case with complicated logic. The maximum hit that it will get will be 2s with from sensitivity + 1s from Thread.sleep(). However, when it's extracted into a separate class sleep time can naturally be an optional constructor argument that tests can pass to decrease the sleep. But I'll prepare that PR, I guess it's best to just see the code.

juhoautio:
Closing this in favor of https://github.com/azkaban/azkaban/pull/2244

PR #2241, opened 2019-05-11
Flow override properties to also override job properties

juhoautio:
Currently job properties always win over flow override properties.

However in practice it's quite handy to be able to override even those.

The UI currently says:

> Flow Parameters / Flow Property Override
> Add temporary flow parameters that are used to override global settings for each job

Also the [docs](https://azkaban.readthedocs.io/en/latest/useAzkaban.html?highlight=override#flow-parameters) read:

> The flow parameters override the global properties for a job, but not the properties of the job itself.

It seems like the intention has been that only flow level props are overridden, and if a job defines the same property, it still wins.

This is kinda confusing. Because on the other hand an overridden flow property is passed to the job, if the job itself doesn't set that property at all.

In other words, if a property that's provided as flow parameter is not (originally) defined by job, it still "bleeds" to that job.

----

As a side note, I would like to implement a feature that additionally allows overriding properties for specific jobs (or sub-flows) only. Currently it of course affects all jobs. On the UI the flow override tab could be extended to be similar with job level SLA settings: there's a dropdown picker for job name etc.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -786,6 +786,13 @@ public class FlowRunner extends EventHandler implements Runnable {
       props = jobSource;
     }
 
+    // 5. If there are any runtime flow overrides, we apply them now.
+    final Map<String, String> flowParam =
+        this.flow.getExecutionOptions().getFlowParameters();
+    if (flowParam != null && !flowParam.isEmpty()) {
+      props.putAll(flowParam);
+    }
+
     node.setInputProps(props);
   }
 
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java
@@ -40,10 +40,10 @@ import org.junit.Test;
  * The tests are contained in execpropstest, and should be resolved in the following fashion, where
  * the later props take precedence over the previous ones.
  *
- * 1. Global props (set in the FlowRunner) 2. Shared job props (depends on job directory) 3. Flow
- * Override properties 4. Previous job outputs to the embedded flow (Only if contained in embedded
- * flow) 5. Embedded flow properties (Only if contained in embedded flow) 6. Previous job outputs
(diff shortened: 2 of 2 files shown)


juhoautio:
Could someone comment on this, please?

juhoautio:
Oh, I was wondering this might not be included lightly. Do you have a good view to how your users are for example leveraging the flow properties? Because this is kind of a major change of behavior. Also if this is merged in, then documentation should be of course updated accordingly.

juhoautio:
@djaiswal83 So yeah I would expect that you would want to check on the usage of flow properties on your side before this could be merged. Do you have any further insight at this point?

juhoautio:
@burgerkingeater @jamiesjc would it be possible to also hear your opinion?

juhoautio:
@djaiswal83 @li-gsalia @abagshaw could you weigh in, please?

juhoautio:
Bump :)

djaiswal83:
Hi. I will bring this up internally and update here.

juhoautio:
@djaiswal83 bump

djaiswal83:
Hi Juho. Sorry abou that. I will bring it up in discussion tomorrow.

djaiswal83:
@juhoautio sorry for the delay in response. I believe this is a great feature. Please submit a PR for it.
I will merge this PR in the meantime. Or if you want to make the change as part of this PR then please let me know.

juhoautio:
@djaiswal83 thanks! Please merge this PR if you feel this is safe enough to do, considering that this is not entirely backward compatible.

If you have concerns about that, just let me know if you'd like me to modify to apply this only if a property (azkaban.FlowOverrideExistingProperties=true) is set. That could be set in the default props so that new users would get this behavior automatically, but existing custom server props that people have would not have it. Code would assume false for backward compatibility if the property is not present.

Regarding the job specific overrides, I've implemented it, and I could prepare a PR once we've dealt with this smaller changed first.

djaiswal83:
Pushed this one.
Looking forward to your PR.

djaiswal83:
@juhoautio we need to revert his PR as we faced issues during deployment.
https://github.com/azkaban/azkaban/pull/2620

Can you please make it configurable and disabled by default?

juhoautio:
@djaiswal83 sure. So you actually had some cases where flow override props are supposed to only apply for cases where the same property is _not_ defined by jobs?

PR #2250, opened 2019-05-17
Fix NPE in dynamically generating job link url.

jamiesjc:
This is related to PR #1695 
The NPE happens when the job link has expired for a job inside an embedded flow.
To fix that, call `getExecutableNodePath` instead of `getExecutableNode` to get the job inside an embedded flow.
Verified the fix on localhost.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -318,7 +318,14 @@ public class ExecutionControllerUtils {
       jobLinkUrl = url.toString();
     } else {
       // If RM job link is invalid or has expired, fetch the job link from JHS or SHS.
-      if (exFlow.getExecutableNode(jobId).getType().equals(SPARK_JOB_TYPE)) {
+      final ExecutableNode node = exFlow.getExecutableNodePath(jobId);
+      if (node == null) {
+        logger.error(
+            "Invalid job link. Job " + jobId + " doesn't exist in " + exFlow.getExecutionId());
+        return null;
+      }
+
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -318,7 +318,14 @@ public static String createJobLinkUrl(final ExecutableFlow exFlow, final String
       jobLinkUrl = url.toString();
     } else {
       // If RM job link is invalid or has expired, fetch the job link from JHS or SHS.
-      if (exFlow.getExecutableNode(jobId).getType().equals(SPARK_JOB_TYPE)) {
+      final ExecutableNode node = exFlow.getExecutableNodePath(jobId);
+      if (node == null) {
```
djaiswal83: In which scenario this is a possibility?

jamiesjc: It should not be null unless the input jobId parameter is incorrect.

ypadron-in:
Not related to this issue but can you close the reader in line 299?

jamiesjc:
@ypadron-in Here it is using `try-with-resources`. `The try-with-resources statement ensures that each resource is closed at the end of the statement. `
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
djaiswal83: Thanks

PR #2251, opened 2019-05-20
Add logs to ProjectManager initialization at web server startup.

ypadron-in:
Add logs to ProjectManager initialization to better understand a consistent latency at web server startup on some of our clusters. 
The suspicions are that this is related to the existing number of project and flows and the fact that currently all active projects and their flows are loaded into memory at web server startup.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -45,13 +45,14 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 @Singleton
 public class ProjectManager {
 
-  private static final Logger logger = Logger.getLogger(ProjectManager.class);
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -117,6 +119,7 @@ public boolean hasFlowTrigger(final Project project, final Flow flow)
 
   private void loadAllProjects() {
     final List<Project> projects;
+    logger.info("Loading active projects.");
```
jamiesjc: Do you think it will be more clear if we print out the time for the loading process instead? Something like:
```
t1 = System.currentTimeMillis();
// loading active projects
t2 = System.currentTimeMillis();
logger.info("Time to load active projects is" + (t2-t1));
```

djaiswal83: Sounds like a good idea. Gives relevant information.

ypadron-in: We can look at the timestamps in the logs to get those times.
I would consider adding time calculations if we were going to take some action in code for example if the delay goes above some threshold.

ypadron-in:
> I just realized that this one used log4j. Maybe it is a good opportunity to update it to use slf4j?

I changed the code to use slf4j instead.

PR #2255, opened 2019-05-28
Fix HDFS viewer authentication issue.

jamiesjc:
Users are unable to access hdfs viewer after 24 hours because of kerberos key expiration. 
`java.io.IOException: Failed on local exception: java.io.IOException: javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]; `
When dynamic job link url is generated, it creates a new login using keytab file which interferes with the hdfs viewer Kerberos ticket refresh process. To fix the issue, use the same login user created by previous hdfs login and refresh ticket when it expires.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/AuthenticationUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/AuthenticationUtils.java
@@ -17,11 +17,7 @@ package azkaban.utils;
 
 import java.net.HttpURLConnection;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.security.PrivilegedExceptionAction;
-import java.util.ArrayList;
-import java.util.List;
-import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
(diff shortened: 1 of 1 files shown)


burgerkingeater:
thanks for the fix. `When dynamic job link url is generated, it creates a new login using keytab file which interferes with the hdfs viewer Kerberos ticket refresh process. ` can you elaborate more on this in the PR description?

jamiesjc:
@burgerkingeater 
I didn't find any documentation to support my theory but that's what I found out from practice. And here is the recommended way of doing Hadoop authentication. Basically, call `loginUserFromKeytab`  on startup and call `checkTGTAndReloginFromKeytab` before every action of calling Hadoop rest API.

References:
> https://stackoverflow.com/questions/34616676/should-i-call-ugi-checktgtandreloginfromkeytab-before-every-action-on-hadoop

> https://stackoverflow.com/questions/33211134/hbase-kerberos-connection-renewal-strategy/33243360#33243360

PR #2256, opened 2019-05-28
FileWatcherTest fix

djaiswal83:
- The test works fine locally, however, it fails on test servers,
  possibly due to permission issues
- Using tried and tested method to create PATH which works in other
  tests.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
+++ b/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
@@ -41,7 +41,7 @@ public class FileWatcherTest {
   private Path setPath() {
     final URL configURL = Resources.getResource("test-conf/azkaban-users-test1.xml");
     final String origpath = configURL.getPath();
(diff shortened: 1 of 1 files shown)


azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -33,9 +35,17 @@
...
 
-  private static final Path PATH = Paths.get("build/tmp/FileWatcherTest");
+  private final Path PATH = setPath();
   private FileWatcher fileWatcher;
 
+  private Path setPath() {
+    final URL configURL = Resources.getResource("test-conf/azkaban-users-test1.xml");
+    final String origpath = configURL.getPath();
+    // Copy the file to keep original file unmodified
+    final String path = origpath.replace("test1", "file_watcher");
```
burgerkingeater: why not getting from `test-conf/azkaban-users-file_watcher.xml` directly?

PR #2259, opened 2019-05-30
Fix flaky FileWatcherTest

juhoautio:
The failure on [Travis](https://travis-ci.org/azkaban/azkaban/builds/539101458?utm_source=github_status&utm_medium=notification) was:

```
  azkaban.user.FileWatcherTest > registerAndTake FAILED
      java.lang.AssertionError: expected:<1> but was:<2>
          at org.junit.Assert.assertEquals(Assert.java:645)
          at org.junit.Assert.assertEquals(Assert.java:631)
          at azkaban.user.FileWatcherTest.registerAndTake(FileWatcherTest.java:75)
```

Which is for line 75:
https://github.com/azkaban/azkaban/blob/505d2847eb88f6dae09fc91127dabc8711a9c391/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java#L74-L75

This is familiar from previous life where we had a multi-threaded test. That test was working because it retried on read error and thus also tolerated 1 or more events from watch service.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
+++ b/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
@@ -16,6 +16,7 @@
 
 package azkaban.user;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 
 import com.google.common.io.Resources;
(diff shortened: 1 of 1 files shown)


djaiswal83:
My bad here.

juhoautio:
@djaiswal83 no, this issue was present already in my initial commit:
https://github.com/azkaban/azkaban/blob/320f3c09efaba4f41a81adb041c16bb9ecaebfab/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java#L65

PR #2261, opened 2019-05-30
Make UserUtils.setupWatch public.

djaiswal83:


burgerkingeater:
can you elaborate the motivation of this change in your PR description?

djaiswal83:
The motivation is to allow setupWatch to be called from outside of azkaban-user package.

PR #2268, opened 2019-06-12
Handle non-existent flows when asking for isLocked property on Flow Execution page load.

ypadron-in:
Loading the Flow Execution page of an execution of a flow that doesn't exist anymore (because an update to a project deleted it for example) causes a NullPointerException because we ask whether that non-existent flow is locked or not to enable/disable "Prepare Execution" button in the view.

More details about "locking flows" feature in #2210.


djaiswal83:
The build failed.

PR #2270, opened 2019-06-14
Prevent login DDoS by enforcing max number of sessions per IP per user

burgerkingeater:
This PR adds a configurable upper limit on the number of sessions per IP per User. This is useful when one misbehaving user is creating too many sessions on the same host, kicking out other users' sessions.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
+++ b/azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
@@ -18,11 +18,13 @@ package azkaban.server.session;
 
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.utils.Props;
+import azkaban.utils.UndefinedPropertyException;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
@@ -45,6 +47,7 @@ public class SessionCache {
   private final Cache<String, Session> cache;
 
   private final long effectiveSessionTimeToLive;
+  private final Optional<Long> maxNumberOfSessionsPerIpPerUser;
 
   /**
    * Constructor taking global props.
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -53,12 +56,26 @@
   public SessionCache(final Props props) {
     this.effectiveSessionTimeToLive = props.getLong(ConfigurationKeys.SESSION_TIME_TO_LIVE,
         DEFAULT_SESSION_TIME_TO_LIVE);
+
+    Long maxNumberOfSessions = null;
+    try {
+      maxNumberOfSessions = props.getLong(ConfigurationKeys.MAX_SESSION_NUMBER_PER_IP_PER_USER);
```
ypadron-in: Should we modify the getLong() method to return an Optional<Long>? It seems this is the only place this method is used so the refactor should be easy?

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -75,11 +92,42 @@ public long getSessionCount() {
...
+   * @return <code>true</code> The number of cached session sharing the same IP and user equals
+   * or greater than the allowed limit if defined;
+   * <code>false</code> otherwise.
+   */
+  private boolean isViolatingMaxNumberOfSessionPerIpPerUser(final Session session) {
+    if (this.maxNumberOfSessionsPerIpPerUser.isPresent()) {
+      // get sessions sharing the same IP and user
+      final Set<Session> sessionsWithSameIP = this.findSessionsByIP(session.getIp());
+      int duplicateSessionCount = 0;
+      for (final Session sessionByIP : sessionsWithSameIP) {
```
ypadron-in: Talked offline about encapsulating this logic in a new method findSessionsByIPByUser or similar name.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
djaiswal83: getLong is too generic as the name suggests. Modifying it to return Optional may not be ideal. A wrapper on top of it is possible.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
burgerkingeater: added more comments to make this piece more understandable.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
burgerkingeater: I agree it's better for getLong to return null instead throwing exception when key is not found, but involves a bigger change, will make it a todo.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
burgerkingeater: updated: extract this piece to a new function.

azkaban-common/src/main/java/azkaban/server/session/SessionCache.java
```diff
@@ -53,12 +57,26 @@
   public SessionCache(final Props props) {
     this.effectiveSessionTimeToLive = props.getLong(ConfigurationKeys.SESSION_TIME_TO_LIVE,
         DEFAULT_SESSION_TIME_TO_LIVE);
+
+    Long maxNumberOfSessions = null;
+    try {
+      maxNumberOfSessions = props.getLong(ConfigurationKeys.MAX_SESSION_NUMBER_PER_IP_PER_USER);
+    } catch (final UndefinedPropertyException exception) {
+      maxNumberOfSessions = null;
```
ypadron-in: This variable is null already. Should we replace this line with a log message?

burgerkingeater: thanks, removed the initializer.

PR #2271, opened 2019-06-18
Delete execution directory on flow finishing

burgerkingeater:
This PR moves execution directory cleanup out from CleanerThread, and deletes execution directory on flow finishing. The motivation is to make sure that execution directory gets deleted on time without leaving any orphan disk-consuming dirs. Previously CleanerThread periodically traverses every execution dir periodically and deletes any directories which haven't been modified for some configurable period. The logic is cumbersome and error-prone as it requires checking against shared data structure like running flows to make sure it's not deleting execution directories which are in use.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -881,8 +881,6 @@ public class FlowRunnerManager implements EventListener,
 
   private class CleanerThread extends Thread {
 
-    // Every 5 mins, clean execution dir.
-    private static final long EXECUTION_DIR_CLEAN_INTERVAL_MS = 5 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


burgerkingeater:
@li-ygerchikov thanks, see my answer inline
Would we ever want to keep execution dir, e.g. for debugging purposes?
- yes, we kept it for debugging purpose but we never actually look at the execution dir for that purpose from my experience. 

Have we captured all the logs and everything that's worth saving from it before we delete it after flow finishes?
- yes, log will be persisted to mysql before the deletion

Are there any situations where handleEvent is not called for a given flow -- e.g. if executor dies or is killed, or if execution terminates abnormally?
- i have tested the case where execution is killed. When executor is shutdown, there's code to handle deletion all execution dirs. 

The main purpose of this code as PR suggests is to simplify the logic and make sure all execution directories get deleted on time to avoid disk penalty.

PR #2276, opened 2019-06-25
Some jobtypess which use UserGroupInformation.doAs do not throw exception gracefully when doAs throws UndeclaredThrowableException

djaiswal83:


jamiesjc:
Have you checked all the places where `UserGroupInformation.doAs` is called?

djaiswal83:
> Have you checked all the places where `UserGroupInformation.doAs` is called?

Yes. These are the only two places in job types code where this exception is not handled. Rest of the places throw Exception which covers everything.

djaiswal83:
Thanks for the review.

PR #2266, opened 2019-06-08
Improve condition on status whether to show job log links

juhoautio:

PR #2285, opened 2019-07-11
Reorder job context menu: move log links to bottom

juhoautio:
As [discussed](https://github.com/azkaban/azkaban/pull/2231#issuecomment-510162091) with @jamiesjc, it makes more sense to show `Open Job` and `Center Job` at the top as this is common to every job and `Open Log` options below them since they are only shown for job nodes that have a start time.

With this change:

Näyttökuva 2019-7-11 kello 10 21 06

For the record, [original suggestion](https://github.com/azkaban/azkaban/pull/2231#issuecomment-509851538) was a bit different

> I would keep the first two entries as Open Job and Open Job in new windows, keep the last item as Center, and add the two new entries in the middle.

However this PR puts also `Center` before the log links, as it is also shown in all occasions.

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
+++ b/azkaban-web-server/src/web/js/azkaban/util/flow-loader.js
@@ -148,23 +148,6 @@ var nodeClickCallback = function (event, model, node) {
     // 1. Flow -> Graph tab
     // 2. Execution -> Graph tab
     menu = [];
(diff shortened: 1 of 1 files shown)


juhoautio:
Oops, conflicts with https://github.com/azkaban/azkaban/pull/2266. It doesn't matter which one is merged first, but other needs to be rebased after.

juhoautio:
@jamiesjc ok, resolved

PR #2290, opened 2019-07-19
AZNewDispatchingLogic - Retry mechanism for flow submission failure.

jamiesjc:
In the new dispatching logic, executors poll executions from DB and submit.
If a flow execution fails to be submitted, e.g. due to some server outage, the current executor should unset the executor id for that flow so that other executors can still pick it up and re-submit. 
This PR implements exponential backoff retries for flow submission failures. Basically, if a flow fails to be submitted, the executor will retry in 1s, 2s, 4s, 8s ... until a successful submission happens, then it will go back to 1s interval. 
Also, use the `update_time` instead of `submit_time` as the ordering in DB so that new flows will get the chance to be submitted.
Verified the PR change in solo server.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -968,10 +968,11 @@ public class FlowRunnerManager implements EventListener,
   @SuppressWarnings("FutureReturnValueIgnored")
   private class PollingService {
 
-    private final long pollingIntervalMs;
     private final ScheduledExecutorService scheduler;
-    private int executorId = -1;
     private final PollingCriteria pollingCriteria;
+    private final long pollingIntervalMs;
+    private int executorId = -1;
+    private int numRetries = 0;
 
     public PollingService(final long pollingIntervalMs, final PollingCriteria pollingCriteria) {
       this.pollingIntervalMs = pollingIntervalMs;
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1002,12 +1003,30 @@ private void pollExecution() {
...
-            submitFlow(execId);
-            FlowRunnerManager.this.commonMetrics.markDispatchSuccess();
+            try {
+              submitFlow(execId);
+              FlowRunnerManager.this.commonMetrics.markDispatchSuccess();
+              this.numRetries = 0;
+            } catch (final ExecutorManagerException e) {
+              // If the flow fails to be submitted, then unset its executor id in DB so that other
+              // executors can pick up this flow and submit again.
+              FlowRunnerManager.this.executorLoader.unsetExecutorIdForExecution(execId);
```
ypadron-in: We are calling unsetExecutorIdForExecution() method from FlowRunner.createFlowRunner() as well. Consider removing the call from there to avoid unsetting executor id twice.

jamiesjc: Good point. Unfortunately, `FlowRunner.createFlowRunner()` is also called from `ExecutorManager` as well, removing it would mess up with the old dispatching logic.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1002,12 +1003,30 @@ private void pollExecution() {
...
+            }
           }
         } catch (final Exception e) {
           FlowRunnerManager.logger.error("Failed to submit flow ", e);
           FlowRunnerManager.this.commonMetrics.markDispatchFail();
+          this.numRetries = this.numRetries + 1;
+          try {
+            // Implement exponential backoff retries when flow submission fails,
+            // i.e., sleep 1s, 2s, 4s, 8s ... before next retries.
+            Thread.sleep((long) (Math.pow(2, this.numRetries) * 1000));
```
jamiesjc: As suggested by @ypadron-in , the backoff time should be implemented inside the scheduler itself. But I didn't find out any existing library that provides this functionality, so just implemented my own backoff time during each scheduled run. Will keep it in the todo list.

PR #2291, opened 2019-07-19
Add a dropdown button in Job Details page to support multiple Hadoop jobs in the Az logs

ypadron-in:
There are cases where a single Azkaban job launches multiple Map-Reduce jobs (for example Pig and Hive jobs) and therefore there can be multiple <application_id>s in the job log.

In the Job Details page, in those cases of multiple <application_id>s, there will be a dropdown button listing all the <application_id>s in the Azkaban job log in the order they were launched with links to their respective logs in Spark/Hadoop history servers.

Screen Shot 2019-07-19 at 10 52 17 AM

This is related to PR #1695

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -32,6 +32,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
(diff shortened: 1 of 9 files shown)


ypadron-in:
> Thanks for improving the job link button! As discussed offline, please capture the thoughts of doing the job link on the client side and also page loading time.

Resolving the url to job logs in Hadoop/Spark servers in the backend means we need to process the job log before replying to the browser with the html code of the page. For long job logs the response from the web server may take a bit longer.

In the future we may want to consider resolving the url in the client side, when the user clicks on the button, to improve/resolve page load time issues.

PR #2305, opened 2019-08-16
Add validation to 'working.dir' and absolute path to chown.

ypadron-in:


scgchiappara:
I used working.dir property to run R scripts that need to run on a specific directory.

What is the proper way to do that now?

PR #2306, opened 2019-08-16
More logging around triggers.

djaiswal83:


azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
```diff
@@ -50,20 +51,26 @@ public Trigger(final int execId,
...
   public void run() {
+    logger.info("Running trigger for " + this);
     if (isTriggerExpired()) {
       logger.info(this + " expired");
       return;
     }
 
+    logger.info("Check if trigger condition met for " + this);
     final boolean isTriggerConditionMet = this.triggerCondition.isMet();
+    logger.info("Trigger condition for execid = " + this.execId + " met? = " + isTriggerConditionMet);
```
jamiesjc: It seems line 62 and 64 can be consolidated.

azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
```diff
@@ -50,20 +51,26 @@ public Trigger(final int execId,
    */
   @Override
   public void run() {
+    logger.info("Running trigger for " + this);
     if (isTriggerExpired()) {
       logger.info(this + " expired");
       return;
     }
 
+    logger.info("Check if trigger condition met for " + this);
```
jamiesjc: Is line 60 really needed?

azkaban-exec-server/src/main/java/azkaban/execapp/Trigger.java
```diff
@@ -50,20 +51,26 @@ public Trigger(final int execId,
...
     final boolean isTriggerConditionMet = this.triggerCondition.isMet();
+    logger.info("Trigger condition for execid = " + this.execId + " met? = " + isTriggerConditionMet);
     if (isTriggerConditionMet) {
       logger.info("Condition " + this.triggerCondition.getExpression() + " met");
       for (final TriggerAction action : this.actions) {
         try {
+          if (action instanceof KillExecutionAction) {
+            logger.info("Killing execution " + this.execId);
```
jamiesjc: There is already logging inside `KillExecutionAction`, so probably don't need this?

PR #2307, opened 2019-08-17
Add support to allow ctor with sigs (java.lang.String,azkaban.utils.Props,azkaban.utils.Props,org.apache.log4j.Logger) in HadoopJavaJobRunnerMain

djaiswal83:
Some jobtypes require this additional signature.


burgerkingeater:
@djaiswal83 what is ctor? Can you elaborate more on the motivation of this PR?

djaiswal83:
Thanks for the review.

PR #2317, opened 2019-08-29
Append flow submit user to Hadoop token principal in email format

djaiswal83:
Append flow submit user to Hadoop token principal in email format which conforms to Kubernetes.

Above can be configured using HadoopSecurityManager and are not mandatory.
A minor fix to log message.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -235,7 +235,7 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
     if (hsmInstance == null) {
       synchronized (HadoopSecurityManager_H_2_0.class) {
         if (hsmInstance == null) {
-          logger.info("getting new instance");
+          logger.info("getting new instance of HadoopSecurityManager");
           hsmInstance = new HadoopSecurityManager_H_2_0(props);
         }
       }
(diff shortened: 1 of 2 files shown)


jamiesjc:
Can you explain the reason why we want to append flow submit user? Also, any reference to the usage pattern in Kubernetes would be helpful.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -470,37 +469,47 @@ public void prefetchToken(final File tokenFile, final Props props, final Logger
...
+            "/" + props.getString(Constants.FlowProperties.AZKABAN_FLOW_SUBMIT_USER) +
+                props.getString(HadoopSecurityManager.SUBMIT_USER_SUFFIX) : "";
+
+    final String userToProxyFQN = userToProxy + suffix;
+    logger.info("Getting hadoop tokens based on props for " + userToProxyFQN);
+
     final Credentials cred = new Credentials();
-    fetchMetaStoreToken(props, logger, userToProxy, cred);
-    fetchJHSToken(props, logger, userToProxy, cred);
+    fetchMetaStoreToken(props, logger, userToProxyFQN, cred);
```
jamiesjc: How would the current `getDelegationToken` work if `userToProxy` is changed?

djaiswal83:
> Can you explain the reason why we want to append flow submit user? Also, any reference to the usage pattern in Kubernetes would be helpful.

This helps in debugging who is the owner of the flow or who scheduled it. If the flow submit user is provided as an email address then Kubernetes can handle it.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: userToProxy would change from something like "azkTest" to "azkTest/deepak@xyz.com"
Delegation tokens work if a valid email ID is provided as tested on actual Azkaban cluster.
The original userToProxy is only used for prepareTokenFile() method as OS user still remains same.

PR #2320, opened 2019-09-03
Add additional logging to track SLA violation

djaiswal83:
Add additional logging to track why once in a while flow level SLA is not honored.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
@@ -69,9 +69,9 @@ public class SlaChecker implements ConditionChecker {
 
   private Boolean isSlaMissed(final ExecutableFlow flow) {
     final SlaType type = slaOption.getType();
-    logger.info("SLA type for flow " + flow.toString() + " is " + type);
+    logger.info("SLA type for flow " + flow.getId() + " is " + type);
     if (flow.getStartTime() < 0) {
-      logger.info("Start time is less than 0 for flow " + flow.toString());
+      logger.info("Start time is less than 0 for flow " + flow.getId());
       return false;
     }
 
@@ -80,12 +80,12 @@ public class SlaChecker implements ConditionChecker {
       logger.info("SLA type is flow.");
       if (this.checkTime < flow.getStartTime()) {
         logger.info("checktime = " + this.checkTime);
-        logger.info("SLA duration = " + slaOption.getDuration().toMillis());
+        logger.info("SLA duration = " + slaOption.getDuration().toMillis() + " ms");
         this.checkTime = flow.getStartTime() + slaOption.getDuration().toMillis();
         logger.info("checktime updated to " + this.checkTime);
       }
       status = flow.getStatus();
-      logger.info("Flow status = " + status);
+      logger.info("Flow status = " + status.toString());
     } else { // JOB
       final ExecutableNode node = flow.getExecutableNode(slaOption.getJobName());
       if (node.getStartTime() < 0) {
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -69,16 +69,23 @@ public static SlaChecker createFromJson(final HashMap<String, Object> obj)
...
+      logger.info("Start time is less than 0 for flow " + flow.toString());
       return false;
     }
 
     Status status;
     if (type.getComponent() == SlaType.ComponentType.FLOW) {
+      logger.info("SLA type is flow.");
       if (this.checkTime < flow.getStartTime()) {
+        logger.info("checktime = " + this.checkTime);
+        logger.info("SLA duration = " + slaOption.getDuration().toMillis());
```
jamiesjc: Consider adding the unit for the logging, e.g. `ms`?

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -69,16 +69,23 @@ public static SlaChecker createFromJson(final HashMap<String, Object> obj)
...
         this.checkTime = flow.getStartTime() + slaOption.getDuration().toMillis();
+        logger.info("checktime updated to " + this.checkTime);
       }
       status = flow.getStatus();
+      logger.info("Flow status = " + status);
```
jamiesjc: Would `status.toString()` be better?

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -92,15 +99,18 @@ private Boolean isSlaMissed(final ExecutableFlow flow) {
...
          case FLOW_SUCCEED:
+           logger.info("isFlowSucceeded?");
             return !isFlowSucceeded(status);
          case JOB_FINISH:
             return !isJobFinished(status);
          case JOB_SUCCEED:
             return !isJobFinished(status);
       }
     } else if (slaOption.getType().getStatus() == StatusType.SUCCEED) {
+      logger.info("slaOption.status = SUCCEED and status = " + status);
```
jamiesjc: Same ^^ `status.toString()`

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
```diff
@@ -162,6 +172,7 @@ public Object isSlaFailed() {
     final ExecutableFlow flow;
     try {
       flow = this.executorLoader.fetchExecutableFlow(this.execId);
+      logger.info("Flow for execid " + this.execId + " is " + flow.toString());
```
jamiesjc: Why to you need to print out `flow.toString()`? Would `flow.getId()` be better?

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
djaiswal83: Looking at line 84, I believed checkTime would be in ms only.

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
djaiswal83: Makes sense, I will do that.

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
djaiswal83: ditto

azkaban-common/src/main/java/azkaban/trigger/builtin/SlaChecker.java
djaiswal83: agreed

PR #2324, opened 2019-09-06
fix minor typo on permissions modal

abagshaw:
The modal for adding a group used to have the input field mislabeled as "User". This fixes that issue and properly labels it as "Group".

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/project-permissions.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/project-permissions.js
@@ -26,7 +26,6 @@ azkaban.PermissionTableView = Backbone.View.extend({
 
   initialize: function (settings) {
     this.group = settings.group;
-    this.proxy = settings.proxy;
   },
 
   render: function () {
@@ -34,8 +33,7 @@ azkaban.PermissionTableView = Backbone.View.extend({
 
   handleChangePermission: function (evt) {
     var currentTarget = evt.currentTarget;
-    changePermissionView.display(currentTarget.id, false, this.group,
-        this.proxy);
+    changePermissionView.display(currentTarget.id, false, this.group);
   }
 });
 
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/web/js/azkaban/view/project-permissions.js
```diff
@@ -171,6 +171,16 @@ azkaban.ChangePermissionView = Backbone.View.extend({
     var executeInput = $("#" + prefix + "-execute-checkbox");
     var scheduleInput = $("#" + prefix + "-schedule-checkbox");
 
+    if (group) {
+      $('#change-type').text("Group");
+    }
+    else if (proxy) {
```
ypadron-in: I think we are using a different modal for proxy users so handling this case here is not necessary. We should also remove other references to proxy user in this method.

ypadron-in:
Thanks for the fix!

ypadron-in:
Also, please add _"?v=\<timestamp\>"_ to the value of "src" attributes in all the views using the modified .js files, like so:
`<script type="text/javascript" src="${context}/js/azkaban/view/project-permissions.js?v=1556216524794"></script>`

This is to force the reload of those files for all clients.

azkaban-web-server/src/web/js/azkaban/view/project-permissions.js
```diff
@@ -148,12 +148,10 @@ azkaban.ChangePermissionView = Backbone.View.extend({
     $('#change-permission-error-msg').hide();
   },
 
-  display: function (userid, newPerm, group, proxy) {
```
ypadron-in: I believe we can delete also lines: 29, 317, 322, 326, 327. Basically 'proxy' property is not used in PermissionTableView.
Also please remove parameter proxy from line 39.

PR #2325, opened 2019-09-06
change checkboxes to use .prop instead of .attr

abagshaw:
.attr() cannot modify the 'checked' property of a checkbox, but .prop() can. So .prop() should be used for this purpose.

Previously this showed up as a minor UI glitch where checking the "Admin" permission when adding or changing a user or group did not properly check all the other checkboxes (i.e. read, write, execute and schedule).


ypadron-in:
Please add "_?v=\<timestamp\>_" in all the views that are using the modified .js files to force the reload of those files for all clients. For example:
<script type="text/javascript" src="${context}/js/azkaban/view/project-permissions.js?v=1556216524794"></script>

abagshaw:
Whoops, forgot that - sorry! Fixed now.

ypadron-in:
LGTM thanks.
This PR actually fixes a bug where permissions not shown/rendered but checked in value were saved incorrectly.

PR #2326, opened 2019-09-09
Force projects to have at least one additional admin user or group

abagshaw:
By setting the `azkaban.forceMultipleUsers` flag to `true` (default is `false`). Administrators will be forced to add another user or group with admin privileges before scheduling or executing any jobs. This reduces the chance that a project is left inaccessible because the only admin leaves the company or for any reason no longer works on the project.

---

UX Changes:
If a user has admin privileges for the current project AND the current project has no other admin users or groups with admin privileges:

1. /manager?project=something, /manager?project=something&permissions, /manager?project=something&logs will all show an alert banner prompting the user to add another admin user or group.
2. Clicking _Execute Flow_ from /manager?project=something will result in a modal prompting the user to add another admin user or group (preventing the requested action).
3. Clicking _Run Job_ or _Run With Dependencies_ on a specific job in a project from /manager?project=something will result in the same modal as above (preventing the requested action).
3. /manager?project=something&flow=otherthing will show the same alert banner as described in 1. and clicking on _Schedule / Execute Flow_ at the top right will result in the same modal as described in 2. and 3.

---

NOTE: This blocking behaviour ONLY applies to admin users. Users who do not have admin privileges, but have execute privileges will still be able to execute the project even if the project has only one admin user. This is because a non-admin user has no ability to add other admin users so they can't do anything to rectify the situation, thus blocking them is pointless and will only create hassle for them.

---

Also, I didn't add any tests because it didn't look like any functionality/methods in the surrounding modified areas were tested either - but I can add some if necessary.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/Project.java
+++ b/azkaban-common/src/main/java/azkaban/project/Project.java
@@ -217,6 +217,17 @@ public class Project {
     return users;
   }
 
+  public List<String> getGroupsWithPermission(final Type type) {
+    final ArrayList<String> groups = new ArrayList<>();
+    for (final Map.Entry<String, Permission> entry : this.groupPermissionMap.entrySet()) {
+      final Permission perm = entry.getValue();
+      if (perm.isPermissionSet(type)) {
+        groups.add(entry.getKey());
+      }
+    }
+    return groups;
+  }
+
   public List<Pair<String, Permission>> getUserPermissions() {
     final ArrayList<Pair<String, Permission>> permissions =
         new ArrayList<>();
--- a/azkaban-common/src/main/java/azkaban/project/Project.java
+++ b/azkaban-common/src/main/java/azkaban/project/Project.java
@@ -207,7 +207,7 @@ public class Project {
   }
 
   public List<String> getUsersWithPermission(final Type type) {
-    final ArrayList<String> users = new ArrayList<>();
+    final List<String> users = new ArrayList<>();
     for (final Map.Entry<String, Permission> entry : this.userPermissionMap.entrySet()) {
       final Permission perm = entry.getValue();
       if (perm.isPermissionSet(type)) {
@@ -218,7 +218,7 @@ public class Project {
   }
 
   public List<String> getGroupsWithPermission(final Type type) {
-    final ArrayList<String> groups = new ArrayList<>();
+    final List<String> groups = new ArrayList<>();
     for (final Map.Entry<String, Permission> entry : this.groupPermissionMap.entrySet()) {
       final Permission perm = entry.getValue();
       if (perm.isPermissionSet(type)) {
@@ -229,8 +229,7 @@ public class Project {
   }
 
   public List<Pair<String, Permission>> getUserPermissions() {
-    final ArrayList<Pair<String, Permission>> permissions =
-        new ArrayList<>();
+    final List<Pair<String, Permission>> permissions = new ArrayList<>();
 
     for (final Map.Entry<String, Permission> entry : this.userPermissionMap.entrySet()) {
       permissions.add(new Pair<>(entry.getKey(), entry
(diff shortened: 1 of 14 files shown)


azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -217,6 +217,17 @@ public boolean hasGroupPermission(final User user, final Type type) {
     return users;
   }
 
+  public List<String> getGroupsWithPermission(final Type type) {
+    final ArrayList<String> groups = new ArrayList<>();
```
djaiswal83: Use " List<String>"

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -119,6 +123,7 @@ public int compare(final Flow f1, final Flow f2) {
   private boolean lockdownCreateProjects = false;
   private boolean lockdownUploadProjects = false;
   private boolean enableQuartz = false;
+  private boolean forceMultipleUsers;
```
djaiswal83: Please initialize to some value.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1976,6 +1994,19 @@ private void lockFlowsForProject(final Project project, final List<String> locke
...
 
+  private void setNeedsMoreUsersIfAppropriate(final Page page, final Project project, final Session session) {
+    // See if the project has only one admin user with no admin groups.
+    // If it does AND the forceMultipleUsers flag is set AND the current user has the ability to add
+    // an admin, let's force the user to do so before allowing any execution or scheduling of flows.
+
+    boolean hasEnoughUsersGroups = project.getUsersWithPermission(Type.ADMIN).size() > 1
+                              || project.getGroupsWithPermission(Type.ADMIN).size() > 0
+                              || !hasPermission(session.getUser(), Type.ADMIN)
+                              || !this.forceMultipleUsers;
```
djaiswal83: Move this.forceMultipleUsers to the front, if the config is off, we fallback to existing logic.

abagshaw: 😄

ypadron-in:
Is this change also preventing project admins from scheduling/executing via apis? It would make little sense to check only in the UI while leaving the api open.
I disagree with the addition of this feature because if the only admin of a project leaves the company Azkaban admins are still capable of managing those projects. Additionally, before we do this I would like know what's the impact of this feature, how many projects with a single admin we have today. Finally, some projects are created only for the purpose of experimentation it might not make sense to force the owners of those projects to add other users in order to run their experiments.

abagshaw:
> Is this change also preventing project admins from scheduling/executing via apis? It would make little sense to check only in the UI while leaving the api open.

Yes it only prevents scheduling and executing through the UI. I thought the API might need to be left open because this might break some automated build processes (i.e. existing build processes that don't involve a user action but use the API)...but maybe I'm wrong?

> I disagree with the addition of this feature because if the only admin of a project leaves the company Azkaban admins are still capable of managing those projects. 

I agree - I was just going off of what the ticket said and the ticket requested this feature as apparently many projects have been left Admin-less requiring Azkaban admins to intervene. I just assumed the ticket creator had done his/her research.

> Additionally, before we do this I would like know what's the impact of this feature, how many projects with a single admin we have today. 

Sure - I don't think I have the expertise to do this though 😄 

> Finally, some projects are created only for the purpose of experimentation it might not make sense to force the owners of those projects to add other users in order to run their experiments.

This is definitely a drawback, although it really isn't that hard to add another admin. The same kind of policy is already included in MPs where mint forces you to add at least 2 owners...so there is a precedent for this.

PR #2330, opened 2019-09-15
Don't use synchronous Ajax call to get job logs

ypadron-in:
A synchronous Ajax call causes the UI thread to block until the Ajax call completes. While the UI thread is blocked, the page is unresponsive as that is the thread in charge of responding to user input. For large job logs, several Ajax calls are needed, which if made in synchronous in a loop cause the page to become unresponsive for too long.
The solution to the problem: use asynchronous Ajax calls.

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobdetailspage.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobdetailspage.vm
@@ -24,7 +24,7 @@
   <script type="text/javascript" src="${context}/js/jquery.twbsPagination.min.js"></script>
 
   <script type="text/javascript" src="${context}/js/azkaban/util/ajax.js"></script>
-  <script type="text/javascript" src="${context}/js/azkaban/model/job-log.js"></script>
+  <script type="text/javascript" src="${context}/js/azkaban/model/job-log.js?v=1568515440"></script>
   <script type="text/javascript" src="${context}/js/azkaban/view/job-details.js"></script>
   <script type="text/javascript">
     var contextURL = "${context}";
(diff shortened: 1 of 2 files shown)


abagshaw:
One small thing unrelated to this change - if we aren't worried about supporting old browsers i.e. IE :) then ES6 support can be assumed.

With ES6, [this](https://github.com/azkaban/azkaban/blob/071bfe87a4cc2c4778819cefa620de4fcac4cfff/azkaban-web-server/src/web/js/azkaban/model/job-log.js#L37) is no longer best practice. Better practice to use an arrow function [here](https://github.com/azkaban/azkaban/blob/071bfe87a4cc2c4778819cefa620de4fcac4cfff/azkaban-web-server/src/web/js/azkaban/model/job-log.js#L39) (i.e. `(data) => {.....}`) and then use `this` instead of `self` inside the arrow function. Arrow functions automatically inherit `this` from their parent scope.

ypadron-in:
> One small thing unrelated to this change but relevant to the file - if we aren't worried about supporting old browsers i.e. IE :) then ES6 support can [probably be assumed](https://caniuse.com/#feat=arrow-functions).
> 
> With ES6, [this](https://github.com/azkaban/azkaban/blob/071bfe87a4cc2c4778819cefa620de4fcac4cfff/azkaban-web-server/src/web/js/azkaban/model/job-log.js#L37) trick is unnecessary. Better practice to use an arrow function [here](https://github.com/azkaban/azkaban/blob/071bfe87a4cc2c4778819cefa620de4fcac4cfff/azkaban-web-server/src/web/js/azkaban/model/job-log.js#L39) (i.e. `(data) => {.....}`) and then use `this` instead of `self` inside the arrow function. Arrow functions automatically inherit `this` from their parent scope.
> 
> But...if ES6 support cannot be relied on, then the pattern currently in place is just fine.

I agree on incorporating ES6 features. Will revisit this later.

PR #2339, opened 2019-09-27
Fix job link urls of embedded jobs in Execute dialog.

ypadron-in:
Urls to embedded jobs in the Execute dialog's "Flow View" tab are not generated correctly due to the way parent flows are determined. One way to reproduce the issue is to use the option "Expand All Flows" followed by "Open Job in New Window" on any embedded job, we will get a page with content "Job XXX not found".

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/exflow.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/exflow.js
@@ -637,7 +637,6 @@ $(function () {
   }
 
   var successHandler = function (data) {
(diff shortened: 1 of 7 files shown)


azkaban-web-server/src/web/js/azkaban/view/exflow.js
```diff
@@ -558,95 +558,6 @@ var logUpdaterFunction = function () {
   }
 }
 
-var exNodeClickCallback = function (event) {
```
ypadron-in: These functions are not bound to any events.

PR #2342, opened 2019-10-01
Use FQN for Kerberos principal only.

djaiswal83:


ypadron-in:
Could you elaborate the motivation behind this change?

djaiswal83:
@ypadron-in, yes I was a bit lazy there. The motive of the original patch which added the FQN with email alias was to use it for Kerberos principal. However, due to my lack of knowledge, I made the changes where I was not supposed to make them, for eg, SSL auth still needs only userToProxy and not the FQN.
This patch reverts back the unnecessary changes to userToProxy.

ypadron-in:
How did we test this?

PR #2313, opened 2019-08-28
[FEATURE] Add Dependency Ramp Manager to support System Dependency Ra…

kxu1026:
…mpping Feature

PR #2343, opened 2019-10-01
[BUGFIX] The generate instance function name is getInstance instead o…

kxu1026:
Fix the bug in previous 'Add Ramp Manager' Feature. The wired hadoopSecurityManagerClass method is getInstance instead of createInstance.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -125,7 +125,7 @@ public class HadoopJobUtils {
     HadoopSecurityManager hadoopSecurityManager = null;
 
     try {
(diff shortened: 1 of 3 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -125,7 +125,7 @@ public static HadoopSecurityManager loadHadoopSecurityManager(Props props, Logge
     HadoopSecurityManager hadoopSecurityManager = null;
 
     try {
-      Method getInstanceMethod = hadoopSecurityManagerClass.getMethod("createInstance", Props.class);
```
ypadron-in: I believe we need to revert this change in some other places. There are more occurrences of `hadoopSecurityManagerClass.getMethod` in https://github.com/azkaban/azkaban/pull/2313

PR #2349, opened 2019-10-11
[MySql] partition execution_{flows, jobs, logs} by exec_id to improve performance and simplify retention

ericsun2:
execution_logs, execution_flows and execution_jobs are the biggest and heavily-accessed tables in Azkaban's MySql backend, DBA has to manually clean up the old records from these tables.

There are 2 tuning aspects we want to tune here, in order to improve the performance without changing its Java backend service.

- Identify the unused indices and remove them
- Partition the PK by range(exec_id)

The partition split can be automated with a monthly cron job in Azkaban Web Server, or it can still be handled by MySql DBA. Partitioning will reduce the maintenance time and effort, since naturally "exec_id" grows monotonically along with "upload_time" together. So the old logs are always kept in the partitions with smaller exec_id values.

- The BTree of PK will be much smaller and shorter because each partition will contain much less entries.
- The archival/pruning operation will be faster and easier because only the oldest partition will be affected. DBA can even run `TRUNCATE PARTITION` now.

execution_logs optimization

A script is provided in update.execution_logs.3.1.sql to migrate the data from existing execution_{logs,flows,jobs} to the partitioned tables.

Each partition will have 5~20M records for 5M exec_id(s), if on average each flow contains 20 jobs. In the future, DBA can split/reorganize the max partition to new range. Please see the sample DDL in update.execution_logs.3.1.sql

### DIFF
--- a/azkaban-sql/src/sql/create.execution_flows.sql
+++ b/azkaban-sql/src/sql/create.execution_flows.sql
@@ -12,9 +12,18 @@ CREATE TABLE execution_flows (
 	enc_type TINYINT,
 	flow_data LONGBLOB,
 	PRIMARY KEY (exec_id)
+)
+PARTITION BY RANGE (exec_id) (
+  PARTITION P004M VALUES LESS THAN (05000000),
+  PARTITION P009M VALUES LESS THAN (10000000),
+  PARTITION P014M VALUES LESS THAN (15000000),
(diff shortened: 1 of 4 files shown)


djaiswal83:
@ericsun2 there are conflicts which need to be resolved. Can you please do that?

PR #2350, opened 2019-10-11
Add Expand All button to job lists

abagshaw:
Add Expand All button to job lists on both graph job sidebar and execution job list. When clicked, that job and all its sub jobs will be expanded recursively.

Note: The button only appears on jobs who have at least one child who has at least one child :smile: In other words, the button only appears on jobs containing more than one level of nested jobs.

Examples:
https://drive.google.com/file/d/1jhCoFm4INzXV9SUx8Sld_57LEsdAsGZi/view?usp=sharing
https://drive.google.com/file/d/1Fq9hRb1qFSC0ftnHEYTg05W-RqGEBAHW/view?usp=sharing

### DIFF
--- a/azkaban-web-server/src/web/js/azkaban/view/job-list.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/job-list.js
@@ -198,16 +198,20 @@ azkaban.JobListView = Backbone.View.extend({
         $(expandDiv).addClass("expandarrow glyphicon glyphicon-chevron-down");
         $(a).append(expandDiv);
 
-        // Add the double down expand all
-        var expandAllDiv = document.createElement("div");
-        $(expandAllDiv).addClass("expandallarrow");
-        var firstArrow = document.createElement("span");
-        $(firstArrow).addClass("glyphicon glyphicon-chevron-down");
-        var secondArrow = document.createElement("span");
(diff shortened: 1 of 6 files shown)


ypadron-in:
Thank you for adding this feature!

azkaban-web-server/src/web/js/azkaban/view/job-list.js
```diff
@@ -240,6 +247,15 @@ azkaban.JobListView = Backbone.View.extend({
     evt.stopImmediatePropagation();
   },
 
+  handleToggleMenuExpandAll: function (evt) {
```
ypadron-in: Is it possible to implement the opposite: Collapse All?

abagshaw: I was thinking about that and, yes it would be possible but I don't think it's desirable.

1. You can always collapse a menu by clicking the single arrow. If you do re-open that menu all menus will still be expanded though.

2. A menu may be in a partially expanded state and you may still want to expand the remainder of the unexpanded children, so changing expand all to collapse all upon opening would remove that ability.

ypadron-in: Talked offline, we'll do only "expand all" for now. If it's confusing for users we'll revisit this.

PR #2356, opened 2019-10-16
Fix ajax call detection for POST requests

abagshaw:
When a valid session could not be identified for a GET request, the existence of a `ajax` query string param was used to determine if a JSON error or if a rendered HTML login page should be returned.

For POST requests, a different method was used to determine if a request was an ajax call. Now both GET and POST requests use the same method for determining if a request is an ajax call:

IF the request has an `ajax` param (in query string for GET or in body for POST) OR if the request has the header `X-Requested-With: XMLHttpRequest`, the request is determined to be an ajax call.

---

Unrelated: this PR also fixes incorrect cURL sample requests in the docs where an endpoint that should be accessed through GET listed a sample cURL request that actually defaulted to POST.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
@@ -121,7 +121,7 @@ public abstract class LoginAbstractAzkabanServlet extends AbstractAzkabanServlet
 
       handleGet(req, resp, session);
     } else {
-      if (hasParam(req, "ajax")) {
+      if (isAjaxCall(req)) {
         final HashMap<String, String> retVal = new HashMap<>();
         retVal.put("error", "session");
         this.writeJSON(resp, retVal);
@@ -481,6 +481,10 @@ public abstract class LoginAbstractAzkabanServlet extends AbstractAzkabanServlet
   }
 
   protected boolean isAjaxCall(final HttpServletRequest req) {
+    if (hasParam(req, "ajax")) {
+      return true;
+    }
+
     final String value = req.getHeader("X-Requested-With");
     if (value != null) {
       logger.info("has X-Requested-With " + value);
--- a/docs/ajaxApi.rst
+++ b/docs/ajaxApi.rst
@@ -1272,7 +1272,7 @@ Here's a curl command sample:
 
 .. code-block:: guess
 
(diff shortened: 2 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -121,7 +121,7 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
 
       handleGet(req, resp, session);
     } else {
-      if (hasParam(req, "ajax")) {
```
ypadron-in: I agree that it's better this way, but this is a breaking change on our api: anyone using our apis outside the website without specifying the header 'X-Requested-With' (not enforced) in POST requests will start getting json instead of html as a response when session expires which can happen often. Api versioning is the way to conduct these type of changes but unfortunately we haven't planned for that yet. I don't know what the impact of this change is but I prefer not to risk it.

docs/ajaxApi.rst
```diff
@@ -1272,7 +1272,7 @@ Here's a curl command sample:
 
 .. code-block:: guess
 
-   curl -k --data "session.id=34ba08fd-5cfa-4b65-94c4-9117aee48dda&ajax=pauseFlow&execid=303" https://localhost:8443/executor
+   curl -k --get --data "session.id=34ba08fd-5cfa-4b65-94c4-9117aee48dda&ajax=pauseFlow&execid=303" https://localhost:8443/executor
```
ypadron-in: Actually for 'pauseFlow' and 'resumeFlow' endpoints what we need to change is the endpoint documentation not the example. These endpoints change state in the server so they should be POST (to be more precise PUT but Azkaban doesn't support it).

azkaban-web-server/src/main/java/azkaban/webapp/servlet/LoginAbstractAzkabanServlet.java
```diff
@@ -121,7 +121,7 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
...
+      if (isAjaxCall(req)) {
```
ypadron-in: I agree that it's better this way, but this is a breaking change on our api: anyone using our apis outside the website without specifying the header 'X-Requested-With' (not enforced) in POST requests will start getting json instead of html as a response when session expires which can happen often. Api versioning is the way to conduct these type of changes but unfortunately we haven't planned for that yet. I don't know what the impact of this change is but I prefer not to risk it.

PR #2363, opened 2019-10-24
Add support for thin archives on web and executor server

abagshaw:
Thin archives allow project zips to include a `/app-meta/startup-dependencies.json` file that specifies JAR dependencies that the project's flows need to be able to run. These dependencies no longer have to be bundled in the zip, significantly reducing its size.

In addition, support has been added for `ProjectValidatorCacheable`, a new specification of a validator that can generate a `cacheKey` for a thin project. Files that have been deemed to be VALID or needing to be REMOVED can be indicated as such in a new `validated_dependencies` db table. Thus files that have already undergone validation do not need to be revalidated on future uploads.

See comment here for detailed description of thin archive processing on web server and how cache keys (validation keys) work: https://github.com/azkaban/azkaban/blob/5f2c98a3d2794ca457790a6dfc70aba3c4b92575/azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java#L48-L77

---

The following configuration parameters have been added (and must be set for both the web AND executor server to enable thin archives):

1. `azkaban.storage.cache.dependency.enabled` (default: `true`), if set to `false` then dependencies downloaded through CachedHttpFileSystem will always be pulled from the origin and never cached. The `azkaban.storage.cache.dependency_root.uri` configuration parameter will be ignored (and can be omitted) if this is set to `false`.
2. `azkaban.storage.origin.dependency_root.uri` The base URI for the origin where dependencies should be downloaded from. NOTE: Scheme should be `chttp://` because this goes through the CachedHttpFileSystem. The scheme will be rewritten to `http://` before the request is made.
3. `azkaban.storage.cache.dependency_root.uri` The base URI for the cache location where dependencies will be cached. If storage type is `HDFS` this should start with `hdfs://`. If storage type is `LOCAL_HADOOP` this should start with `file:///`

Example configuration:
```
azkaban.storage.origin.dependency_root.uri=chttp://www.example.com/path/to/dependencies/
azkaban.storage.cache.dependency_root.uri=hdfs://myhdfs.company.com:9000/dependency/cache
```

NOTE: If all of the above configuration parameters are omitted, then thin archive support will be disabled (the web server will prevent uploads of thin archives)


az-core/src/main/java/azkaban/utils/HashUtils.java
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * @param raw raw hash string
+   * @return lowercase raw hash string
+   * @throws InvalidHashException if the hash is invalid for any of the reasons described above.
+   */
+  public String sanitizeHashStr(final String raw) throws InvalidHashException {
+    if (!raw.matches("^[a-zA-Z0-9]*$")) {
+      throw new InvalidHashException(
+          String.format("Hash %s has invalid characters. Should be only alphanumeric.", raw));
+    } else if (this.type.equals("MD5") && raw.length() != 32) {
```
bxji: nitpick: can you make "MD5" a final constant variable instead, and make 32 a constant as well? it may make the code a little easier to read.

e.g.

private final static int MD5_EXPECTED_LENGTH = 32

or something.

az-core/src/main/java/azkaban/utils/HashUtils.java
```diff
@@ -0,0 +1,133 @@
...
+      throw new InvalidHashException(
+          String.format("MD5 hash %s has incorrect length %d, expected 32", raw, raw.length()));
+    } else if (this.type.equals("SHA1") && raw.length() != 40) {
```
bxji: same thing here for "SHA1" and the number 40

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -0,0 +1,123 @@
...
+      return depValidationStatuses;
+    }
+
+    // Map of (filename + sha1) -> Dependency for resolving the dependencies already cached in the DB
+    // after the query completes.
+    Map<String, Dependency> hashAndFileNameToDep = new HashMap<>();
+
+    PreparedStatement stmnt = this.dbOperator.getDataSource().getConnection().prepareStatement(
+        "select file_name, file_sha1, validation_status from validated_dependencies where validation_key = ? and ("
+            + makeStrWithQuestionMarks(deps.size()) + ")");
```
bxji: Can you parameterize this query or use String.format() instead? it's a little hard to read.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -0,0 +1,123 @@
...
+    // All remaining dependencies in the hashToDep map should be marked as being NEW (because they weren't
+    // associated with any DB entry)
+    hashAndFileNameToDep.values().stream().forEach(d -> depValidationStatuses.put(d, FileValidationStatus.NEW));
+
+    return depValidationStatuses;
+  }
+
+  public void updateValidationStatuses(final Map<Dependency, FileValidationStatus> depValidationStatuses,
+      final String validationKey) throws SQLException {
+    if (depValidationStatuses.size() == 0) {
```
bxji: depValidationStatuses.isEmpty() would be a bit easier to read.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Inject
+  JdbcDependencyManager(final DatabaseOperator dbOperator) {
+    this.dbOperator = dbOperator;
+  }
+
+  public Map<Dependency, FileValidationStatus> getValidationStatuses(final Set<Dependency> deps,
+      final String validationKey) throws SQLException {
+    Map<Dependency, FileValidationStatus> depValidationStatuses = new HashMap<>();
+    if (deps.size() == 0) {
```
bxji: .isEmpty() could be easier to read.

azkaban-common/src/main/java/azkaban/storage/DatabaseStorage.java
```diff
@@ -64,7 +66,23 @@ public String put(final StorageMetadata metadata, final File localFile) {
...
+
+  @Override
+  public boolean dependencyFetchingEnabled() {
+    return false;
+  }
+
+  @Override
+  public String getDependencyRootPath() {
+    return null;
+  }
```
bxji: can you add some comments about these two methods? if they're not being used, you can just say so.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -98,25 +98,40 @@ public String put(final StorageMetadata metadata, final File localFile) {
...
-  public boolean delete(final String key) {
-    final File file = getFile(key);
+  public boolean dependencyFetchingEnabled() {
+    return false;
+  }
+
+  @Override
+  public String getDependencyRootPath() {
+    return null;
+  }
```
bxji: same thing here, please add some comments for why these are not being used(?)

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * handles downloading of dependencies. Used during the thin archive upload process, and upon starting the execution
+ * of a flow defined in a thin archive (to download necessary dependencies)
+ */
+@Singleton
+public class DependencyTransferManager {
+  public static final int MAX_DEPENDENCY_DOWNLOAD_TRIES = 2;
+
+  private static final int NUM_THREADS = 8;
+
```
bxji: Would it be possible to put these in a config file and read these values from there?

azkaban-common/src/main/java/azkaban/AzkabanCommonModuleConfig.java
```diff
@@ -52,8 +52,12 @@ public AzkabanCommonModuleConfig(final Props props) {
     this.storageImplementation = props.getString(AZKABAN_STORAGE_TYPE, this.storageImplementation);
     this.localStorageBaseDirPath = props
         .getString(AZKABAN_STORAGE_LOCAL_BASEDIR, this.localStorageBaseDirPath);
-    this.hdfsRootUri = props.get(AZKABAN_STORAGE_HDFS_ROOT_URI) != null ? props
-        .getUri(AZKABAN_STORAGE_HDFS_ROOT_URI) : null;
+    this.hdfsProjectRootUri = props.get(AZKABAN_STORAGE_HDFS_PROJECT_ROOT_URI) != null ? props
+        .getUri(AZKABAN_STORAGE_HDFS_PROJECT_ROOT_URI, true) : null;
+    this.cacheDependencyRootUri = props.get(AZKABAN_STORAGE_CACHE_DEPENDENCY_ROOT_URI) != null ? props
+        .getUri(AZKABAN_STORAGE_CACHE_DEPENDENCY_ROOT_URI, true) : null;
+    this.originDependencyRootUri = props.get(AZKABAN_STORAGE_ORIGIN_DEPENDENCY_ROOT_URI) != null ? props
```
jamiesjc: Can you use `azkaban.utils.Props#getUri(java.lang.String, java.net.URI)` directly?

azkaban-common/src/main/java/azkaban/HadoopModule.java
```diff
@@ -52,32 +59,111 @@
   @Inject
   @Provides
   @Singleton
-  public Configuration createHadoopConfiguration() {
+  @Named("hdfsConf")
+  public Configuration createHDFSConfiguration() {
```
jamiesjc: This method is for hadoop related configuration, not only for HDFS storage. Will it be better to keep the original name?

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
abagshaw: It would, but there is a precedent throughout our codebase to have thread count values in a non-configurable constant.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
abagshaw: It was renamed to `deleteProject`. Git diff is misleading.

azkaban-common/src/main/java/azkaban/HadoopModule.java
abagshaw: Actually it is only for HDFS storage. But thanks for bringing this up because I discovered I forgot to put the new `@Named` annotation on a `Configuration` injection in `HdfsAuth`. That is fixed now.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -0,0 +1,154 @@
...
+import static azkaban.utils.ThinArchiveUtils.*;
+
+/**
+ * handles downloading of dependencies. Used during the thin archive upload process, and upon starting the execution
+ * of a flow defined in a thin archive (to download necessary dependencies). Provides a thin layer of retry logic,
+ * checksum validation and parallelism on top of the base Storage::getDependency().
+ */
+@Singleton
+public class DependencyTransferManager {
+  public static final int MAX_DEPENDENCY_DOWNLOAD_TRIES = 2;
```
jamiesjc: Is it better to make it configurable instead of hardcoded to some value? Also would you consider some backoff time before the next retry?

azkaban-spi/src/main/java/azkaban/spi/FileValidationStatus.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+// REMOVED = validator removed this file, it is blacklisted
+// VALID = validator gave this file the green light - no modifications made, it's fine as is.
+// NEW = not yet processed by the validator
+public enum FileValidationStatus {
+  REMOVED(0), VALID(1), NEW(2);
```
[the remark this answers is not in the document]
abagshaw: Nope. Byte-Ray will never produce any warnings for VALID files.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
abagshaw: Both good ideas. `MAX_DEPENDENCY_DOWNLOAD_TRIES` is now configurable with the `azkaban.dependency.max.download.tries` property and an exponential delay is added for tries up to the maximum value.

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
```diff
@@ -17,76 +17,76 @@
...
   @Inject
   public LocalStorage(final AzkabanCommonModuleConfig config) {
-    this.rootDirectory = validateRootDirectory(
-        createIfDoesNotExist(config.getLocalStorageBaseDirPath()));
+    this.rootDirectory = validateDirectory(createIfDoesNotExist(config.getLocalStorageBaseDirPath()));
   }
 
-  private static File createIfDoesNotExist(final String baseDirectoryPath) {
-    final File baseDirectory = new File(baseDirectoryPath);
+  private static File createIfDoesNotExist(final File baseDirectory) {
```
djaiswal83: this is only called once at line 58. What is the motive of making this additional function?

azkaban-common/src/main/java/azkaban/utils/DependencyTransferException.java
```diff
@@ -0,0 +1,34 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.utils;
+
+/**
+ * Indicates that a dependency failed to successfully upload or download.
+ */
+public class DependencyTransferException extends RuntimeException {
```
djaiswal83: If you are adding this exception and others like it, better make the best use of them by adding some additional info specific to this exception.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.commons.io.IOUtils;
+
+import static azkaban.utils.ThinArchiveUtils.*;
+
+/**
+ * handles downloading of dependencies. Used during the thin archive upload process, and upon starting the execution
```
djaiswal83: Nit : H capital :P

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -0,0 +1,168 @@
...
+
+  public final int dependencyMaxDownloadTries;
+
+  private final Storage storage;
+
+  @Inject
+  public DependencyTransferManager(final Props props, final Storage storage) {
+    this.storage = storage;
+
+    this.dependencyMaxDownloadTries = props.getInt(Constants.ConfigurationKeys.AZKABAN_DEPENDENCY_MAX_DOWNLOAD_TRIES, 2);
```
djaiswal83: Break into two lines after "="

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
```diff
@@ -0,0 +1,175 @@
...
+          finalDependencies.add(localDepPath);
+        }
+      }
+
+      String localAndRemoteJarSpec = String.join(",", finalDependencies);
+      log.debug("replaceLocalPathsWithStoragePaths: localAndRemoteJarSpec: " + localAndRemoteJarSpec);
+      return localAndRemoteJarSpec;
+    } catch (IOException | InvalidHashException e) {
+      // If something goes wrong, swallow the error and just return the original string.
+      log.warn("Error while opening and parsing startup dependencies file " + startupDependenciesFile.getAbsolutePath());
```
djaiswal83: correct the formatting

azkaban-db/src/main/sql/create.project_versions.sql
```diff
@@ -1,13 +1,14 @@
...
+  project_id           INT           NOT NULL,
+  version              INT           NOT NULL,
+  upload_time          BIGINT        NOT NULL,
+  uploader             VARCHAR(64)   NOT NULL,
+  file_type            VARCHAR(16),
+  file_name            VARCHAR(128),
+  md5                  BINARY(16),
+  num_chunks           INT,
+  resource_id          VARCHAR(512)  DEFAULT NULL,
+  startup_dependencies MEDIUMBLOB    DEFAULT NULL,
```
djaiswal83: Can you provide AJAX API to fetch the cooked dependencies?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -199,25 +208,65 @@ private File createTempDir(final ProjectDirectoryMetadata proj) {
...
+
     try {
       checkState("zip".equalsIgnoreCase(projectFileHandler.getFileType()));
       final File zipFile = requireNonNull(projectFileHandler.getLocalFile());
       final ZipFile zip = new ZipFile(zipFile);
       Utils.unzip(zip, dest);
-      projectDirectoryMetadata.setDirSizeInByte(calculateDirSizeAndSave(dest));
+
+      // Download all startup dependencies. If this is a fat archive, it will be an empty set (so we won't download
+      // anything). Note that we are getting our list of startup dependencies from the DB, NOT from the startup-dependencies.json
```
djaiswal83: wrap over this line.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -199,25 +208,65 @@ private File createTempDir(final ProjectDirectoryMetadata proj) {
...
+   * @param proj project to download
+   * @param execId execution id number
+   * @param folder root of unzipped project
+   * @param dependencies the set of dependencies to download
+   */
+  private void downloadAllDependencies(final ProjectDirectoryMetadata proj, final int execId, final File folder,
+      final Set<Dependency> dependencies) {
+    // Download all of the dependencies from storage
+    LOGGER.info("Downloading {} JAR dependencies... Project: {}, ExecId: {}", dependencies.size(), proj, execId);
+    Set<DependencyFile> depFiles = dependencies.stream().map(d -> getDependencyFile(folder, d)).collect(Collectors.toSet());
```
djaiswal83: reformat this line for better readability

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -85,6 +85,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static azkaban.ServiceProvider.*;
```
djaiswal83: avoid using .* and add specific classes instead.

azkaban-spi/src/main/java/azkaban/spi/Dependency.java
```diff
@@ -0,0 +1,108 @@
...
+  public Dependency(final Map<String, String> m) throws InvalidHashException {
+    this(m.get("file"), m.get("destination"), m.get("type"), m.get("ivyCoordinates"), m.get("sha1"));
+  }
+
+  /**
+   * Make a copy of this dependency
+   *
+   * @return a copy of this dependency
+   */
+  public Dependency makeCopy() {
```
djaiswal83: use name like copy or clone instead.

azkaban-spi/src/main/java/azkaban/spi/Dependency.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  /**
+   * Make a new DependencyFile with the same details as this dependency
+   *
+   * @param f file for DependencyFile
+   * @return the new DependencyFile
+   */
+  public DependencyFile makeDependencyFile(final File f) {
```
djaiswal83: f is too generic, use meaningful param name

azkaban-spi/src/main/java/azkaban/spi/Dependency.java
```diff
@@ -0,0 +1,108 @@
...
+      // This should never happen because we already validated the hash when creating this dependency
+      throw new RuntimeException("InvalidHashException when copying dependency.");
+    }
+  }
+
+  // it makes much more sense for the getter to be getFileName vs getFile, but in the startup-dependencies.json
+  // spec we expect the property to be "file" not "fileName" so we have to annotate this to tell the JSON serializer
+  // to insert it with "file", instead of assuming the name based on the name of the getter like it usually does.
+  @JsonProperty("file")
+  public String getFileName() { return fileName; }
```
djaiswal83: use keyword "this" in these methods before member names.

cached-http-filesystem/src/main/java/azkaban/cachedhttpfilesystem/CachedHttpFileSystem.java
```diff
@@ -0,0 +1,297 @@
...
+ *
+ * FOLDER STRUCTURE ON CACHE EXAMPLE:
+ * CachedHttpFileSystem was initialized with URI: chttp://www.example.com/jars/
+ * CACHE_ROOT_URI is set to: hdfs://localhost:9000/dependencies/
+ *
+ * .open() is called on CachedHttpFileSystem with URI: chttp://www.example.com/jars/some/lib/coollib-1.0.0.jar
+ * the URI for this file on the cache will be: hdfs://localhost:9000/dependencies/some/lib/coollib-1.0.0.jar
+ * and the necessary directories will be created in order to put the file there.
+ */
+public class CachedHttpFileSystem extends FileSystem {
```
djaiswal83: code style recheck
use final wherever applicable.

cached-http-filesystem/src/main/java/azkaban/cachedhttpfilesystem/CachedHttpFileSystem.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  @Override
+  public String getScheme() {
+    return "chttp";
+  }
+
+  @Override
+  public URI getUri() {
+    return uri;
```
djaiswal83: this.uri

cached-http-filesystem/src/main/java/azkaban/cachedhttpfilesystem/CachedHttpFileSystem.java
```diff
@@ -0,0 +1,297 @@
...
+
+  @Override
+  public FSDataInputStream open(Path path, int bufferSize) throws IOException {
+    URI relativeURI = this.uri.relativize(path.toUri());
+    if (relativeURI.isAbsolute()) {
+      throw new IOException("Path must be relative or have same prefix as root origin URI.");
+    }
+
+    URI resolvedOriginURI = this.rootOriginURI.resolve(relativeURI);
+    if (!cachingEnabled) {
```
djaiswal83: use "this" wherever applicable.

cached-http-filesystem/src/main/java/azkaban/cachedhttpfilesystem/CachedHttpFileSystem.java
```diff
@@ -0,0 +1,297 @@
...
+      }
+
+      // Return input stream from cache
+      return this.cacheFS.open(resolvedCachePath, bufferSize);
+    }
+  }
+
+  @Override
+  public FSDataOutputStream create(Path path, FsPermission fsPermission,
+      boolean b, int i, short i1, long l,
```
djaiswal83: short s

az-core/src/main/java/azkaban/utils/HashUtils.java
```diff
@@ -0,0 +1,137 @@
...
+    MessageDigest digest = null;
+    try {
+      digest = MessageDigest.getInstance(getName());
+    } catch (final NoSuchAlgorithmException e) {
+      // Should never get here.
+    }
+    return digest;
+  }
+
+  public String getHashStr(final String str) {
```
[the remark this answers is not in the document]
abagshaw: If you follow the method calls it eventually relies on the type of the ENUM.

az-core/src/main/java/azkaban/utils/HashUtils.java
```diff
@@ -0,0 +1,137 @@
...
+    final byte[] buffer = new byte[BYTE_BUFFER_SIZE];
+
+    int num = 0;
+    do {
+      num = blobStream.read(buffer);
+    } while (num > 0);
+
+    bStream.close();
+
+    return digest.digest();
```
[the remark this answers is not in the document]
abagshaw: `getDigest()` never returns `null`.

az-core/src/main/java/azkaban/utils/HashUtils.java
```diff
@@ -0,0 +1,137 @@
...
+    return digest;
+  }
+
+  public String getHashStr(final String str) {
+    return bytesHashToString(getHashBytes(str)).toLowerCase();
+  }
+
+  public byte[] getHashBytes(final String str) {
+    final MessageDigest digest = getDigest();
+    digest.update(str.getBytes(UTF_8));
```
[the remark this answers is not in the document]
abagshaw: `getDigest()` never returns `null`.

cached-http-filesystem/src/main/java/azkaban/cachedhttpfilesystem/CachedHttpFileSystem.java
abagshaw: This was copied directly from: https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/http/AbstractHttpFileSystem.java

Going to leave as is.

az-core/src/main/java/azkaban/utils/Props.java
```diff
@@ -625,13 +625,22 @@ public double getDouble(final String name) {
   }
 
   /**
-   * Returns the uri representation of the value. If the value is null, then the default value is
-   * returned. If the value isn't a uri, then a IllegalArgumentException will be thrown.
+   * Returns the uri representation of the value. If the value is null, then an
+   * UndefinedPropertyException will be thrown. If the value isn't a uri, then an
```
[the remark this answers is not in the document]
abagshaw: Which exception are you talking about? Both exceptions mentioned in the comment are thrown `IllegalArgumentException` and `UndefinedPropertyException` (both extend `RuntimeException`) could potentially be thrown.

az-core/src/test/java/azkaban/utils/HashUtilsTest.java
```diff
@@ -0,0 +1,119 @@
+/*
```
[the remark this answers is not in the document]
abagshaw: :smile:

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -15,6 +15,8 @@
  */
 package azkaban.jobtype;
 
+import azkaban.spi.Storage;
```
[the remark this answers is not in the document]
abagshaw: They aren't. Good catch :smile:

On a separate PR we can run intellij's optimize imports and it can clean up a gazillion unused imports across the code base. I didn't want to do that for this PR though because it would have made the diff enormous.

azkaban-common/src/main/java/azkaban/AzkabanCommonModuleConfig.java
```diff
@@ -68,7 +70,15 @@ public String getLocalStorageBaseDirPath() {
     return this.localStorageBaseDirPath;
   }
 
-  public URI getHdfsRootUri() {
-    return this.hdfsRootUri;
+  public URI getHdfsProjectRootUri() {
+    return this.hdfsProjectRootUri;
   }
+
+  public URI getCacheDependencyRootUri() { return this.cacheDependencyRootUri; }
```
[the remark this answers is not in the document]
abagshaw: Precedent is to keep them on same line - going to keep it that way.

azkaban-common/src/main/java/azkaban/AzkabanCommonModuleConfig.java
```diff
@@ -68,7 +70,15 @@ public String getLocalStorageBaseDirPath() {
...
+
+  public URI getOriginDependencyRootUri() {
+    return this.originDependencyRootUri;
+  }
+
+  public boolean getDependencyCachingEnabled() { return this.dependencyCachingEnabled; }
```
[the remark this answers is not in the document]
abagshaw: Precedent is to keep them on same line - going to keep it that way.

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * Handles processing of uploaded Thin Archives to the web server.
+ *
+ * This class exposes one public method, validateThinProject() which provides
+ * the all the meat of the processing for Thin Archives. In summary it will:
+ *
+ * 1. Parse the startup-dependencies.json file
+ *
+ * 2. Generate a validation key from the project validators to use for querying the database. For any two projects
```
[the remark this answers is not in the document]
abagshaw: 😄

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
```diff
@@ -0,0 +1,284 @@
...
+    // Delete from the project untouched downloaded dependencies
+    untouchedDownloadedDeps.forEach(d -> d.getFile().delete());
+
+    // Add the cacheReport to the list of reports
+    reports.put(CACHED_VALIDATOR_REPORT_NAME, cacheReport);
+
+    return reports;
+  }
+
+  private void rewriteStartupDependencies(File startupDependenciesFile, Set<Dependency> finalDependencies) {
```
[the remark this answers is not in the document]
abagshaw: Yeah something went wrong I think - pretty sure they were here before and they are all gone now...

azkaban-common/src/main/java/azkaban/project/AzkabanProjectLoader.java
```diff
@@ -95,36 +111,45 @@
...
+        FlowLoaderUtils.cleanUpDir(folder);
         return reports;
       }
 
       // Upload the project to DB and storage.
-      persistProject(project, loader, archive, file, uploader);
+      File startupDependenciesOrNull = startupDependencies.exists() ? startupDependencies : null;
+      persistProject(project, loader, archive, folder, startupDependenciesOrNull, uploader);
+
+      if (startupDependencies.exists()) {
```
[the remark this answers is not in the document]
abagshaw: Personally I think it's more readable this way. Going to keep it like this.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -260,13 +262,14 @@ public void addProjectVersion(
...
       final File localFile,
+      final File startupDependencies,
       final String uploader,
       final byte[] md5,
       final String resourceId) throws ProjectManagerException {
 
     // when one transaction completes, it automatically commits.
     final SQLTransaction<Integer> transaction = transOperator -> {
-      addProjectToProjectVersions(transOperator, projectId, version, localFile, uploader, md5,
+      addProjectToProjectVersions(transOperator, projectId, version, localFile, startupDependencies, uploader, md5,
```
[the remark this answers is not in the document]
abagshaw: It's already less than max length.

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
```diff
@@ -17,102 +17,123 @@
 
 package azkaban.storage;
 
-import static com.google.common.base.Preconditions.checkArgument;
+import static azkaban.utils.StorageUtils.*;
```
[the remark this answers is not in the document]
abagshaw: I changed it to qualify the static methods with `StorageUtils.`

azkaban-common/src/main/java/azkaban/storage/LocalStorage.java
abagshaw: Readability - and it could potentially be used by other methods in the future.

azkaban-db/src/main/sql/create.project_versions.sql
abagshaw: Discussed offline - dropping. You can get the startup dependencies for a project version by downloading the project zip.

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
```diff
@@ -17,102 +17,123 @@
```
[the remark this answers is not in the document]
abagshaw: I think it's very important actually. These `Storage` implementations were originally designed to only be used for projects and they are hardcoded to only support project writing/fetching/deleting - thus they should be named as such.

azkaban-db/src/main/sql/upgrade.3.80.0.to.3.81.0.sql
```diff
@@ -0,0 +1,5 @@
+-- DB Migration from release 3.80.0 to 3.81.0
+-- PR #2363 Implements thin archive support. The thin archive json text must be stored as a blob along with the
+-- project version.
+--
+ALTER TABLE project_versions ADD COLUMN startup_dependencies MEDIUMBLOB DEFAULT NULL;
```
li-ygerchikov: Did you run this locally? I Am not sure if it's "ADD COLUMN" or just "ADD".

azkaban-db/src/main/sql/create.validated_dependencies.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE validated_dependencies (
+  file_name         VARCHAR(100),
```
li-ygerchikov: Why different from `file_name` above? Why not `VARCHAR(128)`?

azkaban-db/src/main/sql/create.validated_dependencies.sql
```diff
@@ -0,0 +1,7 @@
...
+  file_sha1         VARCHAR(40),
```
li-ygerchikov: Is this really a `"VAR"CHAR`? Isn't it fixed length string?

azkaban-db/src/main/sql/create.validated_dependencies.sql
```diff
@@ -0,0 +1,7 @@
...
+  validation_key    VARCHAR(40),
+  validation_status INT,
+  PRIMARY KEY (file_name, file_sha1, validation_key)
```
li-ygerchikov: So here is a thought. It looks like we always query this table by `validation_key` and an arbitrary number of (`file_name`, `file_sha1`) pairs. So if the table is sorted by `file_name` first as implemented, then we'll jump all over the table to fetch all needed rows. Even though each tow will (likely) be accessed by its PK, they are scattered across the table. We'll need to read many table blocks/pages and get (typically) one row from each.
In contrast, if we sorted the table by `validation_key` first, then all relevant rows for each query would always be located sequentially and together, and we'd likely need to read (and keep in buffer cache!) fewer pages, and sequential reads are faster on spinning disks.
Does it make sense? You may want to seek second opinion on this, but if it checks out, please consider changing your PK definition to `(validation_key, file_name, file_sha1)`.

azkaban-db/src/main/sql/upgrade.3.80.0.to.3.81.0.sql
li-ygerchikov: Also: do we want to create `validated_dependencies` here, or do you mean to run this *and* `create.validated_dependencies.sql`, separately?

abagshaw: It appeared that the precedent was that new tables were created separately and not included in the migration.

azkaban-db/src/main/sql/create.validated_dependencies.sql
abagshaw: `128` works fine too. Git diff is misleading - I didn't actually specify the `VARCHAR(128)` for `project_versions`. That was unmodified.

azkaban-db/src/main/sql/create.validated_dependencies.sql
abagshaw: True, both are fixed length.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -211,10 +211,16 @@
...
     public static final String AZKABAN_MAX_FLOW_RUNNING_MINS = "azkaban.server.flow.max.running.minutes";
 
+    // Maximum number of tries to download a dependency (no more retry attempts will be made after this many download failures)
+    public static final String AZKABAN_DEPENDENCY_MAX_DOWNLOAD_TRIES = "azkaban.dependency.max.download.tries";
+
     public static final String AZKABAN_STORAGE_TYPE = "azkaban.storage.type";
     public static final String AZKABAN_STORAGE_LOCAL_BASEDIR = "azkaban.storage.local.basedir";
     public static final String HADOOP_CONF_DIR_PATH = "hadoop.conf.dir.path";
-    public static final String AZKABAN_STORAGE_HDFS_ROOT_URI = "azkaban.storage.hdfs.root.uri";
+    public static final String AZKABAN_STORAGE_HDFS_PROJECT_ROOT_URI = "azkaban.storage.hdfs.root.uri"; // This really should be azkaban.storage.hdfs.project_root.uri
```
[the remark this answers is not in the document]
djaiswal83: The comment seems unchanged. Marking it unresolved.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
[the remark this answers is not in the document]
djaiswal83: Can you create a followup JIRA to track this?

azkaban-db/src/main/sql/upgrade.3.80.0.to.3.81.0.sql
abagshaw: Also yeah, whoops - the query is definitely wrong :)

PR #2364, opened 2019-10-25
Fix Upload Project Zip API command

ld000:
Old command returns a "Login error. Need username and password" error

### DIFF
--- a/docs/ajaxApi.rst
+++ b/docs/ajaxApi.rst
@@ -262,7 +262,7 @@ Here's a curl command sample:
 
 .. code-block:: guess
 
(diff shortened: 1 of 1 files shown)


docs/ajaxApi.rst
```diff
@@ -262,7 +262,7 @@ Here's a curl command sample:
 
 .. code-block:: guess
 
-   curl -k -i -H "Content-Type: multipart/mixed" -X POST --form 'session.id=e7a29776-5783-49d7-afa0-b0e688096b5e' --form 'ajax=upload' --form 'file=@myproject.zip;type=application/zip' --form 'project=MyProject;type/plain' https://localhost:8443/manager
```
ypadron-in: Works for me:
Screen Shot 2019-10-28 at 2 11 02 PM

But we do need to replace ";type/plain" with ";type=text/plain"

ld000: image

I replaced ";type/plain" with ";type=text/plain", but still error, any suggestion?

ypadron-in: I'm not able to reproduce this authentication error but let's make the change anyway as multipart/form-data is better suited here and more commonly supported.

PR #2369, opened 2019-10-31
[BUGFIX] The DB is setup to case-sensitive in Azkaban.

kxu1026:


azkaban-db/src/main/sql/create.ramp_exceptional_job_items.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE ramp_exceptional_job_Items (
+CREATE TABLE ramp_exceptional_job_items (
```
li-ygerchikov: Thank you! May I assume that it's never mentioned in code with uppercase 'I'? Could you confirm? Otherwise we'll be doing this again soon :) .

PR #2371, opened 2019-11-02
[docs] add gradle install link

brandonJY:
add gradle install link

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 
 ## Build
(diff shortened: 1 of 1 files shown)


abagshaw:
This project includes a [gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) which should auto-download and install the correct version of gradle when run. So directing users to the gradle install page might be confusing because there is no need for them to manually install gradle.

PR #2374, opened 2019-11-05
Allow oauth access token fetch by adding custom implemented oauth token provider

nkbuaa:
This change will allow user to implement customized oauth token provider in azkaban to generate access token to access  resources that needs authN. (Example use case: Azure storage).

Users need to set `azkaban.enable.oauth` to true in order to use it. (default to false)

Refactor `HadoopSecurityManager_H_2_0.registerCustomCredential` so it can be reused to register custom credential provider.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -339,18 +339,16 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
   }
 
   private void registerCustomCredential(final Props props, final Credentials hadoopCred, final
-  String userToProxy, final Logger jobLogger) {
+  String userToProxy, final Logger jobLogger, final String customCredentialProviderName) {
     String credentialClassName = "unknown class";
       try {
-        credentialClassName = props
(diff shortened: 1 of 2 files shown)


nkbuaa:
Created a issue on this: https://github.com/azkaban/azkaban/issues/2379

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -339,18 +339,16 @@ public boolean shouldProxy() {
   }
 
   private void registerCustomCredential(final Props props, final Credentials hadoopCred, final
-  String userToProxy, final Logger jobLogger) {
+  String userToProxy, final Logger jobLogger, String customCredentialProviderName) {
```
jamiesjc: Please add `final` keyword.

jamiesjc:
How do you plan to test this change?

nkbuaa:
> How do you plan to test this change?

The oauth flag is default to false so existing flows won't be impacted. We have our own implemented plugin ready in azure vm and will test from there.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
nkbuaa: Thanks. Will change it.

nkbuaa:
Addressed above comment

PR #2377, opened 2019-11-06
Add workflow id to mapreduce tags

hungj:
This is in preparation for Hadoop 2.10.0 upgrade - mechanism for passing workflow id from azkaban to Hadoop was changed (see YARN-9760). Now it uses yarn tags.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -83,6 +83,7 @@ public class HadoopJobUtils {
   public static final String MAPREDUCE_JOB_OTHER_NAMENODES = "mapreduce.job.hdfs-servers";
   // MapReduce config for mapreduce job tags
   public static final String MAPREDUCE_JOB_TAGS = "mapreduce.job.tags";
+  protected static final int APPLICATION_TAG_MAX_LENGTH = 100;
   // Root of folder in storage containing startup dependencies
   public static final String DEPENDENCY_STORAGE_ROOT_PATH_PROP = "dependency.storage.path.prefix";
   // Azkaban property for listing additional namenodes for delegation tokens
@@ -590,10 +591,13 @@ public class HadoopJobUtils {
     }
     if (props.containsKey(CommonJobProperties.PROJECT_NAME)
         && props.containsKey(CommonJobProperties.FLOW_ID)) {
-      keysAndValues[keys.length] = "workflowid:"
+      String constructedTag = "workflowid:"
           + props.get(CommonJobProperties.PROJECT_NAME)
           + HadoopConfigurationInjector.WORKFLOW_ID_SEPERATOR
           + props.get(CommonJobProperties.FLOW_ID);
+      constructedTag = constructedTag.substring(0,
+          Math.min(constructedTag.length(), APPLICATION_TAG_MAX_LENGTH));
+      keysAndValues[keys.length] = constructedTag;
     }
     Joiner joiner = Joiner.on(',').skipNulls();
     return joiner.join(keysAndValues);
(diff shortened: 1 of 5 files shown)


hungj:
Seems the travis-ci error is the same one reported here: https://github.com/azkaban/azkaban/pull/2259#discussion_r341422841

I ran the added test locally (TestHadoopJobUtilsConstructHadoopTags#testWorkflowIdTag), and it passed.

varunsaxena:
@hungj, the test failure doesn't seem related to code change.
But Travis CI is failing for PropsUtilsTest which is setting some props and calling some utils methods and the linked comment talks about race condition for config files. Are these related?
Can we reinvoke the build if the same tests fail?

hungj:
@varunsaxena  where did you see PropsUtilsTest failure? I think it succeeded: ```azkaban.utils.PropsUtilsTest
  SUCCESS (9 tests, 9 passed, 0 failed, 0 skipped)``` https://travis-ci.org/azkaban/azkaban/builds/608470782?utm_source=github_status&utm_medium=notification

The test failure I was referring to is FileWatcherTest:
```azkaban.user.FileWatcherTest > registerAndTake FAILED
    org.junit.ComparisonFailure: expected:<...-conf/azkaban-users-[file_watcher].xml> but was:<...-conf/azkaban-users-[test1_auto_reload_fail].xml>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at azkaban.user.FileWatcherTest.registerAndTake(FileWatcherTest.java:82)
azkaban.user.FileWatcherTest
  FAILURE (1 tests, 0 passed, 1 failed, 0 skipped)```

File: az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java [hunk 1]
```diff
@@ -581,12 +583,22 @@ public static String javaOptStringFromHadoopConfiguration(Configuration conf, St
...
     }
+    if (props.containsKey(CommonJobProperties.PROJECT_NAME)
+        && props.containsKey(CommonJobProperties.FLOW_ID)) {
+      String constructedTag = "workflowid:"
+          + props.get(CommonJobProperties.PROJECT_NAME)
+          + HadoopConfigurationInjector.WORKFLOW_ID_SEPERATOR
+          + props.get(CommonJobProperties.FLOW_ID);
+      constructedTag = constructedTag.substring(0,
+          Math.min(constructedTag.length(), APPLICATION_TAG_MAX_LENGTH));
+      keysAndValues[keys.length] = constructedTag;
```
li-ygerchikov: Does it matter that both the PROJECT_NAME and FLOW_ID are already included in the output, individually, since the three keys passed here are always:
        CommonJobProperties.EXEC_ID,
        CommonJobProperties.FLOW_ID,
        CommonJobProperties.PROJECT_NAME
Do we have to repeat this info one more time as a combined tag?

hungj: Hi @li-ygerchikov , yeah, we need a combined tag for this. The API in Hadoop 2.10 requires a tag of the form "workflowid:...", it is not able to combine the existing tags on the server side (the project_name and flow_id tags are azkaban specific anyway), hence we combine it here.

li-ygerchikov: OK, that makes sense, and it helps -- thanks.
The only other consideration would be whether we'd like to keep this *util* method completely generic, and instead add the specific tag that you need in setupHadoopJobProperties in azkaban/jobtype/AbstractHadoopJavaProcessJob.java?

hungj: @li-ygerchikov  that makes sense. Updated the PR.

hungj:
Seems I do not have write permissions on this repo @li-ygerchikov . Mind merging (and squashing) for me? Thanks!

li-ygerchikov:
@hungj done -- sorry for delay!

hungj:
Thanks Yuliy for review/commit!

PR #2383, opened 2019-11-18
[Logging] Fixed the md5 hash non-printable characters

svnarumugam:
Modified the md5 hash to print the Array of numbers instead of the non-printable characters when there is a mismatch. 

Added few files that are generated locally in the gitignore.

How to test in local?

set the property azkaban.storage.type=LOCAL in the azkaban.properties. This enabled the zip to be stored in LOCAL file system. After that deliberately added few bytes in the validateCheckSum `hash`. When I ran a flow, it failed with IllegalStateException indicating the mismatch between the hashes. 

The flow went through fine when there is a match in the md5 hashes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,9 +14,14 @@ vendor/
 **/velocity.log
 docs/_build/
 
+executor.port
+
 # local & temp directories are ignored so that they can be used to store temporary files such as local testing configurations.
 local/
 temp/
+conf/
+projects/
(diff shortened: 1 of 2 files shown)


svnarumugam:
Removed the unused imports and updated the PR. It turns out that adding test case for this would be tricky as we have modified the log message. Happy to get more thoughts on this.

PR #2386, opened 2019-11-19
[Optimization] Improve web server startup time by optimizing project loading

abagshaw:
Improve the web server startup time by optimizing the project, project permission and project flow loading.

Load all project permissions along with the projects with a `JOIN` and load all the flows through a single transaction.

Tested locally and on staging by adding new projects & flows and executing them.


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -814,19 +762,23 @@ public Flow fetchFlow(final Project project, final String flowId) throws Project
   }
 
   @Override
```
djaiswal83: This API is still useful in certain cases. Instead of deleting it, mark it @Deprecated.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -814,19 +762,23 @@ public Flow fetchFlow(final Project project, final String flowId) throws Project
...
     try {
-      flows = this.dbOperator
-          .query(ProjectFlowsResultHandler.SELECT_ALL_PROJECT_FLOWS, handler, project.getId(),
-              project.getVersion());
+      return this.dbOperator.transaction(transaction);
     } catch (final SQLException e) {
       throw new ProjectManagerException(
-          "Error fetching flows from project " + project.getName() + " version " + project
-              .getVersion(), e);
+          "Error fetching flows for " + projects.size() + " project(s).", e);
```
djaiswal83: This is a step back in terms of debugging. Did you try failing the query and found out what information you could get?

azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
```diff
@@ -40,97 +42,93 @@
...
-              final String jsonString = GZIPUtils.unGzipString(data, "UTF-8");
-              blobObj = JSONUtils.parseJSONFromString(jsonString);
-            } else {
-              final String jsonString = new String(data, "UTF-8");
-              blobObj = JSONUtils.parseJSONFromString(jsonString);
+
+        if (!projects.containsKey(id)) {
+          // This project is new!
+
+          final String name = rs.getString(2);
```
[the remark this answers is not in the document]
abagshaw: I see your point, but I don't really think that helps anything. The standard practice throughout the entire azkaban codebase is to not pull the column numbers into constants but leave them where they are and make sure the `getBoolean`, `getLong`, `getInt`, etc. calls are close to where the query is defined so it's very easy to see how these column numbers map to the query. I think this is a fine practice.

If you disagree with my disagreement 😄, I can submit a separate PR to refactor these column numbers into constants throughout the whole codebase so we maintain consistency. I don't like the idea of having it done one way in some places and different ways in others.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
abagshaw: If the query fails, the underlying SQL error will still bubble up and be printed. The fact that we are moving individual queries into a transaction naturally means we will loose some granularity in error catching - but that's just a tradeoff with the performance increase. In reality, any issues with the query will most likely be due to connection issues, malformed query, or another problem that is not specific to a single row for a project, so the loss of granularity doesn't matter much.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
abagshaw: Which API are you referencing, `getProjectPermissions` or `fetchAllProjectFlows`?

djaiswal83: fetchAllProjectFlows

abagshaw: Done - and I haven't marked it as depreciated, it works just fine.

azkaban-common/src/main/java/azkaban/project/JdbcProjectHandlerSet.java
```diff
@@ -40,97 +41,100 @@
...
-              final String jsonString = GZIPUtils.unGzipString(data, "UTF-8");
-              blobObj = JSONUtils.parseJSONFromString(jsonString);
-            } else {
-              final String jsonString = new String(data, "UTF-8");
-              blobObj = JSONUtils.parseJSONFromString(jsonString);
+
+        // If a project has multiple permissions - the project will be returned multiple times,
+        // one for each permission and we don't need to go through the work of reconstructing the
+        // project object if we've already seen it.
+        if (!projects.containsKey(id)) {
```
li-ygerchikov: No need to change anything now, but this could be simplified to just checking the last fetched ID against the previous one. I think we decided that the resultset would be sorted by ID due to it being [the prefix of] the PK on both tables. This would require an ORDER BY clause in the SELECT just because the order is not guaranteed otherwise, but that would be free given the implementation ...and everything we know :) .

abagshaw: True...it _could_ :)

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
```diff
@@ -218,11 +210,46 @@ public void testUpdatePermission() throws Exception {
...
+    Assert.assertTrue(returnedProject1.getUserPermission("testUser2").isPermissionSet(Permission.Type.EXECUTE));
+    Assert.assertTrue(returnedProject1.getGroupPermission("testGroup1").isPermissionSet(Permission.Type.ADMIN));
+
+    // Make sure proper permissions were returned for project2
+    Assert.assertEquals(returnedProject2.getUserPermissions().size(), 0);
+    Assert.assertTrue(returnedProject2.getGroupPermission("testGroup2").isPermissionSet(Permission.Type.READ));
+  }
+
+  private Project findProjectWithName(List<Project> projects, String name) {
+    return projects.stream().filter(p -> p.getName().equals(name)).findFirst().orElse(null);
```
li-ygerchikov: I'll see your `findFirst()` and I raise you `findAny()`. Rumor has it it's marginally faster *when* there is a difference -- i.e. stream has encounter order *and* may have more than one element. Nit. Ignore me.

azkaban-db/src/main/sql/create.projects.sql
```diff
@@ -12,4 +12,4 @@ CREATE TABLE projects (
 );
 
 CREATE INDEX project_name
-  ON projects (name);
+  ON projects (name);
```
li-ygerchikov: Line-end only diff? Revert?

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -198,9 +204,6 @@ void updateProjectProperty(Project project, Props props)
   Props fetchProjectProperty(int projectId, int projectVer, String propsName)
       throws ProjectManagerException;
 
-  List<Triple<String, Boolean, Permission>> getProjectPermissions(Project project)
```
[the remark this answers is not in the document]
abagshaw: Intellij code search? 😄 Compilation succeeds and tests pass?

azkaban-common/src/test/java/azkaban/project/JdbcProjectImplTest.java
abagshaw: 😄

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -135,14 +121,6 @@ public Project fetchProjectById(final int id) throws ProjectManagerException {
         throw new ProjectManagerException("No project with id " + id + " exists in db.");
       }
       project = projects.get(0);
-
-      // Fetch the user permissions
-      for (final Triple<String, Boolean, Permission> perm : fetchPermissionsForProject(project)) {
-        // TODO kunkun-tang: understand why we need to check permission not equal to 0 here.
```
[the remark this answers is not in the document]
abagshaw: We do - checking to see if `perm.getThird().toFlags() != 0` is essentially checking that the permission isn't useless (if it has a value of 0 then it has no permissions at all...). I checked the DB for both one production cluster and the staging cluster and neither had a single permission with a value of 0. Even if they did, nothing bad should happen...hopefully :)

azkaban-db/src/main/sql/create.projects.sql
abagshaw: Yeah that was an accident...oh well.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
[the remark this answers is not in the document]
li-ygerchikov: Um, I an slightly sceptical. Could some code from another repo reference it? We've just seen some issue when a method was eliminated only to discover that it was used indirectly by injecting its class as a dependency into some other code.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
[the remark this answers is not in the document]
li-ygerchikov: I see. Thanks. That could be a nice comment. Oh well :) .

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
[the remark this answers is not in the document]
abagshaw: I suppose that is possible - I did a global search internally and couldn't find anything that appeared to be depending on this method. We can discuss further in person.

PR #2389, opened 2019-11-26
Add credentials to hadoop UGI if hadoop security is not enabled.

li-leyang:
If hadoop security is off, hadoop proxy user needs OAuth token being available in its UGI credentials. This is to make sure that ABFS(Azure blob file system) driver can still get the token to access azure storage.


nkbuaa:
https://github.com/azkaban/azkaban/issues/2390

nkbuaa:
@jamiesjc @djaiswal83 This is part of OAuth support with Azure storage. It will be great if you guys can take a look.

nkbuaa:
> Where does it check HadoopSecurity is not enabled?

This is checked in [here](https://github.com/azkaban/azkaban/blob/master/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java#L98).

Specifically, the code  is doing the following check against hadoop config `core-site.xml:`
Hadoop security is enabled when `hadoop.security.authentication` != `simple`

li-leyang:
@djaiswal83  Hi Deepak, any comments about this?

PR #2391, opened 2019-11-27
Fix (again) flaky FileWatcherTest

li-ygerchikov:
See PR#2259. Since we watch the entire parent dir, *and* run tests in multiple parallel Gradle threads, at least on Travis, we catch *other* files being changed in FileWatcherTest. Since we do not expect that, test fails (sporadically, and only on Travis). Symptoms:
```
azkaban.user.FileWatcherTest > registerAndTake FAILED
    org.junit.ComparisonFailure: expected:<...-conf/azkaban-users-[file_watcher].xml> but was:<...-conf/azkaban-users-[test1_auto_reload_fail].xml>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at azkaban.user.FileWatcherTest.registerAndTake(FileWatcherTest.java:82)
azkaban.user.FileWatcherTest
  FAILURE (1 tests, 0 passed, 1 failed, 0 skipped)
```

### DIFF
--- a/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
+++ b/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
@@ -16,8 +16,7 @@
 
 package azkaban.user;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import com.google.common.io.Resources;
 import com.sun.nio.file.SensitivityWatchEventModifier;
@@ -73,14 +72,33 @@ public class FileWatcherTest {
 
     final WatchKey key = this.fileWatcher.take();
     final List<WatchEvent<?>> events = this.fileWatcher.pollEvents(key);
-    // depending on the OS & file system there may be at least 1 or 2 events even with 1 write()
-    assertThat(events.size()).isGreaterThanOrEqualTo(1);
 
-    for (WatchEvent<?> event : events) {
+    // Since we watch the entire parent dir, *and* run multiple tests in parallel, we catch *other*
+    // files being changed here. This assert checks that at least one of them is "ours"
+    // (*and* that at least one event is fired in the first place):
+    assertTrue(events.stream().anyMatch(event -> {
       @SuppressWarnings("unchecked") final Path name = ((WatchEvent<Path>) event).context();
       final String resolvedFileName = dir.resolve(name).toString();
-      assertEquals(PATH.toString(), resolvedFileName);
-    }
+      return PATH.toString().equals(resolvedFileName);
+    }));
+
+    // do we really need to convert Paths to Strings in order to compare?
+    assertTrue(events.stream().anyMatch(event -> {
+      @SuppressWarnings("unchecked") final Path name = ((WatchEvent<Path>) event).context();
+      final Path resolvedFile = dir.resolve(name);
+      return PATH.equals(resolvedFile);
+    }));
+
+    // compact but throws info on compilation. Way to suppress?
+    assertTrue(events.stream()
+        .map(event -> dir.resolve(((WatchEvent<Path>) event).context()))
+        .anyMatch(PATH::equals));
+
+    // any more readable?
+    assertTrue(events.stream()
+        .map(event -> ((WatchEvent<Path>) event).context())
+        .map(dir::resolve)
+        .anyMatch(PATH::equals));
   }
 
   private void write() throws IOException {


azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -73,14 +72,33 @@ public void registerAndTake() throws Exception {
...
     final WatchKey key = this.fileWatcher.take();
     final List<WatchEvent<?>> events = this.fileWatcher.pollEvents(key);
-    // depending on the OS & file system there may be at least 1 or 2 events even with 1 write()
-    assertThat(events.size()).isGreaterThanOrEqualTo(1);
 
-    for (WatchEvent<?> event : events) {
+    // Since we watch the entire parent dir, *and* run multiple tests in parallel, we catch *other*
+    // files being changed here. This assert checks that at least one of them is "ours"
+    // (*and* that at least one event is fired in the first place):
+    assertTrue(events.stream().anyMatch(event -> {
```
li-ygerchikov: So... which of the four ways to say the same do we like best? In order from the closest resemblance to the old code to slightly more diverging in style. I'll leave the one with the most votes :) .

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -73,14 +72,33 @@ public void registerAndTake() throws Exception {
 
     final WatchKey key = this.fileWatcher.take();
     final List<WatchEvent<?>> events = this.fileWatcher.pollEvents(key);
-    // depending on the OS & file system there may be at least 1 or 2 events even with 1 write()
-    assertThat(events.size()).isGreaterThanOrEqualTo(1);
 
-    for (WatchEvent<?> event : events) {
+    // Since we watch the entire parent dir, *and* run multiple tests in parallel, we catch *other*
```
abagshaw: Is there no way to make this test write to and watch a different directory that no other tests use (maybe randomly generate a folder name or something) so we don't have to add logic to _ignore_ the noise created by other tests?

juhoautio: Definitely support this. I didn't know other tests are touching the same dir (what are those?). This test could be isolated by using a junit `@Rule` to create a temporary dir, copy the file there, and watch the temporary dir instead of some shared dir. Like here:

https://github.com/azkaban/azkaban/blob/7efe67b0602fdc65f18fc0e2f91e14a57c2d845e/azkaban-exec-server/src/test/java/azkaban/execapp/event/LocalFlowWatcherTest.java#L31-L32

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
juhoautio: assertTrue is maybe the worst(?) because when it fails it doesn't provide any context, it just says that true was expected but got false.

The [contribution style](https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md#style-guides) guide says:

> Use AssertJ as the assertion library in unit tests.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
djaiswal83: Thanks for bringing this up. @li-ygerchikov please let me know if I can be of any use.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -73,14 +72,33 @@ public void registerAndTake() throws Exception {
 
     final WatchKey key = this.fileWatcher.take();
     final List<WatchEvent<?>> events = this.fileWatcher.pollEvents(key);
-    // depending on the OS & file system there may be at least 1 or 2 events even with 1 write()
```
juhoautio: Did you somehow conclude that this isn't true actually?

juhoautio:
> Since we watch the entire parent dir, and run tests in multiple parallel Gradle threads, at least on Travis, we catch other files being changed in FileWatcherTest. Since we do not expect that, test fails (sporadically, and only on Travis).

Do you still have the Travis test log that you used to come to this conclusion? If yes, would be nice to include in the PR desc.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
juhoautio: I would say the tolerance for more than 1 events should be kept.

> The implementation that observes events from the file system is intended to map directly on to the native file event notification facility where available, or to use a primitive mechanism, such as polling, when a native facility is not available. Consequently, many of the details on how events are detected, their timeliness, and whether their ordering is preserved are highly implementation specific. **For example, when a file in a watched directory is modified then it may result in a single ENTRY_MODIFY event in some implementations but several events in other implementations**.

This is from https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchService.html

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
li-ygerchikov: Thanks @juhoautio, I'll look for a better assert.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
li-ygerchikov: All fair points. I hoped to avoid deeper changes, esp since I seem to remember @djaiswal83 citing some reasons to keep the same file/dir management as in other tests... But you both are right in saying that there must be a better way.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
li-ygerchikov: Yes -- it *is* kept. The check, as implemented, verifies that there will be *at least one* event. This is because `anyMatch()` will return `true` is there are more than 1 matching events. So the old additional check on the number of events becomes redundant. Or am I missing something? Thanks for keeping an eye on this, @juhoautio.

li-ygerchikov:
> Do you still have the Travis test log that you used to come to this conclusion? If yes, would be nice to include in the PR desc.

Yes, I do, and yes, I will.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
juhoautio: Yes, sorry. anyMatch works the same as the previous code, and is cleaner. I actually wanted to say that the comment should be kept about why anyMatch is used here. Intuitively one might expect there to be just one event about file modification, and even more so if the developer’s own system only triggers one.

li-ygerchikov: OK, I see what you are saying. We may keep the note of the fact that even with one write to one file we may get multiple events. It will be even more relevant if I refactor this test to use a unique dir. Thanks for pointing that out.

djaiswal83:
I see that XmlUserManagerTest and FileWatcherTest when run in parallel can cause the failure.
Using randomized directory name with test specific prefix or suffix or simply a directory name specific to test would do. Changing only FileWatcherTest is necessary, however, for complete fix, similar changes are needed in XmlUserManagerTest as well. @li-ygerchikov let me know if I can be of any help.

li-ygerchikov:
@djaiswal83 please feel free to take over. I vaguely remember that you had some difficulties generating random/different dir or file names and resorted to copying the same setup code that used the same dir/file names. Something was failing for you in tests. I didn't really love code duplication but you convinced me :) .

PR #2400, opened 2019-12-11
Propagate system properties to HadoopJavaJobRunnerMain so that jobs u…

djaiswal83:
Propagate system properties to HadoopJavaJobRunnerMain so that jobs using HadoopJava's constructor can correctly use them.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
@@ -39,6 +39,7 @@ public abstract class AbstractProcessJob extends AbstractJob {
   public static final String ENV_PREFIX = "env.";
   public static final String WORKING_DIR = "working.dir";
   public static final String JOB_PROP_ENV = "JOB_PROP_FILE";
+  public static final String JOB_SYS_PROP_ENV = "JOB_SYS_PROP_FILE";
   public static final String JOB_NAME_ENV = "JOB_NAME";
   public static final String JOB_OUTPUT_PROP_FILE = "JOB_OUTPUT_PROP_FILE";
   private static final String SENSITIVE_JOB_PROP_NAME_SUFFIX = "_X";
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
```diff
@@ -161,15 +162,21 @@ public Props getJobGeneratedProperties() {
...
     this.jobProps.put(ENV_PREFIX + JOB_PROP_ENV, files[0].getAbsolutePath());
     this.jobProps.put(ENV_PREFIX + JOB_NAME_ENV, getId());
 
     files[1] = createOutputPropsFile(getId(), this._cwd);
     this.jobProps.put(ENV_PREFIX + JOB_OUTPUT_PROP_FILE, files[1].getAbsolutePath());
+
+    // Create job's system properties file.
+    // TODO : ProcessJob.java:254
+    files[2] = createFlattenedPropsFile(this.sysProps, this._cwd, "_job_sys_props_");
+    this.jobProps.put(ENV_PREFIX + JOB_SYS_PROP_ENV, files[2].getAbsolutePath());
```
ypadron-in: How do we know we are not exposing any application(Azkaban) level sensitive configuration by exposing system properties?

ypadron-in:
Can you elaborate more on the issue this PR is trying to solve? What kind of properties do you mean by "system properties? Why do users need access to them?

djaiswal83:
> Can you elaborate more on the issue this PR is trying to solve? What kind of properties do you mean by "system properties? Why do users need access to them?

There are some jobtypes which need properties set at system level and should not be accessible to user. This is a security feature which ensures certain properties are not overridden.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
djaiswal83: These properties consist of only jobtype plugin properties.

ypadron-in: I believe that properties in "commonprivate.properties" file which are properties set by Azkaban, not visible to users and common to all jobtypes are included in _sysProps_. In addition private.properties file which are specific to jobtypes are also in sysProps according to documentation https://azkaban.readthedocs.io/en/latest/getStarted.html#property-overrides. By listing them in a file inside the flow's execution directory we are making them accesible to users. Users have write permissions on that folder.

ypadron-in:
> > Can you elaborate more on the issue this PR is trying to solve? What kind of properties do you mean by "system properties? Why do users need access to them?
> 
> There are some jobtypes which need properties set at system level and should not be accessible to user. This is a security feature which ensures certain properties are not overridden.

How are we ensuring these properties don't get overridden with this change?

djaiswal83:
> > > Can you elaborate more on the issue this PR is trying to solve? What kind of properties do you mean by "system properties? Why do users need access to them?
> > 
> > 
> > There are some jobtypes which need properties set at system level and should not be accessible to user. This is a security feature which ensures certain properties are not overridden.
> 
> How are we ensuring these properties don't get overridden with this change?

There are certain properties which are needed to execute some job types. These properties are only read by the jobtype code and cant be written back. If a user provides any of these properties in DSL, then they go to job props and not sys props.
We can talk about it in more detail offline.

azkaban-common/src/main/java/azkaban/jobExecutor/AbstractProcessJob.java
djaiswal83: Let us discuss this in person.

PR #2406, opened 2019-12-18
Implemented conditions support for properties-based projects

acharneski:
Existing conditional properties in the model simply needed to be set when parsing a properties file. The validation code is copied based closely on the yaml version. Added test cases and test jobs to demonstrate conditional handling.


acharneski:
@djaiswal83 Added test cases based on existing tests in azkaban.execapp.FlowRunnerConditionalFlowTest

PR #2407, opened 2019-12-18
Implementation of timezone parameter for cron scheduling.

acharneski:


acharneski:
I don't think the test failures are related to my change. Is there a build node issue?
```
azkaban.execapp.FlowPreparerTest > testSetupFlowByMultipleThreads
java.nio.file.NoSuchFileException: /tmp/junit9206816566427839738/projects/_temp.10.34.1576793465419/___azkaban_project_dir_size_in_bytes___
```

acharneski:
The failed code coverage validation has too many issues for me to take action on it. It is marking comment lines as covered/not covered and is marking many code lines in ways that are clearly impossible based on normal code flow. I suspect the line numbers are corrupted somehow.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -169,8 +171,9 @@ public synchronized boolean unscheduleJob(final String jobName, final String gro
    *
    * @return true if job has been scheduled, false if the same job exists already.
    */
-  public synchronized boolean scheduleJobIfAbsent(final String cronExpression, final QuartzJobDescription
-      jobDescription) throws SchedulerException {
+  public synchronized boolean scheduleJobIfAbsent(final String cronExpression, TimeZone timeZone,
+      final QuartzJobDescription
```
djaiswal83: Please fit this in 2 lines.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -197,6 +200,7 @@ public synchronized boolean scheduleJobIfAbsent(final String cronExpression, fin
         .newTrigger()
         .withSchedule(
             CronScheduleBuilder.cronSchedule(cronExpression)
+                .inTimeZone(timeZone)
                 .withMisfireHandlingInstructionFireAndProceed()
 //            .withMisfireHandlingInstructionDoNothing()
```
djaiswal83: This maybe a good opportunity to cleanup the dead code.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -125,7 +125,7 @@
       "azkaban.webapp.servlet.LoginAbstractAzkabanServlet";
   private static final Logger logger = Logger.getLogger(AzkabanWebServer.class);
   private static final int MAX_FORM_CONTENT_SIZE = 10 * 1024 * 1024;
-  private static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
+  static final String DEFAULT_TIMEZONE_ID = "default.timezone.id";
```
djaiswal83: What is the need to remove private qualifier?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -611,7 +614,19 @@ private void ajaxScheduleCronFlow(final HttpServletRequest req,
       return;
     }
 
-    final DateTimeZone timezone = DateTimeZone.getDefault();
+    DateTimeZone timezone = DateTimeZone.getDefault();
+    if (hasParam(req, "timezone")) {
+      String timezoneParam = getParam(req, "timezone");
+      try {
+        timezone = DateTimeZone.forID(timezoneParam);
+      } catch (final Exception e) {
```
djaiswal83: It would be nice if you could specify exactly what Exception it is. Applies to both the cases.

azkaban-web-server/src/test/java/azkaban/webapp/FlowTriggerSchedulerTest.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2017 LinkedIn Corp.
```
djaiswal83: 2020

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
acharneski: It is being set in AzkabanWebServerTest to make test cases system independent.

acharneski:
Thanks @djaiswal83 for your review! Implemented requested changed.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
djaiswal83: I am against the philosophy to modify product behavior to make test(s) work. It should be the other way around.

acharneski: Very well, I have copied the string literal and re-marked the field as private.

acharneski:
Made requested change, please re-review. Thank you.

azkaban-common/src/main/java/azkaban/project/CronSchedule.java
```diff
@@ -18,27 +18,36 @@
...
- * FlowTriggerSchedule is the logical representation of a cron-based schedule.
- * It couldn't be changed once gets constructed.
- * It will be used to schedule a trigger.
+ * FlowTriggerSchedule is the logical representation of a cron-based schedule. It couldn't be
+ * changed once gets constructed. It will be used to schedule a trigger.
  */
 public class CronSchedule implements Serializable {
 
-  private static final long serialVersionUID = -1330280892166841227L;
+  private static final long serialVersionUID = -1330280892166841228L;
```
li-ygerchikov: @acharneski  unfortunately we didn't realize that this would break deserialization for all schedules saved so far with the old UID. We need to either revert this UID change (which will break deserialization for all schedules saved with the *new* id since this change), or figure out a migration path for old schedules saved with the old UID. We are currently trying to figure it out under https://github.com/azkaban/azkaban/pull/2457.

acharneski: Sorry and thanks @li-ygerchikov , I didn't realize serialization was being used to persist objects. If that is the case, when old versions are deserialized with this new format, since there is no default constructor, I believe timeZone will be initialized as null. This will very likely cause an NPE later. I see you already reverted the ID. I'll add a PR to impute the value via an accessor to avoid any NPEs.

PR #2408, opened 2019-12-20
Fix project upload failure recovery

abagshaw:
Originally if a project upload failed, the corrupted zip may still be left in HDFS with the finalized name. Then, future uploads will see that the zip already exists and skip uploading it and go ahead and create a new version associated with that zip (causing problems with execution).

Now, we first upload to a temporary file and then rename it only if the upload succeeds.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
+++ b/azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
@@ -20,6 +20,7 @@ package azkaban.storage;
 import static java.util.Objects.requireNonNull;
 
 import azkaban.AzkabanCommonModuleConfig;
+import azkaban.utils.HashUtils;
 import azkaban.utils.StorageUtils;
 import azkaban.spi.Dependency;
 import azkaban.spi.Storage;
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
```diff
@@ -76,18 +83,19 @@ public String putProject(final ProjectStorageMetadata metadata, final File local
...
+      final Path tmpPath = new Path(projectsPath, TMP_PROJECT_UPLOAD_FILENAME);
 
       // Copy file to HDFS
       log.info(String.format("Creating project artifact: meta: %s path: %s", metadata, targetPath));
-      this.hdfs.copyFromLocalFile(new Path(localFile.getAbsolutePath()), targetPath);
+      this.hdfs.copyFromLocalFile(true, true, new Path(localFile.getAbsolutePath()), tmpPath);
+
+      // Rename the tmp file to the final file and overwrite the final file if it already exists
+      // (i.e. if the hash is the same).
+      this.hdfs.rename(tmpPath, targetPath, Options.Rename.OVERWRITE);
```
li-ygerchikov: Does the other `rename` (without options) fail if target exists?

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
```diff
@@ -76,18 +83,19 @@ public String putProject(final ProjectStorageMetadata metadata, final File local
...
-        log.info(
-            String.format("Duplicate Found: meta: %s path: %s", metadata, targetPath));
-        return getRelativeProjectPath(targetPath);
-      }
+      final Path tmpPath = new Path(projectsPath, TMP_PROJECT_UPLOAD_FILENAME);
 
       // Copy file to HDFS
       log.info(String.format("Creating project artifact: meta: %s path: %s", metadata, targetPath));
-      this.hdfs.copyFromLocalFile(new Path(localFile.getAbsolutePath()), targetPath);
+      this.hdfs.copyFromLocalFile(true, true, new Path(localFile.getAbsolutePath()), tmpPath);
```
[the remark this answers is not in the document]
abagshaw: Whoops - I was comparing to wrong overridden API. Original behaviour is to keep the source. I changed it back.

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
abagshaw: Yes it does.

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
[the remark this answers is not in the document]
li-ygerchikov: I hope I am looking at the right API... :)

azkaban-common/src/main/java/azkaban/storage/HdfsStorage.java
li-ygerchikov: Thanks!

PR #2419, opened 2020-01-09
[TASK] Refine Code and Addd Plugin to control Ramp till 1% level gran…

kxu1026:
…ularity

### DIFF
--- a/azkaban-common/src/main/java/azkaban/ramppolicy/SimpleAutoRampPolicy.java
+++ b/azkaban-common/src/main/java/azkaban/ramppolicy/SimpleAutoRampPolicy.java
@@ -22,76 +22,51 @@ import azkaban.utils.TimeUtils;
 
 
 /**
- * Simple Auto Ramp Policy will be divided to 4 stages
+ * Simple Auto Ramp Policy will be divided to 5 stages
  *  stage 1: 5%
  *  stage 2: 20%
  *  stage 3: 50%
- *  stage 4: 100%
+ *  stage 4: 75%
+ *  stage 5: 100%
  */
-public class SimpleAutoRampPolicy extends AbstractRampPolicy {
-  private static int RAMP_STAGE_MAX = 4;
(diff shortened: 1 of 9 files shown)


azkaban-common/src/main/java/azkaban/ramppolicy/SimpleAutoRampPolicy.java
```diff
@@ -22,76 +22,51 @@
...
-      ExecutableRamp executableRamp
-  ) {
-    if (executableRamp.getState().isPaused()) {
-      return false;
-    }
 
-    updateForRampUp(executableRamp, RAMP_STAGE_MAX, ONE_DAY);
+  @Override
+  protected int getMaxRampStage() {
+    return 5;
```
djaiswal83: Can we use enums or constants instead of hardcoded numbers here?
Also, making RAMP_STAGE_MAX protected should be good enough and avoids having this function.

azkaban-common/src/main/java/azkaban/ramppolicy/SimpleQuickRampPolicy.java
```diff
@@ -20,47 +20,34 @@
...
-  ) {
-    if (executableRamp.getState().isPaused()) {
-      return false;
-    }
-
-    int stage = executableRamp.getState().getRampStage();
-    int flowIdHashCode = flow.getId().hashCode();
-    return isInRange(stage, flowIdHashCode);
+  protected int getMaxRampStage() {
+    return 4;
```
djaiswal83: ditto. Please avoid using hardcoded numbers.

azkaban-common/src/main/java/azkaban/ramppolicy/SimpleAutoRampPolicy.java
kxu1026: copied.

PR #2422, opened 2020-01-15
Address a deadlock occurring between the polling query and the query updating running executions

ypadron-in:
Address a deadlock occurring between:
 - the polling query: `SELECT exec_id FROM execution_flows WHERE status = 20 AND executor_id IS NULL AND flow_data IS NOT NULL AND (use_executor IS NULL OR use_executor = ?) ORDER BY flow_priority DESC, update_time ASC, exec_id ASC LIMIT 1 FOR UPDATE; `
 - and the query updating a running execution: `UPDATE execution_flows SET status=?,update_time=?,start_time=?,end_time=?,enc_type=?,flow_data=? WHERE exec_id=?`

By: 
- introducing retry logic in the flow update query to mitigate the effect of any deadlocks.
- lowering the transaction isolation level of the polling query to TRANSACTION_READ_COMMITTED to reduce the chance of the deadlock appearing since it requires less locks to be acquired than MySQL's default read mode REPEATABLE-READ.

### DIFF
--- a/azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
+++ b/azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
@@ -118,7 +118,7 @@ public class DatabaseOperator {
    * @return The number of rows updated.
    */
   public int update(final String updateClause, final Object... params) throws SQLException {
-    int retry = 0;
+    int retryCount = 0;
     SQLException exception;
     String errorMsg = "Update failed: Reached maximum number of retries: " + AzDBUtil.MAX_RETRIES_ON_DEADLOCK;
     do {
@@ -126,21 +126,22 @@ public class DatabaseOperator {
         return this.queryRunner.update(updateClause, params);
       } catch (final SQLException ex) {
         exception = ex;
-        if (ex.getErrorCode() == 1213) { // 1213 -> Deadlock error
-          retry++;
+        if (queryRunner.getDataSource() instanceof MySQLDataSource &&
+            ex.getErrorCode() == MySQLDataSource.MYSQL_ER_LOCK_DEADLOCK) {
+          retryCount++;
           logger.warn("Deadlock detected when trying to execute: " + updateClause + " with values: "
               + Arrays.toString(params));
           try {
             Thread.sleep(AzDBUtil.RETRY_WAIT_TIME);
           } catch (InterruptedException e) {
-            logger.info("Sleep interrupted");
+            logger.info("Sleep during DB operation retry interrupted.");
           }
         } else {
           errorMsg = "update failed";
           break;
         }
       }
-    } while (retry < AzDBUtil.MAX_RETRIES_ON_DEADLOCK);
+    } while (retryCount < AzDBUtil.MAX_RETRIES_ON_DEADLOCK);
 
     logger.error(errorMsg, exception);
     if (this.dbMetrics != null) {
--- a/azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
+++ b/azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
@@ -20,6 +20,7 @@ import static java.util.Objects.requireNonNull;
 
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Arrays;
 import javax.inject.Inject;
 import org.apache.commons.dbutils.DbUtils;
 import org.apache.commons.dbutils.QueryRunner;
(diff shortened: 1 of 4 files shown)


azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
+      } catch (final SQLException ex) {
+        exception = ex;
+        if (ex.getErrorCode() == 1213) { // 1213 -> Deadlock error
+          retry++;
+          logger.warn("Deadlock detected when trying to execute: " + updateClause + " with values: "
+              + Arrays.toString(params));
+          try {
+            Thread.sleep(AzDBUtil.RETRY_WAIT_TIME);
+          } catch (InterruptedException e) {
+            logger.info("Sleep interrupted");
```
li-gsalia: A more detailed message like "Sleep during DB update retry interrupted" would be more useful.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
+      try {
+        return this.queryRunner.update(updateClause, params);
+      } catch (final SQLException ex) {
+        exception = ex;
+        if (ex.getErrorCode() == 1213) { // 1213 -> Deadlock error
+          retry++;
+          logger.warn("Deadlock detected when trying to execute: " + updateClause + " with values: "
+              + Arrays.toString(params));
+          try {
+            Thread.sleep(AzDBUtil.RETRY_WAIT_TIME);
```
li-gsalia: Using exponential backoff while sleeping is preferred, something like Thread.sleep(AzDBUtil.RETRY_WAIT_TIME * retry_count)

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
    */
   public int update(final String updateClause, final Object... params) throws SQLException {
-    try {
-      return this.queryRunner.update(updateClause, params);
-    } catch (final SQLException ex) {
-      // todo kunkun-tang: Retry logics should be implemented here.
-      logger.error("update failed", ex);
-      if (this.dbMetrics != null) {
-        this.dbMetrics.markDBFailUpdate();
+    int retry = 0;
```
li-gsalia: Consider renaming to retryCount or retries, I prefer first one.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
+    SQLException exception;
+    String errorMsg = "Update failed: Reached maximum number of retries: " + AzDBUtil.MAX_RETRIES_ON_DEADLOCK;
+    do {
+      try {
+        return this.queryRunner.update(updateClause, params);
+      } catch (final SQLException ex) {
+        exception = ex;
+        if (ex.getErrorCode() == 1213) { // 1213 -> Deadlock error
```
li-gsalia: Create a const for the error code.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
-      if (this.dbMetrics != null) {
-        this.dbMetrics.markDBFailUpdate();
+    int retry = 0;
+    SQLException exception;
+    String errorMsg = "Update failed: Reached maximum number of retries: " + AzDBUtil.MAX_RETRIES_ON_DEADLOCK;
+    do {
+      try {
+        return this.queryRunner.update(updateClause, params);
+      } catch (final SQLException ex) {
+        exception = ex;
```
li-gsalia: Can we use ex instead of assigning it to exception?

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
```diff
@@ -117,16 +118,35 @@ public DatabaseOperator(final QueryRunner queryRunner) {
...
+          retry++;
+          logger.warn("Deadlock detected when trying to execute: " + updateClause + " with values: "
+              + Arrays.toString(params));
+          try {
+            Thread.sleep(AzDBUtil.RETRY_WAIT_TIME);
+          } catch (InterruptedException e) {
+            logger.info("Sleep interrupted");
+          }
+        } else {
+          errorMsg = "update failed";
```
li-gsalia: Can the error message be enhanced to provide more info about the error?

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
djaiswal83: Would it be better if the actual SQL is also logged? This way, we can learn about which queries get in this kind of situation and how often.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: Update flow query is called many times during a flow execution and it's a synchronized method. By adding exponential backoff here we're adding delays to the execution time that collectively could go into minutes.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: will rename the variable.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: will do, thanks

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: No, `ex` variable is in scope only inside the catch block.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: This message is logged along with the SQL exception detailing the problem.

azkaban-db/src/main/java/azkaban/db/DatabaseOperator.java
ypadron-in: In my opinion there is no relation between the sleep being interrupted and the query being executed, it's just a waiting time that was not honored. Adding the query to the logs would only cause confusion and not add interesting info IMO.
The query should certainly be logged if it fails.

PR #2423, opened 2020-01-16
[v1] Implementation of space api

svnarumugam:
Implementation of space-api. I have commented out the authentication part to make testing easier. I can revert that in the next PR.

### DIFF
--- a/azkaban-web-server/src/main/java/cloudflow/services/SpaceService.java
+++ b/azkaban-web-server/src/main/java/cloudflow/services/SpaceService.java
@@ -1,0 +1,11 @@
+package cloudflow.services;
+
+import azkaban.user.User;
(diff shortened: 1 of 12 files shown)


svnarumugam:
Modified the base branch to cf-control-plane

azkaban-web-server/src/main/java/cloudflow/services/SpaceService.java
```diff
@@ -0,0 +1,11 @@
+package cloudflow.services;
+
+import azkaban.user.User;
+import cloudflow.models.Space;
+import java.util.List;
+
+public interface SpaceService {
+  Space create(Space space, User user);
```
li-gsalia: Shouldn't this return the spaceId instead of the space object?

PR #2425, opened 2020-01-18
Don't allow pausing an execution while processing a KILL action.

ypadron-in:
The KILL action can take a while to be processed. During that waiting time the user was able to click the PAUSE button. If they did the execution would be marked as PAUSED even though all the running jobs would be killed anyway. In addition to persisting an incorrect flow status this causes that long running flows are not killed automatically and that executor server processes waiting for all executions to finish before stopping will be stuck.

In addition to the backend changes the PAUSE button will be disabled in the UI after the KILL button is clicked. The page will look like:

Screen Shot 2020-01-20 at 10 39 10 PM

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -1119,13 +1119,12 @@ public class FlowRunner extends EventHandler implements Runnable {
 
   public void pause(final String user) throws IllegalStateException {
     synchronized (this.mainSyncObj) {
-      this.logger.info("Flow paused by " + user);
+      this.logger.info("Execution pause requested by " + user);
       if (!this.isKilled() && !this.flowFinished) {
-        this.logger.info("Execution " + this.execId + " has been paused.");
         this.flowPaused = true;
         this.flow.setStatus(Status.PAUSED);
-
         updateFlow();
+        this.logger.info("Execution " + this.execId + " has been paused.");
       } else {
         final String errorMessage = "Execution " + this.execId + " with status " +
             this.flow.getStatus() + " cannot be paused.";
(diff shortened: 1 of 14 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1111,16 +1112,20 @@ private void configureJobLevelMetrics(final JobRunner jobRunner) {
     jobRunner.addListener(JmxJobMBeanManager.getInstance());
   }
 
-  public void pause(final String user) {
+  public void pause(final String user) throws IllegalStateException {
     synchronized (this.mainSyncObj) {
-      if (!this.flowFinished) {
-        this.logger.info("Flow paused by " + user);
+      this.logger.info("Flow paused by " + user);
```
li-ygerchikov: Nit: would you like to say "flow pause requested by user"? It's not given that it will be paused at this point, right?

ypadron-in: I'll change the message, thx.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest.java
```diff
@@ -231,6 +231,29 @@ public void execAndCancel() throws Exception {
...
+    assertStatus("job1", Status.SUCCEEDED);
+    assertStatus("job2", Status.SUCCEEDED);
+    waitJobsStarted(this.runner, "job3", "job4", "job6");
+
+    InteractiveTestJob.getTestJob("job3").ignoreCancel();
+    this.runner.kill("me");
+    assertStatus("job3", Status.KILLING);
+    assertFlowStatus(this.runner.getExecutableFlow(), Status.KILLING);
+
+    // Cannot pause a flow that has already been killed. This should throw IllegalStateException.
```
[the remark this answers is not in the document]
ypadron-in: Done.

PR #2426, opened 2020-01-22
[v1] minor space api updates

svnarumugam:
- made the naming convention consistent


svnarumugam:
Updated the PR with @Override changes that was causing build failure. Thanks to Shardool for pointing this out.

PR #2427, opened 2020-01-23
[TASK] Adding more logs for ramp feature + fixing a bug on saving ramp status to DB

kxu1026:
ADD MORE LOGS + FIX ONE BUG WHICH CAUSE THE CACHED STATUS CANNOT BE SAVED INTO DB

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
@@ -16,12 +16,16 @@
 package azkaban.executor;
 
 import com.sun.istack.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * Object of Executable Ramp
  */
 public class ExecutableRamp implements IRefreshable<ExecutableRamp> {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ExecutableRamp.class);
 
   public enum Action {
     IGNORED, SUCCEEDED, FAILED
@@ -344,7 +348,7 @@ public class ExecutableRamp implements IRefreshable<ExecutableRamp> {
   }
 
   synchronized public void cacheResult(Action action) {
-    this.state.numOfTrail++;
+    this.state.cachedNumOfTrail++;
     switch(action) {
       case SUCCEEDED:
         this.state.cachedNumOfSuccess++;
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -365,12 +369,27 @@ synchronized public void cacheResult(Action action) {
             / ((this.state.numOfTrail + this.state.cachedNumOfTrail) * 1.0))
         : (this.state.numOfFailure + this.state.cachedNumOfFailure);
 
+    LOGGER.info(String.format("Cache Ramp Result : [id = %s, action: %s, %s failure: %d, numOfTrail (%d, %d), numOfSuccess: (%d, %d), numOfFailure: (%d, %d), numOfIgnore: (%d, %d)]"
```
djaiswal83: Indentation is not correct. If you have follow the formatting guidelines (in Azkaban github page) it should fix it for you.
Also, break thing into multiple lines as per the guidelines.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -321,7 +320,7 @@ synchronized void saveSettings() {
         .entrySet()
         .stream()
         .forEach(this::updateExecutedRampFlows);
-    rampDataModel.resetEndFlowCount();
+    rampDataModel.resetFlowCountsAfterSave();
```
djaiswal83: resetFlowCountAfterSave, the "s" is not needed.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -401,6 +400,7 @@ synchronized public void configure(ExecutableFlow executableFlow, File flowDirec
                       ExecutableRampStatus.BLACKLISTED.name()
                   )
               );
+              LOGGER.info(String.format("Ramp Flow As BlackListed Item. [rampid = %s, flowName = %s]", rampId, flowName));
```
djaiswal83: Break all these messages into multiple lines as they are too long.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -598,43 +602,39 @@ synchronized private Action convertToAction(Status status) {
     // Host the current processing ramp flows
     // Map.Key = FlowId, Map.Value = Set Of Ramps
     private volatile Map<String, Set<String>> executingFlows = new HashMap<>();
-    private volatile int beginFlowCount; // Incoming Ramp Requests After Latest Database Pull Action;
-    private volatile int endFlowCount; // Incoming Ramp Requests After Latest Database Pull Action;
+    private volatile int[] flowCounts = {0, 0}; // [0] count of started flow [1] count of ended flows
```
djaiswal83: The earlier approach looks cleaner and more readable. Why change it to array?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -665,8 +665,10 @@ public void start() {
     private void pollExecution() {
       if (this.pollingCriteria.shouldPoll()) {
         if (this.pollingCriteria.satisfiesUnsavedDataAvailableCriteria()) {
+          LOGGER.info("Saving Ramp Setting to Database.");
```
djaiswal83: For uniformity, either use "Saving and Loading" or "Save and Load". I prefer the latter.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
kxu1026: addressing

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
kxu1026: good catch. thanks.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
kxu1026: Want to eliminate the potential synchronization problem. I believe volatile int[] way will lock both beginFlowCount and enFlowCount on updating.

PR #2432, opened 2020-01-28
Add Project model, service and servlet files.

li-gsalia:
This change adds servlet, service and model for Project. The plumbing is not complete for project endpoints yet and Dao layer implementation is in the works. This change is to unblock folks working on execution endpoints.

### DIFF
--- a/azkaban-web-server/src/main/java/cloudflow/services/ProjectService.java
+++ b/azkaban-web-server/src/main/java/cloudflow/services/ProjectService.java
@@ -1,0 +1,12 @@
+package cloudflow.services;
+
+import azkaban.user.User;
+import cloudflow.models.Project;
+import java.util.List;
+
+public interface ProjectService {
+
+  String create(Project project, User user);
+  Project get(String projectId);
+  List<Project> getAll();
+}
--- a/azkaban-web-server/src/main/java/cloudflow/servlets/ProjectServlet.java
+++ b/azkaban-web-server/src/main/java/cloudflow/servlets/ProjectServlet.java
@@ -1,0 +1,88 @@
+package cloudflow.servlets;
+
+import azkaban.server.HttpRequestUtils;
+import azkaban.server.session.Session;
+import azkaban.user.User;
+import azkaban.webapp.AzkabanWebServer;
+import azkaban.webapp.servlet.LoginAbstractAzkabanServlet;
+import cloudflow.services.ProjectService;
+import cloudflow.webapp.controllers.ProjectController;
+import cloudflow.models.Project;
+import com.linkedin.jersey.api.uri.UriTemplate;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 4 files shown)


azkaban-web-server/src/main/java/cloudflow/services/ProjectService.java
```diff
@@ -0,0 +1,12 @@
+package cloudflow.services;
+
+import azkaban.user.User;
+import cloudflow.models.Project;
+import java.util.List;
+
+public interface ProjectService {
+
+  String create(Project project, User user);
+  Project get(String projectId);
```
ypadron-in: Consider renaming these methods to _createProject_ and _getProject_ to maintain consistency.

azkaban-web-server/src/main/java/cloudflow/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,88 @@
...
+    this.objectMapper = server.objectMapper();
+  }
+
+  @Override
+  protected void handleGet(HttpServletRequest req, HttpServletResponse resp, Session session)
+      throws IOException {
+    /* there can be two requests here */
+    /* Get all records */
+    if (GET_ALL_PROJECT_URI.equals(req.getRequestURI())) {
+      String response = objectMapper.writeValueAsString(projectService.getAll());
```
ypadron-in: I think we agreed to use snake case for property names in the response. Are we getting that naming style here?

azkaban-web-server/src/main/java/cloudflow/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,88 @@
...
+       * */
+
+      String projectId = projectService.create(project, new User("gsalia"));
+      // TODO: set the projectId in response body
+      resp.setStatus(HttpStatus.SC_CREATED);
+    } catch(Exception e) {
+      log.error("Input is likely missing something", e);
+      resp.setStatus(HttpStatus.SC_BAD_REQUEST);
+    }
+  }
```
ypadron-in: We should add a "Location" header to the response. Also a response is not actually send at this point.

azkaban-web-server/src/main/java/cloudflow/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,88 @@
...
+      resp.setStatus(HttpStatus.SC_NOT_FOUND);
+    }
+  }
+
+  @Override
+  protected void handlePost(HttpServletRequest req, HttpServletResponse resp, Session session)
+      throws ServletException, IOException {
+    try {
+      String body = HttpRequestUtils.getBody(req);
+      Project project = objectMapper.readValue(body, Project.class);
```
ypadron-in: I don't think we're able to parse expected property names like "space_id" here. Also, using the model class to auto map input parameters can be dangerous as malicious users can guess and provide values for model properties not exposed or expected in the user facing form which will potentially be persisted.

PR #2433, opened 2020-01-28
Fixed the flaky FileWatcherTest.

djaiswal83:
The test along with XmlUserManagerTest were using FileWatcher on same directory causing
intermittent failures.
Each test now creates its own directory to watch.
Includes changes done previously by Yuliy.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
+++ b/azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
@@ -16,8 +16,7 @@
 
 package azkaban.user;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import com.google.common.io.Resources;
 import com.sun.nio.file.SensitivityWatchEventModifier;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -57,6 +72,13 @@ public void tearDown() throws Exception {
     if (this.fileWatcher != null) {
       this.fileWatcher.close();
     }
+
+    // Cleanup directories
```
juhoautio: How about using a `TemporaryFolder` `@Rule` for the workDir to simplify code?

For example here:

https://github.com/azkaban/azkaban/blob/7efe67b0602fdc65f18fc0e2f91e14a57c2d845e/azkaban-exec-server/src/test/java/azkaban/execapp/event/LocalFlowWatcherTest.java#L31-L32

djaiswal83: Thanks. I never knew about this being a Java n00b.
I have updated the test cases.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -26,37 +25,55 @@
...
   @After
   public void tearDown() throws Exception {
     if (this.fileWatcher != null) {
       this.fileWatcher.close();
     }
+
+    // Cleanup directories
+    final Path workDir = PATH.getParent();
+    if (Files.exists(workDir)) {
+      Files.deleteIfExists(PATH);
```
li-ygerchikov: Does `TemporaryFolder` not take care of cleanup automagically? Is this not one of the reasons to use it?

azkaban-common/src/test/java/azkaban/user/XmlUserManagerTest.java
```diff
@@ -102,19 +107,27 @@ public void testDoNotExist() throws Exception {
     fail("XmlUserManager should throw an exception when the file doesn't exist");
   }
 
+  private Path getFilePath(final String testName) throws IOException {
+    final URL configURL = Resources.getResource("test-conf/azkaban-users-test1.xml");
+    final String origPathStr = configURL.getPath();
+
+    // Create a new directory and copy the file in it.
+    final Path workDir = temporaryFolder.newFolder().toPath();
```
li-ygerchikov: Same question: with this change do we need to explicitly delete this folder (in two places below)?

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -73,14 +90,32 @@ public void registerAndTake() throws Exception {
 
     final WatchKey key = this.fileWatcher.take();
     final List<WatchEvent<?>> events = this.fileWatcher.pollEvents(key);
-    // depending on the OS & file system there may be at least 1 or 2 events even with 1 write()
-    assertThat(events.size()).isGreaterThanOrEqualTo(1);
-
-    for (WatchEvent<?> event : events) {
+    // Since we watch the entire parent dir, *and* run multiple tests in parallel, we catch *other*
```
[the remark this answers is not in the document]
djaiswal83: Thanks. I will clean this up.

azkaban-common/src/test/java/azkaban/user/FileWatcherTest.java
```diff
@@ -73,14 +90,32 @@ public void registerAndTake() throws Exception {
...
+    // (*and* that at least one event is fired in the first place):
+    assertTrue(events.stream().anyMatch(event -> {
       @SuppressWarnings("unchecked") final Path name = ((WatchEvent<Path>) event).context();
       final String resolvedFileName = dir.resolve(name).toString();
-      assertEquals(PATH.toString(), resolvedFileName);
-    }
+      return PATH.toString().equals(resolvedFileName);
+    }));
+
+    // Do we really need to convert Paths to Strings in order to compare?
```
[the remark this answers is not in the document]
djaiswal83: copy-paste hazard.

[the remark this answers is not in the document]
djaiswal83: I like the last one ;)

li-ygerchikov:
Thanks for the changes. All looks good.
May I assume that you verified that temp dirs do indeed get deleted? If so, approved.

PR #2438, opened 2020-02-06
Unknown timezone now causes api failure during cron scheduling

acharneski:


acharneski:
The function this PR changes (ajaxScheduleCronFlow) does not seem to be covered by any existing test case. I've extracted the logic itself into a separate function, which is tested.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -684,6 +682,20 @@ private void ajaxScheduleCronFlow(final HttpServletRequest req,
     ret.put(PARAM_MESSAGE, projectName + "." + flowName + " scheduled.");
   }
 
+  public static DateTimeZone getTimeZone(String timezoneParam) {
+    try {
+      return DateTimeZone.forID(timezoneParam);
+    } catch (final IllegalArgumentException e) {
+      TimeZone zone = ZoneInfo.getTimeZone(timezoneParam);
+      if(null == zone) return null;
```
djaiswal83: Can you write this in curly braces block instead of 1 line?

acharneski:
Thanks for approving. How can I get this merged?

PR #2442, opened 2020-02-08
Add endpoint for '/execution'

sshardool:
Includes a number of  whitespace changes for formatting to Azkaban coding style. 
Please ignore them by add `?w=1` to the URL during the PR review.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -29,11 +29,15 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.dbutils.ResultSetHandler;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
+
+import static java.util.Objects.*;
+
 
 @Singleton
 public class ExecutionFlowDao {
@@ -232,6 +236,45 @@ public class ExecutionFlowDao {
       query += "  ORDER BY exec_id DESC LIMIT ?, ?";
       params.add(skip);
       params.add(num);
+    }
+
+    try {
+      return this.dbOperator.query(query, new FetchExecutableFlows(), params.toArray());
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error fetching active flows", e);
+    }
+  }
+
+  public List<ExecutableFlow> fetchFlowHistory(Optional<String> projectId, Optional<String> flowId,
+      Optional<String> flowVersion, Optional<String> experimentId) throws ExecutorManagerException {
+    requireNonNull(projectId, "project id is null");
+    requireNonNull(flowId, "flow id is null");
+    requireNonNull(flowVersion, "flow version is null");
+    requireNonNull(experimentId, "experiment id is null");
+
+    String query = FetchExecutableFlows.FETCH_BASE_EXECUTABLE_FLOW_QUERY;
+    final List<Object> params = new ArrayList<>();
+
+    if (projectId.isPresent() || flowId.isPresent() || flowVersion.isPresent() || experimentId
+        .isPresent()) {
+      query += " WHERE 1 = 1 ";
+    }
+
+    if (projectId.isPresent()) {
+      query += " AND project_id = ? ";
+      params.add(projectId.get());
+    }
+    if (flowId.isPresent()) {
+      query += " AND flow_id = ? ";
+      params.add(flowId.get());
+    }
+    if (flowVersion.isPresent()) {
+      query += " AND flow_version = ? ";
+      params.add(flowVersion.get());
+    }
+    if (experimentId.isPresent()) {
+      query += " AND experiment_id = ? ";
+      params.add(experimentId.get());
     }
 
     try {
--- a/azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
+++ b/azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
@@ -1,0 +1,53 @@
+package cloudflow.models;
+
+import static java.util.Objects.requireNonNull;
+
+import azkaban.executor.ExecutableFlow;
+
+
+public class ExecutionBasicResponse {
+  private int executionId;
+  private String submitUser;
+  private long submitTime;
+  private String experimentId;
+  private String concurrentOption;
+  private String failureAction;
+  private boolean isNotifyFailureFirst;
+  private boolean isNotifyFailureLast;
+
+  public ExecutionBasicResponse(ExecutableFlow executableFlow) {
+    requireNonNull(executableFlow, "executable flow is null");
+    this.executionId = executableFlow.getExecutionId();
+    this.submitUser = executableFlow.getSubmitUser();
+    this.submitTime = executableFlow.getSubmitTime();
+    this.concurrentOption = executableFlow.getExecutionOptions().getConcurrentOption();
+    this.failureAction = executableFlow.getExecutionOptions().getFailureAction().toString();
+    this.isNotifyFailureFirst = executableFlow.getExecutionOptions().getNotifyOnFirstFailure();
+    this.isNotifyFailureLast = executableFlow.getExecutionOptions().getNotifyOnLastFailure();
+    this.experimentId = "none"; //todo
+  }
+
+  public int getExecutionId() {
+    return executionId;
+  }
+
+  public String getSubmitUser() {
+    return submitUser;
+  }
+
+  public long getSubmitTime() {
+    return submitTime;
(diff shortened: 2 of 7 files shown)


azkaban-web-server/src/main/java/cloudflow/readme.md
```diff
@@ -28,3 +28,16 @@ Sample body:
     "gsalia"
   ]
 }
+
+Get executions filtered by Project Id and User.
```
li-gsalia: I feel it would be better to create separate files for each endpoint - one for /spaces, one for /executions, etc. That will make it easier to maintain and search.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -0,0 +1,53 @@
...
+    return experimentId;
+  }
+
+  public String getConcurrentOption() {
+    return concurrentOption;
+  }
+
+  public String getFailureAction() {
+    return failureAction;
+  }
```
li-gsalia: get() for some of the fields are not there, is that intentional?

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -22,89 +20,120 @@
...
-            return;
-        }
-
-        // TODO: handle unsupported routes
+  private static final UriTemplate GET_ALL_EXECUTION_URI = new UriTemplate("/executions");
+  private static final UriTemplate GET_SINGLE_EXECUTION_URI = new UriTemplate(
+      "/executions/{executionId}");
+  private static final UriTemplate GET_JOB_EXECUTION_INFO_URI =
+      new UriTemplate(
+          "/executions/{executionId}/jobs/{jobDefinitionId}/{jobPath}"); // TODO: remove jobPath
```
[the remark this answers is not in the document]
sshardool: Done.

azkaban-web-server/src/main/java/cloudflow/readme.md
sshardool: Having a single Readme makes it easier to quickly compare if the request and responses for the different APIs have some conformity. Ideally this should be covered in some unit tests which could be in different files but till then I would prefer a single file. We can discuss further though.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
sshardool: Fixed.

azkaban-web-server/src/main/java/cloudflow/readme.md
li-gsalia: You are right - we will be writing unit tests soon. I don't feel strongly about this so let's leave it as it is.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -1,126 +1,201 @@
...
+    return argValues[0];
+  }
+
+  private ExecutionBasicResponse executionResponseFromFlow(ExecutableFlow executableFlow) {
+    return new ExecutionBasicResponse(executableFlow);
+  }
+
+  @Override
+  public List<ExecutionBasicResponse> getAllExecutions(Map<String, String[]> queryParamMap) {
+    requireNonNull(queryParamMap, "query param map is null");
```
ypadron-in: I think we should handle these exceptions in the servlet and create appropriate responses (400 or 500).

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -1,126 +1,201 @@
...
+    // Currently we support only one value for each of the parameters and return an error otherwise.
+    Optional<String> flowId = Optional.empty();
+    Optional<String> flowVersion = Optional.empty();
+    Optional<String> experimentId = Optional.empty();
+    Optional<String> projectId = Optional.empty();
+
+    for (Map.Entry<String, String[]> entry : queryParamMap.entrySet()) {
+      String key = entry.getKey();
+      String[] value = entry.getValue();
+      if (key.equalsIgnoreCase(FLOW_ID_PARAM)) {
```
ypadron-in: Wondering if it makes sense to make these _equals_ case sensitive to help us enforce naming conventions.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -22,89 +20,121 @@
...
-            sendResponse(resp, HttpServletResponse.SC_OK, toJsonMap(jobExecution));
-            return;
-        }
-
-        // TODO: handle unsupported routes
+  private static final String EXECUTION_ID_KEY = "executionId";
+  private static final String JOB_ID_KEY = "jobDefinitionId";
+  private static final String JOB_PATH_KEY = "jobPath";
+
+  private static final UriTemplate GET_ALL_EXECUTION_URI = new UriTemplate("/executions");
```
ypadron-in: nit: GET_ALL_EXECUTION**S**

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: I am using `requireNonNull` as a check only for the situations where you would get a null pointer exception otherwise. The intention is not to use it for indicating missing values to the end user. In this case it is both checking and documenting that the method param is not expected to be null.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: Good point. Will be specially relevant in case we decide to use `camelCase` params. Will fix.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: Sorry, chose the wrong line to add the comment.
In line 50 above, I believe that encountering a null value there will mean the user wrote something like `/executions?**flow_id=**&project_id=1`(no value for flow_id param) which can be interpreted as a bad request especially if no default values are provided.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -1,126 +1,201 @@
...
 
-        Optional<JobExecutionAttempt> firstAttempt =
-            attempts.stream().filter(a -> a.getId().equals(0)).findFirst();
-        Long firstStartTime =  firstAttempt.isPresent() ? firstAttempt.get().getStartTime() :
-            node.getStartTime();
-        jobExecution.setStartTime(firstStartTime);
+    //todo(sshardool): enable filtering by these fields once they are in the db schema
+    if (experimentId.isPresent() || flowVersion.isPresent()) {
+      throw new CloudFlowException("Filtering by experiment and flow version is currently not "
+          + "supported");
```
ypadron-in: This exception is not captured in the servlet.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: Ah, I see your point. However even at line 50 above the `requireNonNull` is still only ensuring that the String array should not be null. It can be empty but that is handled by subsequent code. There is certainly more detailed param validation that can be performed, but right now the aim was to prevent the service from choking on any invalid inputs. 
(As a quick check I ran this in a debugger and and see that for args without a value are resolved to an empty string)
I do notice that the `RuntimeException` there needs to be replaced by `CloudFlowException`. I will take care of that in a subsequent PR.  
Let me know if you would like to discuss this further.

ypadron-in: I tested it: `req.getParameterMap()` returns an empty array in the cases when no value is provided for a query parameter. Please disregard.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -241,6 +245,45 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
+    requireNonNull(flowId, "flow id is null");
+    requireNonNull(flowVersion, "flow version is null");
+    requireNonNull(experimentId, "experiment id is null");
+
+    String query = FetchExecutableFlows.FETCH_BASE_EXECUTABLE_FLOW_QUERY;
+    final List<Object> params = new ArrayList<>();
+
+    if (projectId.isPresent() || flowId.isPresent() || flowVersion.isPresent() || experimentId
+        .isPresent()) {
+      query += " WHERE 1 = 1 ";
```
ypadron-in: nice!

PR #2445, opened 2020-02-12
Implementation of 'Create execution' endpoint

ypadron-in:


azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -22,8 +22,8 @@ public ExecutionBasicResponse(ExecutableFlow executableFlow) {
     this.submitTime = executableFlow.getSubmitTime();
     this.concurrentOption = executableFlow.getExecutionOptions().getConcurrentOption();
     this.failureAction = executableFlow.getExecutionOptions().getFailureAction().toString();
-    this.isNotifyFailureFirst = executableFlow.getExecutionOptions().getNotifyOnFirstFailure();
-    this.isNotifyFailureLast = executableFlow.getExecutionOptions().getNotifyOnLastFailure();
+    this.notifyOnFirstFailure = executableFlow.getExecutionOptions().getNotifyOnFirstFailure();
+    this.notifyFailureOnExecutionComplete = executableFlow.getExecutionOptions().getNotifyOnLastFailure();
```
ypadron-in: Current names for these properties in Azkaban are unclear or not easy to understand. Let's agree on better names for the POC.

azkaban-web-server/src/main/java/cloudflow/error/CloudFlowValidationException.java
```diff
@@ -1,25 +1,8 @@
 package cloudflow.error;
 
-import java.util.Arrays;
-import java.util.List;
-
 public class CloudFlowValidationException extends CloudFlowException {
 
-    // A list to avoid multiple api calls when there are various validation errors. Especially
-    // useful with endpoints that take several parameters.
-    private List<String> validationErrors;
```
ypadron-in: Supporting list of errors is too much work for the POC but we should definitely consider it in CloudFlow.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -68,6 +68,17 @@
   private ExecutableFlowRampMetadata executableFlowRampMetadata;
   private String flowLockErrorMessage;
 
+  // added for CloudFlow
+  public static final String EXPERIMENTID_PARAM = "experimentId";
+  public static final String DESCRIPTION_PARAM = "description";
+  public static final String FLOWDEFINITIONID_PARAM = "flowDefinitionId";
+  public static final String FLOWVERSION_PARAM = "flowVersion";
```
sshardool: Nit: please add underscores at word boundaries `EXPERIMENT_ID_PARAM`  `FLOW_VERSION_PARAM` etc.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -192,10 +198,74 @@ public JobExecution getJobExecution(String executionId, String jobDefinitionId,
...
   }
+
+  @Override
+  public String createExecution(ExecutionParameters executionParameters) throws CloudFlowException {
+    requireNonNull(executionParameters);
+    // TODO ypadron: check user permissions
+
+    // TODO ypadron: obtain flow and project dynamically
+    Project project = this.projectManager.getProject("yb-test");
+    Flow flow = project.getFlow("flowPriority2");
```
sshardool: Why are these values hard-coded? Is that what the `todo` is for ?

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -132,9 +122,95 @@ protected void handleGet(HttpServletRequest req, HttpServletResponse resp, Sessi
...
+        sendErrorResponse(resp, HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
+        return;
+      } catch (CloudFlowException e) {
+        sendErrorResponse(resp, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+            DEFAULT_500_ERROR_MESSAGE);
+        return;
+      }
+      logger.info(String.format("New execution of flow %s was successfully queued with id %s",
+          executionParameters.getFlowId(), executionId));
+      resp.setHeader("Location", GET_SINGLE_EXECUTION_URI.createURI(executionId));
```
sshardool: Nice! Didn't know this was supported.
Is it common to pass such information in the header ?

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -689,8 +689,9 @@ public String submitExecutableFlow(final ExecutableFlow exflow, final String use
             this.maxConcurrentRunsPerFlowMap);
         if (running.size() > maxConcurrentRuns) {
           this.commonMetrics.markSubmitFlowSkip();
-          throw new ExecutorManagerException("Flow " + flowId
-              + " has more than " + maxConcurrentRuns + " concurrent runs. Skipping",
+          throw new ExecutorManagerException("Flow with id " +
+              (exflow.getFlowDefinitionId() > 0 ? exflow.getFlowDefinitionId() : flowId)
```
sshardool: Is the `flowDefinitionId` sufficient to uniquely identify the flow or does it need to be the tuple `<flowId,flowDefinitionId>` ?

ypadron-in: This is to maintain compatibility with existent code. `flowDefinitionId` will only have a value when creating executions via the new apis.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
ypadron-in: Following the naming conventions in ExecutableFlow file.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: yes, I need to come back here and fix it once flow endpoints are ready.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
ypadron-in: when new resources are creating yes, it's common.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -192,10 +198,74 @@ public JobExecution getJobExecution(String executionId, String jobDefinitionId,
...
     return attempts;
   }
+
+  @Override
+  public String createExecution(ExecutionParameters executionParameters) throws CloudFlowException {
+    requireNonNull(executionParameters);
+    // TODO ypadron: check user permissions
+
+    // TODO ypadron: obtain flow and project dynamically
+    Project project = this.projectManager.getProject("yb-test");
```
li-gsalia: Now that project related code has been checked in, can you check if project can be fetched dynamically?

azkaban-web-server/src/main/java/cloudflow/servlets/Constants.java
```diff
@@ -9,8 +9,13 @@
   public static final String ATTEMPTS_KEY = "attempts";
   public static final String ID_KEY = "id";
 
+  public static final String FLOW_ID_KEY = "flowId";
+  public static final String FLOW_VERSION_KEY = "flowVersion";
+  public static final String EXPERIMENT_ID_KEY = "experimentId";
+
 
 
   public static final String DEFAULT_500_ERROR_MESSAGE = "Unexpected error encountered.";
```
li-gsalia: Nit: Not part of your change, it would be nice to add "server" in the message so it is clear to users who are not familiar with HTTP error codes.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -132,9 +122,95 @@ protected void handleGet(HttpServletRequest req, HttpServletResponse resp, Sessi
...
+      handleAllExecutions(req, resp);
+    } else if (GET_SINGLE_EXECUTION_URI.match(requestUri, templateVariableMap)) {
+      handleSingleExecution(req, resp);
+    } else if (GET_JOB_EXECUTION_INFO_URI.match(requestUri, templateVariableMap)) {
+      handleExecutionWithJobInfo(req, resp, session, templateVariableMap);
+    } else {
+      sendErrorResponse(resp, HttpServletResponse.SC_NOT_FOUND, DEFAULT_404_ERROR_MESSAGE);
+    }
+  }
+
```
li-gsalia: Most of this was part of Shardool's change. You might have to pull and merge.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -132,9 +122,95 @@ protected void handleGet(HttpServletRequest req, HttpServletResponse resp, Sessi
...
+    if(executionParameters.getFlowVersion() == null) {
+      sendErrorResponse(resp, HttpServletResponse.SC_BAD_REQUEST,
+          String.format("Parameter '%s' is required.", FLOW_VERSION_KEY));
+      return null;
+    } else if(executionParameters.getFlowVersion() <= 0) {
+      final String errorMessage = String.format("Parameter '%s' must be a positive number.", FLOW_VERSION_KEY);
+      logger.error(errorMessage);
+      sendErrorResponse(resp, HttpServletResponse.SC_BAD_REQUEST, errorMessage);
+      return null;
+    }
```
li-gsalia: We should discuss moving checks that fall under business logic to the service layer. That way the move to CF is a little easier and servlets can be removed/replaced with less changes.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -68,6 +68,17 @@
   private ExecutableFlowRampMetadata executableFlowRampMetadata;
   private String flowLockErrorMessage;
 
+  // added for CloudFlow
+  public static final String EXPERIMENTID_PARAM = "experimentId";
```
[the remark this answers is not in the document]
ypadron-in: I agree is much more readable. I'll make the change for those I'm adding but I'll probably leave the rest as it is.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -68,6 +68,17 @@
...
+  public static final String DESCRIPTION_PARAM = "description";
+  public static final String FLOWDEFINITIONID_PARAM = "flowDefinitionId";
+  public static final String FLOWVERSION_PARAM = "flowVersion";
+
+  private int experimentId;
```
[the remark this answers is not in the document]
ypadron-in: This is Azkaban's code which most likely won't be reused. It's used to translate MySQL query results into objects. For convenience and for now these ids are stored as integers, so at this layer I think it makes more sense to keep them as integers.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -270,6 +307,9 @@ public void setFlowLockErrorMessage(final String flowLockErrorMessage) {
     flowObj.put(IS_LOCKED_PARAM, this.isLocked);
     flowObj.put(FLOW_LOCK_ERROR_MESSAGE_PARAM, this.flowLockErrorMessage);
 
+    // flowObj.put(EXPERIMENTID_PARAM, this.experimentId);
```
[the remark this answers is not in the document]
ypadron-in: It's not used as part of this endpoint, it will be used by the `get` endpoints so I prefer to add them as comments.

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -283,6 +283,34 @@ public String toJSON() {
   }
 
   public enum FailureAction {
-    FINISH_CURRENTLY_RUNNING, CANCEL_ALL, FINISH_ALL_POSSIBLE
+    FINISH_CURRENTLY_RUNNING("finishCurrent"),
+    CANCEL_ALL("cancelImmediately"),
+    FINISH_ALL_POSSIBLE("finishPossible");
```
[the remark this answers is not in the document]
ypadron-in: These values are the current ones in Azkaban, it's probably not a good idea to modifying them at this point. I agree that we should definitely revisit this in CF.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: It's possible to retrieve projects now but for creating executions what we have is a flow definition id not a project id, we'll be fetching projects by one of their flows.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
ypadron-in: that's done, this change is only a code movement.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
+package cloudflow.services;
```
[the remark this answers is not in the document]
ypadron-in: This class is not a model, it's more like a transport object, we haven't define a place for those yet.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+    @JsonCreator
+    public static FailureAction valueFromName(String name) {
+      for(FailureAction action: FailureAction.values()) {
+        if(action.name.equals(name)) {
+          return action;
+        }
+      }
+      throw new IllegalArgumentException("No FailureAction for name " + name);
```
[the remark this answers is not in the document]
ypadron-in: This exception is for internal consumption. Users won't see this message, they will see "Parameter xxx isn't valid."

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+    @JsonCreator
+    public static ConcurrentOption valueFromName(String name) {
+      for(ConcurrentOption option: ConcurrentOption.values()) {
+        if(option.name.equals(name)) {
+          return option;
+        }
+      }
+      throw new IllegalArgumentException("No ConcurrentOption for name " + name);
```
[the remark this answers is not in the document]
ypadron-in: That is not the message users will see.

azkaban-web-server/src/main/java/cloudflow/servlets/Constants.java
ypadron-in: sure.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: It should be possible to use the `ProjectManager` to get this information, could you please check ?

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
+package cloudflow.services;
+
+import java.util.Map;
+import org.codehaus.jackson.annotate.JsonCreator;
+
+public class ExecutionParameters {
```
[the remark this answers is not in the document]
ypadron-in: I'll do it, thanks.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
...
+            return this.name;
+        }
+
+    @JsonCreator
+    public static FailureAction valueFromName(String name) {
+      for(FailureAction action: FailureAction.values()) {
+        if(action.name.equals(name)) {
+          return action;
+        }
+      }
```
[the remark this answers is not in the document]
ypadron-in: We have the choice of requesting as value (literally) of failureAction param either 
1- FINISH_CURRENTLY_RUNNING which can be processed with valueOf()
2- finishCurrent which has to be processed with valueFromName()
I chose option 2 because that's what's used today and its easier to type. I'm sticking with current Azkaban values because this might change completely in CF but a 3rd option is to define better names now.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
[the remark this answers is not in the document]
ypadron-in: I'll add it.

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -14,6 +14,9 @@ CREATE TABLE execution_flows (
   executor_id INT                   DEFAULT NULL,
   use_executor INT                  DEFAULT NULL,
   flow_priority TINYINT    NOT NULL DEFAULT 5,
+  experiment_id INT                 DEFAULT NULL,
+  flow_definition_id INT,
+  flow_version INT,
```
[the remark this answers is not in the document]
ypadron-in: I don't know. Gaurav should be able to answer it since he's working on flow endpoints.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
```diff
@@ -0,0 +1,155 @@
...
+  private Integer flowVersion;
+  private String description;
+  private String experimentId;
+  private String submitUser;
+
+  private FailureAction failureAction;
+  private Boolean notifyOnFirstFailure;
+  private Boolean notifyFailureOnExecutionComplete;
+  private ConcurrentOption concurrentOption;
+  private Map<String, Map<String, Object>> properties;
```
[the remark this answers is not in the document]
ypadron-in: Added an example of an api call to the readme file.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -264,8 +266,47 @@ private void setAzkabanExecutionOptions(ExecutionParameters executionParameters,
     executionOptions.setNotifyOnLastFailure(executionParameters.isNotifyFailureOnExecutionComplete());
 
     executionOptions.setConcurrentOption(executionParameters.getConcurrentOption().getName());
+    setRuntimeProperties(executionParameters, executionOptions);
```
sshardool: Is this required for the POC or can this be done later?
In that case we may have a better chance of defining how the properties should be structured without tying ourselves to what we provide now.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -264,8 +266,47 @@ private void setAzkabanExecutionOptions(ExecutionParameters executionParameters,
...
+          + executionParameters.getProperties());
+      throw new CloudFlowException("Overwriting properties of jobs or nested flows is not "
+          + "yet supported.");
+    }
+
+    Map<String, String> runtimeProps = executionParameters.getProperties().get("root");
+    if(runtimeProps.containsKey(CommonJobProperties.FAILURE_EMAILS)) {
+      String rawEmails = runtimeProps.get(CommonJobProperties.FAILURE_EMAILS);
+      if (!rawEmails.isEmpty()) {
+        String[] emails = rawEmails.split("\\s*,\\s*|\\s*;\\s*|\\s+");
```
sshardool: May be have the the different emails as an array of strings in the json rather than having to do a regex manipulation here ? ( AFAIR something similar has caused bugs in the recent past)

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -264,8 +266,47 @@ private void setAzkabanExecutionOptions(ExecutionParameters executionParameters,
...
+        executionOptions.setFailureEmails(Arrays.asList(emails));
+        executionOptions.setFailureEmailsOverridden(true);
+      }
+      runtimeProps.remove(CommonJobProperties.FAILURE_EMAILS);
```
sshardool: Why does the property have to be removed? I was expecting `ExecutionParameters` to be an immutable structure (at least in practice even if not implemented that way).

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -264,8 +266,47 @@ private void setAzkabanExecutionOptions(ExecutionParameters executionParameters,
...
+    }
+
+    if(runtimeProps.containsKey(CommonJobProperties.SUCCESS_EMAILS)) {
```
sshardool: Same comments as above.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -143,32 +144,42 @@ protected void handleGet(HttpServletRequest req, HttpServletResponse resp, Sessi
...
       sendErrorResponse(resp, HttpServletResponse.SC_NOT_FOUND, DEFAULT_404_ERROR_MESSAGE);
     }
   }
 
+  private void handleCreateExecution(HttpServletRequest req, HttpServletResponse resp,
+      Session session) throws IOException {
+    ExecutionParameters executionParameters = extractExecutionParameters(req, resp, session);
+    if (executionParameters == null) {
+      return;
+    }
```
sshardool: Shouldn't this be an exception ? (or at least a `requireNonNull` kind of check)

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
li-gsalia: I agree with Shardool, let's avoid string parsing unless there is no other option.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
li-gsalia: Same comment as Shardool

azkaban-db/src/main/sql/create.execution_flows.sql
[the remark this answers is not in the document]
li-gsalia: "project_flows" is the table that keep info flow definition in current schema. In addition to it, a new table will be created to store all the version of a flow. The primary key for the new table with by flow id + flow version.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
ypadron-in: It's intentionally actually, to stop execution of subsequent code. `extracExecutionParameters` takes care of throwing the appropriate exceptions and constructing the correct responses but calling `sendResponse` is not final, execution of `handleCreateExecution` continues after this method returns.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: To maintain compatibility with current Azkaban code. They should not be removed in CF.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: yes, that's definitely better, thanks!

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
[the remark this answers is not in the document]
ypadron-in: Talked offline. We'll rename _models_ to _dtos_ an put all these classes there.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: I think is needed. I agree final structure requires more thoughts that's why I tried to do it as closer to that final structure as I could think of.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: The id of flows table today is the name of that flow. To create an execution in CF we're given a "real" flow id but plumbing for that is not yet implemented so there is no method in ProjectManager that fetches a flow by id.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: To maintain compatibility with current Azkaban code. They should not be removed in CF.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionParameters.java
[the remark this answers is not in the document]
ypadron-in: Talked offline. We agreed on using the strings associated with the enum values.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: I was referring to using something like `flow = project.getFlow(flowId)` until the flow endpoint has progressed further (this code will likely not change once that happens as well). I am ok delaying this to a subsequent PR.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
sshardool: Could please add a comment reflecting this in the code ?

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: The intention of the code `runtimeProps.remove()` is not clear to me if we have to inclue the `FAILURE_EMAILS` property in the azkaban as well as cloudflow.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -192,10 +200,129 @@ public JobExecution getJobExecution(String executionId, String jobDefinitionId,
...
+      Object propValue = runtimeProps.get(CommonJobProperties.SUCCESS_EMAILS);
+      List<String> emails = getNotificationEmails(propValue);
+      executionOptions.setSuccessEmails(emails);
+      executionOptions.setSuccessEmailsOverridden(true);
+      runtimeProps.remove(CommonJobProperties.SUCCESS_EMAILS);
+    }
+
+    Map<String,String> AzRuntimeProps = runtimeProps.entrySet().stream()
```
sshardool: Nit: identifier should begin with lowercase

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -192,10 +200,129 @@ public JobExecution getJobExecution(String executionId, String jobDefinitionId,
...
+    // supporting only flow level properties overwrite for the POC
+    if(executionParameters.getProperties().size() > 1 ||
+        !executionParameters.getProperties().containsKey("root")) {
+      logger.error("Attempt to overwrite properties of nodes other than the root flow detected: "
+          + executionParameters.getProperties());
+      throw new CloudFlowNotImplementedException("Overwriting properties of jobs or nested flows is not "
+          + "yet supported.");
+    }
+
+    Map<String, Object> runtimeProps = executionParameters.getProperties().get("root");
```
sshardool: Use a const for the element `root`.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
sshardool: In general we need to have a larger discussion about for the format of the properties, esp how they will be provided for embedded flows and jobs.  
Also, it appears embedded flows and runtime properties other than emails are not handled right now.
Please add TODOs for these.

azkaban-web-server/src/main/java/cloudflow/readme.md
```diff
@@ -40,6 +40,68 @@ Sample response body:
...
+	"notifyOnFirstFailure": "true",
+	"notifyFailureOnExecutionComplete": "true",
+	"concurrentOption": "skip",
+	"properties": {
+		"root": {
+			"retries": 3,
+			"failure.emails": ["a@a.com", "b@b.com"],
+			"success.emails": ["c@c.com"]
+		}
+	}
```
sshardool: Do we already use this format for properties somewhere ?
This would need to be documented as part of the API as well. Also wondering if there is a more obvious way of providing rather than using a Map containing Maps.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: yes, thanks for catching this typo.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: sure

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: The problem is that the project object is not available without the flow object in this endpoint (we only receive a flow id). I should be getting the project from the flow, not the other way around, that's why ProjectManager alone doesn't help here.
I prefer to wait until flow changes are ready to implement this part properly.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: Yes, ability to overwrite jobs and root/embedded flows properties requires deeper thoughts and discussions. For the POC we're only supporting root flow properties overwrites. It supports what's available in Azkaban today regarding overwrites. 
I would prefer not to add TODOs for job/embedded flows overwrites because that's not going to be addressed in the POC.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: Azkaban currently ignores when these properties (`failure.emails and success.emails`) are set in executions, instead to override the notification emails it uses 4 other properties (`failureEmailsOverride, successEmailsOverride, failureEmails and successEmails`). 
I prefer to remove the ones Azkaban won't use to avoid any unforeseen side effects.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
ypadron-in: ok

azkaban-web-server/src/main/java/cloudflow/readme.md
ypadron-in: > Do we already use this format for properties somewhere ?
No, today we use `application/x-www-form-urlencoded` as the data format of requests and not `application/json`.
> This would need to be documented as part of the API as well. Also wondering if there is a more obvious way of providing rather than using a Map containing Maps.
I can't think of any but please let me know if you have any suggestions.

PR #2447, opened 2020-02-12
Add endpoint for detailed response for a single execution

sshardool:


azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
```diff
@@ -54,10 +55,18 @@ private void handleAllExecutions(HttpServletRequest req, HttpServletResponse res
...
-  private void handleSingleExecution(HttpServletRequest req, HttpServletResponse resp)
+  private void handleSingleExecution(HttpServletRequest req, HttpServletResponse resp, Map<String
+      , String> templateVariableMap)
       throws IOException {
-    // todo(sshardool): endpoint needs to be implemented
-    sendErrorResponse(resp, HttpServletResponse.SC_NOT_IMPLEMENTED, "not implemented");
+    String executionId = templateVariableMap.get(EXECUTION_ID_KEY);
+    ExecutionDetailedResponse executionResponse;
+    try {
+      executionResponse = executorService.getSingleExecution(executionId);
```
ypadron-in: Can you change the method name to `getExecution`? That would be more consistent with the naming pattern we've been using in the service methods.

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
```diff
@@ -101,6 +104,70 @@ private ExecutionBasicResponse executionResponseFromFlow(ExecutableFlow executab
...
+  public ExecutionDetailedResponse getSingleExecution(String executionIdString) {
+    requireNonNull(executionIdString, "execution id is null");
+    int executionId;
+    String errorMessage;
+    try {
+      executionId = Integer.parseInt(executionIdString);
+    } catch (NumberFormatException nfe) {
+      errorMessage = "Execution Id must be an integer";
+      logger.error(errorMessage, nfe);
+      throw new CloudFlowException(errorMessage, nfe);
```
ypadron-in: This looks like a validation exception. Can you handle those and CloudFlowException(500 errors) in the servlet? Currently only not found exceptions are handled.

sshardool: Makes sense. However I would prefer to make that change separately, such that all the `get execution` apis are covered by it rather than having a try-catch block for each flavor.

azkaban-web-server/src/main/java/cloudflow/servlets/ExecutionServlet.java
sshardool: Not opposed to the change but would be good to have short discussion before we agree on a pattern.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -9,12 +9,15 @@
   private int executionId;
```
ypadron-in: This id should be string in CF.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
```diff
@@ -0,0 +1,202 @@
...
+  "submitTime": 1581552354517,
+  "concurrentOption": "skip",
+  "failureAction": "FINISH_CURRENTLY_RUNNING",
+  "experimentId": "none",
+  "description": "none",
+  "previousFlowExecutionId": "none",
+  "notifyFailureFirst": false,
+  "notifyFailureLast": false,
+  "rootFlow": {
+    "nodeId": "flowPriority2",
```
ypadron-in: Can you remove `nodeId` property?

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
```diff
@@ -0,0 +1,202 @@
...
+            "nodeList": null
+          },
+          {
+            "nodeId": "emb_3",
+            "startTime": 1581552355718,
+            "endTime": -1,
+            "status": "RUNNING",
+            "flowInfo": {
+              "flowId": "flowPriority2:emb_2:emb_3",
+              "flowName": "none",
```
ypadron-in: `flowId` should be something like "1a2b3c4" and flowName is "emb_3" in this case.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
sshardool: To clarify - you mean to remove it only from flow and embedded flows sections as it will be captured in the `flowInfo` for those ?
For jobs this information is not captured anywhere else in the output.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
sshardool: Where is the `"1a2b3c4"` like information ? I don't see it in the Flow.java or the `project_flows` table in the db. 
I began to suspect the `flowId` and `flowName` and one of them can be removed.

sshardool: This may change when there is a new table for flows. Have a `todo` for it in ExecutionServiceImpl.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
```diff
@@ -0,0 +1,194 @@
...
+  "executionId": "2",
+  "submitUser": "azkaban",
+  "submitTime": 1581383285656,
+  "concurrentOption": "skip",
+  "failureAction": "FINISH_CURRENTLY_RUNNING",
+  "experimentId": "defaultExperimentId",
+  "description": "defaultDescription",
+  "previousFlowExecutionId": "defaultPreviousExecutionId",
+  "notifyFailureFirst": false,
+  "notifyFailureLast": false,
```
ypadron-in: Can you use `notifyOnFirstFailure` and `notifyFailureOnExecutionComplete` as the names of these properties to maintain consistency with the rest of the endpoints?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -6,28 +6,35 @@
...
     this.concurrentOption = executableFlow.getExecutionOptions().getConcurrentOption();
     this.failureAction = executableFlow.getExecutionOptions().getFailureAction().toString();
     this.isNotifyFailureFirst = executableFlow.getExecutionOptions().getNotifyOnFirstFailure();
     this.isNotifyFailureLast = executableFlow.getExecutionOptions().getNotifyOnLastFailure();
-    this.experimentId = "none"; //todo
+
+    //todo (sshardool): add these missing fields once available.
+    this.experimentId = "defaultExperimentId";
+    this.description = "defaultDescription";
+    this.previousFlowExecutionId = "defaultPreviousExecutionId";
```
ypadron-in: Can you include in the missing fields job execution attempts, runtime properties and partial run related  fields?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionDetailedResponse.java
```diff
@@ -0,0 +1,22 @@
+package cloudflow.models;
+
+import azkaban.executor.ExecutableFlow;
+import com.sun.istack.internal.NotNull;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+public class ExecutionDetailedResponse extends ExecutionBasicResponse {
+
+  @JsonProperty("rootFlow")
+  @NotNull
```
ypadron-in: Is this annotation working/validating for you?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -6,28 +6,35 @@
 
 
 public class ExecutionBasicResponse {
-  private int executionId;
+  private String executionId;
   private String submitUser;
   private long submitTime;
-  private String experimentId;
   private String concurrentOption;
   private String failureAction;
```
ypadron-in: Shouldn't these fields be enums?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
sshardool: Will be included in a subsequent PR to keep the scope of this one limited.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
sshardool: There was an earlier discussion about what should be the exact strings which appear as part of the output(enum values vs names) so I have kept these as strings for now as that can better incorporate both possibilities.
In case we decide to change this it should be part of a separate PR.

sshardool: There was an earlier discussion about what should be the exact strings which appear as part of the output(enum values vs names) so I have kept these as strings for now as that can better incorporate both possibilities.
In case we decide to change this it should be part of a separate PR.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
```diff
@@ -0,0 +1,194 @@
...
+      {
+        "name": "emb_1",
+        "type": "EMBEDDED_FLOW",
+        "startTime": 1581383286307,
+        "endTime": 1581383286581,
+        "status": "SUCCEEDED",
+        "flowInfo": {
+          "flowId": "defaultId",
+          "flowName": "flowPriority2:emb_1",
+          "flowVersion": "1"
```
ypadron-in: Consider removing "flow" prefix from these fields to avoid redundancy.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
ypadron-in: We'll need those additions for the POC. Can you add them as a comment now to make sure we don't forget to do it?

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
```diff
@@ -0,0 +1,194 @@
...
+            "type": "EMBEDDED_FLOW",
+            "startTime": 1581383286560,
+            "endTime": 1581383436881,
+            "status": "SUCCEEDED",
+            "flowInfo": {
+              "flowId": "defaultId",
+              "flowName": "flowPriority2:emb_2:emb_3",
+              "flowVersion": "1"
+            },
+            "nodeList": [
```
ypadron-in: Wondering if something like "executables" would be a better name for this field (I'm looking for something closer to the business domain and less close to the implementation details and the data structure being used).

azkaban-web-server/src/main/java/cloudflow/services/ExecutionServiceImpl.java
ypadron-in: Why do you prefer to throw a NumberFormatException instead of a CloudFlowValidationException?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
ypadron-in: In the code I think the should be enums to maintain consistency with the rest of the classes and because of the benefits of dealing with enums vs strings. This works well with whatever way we decided to use either the exact enum value or a string associated with it, what changes is the method we use to _stringify_ them (`toString() or getName()`)

azkaban-web-server/src/main/java/cloudflow/models/ExecutionNodeResponse.java
```diff
@@ -0,0 +1,265 @@
...
+
+  @JsonIgnore
+  private final long updateTime;
+
+  @Nullable
+  private final FlowBasicResponse flowInfo;
+
+  @JsonIgnore
+  @Nullable
+  private final String condition;
```
ypadron-in: This field is not going to be used. Isn't it better to remove it?

azkaban-web-server/src/main/java/cloudflow/models/ExecutionNodeResponse.java
```diff
@@ -0,0 +1,265 @@
...
+
+  @JsonIgnore
+  @Nullable
+  private final String nestedId;
+
+  @JsonIgnore
+  @Nullable
+  private final List<String> inputNodeIds; // maps to ExecutableNode.inputNodes
```
ypadron-in: I believe this field is also not used.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
sshardool: I was also thinking about an alternative that avoids the term `node` but didn't quite find anything which is less confusing than `nodeList`. I am open to changing this but not sure if `executables` is the best alternative.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
sshardool: I believe all the TODOs left in the code have to be revisited before the POC itself.
Is the `todo` added here not sufficient ?

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
sshardool: Will do.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionNodeResponse.java
sshardool: I decided to remove some of these fields only from the json output as I suspect they will need to be added sooner than later as  users may be quite interested in fields like input-nodes, especially if we decide to make them visible through an `expand` like option in the near future. 
Having this additional info available doesn't hurt.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionDetailedResponse.java
sshardool: JsonProperty annotation works as expected. 
For `NotNull` and 'Nullable' the primary purpose was documentation. The code is not referring to the correct version of these annotations though and I have a ticket for fixing this.

azkaban-web-server/src/main/java/cloudflow/docs/ExecutionDetailSample.md
ypadron-in: ok, let's go with nodeList for now and think about a better name.

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
```diff
@@ -3,16 +3,17 @@
...
 
 public class ExecutionBasicResponse {
   private String executionId;
   private String submitUser;
   private long submitTime;
   private String concurrentOption;
-  private String failureAction;
-  private boolean isNotifyFailureFirst;
-  private boolean isNotifyFailureLast;
+  private FailureAction failureAction;
```
ypadron-in: Thanks for making this field an enum. 
Now we need to use a Jackson annotation (if available) to specify the method we want to use to serialize it (the default is the enum value but we want a string associated with it).

azkaban-web-server/src/main/java/cloudflow/models/ExecutionBasicResponse.java
sshardool: Thanks for pointing out. I have made a note of it and updating it as part of an upcoming PR.

PR #2449, opened 2020-02-19
Decouple Hadoop cluster from Azkaban JVM

haibchen:
The shows the overall idea of how the Hadoop cluster to which jobs are submitted is decoupled from Azkaban Executor Server, 

1. each job is loaded with an instance of JobClassLoader and executed with that classloader during its entire lifetime.  JobClassLoader is where cluster-specific jars and files are injected to jobs while they are running.

2. In addition, HadoopSecurityManager is no longer a singleton shared by all jobs as well as the Azkaban Executor Server. Instead we create an instance of HadoopSecurityManager inside the per job classloader.

3. JobRunner, while creating the classloader instance for a job, also injects cluster jars and conf files into the classpath of the separate JVM process.

The first 6 commits implements the above idea. The remaining commits include various bug fixes and show how Azkaban integration with Robin, a YARN cluster load balancer, can be done on top of multiple Hadoop Cluster Support.


haibchen:
close

PR #2457, opened 2020-02-26
Revert back the serialVersionUID in CronSchedule class

djaiswal83:
Revert back the serialVersionUID in CronSchedule class as the update was not required.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/CronSchedule.java
+++ b/azkaban-common/src/main/java/azkaban/project/CronSchedule.java
@@ -29,7 +29,11 @@ import org.apache.commons.lang.builder.HashCodeBuilder;
  */
 public class CronSchedule implements Serializable {
 
-  private static final long serialVersionUID = -1330280892166841228L;
+  /**
+   * CAUTION : Please do NOT change this serialVersionUID as it may break
+   * backward compatibility.
+   */
+  private static final long serialVersionUID = -1330280892166841227L;
   private final String cronExpression;
   private final String timeZone;
 


djaiswal83:
Addressed the review comment.

azkaban-common/src/main/java/azkaban/project/CronSchedule.java
```diff
@@ -29,7 +29,11 @@
  */
 public class CronSchedule implements Serializable {
 
-  private static final long serialVersionUID = -1330280892166841228L;
+  /**
+   * CAUTION : Please do NOT change this serialVersionUID as it may break
+   * backward compatibility.
+   */
+  private static final long serialVersionUID = -1330280892166841227L;
```
li-ygerchikov: Now it will break for all those who already saved tiggers with UID ...228.

li-ygerchikov: From the [doc](https://docs.oracle.com/javase/7/docs/platform/serialization/spec/version.html): 
```Adding fields - When the class being reconstituted has a field that does not occur in the stream, that field in the object will be initialized to the default value for its type. If class-specific initialization is needed, the class may provide a readObject method that can initialize the field to nondefault values.```
Do we need to provide a (non-`null`) default value to the new `timeZone` field? If so, we may have to provide a custom `readObject` method.

li-ygerchikov: @acharneski Heads-up, this will break Azkaban for you if you already have schedules saved with the new UID. 
Maybe we need to put our heads together and figure out a better approach. We can't be just changing UID (back *or* forth) without also providing migration support in code.

acharneski: I'll wrap any access to this field in an accessor to avoid NPEs. Thanks @li-ygerchikov

acharneski: I'm also going to be preparing a fallback fix for those users who are on the latest version. Will need to catch this specific error and then provide object stream rewriting logic to replace the object id with the one the code will accept. I hope to have it ready today or tomorrow.

li-ygerchikov: @acharneski Thank you for jumping in! My apologies for breaking it for you by reverting the UID. We should've been smarter about it, but who has the time...
Yes, wrapping access is one way to avoid NPEs. The other, and the one that seems to be implied by the doc quoted above, is to implement custom `readObject` method for the affected class -- I imagine it's be something as simple as a call to `defaultReadObject` followed by default initialization if `null` was read for the field in question.
Oh, I _think_ I just realized what you mean by stream rewriting logic. It sounds ...ambitious... Wouldn't this logic have to be aware of the implementation details/format of the default serialization provided by `java.io`? I'd probably experiment with the techniques mentioned in https://docs.oracle.com/javase/8/docs/platform/serialization/spec/input.html first, and hope that either `resolveClass/resolveObject` or `readObjectNoData` could help.
Please ping me if you'd like to discuss any of this, and please keep me in the loop for any reviews.
Thanks again for your help with this!

acharneski: I've prepared a fix that should solve the issue for anyone who used this recent release to create flow triggers: https://github.com/azkaban/azkaban/pull/2459

PR #2458, opened 2020-02-26
Adding logs to debug why executions directories are not getting cleaned up for some executors

jakhani:
Executions are not getting cleaned up on some servers. For same version, it is working for few hosts. I am adding more logs to debug this issue.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -90,17 +90,17 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Execution manager for the server side execution.
- *
+ * <p>
  * When a flow is submitted to FlowRunnerManager, it is the {@link Status#PREPARING} status. When a
  * flow is about to be executed by FlowRunner, its status is updated to {@link Status#RUNNING}
- *
+ * <p>
  * Two main data structures are used in this class to maintain flows.
- *
+ * <p>
  * runningFlows: this is used as a bookkeeping for submitted flows in FlowRunnerManager. It has
  * nothing to do with the executor service that is used to execute the flows. This bookkeeping is
  * used at the time of canceling or killing a flow. The flows in this data structure is removed in
  * the handleEvent method.
- *
+ * <p>
  * submittedFlows: this is used to keep track the execution of the flows, so it has the mapping
  * between a Future<?> and an execution id. This would allow us to find out the execution ids of the
  * flows that are in the Status.PREPARING status. The entries in this map is removed once the flow
@@ -272,11 +272,11 @@ public class FlowRunnerManager implements EventListener,
    * Setting the gid bit on the execution directory forces all files/directories created within the
    * directory to be a part of the group associated with the azkaban process. Then, when users
    * create their own files, the azkaban cleanup thread can properly remove them.
-   *
+   * <p>
    * Java does not provide a standard library api for setting the gid bit because the gid bit is
    * system dependent, so the only way to set this bit is to start a new process and run the shell
    * command "chmod g+s " + execution directory name.
-   *
+   * <p>
    * Note that this should work on most Linux distributions and MacOS, but will not work on
    * Windows.
    */
@@ -612,7 +612,9 @@ public class FlowRunnerManager implements EventListener,
    * delete execution dir pertaining to the given execution id
    */
   private void deleteExecutionDir(final int executionId) {
+    LOGGER.info("Deleting execution directory for " + executionId);
     synchronized (this.executionDirDeletionSync) {
+      LOGGER.info("Starting execution directory deletion for " + executionId);
       final Path flowExecutionDir = Paths.get(this.executionDirectory.toPath().toString(),
           String.valueOf(executionId));
       try {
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -90,17 +90,17 @@
 
 /**
  * Execution manager for the server side execution.
- *
+ * <p>
```
djaiswal83: What does this <p> mean?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -625,10 +626,17 @@ private void deleteExecutionDir(final int executionId) {
 
   @Override
   public void handleEvent(final Event event) {
-    if (event.getType() == EventType.FLOW_FINISHED || event.getType() == EventType.FLOW_STARTED) {
-      final FlowRunner flowRunner = (FlowRunner) event.getRunner();
-      final ExecutableFlow flow = flowRunner.getExecutableFlow();
+    // Adding extra logging for debuggability of execution directory cleanup call
```
djaiswal83: Please add a todo which states this piece of code backs to original state since the structure has changed.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -625,10 +626,17 @@ private void deleteExecutionDir(final int executionId) {
...
+    final FlowRunner flowRunner = (FlowRunner) event.getRunner();
+    final ExecutableFlow flow = flowRunner.getExecutableFlow();
 
+    if (event.getType() != null && flow != null) {
```
li-gsalia: If Azkaban supports configuring the log level then these messages can be added as logger.debug. They can be turned on as needed by changing the log level to debug.

djaiswal83: I believe we have other logs which are debug level too. It would cause a see of logs being added. Not terrible but could be avoided.

sshardool: Nit: both the conditions in the `if` seem to represent an invalid state. Perhaps through an exception for it in a `else` block or at least log the occurrence of it ?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: This means paragraph. It is added by checkstyle automatically.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: @li-gsalia  & @djaiswal83 It will be lot many logs if we will change it to DEBUG. Is it ok if we keep it at info level?

jakhani: Adding logging for the occurrence of it.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: Adding the todo

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: Adding debug logs.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -612,6 +612,7 @@ public ExecutableFlow getExecutableFlow(final int execId) {
    * delete execution dir pertaining to the given execution id
    */
   private void deleteExecutionDir(final int executionId) {
+    LOGGER.debug("Deleting execution directory for " + executionId);
```
li-ygerchikov: Would you like this output to happen *inside* the synchronized block, in case we don't enter it right away (or at all)?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
li-ygerchikov: Alternatively maybe the entire extra work done here (not just the output resulting from this additional work) could be predicated on logging level: `if (!log.getEffectiveLevel().isGreaterOrEqual(Level.INFO)) { ...`

jakhani: Good point @li-ygerchikov. I would like to keep rest of the logs for future purpose as well but i would like to delete these logs once issue is solved. Is that ok with you?

djaiswal83:
Thanks for squashing them into one. Looks good.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
li-ygerchikov: Yes, it's OK. Since you are still logging at INFO level, no objections.

PR #2459, opened 2020-02-28
Implemented workaround for bad serialization id in CronSchedule

acharneski:
1. Added field encapsulation to account for possible NPEs due to legacy object deserialization.
2. Added functionality to rewrite Quartz blobs to fix problematic serialization ids introduced in recent release.
3. Added support for Quartz in solo-server, critical in testing the serialization id hack-fix.


azkaban-solo-server/src/main/resources/quartz_schema.sql
```diff
@@ -0,0 +1,249 @@
...
+-- In your Quartz properties file, you'll need to set 
+-- org.quartz.jobStore.driverDelegateClass = org.quartz.impl.jdbcjobstore.StdJDBCDelegate
+
+CREATE TABLE QRTZ_CALENDARS (
+  SCHED_NAME VARCHAR(120) NOT NULL,
+  CALENDAR_NAME VARCHAR (200)  NOT NULL ,
+  CALENDAR IMAGE NOT NULL
+);
+
+CREATE TABLE QRTZ_CRON_TRIGGERS (
```
jakhani: Can you please use CREATE TABLE IF NOT EXISTS? Even though it is for solo server and we expect that these table won't be existing. Can you please add it for all the tables?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
```diff
@@ -72,8 +73,36 @@ public QuartzScheduler(final Props azProps) throws SchedulerException {
...
-    logger.info("Quartz Scheduler started.");
+    try {
+      this.scheduler.start();
+      logger.info("Quartz Scheduler started.");
+    } catch (SchedulerConfigException e) {
+      if(e.getCause() instanceof org.quartz.JobPersistenceException) {
+        if(e.getCause().getCause() instanceof java.io.InvalidClassException) {
+          if(enableSerializationHack()) {
+            try {
+              this.scheduler.start();
```
jakhani: Are you planning to change old serialized objects in db? or you will follow same way and keep multiple versions of serialized objects? I personally am not comfortable with catching exception and add then constructing old version objects.

jakhani:
@acharneski  Thank you so much for working on this PR very quickly to unblock all of us. I really appreciate your quick response and fix.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
```diff
@@ -0,0 +1,627 @@
+package azkaban.scheduler;
```
li-ygerchikov: Ugh... at the first blush, this is way, *way* too much Quartz internal logic dumped into Azkaban. What happens when Quartz changes `StdJDBCDelegate`? Will we have to forever maintain this piece of code in sync with Quartz changes? Personally I would rather purge the DB and tell users to recreate their schedules if necessary than add this to Azkaban and try to explain to anyone why we did this -- even long after no one has any `...228L` UIDs left in their databases anymore. 
Are we sure there is no better way?

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -163,6 +163,14 @@ public boolean resumeFlowTriggerIfPresent(final int projectId, final String flow
               flowId, flowTrigger, submitUser, quartzTriggers.isEmpty() ? null
               : quartzTriggers.get(0), isPaused, flow.isLocked());
         } catch (final Exception ex) {
+          if (ex instanceof org.quartz.JobPersistenceException) {
+            if (ex.getCause() instanceof java.io.InvalidClassException) {
+              if(this.scheduler.enableSerializationHack()) {
+                logger.info("Enable serialization hack");
+                return getScheduledFlowTriggerJobs();
```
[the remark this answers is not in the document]
li-ygerchikov: I must admit though, the more I look into this, the more I respect what you've accomplished here... a true hack.

For my education: why can't we directly extend `StdJDBCDelegate`, overwrite `getObjectFromBlob` *only* with your filtering version, tell Quartz (via `org.quartz.jobStore.driverDelegateClass` configuration property) to use that delegate from get-go ...and call it good? 

Extra benefit is that anyone who hasn't been exposed to the new UID, or who purged their DB of it, either deliberately or over time, doesn't need to bother and can just use straight `StdJDBCDelegate` again.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
li-ygerchikov: I must admit though, the more I look into this, the more I respect what you've accomplished here... a true hack.

For my education: why can't we directly extend `StdJDBCDelegate`, overwrite `getObjectFromBlob` *only* with your filtering version, tell Quartz (via `org.quartz.jobStore.driverDelegateClass` configuration property) to use that delegate from get-go ...and call it good? 

Extra benefit is that anyone who hasn't been exposed to the new UID, or who purged their DB of it, either deliberately or over time, doesn't need to bother and can just use straight `StdJDBCDelegate` again.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
```diff
@@ -0,0 +1,627 @@
...
+  public Set<String> selectPausedTriggerGroups(Connection conn) throws SQLException {
+    return inner.selectPausedTriggerGroups(conn);
+  }
+
+  @Override
+  protected Object getObjectFromBlob(ResultSet rs, String colName) throws ClassNotFoundException, IOException, SQLException {
+    byte[] bytes = filterBlob(rs.getBytes(colName));
+    if (bytes != null && bytes.length != 0) {
+      InputStream binaryInput = new ByteArrayInputStream(bytes);
+      ObjectInputStream in = new ObjectInputStream(binaryInput);
```
li-ygerchikov: If we get to this point, and if we can control which `ObjectInputStream` implementation we use here, then could we possibly implement our own subclass of `ObjectInputStream` that overrides `readClassDescriptor` that effectively ignores `serialVersionUID`, along the lines of https://stackoverflow.com/a/796589? If we do it for one class name and one UID only, then it feels pretty safe. And we don't need to replace bytes, and we don't need to expose the wire format of serialized objects. Any fundamental reasons this won't work?

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
```diff
@@ -0,0 +1,627 @@
...
+      } finally {
+        in.close();
+      }
+    }
+    return null;
+  }
+
+  private byte[] filterBlob(byte[] original) throws SQLException, IOException {
+    return replace(original,
+        headerBytes(CronSchedule.class, -1330280892166841228L),
```
li-ygerchikov: Another nit, why not compute these two once, and store in `private static final  byte[]` fields?

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
acharneski: If the object is ever persisted again, it will have the new serialization id as that is a static value. To iterate through and overwrite all objects would be significantly more functionality (and thus more possibility of something going wrong.) I am attempting to provide a minimally invasive solution.

azkaban-solo-server/src/main/resources/quartz_schema.sql
acharneski: This is a verbatim copy of the schema definition found in the quartz source code. I'll look into such a change, though the add key statements should also be idempotent. As is, the errors are logged at a low level and the ignored. This case is actually common - any startup of the solo-server past the first one.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
```diff
@@ -55,13 +62,39 @@ public AzkabanSingleServer(final AzkabanWebServer webServer,
...
+      connectionProps.put("user", "sa");
+      connectionProps.put("password", "");
+      DriverManager.registerDriver((Driver) Class.forName("org.h2.Driver").getDeclaredConstructor().newInstance());
+      Connection conn = DriverManager.getConnection("jdbc:h2:./local/h2quartz", connectionProps);
+      String rawSql = IOUtils.toString(ClassLoader.getSystemClassLoader().getResource("quartz_schema.sql"));
+      String noComments = Arrays.stream(rawSql.split("\n")).map(x -> x.trim())
+          .filter(x -> !x.startsWith("--"))
+          .filter(x -> !x.isEmpty())
+          .reduce((a, b) -> a + "\n" + b).get();
+      for (String sql : noComments.split(";")) {
```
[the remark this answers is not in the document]
acharneski: I'll look at the runscript tool, but since we know what exact script we are executing we can easily verify that a simple character-level match is fine here.

azkaban-solo-server/src/main/resources/quartz_schema.sql
```diff
@@ -0,0 +1,249 @@
+-- Thanks to Amir Kibbar and Peter Rietzler for contributing the schema for H2 database,
```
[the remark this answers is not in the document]
acharneski: I didn't see those. If either work on Postgres, should be fine on H2.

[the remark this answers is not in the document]
acharneski: Drop table would defeat the purpose here, since the point is to test persistent behavior and that would erase all data.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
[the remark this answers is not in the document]
acharneski: Yes, one of the scenarios where it will return false is if the targeted inner class is always of the wrapped type. This function will not re-hack the scheduler; it will return false.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
acharneski: There are solutions in the vein of modifying the Quartz database, but that would require manual intervention and worse it would require every user affected by this issue to get into the guts of Quartz. I'm not even sure what such a solution would look like. Clearing the Quartz database could easily be a problematic amount of data loss for any organization using this at any scale in production. Either way, those remain as backup options in case this feature doesn't work. This solution attempts to make it painless for the user. 

Yes, there are many things that could go wrong. Using reflection to change private fields and then doing a search-and-replace on a binary object serialization stream is obviously not ideal. I do believe this will work for pretty much everyone using RDBS-backed Quartz, which should be most users. Quartz code itself is fairly stable here, so I don't expect it to change anytime soon.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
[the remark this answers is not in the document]
acharneski: The main problem here is that I'd have to modify the build script to add h2 to the compile classpath, whereas I'm trying to make this fix minimally invasive. (It is currently part of the runtime classpath only)

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
acharneski: Thanks for the pointer. I've implemented this change.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
acharneski: No longer applicable; changed workaround approach.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
acharneski: I wrapped QuartzScheduler::enableSerializationHack with an extra try/catch to defend against anything unexpected side effects. I also added a comment at the header of this class (renamed to CustomQuartzDeserializer) to discuss the scope and intent.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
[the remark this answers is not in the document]
acharneski: I've also made the exception recognition much more specific, found now in CustomQuartzDeserializer.isSerializationBug

acharneski:
Thanks @li-ygerchikov and @jakhani for the review. I've responded to all your comments and made some changes to address them. Please re-review.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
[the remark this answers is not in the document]
li-ygerchikov: Sorry, I wasn't perfectly clear. I meant the statement approach, http://h2database.com/html/commands.html?highlight=runscript&search=runscript#runscript -- not the org.h2.tools.RunScript tool. The statement approach should not require build changes, should it?

azkaban-solo-server/src/main/resources/quartz_schema.sql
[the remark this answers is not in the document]
li-ygerchikov: > I didn't see those. If either work on Postgres, should be fine on H2. However, they wouldn't be available on the classpath of solo-server.

One of them, the `azkaban-web-server/src/test/resources/create.quartz-all-tables.sql`, is specifically for testing on H2. 

Speaking of which, I am beginning to think that initializing DB in AzkabanSingleServer and pointing it unconditionally to H2 may be wrong. Does Solo Server not use WebServer and ExecServer and do those two not configure and open DB according to properties? So if I have solo server configured to run against a MySQL DB, it will still create a *separate* H2 DB with separate set of Quartz tables. Does it not seem strange? Sorry, this may need more thought.

azkaban-web-server/src/main/java/azkaban/scheduler/CustomDeserializerDelegate.java
li-ygerchikov: Sorry, again I haven't been clear. I am not advocating changing the DB. I am questioning the massive reflection-based wrapper for StdJDBCDelegate that has to implement every method. Can we essentially add an implementation of StdJDBCDelegate as if it was another `MSSQLDelegate` or `OracleDelegate`, and only override `getObjectFromBlob` in this custom class? Then inject that version via the `org.quartz.jobStore.driverDelegateClass` Azkaban property? Is there something that I am missing that makes the wrapper approach necessary and the subclass approach impossible?

acharneski: For some reason your second comment in this thread only just now appeared for me. Such an approach might work, I'll look into it. It would avoid reflection. However, it would need to be manually configured. Also, we'd need to add and read another property to figure out what original driverDelegateClass we are supposed to wrap (or make a lot of subclasses to support each RDBS type). I'm a little worried about things going wrong with that extra logic, but it's certainly no worse than reflection-based injection.

acharneski: I've made changes to allow the explicit configuration of CustomQuartzDeserializer and separated out the reflection-based injector into QuartzHacker. This allows the zero-configuration bug mitigation to exist and be managed separately from the core fix. If triggered, it will output at warn level and explanation that guides the user to explicitly configure the fix.

azkaban-solo-server/src/main/resources/quartz_schema.sql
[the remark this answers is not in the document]
acharneski: I have implemented an additional check that only invokes the H2 script if: 1) Quartz is enabled, 2) The H2 driver is used, and 3) The table prefix is the standard one. The schema script is only executed if the database is uninitialized, as indicated by the existence of the `QRTZ_TRIGGERS` table. The schema script is specified by a configuration property, preconfigured as `azkaban-web-server/src/test/resources/create.quartz-all-tables.sql` in the solo-server config.

acharneski:
@li-ygerchikov Addressed comments

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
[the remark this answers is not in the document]
acharneski: I'll test this out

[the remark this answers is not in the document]
acharneski: Implemented and tested, thanks

acharneski:
@li-ygerchikov @jakhani Just a follow up, I believe all changes requested have been made. Please re-review. Thanks!

acharneski:
> The code appears to be good after all the iterations.
> Please follow the coding guidelines and use wrap around, final etc.
> I think if you configure intelliJ as per the readme, it should do it for you.

Thanks, I have made the formatting changes and cleanups. Please check again @djaiswal83

djaiswal83:
@acharneski can you please rebase your code? We had a test which started failing after daylight saving. It would be nice to have clean travis CI build.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java
jakhani: @acharneski Thank you so much for working on this PR.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
```diff
@@ -163,6 +163,10 @@ public boolean resumeFlowTriggerIfPresent(final int projectId, final String flow
               flowId, flowTrigger, submitUser, quartzTriggers.isEmpty() ? null
               : quartzTriggers.get(0), isPaused, flow.isLocked());
         } catch (final Exception ex) {
+          if (QuartzScheduler.isSerializationBug(ex) && this.scheduler.enableSerializationHack()) {
```
jakhani: If this condition is not met, shall we just throw the exception rather than logging it?
@li-ygerchikov  Your thoughts?

jakhani:
@acharneski I have added one small comment. Everything else looks good to me. Thank you so much for working on this PR.

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
acharneski: Logging it and setting `scheduledFlowTrigger = null` is the existing behavior. Do you think it should be changed?

acharneski:
Rebased as @djaiswal83 requested
Thank you!

azkaban-web-server/src/main/java/azkaban/flowtrigger/quartz/FlowTriggerScheduler.java
li-ygerchikov: Yes, I would retain the existing functionality for the purpose of this PR, @jakhani. If you think we need to revise this behavior (which might be a perfectly valid suggestion), I would do it separately.

jakhani: @li-ygerchikov Sounds good. Thank you so much.
@acharneski Thank you for responding to my comment. It can be taken care of in another PR by us like @li-ygerchikov mentioned.

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
```diff
@@ -86,6 +121,23 @@ public boolean implies(final ProtectionDomain domain, final Permission permissio
       return;
     }
 
+    if (props.getBoolean(Constants.ConfigurationKeys.ENABLE_QUARTZ, true)) {
```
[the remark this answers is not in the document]
acharneski: Change default value to false. This is not just used in unit tests; I've used it when manually starting the standalone server for manually testing this feature. Because it requires a service restart with a code change in-between, an automated unit test doesn't seem feasible. Additionally, it could easily be used by anyone hoping to use the relevant features in a standalone instance for whatever purpose. (Testing, demonstration, small scale deployment, etc)

azkaban-solo-server/src/main/java/azkaban/soloserver/AzkabanSingleServer.java
```diff
@@ -86,6 +121,23 @@ public boolean implies(final ProtectionDomain domain, final Permission permissio
...
+      String initScript = props.getString("azkaban.server.schedule.quartz_schema");
+      if (dataSource != null) {
+        String tablePrefix = props.getString("org.quartz.jobStore.tablePrefix");
+        if ("QRTZ_".equals(tablePrefix)) {
+          String driver = props.getString("org.quartz.dataSource." + dataSource + ".driver");
+          if ("org.h2.Driver".equals(driver)) {
+            String jdbcUrl = props.getString("org.quartz.dataSource." + dataSource + ".URL");
+            String user = props.getString("org.quartz.dataSource." + dataSource + ".user");
+            String password = props.getString("org.quartz.dataSource." + dataSource + ".password");
+            initializeQuartzDb(jdbcUrl, user, password, initScript);
```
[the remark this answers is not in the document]
acharneski: I am not sure where you are coming from here. I don't see this as a test-only feature, and afaik the quartz database would not exist in solo-server without this or some similar change. The additional configuration checks were made as a result of previous review comments expressing concern about assuming things about the configuration.

azkaban-web-server/src/main/java/azkaban/scheduler/QuartzHacker.java
```diff
@@ -0,0 +1,141 @@
...
+      logger.info(String
+          .format("Class of delegate is not StdJDBCDelegate (%s)", delegate.getClass().getName()));
+      return null; // Do not re-wrap!
+    }
+    if (delegate instanceof CustomQuartzDeserializer) {
+      logger.info("Class of delegate is already CustomQuartzDeserializer");
+      return null; // Do not re-wrap!
+    }
+    StdJDBCDelegate inner = (StdJDBCDelegate) delegate;
+    CustomQuartzDeserializer value = new CustomQuartzDeserializer(inner);
```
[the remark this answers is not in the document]
acharneski: The additional overrides are there to make sure that ONLY the relevant code path is intercepted. Every other call does not go into the default implementation of StdJDBCDelegate, but immediately delegates to the wrapped implementation. This provides more assurance that the wrapped class isn't being modified unnecessarily. From my review of the child classes, it does appear this isn't needed, but I wanted to be sure. It also future-proofs against overrides that I can't see yet.

As far as the hacker class, yes it is to provide a transparent, automatic solution. Without it, it would be pretty hard to dig up the needed configurations needed to solve this problem, exactly because this probably only affects a handful of users (if any). Google would probably never uncover this thread.

@li-ygerchikov - To be clear, this issue does not affect my organization, and I'm pretty much done trying to get this change in. I created a mess and I am simply trying to clean it up. I believe I have done my part. The committers of this project can either do their part to clean up this mess (which was included in 2 releases - not entirely my doing) or close this PR and accept this issue. Personally, I do find it theoretically possible that an organization might have grabbed one of the affected versions, locked into it, and might be deploying it to Prod as we speak. They might not discover the hole they are digging themselves into for another year. That possibility is why I put a significant amount of effort into fixing this issue.

PR #2465, opened 2020-03-04
Changing json column from BLOB to MEDIUMBLOB in project_flows table.

jakhani:
BLOB type of json column is small and we are exceeding limit of it for few projects. I am increasing size to MEDIUMBLOB. I have verified that is is not affecting webserver startup time.

### DIFF
--- a/azkaban-db/src/main/sql/migrations/patch-1.sql
+++ b/azkaban-db/src/main/sql/migrations/patch-1.sql
@@ -1,0 +1,3 @@
+-- DB Migration from release 3.83.0
+-- Increasing the side of json column in project_flows table.
+ALTER TABLE project_flows MODIFY COLUMN json MEDIUMBLOB;
--- a/azkaban-db/src/main/sql/migrations/patch-1.sql
+++ b/azkaban-db/src/main/sql/migrations/patch-1.sql
@@ -1,3 +1,3 @@
 -- DB Migration from release 3.83.0
--- Increasing the side of json column in project_flows table.
+-- Increasing the size of json column in project_flows table.
 ALTER TABLE project_flows MODIFY COLUMN json MEDIUMBLOB;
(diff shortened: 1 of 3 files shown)


azkaban-db/src/main/sql/migrations/patch-1.sql
```diff
@@ -0,0 +1,3 @@
+-- DB Migration from release 3.83.0
+-- Increasing the side of json column in project_flows table.
```
sshardool: Nit: `size`

azkaban-db/src/main/sql/migrations/patch-1.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE project_flows MODIFY COLUMN json MEDIUMBLOB;
```
sshardool: Should we also change the CREATE TABLE statement as part of this?

jakhani: We should not. That is how patch should work.

azkaban-db/src/main/sql/migrations/patch-1.sql
jakhani: Good catch. Fixed it.

azkaban-db/src/main/sql/migrations/patch-1.sql
sshardool: Not sure how it has been done historically in Azkaban. 
What happens when a new Azkaban cluster is being initialized ? Is the schema created by always running the original schema creation along with all the patches. Does not seem like a good way of doing this. 
(Not suggesting it needs to be changed at this point - only confirming the current state of things)

jakhani: Totally agree with you. There is no proper process right now. But it will be good to follow some standards for database patches. We can discuss about it and implement it.

ypadron-in: I believe that in cases of schema modifications there is currently a convention of adding them to a file with name like "upgrade.3.xx.0.to.3.yy.0" and since there no mechanism in place for applying changes in those files to newly created DBs automatically we also add the schema change to the create.<table-name>.sql files.

PR #2473, opened 2020-03-16
Moved the db clean up logic out of the push based dispatch-logic

svnarumugam:
Extracted the execution_logs cleanup logic from old dispatcher logic and kept it outside the dispatcher. Independent of the dispatcher logic this clean up would run every 60 minutes. 

Tested manually that the logic is triggered by setting the azkaban.poll.model config to be false and true. I was able to see the ExecutionLogsCleaner getting triggered. Adding unit tests is in progress.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -74,6 +74,7 @@ public class Constants {
   public static final int DEFAULT_SSL_PORT_NUMBER = 8443;
   public static final int DEFAULT_JETTY_MAX_THREAD_COUNT = 20;
 
+
   // One Schedule's default End Time: 01/01/2050, 00:00:00, UTC
   public static final long DEFAULT_SCHEDULE_END_EPOCH_TIME = 2524608000000L;
 
@@ -325,6 +326,10 @@ public class Constants {
     public static final String AZKABAN_RAMP_STATUS_POLLING_INTERVAL = "azkaban.ramp.status.polling.interval";
     public static final String AZKABAN_RAMP_STATUS_POLLING_CPU_MAX = "azkaban.ramp.status.polling.cpu.max";
     public static final String AZKABAN_RAMP_STATUS_POLLING_MEMORY_MIN = "azkaban.ramp.status.polling.memory.min";
+
+    public static final String EXECUTION_LOGS_RETENTION_MS = "execution.logs.retention.ms";
+    public static final String EXECUTION_LOGS_CLEANUP_INTERVAL_SECONDS =
+        "execution.logs.cleanup.interval.seconds";
   }
 
   public static class FlowProperties {
--- a/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
@@ -146,6 +146,7 @@ public class AzkabanWebServer extends AzkabanServer implements IMBeanRegistrable
   private final FlowTriggerScheduler scheduler;
   private final FlowTriggerService flowTriggerService;
   private Map<String, TriggerPlugin> triggerPlugins;
+  private final ExecutionLogsCleaner executionLogsCleaner;
 
   @Inject
   public AzkabanWebServer(final Props props,
@@ -160,7 +161,8 @@ public class AzkabanWebServer extends AzkabanServer implements IMBeanRegistrable
       final VelocityEngine velocityEngine,
       final FlowTriggerScheduler scheduler,
       final FlowTriggerService flowTriggerService,
-      final StatusService statusService) {
+      final StatusService statusService,
+      final ExecutionLogsCleaner executionLogsCleaner) {
     this.props = requireNonNull(props, "props is null.");
     this.server = requireNonNull(server, "server is null.");
     this.executorManagerAdapter = requireNonNull(executorManagerAdapter,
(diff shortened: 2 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionLogsCleaner.java
```diff
@@ -0,0 +1,69 @@
...
+      this.scheduler.scheduleAtFixedRate(() -> cleanExecutionLogs(), 0L, cleanupIntervalInMinutes,
+          TimeUnit.MINUTES);
+   }
+
+   private void cleanExecutionLogs() {
+      logger.info("Cleaning old logs from execution_logs");
+      final long cutoff = System.currentTimeMillis() - this.executionLogsRetentionMs;
+      logger.info("Cleaning old log files before "
+          + new DateTime(cutoff).toString());
+      cleanOldExecutionLogs(System.currentTimeMillis()
```
li-ygerchikov: Can you use the `cutoff` calculated above? If more than one millisecond passes between lines 49 and 52, the log will misstate the cutoff time. (Yes, I know that's how it was before your changes. Still, fix or leave as is?)

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsCleaner.java
```diff
@@ -0,0 +1,69 @@
...
+   private void cleanOldExecutionLogs(final long millis) {
+      final long beforeDeleteLogsTimestamp = System.currentTimeMillis();
+      try {
+         final int count = this.executorLoader.removeExecutionLogsByTime(millis);
+         logger.info("Cleaned up " + count + " log entries.");
+      } catch (final Exception e) {
+         logger.error("log clean up failed. ", e);
+      }
+      logger.info(
+          "log clean up time: " + (System.currentTimeMillis() - beforeDeleteLogsTimestamp) / 1000
```
li-ygerchikov: This is integer division. Will it result in `0  seconds` for anything under 1000 ms? (Also understood this is not new.)

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -146,6 +147,7 @@
   private final FlowTriggerScheduler scheduler;
   private final FlowTriggerService flowTriggerService;
   private Map<String, TriggerPlugin> triggerPlugins;
+  private final ExecutionLogsCleaner executionLogsCleaner;
 
   @Inject
   public AzkabanWebServer(final Props props,
```
li-ygerchikov: Does it matter that we are changing the Ctor signature?

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsCleaner.java
```diff
@@ -0,0 +1,69 @@
+package azkaban.executor;
+
+import azkaban.utils.Props;
+import java.util.concurrent.Executors;
```
li-ygerchikov: Thanks for switching to these -- it's much cleaner and concise this way!

svnarumugam:
Added unit test and addressed the earlier comments.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -325,6 +326,10 @@
     public static final String AZKABAN_RAMP_STATUS_POLLING_INTERVAL = "azkaban.ramp.status.polling.interval";
     public static final String AZKABAN_RAMP_STATUS_POLLING_CPU_MAX = "azkaban.ramp.status.polling.cpu.max";
     public static final String AZKABAN_RAMP_STATUS_POLLING_MEMORY_MIN = "azkaban.ramp.status.polling.memory.min";
+
+    public static final String EXECUTION_LOGS_RETENTION_MS = "execution.logs.retention.ms";
+    public static final String EXECUTION_LOGS_CLEANUP_INTERVAL_SECONDS = "execution.logs.cleanup"
```
li-ygerchikov: Thanks for making these string constants, but please try not to break strings like that -- that makes them unsearchable or at least less findable. Specifically, if I saw `execution.log.cleanup.interval.seconds` in a config file and wanted to figure out how it is used in code, I would not be able to (easily) find any occurrences.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -231,6 +233,8 @@ public static void launch(final AzkabanWebServer webServer) throws Exception {
 
     webServer.executorManagerAdapter.start();
 
+    webServer.executionLogsCleaner.start();
```
li-ygerchikov: Since this is plugged specifically into `AzkabanWebServer`, could you please verify that it will also work properly in `AzkabanSingleServer`? I think it should, since solo server effectively includes web server, but can we double-check just to make sure? Thanks!

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsCleaner.java
```diff
@@ -0,0 +1,69 @@
+package azkaban.executor;
```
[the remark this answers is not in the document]
svnarumugam: My initial thought process was to move the code out and use it in ExecutorManager and ExecutionController. That is why I had placed it here. Now that it runs independently we needn't keep it here. Moved it to web server module. 

My earlier comments were lost because I typed comments and resolve in github thinking that those comments will be recorded!

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
svnarumugam: It doesn't matter. It will just be fine.

az-core/src/main/java/azkaban/Constants.java
svnarumugam: I see your point. Right now if I have to find the occurrence of a key in config it can be a two-step work. 1. Find the const that maps to this string 2. the find the uses of that const. Because I had to use the same const in the test file also I defined it here so that it's not duplicated.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsCleaner.java
svnarumugam: Thank you for your quick reviews Yuliy!

PR #2474, opened 2020-03-18
Fetch flow execution status from database column only.

ypadron-in:
Currently flow execution status is stored in 2 places: a column in the DB and in the execution's JSON object. Because flow execution status is not read from the same source all the times sometimes we see inconsistency in the UI where different pages show different flow execution status.

This PR addresses this inconsistency by fetching the execution status from DB everywhere in the code.

### DIFF
--- a/az-core/src/main/java/azkaban/utils/TimeUtils.java
+++ b/az-core/src/main/java/azkaban/utils/TimeUtils.java
@@ -16,7 +16,9 @@
 package azkaban.utils;
 
 import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import org.joda.time.Days;
@@ -61,6 +63,16 @@ public class TimeUtils {
     final ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(Instant.ofEpochMilli(timestampMs),
         ZoneId.systemDefault());
     return formatter.format(zonedDateTime);
+  }
+
+  /**
+   * Takes a date string formatted as "yyyy-MM-dd HH:mm:ss" and converts it into milliseconds
+   * since the Epoch in UTC
+   */
+  public static long convertDateTimeToUTCMillis(final String dateTime) {
+    final DateTimeFormatter formatter = DateTimeFormatter.ofPattern(DATE_TIME_PATTERN);
+    final LocalDateTime parsedDate = LocalDateTime.parse(dateTime, formatter);
+    return parsedDate.atZone(ZoneOffset.UTC).toInstant().toEpochMilli();
   }
 
   /**
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -409,6 +409,9 @@ public class ExecutionFlowDao {
             final ExecutableFlow exFlow =
                 ExecutableFlow.createExecutableFlowFromObject(
                     GZIPUtils.transformBytesToObject(data, encType));
+
+            // read execution status from the DB column not from the flow data blob to maintain
+            // consistency and have a single source of truth
             Status status = Status.fromInteger(rs.getInt(4));
             exFlow.setStatus(status);
             execFlows.add(exFlow);
@@ -457,6 +460,8 @@ public class ExecutionFlowDao {
                 ExecutableFlow.createExecutableFlowFromObject(
                     GZIPUtils.transformBytesToObject(data, encType));
 
+            // read execution status from the DB column not from the flow data blob to maintain
+            // consistency and have a single source of truth
             Status status = Status.fromInteger(rs.getInt(4));
             exFlow.setStatus(status);
             final ExecutionReference ref = new ExecutionReference(id);
@@ -498,6 +503,9 @@ public class ExecutionFlowDao {
             final ExecutableFlow exFlow =
                 ExecutableFlow.createExecutableFlowFromObject(
                     GZIPUtils.transformBytesToObject(data, encType));
+
+            // read execution status from the DB column not from the flow data blob to maintain
+            // consistency and have a single source of truth
             Status status = Status.fromInteger(rs.getInt(4));
             exFlow.setStatus(status);
             execFlows.add(exFlow);
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -405,15 +405,11 @@ public class ExecutionFlowDao {
 
         if (data != null) {
           final EncodingType encType = EncodingType.fromInteger(encodingType);
+          final Status status = Status.fromInteger(rs.getInt(4));
           try {
             final ExecutableFlow exFlow =
-                ExecutableFlow.createExecutableFlowFromObject(
-                    GZIPUtils.transformBytesToObject(data, encType));
-
-            // read execution status from the DB column not from the flow data blob to maintain
-            // consistency and have a single source of truth
-            Status status = Status.fromInteger(rs.getInt(4));
-            exFlow.setStatus(status);
+                ExecutableFlow.createExecutableFlow(
+                    GZIPUtils.transformBytesToObject(data, encType), status);
             execFlows.add(exFlow);
           } catch (final IOException e) {
             throw new SQLException("Error retrieving flow data " + id, e);
@@ -455,15 +451,11 @@ public class ExecutionFlowDao {
           ExecutionFlowDao.logger.error("Found a flow with empty data blob exec_id: " + id);
         } else {
           final EncodingType encType = EncodingType.fromInteger(encodingType);
+          final Status status = Status.fromInteger(rs.getInt(4));
           try {
             final ExecutableFlow exFlow =
-                ExecutableFlow.createExecutableFlowFromObject(
-                    GZIPUtils.transformBytesToObject(data, encType));
-
-            // read execution status from the DB column not from the flow data blob to maintain
-            // consistency and have a single source of truth
-            Status status = Status.fromInteger(rs.getInt(4));
-            exFlow.setStatus(status);
+                ExecutableFlow.createExecutableFlow(
+                    GZIPUtils.transformBytesToObject(data, encType), status);
             final ExecutionReference ref = new ExecutionReference(id);
             execFlows.add(new Pair<>(ref, exFlow));
           } catch (final IOException e) {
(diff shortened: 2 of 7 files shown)


azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -232,13 +212,11 @@ public void testFetchRecentlyFinishedFlows() throws Exception {
...
-    flow1.setStatus(Status.SUCCEEDED);
-    flow1.setEndTime(DateTimeUtils.currentTimeMillis());
+    final ExecutableFlow flow1 = createExecution(System.currentTimeMillis(), Status.SUCCEEDED);
+    flow1.setEndTime(System.currentTimeMillis());
     this.executionFlowDao.updateExecutableFlow(flow1);
     //Todo jamiesjc: use java8.java.time api instead of jodatime
-
+    
     //Mock flow finished time to be 2 min ago.
     DateTimeUtils.setCurrentMillisOffset(-FLOW_FINISHED_TIME.toMillis());
```
svnarumugam: Did we consider the alternative
_long twoMinutesAgo = currentTime - FLOW_FINISHED_TIME.toMillis();
setEndTime(twoMinutesAgo)_ because it would be more readable?

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -586,11 +565,191 @@ public void testSelectAndUpdateExecutionWithSamePriority() throws Exception {
...
+
+    makeFlowStatusInconsistent(flow1.getExecutionId(), Status.FAILED);
+
+    final List<ExecutableFlow> finishedFlows2 = this.executionFlowDao.fetchRecentlyFinishedFlows(
+        RECENTLY_FINISHED_LIFETIME);
+    assertThat(finishedFlows2).isNotEmpty();
+    assertThat(finishedFlows2.get(0).getStatus()).isEqualTo(Status.FAILED);
+  }
+
+  private void makeFlowStatusInconsistent(int executionId, Status status) {
```
svnarumugam: This method caught my attention because it says "make the status inconsistent".  While it's true that we got the inconsistency because of status column being updated independent of the object we store, we may consider renaming this or add a comment? :)

az-core/src/main/java/azkaban/utils/TimeUtils.java
```diff
@@ -63,6 +65,16 @@ private static String format(final long timestampMs, final String pattern) {
     return formatter.format(zonedDateTime);
   }
 
+  /**
+   * Takes a date string formatted as "yyyy-MM-dd HH:mm:ss" and converts it into milliseconds
+   * since the Epoch in UTC
+   */
+  public static long convertDateTimeToUTCMillis(final String dateTime) {
```
svnarumugam: Thanks Yeni for extracting this to TimeUtils!

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -175,24 +177,24 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
...
     private static final String FETCH_ACTIVE_EXECUTABLE_FLOWS =
-        "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-            + "et.port port, ex.executor_id executorId, et.active executorStatus"
+        "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, ex.status status,"
+            + " et.host host, et.port port, ex.executor_id executorId, et.active executorStatus"
             + " FROM execution_flows ex"
             + " LEFT JOIN "
             + " executors et ON ex.executor_id = et.id"
-            + " Where ex.status NOT IN ("
+            + " WHERE ex.status NOT IN ("
```
jakhani: Thank you for taking care of query standards.

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -266,12 +268,12 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
 
     // Select the flow that is dispatched and not in finished status by execution id
     private static final String FETCH_ACTIVE_EXECUTABLE_FLOW_BY_EXEC_ID =
-        "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, et.host host, "
-            + "et.port port, ex.executor_id executorId, et.active executorStatus"
+        "SELECT ex.exec_id exec_id, ex.enc_type enc_type, ex.flow_data flow_data, ex.status status,"
```
jakhani: If possible, add new column in the end. That way, you won't need to worry about correcting indexes of rest of the fields.

ypadron-in: We're accessing the columns by name when processing the results of this query so it's safe to add the column in the middle. 
We should probably use names in the rest of the queries too.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
ypadron-in: thanks, I added a comment.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
ypadron-in: Fixed it.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -158,23 +158,22 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
+  List<ExecutableFlow> fetchFlowHistory(final String projectNameContains,
+      final String flowNameContains, final String userNameContains, final int status,
+      final long startTime, final long endTime, final int skip, final int num)
       throws ExecutorManagerException {
     String query = FetchExecutableFlows.FETCH_BASE_EXECUTABLE_FLOW_QUERY;
     final List<Object> params = new ArrayList<>();
 
     boolean first = true;
-    if (projContain != null && !projContain.isEmpty()) {
+    if (projectNameContains != null && !projectNameContains.isEmpty()) {
```
sshardool: Nice, updated variable names look much better.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -410,6 +409,8 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
             final ExecutableFlow exFlow =
                 ExecutableFlow.createExecutableFlowFromObject(
                     GZIPUtils.transformBytesToObject(data, encType));
+            Status status = Status.fromInteger(rs.getInt(4));
+            exFlow.setStatus(status);
```
sshardool: Maybe add an explanation in the comments for why the `status` specifically is overwritten by the value from DB column.

ypadron-in:
> If I understand correctly then:
> (1) This change will not make the status serialized ExecutionFlow object consistent with status from the DB column
> (2) It makes sure the value from the DB column is returned whenever one of the `flowHistory...` methods in the DAO are called>
> Is my understanding correct ?
You're correct this change is to read execution status from DB in all the places its currently read and not to make DB and blob status the same. Under normal circumstances these 2 statuses will be the same, only under exceptional circumstances, for example there is a bug in the code, we're in the need to change DB status directly.
A derived change to this one which will prevent inconsistencies all together is to remove status property from the flow data blob, that way there is only **one** status.
The status is read whenever ExecutableFlow object is fetched so calls to read status come from a lot of places not just when fetching the flow history.

> Also - how do we ensure that the value from the DB column is the most accurate one 
Here is the query to update an execution
`UPDATE execution_flows SET status=?, update_time=?, start_time=?, end_time=?, enc_type=?, flow_data=? WHERE exec_id=?`
Every time we update the status column as well as the blob (flow_data) so under normal circumstances it should not happen that one is more accurate than the other one, only when we manually intervene states become inconsistent

> and all the reads of the execution status happen through the methods modified here ?
I can't think of a way to ensure that as long as the status is saved in two different places. 
I think I covered all the places when status is read but if I missed one we'll know soon enough :).

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: done

djaiswal83: This is still bug prone. I see that there are 4 instances of such calls. Why not pass in the status as a parameter to createExecutableFlowFromObject and encapsulate the logic within? This way, any new future call will remain consistent.

ypadron-in:
> Thanks for making this change and my apologies for reviewing it so late.
> I have left a change request.
> Overall, this is a step in the right direction and perhaps, this PR is good enough to go in.
> However, we haven't mitigated the problem yet.
> The inherent problem with keeping status at two different places is that it is difficult to maintain and often causes problems we are facing right now.
> The complete solution (perhaps as a followup PR) would be to rid the status from blob. The status should always be read from one and only one source.
> This PR overwriting it makes the entry in the blob useless anyway.

Thanks for reviewing! I agree with you (see my comment above yours here: https://github.com/azkaban/azkaban/pull/2474#issuecomment-601878470) that a follow up PR is needed to remove status property from the flow data blob, we should **not** store status in 2 different places. I can create a GitHub issue for a follow up PR. 
In fact, we should also create a GitHub issue to fix job execution statuses because the same thing happens there: status is stored in a DB column and in the flow data blob. We experience less inconsistencies with job statuses but there are multiple source of truth for them as well.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: I made the change, thanks.

sshardool:
> > If I understand correctly then:
> > (1) This change will not make the status serialized ExecutionFlow object consistent with status from the DB column
> > (2) It makes sure the value from the DB column is returned whenever one of the `flowHistory...` methods in the DAO are called>
> > Is my understanding correct ?
> 
> You're correct this change is to read execution status from DB in all the places its currently read and not to make DB and blob status the same. Under normal circumstances these 2 statuses will be the same, only under exceptional circumstances, for example there is a bug in the code, we're in the need to change DB status directly.
> A derived change to this one which will prevent inconsistencies all together is to remove status property from the flow data blob, that way there is only **one** status.
> The status is read whenever ExecutableFlow object is fetched so calls to read status come from a lot of places not just when fetching the flow history.
> 
> > Also - how do we ensure that the value from the DB column is the most accurate one
> 
> Here is the query to update an execution
> `UPDATE execution_flows SET status=?, update_time=?, start_time=?, end_time=?, enc_type=?, flow_data=? WHERE exec_id=?`
> Every time we update the status column as well as the blob (flow_data) so under normal circumstances it should not happen that one is more accurate than the other one, only when we manually intervene states become inconsistent
> 
> > and all the reads of the execution status happen through the methods modified here ?
> 
> I can't think of a way to ensure that as long as the status is saved in two different places.
> I think I covered all the places when status is read but if I missed one we'll know soon enough :).

Thanks for the detailed response. I agree that a better longer term approach would be to maintain the status at just one place. 
I am a still concerned that a  number of `ExecutableNode.setStatus` calls in our code may not be updating the status in the db. (There are ~30 of them and not immediately clear if all of them are making it to the database). However this change is improving the current situation at least. 
Could you please create a ticket for the longer term fix which should consider removing the status from the blob please ?
I will approve the change.

PR #2479, opened 2020-03-27
Add remediation for flows assigned to unreachable executors

sshardool:
As part of the remediation flows in an `unfinished` state assinged to an unreachable executor are finalized

Additional changes:
- Adds healthcheck alert as a first-class member of alert related interfaces
   - Current code was repurposing Flow Failure related alert mechanism in a less
     than ideal manner
- Customized email format for healthcheck alert message
   - Assumes that only a `single` kind of healthcheck alert message is required.
     But makes future extensions easier, if required.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
@@ -97,12 +97,7 @@ public class ExecutorHealthChecker {
       final Optional<Executor> executorOption = entry.getKey();
       if (!executorOption.isPresent()) {
         final String finalizeReason = "Executor id of this execution doesn't exist.";
-        for (final ExecutableFlow flow : entry.getValue()) {
-          logger.warn(
-              String.format("Finalizing execution %s, %s", flow.getExecutionId(), finalizeReason));
-          ExecutionControllerUtils
-              .finalizeFlow(this.executorLoader, this.alerterHolder, flow, finalizeReason, null);
-        }
+        finalizeFlows(entry.getValue(), finalizeReason);
         continue;
       }
 
@@ -124,8 +119,23 @@ public class ExecutorHealthChecker {
           }
         }
       } catch (final ExecutorManagerException e) {
-        handleExecutorNotAliveCase(entry, executor, e);
+        handleExecutorNotAliveCase(executor, entry.getValue(), e);
       }
+    }
+  }
+
+  /**
+   * Finalize given flows with the provided reason.
+   *
+   * @param flows
+   * @param finalizeReason
+   */
+  private void finalizeFlows(List<ExecutableFlow> flows, String finalizeReason) {
+    for (ExecutableFlow flow: flows) {
+      logger.warn(
+          String.format("Finalizing execution %s, %s", flow.getExecutionId(), finalizeReason));
+      ExecutionControllerUtils
+          .finalizeFlow(this.executorLoader, this.alerterHolder, flow, finalizeReason, null);
     }
   }
 
@@ -155,25 +165,47 @@ public class ExecutorHealthChecker {
 
   /**
    * Increments executor failure count. If it reaches max failure count, sends alert emails to AZ
-   * admin.
+   * admin and executes any cleanup actions for flows on that executors.
    *
-   * @param entry executor to list of flows map entry
    * @param executor the executor
+   * @param flows flows assigned to the executor
    * @param e Exception thrown when the executor is not alive
    */
-  private void handleExecutorNotAliveCase(
-      final Entry<Optional<Executor>, List<ExecutableFlow>> entry, final Executor executor,
+  private void handleExecutorNotAliveCase(final Executor executor, final List<ExecutableFlow> flows,
       final ExecutorManagerException e) {
     logger.error("Failed to get update from executor " + executor.getId(), e);
     this.executorFailureCount.put(executor.getId(), this.executorFailureCount.getOrDefault
         (executor.getId(), 0) + 1);
-    if (this.executorFailureCount.get(executor.getId()) % this.executorMaxFailureCount == 0
-        && !this.alertEmails.isEmpty()) {
-      entry.getValue().stream().forEach(flow -> flow
-          .getExecutionOptions().setFailureEmails(this.alertEmails));
-      logger.info(String.format("Executor failure count is %d. Sending alert emails to %s.",
-          this.executorFailureCount.get(executor.getId()), this.alertEmails));
-      this.alerterHolder.get("email").alertOnFailedUpdate(executor, entry.getValue(), e);
+    if (this.executorFailureCount.get(executor.getId()) % this.executorMaxFailureCount == 0) {
+      if (!this.alertEmails.isEmpty()) {
+        logger.info(String.format("Executor failure count is %d. Sending alert emails to %s.",
+            this.executorFailureCount.get(executor.getId()), this.alertEmails));
+        this.alerterHolder.get("email")
+            .alertOnFailedExecutorHealthCheck(executor, flows, e,
+                this.alertEmails);
+      }
+      this.cleanupForMissingExecutor(executor, flows);
     }
+  }
+
+  /**
+   * Perform any cleanup required for an unreachable executor.
+   *
+   * Note that ideally we would like to disable the executor such that not further executions are
+   * 'assigned' to it. However with the pull/polling based model there is currently no direct way
+   * of doing this other than the hitting the corresponding ajax endpoint for the executor.
+   * That endpoint is most likely not reachable (hence the repeated healthcheck failures).
+   * Updating the active status or removing the executor from db will not have an impact on any
+   * executor that is still alive and was unreachable temporarily.
+   * For now we limit the action to finalizing any flows assigned to the the unreachable executor.
+   *
+   * @param executor
+   * @param executions
+   */
+  private void cleanupForMissingExecutor(Executor executor, List<ExecutableFlow> executions) {
+    String finalizeReason =
+        String.format("Executor was unreachable, executor-id: %s, executor-host: %s, "
+                + "executor-port: %d", executor.getId(), executor.getHost(), executor.getPort());
+    finalizeFlows(executions, finalizeReason);
   }
 }
--- a/azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
+++ b/azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
@@ -273,21 +273,65 @@ public class DefaultMailCreator implements MailCreator {
       message.println("");
       message.println("<h3>Affected executions</h3>");
       message.println("<ul>");
-      for (final ExecutableFlow flow : flows) {
-        final int execId = flow.getExecutionId();
-        final String executionUrl =
-            scheme + "://" + clientHostname + ":" + clientPortNumber + "/"
-                + "executor?" + "execid=" + execId;
-
-        message.println("<li>Execution '" + flow.getExecutionId() + "' of flow '" + flow.getFlowId()
-            + "' of project '" + flow.getProjectName() + "' - " +
-            " <a href=\"" + executionUrl + "\">Execution Link</a></li>");
-      }
-
+      appendFlowLinksToMessage(message, flows, scheme, clientHostname, clientPortNumber);
       message.println("</ul>");
       return true;
     }
 
     return false;
+  }
+
+  @Override
+  public boolean createFailedExecutorHealthCheckMessage(final List<ExecutableFlow> flows,
+      final Executor executor, final ExecutorManagerException failureException,
+      final EmailMessage message, final String azkabanName,
+      final String scheme, final String clientHostname, final String clientPortNumber,
+      final List<String> emailList) {
+
+    if (emailList == null || emailList.isEmpty()) {
+      return false;
+    }
+    message.addAllToAddress(emailList);
+    message.setMimeType("text/html");
+    message.setSubject(
+        "Alert: Executor is unreachable, " + executor.getHost() + " on " + azkabanName);
+
+    message.println(
+        "<h2 style=\"color:#FFA500\"> Executor is unreachable. Executor host - " + executor
+            .getHost() + " on Cluster - " + azkabanName + "</h2>");
+
+    message.println("Remedial action will be attempted on affected executions - <br>");
+    message.println("Following flows were reported as running on the executor and will be "
+        + "finalized.");
+
+    message.println("");
+    message.println("<h3>Affected executions</h3>");
+    message.println("<ul>");
+    appendFlowLinksToMessage(message, flows, scheme, clientHostname, clientPortNumber);
+    message.println("</ul>");
+
+    message.println("");
+    message.println("<h3>Error detail</h3>");
+    message.println(String.format("Following error was reported for executor-id: %s, "
+            + "executor-host: %s, executor-port: %d", executor.getId(), executor.getHost(),
+        executor.getPort()));
+    message.println("<pre>" + ExceptionUtils.getStackTrace(failureException) + "</pre>");
+
+    return true;
+  }
+
+  private void appendFlowLinksToMessage(final EmailMessage message,
+      final List<ExecutableFlow> flows,
+      final String scheme, final String clientHostname, final String clientPortNumber) {
+    for (final ExecutableFlow flow : flows) {
+      final int execId = flow.getExecutionId();
+      final String executionUrl =
+          scheme + "://" + clientHostname + ":" + clientPortNumber + "/"
+              + "executor?" + "execid=" + execId;
+
+      message.println("<li>Execution '" + flow.getExecutionId() + "' of flow '" + flow.getFlowId()
+          + "' of project '" + flow.getProjectName() + "' - " +
+          " <a href=\"" + executionUrl + "\">Execution Link</a></li>");
+    }
   }
 }
--- a/azkaban-common/src/main/java/azkaban/utils/Emailer.java
+++ b/azkaban-common/src/main/java/azkaban/utils/Emailer.java
@@ -189,6 +189,28 @@ public class Emailer extends AbstractMailer implements Alerter {
   }
 
   /**
+   * Use the default mail creator to send a failed executor healthcheck message to the given list
+   * of addresses. Message includes a list of flows impacted on the executor.
+   */
+  @Override
+  public void alertOnFailedExecutorHealthCheck(Executor executor, List<ExecutableFlow> flows,
+      ExecutorManagerException failureException, List<String> emailList) {
+    if (emailList == null || emailList.isEmpty()) {
+      // We should consider throwing an exception here. For now this follows the model of the rest
+      // of the file and simply returns.
+      logger.error("No email list specified for failed health check alert");
+      return;
+    }
+    MailCreator mailCreator = DefaultMailCreator.getCreator(DefaultMailCreator.DEFAULT_MAIL_CREATOR);
+    final EmailMessage message = this.messageCreator.createMessage();
+    final boolean mailCreated = mailCreator
+        .createFailedExecutorHealthCheckMessage(flows, executor, failureException, message,
+            this.azkabanName, this.scheme, this.clientHostname, this.clientPortNumber, emailList);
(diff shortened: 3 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -169,11 +179,32 @@ private void handleExecutorNotAliveCase(
...
+   *
+   * @param executor
+   * @param executions
+   */
+  private void cleanupForMissingExecutor(Executor executor, List<ExecutableFlow> executions) {
+    String finalizeReason = "Executor was unreachable, executor id: " + executor.getId();
+    finalizeFlows(executions, finalizeReason);
+
+    try {
+      this.executorLoader.removeExecutor(executor.getHost(), executor.getPort());
```
djaiswal83: Can you add a log line for removing executor adding the reason? I looked up removeExecutor and it does not log it.

azkaban-common/src/test/resources/azkaban/executor/mail/faileExecutorHealthCheckMessage.html
```diff
@@ -0,0 +1,12 @@
+<h2 style="color:#FFA500"> Executor is unreachable. Executor host - executor1-host on Cluster - unit-tests</h2>
```
djaiswal83: The file name has typo. I think it is supposed to start with "failed" instead of "faile"

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
```diff
@@ -273,21 +273,62 @@ public boolean createFailedUpdateMessage(final List<ExecutableFlow> flows,
...
+      appendFlowLinksToMessage(message, flows, scheme, clientHostname, clientPortNumber);
       message.println("</ul>");
       return true;
     }
 
     return false;
   }
+
+  @Override
+  public boolean createFailedExecutorHealthCheckMessage(final List<ExecutableFlow> flows,
```
djaiswal83: Where are we using this?

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -169,11 +179,32 @@ private void handleExecutorNotAliveCase(
...
+   * Perform any cleanup required for an unreachable executor.
+   * Currently this accomplishes:
+   *   (1) Finalize the flow executions with appropriate reason
+   *   (2) Remove the corresponding executor entries from the data store.
+   *
+   * @param executor
+   * @param executions
+   */
+  private void cleanupForMissingExecutor(Executor executor, List<ExecutableFlow> executions) {
+    String finalizeReason = "Executor was unreachable, executor id: " + executor.getId();
```
ypadron-in, line 200: Would the executor hostname improve readability of the `finalizeReason`?

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -169,11 +179,32 @@ private void handleExecutorNotAliveCase(
...
     if (this.executorFailureCount.get(executor.getId()) % this.executorMaxFailureCount == 0
         && !this.alertEmails.isEmpty()) {
-      entry.getValue().stream().forEach(flow -> flow
-          .getExecutionOptions().setFailureEmails(this.alertEmails));
       logger.info(String.format("Executor failure count is %d. Sending alert emails to %s.",
           this.executorFailureCount.get(executor.getId()), this.alertEmails));
-      this.alerterHolder.get("email").alertOnFailedUpdate(executor, entry.getValue(), e);
+      this.alerterHolder.get("email").alertOnFailedExecutorHealthCheck(executor, entry.getValue(), e,
+          this.alertEmails);
+      this.cleanupForMissingExecutor(executor, entry.getValue());
```
ypadron-in, line 186: Should we make auto-remediation optional through a configuration variable?

sshardool:
> I have some minor comments related to typos and logging.
> 
> Holistically, we are not solving anything as far as user is concerned. The flow would still fail.
> Is it not possible to put the flows back in queue so that they can be picked up by existing active executors?

We explicitly decided not to attempt a rerun of the flow as with a lost executor we don't quite know the execution status of the child jobs and based on the business logic of the flow it may not be correct to run the flow again.

azkaban-common/src/test/resources/azkaban/executor/mail/faileExecutorHealthCheckMessage.html
sshardool: Will update, thanks for catching this.

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
sshardool: Will fix.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Makes sense. Will add.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Will do.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: I thought about having a config for this originally but started wondering if there would be a good reason for someone to not attempt remediation (mainly the removal of the executor from db).
Configs like these are also have the side effect either resulting a in large number of tests that need to be maintained for different config values or inadequate testing for some config value(s).

I am open to the idea however. Is there any notion of `experimental` configs followed in Azkaban where a new feature is introduced with a config but after a sufficient 'soak time' (in production) the configuration is removed ?

ypadron-in, line 186: The conditional indicates that if no alert emails are configured then `cleanupForMissingExecutor` will not happen. Is that what we want?

sshardool: My initial thought was that we should not perform a cleanup silently without notifying any admins.
But on a closer inspection it makes sense to do so becuase:
- The corresponding message is being logged
- Flow failure emails will still go out to flow admins
- Keeps the behavior more consistent and predictable with or without admin email configured.
Thanks for pointing this out. Will make the change.

sshardool:
> `handleExecutorNotAliveCase()` receives 2 parameters `Entry<Optional<Executor>, List<ExecutableFlow>>` and Executor. Since Executor info seems a bit redundant consider passing just and `Entry` or 2 parameters:` List<ExecutableFlow` and `Executor`.
> 
> Also, could you update the comment of this method to reflect the new cleanup step?

Good suggestion. Will make the change.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -155,25 +165,47 @@ public void checkExecutorHealth() {
...
+      }
+      this.cleanupForMissingExecutor(executor, flows);
     }
   }
+
+  /**
+   * Perform any cleanup required for an unreachable executor.
+   *
+   * Note that ideally we would like to disable the executor such that not further executions are
+   * 'assigned' to it. However with the pull/polling based model there is currently no direct way
```
anish-mal, line 195: Is there a case, that's not a not a false positive, and where an executor continues to poll even though we actually detect it to be unhealthy?

azkaban-common/src/main/java/azkaban/utils/Emailer.java
```diff
@@ -188,6 +188,24 @@ public void alertOnFailedUpdate(final Executor executor, List<ExecutableFlow> fl
     }
   }
 
+  @Override
```
anish-mal, line 191: Please add a function-level description of what this function does.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: In theory it's possible but would be quite unlikely in practice (I am not aware of an instance of this happening).
A made-up example being a temporary network issue where the web server is unable to reach an executor but both the web server as well as the executor are able to reach the mysql db. 
(The current behavior of not disabling the executor should be acceptable in those cases as well)

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -97,12 +97,7 @@ public void checkExecutorHealth() {
       final Optional<Executor> executorOption = entry.getKey();
       if (!executorOption.isPresent()) {
         final String finalizeReason = "Executor id of this execution doesn't exist.";
-        for (final ExecutableFlow flow : entry.getValue()) {
-          logger.warn(
-              String.format("Finalizing execution %s, %s", flow.getExecutionId(), finalizeReason));
-          ExecutionControllerUtils
-              .finalizeFlow(this.executorLoader, this.alerterHolder, flow, finalizeReason, null);
-        }
+        finalizeFlows(entry.getValue(), finalizeReason);
```
anish-mal, line 100: I feel the function renaming is ling overdue:

finalizeFlowsIfExecutorIsUnhealthy() will better reflect what the function does.

Also - midway in this function there is a ToDo from Jamie requesting a metric for Http response times? Do you know what is that for?

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -124,11 +119,26 @@ public void checkExecutorHealth() {
...
 
+  /**
+   * Finalize given flows with the provided reason.
+   *
+   * @param flows
+   * @param finalizeReason
+   */
+  private void finalizeFlows(List<ExecutableFlow> flows, String finalizeReason) {
+    for (ExecutableFlow flow: flows) {
+      logger.warn(
```
anish-mal, line 135: Does it make sense to put the body of the loop in a try-catch in order to continue the loop in case of a runtime exception.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: The private method `finalizeFlows` was added by me as part of this PR as a convenience wrapper over `ExecutionControllerUtils.finalizeFlows`. 
The purpose of the method is specifically to finalize any flows with the given reason. It does not take into account whether there was an error specific to the executor ( a higher level method will make that judgement) 
So I would prefer to keep the method name unchanged. Let me know your thoughts.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Good point.  I thought about exception handling for this. The underlying utility function `finalizeFlow` already captures any `ExecutorManagerExceptions` (which is not a RuntimeException). That is the most likely source of exceptions over here and with that already handled it made sense to pass on any unexpected RuntimeExceptions to the callers as soon as possible rather than swallowing it (even if logged).

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: The todo from Jamie seems to be talking about measuring the response time for the healthcheck. That would be valuable, unless we are already capturing executor response latencies as some other metric.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: The behavior has been modified based on feedback and this is not needed anymore.

azkaban-common/src/main/java/azkaban/executor/mail/DefaultMailCreator.java
sshardool: Used in Emailer. Fixed.

azkaban-common/src/main/java/azkaban/utils/Emailer.java
sshardool: Done.

azkaban-common/src/test/resources/azkaban/executor/mail/faileExecutorHealthCheckMessage.html
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Fixed.

anish-mal:
> > I have some minor comments related to typos and logging.
> > Holistically, we are not solving anything as far as user is concerned. The flow would still fail.
> > Is it not possible to put the flows back in queue so that they can be picked up by existing active executors?
> 
> We explicitly decided not to attempt a rerun of the flow as with a lost executor we don't quite know the execution status of the child jobs and based on the business logic of the flow it may not be correct to run the flow again.

It will be helpful to capture this in a comment.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
anish-mal: I had meant renaming the function checkExecutorHealth(). 
Two suggestions being 
- finalizeFlowsIfExecutorIsUnhealthy()
- finalizeFlowsForUnhealthyExecutor()

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
anish-mal: Not catching a Runtime exception means that we will skip finalizing any subsequent flows. That's the decision you have to make - whether to "miss" finalizing of subsequent Flow or not.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
anish-mal: I have had switch-controlled-feature-introduction on my mind. It requires two steps. First, there is an on-off switch/flag/config for a feature we introduce. Later, in step 2, after sufficient soaking the flag is removed. The drawback is the second step of removing the flag is a manual step, and requires code change as well. 

What approach did you have in mind?

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
anish-mal: Can it happen that subsequently the flow successfully finishes, and the executor updates its status?

In this case will the status go from FAILED --> SUCCESS.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: I had a similar idea in mind but with a few caveats.
(1) Configs introduced in this manner should have something within the config name to indicate that it will be removed in the near future and the only supported  behavior at that point will then be whatever is added as the `default` for the config
(2) This should still be done selectively for features of some magnitude, as it
- Puts an additional burden on someone to remove the config later, by then developers might have lost some context of the change and would make it even harder to debug any issues at that point
- Could result in a combinatorial explosion of tests that need to be maintained or some of the config values will remain untested if the config ends up staying long term

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Ah, got it now. However I feel the current name of the method is still more reflective of what it's intended to do. For instance something like `finalizeFlows..` does immediately reflect that an explicit check (in this case an ajax api call) is included as part of the method.

anish-mal: Doesn't finalizeFlowsIfExecutorIsUnhealthy() not tell the flows will be finalized if the Executor is unhealthy?

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Agree that's the current behavior. However making such a change would mean either:
(1) Any unreported exceptions (including significant ones like NPEs) will at the most simply be logged and things will continue business as usual. (This is part of the reason why we see an increased exceptions in our logs without having much in the way of tracking them)
(2) Alternatively such exceptions could be accumulated/consolidated and thrown to a higher layer but that has additional implementation and maintenance costs.
In this specific case both, the possibility of RuntimeExceptions at this point, and  the drawback of not finalizing subsequent flows seemed low enough that throwing the exception immediately made sense.

anish-mal: What will be the impact of getting a Runtime exception when the flow is being finalized:
- One, is that the flow and any subsequent flows yet to be iterated upon will not get finalized.
- Will some thread get killed or a JVM exit? (basically where will this exception be caught, because that's where things will get impacted)

That's the analysis we will have to weigh.

Also - you are bringing up a very good point (Thanks)

> (This is part of the reason why we see an increased exceptions in our logs without having much in the way of tracking them)

We should assume that people do monitor and get alerted to exception being printed in the log. I am trying to get this addressed (LIHADOOP-52504) internally for us.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Excellent point. 
- `ExecutionControllerUtils.finalizeFlow()` includes checks for validating that the flow should not already be in a final state.
- Flow status being set to SUCCEEDED during the normal execution of the flow also checks if the the flow is already in a final state. Relevant code `FlowRunner.finalizeFlow()`

(1) Both of the above checks make it unlikely that the two path for setting in the final status will step on each other (but there is still a possible race condition as the `testAndSet` is not atomic)
(2) In addition note that executor not being reachable from healthcheck but still running is expected to be a very rare situation. 
The combined probability of (1) and (2) would make this exceedingly unlikely but the probability is still non-zero.
Do note that this is still no worse than a similar remediation being performed manually.

sshardool:
> > > I have some minor comments related to typos and logging.
> > > Holistically, we are not solving anything as far as user is concerned. The flow would still fail.
> > > Is it not possible to put the flows back in queue so that they can be picked up by existing active executors?
> > 
> > 
> > We explicitly decided not to attempt a rerun of the flow as with a lost executor we don't quite know the execution status of the child jobs and based on the business logic of the flow it may not be correct to run the flow again.
> 
> It will be helpful to capture this in a comment.
Done.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: While reviewing existing code for this discovered larger issues with exception handling here. Creating a separate ticket for it and will address this as part of a separate PR.

anish-mal: Thanks. That was the purpose of my pushing you to do this exercise. It's ok to address this in a follow-on. Thanks again for the follow up. I appreciate it.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
sshardool: Decided to keep the current name for now. 
`finalizeFlows...` does not capture the checking and is more reflective of 'how' rather than 'what' the function is doing.  Think about what happens if/when the function starts doing more than just finalizing flows.
We could of course choose something like `checkExecutorHealthAndRemediateIfNeeded` :) but that seems long and unnecessary when a short comment in the function description will suffice.

PR #2480, opened 2020-03-30
If two projects have same flow name and ramp manager event is fired, …

jakhani:
…for the second event, it is getting NullPointerException as executingFlows has flowId as a key which is same for both the flows.

When first flow execution event comes, FlowRampManager is deleting the flow from executingFlows Map. Here the key is flow name. Reference for the same is [here](https://github.com/azkaban/azkaban/blob/56d0fd25cf5def07db12bec7ee745859574fbff0/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java#L647). When second event comes with the same flow name, ramp manager is throwing NullPointerException for ramp.toString() as ramp is null because during previous flow event processing, all the entries were deleted from executionFlows map. As this is the executing in a thread, it is not blocking main thread or raising any exception log as exception is not caught anywhere.


kxu1026:
LGTM. Good Catch!

PR #2481, opened 2020-04-01
Add more metrics to flow and job executions.

ypadron-in:
Metrics added:
-`flow-success-meter`: Records when flow executions are marked as SUCCEEDED.
-`flow-killed-meter`: Records when flow executions are marked as KILLED.
-`flow-killing-counter`: Tracks the number of flow executions in KILLING status (the starting point of the killing procedure).
-`flow-time-to-kill-histogram`: Records the time it takes to kill all the jobs in an execution, in other words the time it takes to transit from KILLING to KILLED status
-`job-success-meter`: Records when job executions are marked as SUCCEEDED.
-`job-fail-meter`: Records when job execution are marked as FAILED.
-`job-killed-meter`: Records when job executions are marked as KILLED.

Additional changes:
-fixed `project-dir-cache-hit-ratio`. Call the correct method at the time of emission for a Gauge metric.
-removed `queue-wait-histogram` until execution times (submission, start, update, end) are not affected by time differences between web and executor servers' machines.

### DIFF
--- a/az-core/src/main/java/azkaban/metrics/MetricsManager.java
+++ b/az-core/src/main/java/azkaban/metrics/MetricsManager.java
@@ -60,13 +60,10 @@ public class MetricsManager {
   }
 
   /**
-   * A {@link Meter} measures the rate of events over time (e.g., “requests per second”). Here we
-   * track 1-minute moving averages.
+   * A {@link Meter} measures the rate of events over time (e.g., “requests per second”).
    */
   public Meter addMeter(final String name) {
-    final Meter curr = this.registry.meter(name);
-    this.registry.register(name + "-gauge", (Gauge<Double>) curr::getOneMinuteRate);
-    return curr;
+    return this.registry.meter(name);
   }
 
   /**
--- a/azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
+++ b/azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
@@ -17,7 +17,6 @@
 package azkaban.metrics;
 
 import com.codahale.metrics.Counter;
-import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -38,7 +37,6 @@ public class CommonMetrics {
   public static final String SUBMIT_FLOW_FAIL_METER_NAME = "submit-flow-fail-meter";
   public static final String SUBMIT_FLOW_SKIP_METER_NAME = "submit-flow-skip-meter";
   public static final String OOM_WAITING_JOB_COUNT_NAME = "OOM-waiting-job-count";
-  public static final String QUEUE_WAIT_HISTOGRAM_NAME = "queue-wait-histogram";
   public static final String UPLOAD_FAT_PROJECT_METER_NAME = "upload-fat-project-meter";
   public static final String UPLOAD_THIN_PROJECT_METER_NAME = "upload-thin-project-meter";
 
@@ -54,7 +52,6 @@ public class CommonMetrics {
   private Meter submitFlowSkipMeter;
   private Meter uploadFatProjectMeter;
   private Meter uploadThinProjectMeter;
-  private Histogram queueWaitMeter;
 
   @Inject
   public CommonMetrics(final MetricsManager metricsManager) {
@@ -72,7 +69,6 @@ public class CommonMetrics {
     this.submitFlowFailMeter = this.metricsManager.addMeter(SUBMIT_FLOW_FAIL_METER_NAME);
     this.submitFlowSkipMeter = this.metricsManager.addMeter(SUBMIT_FLOW_SKIP_METER_NAME);
     this.OOMWaitingJobCount = this.metricsManager.addCounter(OOM_WAITING_JOB_COUNT_NAME);
-    this.queueWaitMeter = this.metricsManager.addHistogram(QUEUE_WAIT_HISTOGRAM_NAME);
     this.uploadFatProjectMeter = this.metricsManager.addMeter(UPLOAD_FAT_PROJECT_METER_NAME);
     this.uploadThinProjectMeter = this.metricsManager.addMeter(UPLOAD_THIN_PROJECT_METER_NAME);
   }
@@ -156,14 +152,5 @@ public class CommonMetrics {
    */
   public void decrementOOMJobWaitCount() {
     this.OOMWaitingJobCount.dec();
-  }
-
-  /**
-   * Add the queue wait time for a flow to the metrics.
-   *
-   * @param time queue wait time for a flow.
-   */
-  public void addQueueWait(final long time) {
-    this.queueWaitMeter.update(time);
   }
 }
--- a/azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
+++ b/azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
@@ -141,14 +141,14 @@ public class CommonMetrics {
   public void markUploadThinProject() { this.uploadThinProjectMeter.mark(); }
 
   /**
-   * Mark the occurrence of an job waiting event due to OOM
+   * Mark the occurrence of a job waiting event due to OOM
    */
   public void incrementOOMJobWaitCount() {
     this.OOMWaitingJobCount.inc();
   }
 
   /**
-   * Unmark the occurrence of an job waiting event due to OOM
+   * Unmark the occurrence of a job waiting event due to OOM
    */
   public void decrementOOMJobWaitCount() {
     this.OOMWaitingJobCount.dec();
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
@@ -18,6 +18,7 @@ package azkaban.execapp;
 
 import azkaban.execapp.metric.ProjectCacheHitRatio;
 import azkaban.metrics.MetricsManager;
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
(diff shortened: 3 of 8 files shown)


az-core/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -60,13 +60,10 @@ private void registerJvmMetrics() {
   }
 
   /**
-   * A {@link Meter} measures the rate of events over time (e.g., “requests per second”). Here we
-   * track 1-minute moving averages.
+   * A {@link Meter} measures the rate of events over time (e.g., “requests per second”).
    */
   public Meter addMeter(final String name) {
-    final Meter curr = this.registry.meter(name);
-    this.registry.register(name + "-gauge", (Gauge<Double>) curr::getOneMinuteRate);
```
ypadron-in, line 68: Removing duplicated info: 1 minute rate values are already emitted by the meter

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1537,6 +1570,23 @@ public synchronized void handleEvent(final Event event) {
                     FlowRunner.this.flow.getExecutionOptions().getSlaOptions()));
       }
     }
+
+    private void reportJobFinishedMetrics(ExecutableNode node) {
```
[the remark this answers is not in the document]
ypadron-in: will add it

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1537,6 +1570,23 @@ public synchronized void handleEvent(final Event event) {
...
+      final Status status = node.getStatus();
+      switch (status) {
+        case SUCCEEDED:
+          FlowRunner.this.execMetrics.markJobSuccess();
```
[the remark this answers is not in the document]
ypadron-in: because `reportJobFinishedMetrics()` is inside an inner class

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -241,6 +245,12 @@ AlerterHolder getAlerterHolder() {
 
   @Override
   public void run() {
+    // Record the time between submission and when the flow execution starts.
+    // Note that since submit time is recorded on the web server and flow execution happens on
+    // the executor, there could be some inaccuracies due to clock skew.
```
li-ygerchikov, line 250: Exactly. I am afraid this issue right here renders this metric not very useful. Not sure if there is a better way.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -282,6 +289,8 @@ public void run() {
         closeLogger();
         updateFlow();
       } finally {
+        reportFlowFinishedMetrics();
```
li-ygerchikov, line 292: Calling this in `finally` means that the flow status may not have been updated properly... is it wise to do it in `finally` *and* still rely on flow status to mark the right metric?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -297,6 +306,26 @@ public void run() {
...
+    switch (status) {
+      case SUCCEEDED:
+        this.execMetrics.markFlowSuccess();
+        break;
+      case FAILED:
+        this.commonMetrics.markFlowFail();
+        break;
+      case KILLED:
+        this.execMetrics.markFlowKilled();
+        final long now = System.currentTimeMillis();
```
li-ygerchikov, line 320: tiny nit: I would embed this into the conditional below because (1) we may not even need it, and (2) in case there is a long pause between execution of this and the next line. (Although optimizer may take care of both... I dunno.)

azkaban-common/src/main/java/azkaban/metrics/CommonMetrics.java
```diff
@@ -157,13 +153,4 @@ public void incrementOOMJobWaitCount() {
   public void decrementOOMJobWaitCount() {
     this.OOMWaitingJobCount.dec();
   }
-
-  /**
-   * Add the queue wait time for a flow to the metrics.
-   *
-   * @param time queue wait time for a flow.
-   */
-  public void addQueueWait(final long time) {
```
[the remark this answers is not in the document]
ypadron-in: `addQueueWait` records the time between submission and when the flow preparation starts, so _queue_ refers to web server queue if old dispatch is enabled or DB queue if new dispatch is enabled. It works regardless of the dispatch model being used.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: I think is safe to place it here because in line 276 execution status is set to FAILED on the occurrence of any exceptions and any further failures won't change that.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
```diff
@@ -29,21 +31,50 @@
...
+  public static final String JOB_FAIL_METER_NAME = "job-fail-meter";
+  public static final String JOB_SUCCESS_METER_NAME = "job-success-meter";
+  public static final String JOB_KILLED_METER_NAME = "job-killed-meter";
 
   private final MetricsManager metricsManager;
   private Timer flowSetupTimer;
   private final ProjectCacheHitRatio projectCacheHitRatio;
+  private Histogram queueWaitHistogram;
+  private Histogram flowTimeToStartHistogram;
+  private Meter flowKillingMeter;
```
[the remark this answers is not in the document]
ypadron-in: You're right, both of them as meters are basically the same thing with a time shift. However, if we have KILLING as a counter it is useful to know how many flow executions are in KILLING status in a point in time and set alerts based on that as executions can get stuck in this status for different reasons.
Kill is a user action, so to me the KILLED metric has more statistical value rather than being an indicator of the health of the system. 
LMKWYT

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: I'm not sure how much clock skew we can get if all machines synchronize their time against a server, therefore I'm not sure how much inaccuracy we can expect from this metric, however it can still be useful in my opinion to for example detect anomalies between releases. While single values could be useless due to clock skew it is still valuable to analyze trends over time, right?
Ideally we should use the clock in the DB as the source of truth for this metric but this means modifying current implementations and DB schema to make the DB record operation's timestamps.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: Will do.

azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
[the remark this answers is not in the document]
li-ygerchikov: Sounds good. Thanks for giving it a thorough consideration.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: Talked offline.
Agreed to removing metrics affected by clock skew issues until these are fixed. Deleted metrics:
-`queue-wait-histogram`: Records the time between submission and the start of flow execution preparation.
-`flow-time-to-start-histogram`: Records the time between submission (a new execution request is inserted in the DB) and the start of flow execution.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -262,12 +266,12 @@ public void run() {
...
         this.logger
-            .error(
-                "An error has occurred during the running of the flow. Quiting.",
-                t);
+            .error("An error has occurred during the running of the flow. Quiting.", t);
+      }
+      if (Status.KILLING.equals(this.flow.getStatus())) {
+        this.execMetrics.decrementFlowKillingCount();
       }
       this.flow.setStatus(Status.FAILED);
```
anish-mal, line 274: I am curious whether this is just you trying to ensure FlowKillingCount stays accurate or have you in the past seen issues with Killing flows.

FlowRunner::kill() sets the status to `KILLING`. If we find the flow in Killing state, this indicates Killing itself failed. This is an interesting case. Since FlowRunner::kill() throws no exception, this can happen only in case of a Runtime Exception or java.lang.Error. Have you in the past seen 'Killing' resulting in these errors?

azkaban-exec-server/src/main/java/azkaban/execapp/ExecMetrics.java
```diff
@@ -29,21 +32,45 @@
...
   private final ProjectCacheHitRatio projectCacheHitRatio;
+  private Counter flowKillingCounter;
+  private Histogram flowTimeToKillHistogram;
+  private Meter flowKilledMeter;
+  private Meter flowSuccessMeter;
+  private Meter jobFailMeter;
+  private Meter jobSuccessMeter;
+  private Meter jobKilledMeter;
+  // TODO ypadron-in: add metrics to measure the time between flow submission and flow execution
+  // preparation/start after clock skew issues in execution times are resolved.
```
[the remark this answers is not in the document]
anish-mal: I am under the impression what we want to lookout for is flow in Preparation state for many minutes, sometimes even hours.  Is that assumption right?

Clock skew should not hamper that I guess. Hence it will be helpful to know what clock skews are we referring here.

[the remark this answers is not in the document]
ypadron-in: Here is the reason for the clock issues: https://github.com/azkaban/azkaban/pull/2481/files#diff-e86078574720dfac4b80b618a6ccafe9L435-L439
I discussed this with @li-ygerchikov and we agreed on removing them until it's fixed. Essentially with that issue metrics using execution times will not be accurate (depending on how servers synchronize their time, the differences could be in the order of minutes) and therefore unreliable.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: I's essentially to keep the killing count dependable. I haven't seen that type of errors before.

PR #2482, opened 2020-04-01
 If two projects have same flow name and ramp manager event is fired,…

jakhani:
… for the second event, it is getting NullPointerException as executingFlows has flowId as a key which is same for both the flows.

When first flow execution event comes, FlowRampManager is deleting the flow from executingFlows Map. Here the key is flow name. When second event comes with the same flow name, ramp manager is throwing NullPointerException for ramp.toString() as ramp is null because during previous flow event processing, all the entries were deleted from executionFlows map. As this is the executing in a thread, it is not blocking main thread or raising any exception log as exception is not caught anywhere. I am changing executingFlows map and making executionId as a key to remove all the conflicts.


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -629,8 +629,8 @@ synchronized private Action convertToAction(Status status) {
 
   private static class RampDataModel {
     // Host the current processing ramp flows
-    // Map.Key = flowName, Map.Value = Set Of Ramps
```
anish-mal, line 632: I think this comment does get impacted by the changes. It will help to modify this as well

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -629,8 +629,8 @@ synchronized private Action convertToAction(Status status) {
 
   private static class RampDataModel {
     // Host the current processing ramp flows
```
anish-mal, line 631:
```suggestion
    // Host the current processing ramp flows
    // Map.key = Any ID that uniquely identifies each execution. (we will use executionId). Map.value = Set of ramps
```

kxu1026:
LGTM.

PR #2483, opened 2020-04-01
Fix missing job callbacks issue.

li-gsalia:
FlowManager event handlers are invoked by one of JobRunner event handler. Fix is not to access the runner object in flow event handler during handling of job events.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/event/EventHandler.java
+++ b/azkaban-common/src/main/java/azkaban/event/EventHandler.java
@@ -17,10 +17,13 @@ package azkaban.event;
 
 import java.util.ArrayList;
 import java.util.HashSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class EventHandler {
 
   private final HashSet<EventListener> listeners = new HashSet<>();
+  private static final Logger logger = LoggerFactory.getLogger(EventHandler.class);
 
   public EventHandler() {
   }
@@ -41,7 +44,12 @@ public class EventHandler {
     final ArrayList<EventListener> listeners =
         new ArrayList<>(this.listeners);
     for (final EventListener listener : listeners) {
-      listener.handleEvent(event);
+      try {
+        listener.handleEvent(event);
+      } catch (RuntimeException e) {
+        logger.warn("Error while calling handleEvent for: " + listener.getClass());
+        logger.warn(e.getMessage(), e);
+      }
     }
   }
 
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackManager.java
@@ -26,7 +26,8 @@ import java.util.Map;
 import java.util.TimeZone;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.message.BasicHeader;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Responsible processing job callback properties on job status change events.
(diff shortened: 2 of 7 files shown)


azkaban-common/src/main/java/azkaban/event/EventHandler.java
```diff
@@ -41,7 +44,12 @@ public void fireEventListeners(final Event event) {
     final ArrayList<EventListener> listeners =
         new ArrayList<>(this.listeners);
     for (final EventListener listener : listeners) {
-      listener.handleEvent(event);
+      try {
+        listener.handleEvent(event);
+      } catch (Exception e) {
+        logger.warn("Error while calling handleEvent for: " + listener.getClass());
+        logger.warn(e.getMessage(), e);
+      }
```
anish-mal, line 52: Wondering what's the benefit of catching java.lang.Exceptions for EventListener.handleEvent() when handleEvent doesn't declare a throw in its signature:

```
public interface EventListener {
  public void handleEvent(Event event);
}
```
I am afraid that this might be dead code, that may never get executed.

abti: It catches perhaps Unchecked exceptions, which would help if there are any Runtime exceptions. 
We dont want other listeners to not get impacted irrespective of failures with one. 
(It does not covers Error, so my question is on the flip: why not catch Throwable?) :)

li-gsalia: @anish-mal  - @abti is correct. This will catch unchecked exceptions. This code path was tested by:
1. Keeping the existing bug and making sure rest of the event handlers were executed.
2. Throwing an exception from a different event handler (JobCallbackManager) and making sure the rest of the handlers were called.

anish-mal: I feel let's catch java.lang.RuntimeException and java.lang.Error explicitly. This makes the code itself very clear of what our intention is.

anish-mal: The current code will not catch  java.lang.Error.

azkaban-common/src/main/java/azkaban/event/EventHandler.java
```diff
@@ -41,7 +44,12 @@ public void fireEventListeners(final Event event) {
     final ArrayList<EventListener> listeners =
         new ArrayList<>(this.listeners);
     for (final EventListener listener : listeners) {
-      listener.handleEvent(event);
+      try {
+        listener.handleEvent(event);
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
anish-mal: If you catch java.lang.RuntimeException and java.lang.Error, please include those here.

azkaban-common/src/main/java/azkaban/event/EventHandler.java
anish-mal: Yes, not sure about catching  java.lang.Error - you are right. So let's leave that aside while we ponder over this. But without loss of what we are trying - java.lang.RuntimeException will clearly indicate our intention. Please consider doing this.

azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackManager.java
```diff
@@ -144,17 +148,13 @@ private void processJobCallOnFinish(final Event event) {
         JobCallbackUtil.buildJobContextInfoMap(event, this.azkabanHostName);
 
     JobCallbackStatusEnum jobCallBackStatusEnum = null;
-    final Logger jobLogger = jobRunner.getLogger();
-
+    final Logger jobLogger = this.logger;
```
djaiswal83, line 151: When we are not using jobLogger, I suppose this is not necessary anymore. Instead use "this.logger" everywhere.

azkaban-exec-server/src/main/java/azkaban/execapp/event/JobCallbackManager.java
```diff
@@ -144,17 +148,13 @@ private void processJobCallOnFinish(final Event event) {
...
     if (jobStatus == Status.SUCCEEDED) {
-
       jobCallBackStatusEnum = JobCallbackStatusEnum.SUCCESS;
-
     } else if (jobStatus == Status.FAILED
         || jobStatus == Status.FAILED_FINISHING || jobStatus == Status.KILLED) {
-
       jobCallBackStatusEnum = JobCallbackStatusEnum.FAILURE;
     } else {
       jobLogger.info("!!!! WE ARE NOT SUPPORTING JOB CALLBACKS FOR STATUS: "
```
djaiswal83, line 160: jobLogger should be replaced by this.logger.

PR #2488, opened 2020-04-07
Adding logs to debug why flows are getting stuck in PREPARING state for some usecases

jakhani:
We have seen couple of usecases where executions are stuck in PREPARING state. To unblock these executions, we had to change executor_id to null in database. I am adding these logs to debug the same issue as there was no exception in executor logs for the failure.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -16,6 +16,10 @@
  */
 package azkaban.execapp;
 
+import static azkaban.utils.ThinArchiveUtils.getDependencyFile;
+import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
+
 import azkaban.execapp.metric.ProjectCacheHitRatio;
 import azkaban.executor.ExecutableFlow;
 import azkaban.executor.ExecutorManagerException;
@@ -42,10 +46,6 @@ import java.util.zip.ZipFile;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static azkaban.utils.ThinArchiveUtils.*;
-import static com.google.common.base.Preconditions.*;
-import static java.util.Objects.*;
 
 
 class FlowPreparer {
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -158,6 +158,11 @@ void setup(final ExecutableFlow flow) throws ExecutorManagerException {
       FileIOUtils.deleteDirectorySilently(tempDir);
       LOGGER.error("Error in preparing flow execution {}", flow.getExecutionId(), ex);
       throw new ExecutorManagerException(ex);
+    } catch (final Throwable t) {
+      FileIOUtils.deleteDirectorySilently(tempDir);
```
djaiswal83, line 162: This looks like a change in behavior. If a flow gets error in prepare stage, it wasn't cleaning up execution directory?

jakhani: Yes that is true.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -158,6 +158,11 @@ void setup(final ExecutableFlow flow) throws ExecutorManagerException {
...
+      LOGGER.error("Error in preparing flow execution {}", flow.getExecutionId(), t);
```
ypadron-in, line 163: Could you check the interface of this and above `error` methods? We might not be getting the stack trace of the exception but rather the `toString()` result of the object.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -158,6 +158,11 @@ void setup(final ExecutableFlow flow) throws ExecutorManagerException {
       FileIOUtils.deleteDirectorySilently(tempDir);
       LOGGER.error("Error in preparing flow execution {}", flow.getExecutionId(), ex);
       throw new ExecutorManagerException(ex);
+    } catch (final Throwable t) {
```
ypadron-in, line 161: If catching Throwable exceptions is not a temporary change please consider merging both catch blocks since they are doing almost the same.

abhisheknath2011: Yes agree with @ypadron-in. In fact, we can have just Throwable catch block if the intention is catch all types of Exception including RuntimeException.

jakhani: Thank you. It was temporary but I have added change to only catch throwable.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
jakhani: Thank you for reviewing. It is giving full cause of the issue.

ypadron-in: Thanks for checking! I got confused while reading some logs locally.

PR #2489, opened 2020-04-08
Enable serial polling for executor using mysql lock

li-pkhandelwal:
Changed polling query to lock single row in mysql by using limit condition in nested query.
Enabled serial polling from execution_flows table using GET_LOCK in mysql.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -125,6 +125,7 @@ public class Constants {
     // Configures Azkaban to use new polling model for dispatching
     public static final String AZKABAN_POLL_MODEL = "azkaban.poll.model";
     public static final String AZKABAN_POLLING_INTERVAL_MS = "azkaban.polling.interval.ms";
+    public static final String AZKABAN_POLLING_LOCK_ENABLED = "azkaban.polling.lock.enabled";
     public static final String AZKABAN_POLLING_CRITERIA_FLOW_THREADS_AVAILABLE =
         "azkaban.polling_criteria.flow_threads_available";
     public static final String AZKABAN_POLLING_CRITERIA_MIN_FREE_MEMORY_GB =
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -32,6 +32,7 @@ import java.util.Collections;
 import java.util.List;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.dbutils.ResultSetHandler;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
@@ -338,13 +339,64 @@ public class ExecutionFlowDao {
     }
   }
 
+  public int selectAndUpdateExecutionWithLocking(final int executorId, final boolean isActive)
+      throws ExecutorManagerException {
+    final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ?, update_time = ? "
+        + "where exec_id = ?";
+    final String selectExecutionForUpdate = isActive ?
+        SelectFromExecutionFlows.SELECT_EXECUTION_FOR_UPDATE_ACTIVE :
+        SelectFromExecutionFlows.SELECT_EXECUTION_FOR_UPDATE_INACTIVE;
+
+    final SQLTransaction<Integer> selectAndUpdateExecution = transOperator -> {
+      int execId = -1;
+      final boolean hasLocked = transOperator.query(PollingLockResultHandler.GET_POLLING_LOCK, new PollingLockResultHandler());
+      logger.info("ExecutionFlow polling lock value: " + hasLocked + " for executorId: " + executorId);
+      if (hasLocked) {
+        try {
+          final List<Integer> execIds = transOperator.query(selectExecutionForUpdate, new SelectFromExecutionFlows(), executorId);
+          if (CollectionUtils.isNotEmpty(execIds)) {
+            execId = execIds.get(0);
+            transOperator.update(UPDATE_EXECUTION, executorId, System.currentTimeMillis(), execId);
+          }
+        } finally {
+          transOperator.query(PollingLockResultHandler.RELEASE_POLLING_LOCK, new PollingLockResultHandler());
+          logger.info("Released polling lock for executorId: " + executorId);
+        }
+      }
+      return execId;
+    };
+
+    try {
+      return this.dbOperator.transaction(selectAndUpdateExecution);
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error selecting and updating execution with executor "
+          + executorId, e);
+    }
+  }
+
+  private static class PollingLockResultHandler implements ResultSetHandler<Boolean> {
+    private static final String POLLING_LOCK_NAME = "execution_flows_polling";
+    private static final int GET_LOCK_TIMEOUT = 5;
+    private static final String GET_POLLING_LOCK = "SELECT GET_LOCK('" + POLLING_LOCK_NAME + "', " + GET_LOCK_TIMEOUT + ")";
+    private static final String RELEASE_POLLING_LOCK = "SELECT RELEASE_LOCK('" + POLLING_LOCK_NAME + "')";
+
+    @Override
+    public Boolean handle(final ResultSet rs) throws SQLException {
+      if (!rs.next()) {
+        return false;
+      }
+      return rs.getBoolean(1);
+    }
+  }
+
   public static class SelectFromExecutionFlows implements
       ResultSetHandler<List<Integer>> {
 
     private static final String SELECT_EXECUTION_FOR_UPDATE_FORMAT =
-        "SELECT exec_id from execution_flows WHERE status = " + Status.PREPARING.getNumVal()
+        "SELECT exec_id from execution_flows WHERE exec_id = (SELECT exec_id from execution_flows"
+            + " WHERE status = " + Status.PREPARING.getNumVal()
             + " and executor_id is NULL and flow_data is NOT NULL and %s"
-            + " ORDER BY flow_priority DESC, update_time ASC, exec_id ASC LIMIT 1 FOR UPDATE";
+            + " ORDER BY flow_priority DESC, update_time ASC, exec_id ASC LIMIT 1) FOR UPDATE";
 
     public static final String SELECT_EXECUTION_FOR_UPDATE_ACTIVE =
         String.format(SELECT_EXECUTION_FOR_UPDATE_FORMAT,
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -42,10 +42,12 @@ public class ExecutionFlowDao {
 
   private static final Logger logger = Logger.getLogger(ExecutionFlowDao.class);
   private final DatabaseOperator dbOperator;
+  private final MysqlNamedLock mysqlNamedLock;
 
   @Inject
-  public ExecutionFlowDao(final DatabaseOperator dbOperator) {
+  public ExecutionFlowDao(final DatabaseOperator dbOperator, final MysqlNamedLock mysqlNamedLock) {
     this.dbOperator = dbOperator;
+    this.mysqlNamedLock = mysqlNamedLock;
   }
 
   public void uploadExecutableFlow(final ExecutableFlow flow)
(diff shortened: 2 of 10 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+    };
+
+    try {
+      return this.dbOperator.transaction(selectAndUpdateExecution);
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error selecting and updating execution with executor "
+          + executorId, e);
+    }
+  }
+
```
abhisheknath2011, line 376: Lets add a comment here regarding the purpose of this class.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+  private static class PollingLockResultHandler implements ResultSetHandler<Boolean> {
```
abhisheknath2011, line 377: Can we make this class generic and move outside as it is trying to get lock and this can be leveraged in other place wherever required.

azkaban-db/src/main/java/azkaban/db/DatabaseTransOperator.java
```diff
@@ -89,6 +89,18 @@ public long getLastInsertId() throws SQLException {
     }
   }
 
+  /**
+   *
```
abhisheknath2011, line 93: Lets add method description.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+    private static final String POLLING_LOCK_NAME = "execution_flows_polling";
+    private static final int GET_LOCK_TIMEOUT = 5;
```
abhisheknath2011, line 379: Can we make the lock timeout config driven?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+    private static final String GET_POLLING_LOCK = "SELECT GET_LOCK('" + POLLING_LOCK_NAME + "', " + GET_LOCK_TIMEOUT + ")";
+    private static final String RELEASE_POLLING_LOCK = "SELECT RELEASE_LOCK('" + POLLING_LOCK_NAME + "')";
+
+    @Override
+    public Boolean handle(final ResultSet rs) throws SQLException {
+      if (!rs.next()) {
```
abhisheknath2011, line 385: Do we need to do null check for the ResultSet?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -1055,8 +1055,14 @@ private void pollExecution() {
         }
       } else if (this.pollingCriteria.shouldPoll()) {
         try {
-          final int execId = FlowRunnerManager.this.executorLoader
-              .selectAndUpdateExecution(this.executorId, FlowRunnerManager.this.active);
+          final int execId;
+          if (FlowRunnerManager.this.azkabanProps.getBoolean(ConfigurationKeys.AZKABAN_POLLING_LOCK_ENABLED, false)) {
```
abhisheknath2011, line 1059: Can we add a comment here as we have conditional lock based and regular execution now?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+          if (CollectionUtils.isNotEmpty(execIds)) {
+            execId = execIds.get(0);
+            transOperator.update(UPDATE_EXECUTION, executorId, System.currentTimeMillis(), execId);
+          }
+        } finally {
+          transOperator.query(PollingLockResultHandler.RELEASE_POLLING_LOCK, new PollingLockResultHandler());
+          logger.info("Released polling lock for executorId: " + executorId);
+        }
+      }
+      return execId;
```
abhisheknath2011, line 366: I think we need to commit before returning the execId as we are doing update before.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
li-pkhandelwal: This is using PreparedStatement.executeQuery(a ResultSet object that contains the data produced by the given query; never null).
https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#executeQuery%28java.lang.String%29

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
li-pkhandelwal: we are passing this lambda expression to this.dbOperator.transaction function. We get the connection, execute lambda expression, commit changes and close connection in the transaction function.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
li-pkhandelwal: Added boolean in case we want to revert. We intend to have only lock based polling after the change is tested on clusters.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+      return rs.getBoolean(1);
+    }
+  }
+
   public static class SelectFromExecutionFlows implements
       ResultSetHandler<List<Integer>> {
 
     private static final String SELECT_EXECUTION_FOR_UPDATE_FORMAT =
-        "SELECT exec_id from execution_flows WHERE status = " + Status.PREPARING.getNumVal()
+        "SELECT exec_id from execution_flows WHERE exec_id = (SELECT exec_id from execution_flows"
```
svnarumugam, line 396: The existing approach seems to be modified here. Is this purposeful?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
svnarumugam, line 379: +1 for a shorter timeout. 5 seconds seems more! It would help if we add some MS or SECONDS in the variable to denote that its seconds!

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
abhisheknath2011: Got it. Thanks for the explanation.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error selecting and updating execution with executor "
+          + executorId, e);
+    }
+  }
+
+  private static class PollingLockResultHandler implements ResultSetHandler<Boolean> {
+    private static final String POLLING_LOCK_NAME = "execution_flows_polling";
+    private static final int GET_LOCK_TIMEOUT = 5;
+    private static final String GET_POLLING_LOCK = "SELECT GET_LOCK('" + POLLING_LOCK_NAME + "', " + GET_LOCK_TIMEOUT + ")";
```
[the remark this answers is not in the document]
li-pkhandelwal: yes, I will update

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
li-pkhandelwal: Locking multiple rows was unintended in the existing implementation, it was the reason for deadlock

azkaban-db/src/main/java/azkaban/db/DatabaseTransOperator.java
```diff
@@ -89,6 +89,18 @@ public long getLastInsertId() throws SQLException {
...
+   * @param querySql
+   * @param resultHandler
+   * @param <T>
+   * @return
+   * @throws SQLException
+   */
+  public <T> T query(final String querySql, final ResultSetHandler<T> resultHandler) throws SQLException {
```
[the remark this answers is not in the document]
li-pkhandelwal: Thanks for pointing it out. I will remove this.

azkaban-common/src/main/java/azkaban/executor/MysqlNamedLock.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.inject.Singleton;
+import org.apache.commons.dbutils.ResultSetHandler;
+
+/**
+ *  Util class for locking using named lock in mysql
+ */
+
+@Singleton
+public class MysqlNamedLock implements ResultSetHandler<Boolean> {
+  private String getLockTemplate = "SELECT GET_LOCK('%s', %s)";
```
jakhani, line 15: Can you please make it final static string? Can you please change name to GET_LOCK_TEMPLATE? Same for release lock.

azkaban-common/src/main/java/azkaban/executor/MysqlNamedLock.java
```diff
@@ -0,0 +1,35 @@
...
+  private String releaseLockTemplate = "SELECT RELEASE_LOCK('%s')";
+
+  public boolean getLock(DatabaseTransOperator transOperator, String lockName, int lockTimeoutInSeconds) throws SQLException {
+    String getLockStatement = String.format(getLockTemplate, lockName, lockTimeoutInSeconds);
```
jakhani, line 19: Better to keep check on lockName and lockTimeoutInSeconds. There fields can not be null or <0 respectively.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +341,53 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
...
+      throws ExecutorManagerException {
+    final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ?, update_time = ? "
+        + "where exec_id = ?";
+    final String selectExecutionForUpdate = isActive ?
+        SelectFromExecutionFlows.SELECT_EXECUTION_FOR_UPDATE_ACTIVE :
+        SelectFromExecutionFlows.SELECT_EXECUTION_FOR_UPDATE_INACTIVE;
+
+    final SQLTransaction<Integer> selectAndUpdateExecution = transOperator -> {
+      final String POLLING_LOCK_NAME = "execution_flows_polling";
+      final int GET_LOCK_TIMEOUT_IN_SECONDS = 5;
```
jakhani, line 354: Should this be in configuration>

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -125,6 +125,7 @@
     // Configures Azkaban to use new polling model for dispatching
     public static final String AZKABAN_POLL_MODEL = "azkaban.poll.model";
     public static final String AZKABAN_POLLING_INTERVAL_MS = "azkaban.polling.interval.ms";
+    public static final String AZKABAN_POLLING_LOCK_ENABLED = "azkaban.polling.lock.enabled";
```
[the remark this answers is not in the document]
li-pkhandelwal: I added this flag in case we want to revert this change, as we would be able to do it by just restarting executors

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -338,13 +339,64 @@ public int selectAndUpdateExecution(final int executorId, final boolean isActive
     }
   }
 
+  public int selectAndUpdateExecutionWithLocking(final int executorId, final boolean isActive)
```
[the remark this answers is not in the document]
li-pkhandelwal: We do not intend to keep both of these function, we will remove the earlier function after running this change on some clusters

azkaban-common/src/main/java/azkaban/executor/MysqlNamedLock.java
li-pkhandelwal: this is a singleton class, do we need to make it static?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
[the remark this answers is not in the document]
li-pkhandelwal: Yes, I have seen this ticket. In my understanding, mysql scans index which is being used to execute the query and locks it(not rows) and then locks other indexes non-atomically. The reason for deadlock was that these two queries were using different indexes for the query(i.e. locking in different order) and "FOR UPDATE" was also scanning row which was being updated concurrently by UPDATE query.
Without serialization, multiple executors running the nested query concurrently can get the same execution_id(however, only one will be able to lock it and thereby other executors not poll anything till the next polling frequency). Serialization adds predictability for polling executions and removes unnecessary db calls

PR #2486, opened 2020-04-05
Add IP Address of the project uploader to project_versions.

djaiswal83:
A new column is added to project_versions table,
uploader_ip_addr varchar(50)

PR #2491, opened 2020-04-10
Refactor WebUtils.getRealClientIpAddr() for better reuse.

djaiswal83:
Use getRealClientIpAddr() to fetch uploader's IP
Cleanup older logic to fetch uploader's IP.
This is a followup to #2486

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
@@ -18,7 +18,9 @@ package azkaban.webapp.servlet;
 
 import azkaban.executor.Status;
 import java.text.NumberFormat;
+import java.util.HashMap;
 import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
 
 public class WebUtils {
 
@@ -74,7 +76,8 @@ public class WebUtils {
   }
 
   /**
-   * Gets the actual client IP address inspecting the X-Forwarded-For HTTP header or using the
+   * Gets the actual client IP address on a best effort basis as user could be sitting
+   * behind a VPN. Get the IP by inspecting the X-Forwarded-For HTTP header or using the
    * provided 'remote IP address' from the low level TCP connection from the client.
    *
    * If multiple IP addresses are provided in the X-Forwarded-For header then the first one (first
@@ -84,6 +87,7 @@ public class WebUtils {
    * @param remoteAddr The client IP address and port from the current request's TCP connection
    * @return The actual client IP address
    */
+  // TODO djaiswal83: Refactor this code and merge into single API
   public static String getRealClientIpAddr(final Map<String, String> httpHeaders,
       final String remoteAddr) {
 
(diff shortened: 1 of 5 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -108,4 +110,27 @@ public static String getRealClientIpAddr(final Map<String, String> httpHeaders,
...
+   * Gets the actual client IP address inspecting the X-Forwarded-For HTTP header or using the
+   * provided 'remote IP address' from the low level TCP connection from the client.
+   *
+   * If multiple IP addresses are provided in the X-Forwarded-For header then the first one (first
+   * hop) is used
+   *
+   * @param req HttpServletRequest
+   * @return The actual client IP address
+   */
+  public static String getRealClientIpAddr(final HttpServletRequest req) {
```
ypadron-in, line 123: Is it possible to do some refactoring and pass the value of the `X-Forwarded-For` header directly to `getRealClientIpAddr(Map<String, String> httpHeaders, String remoteAddr)` instead of passing a map? Passing a map doesn't seem useful to me.

djaiswal83: I just refactored the existing logic and put it here from LoginAbstractAzkabanServlet.java 
This is pre-existing and nothing new.
I did find it weird but kept it as is to make sure nothing breaks.

ypadron-in: I understand you're just reusing existing code but since we're already touching it let's improve it. I don't think you would need to make too many changes.

djaiswal83: This involves changes to the tests in WebUtilsTest.java which is some work to do them properly. Given the priorities I can follow it up.

ypadron-in: If we can't address this now, let's then add a TODO in the code mentioning that `getRealClientIpAddr(HttpServletRequest)` will most likely be removed and it shouldn't be used. Also perhaps that `getRealClientIpAddr(Map<String, String>, String)` needs some refactoring.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -84,6 +86,7 @@ public static String formatStatus(final Status status) {
    * @param remoteAddr The client IP address and port from the current request's TCP connection
    * @return The actual client IP address
    */
+  // TODO : Refactor this code and merge into single API
```
ypadron-in, line 89: Could add your user name to the `TODO` item?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -108,4 +111,27 @@ public static String getRealClientIpAddr(final Map<String, String> httpHeaders,
     return clientIp;
   }
 
+  /**
+   * Gets the actual client IP address inspecting the X-Forwarded-For HTTP header or using the
```
ypadron-in, line 115: nit: Getting the actual IP address is best effort, it's not guaranteed as the client can sit behind a VPN. Could the comment reflect that?

PR #2496, opened 2020-04-15
Fix the NullPointerException thrown when an execution is killed before it has started.

ypadron-in:
This issue can be reproduced when Poll Dispatch model(#2038) is used and `azkaban.polling_criteria.flow_threads_available` isn't enabled (#2195). When there are no more flow threads available executions will be queued on the executor side. If one of them is killed while waiting for a thread a NPE will be seen in the flow logs.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -598,11 +598,11 @@ public class FlowRunner extends EventHandler implements Runnable {
     if (nextNodeStatus == Status.CANCELLED) {
       // if node is root flow
       if (node instanceof ExecutableFlow && node.getParentFlow() == null)  {
-        logger.info(String.format("Flow '%s' was cancelled before execution had started.",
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -596,10 +596,16 @@ private boolean runReadyJob(final ExecutableNode node) throws IOException {
     }
 
     if (nextNodeStatus == Status.CANCELLED) {
-      this.logger.info("Cancelling '" + node.getNestedId()
-          + "' due to prior errors.");
-      node.cancelNode(System.currentTimeMillis());
-      finishExecutableNode(node);
+      // if node is root flow
+      if (node instanceof ExecutableFlow && node.getParentFlow() == null)  {
+        logger.info(String.format("Flow '%s' was cancelled before execution had started.",
```
djaiswal83, line 601: nitpick : this.logger

ypadron-in: will add it.

PR #2499, opened 2020-04-20
Fix NPE when SlaOptions have no actions set.

djaiswal83:
Fix NPE when SlaOptions have no actions set which cause flow serialization to fail.

This failure causes excessive logging as the flow is stuck in Preparing state.
Make sure that in such case the flow is finalized to FAILED state.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -263,7 +263,7 @@ public class ExecutableFlow extends ExecutableFlowBase {
     flowObj.put(SUBMITTIME_PARAM, this.submitTime);
 
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
-    this.executionOptions.getSlaOptions().stream()
+    this.executionOptions.getSlaOptions()
         .forEach((slaOption) -> slaOptions.add(slaOption.toObject()));
 
     flowObj.put(SLAOPTIONS_PARAM, slaOptions);
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -263,8 +263,12 @@ public class ExecutableFlow extends ExecutableFlowBase {
     flowObj.put(SUBMITTIME_PARAM, this.submitTime);
 
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
-    this.executionOptions.getSlaOptions()
-        .forEach((slaOption) -> slaOptions.add(slaOption.toObject()));
+    List<SlaOption> slaOptionList = this.executionOptions.getSlaOptions();
+    if (slaOptionList != null) {
+      for (SlaOption slaOption : this.executionOptions.getSlaOptions()) {
+        slaOptions.add(slaOption.toObject());
+      }
+    }
 
     flowObj.put(SLAOPTIONS_PARAM, slaOptions);
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -265,7 +265,7 @@ public class ExecutableFlow extends ExecutableFlowBase {
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
     List<SlaOption> slaOptionList = this.executionOptions.getSlaOptions();
     if (slaOptionList != null) {
-      for (SlaOption slaOption : this.executionOptions.getSlaOptions()) {
+      for (SlaOption slaOption : slaOptionList) {
         slaOptions.add(slaOption.toObject());
       }
     }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -35,6 +35,7 @@ import javax.inject.Singleton;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.dbutils.ResultSetHandler;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.log4j.Logger;
 
 @Singleton
@@ -270,16 +271,17 @@ public class ExecutionFlowDao {
       updateExecutableFlowStatus(flow);
       throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
           + flow.getExecutionId());
-    } catch (final NullPointerException npe) {
+    } catch (final RuntimeException re) {
       flow.setStatus(Status.FAILED);
       // Likely due to serialization error
-      if ( data == null) {
+      if ( data == null && re instanceof NullPointerException) {
         logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
         + ". Likely due to SLA missing options. Please re-check the flow SLA");
+        logger.warn("NPE stacktrace" + ExceptionUtils.getStackTrace(re));
       }
       updateExecutableFlowStatus(flow);
       throw new ExecutorManagerException("Error encoding the execution flow due to NPE. "
-          + "Execution Id  = " + flow.getExecutionId());
+          + "Execution Id  = " + flow.getExecutionId(), re);
     }
 
     try {
@@ -293,11 +295,12 @@ public class ExecutionFlowDao {
 
   private void updateExecutableFlowStatus(final ExecutableFlow flow)
     throws ExecutorManagerException {
-    final String UPDATE_FLOW_STATUS = "UPDATE execution_flows SET status = ? where exec_id = ?";
+    final String UPDATE_FLOW_STATUS = "UPDATE execution_flows SET status = ?, update_time = ? "
+        + "where exec_id = ?";
 
     try {
       this.dbOperator.update(UPDATE_FLOW_STATUS, flow.getStatus().getNumVal(),
-          flow.getExecutionId());
+          System.currentTimeMillis(), flow.getExecutionId());
     } catch (final SQLException e) {
       throw new ExecutorManagerException("Error updating flow.", e);
     }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -268,20 +268,19 @@ public class ExecutionFlowDao {
       }
     } catch (final IOException e) {
       flow.setStatus(Status.FAILED);
-      updateExecutableFlowStatus(flow);
+      updateExecutableFlowStatusInDB(flow);
       throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
           + flow.getExecutionId());
     } catch (final RuntimeException re) {
       flow.setStatus(Status.FAILED);
       // Likely due to serialization error
       if ( data == null && re instanceof NullPointerException) {
-        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
-        + ". Likely due to SLA missing options. Please re-check the flow SLA");
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId());
         logger.warn("NPE stacktrace" + ExceptionUtils.getStackTrace(re));
       }
-      updateExecutableFlowStatus(flow);
-      throw new ExecutorManagerException("Error encoding the execution flow due to NPE. "
-          + "Execution Id  = " + flow.getExecutionId(), re);
+      updateExecutableFlowStatusInDB(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow due to "
+          + "RuntimeException. Execution Id  = " + flow.getExecutionId(), re);
     }
 
     try {
@@ -293,7 +292,7 @@ public class ExecutionFlowDao {
     }
   }
 
-  private void updateExecutableFlowStatus(final ExecutableFlow flow)
+  private void updateExecutableFlowStatusInDB(final ExecutableFlow flow)
     throws ExecutorManagerException {
     final String UPDATE_FLOW_STATUS = "UPDATE execution_flows SET status = ?, update_time = ? "
         + "where exec_id = ?";
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -255,9 +255,10 @@ public class ExecutionFlowDao {
             + "SET status=?,update_time=?,start_time=?,end_time=?,enc_type=?,flow_data=? "
             + "WHERE exec_id=?";
 
-    final String json = JSONUtils.toJSON(flow.toObject());
     byte[] data = null;
     try {
+      // If this action fails, the execution must be failed.
+      final String json = JSONUtils.toJSON(flow.toObject());
       final byte[] stringData = json.getBytes("UTF-8");
       data = stringData;
       // Todo kunkun-tang: use a common method to transform stringData to data.
@@ -265,13 +266,38 @@ public class ExecutionFlowDao {
         data = GZIPUtils.gzipBytes(stringData);
       }
     } catch (final IOException e) {
-      throw new ExecutorManagerException("Error encoding the execution flow.");
+      flow.setStatus(Status.FAILED);
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
+          + flow.getExecutionId());
+    } catch (final NullPointerException npe) {
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
+        + ". Likely due to SLA missing options. Please re-check the flow SLA");
+      }
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow due to NPE. "
+          + "Execution Id  = " + flow.getExecutionId());
     }
 
     try {
       this.dbOperator.update(UPDATE_EXECUTABLE_FLOW_DATA, flow.getStatus()
           .getNumVal(), flow.getUpdateTime(), flow.getStartTime(), flow
           .getEndTime(), encType.getNumVal(), data, flow.getExecutionId());
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error updating flow.", e);
+    }
+  }
+
+  private void updateExecutableFlowStatus(final ExecutableFlow flow)
+    throws ExecutorManagerException {
+    final String UPDATE_FLOW_STATUS = "UPDATE execution_flows SET status = ? where exec_id = ?";
+
+    try {
+      this.dbOperator.update(UPDATE_FLOW_STATUS, flow.getStatus().getNumVal(),
+          flow.getExecutionId());
     } catch (final SQLException e) {
       throw new ExecutorManagerException("Error updating flow.", e);
     }
--- a/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
@@ -778,11 +778,11 @@ public class ExecutionFlowDaoTest {
     } catch (ExecutorManagerException e) {
        assert e.getMessage().contains("NPE");
     }
-    // Fetch flow again, the status must be FAILED
-    final ExecutableFlow failedFlow =
+    // Fetch flow again, the status must be READY not PREPARING
+    final ExecutableFlow readyFlow =
         this.executionFlowDao.fetchExecutableFlow(fetchFlow.getExecutionId());
 
-    assertThat(failedFlow.getStatus()).isEqualTo(Status.FAILED);
+    assertThat(readyFlow.getStatus()).isEqualTo(Status.READY);
   }
 
   /*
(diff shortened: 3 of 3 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +255,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
             + "SET status=?,update_time=?,start_time=?,end_time=?,enc_type=?,flow_data=? "
             + "WHERE exec_id=?";
 
-    final String json = JSONUtils.toJSON(flow.toObject());
     byte[] data = null;
     try {
+      // If this action fails, the execution must be failed.
+      final String json = JSONUtils.toJSON(flow.toObject());
```
abhisheknath2011, line 261: I believe flow.toObject() method is throwing NPE if SLA option list (i.e. this.executionOptions.getSlaOptions()) is null or one of the value in the list (i.e. this.executionOptions.getSlaOptions()) is null. If that is the case we could fix root cause of NPE in the ExecutableFlow.toObject() method. The corresponding code to populate SLA options can be modified as follows.
   final List<Map<String, Object>> slaOptions =
        Optional.ofNullable(this.executionOptions.getSlaOptions()).orElse(Collections.emptyList()).stream()
        .filter(Objects::nonNull)
        .map(slaOption -> slaOption.toObject())
        .collect(Collectors.toList());

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +255,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
       if (encType == EncodingType.GZIP) {
         data = GZIPUtils.gzipBytes(stringData);
       }
     } catch (final IOException e) {
-      throw new ExecutorManagerException("Error encoding the execution flow.");
+      flow.setStatus(Status.FAILED);
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
+          + flow.getExecutionId());
+    } catch (final NullPointerException npe) {
```
abhisheknath2011, line 273: While using JSONUtils to read Object to string format it will throw RuntimeException and it is better to catch RuntimeException as it is parent of all the RuntimeException including NPE. And we could address the NPE here by adding the code mentioned in the earlier comment.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +255,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
+        + ". Likely due to SLA missing options. Please re-check the flow SLA");
+      }
+      updateExecutableFlowStatus(flow);
```
abhisheknath2011, line 280: This is invoked two times and can be moved to a finally block. Or do we want to update only during failure or exception?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -277,6 +291,18 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
     }
   }
 
+  private void updateExecutableFlowStatus(final ExecutableFlow flow)
+    throws ExecutorManagerException {
+    final String UPDATE_FLOW_STATUS = "UPDATE execution_flows SET status = ? where exec_id = ?";
```
abhisheknath2011, line 296: I believe we need to add update time as well.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -263,7 +263,7 @@ public void setFlowLockErrorMessage(final String flowLockErrorMessage) {
     flowObj.put(SUBMITTIME_PARAM, this.submitTime);
 
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
-    this.executionOptions.getSlaOptions().stream()
+    this.executionOptions.getSlaOptions()
```
abhisheknath2011, line 266: This code can be changed as mentioned in one of the comment.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: Thanks! It was supposed to be done and I completely forgot!
I did not do it earlier to test if the catch logic works properly and was supposed to work on it after that.
Good catch!

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: Will do!

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: We only want to update during failure. The normal case is when we update multiple entries including flow_data

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: Yes!

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
djaiswal83: Will do

djaiswal83:
> Creating an SLA rule with no action (kill execution or notify user via email) causes a chain of events, one of them is preventing execution records from being updated with flow data.
> I believe this PR takes care of the handling of non SQL exceptions when updating execution records with flow data, but it doesn't cover preventing users from creating SLA rules with no action which is the root cause.
> Could you add UI and server validation to the "Set SLA" dialog to require users to pick one action?

This PR is limited to handling any exception be it SLA or anything else while updating flow in DB and recovering from it.
The setting of SLA properly in UI needs to be handled separately.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +255,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+    } catch (final RuntimeException re) {
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
+        + ". Likely due to SLA missing options. Please re-check the flow SLA");
+      }
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow due to NPE. "
+          + "Execution Id  = " + flow.getExecutionId());
```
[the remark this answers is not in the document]
djaiswal83: Technically it is possible. I will add some checks to log this message when it is NPE

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +255,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+      throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
+          + flow.getExecutionId());
+    } catch (final RuntimeException re) {
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
+        + ". Likely due to SLA missing options. Please re-check the flow SLA");
+      }
+      updateExecutableFlowStatus(flow);
```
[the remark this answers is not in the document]
djaiswal83: We cant throw NPE, that is what caused the whole issue. I am logging the stacktrace.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -263,8 +263,12 @@ public void setFlowLockErrorMessage(final String flowLockErrorMessage) {
     flowObj.put(SUBMITTIME_PARAM, this.submitTime);
 
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
-    this.executionOptions.getSlaOptions().stream()
-        .forEach((slaOption) -> slaOptions.add(slaOption.toObject()));
+    List<SlaOption> slaOptionList = this.executionOptions.getSlaOptions();
+    if (slaOptionList != null) {
+      for (SlaOption slaOption : this.executionOptions.getSlaOptions()) {
+        slaOptions.add(slaOption.toObject());
```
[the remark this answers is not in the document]
djaiswal83: No. the element can't be null. The list however, can be that is why we have the check.
Overall, I am adding code to the UI to ensure we always get correct SlaOptions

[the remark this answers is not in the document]
anish-mal: Got it. Thanks.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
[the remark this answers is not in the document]
anish-mal: You might be able to do it by using Chained Exceptions. Not only does it provide better debuggabilty , it also has other advantages and leads to better architected software modules. The current approach on the other hand might introduce some other defects.

Please take a look at the following code in ExecutionController.java, and see how the metrics will break here.       
```
      // The exflow id is set by the loader. So it's unavailable until after
      // this call.
      this.executorLoader.uploadExecutableFlow(exflow);       
      this.commonMetrics.markSubmitFlowSuccess();
      message += "Execution queued successfully with exec id " +exflow.getExecutionId();      
      return message;

```
Please also check out how the call backs are setup, so that we are able to make the expected callbacks. That is one more thing to worry about, which ought to be checked.

Here is some more info, in more generic terms

_Here are more details of how printing out the error, and/or aborting the operation/returning to the parent function can led to other issues. Here is a depiction of what can happen. 
```
class Loader {
    void uploadExecutableFlow() throwExecutorManagerException;
} 

public static void putFlowInQueue(Loader loader, Executableflowflow) {
  flow.SetCallbacksUponFinishing();
   try {
     loader.uploadExecutableFlow(flow);
     successfulFLowUploadMetric.increment(1);
   } catch (throwExecutorManagerExceptionex) {  }
}
``` 

The expectation above is that metrics will be accurate e.g one could be calculating (successfulFLowUploadMetric – startedMetric). One would be expecting proper call backs to be made. 

However, failing the operation uploadExecutableFlow() under the hood, when the caller thinks the operation is succesful can lead to writing of buggy code at the parent level. A better approach will be to chain the exceptions. 

Use
public Exception(String message,Throwable cause) 
https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html#Exception(java.lang.String,%20java.lang.Throwable)._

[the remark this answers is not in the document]
Comment (djaiswal83) [reply azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java:280]: As discussed offline, I will add chained exception which is definitely a better way to propagate errors.

File: azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java [hunk 9]
```diff
@@ -263,8 +263,12 @@ public void setFlowLockErrorMessage(final String flowLockErrorMessage) {
     flowObj.put(SUBMITTIME_PARAM, this.submitTime);
 
     final List<Map<String, Object>> slaOptions = new ArrayList<>();
-    this.executionOptions.getSlaOptions().stream()
-        .forEach((slaOption) -> slaOptions.add(slaOption.toObject()));
+    List<SlaOption> slaOptionList = this.executionOptions.getSlaOptions();
+    if (slaOptionList != null) {
+      for (SlaOption slaOption : this.executionOptions.getSlaOptions()) {
```
li-gsalia, line 268: Nit: Use slaOptionList instead of using this.executionOptions.getSlaOptions() again.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +256,32 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+      flow.setStatus(Status.FAILED);
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
+          + flow.getExecutionId());
+    } catch (final RuntimeException re) {
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null && re instanceof NullPointerException) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId()
+        + ". Likely due to SLA missing options. Please re-check the flow SLA");
```
li-gsalia, line 279: Unless a NPE won't be thrown by most of the other fields being null, I would refrain from adding SLA as missing, since it may send someone who is debugging this in the wrong direction.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +256,32 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+        logger.warn("NPE stacktrace" + ExceptionUtils.getStackTrace(re));
+      }
+      updateExecutableFlowStatus(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow due to NPE. "
```
li-gsalia, line 283: This should be "due to RuntimeException" since the catch block is for all Runtime exceptions. Since the exception is added in the next line, maybe this text is not needed.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -277,6 +293,19 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
     }
   }
 
+  private void updateExecutableFlowStatus(final ExecutableFlow flow)
```
li-gsalia, line 296: Nit: Consider renaming this function to "updateExecutableFlowStatusInDB". Looking at the current name I thought the function will update the status in flow object.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -762,6 +762,29 @@ public void testFlowStatusWithFetchRecentlyFinishedFlows() throws Exception {
     assertThat(finishedFlows2.get(0).getStatus()).isEqualTo(Status.FAILED);
   }
 
+  @Test
```
li-gsalia, line 765: Thanks for adding a test! A short description for the test would be useful.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
djaiswal83: will do.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: Makes sense

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: remnant of earlier draft. Thanks

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
djaiswal83: Will do

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
djaiswal83: I thought the name was self explanatory, however, I will add some comments.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -268,20 +268,19 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
     } catch (final IOException e) {
       flow.setStatus(Status.FAILED);
-      updateExecutableFlowStatus(flow);
+      updateExecutableFlowStatusInDB(flow);
       throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
           + flow.getExecutionId());
     } catch (final RuntimeException re) {
       flow.setStatus(Status.FAILED);
       // Likely due to serialization error
       if ( data == null && re instanceof NullPointerException) {
```
abhisheknath2011, line 277: Are we expecting NullPointerException here? Looks like it is handled in this PR and also in the frontend code.

djaiswal83: We are expecting that if there is an exception we should be able to handle it.
Since NPE is most likely to happen, we want to log the stack trace immediately for better debuggability.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -762,6 +762,33 @@ public void testFlowStatusWithFetchRecentlyFinishedFlows() throws Exception {
...
+   */
+  @Test
+  public void testUpdateExecutableFlowNullSLAOptions() throws Exception {
+    final ExecutableFlow flow = createTestFlow();
+    this.executionFlowDao.uploadExecutableFlow(flow);
+
+    final ExecutableFlow fetchFlow =
+        this.executionFlowDao.fetchExecutableFlow(flow.getExecutionId());
+
+    // set null sla option
```
anish-mal, line 777: This comment gives no more information than what is clearly seen in the code itself. The comment, being unnecessary should be removed.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -762,6 +762,33 @@ public void testFlowStatusWithFetchRecentlyFinishedFlows() throws Exception {
...
+    fetchFlow.getExecutionOptions().setSlaOptions(null);
+    // Try updating flow
```
anish-mal, line 779: This comment gives no more information than what is clearly seen in the code itself. The comment, being unnecessary should be removed.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -762,6 +762,33 @@ public void testFlowStatusWithFetchRecentlyFinishedFlows() throws Exception {
...
+    try {
+      this.executionFlowDao.updateExecutableFlow(fetchFlow);
+    } catch (ExecutorManagerException e) {
+       assert e.getMessage().contains("NPE");
+    }
+    // Fetch flow again, the status must be READY not PREPARING
```
anish-mal, line 785: This comment gives no more information than what is clearly seen in the code itself. 

Out here, rather than saying the status must be ready - it will be helpful to know why the status must be ready.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
```diff
@@ -762,6 +762,33 @@ public void testFlowStatusWithFetchRecentlyFinishedFlows() throws Exception {
     assertThat(finishedFlows2.get(0).getStatus()).isEqualTo(Status.FAILED);
   }
 
+  /**
+   * Test the resiliency of ExecutableFlow when Sla Option is set to NULL.
+   * Make sure that the system overcomes it and the flow proceeds to next valid state.
```
anish-mal, line 767: System overcoming can be replaced by what you intend the behavior to be.

azkaban-common/src/test/java/azkaban/executor/ExecutionFlowDaoTest.java
djaiswal83: I Would prefer to keep it this way. Ditto for the other three comments.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +256,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
+          + flow.getExecutionId());
+    } catch (final RuntimeException re) {
+      flow.setStatus(Status.FAILED);
+      // Likely due to serialization error
+      if ( data == null && re instanceof NullPointerException) {
+        logger.warn("Failed to serialize executable flow for " + flow.getExecutionId());
+        logger.warn("NPE stacktrace" + ExceptionUtils.getStackTrace(re));
+      }
+      updateExecutableFlowStatusInDB(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow due to "
```
anish-mal, line 282: The exception thrown by updateExecutableFlowStatusInDB(flow), line 281, will suppress the Runtime Exception that has been caught in this catch block. The latter can be added as a suppressed exception into the former.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -255,17 +256,31 @@ private void updateExecutableFlow(final ExecutableFlow flow, final EncodingType
...
       data = stringData;
       // Todo kunkun-tang: use a common method to transform stringData to data.
       if (encType == EncodingType.GZIP) {
         data = GZIPUtils.gzipBytes(stringData);
       }
     } catch (final IOException e) {
-      throw new ExecutorManagerException("Error encoding the execution flow.");
+      flow.setStatus(Status.FAILED);
+      updateExecutableFlowStatusInDB(flow);
+      throw new ExecutorManagerException("Error encoding the execution flow. Execution Id  = "
```
anish-mal, line 272: The exception thrown by updateExecutableFlowStatusInDB(flow), line 271, will suppress the IOException that has been caught in this catch block. The latter can be added as a suppressed exception into the former.

PR #2500, opened 2020-04-20
1. Adding More Tests 2. BugFix for SimpleRampPolicy 3. Refine Code

kxu1026:


azkaban-common/src/main/java/azkaban/executor/BaseRefreshableMap.java
```diff
@@ -60,4 +60,9 @@
   public BaseRefreshableMap<K, V> clone() {
     return (BaseRefreshableMap<K, V>) super.clone();
   }
+
+  @Override
+  public int elementCount() {
+    return this.values().stream().mapToInt(elem -> elem.elementCount()).sum();
```
jakhani, line 66: What if there is no element?

azkaban-common/src/main/java/azkaban/executor/ExecutableRampDependency.java
```diff
@@ -78,4 +78,9 @@ public ExecutableRampDependency clone() {
         .setDefaultValue(this.getDefaultValue())
         .setAssociatedJobTypes(clonedAssociatedJobTypes);
   }
+
+  @Override
+  public int elementCount() {
+    return 1;
```
jakhani, line 84: Why are you returning 1?

azkaban-common/src/main/java/azkaban/executor/BaseRefreshableMap.java
kxu1026: it will return zero. The values is a map so, it could not be NPE.

azkaban-common/src/main/java/azkaban/executor/ExecutableRampDependency.java
kxu1026: This is a simple object. The elementCount is always 1.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -297,8 +294,29 @@ public void insertAction(final String tableName, Map<String, Object> actionData)
       );
     }
 
-
     try {
+      if ("ramp".equalsIgnoreCase(tableName)) {
```
kxu1026, line 298: This block will make sure the default value being added when appending record.
Let's say, if the client didn't checkin the startTime field for Ramp Table, the system will auto append the timestamp.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -393,6 +411,28 @@ public void updateAction(final String tableName, Map<String, Object> actionData,
     }
 
     try {
+      if ("ramp".equalsIgnoreCase(tableName)) {
```
kxu1026, line 414: This block will make sure the default value being added when changing record.
Let's say, if the client didn't checkin the lastUpdatedTime field for Ramp Table, the system will auto append the timestamp.

azkaban-common/src/main/java/azkaban/executor/ExecutableRampExceptionalItems.java
```diff
@@ -54,6 +57,22 @@ public ExecutableRampStatus getStatus(final String key) {
     return Optional.ofNullable(get(key)).map(RampRecord::getStatus).orElse(ExecutableRampStatus.UNDETERMINED);
   }
 
+  public boolean exists(final String key) {
+    return get(key) != null;
```
djaiswal83, line 61: contains serves better purpose than get method.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -70,23 +71,19 @@ public ExecutableRampMap handle(final ResultSet resultSet) throws SQLException {
             ExecutableRamp.createInstance(
                 resultSet.getString(1),
```
djaiswal83, line 72: Can you please convert the hardcoded numbers into meaningful enum?

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
djaiswal83: Please add appropriate comment to explain the functionality.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
djaiswal83: Please add a comment in the code as well.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -525,11 +576,7 @@ public void updateExecutedRampFlows(final String ramp, ExecutableRampExceptional
...
     try {
 
-      Object[][] parameters = executableRampExceptionalItems
-          .getItems()
-          .entrySet()
-          .stream()
-          .filter(item -> item.getValue().isCachedOnly())
+      Object[][] parameters = executableRampExceptionalItems.getCachedItems().stream()
           .map(item -> {
             ArrayList<Object> object = new ArrayList<>();
```
djaiswal83, line 581: Should it be List instead of ArrayList on LHS?

azkaban-common/src/test/java/azkaban/executor/ExecutableRampDependencyMapTest.java
```diff
@@ -0,0 +1,132 @@
...
+    Assert.assertFalse(executableMap.isValidJobType(DEPENDENCY_NAME1, JOB_TYPE_MR));
+    Assert.assertTrue(executableMap.isValidJobType(DEPENDENCY_NAME1, JOB_TYPE_PIG));
+  }
+
+  @Test
+  public void testClone() {
+    executableMap.add(DEPENDENCY_NAME1, DEPENDENCY_VALUE11, JOB_TYPE_SPARK);
+    Assert.assertEquals(1, executableMap.size());
+    ExecutableRampDependencyMap clonedExecutableMap = executableMap.clone();
+    Assert.assertEquals(clonedExecutableMap, executableMap);
```
djaiswal83, line 120: Please swap the params, assertEquals takes (expected, actual). executableMap is expected. clonedExecutableMap is actual.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
```diff
@@ -626,23 +629,19 @@ synchronized private boolean isDatabasePushingActionRequired() {
...
   synchronized private Action convertToAction(Status status) {
-    switch (status) {
-      case FAILED:
-        return Action.FAILED;
-      case SUCCEEDED:
-        return Action.SUCCEEDED;
-      default:
-        return Action.IGNORED;
-    }
+    if (Status.FAILED.equals(status)) return Action.FAILED;
```
djaiswal83, line 638: The switch case is more scalable for future. Any reason why moved to if...else construct?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -906,12 +907,18 @@ private void runExecutableNode(final ExecutableNode node) throws IOException {
...
     node.setStatus(Status.QUEUED);
 
+    // Attach Ramp Props if there is any desired properties
     String jobId = node.getId();
-    String jobType = node.getInputProps().getString("type");
-    Props rampProps = this.flow.getRampPropsForJob(jobId, jobType);
-    if (rampProps != null) {
-      this.logger.info("Selected Ramp Props in Flow : [" + rampProps + "]");
-      node.setRampProps(rampProps);
+    String jobType = Optional.ofNullable(node.getInputProps()).map(props -> props.getString("type")).orElse(null);
```
djaiswal83, line 912: if jobId could be null, then this line should be inside the if condition and jobtype's null check should be nested in it.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -15,23 +15,30 @@
...
   public enum Action {
     IGNORED, SUCCEEDED, FAILED
   }
 
-  public static class State implements IRefreshable<State> {
+  public enum CountType {
+    TRAIL, SUCCESS, FAILURE, IGNORED
+  }
+
+  private static class State implements IRefreshable<State> {
```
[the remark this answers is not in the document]
kxu1026: sure

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -43,6 +50,7 @@
     private volatile int numOfIgnored = 0;
     private volatile boolean isPaused = false;
     private volatile int rampStage = 0;
+    private volatile long lastRampDownTime = 0;
```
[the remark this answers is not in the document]
kxu1026: adding more comment. This variable is used to tracking the 'automatically ramp down' mechanism. Let's say, we change the Status to ramp up from Stage 1 to Stage 2. And, during the ramp up period, the system detected more workflow failure. Therefore, the system will be capable to automatically ramp down from Stage 2 to Stage 1. However, to avoid non-stop ramp down from Stage 1 to Stage 0, we need an extra condition to determine if we need to continue ramping down. lastRampDownTime is added on that purpose. When the ramp has been ramped down in 1 day (by comparing the saved timestamp in lastRampDownTime), the system will not be automatically ramped down again.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -51,226 +59,106 @@
...
 
       this.numOfTrail = source.numOfTrail;
       this.numOfSuccess = source.numOfSuccess;
       this.numOfFailure = source.numOfFailure;
-      this.numOfIgnored = source.numOfFailure;
+      this.numOfIgnored = source.numOfIgnored;
 
       this.isPaused = source.isPaused;
-      this.rampStage = (source.rampStage > this.rampStage) ? source.rampStage : this.rampStage;
+      if (source.rampStage > this.rampStage) {
```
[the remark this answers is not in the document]
kxu1026: It is a fixing bug. The existing logic will not block the ramp down process during the ramp. Let's say, we are in Stage 2 ramping. And, we would like to manually ramp down to Stage 1. The rampStage will not be updated due to (source.rampStage <= this.rampStage).

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -51,226 +59,106 @@
...
+        this.lastRampDownTime = 0;
+      }
+      this.rampStage = source.rampStage;
```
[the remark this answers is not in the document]
kxu1026: As I explained in previous comments, it is a bug which I am fixing. 
There is ramp status control feature : 'When there are more workflow failures detected during the ramp on particular stage, we had better to auto ramp down from Stage N to Stage N-1 to limit the disaster'

The original code has a bug. It will block the action to manually ramp down because when  source.rampStage < this.rampStage, the rampStage didn't get updated.
Right now, after introducing the lastRampDownTime, the problem can be addressed. The rampStage can be updated directly. However, to make sure any immediately ramp up can be automatically ramp down when more error detected, lastRampDownTime need to be reset. If you need more detail information, we can discuss that offline.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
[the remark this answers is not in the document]
djaiswal83: I meant, please add comments in the code itself.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -51,226 +59,106 @@
...
+          && (this.cachedNumOfSuccess == 0)
+          && (this.cachedNumOfTrail == 0);
+
       return this;
     }
 
     @Override
     public State clone() {
-      return null;
+      State cloned = new State();
```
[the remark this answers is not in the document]
kxu1026: sure. I can address that.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -51,226 +59,106 @@
...
+    }
+
+    @Override
+    public int elementCount() {
+      return 1;
     }
   }
 
-  public static class Metadata implements IRefreshable<Metadata> {
+  private static class Metadata implements IRefreshable<Metadata> {
```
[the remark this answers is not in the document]
kxu1026: addessing

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -284,7 +172,12 @@ public Metadata refresh(Metadata source) {
 
     @Override
     public Metadata clone() {
-      return null;
+      return new Metadata(this.maxFailureToPause, this.maxFailureToRampDown, this.isPercentageScaleForMaxFailure);
+    }
+
+    @Override
+    public int elementCount() {
```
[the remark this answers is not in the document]
kxu1026: Since it is a single object, it is always return 1.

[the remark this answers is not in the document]
kxu1026: i added comment.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -293,143 +186,244 @@ public Metadata clone() {
...
-
   public String getPolicy() {
     return policy;
   }
 
-  public ExecutableRamp setPolicy(String policy) {
-    this.policy = policy;
-    return this;
+  public boolean isActive() {
+    long diff = this.state.startTime - System.currentTimeMillis();
```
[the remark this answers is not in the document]
kxu1026: addressed.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -293,143 +186,244 @@ public Metadata clone() {
...
+    lock.lock();
+    try {
+      int currentRampStage = this.state.rampStage;
+      this.state.rampStage = currentRampStage + 1;
+      this.state.lastUpdatedTime = System.currentTimeMillis();
+      this.state.lastRampDownTime = 0;
+      if (this.state.rampStage >= maxStage) {
+        this.state.endTime = this.state.lastUpdatedTime;
       }
+      this.state.isSynchronized = false;
```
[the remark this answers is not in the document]
kxu1026: adding that.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -293,143 +186,244 @@ public Metadata clone() {
...
-        LOGGER.info("[RAMP STOP] (rampId = {}, failure = {}, threshold = {}, timestamp = {})",
-            this.getId(), failure, this.metadata.maxFailureToPause, System.currentTimeMillis());
-      }
+  synchronized public void rampDown() {
+    lock.lock();
+    try {
+      int currentStage = this.state.rampStage;
+      this.state.rampStage = currentStage - 1;
+      this.state.lastRampDownTime = System.currentTimeMillis();
+      this.state.isSynchronized = false;
```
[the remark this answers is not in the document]
kxu1026: adding that.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -293,143 +186,244 @@ public Metadata clone() {
...
+    } finally {
+      lock.unlock();
+    }
   }
 
   synchronized public void cacheSaved() {
-    this.getState().markDataSaved();
+    lock.lock();
+    try {
+      int ttlTrail = this.state.numOfTrail + this.state.cachedNumOfTrail;
```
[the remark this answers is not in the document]
kxu1026: We should protect that. The ramp status loading from db is independent to the ramp up/down. It could be possible one running flow trigged the ramp up/down and in the meantime the polling service is trying to load the status. Let's keep the lock to make sure there is no conflict.

azkaban-common/src/main/java/azkaban/executor/ExecutableRampDependencyMap.java
```diff
@@ -90,10 +89,10 @@ synchronized public boolean isValidJobType(@NotNull final String dependency,
       @NotNull final String jobType) {
     // If no specified job type associated, it means the ramp is valid for all job types
     return Optional.ofNullable(this.get(dependency))
-        .map(ExecutableRampDependency::getAssociatedJobTypes)
-        .filter(CollectionUtils::isNotEmpty)
-        .map(set -> set.contains(jobType))
-        .orElse(true);
+        .map(dp -> Optional.ofNullable(dp.getAssociatedJobTypes())
```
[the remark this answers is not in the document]
kxu1026: haha. As long as both works, I am fine with the new style. let's keep this modern style.

azkaban-common/src/main/java/azkaban/executor/ExecutableRampExceptionalItems.java
kxu1026: addressed.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
kxu1026: No. It need to convert to array by toArray(). It will be better to keep ArrayList instead.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRampManager.java
kxu1026: yes. I want to leverage the Status.isStatusSucceeded function. Status is actually a non-ramp object. If the Status.isStatusSucceeded's enum list has been updated later, we definitely would like the ramp feature will carry the change automatically without any other extra change.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -293,143 +186,244 @@ public Metadata clone() {
...
   }
 
   @Override
   public ExecutableRamp clone() {
-    return null;
+    return new ExecutableRamp(this.id, this.policy, this.metadata.clone(), this.state.clone());
+  }
+
+  @Override
+  public int elementCount() {
```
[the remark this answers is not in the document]
kxu1026: because it is a single object.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
kxu1026: The number is mapping with SQL statement well. And, all other DAO in azkaban is using the same pattern. But, sure, if you have concern, let's address the problem in a seperate PR.

azkaban-common/src/test/java/azkaban/executor/ExecutableRampDependencyMapTest.java
kxu1026: good catch

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
kxu1026: No. JobId and jobType should not be Null. I just want to put more safe guard in case any accident which I don't know.

kxu1026: Let me add warning when jobId or JobType is null, which should be impossible.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -70,23 +71,19 @@ public ExecutableRampMap handle(final ResultSet resultSet) throws SQLException {
...
-                    resultSet.getBoolean(13),
-                    resultSet.getInt(14),
-                    resultSet.getBoolean(15)
-                )
+                resultSet.getInt(3),
+                resultSet.getInt(4),
+                resultSet.getBoolean(5),
+                resultSet.getLong(6),
+                resultSet.getLong(7),
+                resultSet.getLong(8),
```
jakhani, line 79: Can you please change reference using index to column name? This is very risky and error prone.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
```diff
@@ -70,23 +71,19 @@ public ExecutableRampMap handle(final ResultSet resultSet) throws SQLException {
...
+                resultSet.getInt(9),
+                resultSet.getInt(10),
+                resultSet.getInt(11),
+                resultSet.getInt(12),
+                resultSet.getBoolean(13),
+                resultSet.getInt(14),
+                resultSet.getBoolean(15)
```
jakhani, line 86: It is very difficult to follow which values are getting set in getInstance method as there are lot of values. Have you considered using builder pattern for this?

azkaban-common/src/test/java/azkaban/ramppolicy/SimpleRampPolicyTest.java
```diff
@@ -0,0 +1,224 @@
...
+  @Test
+  public void Test_NegativePercentRamp() throws Exception {
+    Props sysProps = new Props();
+    Props privateProps = new Props();
+    RampPolicy policy = new SimpleRampPolicy(sysProps, privateProps);
+
+    Project project = new Project(1, PROJECT_ID);
+    Flow flow = new Flow(FLOW_ID);
+    ExecutableFlow executableFlow = new ExecutableFlow(project, flow);
+    ExecutableRamp executableRamp = ExecutableRamp.createInstance(
```
jakhani, line 114: Can you please consider using builder pattern for ExecutableRamp rather than createInstance?

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
kxu1026: will do in a seperate PR.

azkaban-common/src/main/java/azkaban/executor/ExecutionRampDao.java
kxu1026: Deepark had the same concern. I just tried to follow the existing pattern in Azkaban. As we both agreed, we can address this in a seperate PR.

azkaban-common/src/test/java/azkaban/ramppolicy/SimpleRampPolicyTest.java
kxu1026: Agreed to address it in a seperate PR.
https://jira01.corp.linkedin.com:8443/browse/LIHADOOP-53102

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -114,24 +150,10 @@ public State refresh(State source) {
...
-      cloned.isPaused = this.isPaused;
-      cloned.rampStage = this.rampStage;
-      cloned.lastRampDownTime = this.lastRampDownTime;
-      cloned.cachedNumOfTrail = this.cachedNumOfTrail;
-      cloned.cachedNumOfSuccess = this.cachedNumOfSuccess;
-      cloned.cachedNumOfFailure = this.cachedNumOfFailure;
-      cloned.cachedNumOfIgnored = this.cachedNumOfIgnored;
-      cloned.isActive = true;
-      return cloned;
+      return new State(this.rampStage, this.isActive, this.isPaused, this.isSynchronized,
```
djaiswal83, line 153: this works too. What I meant was to use existing constructor to create State object and add the additional values and avoid using empty constructor.

PR #2501, opened 2020-04-21
Currently there is no limit on number of records getting deleted by cleanup script. Adding limit to make sure database is not affected by cleanup script execution.

jakhani:
When cleanup script was executed for the first time, it has deleted so many records that it affected slave server. I am adding limit to make sure that no matter how frequently we run cleanup script, it only deletes limited records and slave server does not get affected.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
@@ -16,9 +16,9 @@
 
 package azkaban.executor;
 
-import azkaban.db.EncodingType;
 import azkaban.db.DatabaseOperator;
 import azkaban.db.DatabaseTransOperator;
+import azkaban.db.EncodingType;
 import azkaban.db.SQLTransaction;
 import azkaban.utils.FileIOUtils;
 import azkaban.utils.FileIOUtils.LogData;
@@ -136,6 +136,19 @@ public class ExecutionLogsDao {
   }
 
   int removeExecutionLogsByTime(final long millis, final int recordCleanupLimit)
+      throws ExecutorManagerException {
+    int totalRecordsRemoved = 0;
+    int removedRecords;
+    do {
+      removedRecords = removeExecutionLogsBatch(millis, recordCleanupLimit);
+      logger.debug("Removed batch of execution logs. Count of records removed in this batch: "
+          + removedRecords);
+      totalRecordsRemoved = totalRecordsRemoved + removedRecords;
+    } while (removedRecords == recordCleanupLimit);
+    return totalRecordsRemoved;
+  }
+
+  int removeExecutionLogsBatch(final long millis, final int recordCleanupLimit)
       throws ExecutorManagerException {
     final String DELETE_BY_TIME =
         "DELETE FROM execution_logs WHERE upload_time < ? LIMIT ?";
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
@@ -135,12 +135,12 @@ public class ExecutionLogsDao {
     }
   }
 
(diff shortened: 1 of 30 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/ExecutionLogsCleaner.java
```diff
@@ -25,10 +25,14 @@
    private static final long DEFAULT_EXECUTION_LOGS_RETENTION_MS = 3 * 4 * 7
        * 24 * 60 * 60 * 1000L;
 
-   // 1 hour
-   private static final long DEFAULT_LOG_CLEANUP_INTERVAL_SECONDS = 60 * 60;
+   // 1 minute
+   private static final long DEFAULT_LOG_CLEANUP_INTERVAL_SECONDS = 60;
```
li-pkhandelwal, line 29: cleaning up every minute for something which has retention of 12 weeks seems too frequent. Do we know how long does it take on avg to execute this statement for 1000 deletions, and do we have over 1000 executions for clusters in a minute. 
This is not cleaning up everything too, we could probably run this clean up every hour with some batch_size in a loop while there are records to delete and committing each batch.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -47,6 +43,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static azkaban.utils.ThinArchiveUtils.*;
```
abhisheknath2011: Looks like the the * imports got introduced by IDE or by formatting. Might want to change these to individual imports.

azkaban-web-server/src/main/java/azkaban/webapp/ExecutionLogsCleaner.java
djaiswal83: 48K log entries seems too less for clusters like UMP where at any given time 500-1500 flows are running with each flow running multiple jobs. However, I agree 1 minute is too aggressive. Lets find out how many logs are added in 24 hrs in UMP and then work on the frequency backwards.

jakhani: I will keep 1 hour as default. With that, we can set this differently based on load on cluster by setting config value for this frequency.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
jakhani: This file change should not have been committed. I have reverted it.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
```diff
@@ -135,12 +135,12 @@ private void uploadLogFile(final DatabaseTransOperator transOperator, final int
     }
   }
 
-  int removeExecutionLogsByTime(final long millis)
+  int removeExecutionLogsByTime(final long millis, final int recordCleanupLimit)
       throws ExecutorManagerException {
     final String DELETE_BY_TIME =
-        "DELETE FROM execution_logs WHERE upload_time < ?";
+        "DELETE FROM execution_logs WHERE upload_time < ? LIMIT ?";
```
li-pkhandelwal, line 141: This is not maintaining retention time for logs as we are deleting only 1000 entries and we might end up with huge number of rows spanning over multiple months over a period of time if the cluster generates more rows as we are restricting ourselves to delete only 24k rows/day. 
Like I pointed out earlier a better would be to probably run this clean up every hour with some batch_size in a loop while there are records to be deleted(i.e older than 12 weeks) and committing each batch.

li-pkhandelwal: I understand this limit is in config, but deleting fixed number of rows is not deterministic for enforcing retention period

PR #2502, opened 2020-04-23
Use updated kerberos principal to fetch NameNode Tokens

djaiswal83:
Separate out logger in its own try catch block. This helps in logging out messages when HadoopJavaJob throws exception.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJobRunnerMain.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJobRunnerMain.java
@@ -91,17 +91,23 @@ public class HadoopJavaJobRunnerMain {
       }
     });
 
+    // Separate try catch for logger
     try {
-      _jobName = System.getenv(ProcessJob.JOB_NAME_ENV);
-      String jobPropsFile = System.getenv(ProcessJob.JOB_PROP_ENV);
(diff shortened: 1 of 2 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJobRunnerMain.java
```diff
@@ -91,17 +91,23 @@ public void run() {
...
-      String privatePropsFile = System.getenv(ProcessJob.JOBTYPE_PRIVATE_PROP_ENV);
-
       _logger = Logger.getRootLogger();
       _logger.removeAllAppenders();
       ConsoleAppender appender = new ConsoleAppender(DEFAULT_LAYOUT);
       appender.activateOptions();
       _logger.addAppender(appender);
       _logger.setLevel(Level.INFO); //Explicitly setting level to INFO
+    } catch (Exception e) {
+      throw e;
```
svnarumugam, line 103: Clarification: Do we wrap this logger init steps in exception to continue executing job even if the log set up fails? Should we not print in az log that we failed to init?

djaiswal83: Earlier everything was in one single try catch block. This separate out the logger in its own try catch. This way it can be used in catch block of the larger try catch.
We cant log anything anyway as we need a logger to begin with :)

PR #2504, opened 2020-04-24
Currently there is no limit on number of records getting deleted by cleanup script. Adding limit to make sure database is not affected by cleanup script execution.

jakhani:
When cleanup script was executed for the first time, it has deleted so many records that it affected slave server. I am adding limit to make sure that no matter how frequently we run cleanup script, it only deletes limited records and slave server does not get affected.
This PR covers all the comments captured in https://github.com/azkaban/azkaban/pull/2501.


azkaban-common/src/main/java/azkaban/executor/IntHandler.java
```diff
@@ -0,0 +1,24 @@
+package azkaban.executor;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import org.apache.commons.dbutils.ResultSetHandler;
+
+public class IntHandler implements ResultSetHandler<Integer> {
+  public static final String NUM_EXECUTIONS =
```
li-pkhandelwal, line 8: The string variables should probably be defined in respective dao class as private static field and not in this generic class.

azkaban-common/src/main/java/azkaban/executor/IntHandler.java
```diff
@@ -0,0 +1,24 @@
...
+      "SELECT COUNT(1) FROM execution_flows";
+  public static final String NUM_FLOW_EXECUTIONS =
+      "SELECT COUNT(1) FROM execution_flows WHERE project_id=? AND flow_id=?";
+  public static final String NUM_JOB_EXECUTIONS =
+      "SELECT COUNT(1) FROM execution_jobs WHERE project_id=? AND job_id=?";
+  public static final String FETCH_EXECUTION_LOGS_COUNT = "SELECT COUNT(1) execution_logs WHERE "
```
li-pkhandelwal, line 14: please remove, not being used

azkaban-web-server/src/main/java/azkaban/webapp/ExecutionLogsCleaner.java
```diff
@@ -29,6 +29,10 @@
    private static final long DEFAULT_LOG_CLEANUP_INTERVAL_SECONDS = 60 * 60;
    private long cleanupIntervalInSeconds;
 
+   // 1000 records
```
li-pkhandelwal, line 32: can be removed, this looks implicit from variable name

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
```diff
@@ -135,12 +135,25 @@ private void uploadLogFile(final DatabaseTransOperator transOperator, final int
...
-  int removeExecutionLogsByTime(final long millis)
+  int removeExecutionLogsByTime(final long millis, final int recordCleanupLimit)
+      throws ExecutorManagerException {
+    int totalRecordsRemoved = 0;
+    int removedRecords;
+    do {
+      removedRecords = removeExecutionLogsBatch(millis, recordCleanupLimit);
+      logger.debug("Removed batch of execution logs. Count of records removed in this batch: "
+          + removedRecords);
+      totalRecordsRemoved = totalRecordsRemoved + removedRecords;
```
svnarumugam, line 146: could we consider adding `sleep of 1` second here as recommended by the DBA?
Reference: http://mysql.rjweb.org/doc.php/deletebig

azkaban-common/src/main/java/azkaban/executor/IntHandler.java
```diff
@@ -0,0 +1,24 @@
+package azkaban.executor;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import org.apache.commons.dbutils.ResultSetHandler;
```
svnarumugam, line 5: Do we need the changes in this file? FETCH_EXECUTION_LOGS_COUNT is not used in this change.

azkaban-common/src/test/java/azkaban/executor/ExecutionLogsDaoTest.java
```diff
@@ -109,4 +109,32 @@ public void testLargeUploadLog() throws ExecutorManagerException {
...
+    assertThat(logsResult2).isNotNull();
+    assertThat(logsResult2.getLength()).isEqualTo(64000);
+
+    final LogData logsResult3 = this.executionLogsDao.fetchLogs(1, "largeFiles", 0, 150000, 250000);
+    assertThat(logsResult3).isNotNull();
+    assertThat(logsResult3.getLength()).isEqualTo(185493);
+
+    final long currentTimeMillis = System.currentTimeMillis();
+    int totalRemovedRecords = executionLogsDao.removeExecutionLogsByTime(currentTimeMillis, 3);
+    assertThat(totalRemovedRecords).isEqualTo(7);
```
svnarumugam, line 138: Clarification: How are we asserting that it should be 7? couldn't get the earlier test lines.

azkaban-common/src/test/java/azkaban/executor/ExecutionLogsDaoTest.java
```diff
@@ -109,4 +109,32 @@ public void testLargeUploadLog() throws ExecutorManagerException {
...
+
+    final LogData logsResult = this.executionLogsDao.fetchLogs(1, "largeFiles", 0, 0, 64000);
+    assertThat(logsResult).isNotNull();
+    assertThat(logsResult.getLength()).isEqualTo(64000);
+
+    final LogData logsResult2 = this.executionLogsDao.fetchLogs(1, "largeFiles", 0, 1000, 64000);
+    assertThat(logsResult2).isNotNull();
+    assertThat(logsResult2.getLength()).isEqualTo(64000);
+
+    final LogData logsResult3 = this.executionLogsDao.fetchLogs(1, "largeFiles", 0, 150000, 250000);
```
li-pkhandelwal, line 132: should testing fetchLogs be part of testing cleanup? There is a testcase for testing log upload by calling fetchLogs functionality

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
```diff
@@ -135,12 +135,25 @@ private void uploadLogFile(final DatabaseTransOperator transOperator, final int
...
   }
 
-  int removeExecutionLogsByTime(final long millis)
+  int removeExecutionLogsByTime(final long millis, final int recordCleanupLimit)
+      throws ExecutorManagerException {
+    int totalRecordsRemoved = 0;
+    int removedRecords;
+    do {
+      removedRecords = removeExecutionLogsBatch(millis, recordCleanupLimit);
+      logger.debug("Removed batch of execution logs. Count of records removed in this batch: "
```
li-pkhandelwal, line 144: logger.info?

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
```diff
@@ -135,12 +135,25 @@ private void uploadLogFile(final DatabaseTransOperator transOperator, final int
...
+          + removedRecords);
+      totalRecordsRemoved = totalRecordsRemoved + removedRecords;
+    } while (removedRecords == recordCleanupLimit);
```
abhisheknath2011, line 147: Let us say we have 100K records to be deleted. Does this mean that in batch of 1000 it will run delete in while loop 100 times? Should we put upper limit as well in one run, because this thread will run every one hour? I feel it would be quite a lot of delete in one run. Also not sure how much time it will take to delete 100K records with delay in between. Just thinking about the performance and load in the DB with so many deletes at time.

azkaban-common/src/main/java/azkaban/executor/IntHandler.java
jakhani: My bad. I am removing this change.

azkaban-common/src/main/java/azkaban/executor/IntHandler.java
jakhani: I am removing this file.. This file change is not needed.

azkaban-common/src/main/java/azkaban/executor/IntHandler.java
jakhani: My bad. I am removing this file change.

azkaban-common/src/test/java/azkaban/executor/ExecutionLogsDaoTest.java
jakhani: Not needed. I will remove it.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
jakhani: If there are 2million records, I don't want to fill logs with this line. That is why I kept it as debug.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
jakhani: Thank you. Adding it.

azkaban-web-server/src/main/java/azkaban/webapp/ExecutionLogsCleaner.java
jakhani: Sounds good. Removing it.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
jakhani: Thank you for reviewing this RB.
For clusters with huge execution_logs, I am planning to set 10K as limit. If we do so and have index on upload_time, it will take hardly 1 sec to delete 10K records. So to delete 100K records, delete query will be executed 10 times with delay of 1 sec. So it will take around 20 secs to delete all 100K records. Mysql can handle this much load so we should be good.

azkaban-common/src/test/java/azkaban/executor/ExecutionLogsDaoTest.java
jakhani: The size of logFile 2 is very small. It makes only one entry in execution_logs table. The size of logFile 1 makes 3 chunks in database. I have updated test case to simplify it.

azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java
abhisheknath2011: Sounds good. Thanks for the clarification Janki.

PR #2506, opened 2020-04-27
Make sure SLA options are set correctly.

djaiswal83:
When a user sets SLA, there are two options,
1. kill action
2. Email action.
User must select one of these two.
If user selects email action, then there must be atleast one email address provided.

Also, make sure SlaOption is not null before serializing it.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
@@ -268,14 +268,13 @@ public class ScheduleServlet extends LoginAbstractAzkabanServlet {
           "Unable to parse duration for a SLA that needs to take actions!", e);
     }
 
-    if (actions.size() > 0) {
-      logger.info("Parsing sla as id:" + id + " type:" + type + " sla:"
-          + rule + " Duration:" + duration + " actions:" + actions);
-      return new SlaOptionBuilder(type, flowName, dur).setJobName(id).setActions(actions)
-          .setEmails(emails).createSlaOption();
+    if (actions.size() == 0) {
+      throw new ScheduleManagerException("Unable to create SLA as there is no action set");
     }
-    return null;
-
+    logger.info("Parsing sla as id:" + id + " type:" + type + " sla:"
+        + rule + " Duration:" + duration + " actions:" + actions);
+    return new SlaOptionBuilder(type, flowName, dur).setJobName(id).setActions(actions)
+        .setEmails(emails).createSlaOption();
   }
 
   private Duration parseDuration(final String duration) {
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
@@ -268,7 +268,7 @@ public class ScheduleServlet extends LoginAbstractAzkabanServlet {
           "Unable to parse duration for a SLA that needs to take actions!", e);
     }
 
-    if (actions.size() == 0) {
+    if (actions.isEmpty()) {
       throw new ScheduleManagerException("Unable to create SLA as there is no action set");
     }
     logger.info("Parsing sla as id:" + id + " type:" + type + " sla:"
--- a/azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
+++ b/azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
@@ -189,6 +189,14 @@ azkaban.ChangeSlaView = Backbone.View.extend({
       var duration = rFlowRule.cells[2].firstChild.value;
       var email = rFlowRule.cells[3].firstChild.checked;
       var kill = rFlowRule.cells[4].firstChild.checked;
+      if (email === false && kill === false) {
+        alert("Unable to create SLA Action. Please set either Kill or Email"
+            + " action");
+      }
+      if (email === true && slaEmails === "") {
+        alert("Unable to create SLA Action. Please provide atleast one email"
+            + " address for Email action")
+      }
       settings[row] = id + "," + rule + "," + duration + "," + email + ","
           + kill;
     }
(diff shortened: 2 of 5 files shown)


ypadron-in:
Also, please add `v=<current-timestamp>` in all the _.vm_ files where _schedule-sla.js_ is used. It should look similar to this: 
`<script type="text/javascript" src="${context}/js/azkaban/view/flow-execute-dialog.js?v=1569610022"></script>`

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
```diff
@@ -189,6 +189,14 @@ azkaban.ChangeSlaView = Backbone.View.extend({
       var duration = rFlowRule.cells[2].firstChild.value;
       var email = rFlowRule.cells[3].firstChild.checked;
       var kill = rFlowRule.cells[4].firstChild.checked;
+      if (email === false && kill === false) {
+        alert("Unable to create SLA Action. Please set either Kill or Email"
```
ypadron-in, line 193: These should be Bootstrap modal dialogs instead of `alert()` functions to match Azkaban's styles but maybe it can be done separately.

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
```diff
@@ -189,6 +189,14 @@ azkaban.ChangeSlaView = Backbone.View.extend({
...
+            + " action");
```
ypadron-in, line 194: typo: _actions_, in plural I believe.

arvindpruthi: I don't think it should be plural here as this code will generate a separate alert per bad entry. Is that right Deepak?

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
```diff
@@ -189,6 +189,14 @@ azkaban.ChangeSlaView = Backbone.View.extend({
       var duration = rFlowRule.cells[2].firstChild.value;
       var email = rFlowRule.cells[3].firstChild.checked;
       var kill = rFlowRule.cells[4].firstChild.checked;
+      if (email === false && kill === false) {
```
ypadron-in, line 192: Please use:
```
if (!email && !kill) {
        alert("....");
        return false;
}
```
`return false;` to stop event propagation, otherwise the code keeps running after clicking "Ok" in the alert  message and the API call is made anyways.

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
```diff
@@ -189,6 +189,14 @@ azkaban.ChangeSlaView = Backbone.View.extend({
...
+        alert("Unable to create SLA Action. Please set either Kill or Email"
+            + " action");
+      }
+      if (email === true && slaEmails === "") {
+        alert("Unable to create SLA Action. Please provide atleast one email"
```
ypadron-in, line 197: typo: _atleast_

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
ypadron-in: Please consider rephrasing as: "Unable to create SLA Action. Please check at least one action."
(because is possible to check both actions)

djaiswal83: Yes. This is running in a for loop for each SLA action.

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
djaiswal83: In my testing, the API call was not made, but I agree. I will test this out. Thanks.

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
djaiswal83: good catch!

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
djaiswal83: Can you provide an example where we use them?

djaiswal83: As discussed offline, this is being deferred.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
```diff
@@ -268,14 +268,13 @@ private SlaOption parseSlaSetting(final String set, String flowName, List<String
           "Unable to parse duration for a SLA that needs to take actions!", e);
     }
 
-    if (actions.size() > 0) {
-      logger.info("Parsing sla as id:" + id + " type:" + type + " sla:"
-          + rule + " Duration:" + duration + " actions:" + actions);
-      return new SlaOptionBuilder(type, flowName, dur).setJobName(id).setActions(actions)
-          .setEmails(emails).createSlaOption();
+    if (actions.size() == 0) {
```
ypadron-in, line 271: nit: consider using `isEmpty()`

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
```diff
@@ -189,6 +189,16 @@ azkaban.ChangeSlaView = Backbone.View.extend({
       var duration = rFlowRule.cells[2].firstChild.value;
       var email = rFlowRule.cells[3].firstChild.checked;
       var kill = rFlowRule.cells[4].firstChild.checked;
+      if (!email && !kill) {
+        alert("Unable to create SLA Action. Please set either Kill or Email"
+            + " action");
+        return false;
+      }
+      if (!email && slaEmails === "") {
```
ypadron-in, line 197: Please use:
```
var slaEmails = $('#slaEmails').val().trim();
```
We might want to _trim_ the value to discard white spaces. 
Also, we're not validating the emails specified but is fine to do it in another time as this hole SLA code should probably be revisited and refactored to use jQuery or Backbone instead of plain Javascript.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java
djaiswal83: Will do

azkaban-web-server/src/web/js/azkaban/view/schedule-sla.js
djaiswal83: For now, I will just trim the white spaces.

PR #2508, opened 2020-04-28
Ensure main thread is not interrupted when running testKillBeforeStart test

ypadron-in:
When the main thread is interrupted subsequent tests fail due to an IO channel being closed by the generated _InterruptedException_ in that test.

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
@@ -1092,11 +1092,12 @@ public class FlowRunnerTest2 extends FlowRunnerTestBase {
       }
     });
 
-    this.runner.run();
+    FlowRunnerTestUtil.startThread(this.runner).join();
     // children jobs shouldn't start
     assertStatus("joba", Status.READY);
     assertStatus("joba1", Status.READY);
     waitForAndAssertFlowStatus(Status.KILLED);
+    this.runner = null;
   }
 
 }


azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
```diff
@@ -1092,11 +1092,12 @@ public void testKillBeforeStart() throws Exception {
       }
     });
 
-    this.runner.run();
+    FlowRunnerTestUtil.startThread(this.runner).join();
```
svnarumugam, line 1095: Nice fix. The earlier this.runner.run() runs as part of the main thread itself and now it's running as a separate thread. For my education, how did we find this specific place causing problem? :)

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
```diff
@@ -1092,11 +1092,12 @@ public void testKillBeforeStart() throws Exception {
...
     // children jobs shouldn't start
     assertStatus("joba", Status.READY);
     assertStatus("joba1", Status.READY);
     waitForAndAssertFlowStatus(Status.KILLED);
+    this.runner = null;
```
svnarumugam, line 1100: Clarification: why are we doing this? On a different note, the base class has runner instance variable that is shared by many tests. what is the reason for having this as an instance variable rather than creating it for every unit test?

ypadron-in: Setting to _null_ to prevent other tests from using this particular instance.
I don't know the reason why _runner_ is an instance variable to be honest but that's something we should look into because as you mentioned tests should run in isolation from each other.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerTest2.java
ypadron-in: I added this test recently and that was the only thing different from the rest of the tests.

PR #2514, opened 2020-04-30
Use String Interning to reuse copy from String Constant Pool for Flow…

li-gsalia:
…Props class.

Problem: Analysis of Azkaban heap dump from 3 internal clusters revealed a high percentage of Heap memory usage (between 17% to 40% of heap usage) is due to duplicate strings.

Tops 2 contributors to duplicate strings account for 99% of total duplicates. The top 2 contributors are:
1.	HashMap keys (the HashMap is in azkaban.flow.Flow.flowProps)
2.	azkaban.flow.Flow.flowProps.parentSource

Both can be traced to Flow.flowProps data fields. Change how Flow.flowProps is initialized and dedup the stringsi by using intern().

### DIFF
--- a/azkaban-common/src/main/java/azkaban/flow/FlowProps.java
+++ b/azkaban-common/src/main/java/azkaban/flow/FlowProps.java
@@ -27,8 +27,20 @@ public class FlowProps {
   private Props props = null;
 
   public FlowProps(final String parentSource, final String propSource) {
-    this.parentSource = parentSource;
-    this.propSource = propSource;
+    /**
+     * Use String interning so that just 1 copy of the string value exists in String Constant Pool
+     * and the value is reused. Azkaban Heap dump analysis indicated a  high percentage of heap
+     * usage is coming from duplicate strings of FlowProps fields.
+     *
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/flow/FlowProps.java
```diff
@@ -27,8 +27,20 @@
   private Props props = null;
 
   public FlowProps(final String parentSource, final String propSource) {
```
anish-mal, line 29: Does this mean we have millions of FlowProps objects in the Heap simultaneously? Is that necessary?

djaiswal83: Yes, once a flow is brought in-memory, the FlowProps objects stay in until either the flow is removed or web server restarts. Changing that will require an overhaul.

anish-mal: This pull request is super helpful and we should get it in.

I feel we should not just stop here. I think there have to be million(s) of unnecessary objects being created (and never deleted) for these duplicate string to turn into long-lived objects in the heap.

PR #2516, opened 2020-05-01
Overhaul Project Cache Cleanup Code to fix inefficiencies

arvindpruthi:
The previous code in ProjectCacheCleaner class had several deficiencies:
1. When a watermark for cache cleaning hits, the code did LRU evictions by deleting old
   project directories in a *blocking context*. This can cause flows to get stuck in
   preparing state.
2. Every time this class is invoked in response to a cache miss; the code does equivalent
   of a "du" command on the project cache directory. On large clusters this directory tends
   to get huge thereby making the flows get stuck for several seconds. In extreme cases
   for minutes or even hours.
3. The calculation of watermark (deletion thresh-hold) itself was flawed. The thresh-hold
   is getting calculated from the partition size; when other things like execution directories
   share the same partition. So in many cases the commonly used thresh-hold value of 70% is
   impossible to hit as > 30% space is consumed by non-cache directories.

The code changes overhaul the ProjectCacheCleaner class to fix all 3 issues mentioned above:
1. Deletion is now done in a *non-blocking* context. Also a *throttle-watermark* is introduced
   to address extreme scenarios such as 95% cache full when it actually does make sense to block
   till cache is fully clean.
2. A full "du" is done only on first cache miss. Subsequent invocations of this class will
   only calculate directory space for newly added projects since the previous invocation.
3. The watermarks are now made dynamic. They feed off from:
   "Current Cache Space + Available capacity in partition". Hence, if say execution directory
   is taking more space than usual; then cache watermarks will re-adjust automatically.

TESTING:
1. Adjusted existing unit test cases for the non-blocking behavior. Also added a new test case
   to test throttle condition.
2. Tested on a real cluster by artifically setting low thresh-holds to make sure the new code
   is working as expected and all possible conditions are getting fired.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -298,11 +298,11 @@ public class Constants {
         + ".max_number_per_ip_per_user";
 
     // allowed max size of shared project dir (percentage of partition size), e.g 0.8
-    public static final String PROJECT_CACHE_SIZE_PERCENTAGE = "azkaban"
-        + ".project_cache_size_percentage_of_disk";
+    public static final String PROJECT_CACHE_SIZE_PERCENTAGE =
+        "azkaban.project_cache_size_percentage_of_disk";
 
-    public static final String PROJECT_CACHE_THROTTLE_PERCENTAGE = "azkaban"
-        + ".project_cache_throttle_percentage";
+    public static final String PROJECT_CACHE_THROTTLE_PERCENTAGE =
+        "azkaban.project_cache_throttle_percentage";
 
     // how many older versions of project files are kept in DB before deleting them
     public static final String PROJECT_VERSION_RETENTION = "project.version.retention";
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -301,6 +301,9 @@ public class Constants {
     public static final String PROJECT_CACHE_SIZE_PERCENTAGE = "azkaban"
         + ".project_cache_size_percentage_of_disk";
 
+    public static final String PROJECT_CACHE_THROTTLE_PERCENTAGE = "azkaban"
+        + ".project_cache_throttle_percentage";
+
     // how many older versions of project files are kept in DB before deleting them
     public static final String PROJECT_VERSION_RETENTION = "project.version.retention";
 
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -228,11 +228,15 @@ public class FlowRunnerManager implements EventListener,
             getClass().getClassLoader());
 
     ProjectCacheCleaner cleaner = null;
+    this.LOGGER.info("Configuring Project Cache");
     try {
       final double projectCacheSizePercentage =
           props.getDouble(ConfigurationKeys.PROJECT_CACHE_SIZE_PERCENTAGE);
+      this.LOGGER.info("Configuring Cache Cleaner with {} % as threshold", projectCacheSizePercentage);
       cleaner = new ProjectCacheCleaner(this.projectDirectory, projectCacheSizePercentage);
+      this.LOGGER.info("ProjectCacheCleaner configured.");
     } catch (final UndefinedPropertyException ex) {
+      ex.printStackTrace();
     }
 
     // Create a flow preparer
@@ -892,6 +896,7 @@ public class FlowRunnerManager implements EventListener,
         LOGGER.error(e.getMessage());
       }
     }
+    flowPreparer.shutdown();
     LOGGER.warn("Shutdown FlowRunnerManager complete.");
   }
 
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -229,14 +229,27 @@ public class FlowRunnerManager implements EventListener,
 
     ProjectCacheCleaner cleaner = null;
     this.LOGGER.info("Configuring Project Cache");
+    double projectCacheSizePercentage = 0.0;
+    double projectCacheThrottlePercentage = 0.0;
     try {
-      final double projectCacheSizePercentage =
+      projectCacheSizePercentage =
           props.getDouble(ConfigurationKeys.PROJECT_CACHE_SIZE_PERCENTAGE);
+      projectCacheThrottlePercentage =
+          props.getDouble(ConfigurationKeys.PROJECT_CACHE_THROTTLE_PERCENTAGE);
       this.LOGGER.info("Configuring Cache Cleaner with {} % as threshold", projectCacheSizePercentage);
-      cleaner = new ProjectCacheCleaner(this.projectDirectory, projectCacheSizePercentage);
+      cleaner = new ProjectCacheCleaner(this.projectDirectory,
+          projectCacheSizePercentage,
+          projectCacheThrottlePercentage);
       this.LOGGER.info("ProjectCacheCleaner configured.");
     } catch (final UndefinedPropertyException ex) {
-      ex.printStackTrace();
+      if (projectCacheSizePercentage == 0.0) {
+        this.LOGGER.info("Property {} not set. Project Cache directory will not be auto-cleaned as it gets full",
+            ConfigurationKeys.PROJECT_CACHE_SIZE_PERCENTAGE);
+      } else {
+        // Exception must have been fired because Throttle percentage is not set. Initialize the cleaner
+        // with the default throttle value
+        cleaner = new ProjectCacheCleaner(this.projectDirectory, projectCacheSizePercentage);
+      }
     }
 
     // Create a flow preparer
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
@@ -248,6 +248,8 @@ public class FlowRunnerManager implements EventListener,
       } else {
         // Exception must have been fired because Throttle percentage is not set. Initialize the cleaner
         // with the default throttle value
+        this.LOGGER.info("Property {} not set. Initializing with default value of Throttle Percentage",
+            ConfigurationKeys.PROJECT_CACHE_THROTTLE_PERCENTAGE);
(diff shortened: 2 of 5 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -72,141 +152,192 @@ public boolean accept(final File dir, final String name) {
...
+   * This method will block until all active cleanup threads finish deleting submitted
+   * cleanup jobs.
+   */
+  public void finishPendingCleanup() {
+    final long start = System.currentTimeMillis();
+    try {
+      try {
+        barrier.lock();
+        while (!projectsUnderDeletion.isEmpty()) {
+          emptyQCond.await(1, TimeUnit.SECONDS);
```
djaiswal83, line 252: Please add a long line every time the code enters the while loop.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -72,141 +152,192 @@ public boolean accept(final File dir, final String name) {
...
-    } else {
-      log.info(
-          "Project cache usage[{} MB] < cache limit[{} MB], no need to delete any project dir",
-          bytesToMB(spaceToUseInBytes),
-          bytesToMB(projectCacheMaxSizeInByte));
+    loadAllProjects();
+    log.info("Loading {} project dirs metadata completed in {} msecs",
+        cachedProjects.size(), System.currentTimeMillis() - start);
+
+    long currentCacheSize = getProjectDirsTotalSizeInBytes();
```
djaiswal83, line 285: nit final, applies to next line too.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -72,141 +152,192 @@ public boolean accept(final File dir, final String name) {
...
+    long projectCacheDirCapacity = currentCacheSize + availablePartitionSize;
+    boolean throttleAfterDeletion = false;
+
+    final long highWatermark = (long) (projectCacheDirCapacity * this.percentageOfDisk);
+    final long throttleWatermark = (long) (projectCacheDirCapacity * throttlePercentage);
```
djaiswal83, line 290: nit : this.throttlePercentage

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -72,141 +152,192 @@ public boolean accept(final File dir, final String name) {
...
     return bytes / (1024 * 1024);
   }
 
   /**
-   * Deleting least recently accessed project dirs when there's no room to accommodate new project
+   *
+   * This method will block until all active cleanup threads finish deleting submitted
+   * cleanup jobs.
+   */
+  public void finishPendingCleanup() {
```
djaiswal83, line 246: Please use annotation "@VisibleForTesting" and make this method private.

arvindpruthi:
Addressed all the code review comments. Also made the logic for fetchMetadata cleaner/simpler than before.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -229,14 +229,27 @@ public FlowRunnerManager(final Props props,
...
+          projectCacheThrottlePercentage);
       this.LOGGER.info("ProjectCacheCleaner configured.");
     } catch (final UndefinedPropertyException ex) {
-      ex.printStackTrace();
+      if (projectCacheSizePercentage == 0.0) {
+        this.LOGGER.info("Property {} not set. Project Cache directory will not be auto-cleaned as it gets full",
+            ConfigurationKeys.PROJECT_CACHE_SIZE_PERCENTAGE);
+      } else {
+        // Exception must have been fired because Throttle percentage is not set. Initialize the cleaner
+        // with the default throttle value
```
djaiswal83, line 250: Please add a log line for this case similar to if condition.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -229,14 +229,27 @@ public FlowRunnerManager(final Props props,
...
-      cleaner = new ProjectCacheCleaner(this.projectDirectory, projectCacheSizePercentage);
+      cleaner = new ProjectCacheCleaner(this.projectDirectory,
+          projectCacheSizePercentage,
+          projectCacheThrottlePercentage);
       this.LOGGER.info("ProjectCacheCleaner configured.");
     } catch (final UndefinedPropertyException ex) {
-      ex.printStackTrace();
+      if (projectCacheSizePercentage == 0.0) {
+        this.LOGGER.info("Property {} not set. Project Cache directory will not be auto-cleaned as it gets full",
+            ConfigurationKeys.PROJECT_CACHE_SIZE_PERCENTAGE);
```
djaiswal83, line 247: To maintain the behavior, please print the stacktrace. ex.printStackTrace()

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -229,14 +229,27 @@ public FlowRunnerManager(final Props props,
 
     ProjectCacheCleaner cleaner = null;
     this.LOGGER.info("Configuring Project Cache");
+    double projectCacheSizePercentage = 0.0;
+    double projectCacheThrottlePercentage = 0.0;
```
djaiswal83, line 233: This could be declared inside try block as a final variable.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -55,25 +59,25 @@
...
-  private HashMap<Path, ProjectDirectoryMetadata> cachedProjects;
+  private final Map<Path, ProjectDirectoryMetadata> cachedProjects = new HashMap<>();
 
   // A record of projects under deletion
-  private ConcurrentHashMap<Path, File> projectsUnderDeletion;
+  private final ConcurrentMap<Path, File> projectsUnderDeletion = new ConcurrentHashMap<>();
 
   // Executor service responsible for cache cleanup
-  ExecutorService deletionService;
+  private ExecutorService deletionService;
```
djaiswal83, line 75: this could be final as well as it is defined in ctor.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -152,9 +159,8 @@ public boolean accept(final File dir, final String name) {
       }
     })) {
```
djaiswal83, line 160: This logic which is pre-existing is flawed, it can cause NPE if there is no match.  This could be handled by,
Objects
        .requireNonNull(this.projectCacheDir.listFiles(new FilenameFilter() {
          String pattern = "[0-9]+\\.[0-9]+";

          @Override
          public boolean accept(final File dir, final String name) {
            return name.matches(this.pattern);
          }
        }))) {

Nit : the indentation needs to be fixed.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -152,9 +159,8 @@ public boolean accept(final File dir, final String name) {
...
       if (project.exists() && project.isDirectory()) {
-        ProjectDirectoryMetadata projectDirectoryMetadata = cachedProjects.get(project.toPath());
         if (!projectsUnderDeletion.containsKey(project.toPath())) {
```
djaiswal83, line 162: Conditions in line 161 and 162 can be combined into one.

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -152,9 +159,8 @@ public boolean accept(final File dir, final String name) {
...
-          projectDirectoryMetadata = fetchMetadataUpdate(project.toPath(), projectDirectoryMetadata);
+          ProjectDirectoryMetadata projectDirectoryMetadata = fetchProjectMetadata(project.toPath());
```
djaiswal83, line 163: Suggestion: Line 170, please add a comment for better readability like,
// end for loop

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -243,16 +249,19 @@ private long bytesToMB(final long bytes) {
    * This method will block until all active cleanup threads finish deleting submitted
    * cleanup jobs.
    */
-  public void finishPendingCleanup() {
+  @VisibleForTesting
+  void finishPendingCleanup() {
```
djaiswal83, line 253: nit : please add private qualifier

azkaban-exec-server/src/main/java/azkaban/execapp/ProjectCacheCleaner.java
```diff
@@ -243,16 +249,19 @@ private long bytesToMB(final long bytes) {
...
     final long start = System.currentTimeMillis();
     try {
```
djaiswal83, line 255: I think this whole logic can be written inside one single try catch finally like this,

try {
        barrier.lock();
        while (!projectsUnderDeletion.isEmpty()) {
          log.info("{} entries left in the cache directory deletion Q. Waiting for the cleanup to finish",
              this.projectsUnderDeletion.size());
          this.emptyQCond.await(10, TimeUnit.SECONDS);
        }
      log.info("Took {} ms to complete ongoing cache cleanup.", (System.currentTimeMillis() - start));
    } catch (InterruptedException e) {
      e.printStackTrace();
    } finally {
      this.barrier.unlock();
    }

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -301,6 +301,9 @@
     public static final String PROJECT_CACHE_SIZE_PERCENTAGE = "azkaban"
         + ".project_cache_size_percentage_of_disk";
 
+    public static final String PROJECT_CACHE_THROTTLE_PERCENTAGE = "azkaban"
+        + ".project_cache_throttle_percentage";
```
svnarumugam, line 305: Please consider keeping the config key in one line as it will help in case someone is looking at how this config is used.
public static final String PROJECT_CACHE_THROTTLE_PERCENTAGE = 
        "azkaban.project_cache_throttle_percentage";

PR #2518, opened 2020-05-02
Fixed the broken link in the schedule panel page

svnarumugam:
The current link we are pointing doesn't exist anymore. Pointed to the right link. Tested if it's taking me to the correct page by running the Azkaban WS locally.


svnarumugam:
Thanks for taking a look at @djaiswal83 . It's true that Azkaban is using 2.2.2. However, my understanding is that the detailed instruction link we show is mainly for the cron expressions. If that is the case then the above link will do.

PR #2519, opened 2020-05-04
Rollback timezone feature in schedules.

ypadron-in:
This feature was added in PR #2407.
We're in the need to roll it back for now because it's causing too much confusion among users. The confusion is mainly due to:
-currently in Azkaban most of the times shown in the UI do not include the timezone.
-preview of 10 next runs in the schedule page is calculated based on the specified schedule timezone but after the schedule is created "Next Execution time" is shown in the server's timezone which doesn't necessarily match the preview.
Azkaban users do access and use the application from different timezones but some analysis is needed to come up with a solution that effectively helps and improves experience for those users.


azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/schedulepanel.vm
```diff
@@ -53,6 +52,8 @@
           <div class="form-group">
 
             <div class="col-sm-12" style="height:55px;">
+              <h4 style="color:Coral; font-style:italic;">All schedules are based on the server
```
djaiswal83, line 55: thanks for fixing the typo :)

ZhangchengHu0923:
I was almost crazy by this feature

PR #2520, opened 2020-05-05
Enhance the logic to scan for MR Jobs to kill to handle multiple logs

djaiswal83:
Enhance the logic to scan for MR Jobs to kill to handle multiple job log files.

The core logic is in findApplicationIdFromLog()
The test is testMultipleLogFilesMultipleApplicationIds()

Lot of code style updates due to Save Actions.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -468,22 +468,22 @@ public class HadoopJobUtils {
     BufferedReader br = null;
     final Set<String> applicationIds = new HashSet<>();
 
-    try {
-      // There can be more than one log file. Go through each one of them.
-      for (final File curLogFile : logFiles) {
-        // Start with sanity checks
-        if (!curLogFile.exists()) {
-          throw new IllegalArgumentException("the logFilePath does not exist: "
-              + curLogFile.getAbsolutePath());
-        }
-        if (!curLogFile.isFile()) {
-          throw new IllegalArgumentException("the logFilePath specified  is not a valid file: "
-              + curLogFile.getAbsolutePath());
-        }
-        if (!curLogFile.canRead()) {
-          throw new IllegalArgumentException(
-              "unable to read the logFilePath specified: " + curLogFile.getAbsolutePath());
-        }
+    // There can be more than one log file. Go through each one of them.
+    for (final File curLogFile : logFiles) {
+      // Start with sanity checks
+      if (!curLogFile.exists()) {
+        throw new IllegalArgumentException("the logFilePath does not exist: "
+            + curLogFile.getAbsolutePath());
+      }
+      if (!curLogFile.isFile()) {
+        throw new IllegalArgumentException("the logFilePath specified  is not a valid file: "
+            + curLogFile.getAbsolutePath());
+      }
+      if (!curLogFile.canRead()) {
+        throw new IllegalArgumentException(
+            "unable to read the logFilePath specified: " + curLogFile.getAbsolutePath());
+      }
+      try {
         br = new BufferedReader(new InputStreamReader(
             new FileInputStream(curLogFile), StandardCharsets.UTF_8));
         String line;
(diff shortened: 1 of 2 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -433,52 +442,65 @@ public Void run() throws Exception {
...
-          "unable to read the logFilePath specified: " + logFilePath);
-    }
+
+    // Get the directory and file name
+    final int lastSlash = logFilePath.lastIndexOf('/');
+    final String dirPath = logFilePath.substring(0, lastSlash);
+    final String logFileName = logFilePath.substring(lastSlash + 1);
+
+    final File[] logFiles =
+        new File(dirPath).listFiles((dir, name) -> name.startsWith(logFileName));
```
li-gsalia, line 460: Can you add a comment with few files names as an example so it becomes clear what this code does?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -433,52 +442,65 @@ public Void run() throws Exception {
...
+              final String appId = m.group(1);
               applicationIds.add(appId);
             }
-          }
+          } // end for loop
         }
       }
-    } catch (IOException e) {
+    } catch (final IOException e) {
       log.error("Error while trying to find applicationId for log", e);
```
li-gsalia, line 497: Change this to a warning instead of error since we just log the exception.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -433,52 +442,65 @@ public Void run() throws Exception {
...
-          new FileInputStream(logFile), StandardCharsets.UTF_8));
-      String line;
-
-      // finds all the application IDs
-      while ((line = br.readLine()) != null) {
-        String[] inputs = line.split("\\s");
-        if (inputs != null) {
-          for (String input : inputs) {
-            Matcher m = APPLICATION_ID_PATTERN.matcher(input);
+      // There can be more than one log file. Go through each one of them.
```
ypadron-in, line 465: In what scenarios are there more than one log file?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -433,52 +442,73 @@ public Void run() throws Exception {
...
+      for (final File curLogFile : logFiles) {
+        // Start with sanity checks
+        if (!curLogFile.exists()) {
```
ypadron-in, line 475: When are rolled over files deleted? Is it possible that one gets deleted right after it's being listed but before we try to read it? In that case shouldn't we try reading the next file available instead of failing fast?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -433,52 +442,73 @@ public Void run() throws Exception {
...
+          throw new IllegalArgumentException("the logFilePath does not exist: "
+              + curLogFile.getAbsolutePath());
+        }
```
ypadron-in, line 478: When are rolled over log files deleted? Is it possible that one gets deleted right after it's being listed but before we try to read it? In that case shouldn't we try reading the next file available instead of failing fast?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
djaiswal83: When the log file goes over 5MB. Jobrunner uses RollingAppender for that purpose.
That is why the MR jobs were leaking.
As can be seen in the test, the file names are named according to how they would be in real use case.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
djaiswal83: The files stay there until the execution directory is deleted, just like the regular log file. Unless someone manually deletes the directory or file, this case will not happen.

Issue #2509, opened 2020-04-28
Support SLA options in execute API


juhoautio:
At the moment SLA options can be only set for schedules, because in the API it's only implemented in `ajaxSetSla` of `ScheduleServlet`.

However, SLA options are stored in `ExecutionOptions` of `ExecutableFlow`, so internally Azkaban execution model would naturally support SLA also for non-scheduled executions.

Would you be willing to review a contribution (pull request) where `ajaxExecuteFlow` of `ExecutorServlet` is extended to also allow passing SLA options as request params?

djaiswal83:
This definitely makes sense. Will run this by the team and update here.

juhoautio:
@djaiswal83 thanks, any update?

djaiswal83:
@juhoautio Due to priorities I am not sure if we can pick this up anytime soon.

juhoautio:
@djaiswal83 just to confirm, do you mean that you don't even have time to review the PR at this point if I make one?

( + maybe some QA work that is not visible to the public )

djaiswal83:
No, I meant that it is low in our priorities to pick this up. If there is a PR, we are happy to look into.

juhoautio:
Thanks, nice! That was my question actually:

> Would you be willing to review a contribution (pull request)

djaiswal83:
Oh yes!

PR #2522, opened 2020-05-06
Refactoring only: extract SLA parsing to new SlaRequestUtils

juhoautio:
This is groundwork for https://github.com/azkaban/azkaban/issues/2509: _Support SLA options in execute API_.

Just moved the code into a util class so that it can be next adjusted to be used also in `ExecutorServlet`.

Issue #1577, opened 2017-12-06
Azkaban 3.37.0 SLA options => Flow SLA Rules email action issue


dinusha00:
I am using Azkaban 3.37.0 (two server mode) SLA options => Flow SLA Rules => Email Action

When setting a "Duration" by selecting only "Email Action" will not receive any email alert when a job/flow takes more that the given time. Anyway if we select both "Email Action" and "Kill Action", then it will kill the job and will send an email correctly. 

So due to this bug, if we don't want to kill the long running process, but receive an email when a job takes longer time, will not work.

This works well with older versions but not with this new version

Below is the 'working' settings screenshot
![email working](https://user-images.githubusercontent.com/7746532/33695238-29b8a7c0-db37-11e7-9d95-40d20efaaeda.png)

Below is the 'not working' settings screenshot
![email not working](https://user-images.githubusercontent.com/7746532/33695253-41a8a222-db37-11e7-8641-a9017468e8e6.png)

Thanks!

chengren311:
hey, not able to get a reproduce either. Did you observe anything outstanding in the log?(e.x email sending failure)

dinusha00:
Hi, I will try with newer version and will post if i was able to find anything from logs. thanks for the message. Closing the ticket, since its not reproducible for others

yangxu1362850636:
hi, i used azkaban sla to alert my job, but it did not works,only killed the job which timeout, but did not send any emails to me, my azkaban's version is 3.30.1 @chengren311

chengren311:
@yangxu1362850636 3.30.1 is pretty old. Please consider upgrading to latest. Also what's your email setting, did you configure your email setting correctly in your azkaban.properties file? Can the email address receive flow failure/success email? Did you observe anything outstanding in the log when job killing by SLA happens?

yangxu1362850636:
@chengren311  thanksfor your reply, I'm sure my configuration of azkaban emails is correct,and , i can receive failure/success email from my azkaban, but , i can't receive any sla emails from sla,and The sla timeout  kill has no effect on the spark task submitted to yarn. I am very confused

chengren311:
@yangxu1362850636 can you try latest version?

yangxu1362850636:
i did not, this is about versions???

PR #2523, opened 2020-05-06
Support sla settings in execute & schedule API actions

juhoautio:
This implements https://github.com/azkaban/azkaban/issues/2509 _Support SLA options in execute API_.

As a bonus SLA options are now supported also in the schedule call because it's using the same HttpRequestUtils. This is nice because if using the API directly to add schedules (like many users do), the API now allows ensuring that a scheduled execution doesn't start running before SLA has been set for it.

The separate set sla API for schedules remains as well.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
@@ -19,6 +19,7 @@ import azkaban.executor.DisabledJob;
 import azkaban.executor.ExecutionOptions;
 import azkaban.executor.ExecutionOptions.FailureAction;
 import azkaban.executor.ExecutorManagerException;
+import azkaban.sla.SlaOption;
 import azkaban.user.Permission;
 import azkaban.user.Permission.Type;
 import azkaban.user.Role;
@@ -38,7 +39,11 @@ import org.apache.commons.lang.StringUtils;
 
 public class HttpRequestUtils {
 
-  public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
+  public static final String PARAM_SLA_EMAILS = "slaEmails";
+  public static final String PARAM_SLA_SETTINGS = "slaSettings";
+
+  public static ExecutionOptions parseFlowOptions(final HttpServletRequest req,
+      final String flowName)
       throws ServletException {
     final ExecutionOptions execOptions = new ExecutionOptions();
 
@@ -99,6 +104,15 @@ public class HttpRequestUtils {
     if (hasParam(req, "mailCreator")) {
       final String mailCreator = getParam(req, "mailCreator");
       execOptions.setMailCreator(mailCreator);
+    }
+
+    final Map<String, String> slaSettings = getParamGroup(req, PARAM_SLA_SETTINGS);
+    // emails param is optional
+    final String emailStr = getParam(req, PARAM_SLA_EMAILS, null);
+    final List<SlaOption> slaOptions = SlaRequestUtils.parseSlaOptions(flowName, emailStr,
+        slaSettings);
+    if (!slaOptions.isEmpty()) {
+      execOptions.setSlaOptions(slaOptions);
     }
 
     final Map<String, String> flowParamGroup = getParamGroup(req, "flowOverride");
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -946,7 +946,14 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
     final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
     exflow.setSubmitUser(user.getUserId());
 
-    final ExecutionOptions options = HttpRequestUtils.parseFlowOptions(req, flowId);
+    final ExecutionOptions options;
+    try {
+      options = HttpRequestUtils.parseFlowOptions(req, flowId);
+    } catch (final ServletException e) {
+      logger.info("parseFlowOptions failed", e);
+      ret.put("error", "Error parsing flow options: " + e.getMessage());
+      return;
+    }
     exflow.setExecutionOptions(options);
     if (!options.isFailureEmailsOverridden()) {
       options.setFailureEmails(flow.getFailureEmails());
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -946,7 +946,7 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
     final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
     exflow.setSubmitUser(user.getUserId());
 
-    final ExecutionOptions options = HttpRequestUtils.parseFlowOptions(req);
+    final ExecutionOptions options = HttpRequestUtils.parseFlowOptions(req, flowId);
     exflow.setExecutionOptions(options);
     if (!options.isFailureEmailsOverridden()) {
       options.setFailureEmails(flow.getFailureEmails());
--- a/azkaban-web-server/src/test/java/azkaban/webapp/servlet/ExecutorServletTest.java
+++ b/azkaban-web-server/src/test/java/azkaban/webapp/servlet/ExecutorServletTest.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.webapp.servlet;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import azkaban.sla.SlaAction;
+import azkaban.sla.SlaOption;
+import azkaban.sla.SlaType;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import javax.servlet.ServletException;
+import org.codehaus.jackson.JsonNode;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ExecutorServletTest extends LoginAbstractAzkabanServletTestBase {
+
+  private ExecutorServlet executorServlet;
+
+  @Override
+  @Before
+  public void setUp() throws Exception {
+    super.setUp();
+    this.executorServlet = new ExecutorServlet();
+    this.executorServlet.init(this.servletConfig);
+  }
+
+  /**
+   * When executeFlow is called with SLA options, a new execution should be submitted with SLA.
+   */
+  @Test
+  public void postAjaxExecuteFlowWithSlaSettings() throws Exception {
+    mockTestProjectAndFlow();
+    mockSubmitExecution();
+    this.req.addParameter("ajax", "executeFlow");
+    this.req.addParameter("project", "testProject");
+    this.req.addParameter("flow", "testFlow");
+    this.req.addParameter("slaSettings[1]", ",FINISH,2:30,true,false");
+    this.req.addParameter("slaEmails", "sla1@example.com,sla2@example.com");
+
+    this.executorServlet.handlePost(this.req, this.res, this.session);
+
+    assertEquals(200, this.res.getStatus());
+
+    final JsonNode json = this.res.getResponseJson();
+    assertEquals("testProject", json.path("project").asText());
+    assertEquals("testFlow", json.path("flow").asText());
+    assertEquals("Submitted (mocked)", json.path("message").asText());
+    assertEquals(99, json.path("execid").asInt());
+
+    final List<SlaOption> slaOptions = this.exFlow.getValue().getExecutionOptions().getSlaOptions();
+    final List<SlaOption> expected = Arrays.asList(new SlaOption(SlaType.FLOW_FINISH, "testFlow",
+        "", Duration.ofMinutes(150), ImmutableSet.of(SlaAction.ALERT),
+        ImmutableList.of("sla1@example.com", "sla2@example.com")));
+    Assert.assertEquals(expected, slaOptions);
+  }
+
+  /**
+   * When SLA options can't be parsed, API is expected to return 200 OK with JSON that has "error".
+   */
+  @Test
+  public void postAjaxExecuteFlowWithInvalidSlaSettings() throws Exception {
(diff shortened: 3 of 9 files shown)


djaiswal83:
@juhoautio can we wait couple of weeks?

juhoautio:
@djaiswal83 it's of course up to. Using this internally already, so it doesn't matter too much now how long this takes to be merged. Of course looking forward to reducing the diff between internal & public repo as soon as possible.

If I may ask though, would you be comfortable with at least merging the refactoring-only PR https://github.com/azkaban/azkaban/pull/2522? If new commits appear and cause conflicts it may become a bit tedious for me.

On the other hand even the actual change in this PR is only adding a new feature and in that sense is quite safe.

juhoautio:
@djaiswal83 some weeks passed.. how about now?

djaiswal83:
I will take a look.

djaiswal83:
@juhoautio this code seems to have conflict. Can you please resolve it prior to review process?

juhoautio:
@djaiswal83 I've rebased this. The new PR #2562 can be merged first if you like to have unrelated save actions formatting separately.

djaiswal83:
Thanks @juhoautio . I will review this asap.

ypadron-in:
How are we handling the scenario where a user creates a schedule with SLA settings specified and then calls the _setSLA_ api? Which SLA rules are applied ultimately? How are the conflicts resolved?

ypadron-in:
Sorry, I accidentally clicked the close PR button.

juhoautio:
> How are we handling the scenario where a user creates a schedule with SLA settings specified and then calls the setSLA api? Which SLA rules are applied ultimately? How are the conflicts resolved?

Well, any call to setSLA overwrites the existing SLA settings. But then It's not like one could only call setSLA once. Or am I missing something?

So, if SLA is set as part of the schedule request, the result is that same as if scheduling first without any SLA and then calling setSLA. Following calls to setSLA will replace any previous SLA settings of the given schedule.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -97,12 +101,20 @@ public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
...
-      mailCreator = getParam(req, "mailCreator");
+      final String mailCreator = getParam(req, "mailCreator");
       execOptions.setMailCreator(mailCreator);
     }
 
+    final Map<String, String> slaSettings = getParamGroup(req, PARAM_SLA_SETTINGS);
+    // emails param is optional
+    final String emailStr = getParam(req, PARAM_SLA_EMAILS, null);
+    final List<SlaOption> slaOptions = SlaRequestUtils.parseSlaOptions(flowName, emailStr,
+        slaSettings);
```
ypadron-in, line 113: _parseSlaOptions_ throws an exception if a SLA rule is specified without any actions (for example: send an email, kill execution). How are those validation errors notified to the user?

Also, I noticed _SlaRequestUtils.parseSlaSetting_ throws a `ScheduleManagerException`. Will `ServletException` (in the absence of validation exceptions) be more appropriate for this utility method?

ypadron-in:
`flowInfo` api (`/executor?ajax=flowInfo&execid=12345`) returns execution options of an execution. Currently SLA details are not included in the response and I think they should, specially now that a manual execution could have SLA. Adding SLA info to this api will be very useful for debugging also since there is no UI support for this feature yet.

ypadron-in:
What happens if I create an execution specifying SLA rules at job level and I accidentally disabled one of the jobs used in the SLA rules?

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
juhoautio: > How are those validation errors notified to the user?

Same as usual:

https://github.com/azkaban/azkaban/blob/2a3f861d33f5df9a251086837a6eeb7e80f75614/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ScheduleServlet.java#L200-L201

> I noticed SlaRequestUtils.parseSlaSetting throws a ScheduleManagerException

Not true. `SlaRequestUtils.parseSlaSetting` throws `ServletException`:

https://github.com/azkaban/azkaban/blob/2a3f861d33f5df9a251086837a6eeb7e80f75614/azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java#L22-L23

juhoautio:
> flowInfo api (/executor?ajax=flowInfo&execid=12345) returns execution options of an execution. Currently SLA details are not included in the response and I think they should, specially now that a manual execution could have SLA. Adding SLA info to this api will be very useful for debugging also since there is no UI support for this feature yet.

I agree, but that could be added as a separate PR.

> What happens if I create an execution specifying SLA rules at job level and I accidentally disabled one of the jobs used in the SLA rules?

I'm not sure, but my guess is that it's allowed and SLA trigger is just skipped if a job is disabled. Indeed I think it's natural that the behavior remains the same when using SLA with the new APIs that are added in this PR, and thus there's no reason to add any new validation that would prevent having job-specific SLA for a job that is disabled in the schedule params.

**Note that even if you set SLA on a job and don't ever run that job as part of a _scheduled_ execution, you could definitely trigger a new execution of the same flow via the API where the job with SLA _is_ enabled. In that case the job-specific SLA would apply.** I think the flexibility to allow this special use case is a plus. 

The SLA rules of Azkaban only apply to actual executions of flows & jobs, and can only monitor for the duration to not be more than a specified limit. SLA rules can **not** be used to monitor for the absence of executing a flow or a job, for example.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
ypadron-in: Thanks for generalizing the exception thrown in the utility class. 

SLA validation errors are handled correctly when creating schedules as you pointed out but not when creating a flow execution. `ExecutorServlet::ajaxExecuteFlow()` replies with a 500 response if `HttpRequestUtils.parseFlowOptions(req)` throws an exception. Could you make the necessary changes to handle validation errors there as well?

ypadron-in:
> > flowInfo api (/executor?ajax=flowInfo&execid=12345) returns execution options of an execution. Currently SLA details are not included in the response and I think they should, specially now that a manual execution could have SLA. Adding SLA info to this api will be very useful for debugging also since there is no UI support for this feature yet.
> 
> I agree, but that could be added as a separate PR.

Absolutely. 

> > What happens if I create an execution specifying SLA rules at job level and I accidentally disabled one of the jobs used in the SLA rules?
> 
> I'm not sure, but my guess is that it's allowed and SLA trigger is just skipped if a job is disabled. Indeed I think it's natural that the behavior remains the same when using SLA with the new APIs that are added in this PR, and thus there's no reason to add any new validation that would prevent having job-specific SLA for a job that is disabled in the schedule params.
> 
> ~Note that even if you set SLA on a job and don't ever run that job as part of a _scheduled_ execution, you could definitely trigger a new execution of the same flow via the API where the job with SLA _is_ enabled. In that case the job-specific SLA would apply. I think the flexibility to allow this special use case is a plus.~ Edit: So actually manually triggered executions haven't inherited the SLA options from the existing schedule. Maybe that's the most natural behavior. So no reason to change it. But from now on users will be able to for example submit the same SLA options as with the schedule when launching executions manually (directly with the API).
> 
> The SLA rules of Azkaban only apply to actual executions of flows & jobs, and can only monitor for the duration to not be more than a specified limit. SLA rules can **not** be used to monitor for the absence of executing a flow or a job, for example.

Thanks, I think you're right, no additional validation is required as SLA rules at job level are only set up if that job runs.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -97,12 +101,20 @@ public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
...
     if (hasParam(req, "mailCreator")) {
-      mailCreator = getParam(req, "mailCreator");
+      final String mailCreator = getParam(req, "mailCreator");
       execOptions.setMailCreator(mailCreator);
     }
 
+    final Map<String, String> slaSettings = getParamGroup(req, PARAM_SLA_SETTINGS);
+    // emails param is optional
+    final String emailStr = getParam(req, PARAM_SLA_EMAILS, null);
+    final List<SlaOption> slaOptions = SlaRequestUtils.parseSlaOptions(flowName, emailStr,
```
ypadron-in, line 112: Could you please add license details to the new class `SlaRequestUtils`?

juhoautio: Will do, sorry. I thought I had that configured in my IDE but it must've been lost at some point.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
juhoautio: Nicely spotted. I'll add it.

Btw, did you have some set of commands for manually testing `ExecutorServlet::ajaxExecuteFlow()` like that?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012 LinkedIn Corp.
+ * Copyright 2020 LinkedIn Corp.
```
juhoautio, line 2: @ypadron-in after fixing my IDE settings, this was automatically updated. Is this the right thing to do, and shall I do this for all files affected by this PR? I would suggest merging https://github.com/azkaban/azkaban/pull/2562 before this PR btw.

ypadron-in: I'm not an expert in the copyright topic. What I can say is that we haven't been updating the year when making a change in a file only when creating new ones. Maybe we should run this by a company lawyer to avoid surprises. 

I'll take a look at PR #2562

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
ypadron-in: sorry, I don't have any fancy commands :)

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
juhoautio: All right. Note that in the [wiki](https://github.com/azkaban/azkaban/wiki/Developer-Tools-and-Tips#use-intellij-to-create-and-update-copyright-automatically) it's instructed to have IDE settings update copyright on commit.

I can revert this change to not block this PR by the legal check?

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
juhoautio: Ookay, so did you deduce that there would be a 500 response simply by looking at the code?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
ypadron-in: Thank you. I'll double check the copyright matter asap and update the wiki if necessary.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
ypadron-in: yes, while working on apis changes, I've seen this is what happens when exceptions just bubble up.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ExecutorServletTest.java
```diff
@@ -0,0 +1,125 @@
...
+   * When executeFlow is called without project param, API is expected to raise an exception.
+   */
+  @Test
+  public void postAjaxExecuteFlow() throws Exception {
+    this.req.addParameter("ajax", "executeFlow");
+    try {
+      this.executorServlet.handlePost(this.req, this.res, this.session);
+      fail("Expected exception was not thrown");
+    } catch (final ServletException e) {
+      assertEquals("Missing required parameter 'project'.", e.getMessage());
```
juhoautio, line 107: @ypadron-in I added this unit test for ExecutorServlet. Something I noticed is that `executeFlow` raises an exception in some cases. Would the right thing to do even in this case to return 200 OK with an "error" JSON? Any way, I'm _not_ proposing to change any of that existing behavior as part of this PR.

juhoautio:
@ypadron-in could you check the latest changes, please?

juhoautio:
Rebased to clean up the diff: to discard the commit that was part of PR https://github.com/azkaban/azkaban/pull/2562, which was already merged.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTestBase.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Mock
+  protected ServletConfig servletConfig;
+  @Mock
+  protected Session session;
+  protected AzkabanMockHttpServletRequest req;
+  protected AzkabanMockHttpServletResponse res;
+
+  @Captor
+  protected ArgumentCaptor<ExecutableFlow> exFlow;
```
ypadron-in, line 82: What's this used for?

build.gradle
```diff
@@ -117,6 +117,7 @@ ext.deps = [
     slf4j                : 'org.slf4j:slf4j-api:' + versions.slf4j,
     slf4jLog4j           : 'org.slf4j:slf4j-log4j12:' + versions.slf4j,
     sparkCore            : 'org.apache.spark:spark-core_2.10:1.4.0',
+    springTest           : 'org.springframework:spring-test:3.2.18.RELEASE',
```
ypadron-in, line 120: Why do we need to add this dependency?

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ExecutorServletTest.java
ypadron-in: Thanks for adding the tests! 
You're right, 200 status code shouldn't be returned when there is an error but setting the right status code means adding a breaking change for the API and unfortunately there isn't a versioning strategy yet.
Returning an exception instead of a json response is also incorrect :(.

build.gradle
juhoautio: For mock implementations of servlet request & response. Those are big interfaces and tedious to mock with mockito.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/LoginAbstractAzkabanServletTestBase.java
ypadron-in: never mind I just found it.

juhoautio: To catch value of such arg in mocked methods. The value is read in the test postAjaxExecuteFlowWithSlaSettings.

juhoautio:
Thanks! Are you expecting something else before merging?

xuelianger:
This version  can fix [Azkaban 3.37.0 SLA options => Flow SLA Rules email action issue #1577](https://github.com/azkaban/azkaban/issues/1577)

xuelianger:
This version can fix [Azkaban 3.37.0 SLA options => Flow SLA Rules email action issue #1577](https://github.com/azkaban/azkaban/issues/1577) ??

juhoautio:
@xuelianger not that I'm aware at least.

I have not encountered https://github.com/azkaban/azkaban/issues/1577 with any recent versions – not even those before this PR. I would have to assume that such bug doesn't exist, at least not any more.

PR #2528, opened 2020-05-13
Fix exception handling of executor health checker

sshardool:
Currently in case of any uncaught exceptions during executor health check
the health checking stops and the exception is never reported.
This is fixed by introducing a wrapper routine for catching and reporting any
uncaught exceptions.
Includes additional exception handling when finalizing flows for an
unreachable executor.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
@@ -18,18 +18,19 @@ package azkaban.executor;
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import com.google.common.annotations.VisibleForTesting;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -71,11 +72,13 @@ public class ExecutorHealthChecker {
 
   public void start() {
     logger.info("Starting executor health checker.");
-    this.scheduler.scheduleAtFixedRate(() -> checkExecutorHealth(), 0L, this.healthCheckIntervalMin,
+    this.scheduler.scheduleAtFixedRate(this::checkExecutorHealthQuietly, 0L,
+        this.healthCheckIntervalMin,
         TimeUnit.MINUTES);
   }
 
   public void shutdown() {
+    logger.info("Terminating executor health checker.");
     this.scheduler.shutdown();
     try {
       if (!this.scheduler.awaitTermination(60, TimeUnit.SECONDS)) {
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -71,11 +72,13 @@ public ExecutorHealthChecker(final Props azkProps, final ExecutorLoader executor
...
   public void start() {
     logger.info("Starting executor health checker.");
-    this.scheduler.scheduleAtFixedRate(() -> checkExecutorHealth(), 0L, this.healthCheckIntervalMin,
+    this.scheduler.scheduleAtFixedRate(this::checkExecutorHealthQuietly, 0L,
+        this.healthCheckIntervalMin,
         TimeUnit.MINUTES);
   }
 
   public void shutdown() {
+    logger.info("Terminating executor health checker.");
```
anish-mal, line 81: The function name is shutdown and message says 'Terminating'. Possible to choose one of the two words and make it consistent?

Shutdown has a meaning of a more orderly than 'terminating.'

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -87,11 +90,27 @@ public void shutdown() {
...
+   * Wrapper for capturing and logging errors and exceptions thrown during healthcheck.
+   * {@code ScheduledExecutorService} stops the scheduled invocations of a given method in
+   * case it throws an exception.
+   * Any exceptions are not expected at this stage however in case any errors and unchecked
+   * exceptions do occur, we still don't want subsequent healthchecks to stop.
+   */
+  public void checkExecutorHealthQuietly() {
+    try {
+      checkExecutorHealth();
+    } catch (Throwable t) {
```
anish-mal, line 103: I feel we need to differentiate between Runtime Exception and java.lang.Error out here. We need to only handle Runtime Exceptions. For java.lang.Error shouldn't we stop the server or raise an alert? (Handling if java.lang.Error can be done as a follow-on, and tracked by filing a ticket).

azkaban-common/src/test/java/azkaban/executor/ExecutorHealthCheckerTest.java
```diff
@@ -145,4 +146,46 @@ public void checkExecutorHealthConsecutiveFailures() throws Exception {
...
+   * Test that the wrapper routine swallows any exceptions reported by underlying health checker.
+   */
+  @Test
+  public void testCheckExecutorHealthWrapperExceptionHandling() throws Exception {
+    this.activeFlows.put(EXECUTION_ID_11, new Pair<>(
+        new ExecutionReference(EXECUTION_ID_11, this.executor1), this.flow1));
+    when(this.apiGateway.callWithExecutionId(this.executor1.getHost(), this.executor1.getPort(),
+        ConnectorParams.PING_ACTION, null, null)).thenThrow(new Error("test error"));
+
+    // this will throw causing the test to fail in case the error is not caught correctly
```
anish-mal, line 160: A comma is missing between throw and causing.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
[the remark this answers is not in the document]
sshardool: Will do.

sshardool: Will do.

azkaban-common/src/test/java/azkaban/executor/ExecutorHealthCheckerTest.java
sshardool: Will do.

PR #2529, opened 2020-05-14
Flow and job events

djaiswal83:
Added upload related attributes to Flow events.
Added projectName and attemptID in job events.


azkaban-common/src/main/java/azkaban/project/ProjectFileHandler.java
```diff
@@ -46,7 +47,8 @@ public ProjectFileHandler(
       final int numChunks,
       final byte[] MD5Hash,
       final Set<Dependency> startupDependencies,
-      final String resourceId) {
+      final String resourceId,
```
abti, line 50: this seems to be changing public constructor. recommended approach for such situations is to create a new constructor with expanded set of vars and wrap it. eg: 

Change THIS: 
public foo(x, y) {
 // yada yada logic
}

TO: 

public foo(x, y) {
 foo(x, y, null);
}
public foo(x, y, z) {
 // yada yada logic
}

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1531,6 +1538,8 @@ public JobRunnerEventListener() {
           props.getString("jetty.hostname", "localhost")));
       metaData.put("jobProxyUser",
           jobRunner.getProps().getString(JobProperties.USER_TO_PROXY, null));
+      // attempt id
+      metaData.put("attempt ID", String.valueOf(node.getAttempt()));
```
abti, line 1542: is no space between attempt and ID in "attempt ID" intentional?

djaiswal83: Yes.

azkaban-common/src/main/java/azkaban/project/ProjectFileHandler.java
djaiswal83: Agreed. However, that approach is good when the constructor is called at many places. In this instance, it is called exactly at one place and the change is required there only.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1464,6 +1458,15 @@ public FlowRunnerEventListener() {
       metaData.put("startTime", String.valueOf(flow.getStartTime()));
       metaData.put("submitTime", String.valueOf(flow.getSubmitTime()));
 
+      // project upload info
+      final ProjectFileHandler handler = flowRunner.projectFileHandler;
+      if (handler != null) {
```
sshardool, line 1463: Can the file handler be null and shouldn't we throw an exception in that case ?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in, line 1542: Currently a job execution attempt is a regular integer in the ExecutableNode class not a separate entity. Wondering if "attemptId" is a good name for it in that case. Also if there were multiple execution attempts will `getAttempt()` return the last of them?

djaiswal83: Got it. I was thinking the other way round. There should not be any space as per the norm.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1519,8 +1522,12 @@ public JobRunnerEventListener() {
       final Props props = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class);
       final Map<String, String> metaData = new HashMap<>();
       metaData.put("jobId", node.getId());
-      metaData.put("executionID", String.valueOf(node.getExecutableFlow().getExecutionId()));
-      metaData.put("flowName", node.getExecutableFlow().getId());
+      // Flow specific properties
+      final ExecutableFlow flow = node.getExecutableFlow();
+      metaData.put("executionID", String.valueOf(flow.getExecutionId()));
```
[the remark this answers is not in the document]
djaiswal83: Will do

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1519,8 +1522,12 @@ public JobRunnerEventListener() {
       final Props props = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class);
       final Map<String, String> metaData = new HashMap<>();
       metaData.put("jobId", node.getId());
-      metaData.put("executionID", String.valueOf(node.getExecutableFlow().getExecutionId()));
-      metaData.put("flowName", node.getExecutableFlow().getId());
+      // Flow specific properties
+      final ExecutableFlow flow = node.getExecutableFlow();
```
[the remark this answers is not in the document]
djaiswal83: Lets discuss the embedded flows.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: This is to keep the existing unit tests happy as the info comes from db where it does not exist. If there is a way to circumvent in tests, I would like to know.
Same reason why there is no unit test for this change.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: Yes, I am renaming it to attemptId.
Yes, the getAttempt() should give the last one. The attempt in ExecutableNode is Atomic Integer which must contain the latest value.

djaiswal83:
> Are there any unit tests that can be updated for this?
> Noted a few other questions/concerns in comments. (Looks fine otherwise)

Unit tests cant be updated as data needs to be fetched from DB.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1520,8 +1523,12 @@ public JobRunnerEventListener() {
       final Props props = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class);
       final Map<String, String> metaData = new HashMap<>();
       metaData.put("jobId", node.getId());
+      // Flow specific properties
+      final ExecutableFlow flow = node.getExecutableFlow();
```
ypadron-in, line 1527: line #1522 above defines the same variable. Can you remove one?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: What I wanted to convey is that attempt is a just number not an Id. Why do you prefer naming it "attemptId"? 
BTW, this number starts in zero, should we report it with a +1? Since attempt 0 might not be make much sense.
How did you verify that `getAttempt()` returns the last attempt?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
ypadron-in: How many tests fail without the conditional? I think we should fix the tests

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1464,6 +1458,15 @@ public FlowRunnerEventListener() {
       metaData.put("startTime", String.valueOf(flow.getStartTime()));
       metaData.put("submitTime", String.valueOf(flow.getSubmitTime()));
 
+      // project upload info
+      final ProjectFileHandler handler = flowRunner.projectFileHandler;
```
ypadron-in, line 1462: Why we're not using a getter to get the projectFileHandler?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1464,6 +1458,15 @@ public FlowRunnerEventListener() {
...
+      if (handler != null) {
+        metaData.put("uploadUser", handler.getUploader());
+        metaData.put("uploaderIpAddr", handler.getUploaderIpAddr());
```
ypadron-in, line 1465: Should we add prefix "_projectFile_" to "uploadUser" and "uploaderIpAddr" properties to maintain consistency with the rest of them?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: I replied to you on slack which should answer the last question. Please take a look at lowRunner.retryJobIfPossible() , ExecutableNode.resetForRetry() as well ExecutionAttempt class.
I don't understand what you mean by attempt is just a number not an id? This number is specific to this particular job and identifies which attempt it is for that job, hence attemptId.
Adding +1 to it may confuse the users as we show attempt 0 for job logs in UI. Also it is pretty common to start the count with 0.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: Good catch. Missed during conflict resolution.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: As it is accessible directly. There is similar usecase in line 1479 for sharedProps object.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: It was becoming too big, but I dig it.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
djaiswal83: flowEventMetadata fails as it tries to generate metadata. The FlowRunner object has projectFileHandler set to NULL as there is no db to pull it from. I dont know how to make it happen for test that is why I kept that null check.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1464,6 +1458,13 @@ public FlowRunnerEventListener() {
       metaData.put("startTime", String.valueOf(flow.getStartTime()));
       metaData.put("submitTime", String.valueOf(flow.getSubmitTime()));
 
+      // project upload info
+      final ProjectFileHandler handler = flowRunner.projectFileHandler;
+      metaData.put("projectFilUploadUser", handler.getUploader());
```
sshardool, line 1463: Fix typo - missing `e` in projectFil...

PR #2542, opened 2020-05-22
Add metric NumAgedQueuedFlows: Counts flows submitted long ago, but n…

anish-mal:
This introduces a metric to track number of flows that have not started running for more than 20 minutes after the web server put those in the queue.

Unit-Test
=======
Writing unit test for duration-dependent functionality is flaky if the underlying code is using System.currentTimeMillis() calls, which the existing code was already doing. So I will be following this up with another change to restructure the existing code and add a unit test.

The follow on change will also make the definition of age configurable through property file. Right now minimum age is hard coded to 20 minutes.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -83,6 +83,8 @@ public class Constants {
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
+  public static final Duration MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED = Duration.ofMinutes(20);
+
   // The flow exec id for a flow trigger instance which hasn't started a flow yet
   public static final int UNASSIGNED_EXEC_ID = -1;
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -123,6 +123,16 @@ public class ExecutionFlowDao {
     }
   }
 
+  public List<Pair<ExecutionReference, ExecutableFlow>> fetchAgedQueuedFlows(final Duration minAge)
+      throws ExecutorManagerException {
+    try {
+      return this.dbOperator.query(FetchAgedQueuedExecutableFlows.FETCH_FLOWS_QUEUED_FOR_LONG_TIME,
+          new FetchQueuedExecutableFlows(), System.currentTimeMillis() - minAge.toMillis());
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error fetching active flows", e);
+    }
+  }
+
   /**
    * fetch flow execution history with specified {@code projectId}, {@code flowId} and flow start
    * time >= {@code startTime}
@@ -502,6 +512,50 @@ public class ExecutionFlowDao {
     private static final String FETCH_QUEUED_EXECUTABLE_FLOW =
         "SELECT exec_id, enc_type, flow_data, status FROM execution_flows"
             + " WHERE executor_id is NULL AND status = "
+            + Status.PREPARING.getNumVal();
+
+    @Override
+    public List<Pair<ExecutionReference, ExecutableFlow>> handle(final ResultSet rs)
+        throws SQLException {
+      if (!rs.next()) {
+        return Collections.emptyList();
+      }
+
+      final List<Pair<ExecutionReference, ExecutableFlow>> execFlows =
+          new ArrayList<>();
+      do {
+        final int id = rs.getInt(1);
+        final int encodingType = rs.getInt(2);
+        final byte[] data = rs.getBytes(3);
+
+        if (data == null) {
+          ExecutionFlowDao.logger.error("Found a flow with empty data blob exec_id: " + id);
+        } else {
+          final EncodingType encType = EncodingType.fromInteger(encodingType);
+          final Status status = Status.fromInteger(rs.getInt(4));
+          try {
+            final ExecutableFlow exFlow =
+                ExecutableFlow.createExecutableFlow(
+                    GZIPUtils.transformBytesToObject(data, encType), status);
+            final ExecutionReference ref = new ExecutionReference(id);
+            execFlows.add(new Pair<>(ref, exFlow));
+          } catch (final IOException e) {
+            throw new SQLException("Error retrieving flow data " + id, e);
+          }
+        }
+      } while (rs.next());
+
+      return execFlows;
+    }
+  }
+
+  private static class FetchAgedQueuedExecutableFlows implements
+      ResultSetHandler<List<Pair<ExecutionReference, ExecutableFlow>>> {
+
+    // Select queued unassigned flows
+    private static final String FETCH_FLOWS_QUEUED_FOR_LONG_TIME =
+        "SELECT exec_id, enc_type, flow_data, status FROM execution_flows"
+            + " WHERE submit_time < ? AND status = "
             + Status.PREPARING.getNumVal();
 
     @Override
--- a/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
+++ b/azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
@@ -441,6 +441,24 @@ public class MockExecutorLoader implements ExecutorLoader {
   }
 
   @Override
+  public List<Pair<ExecutionReference, ExecutableFlow>> fetchAgedQueuedFlows(final Duration minAge)
+      throws ExecutorManagerException {
+    final List<Pair<ExecutionReference, ExecutableFlow>> agedQueuedFlows =
+        new ArrayList<>();
+
+    long timeThreshoold = System.currentTimeMillis() - minAge.toMillis();
+    for (final int execId : this.refs.keySet()) {
+      if (!this.executionExecutorMapping.containsKey(execId)) {
+        ExecutableFlow agedFlow = this.flows.get(execId);
+        if (agedFlow.getSubmitTime() < timeThreshoold) {
+          agedQueuedFlows.add(new Pair<>(this.refs.get(execId), this.flows.get(execId)));
+        }
+      }
+    }
+    return agedQueuedFlows;
+  }
+
+  @Override
   public void unassignExecutor(final int executionId) throws ExecutorManagerException {
     this.executionExecutorMapping.remove(executionId);
   }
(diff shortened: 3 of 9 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -83,6 +83,8 @@
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
+  public static final Duration MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED = Duration.ofMinutes(20);
```
sshardool, line 86: 20 minutes seems a little too high to me. Is there any empirical evidence guiding this ? (why not 5, 10 or 30 minutes, for example)

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -539,6 +549,50 @@ public int selectAndUpdateExecutionWithLocking(final int executorId, final boole
...
+                ExecutableFlow.createExecutableFlow(
+                    GZIPUtils.transformBytesToObject(data, encType), status);
+            final ExecutionReference ref = new ExecutionReference(id);
+            execFlows.add(new Pair<>(ref, exFlow));
+          } catch (final IOException e) {
+            throw new SQLException("Error retrieving flow data " + id, e);
+          }
+        }
+      } while (rs.next());
+
```
sshardool, line 591: This seems to have significant duplication with `FetchQueuedExecutableFlows`. Could you please see if the existing class can be re-used with an additional query/query-parameter ?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -440,6 +440,24 @@ public Executor fetchExecutorByExecutionId(final int execId) throws ExecutorMana
     return queuedFlows;
   }
 
+  @Override
+  public List<Pair<ExecutionReference, ExecutableFlow>> fetchAgedQueuedFlows(final Duration minAge)
```
sshardool, line 444: Is this currently used anywhere or is it for future unit tests ?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -510,6 +511,13 @@ private void addThreadPoolGauges(final QueuedThreadPool threadPool) {
   private void startWebMetrics() throws Exception {
     this.metricsManager
         .addGauge("WEB-NumQueuedFlows", this.executorManagerAdapter::getQueuedFlowSize);
+
+    // Metric for flows that have been submitted, but haven't started for more than 20 minutes.
+    if (this.props.getBoolean(ConfigurationKeys.AZKABAN_POLL_MODEL, false)) {
```
sshardool, line 516: Would your change work as-is for the older push-based dispatch model? If yes, then should we extend this to include that ? (there will be some testing overhead I suppose) as I believe that model is still used by a couple of instances (and as also more in-line with future containerization related work).
Doesn't necessarily have to be part of this PR, but worth checking.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -510,6 +511,13 @@ private void addThreadPoolGauges(final QueuedThreadPool threadPool) {
...
+      this.metricsManager
+          .addGauge("WEB-NumAgedQueuedFlows", this.executorManagerAdapter::getAgedQueuedFlowSize);
```
sshardool, line 518: Does this automatically care of emitting this metric periodically ?

az-core/src/main/java/azkaban/Constants.java
anish-mal: This is to account for a sudden spike that happens at turn of the hour and day and then ramps down. This number will be made configurable because this is likely to be different. 20 seemed the best default based on the data I saw.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
anish-mal: This is for future unit tests.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
anish-mal: Yes, It gets handled by the Dropwizard library being used.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
anish-mal: Done. Merged it with FetchExecutableFlows.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
anish-mal: Sure. Makes sense, I will look into this. 
There is a 2nd part to this PR. So can check this out too.

The reason, the metric for new dispatch logic is more important is because if recent fixes that are supposed to improve the situation of flows getting stuck in PREPARING state - it will be good to have a metric that measures that aspect well.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -539,6 +549,50 @@ public int selectAndUpdateExecutionWithLocking(final int executorId, final boole
...
   }
 
+  private static class FetchAgedQueuedExecutableFlows implements
+      ResultSetHandler<List<Pair<ExecutionReference, ExecutableFlow>>> {
+
+    // Select queued unassigned flows
+    private static final String FETCH_FLOWS_QUEUED_FOR_LONG_TIME =
+        "SELECT exec_id, enc_type, flow_data, status FROM execution_flows"
+            + " WHERE submit_time < ? AND status = "
+            + Status.PREPARING.getNumVal();
```
ypadron-in, line 559: If we're only interested in unassigned executions "executor_id is NULL" must be added to the WHERE clause. Without it executions that are already dispatched will be retuned as well.

az-core/src/main/java/azkaban/Constants.java
ypadron-in: Defaulting to 20 mins duration leads me to think that the metric "WEB-NumQueuedFlows" quite often during peak hours reports more than 1 queued execution for more than 20 minutes, steadily. Is this or close to this what you're seeing in our metrics?

ypadron-in: I think the right threshold for duration is very much related to a cluster's load and therefore IMO it should be configurable from the beginning.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
ypadron-in, line 518: I believe web server metrics are prefixed and grouped under "AZ-WEB" so adding "WEB" in the metric's name is somewhat redundant. Please, consider renaming this metric to  "num-aged-queued-flows" or similar.

anish-mal: I am only following the convention that's already present in the file. I wouldn'y have settled for this convention to begin with. But at this point, it will lead to a mix.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -123,6 +123,16 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
     }
   }
 
+  public List<Pair<ExecutionReference, ExecutableFlow>> fetchAgedQueuedFlows(final Duration minAge)
+      throws ExecutorManagerException {
+    try {
+      return this.dbOperator.query(FetchAgedQueuedExecutableFlows.FETCH_FLOWS_QUEUED_FOR_LONG_TIME,
+          new FetchQueuedExecutableFlows(), System.currentTimeMillis() - minAge.toMillis());
+    } catch (final SQLException e) {
+      throw new ExecutorManagerException("Error fetching active flows", e);
```
ypadron-in, line 132: Can the error message reflect the purpose of the query more closely? Something like "Error fetching executions queued for a long time"

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
anish-mal: The intent is to gather both the unassigned and assigned ones.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: Has this change been committed? I can't find the commit.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -510,6 +511,13 @@ private void addThreadPoolGauges(final QueuedThreadPool threadPool) {
   private void startWebMetrics() throws Exception {
     this.metricsManager
         .addGauge("WEB-NumQueuedFlows", this.executorManagerAdapter::getQueuedFlowSize);
+
+    // Metric for flows that have been submitted, but haven't started for more than 20 minutes.
```
ypadron-in, line 515: Based on the query it's more accurate to say "Metric for executions that have been submitted, but haven't been dispatched..."

ypadron-in: If we say "started" those 20 minutes include dispatch and preparation steps not just dispatch which I believe is the intention here.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
sshardool: @anish-mal how was this resolved? I am unable to see the change either. 
Perhaps the commit was not pushed ?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
sshardool: In that case could you please add a TODO to reflect this. Something like `// todo(anish-mal): ...` so that we have a explanation for this until the subsequent unit test change is made.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: In that case could you rephrase the comment above the query to reflect this? It's confusing as the comment indicates something that the actual query is not fulfilling.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
ypadron-in: I understand there is currently a mix of naming conventions around metric names. I was hoping to follow the one applied to the majority of metrics in CommonMetrics.java and ExecMetrics.java further on and maybe at some point fix the legacy names.

ypadron-in:
Closing as these changes are already merged in #2549

PR #2543, opened 2020-05-23
Adding an option to disable an option to set flow priority as admin only role

jakhani:
This will allow to enable this feature by setting azkaban.disable.admin.only.permission.for.flow.priority to true.
Once this property is set, users who have rights to execute or schedule a flow, will be able to set flowPriority as flow parameter.


azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -130,20 +132,25 @@ public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
...
    */
   public static void filterAdminOnlyFlowParams(final UserManager userManager,
-      final ExecutionOptions options, final User user) throws ExecutorManagerException {
+      final ExecutionOptions options, final User user, final Props props) throws ExecutorManagerException {
     if (options == null || options.getFlowParameters() == null) {
       return;
     }
 
     final Map<String, String> params = options.getFlowParameters();
+    boolean hasAdminPermission = hasPermission(userManager, user, Type.ADMIN);
```
djaiswal83, line 141: nit : missing final qualifier.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -130,20 +132,25 @@ public static ExecutionOptions parseFlowOptions(final HttpServletRequest req)
...
     // is azkaban Admin
-    if (!hasPermission(userManager, user, Type.ADMIN)) {
-      params.remove(ExecutionOptions.FLOW_PRIORITY);
+    if (!hasAdminPermission) {
       params.remove(ExecutionOptions.USE_EXECUTOR);
     } else {
-      validateIntegerParam(params, ExecutionOptions.FLOW_PRIORITY);
       validateIntegerParam(params, ExecutionOptions.USE_EXECUTOR);
     }
+    boolean disableAdminOnlyForFlowPriority = props.getBoolean(ConfigurationKeys.AZKABAN_DISABLE_ADMIN_ONLY_PERMISSION_FOR_FLOW_PRIORITY, false);
```
djaiswal83, line 148: Indentation.

azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
```diff
@@ -62,26 +64,45 @@ public void TestFilterNonAdminOnlyFlowParams() throws IOException,
     final ExecutableFlow flow = createExecutableFlow();
     final UserManager manager = TestUtils.createTestXmlUserManager();
     final User user = manager.getUser("testUser", "testUser");
-
+    Props props = new Props();
```
djaiswal83, line 67: nit: final, applies to method below.

azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
```diff
@@ -62,26 +64,45 @@ public void TestFilterNonAdminOnlyFlowParams() throws IOException,
...
-        flow.getExecutionOptions(), user);
+        flow.getExecutionOptions(), user, props);
 
     Assert.assertFalse(flow.getExecutionOptions().getFlowParameters()
         .containsKey(ExecutionOptions.FLOW_PRIORITY));
     Assert.assertFalse(flow.getExecutionOptions().getFlowParameters()
         .containsKey(ExecutionOptions.USE_EXECUTOR));
   }
 
+  /* Test that flow priority property is not removed for non-admin user when
```
djaiswal83, line 77: nit : can you please format this like line 45?

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
djaiswal83, line 148: Is it necessary to pass the whole props reference when only the boolean property could be sent?

djaiswal83:
> Because the use of this property is restricted to Azkaban admins only at the time this feature was implemented for the new Dispatch logic (#2140) it was decided to not take into consideration starvation of low priority executions. If more users will have permissions to change priority of an execution I believe we should provide a fix for this issue or at least a way to detect/monitor occurrences of it.
> Please refer to [#2140 (comment)](https://github.com/azkaban/azkaban/pull/2140#discussion_r266620159) for one possible solution for starvation of low priority executions.

Good to know!

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
jakhani: To keep it generic to enable or disable any other property in future, I am passing props.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -336,6 +336,9 @@
         "execution.logs.cleanup.interval.seconds";
     public static final String EXECUTION_LOGS_CLEANUP_RECORD_LIMIT =
         "execution.logs.cleanup.record.limit";
+
```
djaiswal83, line 339: Can you please add a comment warning users about this configuration?

PR #2544, opened 2020-05-27
Remove dependency on JavaFX: use Pair class from azkaban-common

xandris:
Not all JDKs have JavaFX, and there's no reason to use the Pair class from JavaFX over the one from `azkaban-common`.


xandris:
Is there something blocking this? Let's discuss.

PR #2545, opened 2020-05-27
Fix Executor's object identity

xandris:
An Executor's hash code can change if it's `isActive` field changes, which causes it to be misplaced in hashing collections. This damages the failover mechanism, which uses a HashSet. This PR changes `equals` and `hashCode` to only consider immutable fields.


xandris:
Is there something blocking this? Let's discuss.

PR #2548, opened 2020-05-29
Revert back changes reverted earlier for PR#2502 and PR#2546

djaiswal83:
Disable FileSystem Cache for HadoopSecurityManager.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -36,6 +36,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
+import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.security.PrivilegedAction;
@@ -65,7 +66,6 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.JobClient;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.Master;
-import org.apache.hadoop.mapreduce.security.TokenCache;
 import org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier;
 import org.apache.hadoop.mapreduce.server.jobtracker.JTConfig;
 import org.apache.hadoop.mapreduce.v2.api.HSClientProtocol;
@@ -115,6 +115,8 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
   public static final String TOKEN_FILE_PERMISSIONS = "460";
   private static final String FS_HDFS_IMPL_DISABLE_CACHE =
       "fs.hdfs.impl.disable.cache";
+  private static final String FS_FAILOVER_IMPL_DISABLE_CACHE =
+          "fs.failover.impl.disable.cache";
   private static final String OTHER_NAMENODES_TO_GET_TOKEN = "other_namenodes";
   private static final String AZKABAN_KEYTAB_LOCATION = "proxy.keytab.location";
   private static final String AZKABAN_PRINCIPAL = "proxy.user";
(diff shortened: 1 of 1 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -565,40 +566,74 @@ private void fetchJobTrackerToken(final String userToProxy, final Props props,
...
+   * @throws HadoopSecurityManagerException
+   */
+
+  private void fetchNameNodeTokenInternal(final String renewer, final Credentials cred,
+          final String userToProxy, final URI uri)
+          throws IOException, HadoopSecurityManagerException {
+    FileSystem fs = null;
+    try {
+      if (uri == null) {
+        fs = FileSystem.get(conf);
```
ibuenros, line 615: nit: Can you put a comment here explaining that `get(conf)` is important as opposed to `newInstance(...)` because the disable cache flag is only checked on the former?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -115,6 +115,8 @@
   public static final String TOKEN_FILE_PERMISSIONS = "460";
   private static final String FS_HDFS_IMPL_DISABLE_CACHE =
       "fs.hdfs.impl.disable.cache";
+  private static final String FS_FAILOVER_IMPL_DISABLE_CACHE =
```
ibuenros, line 118: Can you add comments explaining why you're doing this?

abhishekdas99:
LGTM..

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -115,6 +115,9 @@
   public static final String TOKEN_FILE_PERMISSIONS = "460";
   private static final String FS_HDFS_IMPL_DISABLE_CACHE =
       "fs.hdfs.impl.disable.cache";
+  // Some hadoop clusters have failover name nodes.
+  private static final String FS_FAILOVER_IMPL_DISABLE_CACHE =
```
virajith, line 119: There might be other schemes in use here and cache has to be disabled for those as well. You would have to look at `fs.defaultFS` in Configuration and disable for the associated scheme as well.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -170,12 +176,17 @@ private HadoopSecurityManager_H_2_0(final Props props)
...
-    }
+    // Disable FileSystem Cache for HadoopSecurityManager
+    this.conf.setBoolean(FS_HDFS_IMPL_DISABLE_CACHE, true);
+    this.conf.setBoolean(FS_FAILOVER_IMPL_DISABLE_CACHE, true);
+    // Get the default scheme
+    final String defaultFS = conf.get(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY);
+    final String scheme = new Path(defaultFS).toUri().getScheme();
+    // Construct the property name
+    final String FS_DEFAULT_IMPL_DISABLE_CACHE =
+            "fs." + scheme + IMPL_DISABLE_CACHE_SUFFIX;
```
virajith, line 187: You need a null check here. Also, does using a static Configuration make sense here?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -170,12 +176,17 @@ private HadoopSecurityManager_H_2_0(final Props props)
...
-          + props.get(FS_HDFS_IMPL_DISABLE_CACHE));
-      this.conf.setBoolean(FS_HDFS_IMPL_DISABLE_CACHE,
-          Boolean.valueOf(props.get(FS_HDFS_IMPL_DISABLE_CACHE)));
-    }
+    // Disable FileSystem Cache for HadoopSecurityManager
+    this.conf.setBoolean(FS_HDFS_IMPL_DISABLE_CACHE, true);
+    this.conf.setBoolean(FS_FAILOVER_IMPL_DISABLE_CACHE, true);
+    // Get the default scheme
+    final String defaultFS = conf.get(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY);
+    final String scheme = new Path(defaultFS).toUri().getScheme();
```
virajith, line 184: null check here as well for `defaultFS`

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -565,40 +578,76 @@ private void fetchJobTrackerToken(final String userToProxy, final Props props,
...
+    } else {
+      logger.info(
+              HadoopSecurityManager_H_2_0.OTHER_NAMENODES_TO_GET_TOKEN + " was not configured");
+    }
+  }
+
+  /**
+   * fetchNameNodeInternal - With modified UGI which is of the format,
+   * <userToProxy>/az_<host name>_<exec_id>
+   * Due to this change, the FileSystem cache creates an entry per execution instead of an entry
```
sshardool, line 620: Nit: update the comment to reflect that cache should be disabled before this methods is executed. As FS.get() will make an entry in the cache otherwise.

djaiswal83: good one!

PR #2549, opened 2020-05-30
Add metric NumAgedQueuedFlows: Count flows that are in queue for a long time

anish-mal:
Add a metric to track number of flows that have not started running for more than 20 minutes (configurable) after the web server put those in the queue.

The metric provides a strong signal that an underlying problem (system overload or sluggish/overworked executors) have started impacting the overall performance of the system.

This is different from:

- The state where large number of flows are in the queue, but all the flows in there are young.
- Measuring polling interval of executors, which does not indicate whether a slowed down polling has impacted the intake from the queue (Please note there could be multiple executors) 

Unit-Test
Writing unit test for duration-dependent functionality is flaky if the underlying code is using System.currentTimeMillis() calls, which the existing code was already doing. So I will be following this up with another change to restructure the existing code and add a unit test.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -83,6 +83,8 @@ public class Constants {
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
+  public static final int MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = 20;
+
   // The flow exec id for a flow trigger instance which hasn't started a flow yet
   public static final int UNASSIGNED_EXEC_ID = -1;
 
@@ -182,6 +184,8 @@ public class Constants {
     public static final String METRICS_SERVER_URL = "azkaban.metrics.server.url";
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
+    public static final String MIN_AGE_FOR_CLASSIFYING_A_JOB_AGED_MINUTES = "azkaban.metrics"
+        + ".min_age_for_classifying_a_job_aged_minutes";
 
     // User facing web server configurations used to construct the user facing server URLs. They are useful when there is a reverse proxy between Azkaban web servers and users.
     // enduser -> myazkabanhost:443 -> proxy -> localhost:8081
@@ -240,6 +244,7 @@ public class Constants {
         "azkaban.event.reporting.kafka.topic";
     public static final String AZKABAN_EVENT_REPORTING_KAFKA_SCHEMA_REGISTRY_URL =
         "azkaban.event.reporting.kafka.schema.registry.url";
+
 
     /*
      * The max number of artifacts retained per project.
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -83,7 +83,7 @@ public class Constants {
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
-  public static final int MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = 20;
+  public static final int DEFAULT_MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = 20;
 
   // The flow exec id for a flow trigger instance which hasn't started a flow yet
   public static final int UNASSIGNED_EXEC_ID = -1;
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -184,8 +184,8 @@ public class Constants {
     public static final String METRICS_SERVER_URL = "azkaban.metrics.server.url";
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
-    public static final String MIN_AGE_FOR_CLASSIFYING_A_JOB_AGED_MINUTES = "azkaban.metrics"
-        + ".min_age_for_classifying_a_job_aged_minutes";
+    public static final String MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = "azkaban.metrics"
+        + ".min_age_for_classifying_a_flow_aged_minutes";
 
     // User facing web server configurations used to construct the user facing server URLs. They are useful when there is a reverse proxy between Azkaban web servers and users.
     // enduser -> myazkabanhost:443 -> proxy -> localhost:8081
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -15,6 +15,7 @@
  */
 package azkaban.executor;
 
+import azkaban.Constants;
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.event.EventHandler;
 import azkaban.flow.FlowUtils;
@@ -319,6 +320,21 @@ public class ExecutionController extends EventHandler implements ExecutorManager
       size = this.executorLoader.fetchQueuedFlows().size();
     } catch (final ExecutorManagerException e) {
       this.logger.error("Failed to get queued flow size.", e);
+    }
+    return size;
+  }
+
+  @Override
+  public long getAgedQueuedFlowSize() {
+    long size = 0L;
+    int minimum_age_minutes = this.azkProps.getInt(
+        ConfigurationKeys.MIN_AGE_FOR_CLASSIFYING_A_JOB_AGED_MINUTES,
+        Constants.MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES);
+    try {
+      size = this.executorLoader.fetchAgedQueuedFlows(Duration.ofMinutes(minimum_age_minutes))
+          .size();
+    } catch (final ExecutorManagerException e) {
+      this.logger.error("Failed to get flows queued for a long time.", e);
     }
     return size;
   }
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
@@ -32,7 +32,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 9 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -323,6 +324,21 @@ public long getQueuedFlowSize() {
     return size;
   }
 
+  @Override
+  public long getAgedQueuedFlowSize() {
+    long size = 0L;
+    int minimum_age_minutes = this.azkProps.getInt(
```
sshardool, line 330: Are invalid values for this (such a negative) being checked anywhere in the stack ?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
```diff
@@ -93,6 +93,7 @@
 import org.apache.log4j.jmx.HierarchyDynamicMBean;
 import org.apache.velocity.app.VelocityEngine;
 import org.joda.time.DateTimeZone;
+import org.joda.time.Duration;
```
sshardool, line 96: Where is this used ?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -182,6 +184,8 @@
     public static final String METRICS_SERVER_URL = "azkaban.metrics.server.url";
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
+    public static final String MIN_AGE_FOR_CLASSIFYING_A_JOB_AGED_MINUTES = "azkaban.metrics"
+        + ".min_age_for_classifying_a_job_aged_minutes";
```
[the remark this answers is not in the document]
anish-mal: Here are the occurrences on which I based the naming:

public static final String RESOURCE_MANAGER_JOB_URL = ```
"azkaban.server.external.resource_manager_job_url";
    public static final String HISTORY_SERVER_JOB_URL = "azkaban.server.external.history_server_job_url";
    public static final String SPARK_HISTORY_SERVER_JOB_URL = "azkaban.server.external.spark_history_server_job_url";
```

I do see inconsistency, but this seemed the most common and most readable friendly.

File: az-core/src/main/java/azkaban/Constants.java [hunk 4]
```diff
@@ -83,6 +83,8 @@
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
+  public static final int MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = 20;
```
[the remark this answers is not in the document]
anish-mal: Added 'default' to the one that's the default.

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
anish-mal: There are a lot of property where a non negative number will be expected, and supporting this within the props file will be more appropriate I think. Not having this in props is a shortcoming in the current system, for e.g:

```
final int connectionRequestTimeout =
        props.getInt(JOBCALLBACK_CONNECTION_REQUEST_TIMEOUT, DEFAULT_TIME_OUT_MS);
    final int connectionTimeout = props.getInt(JOBCALLBACK_CONNECTION_TIMEOUT, DEFAULT_TIME_OUT_MS);
    final int socketTimeout = props.getInt(JOBCALLBACK_SOCKET_TIMEOUT, DEFAULT_TIME_OUT_MS);
    this.responseWaitTimeoutMS =
        props.getInt(JOBCALLBACK_RESPONSE_WAIT_TIMEOUT, DEFAULT_RESPONSE_WAIT_TIME_OUT_MS);
```

There are no positivity or non-negativity checks for our props.We can introduce methods in props file such as:

getPositiveInt()
getNonNegativeInt().

What do you think?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java
anish-mal: Deleted. Remnant of an earlier change.

az-core/src/main/java/azkaban/Constants.java
ypadron-in, line 188: Please consider using FLOW instead of JOB in the name as the new metric detects delayed flow executions not job executions.

anish-mal: Done

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -461,6 +471,11 @@ public int selectAndUpdateExecutionWithLocking(final int executorId, final boole
         "SELECT exec_id, enc_type, flow_data, status FROM execution_flows "
             + "WHERE project_id=? AND flow_id=? AND status=? "
             + "ORDER BY exec_id DESC LIMIT ?, ?";
+    // Fetch flows that are in preparing state for more than a certain duration.
+    private static final String FETCH_FLOWS_QUEUED_FOR_LONG_TIME =
+        "SELECT exec_id, enc_type, flow_data, status FROM execution_flows"
+            + " WHERE submit_time < ? AND status = "
+            + Status.PREPARING.getNumVal();
```
ypadron-in, line 478: Is it possible to return the number of executions matching the criteria only? Since this query is executed every minute and it's only used for monitoring purposes it would be good to avoid the overhead of converting the `flow_data` blob into a Java object.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -549,6 +549,13 @@ public long getQueuedFlowSize() {
     return this.queuedFlows.size();
   }
 
+  @Override
+  public long getAgedQueuedFlowSize() {
+    logger.error("Unsupported Function: getQueuedFlowSize() - Retrieving of Aged flows that are "
```
ypadron-in, line 554: Did you mean get**Aged**QueuedFlowSize() in the message?

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
```diff
@@ -440,6 +440,26 @@ public Executor fetchExecutorByExecutionId(final int execId) throws ExecutorMana
     return queuedFlows;
   }
 
+  @Override
+  // TODO(anish-mal) To be used in a future unit test, once System calls to obtain
+  // current time have been replaced by Clocks. Clocks are needed in order to write
+  // unit tests for duration based features. Without it, the tests end up being flaky.
+  public List<ExecutableFlow> fetchAgedQueuedFlows(final Duration minAge)
+      throws ExecutorManagerException {
```
ypadron-in, line 448: how we're using this method in the tests?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
anish-mal: Originally, I started in the direction suggested by you. It involved creating a different class parallel to FetchExecutableFlows. This new class was a 3rd class for operations that were very similar. I was asked to merge the class into the existing one. 

The existing classes, one of which I am reusing, were doing redundant work when it came to computing number of flows:
- "WEB-NumQueuedFlows",  is using this.executorManagerAdapter::getQueuedFlowSize.
- "WEB-NumRunningFlows", is using this.executorManagerAdapter.getRunningFlows().size()

(These existing gauges, in fact, have an added redundant burden of pairing up with ExecutionReference as well.)

I then thought I will reuse existing class for now, especially because this wasn't having any material impact on the performance of Web server's core operations, and then follow-up to pull out NumQueuedFlow and NumQueuedFlows.  Let's do this as a follow-on.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
anish-mal: Yes, changed.

azkaban-common/src/test/java/azkaban/executor/MockExecutorLoader.java
anish-mal: We are not using it yet because any unit test created will be flaky. Plan is to first start using Clock rather than System.currentTime(). This is explained in the function comments just above these two lines.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
ypadron-in: Ok. Could you add a TODO comment summarizing what's pending?

anish-mal: Done

azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
sshardool: I would think that for the timeout related configs negative values are acceptable and would typically imply `no-timeout`
For most other 'time' based configs typically the underlying libraries will take care of looking for illegal arguments e.g In this case `scheduleAtFixedRate` has an explicit check for non-positive values  internally.
 ```    public void start() {
      this.scheduler.scheduleAtFixedRate(() -> pollExecution(), 0L, this.pollingIntervalMs,
          TimeUnit.MILLISECONDS);
    }
```
In your specific usage things will fail silently with the db simply not returning any values. 
A simple check within this method would be fine for this. More exhaustive solutions like modifying the `props` are likely not needed and can be taken up in future if really required.

Comment (anish-mal) [reply azkaban-common/src/main/java/azkaban/executor/ExecutionController.java:332]: I have made the change. But I still wanted to post a clarification  

"> In your specific usage things will fail silently with the db simply not returning any values."

Values would have been returned. Let's go through an example to see what would have happened:

A negative value will have the same effect as a zero value.
Let's say the 
age = -50.
current_time = 800
cutoff = current_time - age = 800 - (-50) = 850
For the SQL query, the WHERE clause for submit_time will become:
(submit_time < 850).

The query wouldn't have failed. It would have returned valid results. Now, Semantic of negative age is wierd (same as negative timeouts or many other thresholds), but that wouldn't have made the query return no results.

Comment (sshardool) [reply azkaban-common/src/main/java/azkaban/executor/ExecutionController.java:332]: You are right, for this query the result set would not be empty, whether it's valid would depend on semantics we want and documented accordingly (at least in code)
Thanks for the clarification.

File: azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServer.java [hunk 8]
```diff
@@ -515,8 +515,17 @@ private void startWebMetrics() throws Exception {
     // (N is configurable by MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES).
     // ToDo(anish-mal) Enable this for push based dispatch logic.
     if (this.props.getBoolean(ConfigurationKeys.AZKABAN_POLL_MODEL, false)) {
-      this.metricsManager
-          .addGauge("WEB-NumAgedQueuedFlows", this.executorManagerAdapter::getAgedQueuedFlowSize);
+      int minAge =
+          this.props.getInt(ConfigurationKeys.MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES,
+              Constants.DEFAULT_MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES);
+      if (minAge >= 0) {
```
sshardool, line 521: Would have preferred this check in the Dao so that any other future users of the method are subject to the same check. 
However likelihood of running into this is low and given that you are planning subsequent changes to this, I am fine with retaining it here.

PR #2550, opened 2020-06-01
Fix DB connection leak in getValidationStatuses

li-gsalia:
getValidationStatuses() gets a DB connection directly from the connection pool
instead of using DatabaseOperator.query. If a connection is fetched directly then
the method is also responsible for managing the lifecycle of the DB connection.

This change adds lifecycle management logic for Connection, ResultSet and
PreparedStatement objects.

Added a TODO to move to DatabaseOperator.query() since it takes care of the
lifecycle management for all these objects. This fix is time sensitive and
hence change to use query() will be done in a follow-up PR.

Tested the fix and verified no DB connection leak is present in getValidationStatuses().

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
@@ -19,6 +19,7 @@ package azkaban.project;
 import azkaban.db.DatabaseOperator;
 import azkaban.spi.Dependency;
 import azkaban.spi.FileValidationStatus;
+import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -27,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.dbutils.DbUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -56,35 +58,59 @@
...
+      hashAndFileNameToDep.values().stream()
+          .forEach(d -> depValidationStatuses.put(d, FileValidationStatus.NEW));
+    } catch (final SQLException ex) {
+      log.error("Transaction failed: ", ex);
+      throw ex;
+    } finally {
+        // Replicate the order of closing in org.apache.commons.dbutils.QueryRunner#query
+        DbUtils.closeQuietly(rs);
+        DbUtils.closeQuietly(stmnt);
+        DbUtils.closeQuietly(conn);
```
sshardool, line 111: Nit 1 -  consider replacing with: `public static void closeQuietly(Connection conn, Statement stmt, ResultSet rs)` in DBUtils.
Nit 2 - include a log.debug() indicating that db connection resources are being cleaned up.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -56,35 +58,59 @@
...
+
+    // TODO: Use azkaban.db.DatabaseOperator.query() instead of getting the DB connection and
+    // dealing with connection lifecycle.
+
+    try {
+      conn = this.dbOperator.getDataSource().getConnection();
+      if (conn == null) {
+        throw new SQLException("Null connection");
+      }
+      conn.setAutoCommit(false);
```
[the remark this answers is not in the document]
li-gsalia: Removed setting autocommit to false.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
li-gsalia: Addressed 1. 

For 2, it will be better to add this in query(). Since this change will be replaced by query() soon, not adding it here.

PR #2555, opened 2020-06-11
Add API documentation for fetchuserprojects.

haroldl:
This API was undocumented but I found it by looking through
the code and I'm using it, so I thought we should add docs.

Here is a screenshot of the rendered HTML after running make html:

fetchuserprojects


haroldl:
I rebased to pick up the azkaban/master commits that have happened since I first wrote this up. Also, somehow a commit from anish-mal got into my fork and this pull request so I've pulled out that commit. This is a documentation only commit.

djaiswal83 (approved):
Thanks for making this change. Your contribution is appreciated by the community!

PR #2560, opened 2020-06-18
use POST instead of GET for executing flow

wilson-lauw:
because GET has url length limitation, this will cause problem when we have a large flow, and we disable a lot of the job inside, as the disabled jobs list will need to be included in the request param.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -83,6 +83,8 @@ public class Constants {
 
   public static final Duration MIN_FLOW_TRIGGER_WAIT_TIME = Duration.ofMinutes(1);
 
+  public static final int DEFAULT_MIN_AGE_FOR_CLASSIFYING_A_FLOW_AGED_MINUTES = 20;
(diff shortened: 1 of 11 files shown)


wilson-lauw:
Updated the documentation as well.

ypadron-in:
Hi @wilson-lauw,

So sorry for the delays! 
Could you please update the curl example in the documentation [here](https://azkaban.readthedocs.io/en/latest/ajaxApi.html#execute-a-flow) to use the POST HTTP method instead? 

Also, could you please resolve the conflicts to merge this PR? I'm not sure why commit https://github.com/azkaban/azkaban/pull/2560/commits/6126379ef6386a630aa9068e9b652810ebeddd2e which belongs to PR #2549 is showing up in this PR.

wilson-lauw:
let me check on that, seems like someone force push on the master branch on the main repo causing that commit to be gone from the master branch

wilson-lauw:
I recreated the PR here: https://github.com/azkaban/azkaban/pull/2613

PR #2562, opened 2020-06-26
Formatting only: apply Save Actions on some files

juhoautio:


juhoautio:
@djaiswal83 @ypadron-in how about merging this? To have a cleaner diff in https://github.com/azkaban/azkaban/pull/2523.

juhoautio:
@ypadron-in, thanks for approving. Anything to prevent merging this? I don’t have the permission.

PR #2580, opened 2020-08-12
[UI] Improve readability of long job property values

ypadron-in:
Job property values are presented in the UI as a single line of characters. This makes it difficult to read values such as: classpaths separated by comma or colon, list of options that are passed to a command/process.
To improve readability in such cases values following below patterns will be formatted in the UI by adding a new line where appropriate.
```
--param1 value1 --param2 --param3 value3
-param1 value -param2 -param3
something,something,something
something:something:something
```

Example:
Before this change:
Screen Shot 2020-08-12 at 1 34 45 AM

After:
Screen Shot 2020-08-12 at 12 37 34 AM


azkaban-web-server/src/web/js/azkaban/view/job-edit.js
```diff
@@ -260,4 +259,44 @@ $(function () {
...
+
+  // Format values like: something,something,something
+  patternsToFormat.set(/^[^,\s]+(,[^,\s]+)+$/, function(str) {
+    return str.replace(/,/g, ",</br>");
+  });
+
+  // Format values like: something:something:something
+  patternsToFormat.set(/^[^,\s;]+(:[^,\s;]+){5,}$/, function(str) {
+    return str.replace(/:/g, ":</br>");
+  });
```
li-gsalia, line 290: Since the only different between the 2 regexs is , and : - can you factor it out in a function that takes the delimiting character as input?

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
```diff
@@ -260,4 +259,44 @@ $(function () {
...
 });
+
+
+// Improve readability of long job property values
+function formatJobPropertyValues() {
+  var patternsToFormat = new Map();
+
+  // Format values like:
+  // 1) --param1 value1 --param2 --param3 value3
+  // 2) -param1 value -param2 -param3
```
li-gsalia, line 274: Can you also add the properties after formatting in the comment (here and below). That way the before and after are both clear.

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
```diff
@@ -260,4 +259,44 @@ $(function () {
   jobEditView = new azkaban.JobEditView({
     el: $('#job-edit-pane')
   });
+
+  formatJobPropertyValues();
+
 });
+
+
+// Improve readability of long job property values
```
li-gsalia, line 268: Can you please add some more details about how the long properties are formatted? The PR description has a very good summary, you can copy text from there :)

The common theme for the function is that instead of multiple properties on 1 line it will split them up so that there is only 1 property per line. A comment mentioning that will provide good context about the function.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobpage.vm
```diff
@@ -21,7 +21,7 @@
   #parse ("azkaban/webapp/servlet/velocity/style.vm")
   #parse ("azkaban/webapp/servlet/velocity/javascript.vm")
 
-  <script type="text/javascript" src="${context}/js/azkaban/view/job-edit.js"></script>
+  <script type="text/javascript" src="${context}/js/azkaban/view/job-edit.js?v=1597214414"></script>
```
li-gsalia, line 24: What does this change do? Is "1597214414" a well known number in velocity?

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
djaiswal83: You beat me to it :)

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobpage.vm
djaiswal83: It is the timestamp. This timestamp guarantees that a page older than it will reload using newer code.

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
ypadron-in: Actually, the "," and ":" are not the only difference.
-the regex for comma separated values use: [^,\s]
-the regex for colon separated values use: [^,\s;] - this is intentional, to avoid splitting values by colon where the real delimiter might be a semi-colon, a comma or a space.
Even if it's safe to add ; to the regex for comma separated values for now I would prefer not to encapsulate this in a function because I feel a variable for the delimiter will make the regular expression harder to understand.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobpage.vm
ypadron-in: Deepak is correct. That's a trick to change the url of the js file and therefore force its reload for all clients.

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
ypadron-in: Done, thanks.

azkaban-web-server/src/web/js/azkaban/view/job-edit.js
ypadron-in: Done.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobpage.vm
li-gsalia: Thanks for the explanation.

PR #2583, opened 2020-08-14
Logging improvements for job callbacks

juhoautio:
Especially, if a bad URL was set, the message was confusing: "Encountered error while waiting for job callback to complete" and didn't include job id for contex. Also the callback URL was not logged at any point, only the "postActualBody" was.

So, added:
- log the formatted callback URL before trying to execute the request
- include job id for context in case of failure


djaiswal83:
Will leave it for another day for others to take a look.

PR #2585, opened 2020-08-17
Bring back Multiple Hadoop cluster (Robin) changes and fix LocalFileSystem leak.

haibchen:
Multiple Hadoop cluster support changes are previously reverted due to a LocalFileSystem memory leak in FileSystem cache caused by changing HadoopSecurityManager from a singleton to a non-singleton.

This PR pulls back all these changes with a proper fix of the memory leak.


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -373,14 +393,158 @@ public Job buildJobExecutor(final String jobId, Props jobProps, final Logger log
...
+   * Create an instance of Job with the given parameters, job id and job logger.
+   */
+  public static Job createJob(final String jobId, final JobParams jobParams, final Logger logger) {
+    Job job;
+    try {
+      job =
+          (Job) Utils.callConstructor(jobParams.jobClass, jobId, jobParams.pluginLoadProps,
+              jobParams.jobProps, jobParams.pluginPrivateProps, logger);
+    } catch (final Exception e) {
+      logger.info("Failed with 5 inputs with exception e = "
```
jakhani, line 446: 6 inputs

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -373,14 +393,158 @@ public Job buildJobExecutor(final String jobId, Props jobProps, final Logger log
...
+
+    final String javaLibPath = Cluster.getJavaLibraryPath(sourceProps, components);
+    if (javaLibPath != null && !javaLibPath.isEmpty()) {
+      clusterProps.put(CommonJobProperties.TARGET_CLUSTER_CLASSPATH, javaLibPath);
+    }
+    final String nativeLibPath = Cluster.getNativeLibraryPath(sourceProps, components);
+    if (nativeLibPath != null && !nativeLibPath.isEmpty()) {
+      clusterProps.put(CommonJobProperties.TARGET_CLUSTER_NATIVE_LIB, nativeLibPath);
+    }
+    final String hadoopSecurityManagerClazz =
```
jakhani, line 510: Can you please rename it Clazz to class?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -231,6 +198,7 @@ private HadoopSecurityManager_H_2_0(final Props props)
   private void disableFSCache() {
     this.conf.setBoolean(FS_HDFS_IMPL_DISABLE_CACHE, true);
     this.conf.setBoolean(FS_FAILOVER_IMPL_DISABLE_CACHE, true);
+    this.conf.setBoolean(FS_LOCAL_IMPL_DISABLE_CACHE, true);
```
jakhani, line 201: How is this change is tested?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopProxy.java
```diff
@@ -58,11 +61,24 @@ private void init(Props sysProps, Props jobProps, final Logger logger) {
     if (shouldProxy) {
       logger.info("Initiating hadoop security manager.");
       try {
-        hadoopSecurityManager = HadoopJobUtils.loadHadoopSecurityManager(sysProps, logger);
-      } catch (RuntimeException e) {
-        e.printStackTrace();
+        String hadoopSecurityClassName = jobProps.get(HadoopJobUtils.HADOOP_SECURITY_MANAGER_CLASS_PARAM);
```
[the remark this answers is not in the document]
haibchen: addressed

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
haibchen: addressed

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
haibchen: I reproduced the memory leak without this first fix first, verified through heap dump that there are leaked LocalFileSystem instances in File System cache.

Then deployed a version with this fix, verified through heap dump that these instances no longer exist.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
haibchen: addressed

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
jakhani: Thank you for thorough testing.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopProxy.java
```diff
@@ -58,11 +61,24 @@ private void init(Props sysProps, Props jobProps, final Logger logger) {
...
       logger.info("Initiating hadoop security manager.");
       try {
-        hadoopSecurityManager = HadoopJobUtils.loadHadoopSecurityManager(sysProps, logger);
-      } catch (RuntimeException e) {
-        e.printStackTrace();
+        // hadoop.security.manager.class can be configured for each Hadoop cluster and it is exposed to
+        // Hadoop jobs as a job property (See JobTypeManager.getClusterSpecificJobProps()).If no cluster
+        // is configured, we can not find it from job properties. Instead, try to load it from sysProps.
+        // This preserves the existing behavior when no clusters (& routing) are configured.
+        final String hadoopSecurityClassName =
```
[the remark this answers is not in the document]
haibchen: Jobs require their own instance of classloader to support Multiple Hadoop clusters, in order to support multiple versions of Spark/Hive on the same cluster (hence we cannot construct a per-cluster classloader to be shared by all jobs routed to a cluster)

To support one security manager instance per cluster, we would create a per-cluster classloader  in which the security manager instance can be loaded.

Each job gets its own classloader (JobClassloader) and then references to the security manager instance that is loaded in a per-cluster classloader.   This paradigm deviates from the standard parent-child classloader hierarchy, making it unmaintainable and extremely difficult to debug any classloading issues, based on my initial attempt to implement this idea.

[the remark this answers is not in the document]
haibchen: I have created https://issues.apache.org/jira/browse/HADOOP-17214
to globally disable FileSystem caching in HadoopSecurityManager, which is the source of concern.

PR #2588, opened 2020-08-18
Remove null from startup-dependencies.json while writing and reading …

li-gsalia:
…the file.

A couple of users complained about not being able to execute flows. Debugging the
issue revealed that thinZip was enabled for this project and startup dependencies
that are saved in the DB contains a 'null'.

When 'null' is encountered in the dependencies json stored in the DB, an error occurs
while created Dependency object from the json since 'null' is not expected.

This change adds filters and checks to remove 'null' while writing and reading startup
dependencies. Also added unit test to ensure the code checks for null while writing
startup-dependencies.json to disk and while reading the dependencies json object
from the DB.


test/src/main/java/azkaban/test/executions/ThinArchiveTestUtils.java
```diff
@@ -120,6 +130,22 @@ public static String getRawJSONDepsABC() {
         "}";
   }
 
+  /**
+   * Create a string of valid raw json that adds a null to the list of valid dependencies.
+   *
+   * @return raw json string containing valid dependencies and a null.
+   */
+  public static String getRawJSONDepsNullAndABC() {
```
jakhani, line 138: Why the method name has ABC? Can you please share what does A B and C denotes?

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
```diff
@@ -173,4 +178,18 @@ public static void validateDependencyHash(final File f, final Dependency d)
...
   }
+
+  /**
+   * Helper method to filer out null from the set containing Dependency objects.
+   * Eg: If input set: {null, obj1, obj2 }, then output set: {obj1, obj2}
+   *
+   * @param dependencies Set of Dependency objects from which null has to be filtered out.
+   * @return Set of Dependency without any null in it.
+   */
+  public static Set<Dependency> filterNullFromDeps(Set<azkaban.spi.Dependency> dependencies) {
```
jakhani, line 189: Can you please import azkaban.spi package?

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
```diff
@@ -173,4 +178,18 @@ public static void validateDependencyHash(final File f, final Dependency d)
...
+    return dependencies.
```
jakhani, line 190: Can dependencies set itself be null?

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
```diff
@@ -73,7 +77,8 @@ public static DependencyFile getDependencyFile(final File projectFolder, final D
   public static void writeStartupDependencies(final File f,
       final Set<Dependency> dependencies) throws IOException {
     Map<String, Set<Dependency>> outputFormat = new HashMap<>();
-    outputFormat.put("dependencies", dependencies);
+    Set<Dependency> nonNullDependencies = filterNullFromDeps(dependencies);
+    outputFormat.put("dependencies", nonNullDependencies);
```
jakhani, line 81: Can you please replace with outputFormat.put("dependencies", filterNullFromDeps(dependencies));

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
```diff
@@ -91,8 +91,13 @@
       while (rs.next()) {
         // Columns are (starting at index 1): file_name, file_sha1, validation_status
         Dependency d = hashAndFileNameToDep.remove(rs.getString(1) + rs.getString(2));
-        FileValidationStatus v = FileValidationStatus.valueOf(rs.getInt(3));
-        depValidationStatuses.put(d, v);
+
+        // HashMap.remove will return null if the key is not found, hence check for it before
+        // adding to depValidationStatuses.
+        if (d != null) {
```
jakhani, line 97: It will be better to add contains check for key in hashAndFileNameToDep instead of null check over here. All of these should be inside that contains if case.

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
```diff
@@ -272,6 +276,8 @@ private void updateValidationStatuses(final Set<? extends Dependency> untouchedD
         .keySet()
         .stream()
         .filter(d -> validationStatuses.get(d) == status)
+        .filter(Objects::nonNull) // Filter out null since the Set should contain Dependency
```
jakhani, line 279: nonNull should be above filter for validationStatuses

jakhani, line 279: nonNull should be above filter for validationStatuses

test/src/main/java/azkaban/test/executions/ThinArchiveTestUtils.java
li-gsalia: This is a method used only for Unit test and follows the convention in the file. The pre-existing test dependencies have names A, B and C and various methods in the file have the dependency names in the method name.

Since this is for unit test, all the dependencies are dummy and do not have real world significance.

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
li-gsalia: Done.

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
li-gsalia: The Set can be empty, but not null.

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
li-gsalia: Changed.

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
li-gsalia: Changed the order

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
```diff
@@ -281,4 +287,5 @@ private void updateValidationStatuses(final Set<? extends Dependency> untouchedD
         .map(d -> String.format("Removed blacklisted file %s", d.getFileName()))
         .collect(Collectors.toSet());
   }
+
```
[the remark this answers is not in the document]
li-gsalia: It was added accidentally, thanks for catching it. Removed.

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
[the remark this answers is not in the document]
li-gsalia: Removed the FQN.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
li-gsalia: Contains check will lead to 2 lookups in the hasMap for every dependency instead of 1 which is the current case. Let me know if you feel strongly about using contains, I can make the change.

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
djaiswal83: @li-gsalia the order appears same. Did you push the change?

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
djaiswal83: I agree with @li-gsalia about it.

abhisheknath2011: If we are expecting more key miss i.e. keys not present in hashAndFileNameToDep, then contains check looks more logical. But I also agree with @li-gsalia. So will leave it up to @li-gsalia to take a call on this.

azkaban-common/src/main/java/azkaban/utils/ThinArchiveUtils.java
abhisheknath2011: looks like the changes are not published.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
jakhani: +1 with @abhisheknath2011

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
li-gsalia: Pushed the change now, laptop had issues yesterday.

azkaban-common/src/main/java/azkaban/project/JdbcDependencyManager.java
li-gsalia: There should be no miss based on the current code. Added the null check since set.remove() can return null. This will future proof the code in case future code changes lead to a miss in set.

azkaban-common/src/main/java/azkaban/project/ArchiveUnthinner.java
li-gsalia: Pushed the change now, had some laptop issues yesterday.

abhisheknath2011 (approved):
Thanks for identifying the issue and making the changes. Changes looks good to me.

abagshaw:
@li-gsalia Thanks for fixing my bug 😄

An unsolicited 2 cents (maybe you already looked into this, but just in case...): I'd keep an eye out for issues like flows failing with `ClassNotFoundException` in projects that were hitting this `null` dependency issue. It might be the case that the root cause of the problem is on the Gradle side of things (the thinning step at build time that generates the `startup-dependencies.json` file) and some JAR is being thinned (removed from the zip) but failing to be added to the list of thinned dependencies (i.e. it's being added as `null` for some reason) which could result it it being missing from the classpath at runtime.

li-gsalia:
@abagshaw it's good to see you being active on Azkaban! :) 

For the projects where this issue was seen, `startup-dependencies.json` does not have a `null` in it after the build completes. In fact, the behavior is not reproducible for the same `startup-dependencies.json` file. The null gets inserted during byte-ray processing of the dependencies. The new checks added will ensure `null` is not saved in the DB.

Thanks for pointing out the possibility of `null` being added during build time. Will keep an eye out for it.

PR #2590, opened 2020-08-20
Add metrics to API endpoint requests.

ypadron-in:
Following metrics are reported for every API call received by the web server:
-`num-app-get-req`: Records the number of GET requests made by the Azkaban Web application in the last reporting interval.
-`num-app-post-req`: Records the number of POST requests made by the Azkaban Web application in the last reporting interval.
-`num-nonApp-get-req`: Records the number of GET requests NOT made by the Azkaban Web application in the last reporting interval (ex. curl requests).
-`num-nonApp-post-req`: Records the number of POST requests NOT made by the Azkaban Web application in the last reporting interval (ex. curl requests).
-`resp-time-histogram`: Records the time it takes to fulfill the request.

(reapplying a temporarily reverted commit (PR #2582))


ypadron-in:
> Looks familiar -- I'll assume that it's an exact replay of the old PR. As such, approving it "wholesale".

Thanks! It is the same _diff_ as before.

PR #2591, opened 2020-08-24
Add flow execution source to flow properties

li-pkhandelwal:


azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
```diff
@@ -0,0 +1,3 @@
+-- DB Migration from release 3.85.0 to 3.86.0
+-- Adding ExecutionType column in execution_flows
+alter table execution_flows add column execution_type varchar(32) default null;
```
jakhani, line 3: Can you please create enum in mysql? With that release will be different. You might need to change name of upgrade file.

azkaban-common/src/main/java/azkaban/flow/FlowExecutionType.java
```diff
@@ -0,0 +1,27 @@
+package azkaban.flow;
+
+public enum FlowExecutionType {
+  SCHEDULED("scheduled"), ADHOC("adhoc");
+
+  private final String value;
```
jakhani, line 6: If you create enum in database with SCHEDULED and ADHOC as values, there is no need of value instance variable. Name of enum will be enough.

jakhani (commented):
Can you please add unit tests for this PR?

ypadron-in (changes requested):
A few notes on this. 
It's my understanding that the goal of this new property is to record the origin of a particular execution. For example if it was triggered by a user manually or initiated by Azkaban based on an existing schedule. Based on this I would say we're talking about a new property "executionOrigin", "executionSource",  or "executionInitiator" maybe but not exactly "executionType" I believe. For execution types we have identified: executions that reuse data from a previous execution (reruns) and executions which don’t (new/pristine).

Also, I don't think it would be a good idea to define this new property as an enum because they are static and even in today's Azkaban we kind of need an extensible list of origins. Currently, an execution can be started 
-manually (in the UI or via APIs)
-by the scheduler
-by the [FlowTrigger service](https://github.com/azkaban/azkaban/blob/5b8eb08f95374e00459a88fc76a07b13c5d78fb9/azkaban-web-server/src/main/java/azkaban/flowtrigger/TriggerInstanceProcessor.java#L75) 
The FlowTrigger service (https://azkaban.readthedocs.io/en/latest/eventTrigger.html) is essentially an event based trigger mechanism which follows the plugin model. Because it's a plugin, we don't really know the "origin" (there could be multiple even) of an execution started by that service. We could define a default initiator value for it like "EventTriggered" but ideally I think we should allow the plugin to specify a less ambiguous origin name.

azkaban-common/src/main/java/azkaban/flow/FlowExecutionType.java
djaiswal83: +1. Came to say exactly this.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -129,6 +129,11 @@
   // Used (or should be used) wherever a string representation of UTF_8 charset is needed:
   public static final String UTF_8 = java.nio.charset.StandardCharsets.UTF_8.toString();
 
+  // Specifies the source(adhoc, scheduled, flow_trigger) from where flow execution is triggered
+  public static final String EXECUTION_SOURCE_ADHOC = "adhoc";
+  public static final String EXECUTION_SOURCE_SCHEDULED = "scheduled";
+  public static final String EXECUTION_SOURCE_FLOW_TRIGGER = "flow_trigger";
```
ypadron-in, line 135: Shouldn't "source" be a noun? How about?
-USER
-SCHEDULE
-EVENT (because it's an event based trigger mechanism)

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -50,11 +50,13 @@
   public static final String AZKABANFLOWVERSION_PARAM = "azkabanFlowVersion";
   public static final String IS_LOCKED_PARAM = "isLocked";
   public static final String FLOW_LOCK_ERROR_MESSAGE_PARAM = "flowLockErrorMessage";
+  public static final String EXECUTION_SOURCE = "execution_source";
```
ypadron-in, line 53: Could you please use `camelCase` in the property name to follow the naming convention used in the rest of the properties?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -65,8 +65,9 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
 
     final String INSERT_EXECUTABLE_FLOW = "INSERT INTO execution_flows "
         + "(project_id, flow_id, version, status, submit_time, submit_user, update_time, "
-        + "use_executor, flow_priority) values (?,?,?,?,?,?,?,?,?)";
+        + "use_executor, flow_priority, execution_type) values (?,?,?,?,?,?,?,?,?,?)";
```
ypadron-in, line 68: Did you mean `execution_source`?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -77,7 +78,7 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
     final SQLTransaction<Long> insertAndGetLastID = transOperator -> {
       transOperator.update(INSERT_EXECUTABLE_FLOW, flow.getProjectId(),
           flow.getFlowId(), flow.getVersion(), flow.getStatus().getNumVal(),
-          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority);
+          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority, executionType);
```
ypadron-in, line 81: `executionSource`?

azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
```diff
@@ -0,0 +1,3 @@
+-- DB Migration from release 3.85.0 to 3.86.0
+-- Adding ExecutionType column in execution_flows
```
ypadron-in, line 2: execution source column?

azkaban-common/src/main/java/azkaban/flow/CommonJobProperties.java
```diff
@@ -140,6 +140,10 @@
    */
   public static final String SUBMIT_USER = "azkaban.flow.submituser";
 
+  /**
+   * Represents whether the flow is a scheduled flow or an adhoc execution
```
ypadron-in, line 144: Could you also mention the "event triggered" source?

li-pkhandelwal: Done

azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
li-pkhandelwal: Done

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
li-pkhandelwal: Thanks, done

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
li-pkhandelwal: yes, updated

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
li-pkhandelwal: yes, thanks for pointing

az-core/src/main/java/azkaban/Constants.java
li-pkhandelwal: The earlier names were for execution_type. Makes sense, updated the source names. Since the execution can be triggered from command line which might not be from the user manually, keeping the name as adhoc.

ypadron-in (approved):
Thanks for addressing the comments! LGTM.

PR #2598, opened 2020-09-15
Adding enum for Dispatch Method. It can be PUSH, POLL or CONTAINERIZED.

jakhani:
Adding new property for Dispatch Method. Property is azkaban.execution.dispatch.method.
Values of this property can be PUSH, POLL or PUSH_CONTAINERIZED.

1. If it is PUSH then web server will push executions to executors based on capacity of executor service
2. If it is POLL then executors will poll executions based on interval mentioned by azkaban.polling.interval.ms property.
3. If it is PUSH_CONTAINERIZED then web server will dispatch the executions to containerized infrastructure like Kubernetes.


azkaban-common/src/main/java/azkaban/DispatchMethod.java
```diff
@@ -0,0 +1,14 @@
...
+import azkaban.utils.Props;
+
+public enum DispatchMethod {
+    PUSH,
+    POLL,
+    CONTAINERIZED;
+
+    public static DispatchMethod getDispatchModel(Props azkabanProps) {
+        return DispatchMethod.valueOf(
+                azkabanProps.getString(Constants.ConfigurationKeys.AZKABAN_EXECUTION_DISPATCH_METHOD, "PUSH").toUpperCase());
```
djaiswal83, line 12: Should the default be Poll, considering that is what is used in Azkaban these days?

azkaban-common/src/test/java/azkaban/DispatchMethodTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_EXECUTION_DISPATCH_METHOD;
+
+public class DispatchMethodTest {
+    @Test
+    public void testDispatchMethod() {
+        Props props = new Props();
+        props.put(AZKABAN_EXECUTION_DISPATCH_METHOD, "CONTAINERIZED");
+        Assert.assertEquals(DispatchMethod.getDispatchModel(props), DispatchMethod.CONTAINERIZED);
+        props = new Props();
```
djaiswal83, line 15: Nit : Use the same props object and override the value? I know this aint C++, but I am hard wired like that. Feel free to ignore it.

djaiswal83 (changes requested):
Overall the code looks good. Please address couple of nitpicks.

azkaban-common/src/main/java/azkaban/DispatchMethod.java
ypadron-in: I feel like the `getDispatchModel` method would be better suited for a service/"server initialization" class so that enum class doesn't need to know about the existence of an Azkaban properties file, it's just a data type for me. What do you think?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -265,7 +266,7 @@ public FlowRunnerManager(final Props props,
     this.cleanerThread = new CleanerThread();
     this.cleanerThread.start();
 
-    if (this.azkabanProps.getBoolean(ConfigurationKeys.AZKABAN_POLL_MODEL, false)) {
+    if (DispatchMethod.getDispatchModel(azkabanProps) == DispatchMethod.POLL) {
```
ypadron-in, line 269: What happens if the `azkaban.execution.dispatch.method` property isn't well configured and `DispatchMethod.valueOf` throws an IllegalArgumentException exception?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
```diff
@@ -87,8 +89,15 @@ protected void configure() {
   }
 
   private Class<? extends ExecutorManagerAdapter> resolveExecutorManagerAdaptorClassType() {
-    return this.props.getBoolean(ConfigurationKeys.AZKABAN_POLL_MODEL, false)
-        ? ExecutionController.class : ExecutorManager.class;
+    switch(DispatchMethod.getDispatchModel(this.props)) {
```
ypadron-in, line 92: What do you think of reading the `azkaban.execution.dispatch.method` at initialization time here to inject the right class and defining a `getDispatchMethod` method in the ExecutorManagerAdapter interface which will then be implemented as `return DispatchMethod.XXX` by every class extending ExecutorManagerAdapter.

azkaban-common/src/main/java/azkaban/DispatchMethod.java
jakhani: @ypadron-in Let us have two methods. One is for service and another one for property. The reason is I have seen couple of places where property for poll is used in if cases.
@djaiswal83 I have kept PUSH as default because currently if poll is not set deliberately then PUSH is selected by default. So that will maintain consistency with existing behavior.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: Thanks. Took care of this comment.

azkaban-common/src/test/java/azkaban/DispatchMethodTest.java
jakhani: Took care of it. Thank you.

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
jakhani: I have made that change for web server. For executor server, property and getDispatchMethod is needed. I have changed the method definition. It won't take props object anymore.

djaiswal83 (approved):
Thanks for addressing the review comments.
Looks good to me.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
```diff
@@ -265,10 +266,12 @@ public FlowRunnerManager(final Props props,
     this.cleanerThread = new CleanerThread();
     this.cleanerThread.start();
 
-    if (this.azkabanProps.getBoolean(ConfigurationKeys.AZKABAN_POLL_MODEL, false)) {
+    if (DispatchMethod.getDispatchMethod(azkabanProps
+        .getString(Constants.ConfigurationKeys.AZKABAN_EXECUTION_DISPATCH_METHOD, "PUSH"))
+        == DispatchMethod.POLL) {
```
ypadron-in, line 271: nit: is it possible to encapsulate this in a method or maybe create a private attribute `dispatchMethod`?

azkaban-web-server/src/main/java/azkaban/webapp/AzkabanWebServerModule.java
ypadron-in: Sorry, I didn't realize that most of the usages of `getDispatchMethod` were on the executor module were there is no ExecutorManagerAdapter class available. Then, for now, there is no need for a "getDispatchMethod" in the that interface because once the right class is injected branching logic based on dispatch model isn't needed.

jakhani: Yes agree. I am using getDispatchMethod for now in Test case to validate the correct class is injected based on property. So let's keep that method.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java
jakhani: Thank you for reviewing the PR. I have take care of comments.

ypadron-in (approved):
LGTM. Thank you for addressing the comments!

PR #2599, opened 2020-09-16
Add metric to measure the executor polling frequency

ypadron-in:
It will be used to detect when executors stop polling for new executions.


li-gsalia (approved):
Change looks good!

PR #2600, opened 2020-09-18
Restore flow lock message after project upload.

ypadron-in:
Before, only the "isLocked" attribute was set again so lock messages were lost.


djaiswal83 (approved):
Good catch! The code looks good to me.

PR #2603, opened 2020-09-18
Adding enum for Dispatch Method. It can be PUSH, POLL or CONTAINERIZED

jakhani:
Adding new property for Dispatch Method. Property is azkaban.execution.dispatch.method.
Values of this property can be PUSH, POLL or PUSH_CONTAINERIZED.

If it is PUSH then web server will push executions to executors based on capacity of executor service
If it is POLL then executors will poll executions based on interval mentioned by azkaban.polling.interval.ms property.
If it is PUSH_CONTAINERIZED then web server will dispatch the executions to containerized infrastructure like Kubernetes.


ypadron-in (approved):
LGTM. These are the same changes as PR https://github.com/azkaban/azkaban/pull/2598.

PR #2604, opened 2020-09-20
Fix flaky queue duration condition in JobRunnerTest

juhoautio:
Fixes test failure like this:
```
azkaban.execapp.JobRunnerTest > testBasicRun FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at azkaban.execapp.JobRunnerTest.testBasicRun(JobRunnerTest.java:112)
```

Which points to this line:
https://github.com/azkaban/azkaban/blob/783f88d490a65ecbc1283c317eb9f662056f4fba/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java#L112

At least locally this test was often executed so fast that time in queue was set to 0. That must mean that `System.currentTimeMillis()` had not advanced between the times when job was queued and executed. Add 1 ms sleep to ensure that queue time is always at least 1.

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/JobRunnerTest.java
@@ -88,7 +88,7 @@ public class JobRunnerTest {
   }
 
   @Test
-  public void testBasicRun() throws IOException {
+  public void testBasicRun() throws Exception {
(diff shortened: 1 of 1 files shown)


djaiswal83 (approved):
Thanks for fixing the UT.

PR #2607, opened 2020-09-22
Adding utility method to check PUSH and PUSH_CONTAINERIZED dispatch m…

jakhani:
…ethod.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/DispatchMethod.java
+++ b/azkaban-common/src/main/java/azkaban/DispatchMethod.java
@@ -1,12 +1,13 @@
 package azkaban;
 
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public enum DispatchMethod {
   PUSH,
   POLL,
   PUSH_CONTAINERIZED;
-  private static final Logger logger = Logger.getLogger(DispatchMethod.class);
+  private static final Logger logger = LoggerFactory.getLogger(DispatchMethod.class);
 
   public static DispatchMethod getDispatchMethod(String value) {
     try {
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/DispatchMethod.java
```diff
@@ -22,4 +22,12 @@ public static DispatchMethod getDispatchMethod(String value) {
...
     return DispatchMethod.getDispatchMethod(dispatchMethod) == DispatchMethod.POLL;
   }
+
+  public static boolean isPushMethodEnabled(String dispatchMethod) {
+    return DispatchMethod.getDispatchMethod(dispatchMethod) == DispatchMethod.PUSH;
+  }
+
+  public static boolean isPushContainerizedMethodEnabled(String dispatchMethod) {
+    return DispatchMethod.getDispatchMethod(dispatchMethod) == DispatchMethod.PUSH_CONTAINERIZED;
+  }
```
ypadron-in, line 32: Is it possible to rename the methods to `isXXXMethod` to keep the usage by the caller generic? The caller could use the result of these methods to verify whether a feature is enabled or to validate an input value for example.

ypadron-in:
As @djaiswal83 suggested offline, can we use "slf4j" for the logger creation in DispatchMethod.java?

azkaban-common/src/main/java/azkaban/DispatchMethod.java
djaiswal83: +1 on that.

jakhani: There is already generic method called getDispatchMethod and where the validation can be done. I am adding more utility methods so that it can directly be used. It will be useful for containerization specifically.

jakhani:
> As @djaiswal83 suggested offline, can we use "slf4j" for the logger creation in DispatchMethod.java?

Thanks. I have fixed it.

djaiswal83 (approved):
Thanks for implementing the requested changes.
The code looks good to me.

ypadron-in (approved):

PR #2609, opened 2020-09-22
Update Travis CI configs.

ypadron-in:
-use a more recent Linux distribution.
-replace travis-ci.org with travis-ci.com in build badge reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 languages: java
-# TODO: Add openjdk8 once it is available in Travis.
 jdk:
-  - oraclejdk8
-dist: trusty
-sudo: false
+  - openjdk8
+os: linux
+dist: xenial
+
+before_install:
+  - sudo apt update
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,9 +1,13 @@
...
-dist: trusty
-sudo: false
+  - openjdk8
+os: linux
+dist: xenial
+
+before_install:
+  - sudo apt update
+  - sudo apt install openjfx
+
```
ypadron-in, line 10: oraclejdk8 isn't supported on Xenial (https://travis-ci.community/t/install-of-oracle-jdk8-is-failing/4365)
OpenJDK 8 doesn't include JavaFX (https://stackoverflow.com/questions/35974003/javafx-comes-with-jdk-8)

djaiswal83 (approved):
Thanks for making the change and providing links.
Looks good to me.

PR #2611, opened 2020-09-28
Add CSRF token validation for modifying project permissions

aditya1105:
CSRF token validation for modifying project permissions
We have used "Synchronizer token pattern" for CSRF validation. This code generates CSRF token by hashing session-id hence csrf tokens are not kept in memory. 
CSRF token will be sent as an HTTP header rather than request parameters, as it is the more recommended approach


svnarumugam (approved):
This is a copy of the change that was already reviewed. The changes look good to me.

PR #2612, opened 2020-09-28
Fixes failing WebUtils unit test

li-ygerchikov:


ypadron-in (approved):
In the future let's try to find a way to remove Guice dependencies from the WebUtils utility class.
```private static final AzkabanEventReporter azkabanEventReporter =
       ServiceProvider.SERVICE_PROVIDER.getInstance(AzkabanEventReporter.class);```

li-ygerchikov:
> In the future let's try to find a way to remove Guice dependencies from the WebUtils utility class.

Yes, let's. I'm open to suggestions!

PR #2613, opened 2020-09-29
use POST instead of GET for executing flow

wilson-lauw:
because GET has url length limitation, this will cause problem when we have a large flow, and we disable a lot of the job inside, as the disabled jobs list will need to be included in the request param.

### DIFF
--- a/docs/ajaxApi.rst
+++ b/docs/ajaxApi.rst
@@ -622,7 +622,7 @@ This API executes a flow via an ajax call, supporting a rich selection
 of different options. Running an individual job can also be achieved via
 this API by disabling all other jobs in the same flow.
 
--  **Method:** GET
+-  **Method:** POST
 -  **Request URL:** /executor?ajax=executeFlow
 -  **Parameter Location:** Request Query String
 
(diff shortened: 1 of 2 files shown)


docs/ajaxApi.rst
```diff
@@ -622,7 +622,7 @@ This API executes a flow via an ajax call, supporting a rich selection
 of different options. Running an individual job can also be achieved via
 this API by disabling all other jobs in the same flow.
 
--  **Method:** GET
+-  **Method:** POST
 -  **Request URL:** /executor?ajax=executeFlow
 -  **Parameter Location:** Request Query String
```
ypadron-in, line 627: Request `body` maybe?

ismailsimsek:
Hi @wilson-lauw when submitting large workflow with a lot of disabled jobs, nothing is happening submit button is not working, could this be reason?

ismailsimsek (approved):
LGTM

wilson-lauw:
Hi @ismailsimsek , yes it's because of this

ismailsimsek:
cc: @ypadron-in

docs/ajaxApi.rst
ypadron-in: Thanks for fixing the docs for the login endpoint, however, we also need to specify "Request body" instead of "Request query string" for the execute flow endpoint. 
In addition, we should fix the uris of both endpoints as parameters go in the request body for POST requests:
Instead of `/?action=login` it should be just `/`
Instead of  `/executor?ajax=executeFlow` it should be `/executor`

I'll make the change myself in a separate PR.

ypadron-in (approved):

ypadron-in:
Thanks for fixing this!

PR #2614, opened 2020-10-02
Fix startup errors in some environments

li-ygerchikov:
1. Some configuration errors resulted in uncaught exceptions from WebUtils that prevented WS from starting.
2. In some environments Velocity template engine failed to process unescaped dollar sign resulting in 500 error to UI.

Tested (1) on local soloserver, (2) on Pokemon and (3) in "certain" environment (where it failed before) by patching `javascript.vm`, splicing patched file into WS JAR file in `lib/, restarting WS and observing the final rendered templates via Dev Tools in Chrome. In all environments the template rendered successfully and correctly.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
@@ -47,7 +47,7 @@ public class WebUtils {
     try {
       azkabanEventReporter = ServiceProvider.SERVICE_PROVIDER
           .getInstance(AzkabanEventReporter.class);
-    } catch (NullPointerException | ConfigurationException e) {
+    } catch (Exception e) {
       Logger.getLogger(WebUtils.class.getName()).warn("AzkabanEventReporter not configured", e);
     }
   }
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/javascript.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/javascript.vm
@@ -29,7 +29,8 @@
      * used by all the views whenever APIs need to be called. Those wrapper functions should
      * apply application related parameters common to all requests such as this header.
     *#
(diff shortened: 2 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -47,7 +47,7 @@
     try {
       azkabanEventReporter = ServiceProvider.SERVICE_PROVIDER
           .getInstance(AzkabanEventReporter.class);
-    } catch (NullPointerException | ConfigurationException e) {
+    } catch (Exception e) {
```
li-ygerchikov, line 50: We really, really want to catch _everything_ here. It is safe to leave `azkabanEventReporter` set to null.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/javascript.vm
```diff
@@ -29,7 +29,8 @@
      * used by all the views whenever APIs need to be called. Those wrapper functions should
      * apply application related parameters common to all requests such as this header.
     *#
-  $.ajaxSetup({
+  #set ( $d = "$" )
```
li-ygerchikov, line 32: Alternative escaping method:
```
#[[$]]#.ajaxSetup({
```
IMO it's uglier, but works too.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/javascript.vm
```diff
@@ -29,7 +29,8 @@
...
+  ${d}.ajaxSetup({
```
li-ygerchikov, line 33: Alternative escaping method:
```
#[[$]]#.ajaxSetup({
```
IMO it's uglier, but works too, if you/we/anyone prefers this form -- please let me know.

ypadron-in (approved):
LGTM

PR #2622, opened 2020-10-23
Make MetricsManager more robust to changes in metrics reporter Ctor

li-ygerchikov:
When we added second public Ctor with a different signature
to the reporter class, MetricsManager started failing sporadically
depending on which Ctor was the [0]-th in the list.
Tested on local soloserver:
2020-10-23 01:33:28,768 INFO [azkaban.metrics.MetricsManager main] metricsReporterClassName: com.linkedin.amf.reporter.AmfReporter
2020-10-23 01:33:28,771 INFO [com.linkedin.amf.reporter.AmfReporter main] AmfReporter Initializing. serverUrl: ltx1-amf.prod.linkedin.com

### DIFF
--- a/az-core/src/main/java/azkaban/metrics/MetricsManager.java
+++ b/az-core/src/main/java/azkaban/metrics/MetricsManager.java
@@ -120,10 +120,11 @@ public class MetricsManager {
     if (metricsReporterClassName != null && metricsServerURL != null) {
       try {
         log.info("metricsReporterClassName: " + metricsReporterClassName);
-        final Class metricsClass = Class.forName(metricsReporterClassName);
+        final Class<?> metricsClass = Class.forName(metricsReporterClassName);
 
-        final Constructor[] constructors = metricsClass.getConstructors();
-        constructors[0].newInstance(reporterName, this.registry, metricsServerURL);
+        final Constructor<?> ctor = metricsClass.getConstructor(reporterName.getClass(),
+            this.registry.getClass(), metricsServerURL.getClass(), boolean.class);
+        ctor.newInstance(reporterName, this.registry, metricsServerURL, true);
 
       } catch (final Exception e) {
         log.error("Encountered error while loading and instantiating "


svnarumugam (approved):
Thank you so much for making this change @li-ygerchikov !

az-core/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -120,11 +120,8 @@ public synchronized void startReporting(final String reporterName, final Props p
     if (metricsReporterClassName != null && metricsServerURL != null) {
       try {
         log.info("metricsReporterClassName: " + metricsReporterClassName);
-        final Class metricsClass = Class.forName(metricsReporterClassName);
-
-        final Constructor[] constructors = metricsClass.getConstructors();
-        constructors[0].newInstance(reporterName, this.registry, metricsServerURL);
-
+        final Class<?> metricsClass = Class.forName(metricsReporterClassName);
+        Utils.callConstructor(metricsClass, reporterName, this.registry, metricsServerURL);
```
ypadron-in, line 124: Thanks for improving the initialization of metrics! 
Since we're at it :) , could you please add a 4th **boolean** parameter to the constructor? The signature of the constructor we should use is:
 `public AmfReporter(String groupName, MetricRegistry registry, String amfServerUrl, boolean reportCounterAsGauge) {..}`

The parameter reportCountersAsGauge should be **true**.

li-ygerchikov: Added. Unfortunately, the Utils helper that I tried initially can't handle primitive types, so it's back to fetching and calling Ctor explicitly. If we want to do anything smarter (Ctor that takes context object? builder instead of Ctor?), let's do it separately. @ypadron-in

ypadron-in: I agree, thanks.

ypadron-in (approved):
LGTM

PR #2623, opened 2020-10-30
Removing duplicate code in ExecutorManager and ExecutionController

jakhani:
There is duplicate code in ExecutorManager and ExecutionController. Created AbstractExecutorManagerAdapter which has common code.


djaiswal83 (approved):
Thanks for refactoring the code. This is something we should have probably done when ExecutionController was introduced.
The changes look good.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -0,0 +1,450 @@
...
+          this.apiGateway.callWithReferenceByUser(pair.getFirst(),
+              ConnectorParams.MODIFY_EXECUTION_ACTION, userId,
+              new Pair<>(
+                  ConnectorParams.MODIFY_EXECUTION_ACTION_TYPE, command));
+    }
+    return response;
+  }
+
+  /* Helper method for getRunningFlows */
+  protected List<Integer> getRunningFlowsHelper(final int projectId, final String flowId,
```
ypadron-in, line 367: Thanks for working on this refactoring!
Can this method be moved to `ExecutionControllerUtils` or `ExecutorUtils` class as an static method? Also, is it possible to merge those 2 classes?

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -66,22 +59,11 @@
  */
 @Singleton
 @Deprecated
-public class ExecutorManager extends EventHandler implements
-    ExecutorManagerAdapter {
+public class ExecutorManager extends AbstractExecutorManagerAdapter {
```
ypadron-in, line 62: If the PUSH dispatch model will still be supported maybe we should remove the @Deprecated annotation on top.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
jakhani: Sure. I have moved getRunningFlowsHelper  to ExecutorUtils. ExecutorUtils is for helper methods for executor related code. But ExecutionControllerUtils is related to executions. It is better to keep it separate.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
jakhani: It is good to keep it deprecated as it has many limitations. Preferable solution for now is POLL and in future CONTAINERIZED will be added.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -0,0 +1,436 @@
...
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractExecutorManagerAdapter extends EventHandler implements
```
ypadron-in, line 28: One alternative way of encapsulating this code which doesn't introduce inheritance could be declaring an intermediate interface with package access only which extends the `ExecutorManagerAdapter` interface and it is implemented by the `ExecutionController` and `ExecutorManager` classes. Something like:

```
interface IntermediateExecutorManagerAdapter extends ExecutorManagerAdapter {
  ExecutorLoader getExecutorLoader();
  
  default ExecutableFlow getExecutableFlow(final int execId)
    return getExecutorLoader().fetchExecutableFlow(execId);
  }
}
```

```
public class ExecutionController extends EventHandler implements IntermediateExecutorManagerAdapter {
@Override
  public ExecutorLoader getExecutorLoader() {
    return this.executorLoader;
  }
}
```

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
ypadron-in: The 3 methods inside `ExecutorUtils` are used both in `ExecutionController` and `ExecutorManager` classes therefore I thought about merging the 2 utility classes.

ypadron-in: Aren't `ExecutionControllerUtils` and `ExecutorUtils` just utility classes for `ExecutionController` and `ExecutorManager` respectively?

ypadron-in (approved):
LGTM

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
jakhani: Thank you. Let me think about this change and cover it in next PR. I want to keep this PR simple to reduce the amount of effect it could cause.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
jakhani: Not really. ExecutorUtils is for utilities related to executors so it is used by both ExecutionController and ExecutorManager.
ExecutionControllerUtils is more like utilities for executions and it is used by FlowRunner in executor.

PR #2626, opened 2020-11-02
Fixed the PDT-->PST time change tests. They should pass until atleast 2029

djaiswal83:


li-gsalia (approved):
Thanks for fixing this, change looks good!

PR #2627, opened 2020-11-05
hadoop clients are not closed properly in HadoopSecurityManager_H_2_0

haibchen:


jakhani (approved):
Thank you for taking care of closing these connections.

PR #2630, opened 2020-11-06
Expand the length of some fields related to Ramp Feature

kxu1026:
Expand the length of some fields related to Ramp Feature.
One of our client reported an oversize problem when using the ramp feature. Let's expand some field sizes to meet the requirement.


azkaban-db/src/main/sql/create.ramp_dependency.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE ramp_dependency (
-    dependency VARCHAR(45) NOT NULL,
-    defaultValue VARCHAR (500),
+    dependency VARCHAR(200) NOT NULL,
+    defaultValue VARCHAR (1000),
```
ypadron-in, line 3: Just curious, is there a retention policy setup at DB or app level for these tables?

azkaban-db/src/main/sql/upgrade.3.86.0.to.3.87.0.sql
```diff
@@ -0,0 +1,6 @@
+-- DB Migration from release 3.86.0 to 3.87.0
+-- Expand the length of fields related to ramp feature
+ALTER TABLE ramp_dependency CHANGE COLUMN dependency dependency VARCHAR(200) NOT NULL , CHANGE COLUMN defaultValue defaultValue VARCHAR(1000) NULL DEFAULT NULL;
+ALTER TABLE ramp_items CHANGE COLUMN dependency dependency VARCHAR(200) NOT NULL , CHANGE COLUMN rampValue rampValue VARCHAR(1000) NULL DEFAULT NULL;
```
ypadron-in, line 4: Here `rampValue` column is set to NULL, however in the create sql file is NOT NULL. Can we make constraints the same in both places?

azkaban-db/src/main/sql/upgrade.3.86.0.to.3.87.0.sql
```diff
@@ -0,0 +1,6 @@
+-- DB Migration from release 3.86.0 to 3.87.0
```
ypadron-in, line 1: Current release is 3.104.3. Could you please rename the file to `upgrade.3.104.0.to.3.105.0.sql` and modify this line as well.

azkaban-db/src/main/sql/create.ramp_dependency.sql
kxu1026: what do you mean? The constraint check is on DB only. We don't setup the limitation upon the app level.

azkaban-db/src/main/sql/upgrade.3.86.0.to.3.87.0.sql
kxu1026: thanks. good catch.

azkaban-db/src/main/sql/create.ramp_dependency.sql
ypadron-in: Are the records of these tables deleted after some time, for example after 6 months?

ypadron-in: yes. From @kxu1026:

> data will be removed after even ramping cycle.
> so, no accumulated items will remain even after couple of weeks.

ypadron-in (approved):
LGTM

PR #2632, opened 2020-11-07
Bring back "Flow override properties to also override job properties"

juhoautio:
This feature was reverted by @djaiswal83 in https://github.com/azkaban/azkaban/pull/2620.

Now adding it back as [requested](https://github.com/azkaban/azkaban/pull/2241#issuecomment-713893621):

> Can you please make it configurable and disabled by default?

See the PR description of https://github.com/azkaban/azkaban/pull/2241 for the background/rationale of this change.

Content of this PR:

- First, revert commit 21d33f0812c22df0883a3e56fc508ddd5abdc621 ie. PR https://github.com/azkaban/azkaban/pull/2620 to add back the code changes of https://github.com/azkaban/azkaban/pull/2241
- Then, make it configurable with a new property `executor.props.resolve.overrideExisting.enabled`

So, by default there is no change in behavior. That is to **not** override existing job props.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
+++ b/azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
@@ -161,11 +161,12 @@ public final class HttpRequestUtilsTest {
   @Test
   public void testGetJsonBodyForSingleMapObject() throws IOException, ServletException {
     final HttpServletRequest httpRequest = Mockito.mock(HttpServletRequest.class);
-    final String originalString = "  {\n" + "    \"action\": \"update\",\n" + "    \"table\": \"ramp\",\n"
-        + "    \"conditions\" : {\n" + "      \"rampId\" : \"dali\"\n" + "    },\n"
-        + "    \"values\": {\n"
-        + "      \"rampStage\": 2,\n" + "      \"lastUpdatedTime\": 1566259437000\n" + "    }\n"
-        + "  }\n";
+    final String originalString =
+        "  {\n" + "    \"action\": \"update\",\n" + "    \"table\": \"ramp\",\n"
+            + "    \"conditions\" : {\n" + "      \"rampId\" : \"dali\"\n" + "    },\n"
+            + "    \"values\": {\n"
+            + "      \"rampStage\": 2,\n" + "      \"lastUpdatedTime\": 1566259437000\n" + "    }\n"
+            + "  }\n";
     final InputStream inputStream = new ByteArrayInputStream(originalString.getBytes(UTF_8));
     final InputStreamReader inputStreamReader = new InputStreamReader(inputStream, UTF_8);
     final BufferedReader bufferedReader = new BufferedReader(inputStreamReader);
@@ -185,13 +186,13 @@ public final class HttpRequestUtilsTest {
 
   @Test
   public void testParseFlowOptionsSla() throws Exception {
-    HttpServletRequest req = mockRequestWithSla(ImmutableMap.of(
+    final HttpServletRequest req = mockRequestWithSla(ImmutableMap.of(
         // job_name, status, duration, is_email, is_kill
         "slaSettings[1]", ",FINISH,2:30,true,false",
         "slaSettings[2]", "test_job,SUCCESS,12:00,false,true",
         "slaSettings[3]", ",SUCCESS,12:00,true,true"));
-    ExecutionOptions options = HttpRequestUtils.parseFlowOptions(req, "test-flow");
-    List<SlaOption> slaOptions = options.getSlaOptions();
+    final ExecutionOptions options = HttpRequestUtils.parseFlowOptions(req, "test-flow");
+    final List<SlaOption> slaOptions = options.getSlaOptions();
     final List<SlaOption> expected = Arrays.asList(
         new SlaOption(SlaType.FLOW_FINISH, "test-flow", "", Duration.ofMinutes(150),
             ImmutableSet.of(SlaAction.ALERT), ImmutableList.of()),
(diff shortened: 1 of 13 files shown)


juhoautio:
@djaiswal83 does this match with what you had in mind?

djaiswal83:
@juhoautio Please allow us to go through the PR this week.

azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
```diff
@@ -161,11 +161,12 @@ public void testGetJsonBodyForListOfMapObject() throws IOException, ServletExcep
   @Test
   public void testGetJsonBodyForSingleMapObject() throws IOException, ServletException {
     final HttpServletRequest httpRequest = Mockito.mock(HttpServletRequest.class);
-    final String originalString = "  {\n" + "    \"action\": \"update\",\n" + "    \"table\": \"ramp\",\n"
-        + "    \"conditions\" : {\n" + "      \"rampId\" : \"dali\"\n" + "    },\n"
-        + "    \"values\": {\n"
-        + "      \"rampStage\": 2,\n" + "      \"lastUpdatedTime\": 1566259437000\n" + "    }\n"
-        + "  }\n";
+    final String originalString =
```
aditya1105, line 164: I would suggest putting this in the resources directory with the same package as of this test class. Then you can utilize the method
azkaban.utils.TestUtils#readResource to read it as string.

aditya1105:
@juhoautio , I have added one minor suggestion. Apart from that, PR looks good to me. Thanks for making this change. I am approving it.

aditya1105 (approved):

azkaban-common/src/test/java/azkaban/server/HttpRequestUtilsTest.java
juhoautio: > I would suggest putting this in the resources directory

Ok, I'll do that. But as you can see this wasn't added as part of this PR. Formatting changed by Save Actions.

juhoautio:
@aditya1105 thanks! I have moved the json strings to resource files.

juhoautio:
@aditya1105 @djaiswal83 how about this, did you have any other remarks? I can resolve the conflicts.

djaiswal83 (dismissed):
Thanks for making this change. Please resolve the conflict and I will re-approve and merge.

juhoautio:
@djaiswal83 rebased & resolved conflicts.

djaiswal83 (approved):
Thanks for adding the feature flag. The PR now looks good.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -879,12 +893,24 @@ private void prepareJobProperties(final ExecutableNode node) throws IOException
...
+        ConfigurationKeys.EXECUTOR_PROPS_RESOLVE_OVERRIDE_EXISTING_ENABLED, false)) {
+      // Flow override props are configured to also override existing job props
+      // =>
+      // 5. If there are any runtime flow overrides, we apply them now.
+      final Map<String, String> flowParam =
+          this.flow.getExecutionOptions().getFlowParameters();
+      if (flowParam != null && !flowParam.isEmpty()) {
+        props.putAll(flowParam);
+      }
+    }
```
juhoautio, line 906: @ypadron-in wrote in https://github.com/azkaban/azkaban/pull/2597#discussion_r566636794:

> With this block of code I think we're mixing the ability to overwrite properties with the concept of flow and job level properties.
>
> -By providing runtime properties we are overwriting the value of already defined properties (or adding new ones).
-That should NOT interfere with the notion of flow and job level properties. I believe flow level properties are meant to define properties to be applied to the entire DAG and as a side effect offer a convenient way of defining job level properties in one place that will be applied to all jobs. However, job level properties take precedence over the same properties defined at flow level.
>
> I don't think it's ok to add this piece of code because we're saying that flow level overwrites take precedence over the existing hierarchy of flow/job level properties. Flow level overwrites should only take precedence over flow level definition properties.
>
> With the "job specific flow overwrites" feature added by ~~this~~ PR (https://github.com/azkaban/azkaban/pull/2597) we're adding the missing piece in the puzzle which is the ability to overwrite job level definition properties.

and

> I think we should revert ~~that~~ this PR. Do you understand the issue there?

I understand that what you have described is the old/existing behavior, and it also remains the default. As you can see, this behavior is controlled by azkaban server property, and is disabled by default.

But the ability to override also existing job properties globally is the whole point of this PR. It's just very handy in practice to be able to override globally instead of listing all jobs (or nodes) that need some override separately.

I have actually failed to see what the real-life use cases are for overriding "flow level props", but still not overriding if that property is present in some jobs. I understand that the default behavior cannot be changed without ensuring that users understand it though.

To continue on the use case, to me it would be more natural to not set a flow override but just set job-specific (or node-specific) overrides where needed and not set a global (flow) override at all 🤷‍♂️ But of course either way can work, more or less.

azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java
```diff
@@ -107,15 +110,15 @@ private void assertProperties(final boolean isAzkabanFlowVersion20) throws Excep
...
+    Assert.assertEquals("flow7", job2Props.get("props7"));
     Assert.assertEquals("flow5", job2Props.get("props5"));
     Assert.assertEquals("flow6", job2Props.get("props6"));
     Assert.assertEquals("shared4", job2Props.get("props4"));
     Assert.assertEquals("shared8", job2Props.get("props8"));
 
     // Job 1 is inside another flow, and is nested in a different directory
     // The priority order should be:
-    // job1->innerflow->job2.output->flow.overrides->job1 shared props
+    // job1->innerflow->job2.output->job1.sharedprops->flow.overrides
```
ypadron-in, line 121: @juhoautio with `EXECUTOR_PROPS_RESOLVE_OVERRIDE_EXISTING_ENABLED` set to true I believe this comment  (and similar below) is not correct, flow overrides now will win over everything so it should be: 
`flow.overrides->job1->innerflow->job2.output->job1.sharedprops`

PR #2641, opened 2020-11-17
Add metrics for jobtype executions

li-pkhandelwal:
Added following metrics for jobtype:
*JobTypeName*-job-success-meter: Records when jobtype executions are marked as SUCCEEDED.
*JobTypeName*-job-fail-meter: Records when jobtype execution are marked as FAILED.
*JobTypeName*-job-killed-meter: Records when jobtype executions are marked as KILLED.


djaiswal83 (changes requested):
Can you please add unit tests?

PR #2652, opened 2020-11-19
BDP-3878: Magnet on flow ramping regardless failure

kxu1026:


kxu1026:
The requirement is ' If the workflow fails during the ramp, DONNOT stop ramping it and do not logging this failed flow into the exclude-ramp list.' I am leveraging the maxFailureToPause and maxFailureToRampDown configure to make it happen. When both maxFailureToPause and maxFailureToRampDown are set to negative value, the failed workflow will not be logged into the exclude-ramp list.

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -376,6 +376,13 @@ public boolean isNotTestable() {
     return (!this.state.isActive || this.state.isPaused || (this.state.rampStage <= 0));
   }
 
+  /**
+   * Return true iif this ramp will not be required to put those failed workflow into the exclude-ramp list
+   */
+  public boolean isMagneticTest() {
```
zhouyejoe, line 382: Function name should be generic, maybe "ignoreFlowFailure".

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -376,6 +376,13 @@ public boolean isNotTestable() {
     return (!this.state.isActive || this.state.isPaused || (this.state.rampStage <= 0));
   }
 
+  /**
+   * Return true iif this ramp will not be required to put those failed workflow into the exclude-ramp list
```
jakhani, line 380: nit: if

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
jakhani, line 382: Shall this method name be changed? Magnetic is very specific to Spark right?

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
kxu1026: iif = if and only if. right?

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
kxu1026: what is your suggestion?

kxu1026: changed to ignoreTestFailure

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
kxu1026: changed to if and only if

azkaban-common/src/main/java/azkaban/executor/ExecutableRamp.java
```diff
@@ -376,6 +376,13 @@ public boolean isNotTestable() {
...
+   */
+  public boolean ignoreTestFailure() {
+    return (this.metadata.maxFailureToPause < 0) && (this.metadata.maxFailureToRampDown < 0);
```
jakhani, line 383: <= ?

jakhani (dismissed):

djaiswal83 (changes requested):
A minor typo. Rest looks good.

djaiswal83 (approved):

PR #2653, opened 2020-11-20
Introducing version set and added skeleton for version set loader

aditya1105:
Introducing version set and added skeleton for version set loader

LocalVersionSetLoader.java is added as of now for functionality testing. Later Actual JDBC-based implementation will be added.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
+++ b/azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.version;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+@VisibleForTesting
+public class LocalVersionSetLoader implements VersionSetLoader {
+    private Map<String, Integer> versionSetMapping = new HashMap<>();
+    private int max = 0;
+
+    @Override
+    public int insertAndGetVersionSetNum(String versionSetJsonString, String versionSetMd5Hex) throws IOException {
+        this.versionSetMapping.put(versionSetMd5Hex, ++max);
+        return this.versionSetMapping.get(versionSetMd5Hex);
+    }
+
+    @Override
+    public boolean removeVersionSet(String versionSetMd5Hex) throws IOException {
+        if (!this.versionSetMapping.containsKey(versionSetMd5Hex)) {
+            // default to true
+            return true;
+        }
+        return this.versionSetMapping.remove(versionSetMd5Hex, this.versionSetMapping.get(versionSetMd5Hex));
+    }
+
+    @Override
+    public int getVersionSetNum(String versionSetJsonString, String versionSetMd5Hex) throws IOException {
+        if (this.versionSetMapping.containsKey(versionSetMd5Hex)) {
+            return this.versionSetMapping.get(versionSetMd5Hex);
+        }
+        return this.insertAndGetVersionSetNum(versionSetJsonString, versionSetMd5Hex);
+    }
+}
--- a/azkaban-common/src/main/java/azkaban/version/VersionSet.java
+++ b/azkaban-common/src/main/java/azkaban/version/VersionSet.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.version;
+
+/**
+ * VersionSet is defined as the set of key-val pairs to uniquely define the execution environment.
+ * This version set will have azkaban version, azkaban configuration version, job type versions, etc.
+ *
+ * Example version set json:
+ * {"azConf":"0.0.5","azkaban":"0.0.2","sparkJobType":"0.1.2"}
+ */
+public class VersionSet {
+    private final String versionSetJsonString;
+    private final String versionSetMd5Hex;
+    private final int versionSetNum;
+
+    public VersionSet(String versionSetJsonString, String versionSetMd5Hex, int versionSetNum) {
+        this.versionSetJsonString = versionSetJsonString;
+        this.versionSetMd5Hex = versionSetMd5Hex;
+        this.versionSetNum = versionSetNum;
+    }
+
+    public String getVersionSetJsonString() {
+        return this.versionSetJsonString;
+    }
+
+    public String getVersionSetMd5Hex() {
+        return this.versionSetMd5Hex;
+    }
+
+    public int getVersionSetNum() {
+        return this.versionSetNum;
+    }
+}
--- a/azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
+++ b/azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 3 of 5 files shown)


azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+@VisibleForTesting
+public class LocalVersionSetLoader implements VersionSetLoader {
+    private Map<String, Integer> versionSetMapping = new HashMap<>();
+    private int max = 0;
```
abhisheknath2011, line 27: static final?

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public int insertVersionSet(String versionSetJsonString, String versionSetMd5Hex) throws IOException {
```
abhisheknath2011, line 30: As this method is internally invoked I think we can make this private.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+    public boolean removeVersionSet(String versionSetMd5Hex) throws IOException {
+        if (!this.versionSetMapping.containsKey(versionSetMd5Hex)) {
+            // default to true
+            return true;
+        }
+        return this.versionSetMapping.remove(versionSetMd5Hex, this.versionSetMapping.get(versionSetMd5Hex));
+    }
+
+    @Override
+    public int getVersionSetNum(String versionSetJsonString, String versionSetMd5Hex) throws IOException {
```
abhisheknath2011, line 45: Can we rename this method to insertAndGetVersionSetNum as this method returns versionSetNum if exists else inserts and returns?

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+        return this;
+    }
+
+    public VersionSetBuilder addElements(Map<String, String> keyVals) {
+        this.versionSetElements.putAll(keyVals);
+        return this;
+    }
+
+    public VersionSet build() throws IOException {
+        ObjectMapper objectMapper = new ObjectMapper();
```
abhisheknath2011, line 49: It would be good if we can pass ObjectMapper as objectMapper instance is already created and available as part of AzkabanWebServer. Other option could be through Injection. Else whenever builder is used ObjectMapper instance will be created. Might want to check this.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+
+package azkaban.version;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+@VisibleForTesting
+public class LocalVersionSetLoader implements VersionSetLoader {
```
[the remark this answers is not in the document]
aditya1105: This class is only meant for functionality testing.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+    private Map<String, Integer> versionSetMapping = new HashMap<>();
```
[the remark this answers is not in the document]
aditya1105: This class is only meant for functionality testing.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
aditya1105: This class is only meant for functionality testing.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
aditya1105: This class is only meant for functionality testing.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
aditya1105: This class is only meant for functionality testing. However, you are right and this can be renamed in the interface.

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
aditya1105: ObjectMapper is made threadsafe and hence it ends up blocking. Passing a single instance everywhere may not be recommended.
Since the object is method local, it will be cleaned

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This builder class is responsible for creating key-ordered json string for the version set,
+ * to ensure that the same execution environment will have the same string and hence the
+ * md5 digest will remain the same.
+ */
+public class VersionSetBuilder {
+    private final Map<String, String> versionSetElements = new TreeMap();
+    private final VersionSetLoader loader;
+
+    public VersionSetBuilder(VersionSetLoader loader) {
```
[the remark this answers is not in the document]
aditya1105: As you mentioned, when VersionSetLoader implementation will be added, it will be made singleton. Hence caller class will pass the same instance to this constructor.

abhisheknath2011 (approved):

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
jakhani, line 25: Can you please add documentation for the classes and public methods?

azkaban-common/src/main/java/azkaban/version/VersionSet.java
```diff
@@ -0,0 +1,48 @@
...
+ * VersionSet is defined as the set of key-val pairs to uniquely define the execution environment.
+ * This version set will have azkaban version, azkaban configuration version, job type versions, etc.
+ *
+ * Example version set json:
+ * {"azConf":"0.0.5","azkaban":"0.0.2","sparkJobType":"0.1.2"}
+ */
+public class VersionSet {
+    private final String versionSetJsonString;
+    private final String versionSetMd5Hex;
+    private final int versionSetNum;
```
jakhani, line 29: Is this id, primary key in a table?

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
```diff
@@ -0,0 +1,51 @@
...
+        return this.versionSetMapping.get(versionSetMd5Hex);
+    }
+
+    @Override
+    public boolean removeVersionSet(String versionSetMd5Hex) throws IOException {
+        if (!this.versionSetMapping.containsKey(versionSetMd5Hex)) {
+            // default to true
+            return true;
+        }
+        return this.versionSetMapping.remove(versionSetMd5Hex, this.versionSetMapping.get(versionSetMd5Hex));
```
jakhani, line 41: Shall it be concurrent hashmap?

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
jakhani, line 34: Loader should be injected right?

azkaban-common/src/main/java/azkaban/version/VersionSet.java
aditya1105: versionSetNum will be an autoincrement field.

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
aditya1105: This is a class only meant for functional testing. So kept simple

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
aditya1105: This is a class only meant for functional testing.

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
aditya1105: Since VersionSetLoader will be created multiple times across and is not even a singleton. The loader will be a singleton and it will be created using guice. However, since the loader will be a singleton, it can be passed across to VersionSetBuilder from anywhere.

azkaban-common/src/main/java/azkaban/version/VersionSet.java
jakhani: Yup so it's name should be id right?

azkaban-common/src/main/java/azkaban/version/LocalVersionSetLoader.java
jakhani: Then please add the detail that it is for functional testing?

azkaban-common/src/main/java/azkaban/version/VersionSetBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+        this.loader = loader;
+    }
+
+    public VersionSetBuilder addElement(String key, String version) {
```
arvindpruthi, line 38: I feel the name "addElement" is too generic. I think a better name is: 
    addImageVersion(String imageType, String version)

azkaban-common/src/main/java/azkaban/version/VersionSet.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public VersionSet(String versionSetJsonString, String versionSetMd5Hex, int versionSetNum) {
```
arvindpruthi, line 31: No matter how obvious. We should definitely have a proper javadoc style header for at least every public method. 

Each private method must also be adequately documented.

PR #2651, opened 2020-11-19
Image Management: Implementation for create image type, image version and get image version REST APIs

abhisheknath2011:
Included changes
1. Database changes for containerization and image management - https://jira01.corp.linkedin.com:8443/browse/BDP-3108
2. Implementation of create image type REST API - https://jira01.corp.linkedin.com:8443/browse/BDP-3655
3. Implementation of create image version REST API - https://jira01.corp.linkedin.com:8443/browse/BDP-3654
4. Implementation of get image version REST API - https://jira01.corp.linkedin.com:8443/browse/BDP-3712
5. Incorporated review comments on initial PR - https://github.com/azkaban/azkaban/pull/2637

Initial first cut pull request : https://github.com/azkaban/azkaban/pull/2637
Initial first cut commit: https://github.com/azkaban/azkaban/pull/2637/commits/aa370b7d5a980cc64c1838b34d10ec27639daadf

Image management REST API design doc: https://docs.google.com/document/d/1zbA8tZ7jjT3owWNPFLyhtU9A-8RKxRU6Rkqq1LA1G6o/edit#

PR #2655, opened 2020-11-25
Added version_set table DDL and JdbcVersionSetLoaderImpl class for interaction

aditya1105:
Added sql/create.version_set.sql for DDL
Added JdbcVersionSetLoaderImpl for interaction with version_set table
Removed LocalVersionSetLoader as it was only temporary until JdbcVersionSetLoaderImpl is added
Added test JdbcVersionSetLoaderImplTest


azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+                throw new SQLException("Expected only one VersionSet for newly inserted versionSetMd5Hex " + versionSetMd5Hex);
+            }
+            // Insert if it doesn't
+            if (versionSets.isEmpty()) {
+                transOperator.update(INSERT_VSET, versionSetMd5Hex, versionSetJsonString);
+            }
+
+            // Query for versionSet again after insert
+            versionSets = transOperator.query(SELECT_VSET_FROM_MD5, versionSetHandler, versionSetMd5Hex);
+            if (versionSets.size() > 1) {
```
jakhani, line 73: You should throw exception when versionSets.size() != 1.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+                throw new SQLException("Expected only one VersionSet for newly inserted versionSetMd5Hex " + versionSetMd5Hex);
+            }
+            if (versionSets.isEmpty()) {
```
jakhani, line 76: Empty should not be a valid case. exception should be thrown like I mentioned in above comment. Can you please follow https://github.com/azkaban/azkaban/blob/1e8c84c13d20af27bcd47da5d42040d1ec996a00/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java#L82 and write insert related code using transOperator.update method? You will get id in return and won't need to get version set again. It will save a query.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+            logger.error("Exception occurred while inserting version set and getting version number", e);
+            throw new IOException("Unable to insert and get versionSetNum", e);
+        }
+    }
+
+    /**
+     * Removes the versionSet row from the table version_set corresponding to versionSetMd5Hex.
+     */
+    @Override
+    public synchronized boolean removeVersionSet(String versionSetMd5Hex) throws IOException {
```
jakhani, line 99: version set can be deleted by id as well once it become invalid. Can you please rename this method to deleteVersionSet?

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+        final SQLTransaction<Integer> insertVersionSet = transOperator -> transOperator.update(DELETE_VSET, versionSetMd5Hex);
+        try {
+            Integer rowsDeleted = this.dbOperator.transaction(insertVersionSet);
+            if (this.md5ToVersionSet.containsKey(versionSetMd5Hex)) {
+                VersionSet versionSet = this.md5ToVersionSet.get(versionSetMd5Hex);
+                this.md5ToVersionSet.remove(versionSetMd5Hex);
```
jakhani, line 105: You can directory remove the key. No need to get. Remove returns a value that will be removed. You can use it as a reference to remove it from numToVersionSet map.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+            return rowsDeleted != 0;
+        } catch (SQLException e) {
+            logger.error("Failed to execute: " + DELETE_VSET, e);
+            throw new IOException("Exception occurred while removing version set for " + versionSetMd5Hex, e);
+        }
+    }
+
+    /**
+     * This method first checks if the versionSetNum already exists in the local copy of the versionSets,
+     * otherwise, it insert the versionSet into the table and get the resulting versionSetNum.
```
jakhani, line 117: Nit: it will insert

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    @Override
+    public synchronized int getVersionSetNum(String versionSetJsonString, String versionSetMd5Hex) throws IOException {
+        if (this.md5ToVersionSet.containsKey(versionSetMd5Hex)) {
+            VersionSet versionSet = this.md5ToVersionSet.get(versionSetMd5Hex);
+            return versionSet.getVersionSetNum();
+        }
+        return insertAndGetVersionSetNum(versionSetJsonString, versionSetMd5Hex);
```
jakhani, line 125: Nit: Can you please change the order of parameters? (versionSetMd5Hex, versionSetJsonString)

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    @Override
+    public synchronized VersionSet getVersionSet(int versionSetNum) throws IOException {
+        if (this.numToVersionSet.containsKey(versionSetNum)) {
+            return this.numToVersionSet.get(versionSetNum);
+        }
+        final VersionSetHandler versionSetHandler = new VersionSetHandler();
+        try {
+            List<VersionSet> versionSets = this.dbOperator.query(SELECT_VSET_FROM_NUM, versionSetHandler, versionSetNum);
+            if (versionSets.size() > 1) {
```
jakhani, line 169: There is a common code in some get methods. Can you please extract common code in another method and use it everywhere?

azkaban-common/src/main/java/azkaban/version/VersionSetLoader.java
```diff
@@ -17,21 +17,37 @@
...
+
+    /**
+     * @return {@link VersionSet} corresponding to versionSetMd5Hex.
+     */
+    VersionSet getVersionSet(String versionSetMd5Hex) throws IOException;
+
+    /**
+     * @return {@link VersionSet} corresponding to versionSetNum.
+     */
+    VersionSet getVersionSet(int versionSetNum) throws IOException;
```
jakhani, line 47: Can you please rename this method to getVersionSetById?

aditya1105: Done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Common code is there with the exchange of num and Id, along with log messages, etc. Hence it will be better to keep it segregated

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Done.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
aditya1105: Removed this block as it won't be required after using getLastInsertId

azkaban-db/src/main/sql/create.version_set.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE version_set
+  (
+     num  INT NOT NULL auto_increment,
+     md5  CHAR(32) NOT NULL,
```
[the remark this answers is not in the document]
aditya1105: Since the Md5 is always fixed to 32 bits, it is better to use char. Varchar is useful when fields are of varying size which varchar optimizes by using one extra bit.

azkaban-db/src/main/sql/create.version_set.sql
```diff
@@ -0,0 +1,9 @@
...
+     json VARCHAR(4096) NOT NULL,
+     PRIMARY KEY (num)
+  );
+
+CREATE UNIQUE INDEX idx_md5 ON version_set (md5);
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-db/src/main/sql/create.version_set.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE version_set
+  (
+     num  INT NOT NULL auto_increment,
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+ */
+@Singleton
+public class JdbcVersionSetLoaderImpl implements VersionSetLoader {
+    private static final Logger logger = Logger.getLogger(JdbcVersionSetLoaderImpl.class);
+
+    private static final String SELECT_VSET_FROM_NUM = "SELECT vset.num, vset.md5, vset.json FROM version_set vset WHERE vset.num=?";
+    private static final String SELECT_VSET_FROM_MD5 = "SELECT vset.num, vset.md5, vset.json FROM version_set vset WHERE vset.md5=?";
+    private static final String SELECT_ALL_VSET = "SELECT vset.num, vset.md5, vset.json FROM version_set vset";
+    private static final String INSERT_VSET = "INSERT INTO version_set (md5, json) values (?,?)";
+    private static final String DELETE_VSET = "DELETE FROM version_set vset WHERE vset.md5=?";
```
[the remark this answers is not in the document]
aditya1105: version_set has no concept of active/deprecated. 
It is only stored in the db to identify if the exact combination was used earlier as well and hence to determine the id allocated for that. 

Deletion functionality may never be utilized by added so that in future if the table requires clean up, then this can be leveraged.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+ * This class keeps the copy of entries in the version_set table locally
+ * as well to reduce the number of queries to the db.
+ *
+ * This is a singleton class to be instantiated using dependency injection.
+ */
+@Singleton
+public class JdbcVersionSetLoaderImpl implements VersionSetLoader {
+    private static final Logger logger = Logger.getLogger(JdbcVersionSetLoaderImpl.class);
+
+    private static final String SELECT_VSET_FROM_NUM = "SELECT vset.num, vset.md5, vset.json FROM version_set vset WHERE vset.num=?";
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * A JDBC based implementation for loading VersionSet from the backed db.
+ * This class keeps the copy of entries in the version_set table locally
+ * as well to reduce the number of queries to the db.
+ *
+ * This is a singleton class to be instantiated using dependency injection.
+ */
+@Singleton
+public class JdbcVersionSetLoaderImpl implements VersionSetLoader {
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.commons.dbutils.ResultSetHandler;
+import org.apache.log4j.Logger;
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
+package azkaban.version;
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-db/src/main/sql/create.version_set.sql
```diff
@@ -0,0 +1,9 @@
...
+     md5  CHAR(32) NOT NULL,
+     json VARCHAR(4096) NOT NULL,
```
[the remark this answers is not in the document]
aditya1105: Version set table is only used to identify if the same combination of versions were used earlier as well, and if yes, what was the assigned id to it. Hence, the only use case for this table to determine the id based on md5, and along with id, we may need what does id describes? (json string).

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
```diff
@@ -0,0 +1,217 @@
...
+    private static final Logger logger = Logger.getLogger(JdbcVersionSetLoaderImpl.class);
+
+    private static final String SELECT_VSET_FROM_NUM = "SELECT vset.num, vset.md5, vset.json FROM version_set vset WHERE vset.num=?";
+    private static final String SELECT_VSET_FROM_MD5 = "SELECT vset.num, vset.md5, vset.json FROM version_set vset WHERE vset.md5=?";
+    private static final String SELECT_ALL_VSET = "SELECT vset.num, vset.md5, vset.json FROM version_set vset";
+    private static final String INSERT_VSET = "INSERT INTO version_set (md5, json) values (?,?)";
+    private static final String DELETE_VSET = "DELETE FROM version_set vset WHERE vset.md5=?";
+
+    private final DatabaseOperator dbOperator;
+    private final Map<String, VersionSet> md5ToVersionSet = new HashMap<>();
```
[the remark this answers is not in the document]
aditya1105: If the user provides versionSetId/versionSetNum as part of flow parameters and we want to determine the version set corresponding to it, then we need this mapping (otherwise we may have to iterate over all values and find it). Also, the versionSet object itself is common for both the maps.

Even with LRU, we may have to end up either using two caches or maintaining one cache and one map and keeping them in sync. Also, it may be overkill in the beginning. I think we should see the size of the version set table and then update the code to the LRU cache accordingly.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A JDBC based implementation for loading VersionSet from the backed db. This class keeps the copy
+ * of entries in the version_set table locally as well to reduce the number of queries to the db.
+ * <p>
+ * This is a singleton class to be instantiated using dependency injection.
+ */
+@Singleton
+public class JdbcVersionSetLoader implements VersionSetLoader {
```
abhisheknath2011, line 41: General comment: Should we consider adding version set implementation as part of imagemgmt package as version set implementation is part of image management.

azkaban-db/src/main/sql/create.version_set.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE version_set
```
abhisheknath2011, line 1: image_version_set table already added as part of the PR: https://github.com/azkaban/azkaban/pull/2651

aditya1105: That is not in the master branch. Second, I think the schema we anyway have to change as created_on and created_by is not required with this definition of the version set.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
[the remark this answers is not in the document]
jakhani: What if there is any image version which is faulty. There is version set already populated with that version. User is using version set number in flow parameter , how will we stop them from using faulty version of any image?

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
jakhani: +1

azkaban-db/src/main/sql/create.version_set.sql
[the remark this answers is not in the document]
jakhani: Sometimes, it will help us to find bugs based on when version set is created. In case of faulty image version or something, we can filter by date and all. It will be useful to keep.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
abhisheknath2011, line 41: This class can be renamed to something like VersionSetLoaderImpl or VersionSetDaoImpl. So it would make sense to move this peace of code to cf-engine-k8s branch so that it can be place under imagemgmt package and similar structure and format can be followed.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(JdbcVersionSetLoader.class);
+
+  private static final String SELECT_VSET_FROM_ID = "SELECT vset.id, vset.md5, vset.json FROM "
```
abhisheknath2011, line 45: As we are using single table and there is no join involved so there is not possibility of column name conflicts. Hence, usage of alias can be avoided here. Same for all the other queries.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+  @Override
+  public synchronized int insertAndGetVersionSetId(String versionSetMd5Hex,
+      String versionSetJsonString) throws IOException {
+    final SQLTransaction<Integer> insertVersionSet = transOperator -> {
+      final VersionSetHandler versionSetHandler = new VersionSetHandler();
+      // Check if the versionSet already exists for versionSetMd5Hex
+      List<VersionSet> versionSets = transOperator
+          .query(SELECT_VSET_FROM_MD5, versionSetHandler, versionSetMd5Hex);
+      if (versionSets.size() > 1) {
+        throw new SQLException(
```
abhisheknath2011, line 86: Application specific custom exception can be thrown here. From the old code ExecutorManagerException is thrown from DAO. In this case as this will be part of imagemgmt package, ImageMgmtDaoException can be thrown here.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+    };
+    try {
+      Integer versionSetId = this.dbOperator.transaction(insertVersionSet);
+      VersionSet versionSet = new VersionSet(versionSetJsonString, versionSetMd5Hex, versionSetId);
+      this.md5ToVersionSet.put(versionSetMd5Hex, versionSet);
+      this.idToVersionSet.put(versionSetId, versionSet);
+      return versionSetId;
+    } catch (SQLException e) {
+      logger.error("Exception occurred while inserting version set and getting version id", e);
+      throw new IOException("Unable to insert and get versionSetId", e);
```
abhisheknath2011, line 104: Why IOException? Again dao specific exception can be thrown here.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+    try {
+      Integer rowsDeleted = this.dbOperator.transaction(insertVersionSet);
+      if (this.md5ToVersionSet.containsKey(versionSetMd5Hex)) {
+        VersionSet removedVersionSet = this.md5ToVersionSet.remove(versionSetMd5Hex);
+        this.idToVersionSet.remove(removedVersionSet.getVersionSetId());
+      }
+      return rowsDeleted != 0;
+    } catch (SQLException e) {
+      logger.error("Failed to execute: " + DELETE_VSET, e);
+      throw new IOException("Exception occurred while removing version set for " + versionSetMd5Hex,
```
abhisheknath2011, line 124: Same as above.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+  public synchronized VersionSet getVersionSet(String versionSetMd5Hex) throws IOException {
+    if (this.md5ToVersionSet.containsKey(versionSetMd5Hex)) {
+      return this.md5ToVersionSet.get(versionSetMd5Hex);
+    }
+    final VersionSetHandler versionSetHandler = new VersionSetHandler();
+    try {
+      List<VersionSet> versionSets = this.dbOperator
+          .query(SELECT_VSET_FROM_MD5, versionSetHandler, versionSetMd5Hex);
+      if (versionSets.size() > 1) {
+        throw new IOException(
```
abhisheknath2011, line 158: ImageMgmtDaoException

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * This constructor should be instantiated using dependency injection only to make sure that that
+   * class remains Singleton.
+   *
+   * @param dbOperator Interface to perform db related operations.
+   */
+  @Inject
+  public JdbcVersionSetLoader(DatabaseOperator dbOperator) throws IOException {
+    this.dbOperator = dbOperator;
+    for (VersionSet versionSet : fetchAllVersionSets()) {
```
abhisheknath2011, line 67: As this is a Singleton class, fetching all the version set will happen only once during instance creation. We can improve on the caching part by using guava cache suppliers. Suppliers provide memoiseWithExpiration method, using which we can trigger db callback at specified duration say 1 hr or so. This will ensure that cache is updated at specified interval.

abhisheknath2011: I think this comment can be ignored as same class is populating the cache as well. I believe from nowhere else db will be populated.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
abhisheknath2011, line 41: One Suggestion: I think the cache part can be improved. Currently the cache and the database access part is fully coupled together. Cache can be build on top of database as a wrapper and allow caller to invoke the cache methods directly. If the requested versionSet is not present in the cache access from database and populate the cache. In case of insert first insert into cache and then database. This should not cause any issue as direct db access is not allowed as it will be always via cache.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoaderImpl.java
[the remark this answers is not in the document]
aditya1105: Reusing the same version set is a functionality added only to provide the user with the same environment as of some previous execution if it is mentioned explicitly, and hence if the user wants to use the same version set, it should be allowed. If this is not explicitly provided, then anyway the versions will be calculated at runtime and the deprecated jobtype version won't be utilized.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: It does make sense to change the package to imagemgmt, however, since the code itself doesn't have any dependency with the code in cf-engine-k8s, it can be kept in the master branch itself.

aditya1105: will rename the package

aditya1105: Since it exetends VersionSetLoader, changing it to VersionSetDaoImpl would be odd. This name is derived from the pair FlowTriggerInstanceLoader and JdbcFlowTriggerInstanceLoaderImpl already existing in the code base.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: Since VersionSetLoader interface has the definition to throw IOException, it is kept along the same lines.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: Since VersionSetLoader interface has the definition to throw IOException, it is kept along the same lines.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: Since VersionSetLoader interface has the definition to throw IOException, it is kept along the same lines.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: Since VersionSetLoader interface has the definition to throw IOException, it is kept along the same lines.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: Yes, the same class will update the local maps as well as the db every time. Hence only once it needs to be loaded from the db.

azkaban-common/src/main/java/azkaban/version/JdbcVersionSetLoader.java
aditya1105: I agree on the decoupling part which will allow us to have any implementation of the cache in the future as well. However, I am not sure about the requirement of the cache itself. This is similar to schedule based trigger where we keep all the schedules in the memory and always update both the local map and the db through the same class. Since the code is only accessed by one class, we can refactor it easily in the future if the cache requirement arises.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+    try {
+      List<VersionSet> versionSets = this.dbOperator
+          .query(SELECT_VSET_FROM_MD5, versionSetHandler, versionSetMd5Hex);
+      if (versionSets.size() > 1) {
+        throw new IOException(
+            "Expected only one VersionSet for versionSetMd5Hex " + versionSetMd5Hex);
+      }
+      if (versionSets.isEmpty()) {
+        logger.info("No VersionSet exists for versionSetMd5Hex " + versionSetMd5Hex);
+        return null;
```
jakhani, line 163: Shall we use Optional here?

azkaban-db/src/main/sql/create.version_set.sql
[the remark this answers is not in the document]
jakhani: This is pending.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * This constructor should be instantiated using dependency injection only to make sure that that
+   * class remains Singleton.
+   *
+   * @param dbOperator Interface to perform db related operations.
+   */
+  @Inject
+  public JdbcVersionSetLoader(DatabaseOperator dbOperator) throws IOException {
+    this.dbOperator = dbOperator;
+    for (VersionSet versionSet : fetchAllVersionSets()) {
```
arvindpruthi, line 67: I would prefer to discourage the "pre-loading" pattern as in recent future we want to move to distributed web server model.

I understand that you are not fully depending on the cache and have a cache-miss processing in place. But make it clear either way that cache is not authoritative either by not doing pre-loading or by commenting very clearly that is the case. I do understand you think this pre-loading is desirable for performance reasons.

azkaban-common/src/main/java/azkaban/imagemgmt/version/VersionSet.java
```diff
@@ -14,35 +14,36 @@
...
- * This version set will have azkaban version, azkaban configuration version, job type versions, etc.
- *
- * Example version set json:
- * {"azConf":"0.0.5","azkaban":"0.0.2","sparkJobType":"0.1.2"}
+ * This version set will have azkaban version, azkaban configuration version, job type versions,
+ * etc.
+ * <p>
+ * Example version set json: {"azConf":"0.0.5","azkaban":"0.0.2","sparkJobType":"0.1.2"}
  */
 public class VersionSet {
```
arvindpruthi, line 26: It will be great to have a Map of imagetypes & versions as part of VersionSet. If we have to look up the version for a specific image, parsing json to pull that information is untidy.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A JDBC based implementation for loading VersionSet from the backed db. This class keeps the copy
+ * of entries in the version_set table locally as well to reduce the number of queries to the db.
+ * <p>
+ * This is a singleton class to be instantiated using dependency injection.
+ */
+@Singleton
+public class JdbcVersionSetLoader implements VersionSetLoader {
```
arvindpruthi, line 41: Should we separate out Dao from the VersionSetLoader? It's not strictly required and I like this implementation as well. Will go with either opinion.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * A Handler class to modify the result set into List of VersionSet.
+   */
+  public static class VersionSetHandler implements ResultSetHandler<List<VersionSet>> {
+
+    @Override
+    public List<VersionSet> handle(ResultSet rs) throws SQLException {
+      List<VersionSet> versionSets = new ArrayList<>();
+      while (rs.next()) {
+        int versionSetId = rs.getInt(1);
```
arvindpruthi, line 229: Would be nice to use mnemonics instead of hard-coded column numbers.

arvindpruthi (changes requested):

azkaban-common/src/main/java/azkaban/imagemgmt/version/VersionSetLoader.java
```diff
@@ -0,0 +1,54 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.imagemgmt.version;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface VersionSetLoader {
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
aditya1105: done

azkaban-common/src/main/java/azkaban/imagemgmt/version/VersionSet.java
aditya1105: Not sure if this is the correct place, as this VersionSet object is very similar to a POJO. VersionSetBuilder already accepts a Map to construct this JSON string, hence the Map will be there somewhere which can be leveraged.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
aditya1105: I added the comment for the same.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
aditya1105: We are not doing much modification on top of what is returned from the DB. Hence, in this case, separating out DAO from the loader will not be much useful.

azkaban-common/src/main/java/azkaban/imagemgmt/version/JdbcVersionSetLoader.java
aditya1105: done

jakhani (approved):

arvindpruthi (approved):

PR #2664, opened 2020-12-08
Use CredentialProvider instead of CredentialProviderWithKeyStore unless needed.

djaiswal83:
Tested on test cluster.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
@@ -310,7 +310,7 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
     return this.shouldProxy;
   }
 
-  private CredentialProviderWithKeyStore getCustomCredentialProvider(final Props props, final Credentials hadoopCred,
+  private CredentialProvider getCustomCredentialProvider(final Props props, final Credentials hadoopCred,
       final Logger jobLogger, final String customCredentialProviderName) {
     String credentialClassName = "unknown class";
     try {
@@ -321,7 +321,7 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
       // The credential class must have a constructor accepting 3 parameters, Credentials,
       // Props, and Logger in order.
       final Constructor constructor = credentialClass.getConstructor(Credentials.class, Props.class, Logger.class);
-      final CredentialProviderWithKeyStore customCredential = (CredentialProviderWithKeyStore) constructor
+      final CredentialProvider customCredential = (CredentialProviderWithKeyStore) constructor
               .newInstance(hadoopCred, props, jobLogger);
       return customCredential;
     } catch (final Exception e) {
@@ -334,12 +334,22 @@ public class HadoopSecurityManager_H_2_0 extends HadoopSecurityManager {
 
   private void registerCustomCredential(final Props props, final Credentials hadoopCred,
       final String userToProxy, final Logger jobLogger, final String customCredentialProviderName) {
-    final CredentialProviderWithKeyStore customCredential = getCustomCredentialProvider(
+    final CredentialProvider customCredential = getCustomCredentialProvider(
             props, hadoopCred, jobLogger, customCredentialProviderName);
     final KeyStore keyStore = KeyStoreManager.getInstance().getKeyStore();
     if (keyStore != null) {
       // Containerized Execution.
-      customCredential.setKeyStore(keyStore);
+      try {
+        ((CredentialProviderWithKeyStore)customCredential).setKeyStore(keyStore);
+      } catch (ClassCastException e) {
+        logger.error("Encountered error while casting to "
+                + customCredentialProviderName, e);
+        throw new IllegalStateException("Encountered error while casting to "
+                + customCredentialProviderName, e);
+      } catch (final Exception e) {
+        logger.error("Unknown error occurred while setting keyStore");
+        throw new IllegalStateException("Unknown error occurred while setting keyStore");
+      }
     }
     customCredential.register(userToProxy);
   }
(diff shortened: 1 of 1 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -334,12 +334,22 @@ private CredentialProviderWithKeyStore getCustomCredentialProvider(final Props p
...
-      customCredential.setKeyStore(keyStore);
+      try {
+        ((CredentialProviderWithKeyStore)customCredential).setKeyStore(keyStore);
+      } catch (ClassCastException e) {
+        logger.error("Encountered error while casting to "
+                + customCredentialProviderName, e);
+        throw new IllegalStateException("Encountered error while casting to "
+                + customCredentialProviderName, e);
+      } catch (final Exception e) {
+        logger.error("Unknown error occurred while setting keyStore");
```
li-ygerchikov, line 350: Would you like to also report `e` here (and on the next line), so that we know what actually failed and why?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -334,12 +334,22 @@ private CredentialProviderWithKeyStore getCustomCredentialProvider(final Props p
...
+    final CredentialProvider customCredential = getCustomCredentialProvider(
             props, hadoopCred, jobLogger, customCredentialProviderName);
     final KeyStore keyStore = KeyStoreManager.getInstance().getKeyStore();
     if (keyStore != null) {
       // Containerized Execution.
-      customCredential.setKeyStore(keyStore);
+      try {
+        ((CredentialProviderWithKeyStore)customCredential).setKeyStore(keyStore);
+      } catch (ClassCastException e) {
+        logger.error("Encountered error while casting to "
```
li-ygerchikov, line 345: Nit: no, this is not the error -- we aren't casting *to* customCredentialProviderName, we are casting customCredentialProviderName to CredentialProviderWithKeyStore. 
But maybe more is less here, and we don't needs to say much at all, because the cause exception `e` will have that info -- see what the message looks like on the ticket.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -348,8 +358,9 @@ private void registerCustomCredential(final Props props, final Credentials hadoo
   public KeyStore getKeyStore(final Props props) {
     logger.info("Prefetching KeyStore for the flow");
     final Credentials cred = new Credentials();
-    final CredentialProviderWithKeyStore customCredential = getCustomCredentialProvider(
-            props, cred, logger, Constants.ConfigurationKeys.CUSTOM_CREDENTIAL_NAME);
+    final CredentialProviderWithKeyStore customCredential = (CredentialProviderWithKeyStore)
```
li-ygerchikov, line 361: No need to guard this cast because getKeyStore() should never be called with non-keystore provider, right? Probably fine... but will bad things happen if that assumption is violated? Again, probably fine.

li-ygerchikov (changes requested):
Thanks for making this change so quickly.
Two minor comments and a question in code, none about the essence of the change.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -334,12 +334,22 @@ private CredentialProviderWithKeyStore getCustomCredentialProvider(final Props p
 
   private void registerCustomCredential(final Props props, final Credentials hadoopCred,
       final String userToProxy, final Logger jobLogger, final String customCredentialProviderName) {
-    final CredentialProviderWithKeyStore customCredential = getCustomCredentialProvider(
+    final CredentialProvider customCredential = getCustomCredentialProvider(
             props, hadoopCred, jobLogger, customCredentialProviderName);
     final KeyStore keyStore = KeyStoreManager.getInstance().getKeyStore();
     if (keyStore != null) {
       // Containerized Execution.
```
jakhani, line 341: Nit: Add detail comment when keystore is expected. It can be used even without containerization.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -348,8 +358,9 @@ private void registerCustomCredential(final Props props, final Credentials hadoo
   public KeyStore getKeyStore(final Props props) {
```
jakhani, line 358: Nit: Add comment for this method

jakhani:
@li-ygerchikov How can this be tested to make sure this fix will work for you?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: Will do.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: agree, will do.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: I had it initially like that, it wont even compile without the cast.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: Will do

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
djaiswal83: I was planning to address it in the other PR but will do it here for this one.

jakhani:
Is this change covered as a part of unit test?

li-ygerchikov:
> Is this change covered as a part of unit test?

Great question, kind of echoes back to where this situation started. I am torn between asking for more tests (specifically, for a test that would catch this breakage) and wanting to take this change soon because it's blocking people. Maybe merge it now and file a follow up ticket to add a test (if we agree it's a good idea, _and if we actually do it_ at some point)?

jakhani:
Sounds good @li-ygerchikov & @djaiswal83

jakhani (approved):

li-ygerchikov (approved):

PR #2667, opened 2020-12-08
Added class for creation of kubernetes service object using template

aditya1105:
Added a class which will use the template packaged as a resource to create a Service Kubernetes object

### DIFF
--- a/azkaban-common/src/main/resources/azkaban/container/models/v1Service.yaml
+++ b/azkaban-common/src/main/resources/azkaban/container/models/v1Service.yaml
@@ -1,0 +1,22 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: fc-svc-${execid}
+  namespace: ${namespace}
+  annotations:
+    getambassador.io/config: |
+      ---
+      apiVersion: ${apiVersion}
+      kind:  ${kind}
+      name: fc-mapping-${execid}
+      prefix: /${execid}
+      service: fc-svc-${execid}}:${port}
+      rewrite: /
+      timeout_ms: ${timeout_ms}
+spec:
+  selector:
+    service: flow-${execid}
+  ports:
+    - name: flowport
+      port: ${port}
+      targetPort: flowport
--- a/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1ServiceBuilder.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1ServiceBuilder.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.container.models;
+
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
+import io.kubernetes.client.openapi.models.V1Service;
+import io.kubernetes.client.util.Yaml;
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.text.StrSubstitutor;
+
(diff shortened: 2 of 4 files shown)


azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  /**
+   * @param valuesMap Map containing the key-value pairs used for substitution in the template
+   * @return {@link V1Service} object by replacing the values from the valuesMap within the template
+   * @throws IOException if the object can't be constructed from the substituted YAML String
+   */
+  public V1Service getServiceObject(Map<String, String> valuesMap) throws IOException {
+    StrSubstitutor strSubstitutor = new StrSubstitutor(valuesMap);
```
sshardool, line 55: Does V1Service provide a builder that we could use instead of string substitution?
If there is a reason for not using it please mention that in the comments.

azkaban-common/src/main/resources/azkaban/container/models/v1Service.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: fc-svc-${execid}
```
sshardool, line 4: By providing the replacement values as strings we will miss out on any type-checking that could be performed as by a builder.
Even if we use a map for the string substitution could we overlay with it our own builder, something like `AzKubernetesV1ServiceBuilder` and provide methods like
```
builder.withExeciId(execid)
.withPort(port)
.withNamespace(namespace)
...
.build()
``` 
This will provide type safety as well as keep the interface in conformance with what's done for the PodSpec.

sshardool (commented):
Thanks for taking this up!

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private final String template;
+
+  /**
+   * Constructor for instantiation of this class
+   *
+   * @param templateName name of the resource within the package
+   * @throws IOException if the resource file is not readable
+   */
+  public AzKubernetesV1Service(String templateName) throws IOException {
```
jakhani, line 43: As the template is going to be constant for a release, it should be loaded only once and cached.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
```diff
@@ -0,0 +1,59 @@
...
+      this.template = IOUtils.toString(is, Charsets.UTF_8).trim();
+    }
+  }
+
+  /**
+   * @param valuesMap Map containing the key-value pairs used for substitution in the template
+   * @return {@link V1Service} object by replacing the values from the valuesMap within the template
+   * @throws IOException if the object can't be constructed from the substituted YAML String
+   */
+  public V1Service getServiceObject(Map<String, String> valuesMap) throws IOException {
```
jakhani, line 54: Can you please rename valuesMap parameter? templateValuesMap or something more better?

azkaban-common/src/main/resources/azkaban/container/models/v1Service.yaml
arvindpruthi: +1

arvindpruthi (commented):
Love the craftsmanship around the code. But let's see if we can provide a builder pattern as opposed to replacement semantics as Shardool and Janki pointed out.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
aditya1105: We still have to end up using string substitution for the creation of annotation value which is the major chunk for the spec, hence it is better to use a template than builder itself.

azkaban-common/src/main/resources/azkaban/container/models/v1Service.yaml
aditya1105: Thanks for the suggestion. Updated the PR with the same.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
aditya1105: The caller class can make sure to use a single instance everywhere. The intention is that we may have more than one template in the future.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1Service.java
aditya1105: done

jakhani (approved):

sshardool (approved):
Thanks a lot for making the changes!

sshardool (commented):
Looks good apart from a minor nit.

PR #2671, opened 2020-12-10
Add periodic cleanup of containers

sshardool:
Creates a background `ExecutorService` that will be responsible for
cleaning up execution containers running for longer than a configurable
duration.  Note that the background service will be started in a
subsequent commit after (1) K8S container deletion (2) FlowContainer
execution has been implemented.


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This provides the ability to cleanup up execution containers which exceed the maximum allowed
```
jakhani, line 36: Nit: to cleanup execution

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+ * duration for a flow. Following extensions will be considered in future. (1) Convert this to a
+ * general background process for any other containerization related periodic tasks. (2) Add ability
+ * to track containers which may not be have been deleted despite prior attempts. (3) It's assumed
```
jakhani, line 39: Nit: which may not have been

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+ * final status updates to the db are handled by other parts, such as flow-executor. That can also
```
jakhani, line 40: Nit: such as flow-container

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+@Singleton
+public class ContainerCleanupManager {
+
+  private static final Logger logger = LoggerFactory.getLogger(ContainerCleanupManager.class);
+  private static final Duration DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL = Duration.ofMinutes(10);
+  private static final Duration DEFAULT_STALE_CONTAINER_AGE = Duration.ofDays(10);
+  private final long cleanupIntervalMin;
+  private final long staleContainerAgeDays;
+  private final ScheduledExecutorService cleanupService;
```
jakhani, line 51: Will you be adding service clean up as a part of this manager? You should create different thread for service and container. In fact different thread for any different types of clean up.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  /**
+   * Start periodic deletions of per-container resources for stale executions.
+   */
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public void start() {
+    logger.info("Start container cleanup service");
+    this.cleanupService.scheduleAtFixedRate(this::terminateStaleContainers, 0L,
```
jakhani, line 113: Why don't you use scheduleWithFixedDelay instead of scheduleAtFixedRate?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+
+  // Deletes the container specified by executionId while logging and consuming any exceptions.
+  // Note that while this method is not async it's still expected to return 'quickly'. This is true
+  // for Kubernetes as it's declarative API will only submit the request for deleting container
+  // resources. In future we can consider making this async to eliminate any chance of the cleanup
+  // thread getting blocked.
+  private void deleteContainerQuietly(final int executionId) {
+    try {
+      this.containerizedImpl.deleteContainer(executionId);
+    } catch (final ExecutorManagerException eme) {
```
jakhani, line 100: If possible, you should catch ApiException and use e.getResponseBody() for detail message

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // Deletes the container specified by executionId while logging and consuming any exceptions.
+  // Note that while this method is not async it's still expected to return 'quickly'. This is true
+  // for Kubernetes as it's declarative API will only submit the request for deleting container
+  // resources. In future we can consider making this async to eliminate any chance of the cleanup
+  // thread getting blocked.
+  private void deleteContainerQuietly(final int executionId) {
+    try {
+      this.containerizedImpl.deleteContainer(executionId);
```
jakhani, line 99: Let's see if Kubernetes client has method to delete multiple containers in single call. In that case, you can pass list of executions. I need to add deleteAllContainer method for it. Let's use that one over here if possible.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -496,5 +497,13 @@
         AZKABAN_CONTAINERIZED_PREFIX + "execution.processing.thread.pool.size";
     public static final String CONTAINERIZED_CREATION_RATE_LIMIT =
         AZKABAN_CONTAINERIZED_PREFIX + "creation.rate.limit";
+    // Periodicity of lookup and cleanup of  stale executions.
```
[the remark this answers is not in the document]
sshardool: Will do.  (I do have the save-actions enabled, not sure why this wasn't caught)

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -137,6 +138,30 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
     }
   }
 
+  public List<ExecutableFlow> fetchStaleFlows(final long beforeInMs)
+      throws ExecutorManagerException {
+    final List<Status> finalStates = ImmutableList
+        .of(Status.KILLED, Status.FAILED, Status.SUCCEEDED);
```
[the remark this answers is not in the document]
sshardool: Good point. I use `azkaban.executor.ExecutionControllerUtils::isFinished` as a reference for the final states, but `Status::isStatusFailed` is more exhaustive, will change.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -137,6 +138,30 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
+    final StringBuilder query = new StringBuilder(FetchExecutableFlows.FETCH_FLOWS_STARTED_BEFORE);
```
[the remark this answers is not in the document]
sshardool: Thanks for bringing this up. The table as indices on both start_time and status.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: I designed this to be agnostic to the specifics of the container implementation and the code here will interact only with the high level interface `ContainerImpl`. If a service needs to be deleted, that should be handled by the implementation of `ContainerImpl::deleteContainer()`. The K8S ContainerImpl does take care of it. 
For the same reason I don't think having a background thread for cleanup of every type of resource is necessary either. Please let me know if you think otherwise.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: `ApiException` will be specific to the k8s Containerized implementation. This layer is interacting through the interface `ContainerizedImpl` (similar to a previous comment)

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: Good point. That's worth checking.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -496,5 +497,13 @@
...
+    public static final String CONTAINERIZED_STALE_EXECUTION_CLEANUP_INTERVAL_MIN =
+        AZKABAN_CONTAINERIZED_PREFIX + "stale.execution.cleanup.interval.min";
+    // Age of an execution for it to be considered 'stale'.
+    // Note: worth considering providing the age as {@code Duration} instead of days.
+    //   The method {@code java.time.Duration.parse} specifies a format for it.
+    public static final String CONTAINERIZED_STALE_EXECUTION_AGE_DAYS =
```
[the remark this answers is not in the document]
sshardool: I expect we may have to configure this value differently from what's done for the bare-metal executor. I recall there was a discussion around basing this on the validity duration of the ksudo token. 
I would prefer to retain this config variable though we can change it to pick the default value from the currently existing constant. Please let me know what you think.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: `scheduleAtFixedRate` will provide better consistency for the periodicity of the cleanup check. With  `scheduleWithFixedDelay` any unexpected delays in the executions of the cleanup checks and/or the `deleteContainer` procedure will result in an avoidable delay in the next execution.

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
djaiswal83: Yes, we may have to do 7 days if needed.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -137,6 +138,30 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
+    query.append(" AND status NOT IN (");
+    query.append(
+        finalStates.stream()
+            .map(s -> String.valueOf(s.getNumVal()))
+            .collect(Collectors.joining(", ")));
+    query.append(")");
+
```
djaiswal83, line 152: Can you add a comment stating how the query will look in final form?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -137,6 +138,30 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
     }
   }
 
+  public List<ExecutableFlow> fetchStaleFlows(final long beforeInMs)
```
djaiswal83, line 141: beforeInMs is not clear, is it microseconds or milliseconds?
Please consider using full name.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This provides the ability to cleanup up execution containers which exceed the maximum allowed
+ * duration for a flow. Following extensions will be considered in future. (1) Convert this to a
```
djaiswal83, line 37: Please put the numbers points in new line for better readability.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+  // Deletes the container specified by executionId while logging and consuming any exceptions.
+  // Note that while this method is not async it's still expected to return 'quickly'. This is true
+  // for Kubernetes as it's declarative API will only submit the request for deleting container
+  // resources. In future we can consider making this async to eliminate any chance of the cleanup
+  // thread getting blocked.
+  private void deleteContainerQuietly(final int executionId) {
+    try {
+      this.containerizedImpl.deleteContainer(executionId);
+    } catch (final ExecutorManagerException eme) {
+      logger.warn("ExecutorManagerException while deleting container.", eme);
```
djaiswal83, line 101: I believe any exception should be reported as error?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public void terminateStaleContainers() {
+    try {
+      final List<ExecutableFlow> staleFlows = this.executorLoader.fetchStaleFlows(Duration.ofDays(
+          this.staleContainerAgeDays));
+      for (final ExecutableFlow flow : staleFlows) {
+        deleteContainerQuietly(flow.getExecutionId());
+      }
+    } catch (final Exception e) {
+      logger.error("Unexpected exception during container cleanup." + e);
```
djaiswal83, line 88: nit, no need to use concatenation. Exception could be the next param.

djaiswal83 (changes requested):
I have added few comments.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
sshardool: Will do.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: This is done by the save-actions plugin. Unfortunately I didn't find an easy way of disabling this behavior.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: I chose `warn` as failure to delete a container did not seem like a severe enough issue, for example the container may already be gone and only the database was not updated.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: Will do.

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -327,9 +340,10 @@ ExecutableRampDependencyMap fetchExecutableRampDependencyMap()
       throws ExecutorManagerException;
 
   ExecutableRampExceptionalFlowItemsMap fetchExecutableRampExceptionalFlowItemsMap()
-    throws ExecutorManagerException;
+      throws ExecutorManagerException;
```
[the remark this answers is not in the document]
sshardool: This was again done by the save-actions plugin, other exceptions in the same file have a similar indentation. 
I tried avoiding unrelated formatting changes by configuring the plugin settings but they don't seem to apply to these lines :|

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
sshardool: Will do.

djaiswal83 (dismissed):
Thanks for addressing all the comments.
Looks good to me.

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
jakhani: No. It is not required to be 7 days as we are capturing delegation token at the start of job. 7 days is only for job. For flow, it is 10 days. If for existing limit, if property doesn't exist then you should add it for it and reuse it over here. If it exists, then use that where you need for stale flows.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
jakhani: Yes deleteContainer is taking care of deleting pod as well as service. 

Let's say flow has finished within 10 days then how will we terminate service for it?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
jakhani: Until the previous run has finished, if you will start another run then it will bring similar execution if it is not yet processed in previous run and status is not yet updated in db. How will you handle that?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
jakhani: Point noted. I already have added logs for error this case. It will be taken care of.

az-core/src/main/java/azkaban/Constants.java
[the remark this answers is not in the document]
sshardool: Updated to use the same  config `azkaban.server.flow.max.running.minutes`

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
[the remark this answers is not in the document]
sshardool: Added a static `Set` in Status that now provides the correct non-final status to the containerized cleanup as well as the FlowRunnerManager::isFlowRunningLongerThan

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: You're right, that needs to be supported directly within `deleteContainer` implementation. It will be added once the flow-status and healthcheck related changes are committed, as we discussed offline. I will create a ticket for it if it's not already tracked.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: As discussed, `scheduleAtFixedRate` guarantees no concurrent executions will take place, so we should run into this.

jakhani (approved):

PR #2675, opened 2020-12-11
Expose Azkaban Executor Server configuration to the cluster router

haibchen:
The cluster router may need to reuse some configurations that have been defined for Azkaban Executor Server


jakhani (approved):
Thank you for taking care of this fix.

PR #2679, opened 2020-12-16
Add TLS support to ExecutorApiGateway

sshardool:
Modifies RestfulApiClient and ExecutorApiClient to use TLS for
connecting to https based servers.
- Introduces a protected method createHttpClient()
- The method is overridden in ExecutorApiClient to create a tls client
- Adds multiple sanity tests for the ExecutorApiClient
- Adds self-signed certificates, keys in a keystore and trustore for
  tests

Note that tls is not enabled on the jetty ExecutorServer as part of this
change. That support will be considered in future.  For now the
assumption is that an intervening proxy will provide the https support,
which will be the case with containerized deployments.


azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -31,6 +51,63 @@
 @Singleton
 public class ExecutorApiClient extends RestfulApiClient<String> {
 
+  final private static Logger logger = Logger.getLogger(ExecutorApiClient.class);
```
jakhani, line 54: Can you please use slf4j for logging?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -31,6 +51,63 @@
...
+  final private String truststorePath;
+  final private String truststorePassword;
+  private SSLConnectionSocketFactory tlsSocketFactory;
+
+  @Inject
+  public ExecutorApiClient(final Props azkProps) {
+    this.isTlsEnabled = azkProps.getBoolean(EXECUTOR_CONNECTION_TLS_ENABLED, false);
+    this.truststorePath = azkProps
+        .getString(Constants.JETTY_TRUSTSTORE_PATH, this.DEFAULT_TRUSTSTORE_PATH);
+    this.truststorePassword = azkProps.getString(Constants.JETTY_TRUSTSTORE_PASSWORD,
```
jakhani, line 68: Truststore password is already available in private properties. Property name for it is azkaban.csr.truststore.password. Will you be able to reuse it here?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -31,6 +51,63 @@
...
+  final private boolean isTlsEnabled;
+  final private String truststorePath;
+  final private String truststorePassword;
+  private SSLConnectionSocketFactory tlsSocketFactory;
+
+  @Inject
+  public ExecutorApiClient(final Props azkProps) {
+    this.isTlsEnabled = azkProps.getBoolean(EXECUTOR_CONNECTION_TLS_ENABLED, false);
+    this.truststorePath = azkProps
+        .getString(Constants.JETTY_TRUSTSTORE_PATH, this.DEFAULT_TRUSTSTORE_PATH);
```
jakhani, line 67: There is already a property name azkaban.csr.truststore.location for truststore location. Will it be possible to reuse it here?

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -31,6 +51,63 @@
 @Singleton
 public class ExecutorApiClient extends RestfulApiClient<String> {
 
+  final private static Logger logger = Logger.getLogger(ExecutorApiClient.class);
+
+  final private String DEFAULT_TRUSTSTORE_PATH = "keystore";
```
jakhani, line 56: Can you please make it static? Same for DEFAULT_TRUSTSTORE_PASSWORD.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -46,8 +123,9 @@ protected String parseResponse(final HttpResponse response)
 
     if (statusLine.getStatusCode() >= 300) {
 
-      logger.error(String.format("unable to parse response as the response status is %s",
-          statusLine.getStatusCode()));
+      RestfulApiClient.logger
+          .error(String.format("unable to parse response as the response status is %s",
```
jakhani, line 127: Nit: **U**nable. U capital.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiClientTest.java
```diff
@@ -0,0 +1,170 @@
+package azkaban.executor;
```
jakhani, line 1: Can you please add open source disclaimer.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiClientTest.java
```diff
@@ -0,0 +1,170 @@
...
+      resp.getWriter().print(this.GET_RESPONSE_STRING);
+    }
+
+    @Override
+    protected void doPost(final HttpServletRequest req, final HttpServletResponse resp)
+        throws ServletException, IOException {
+      resp.getWriter().print(this.POST_RESPONSE_STRING);
+    }
+  }
+}
```
jakhani, line 170: Nit: New line in the end.

jakhani (commented):
Unit Test cases are very well covered. Everything looks good. Just minor comments are added.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -46,8 +123,9 @@ protected String parseResponse(final HttpResponse response)
 
     if (statusLine.getStatusCode() >= 300) {
```
djaiswal83, line 124: nit: This line is from old code but would it be possible to move the constant to a static variable?

djaiswal83 (changes requested):
The overall code looks good. I have a few minor comments.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
sshardool: Will do.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
sshardool: Good catch. Will fix.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -31,6 +51,63 @@
...
+  private void setupTlsSocketFactory() {
+    SSLContextBuilder sslContextBuilder = SSLContexts.custom();
+    SSLContext sslContext = null;
+    try {
+      sslContextBuilder = sslContextBuilder
+          .loadTrustMaterial(new File(this.truststorePath), this.truststorePassword.toCharArray());
+      sslContext = sslContextBuilder.build();
+    } catch (final NoSuchAlgorithmException | KeyStoreException |
+        CertificateException | IOException | KeyManagementException ex) {
+      logger.error("ExecutorApiClient could not be created due to exception: ", ex);
```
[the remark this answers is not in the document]
sshardool: Good catch, intended to re-throw. Will fix.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
sshardool: The comparison `>= 300` is for http error codes for non-success.  
I think in this instance converting it to a static variable will make the usage less clear for someone looking at this in future, esp as it will retain the comparison `>=`. 
A better alternative would be to replace with a method like `boolean isHttpReturnCodeSuccess(int code)` which encapsulates the comparison but that should be in a separate PR.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
sshardool: `azkaban.csr.truststore.password ` and `azkaban.csr.truststore.location` are used within the mp az-grestin-client for bare-metal executor's interaction with Datavault.  (The `csr` in the name stands for Certificate Signing Request).
I would prefer not to overload these properties for the purpose of the ApiClient.

1. The properties are not very meaningful on the webserver (which shouldn't be fetching certificates from grestin) 
2. It's quite possible (and likely) that we use different keys and certificates for communicating with flow-containers vs the grestin.

Please let me know what you think.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
sshardool: Fixed.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiClientTest.java
sshardool: Done.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiClientTest.java
sshardool: Fixed.

djaiswal83 (commented):
Thanks for addressing my comments. Will leave it open for other pending reviews.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -51,17 +52,18 @@
 @Singleton
 public class ExecutorApiClient extends RestfulApiClient<String> {
 
-  final private static Logger logger = Logger.getLogger(ExecutorApiClient.class);
+  final private static Logger logger = LoggerFactory.getLogger(ExecutorApiClient.class);
+  final private static String DEFAULT_TRUSTSTORE_PATH = "keystore";
```
jakhani, line 56: Nit: better to keep private first and then final for consistency as well scope should come first.

sshardool: Fixed. Thanks for pointing out.

djaiswal83 (approved):
Thanks for addressing all the comments.

jakhani (approved):

PR #2688, opened 2020-12-21
Add a date-time utility method and do a few refactors.

ypadron-in:
-Add a method to format date-time values using the ISO-8601 format "YYYY-MM-DDThh:mm:ssTZD".
-The refactors should help prevent NullPointerException without changing existing code behavior.

### DIFF
--- a/azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
+++ b/azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
@@ -1,3 +1,4 @@
 -- DB Migration from release 3.85.0 to 3.86.0
 -- Adding ExecutionSource column in execution_flows
 alter table execution_flows add column execution_source varchar(32) default null;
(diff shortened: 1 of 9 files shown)


li-gsalia (approved):
Change looks good. Thanks for porting it and refactoring other code to improve the craftsmanship!

azkaban-db/src/main/sql/upgrade.3.85.0.to.3.86.0.sql
```diff
@@ -1,3 +1,4 @@
 -- DB Migration from release 3.85.0 to 3.86.0
 -- Adding ExecutionSource column in execution_flows
 alter table execution_flows add column execution_source varchar(32) default null;
+ALTER TABLE execution_flows ADD COLUMN execution_source VARCHAR(32) DEFAULT NULL;
```
juhoautio-rovio, line 4: This line was most likely supposed to replace the existing one? Functionally it's exactly the same, just uppercase keywords.

ypadron-in: yes, thanks!

PR #2690, opened 2020-12-23
Added regex validation for image version, added validation for specifying at least two owners  during image type creation, minor renaming and removed dead code.

abhisheknath2011:
1. Added regex validation for image version during version create to ensure version is created in major.minor.patch format. 
2. Added validation for specifying at least two owners during image type creation. This would be helpful as we are planning to add add/remove owners API sometime later in future.
3. Minor renaming.
4. Removed dead code.


azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
```diff
@@ -87,7 +88,8 @@ public int createImageType(final ImageTypeDTO imageType) throws ImageMgmtExcepti
   private boolean validateOwnership(final ImageTypeDTO imageType)
       throws ImageMgmtValidationException {
     // Check if ownership record exists
-    if (imageType.getOwnerships() == null) {
+    if (CollectionUtils.isEmpty(imageType.getOwnerships())
```
jakhani, line 91: Can image type owners add owners one by one? Won't this validation stop them from doing that?

jakhani (commented):
Thank you!

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
abhisheknath2011: This validation is triggered during image type creation which is one time activity. During image type creation at least two owners needs to be specified. Also we don't have separate API for adding/removing owners to start with and this will be added sometime in future. Hence, it will be good to have at least two owners to for image types. Moreover, mandate is at least one of them must be ADMIN.

jakhani: Agree with you. Thank you!

jakhani (approved):

PR #2691, opened 2020-12-23
Prevent kill action from leaving executions stuck in KILLING status.

ypadron-in:
When a user requests to kill an execution a check that it's still running is made several times however the execution may finish before the kill action is processed after those checks causing it to stay in KILLING status forever.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -1268,7 +1268,9 @@ public class FlowRunner extends EventHandler<Event> implements Runnable {
 
   public void kill() {
     synchronized (this.mainSyncObj) {
-      if (isKilled()) {
+      if (isKilled() || this.flowFinished) {
+        this.logger.info(
+            "Dropping Kill action as execution " + this.execId + " is already finished.");
         return;
       }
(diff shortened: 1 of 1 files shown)


sshardool (approved):
Looks good. Thanks for taking this up!

djaiswal83 (approved):
Looks good to me. Nice catch!

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1268,7 +1268,9 @@ public void kill(final String user) {
 
   public void kill() {
     synchronized (this.mainSyncObj) {
-      if (isKilled()) {
+      if (isKilled() || this.flowFinished) {
```
jakhani, line 1271: Will this.flowFinished verify with every finished state stated in Status.isStatusFinished method? If so, no need to verify isKilled.

ypadron-in: No, `flowFinished` is an in-memory boolean attribute which is set to true when a flow execution is finalized.

jakhani: Thank you for explaining that.

jakhani (approved):

PR #2692, opened 2020-12-24
Adding nscd volume and environment variable for version set id.

jakhani:
1. Adding a way to setup volume for nscd which is required on Kubernetes.
2. Adding a way to set environment variables in pod spec builder.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -83,6 +83,10 @@ public class KubernetesContainerizedImpl implements ContainerizedImpl {
       "/export/apps/azkaban/azkaban-exec-server/current/plugins/jobtypes";
   public static final String IMAGE = "image";
   public static final String VERSION = "version";
+  public static final String NSCD_SOCKET_VOLUME_NAME = "nscd-socket";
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -83,6 +83,10 @@
       "/export/apps/azkaban/azkaban-exec-server/current/plugins/jobtypes";
   public static final String IMAGE = "image";
   public static final String VERSION = "version";
+  public static final String NSCD_SOCKET_VOLUME_NAME = "nscd-socket";
+  public static final String NSCD_SOCKET_HOST_PATH = "/var/run/nscd/socket";
```
aditya1105, line 87: Should we have this path passed from the Azkaban properties? 
If this is the standard path for nscd even outside LinkedIn, then we can keep it.

aditya1105 (approved):

aditya1105 (approved):

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
jakhani: You are right. Fixed it. Thanks.

PR #2694, opened 2020-12-24
Changes for incorporating the Jetty Server to serve control commands in FlowContainer

arvindpruthi:
- Added Jetty Server and Container Servlet to FlowContainer
    - Added handling of Cancel and fetch log control commands


azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,183 @@
...
+import org.codehaus.jackson.map.ObjectMapper;
+
+import static azkaban.server.HttpRequestUtils.*;
+import static azkaban.common.ServletUtils.*;
+
+
+public class ContainerServlet extends HttpServlet implements ConnectorParams {
+
+  public static final String JSON_MIME_TYPE = "application/json";
+  private static final Logger logger = Logger.getLogger(ContainerServlet.class);
```
abhisheknath2011, line 42: Can we use slf4j logger here as they are used in other part of the codebase?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -143,16 +169,22 @@ private FlowContainer(final Props props) throws ExecutorManagerException {
...
     // setup executor service
     this.executorService = Executors.newSingleThreadExecutor();
 
+    // jetty server
+    this.jettyServer = jettyServer;
+    this.containerContext = context;
+
+    this.eventReporter = eventReporter;
+
     this.jobLogChunkSize = this.azKabanProps.getString("job.log.chunk.size",
```
abhisheknath2011, line 188: Can we use constant here?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -143,16 +169,22 @@ private FlowContainer(final Props props) throws ExecutorManagerException {
...
             DEFAULT_LOG_CHUNK_SIZE);
     this.jobLogNumFiles = this.azKabanProps.getInt("job.log.backup.index", DEFAULT_LOG_NUM_FILES);
```
abhisheknath2011, line 190: Same as above.

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,183 @@
...
+  private FlowContainer application;
```
abhisheknath2011, line 43: nit: is there any specific reason to call as application? Can me rename this to flowContainer?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -111,10 +124,16 @@
...
   private final ProjectLoader projectLoader;
   private final JobTypeManager jobTypeManager;
+  private final Server jettyServer;
+  private final Context containerContext;
+  private final AzkabanEventReporter eventReporter;
   private final Props azKabanProps;
   private Props globalProps;
   private final int numJobThreadPerFlow;
   private String execDirPath;
+  private int port; // Listener port for incoming control & log messages (ContainerServlet)
```
sshardool, line 134: FlowContainer is expected to listen on a fixed local port, which I don't think happens currently. (Reverse-proxy settings need this to be fixed)

Would be ok to do it in a separate PR, please create a TODO for it in that case.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -414,7 +451,113 @@ private void setupKeyStore() throws ExecutorManagerException {
         throw new ExecutorManagerException("Failed to delete the cert file");
       }
     }
+  }
+
+  public void cancelFlow(final int execId, final String user)
```
sshardool, line 456: Similar to the previous comment about FlowContainerServlet. There is overlap in implementation of cancelFlow, readFlowLog, readJobLog, readJobMetadata with the ones in AzkabanExecutorServer and FlowRunnerManager. 
Please see the class can be structured to minimize the overlap.

sshardool (commented):
Thanks a lot for taking this up. Overall implementation looks quite sound. (I understand additional tests will be in a separate PR)
Adding a few comments/questions.

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private void handleFetchLogEvent(final int execId, final HttpServletRequest req,
+      final HttpServletResponse resp, final Map<String, Object> respMap)
+      throws ServletException {
+    final String type = getParam(req, "type");
+    final int startByte = getIntParam(req, "offset");
+    final int length = getIntParam(req, "length");
+
+    resp.setContentType("text/plain");
```
jakhani, line 135: Can you please create constant for this?

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    if (this.application == null) {
+      throw new IllegalStateException(
+          "No flow container defined in this servlet context!");
+    }
+  }
+
+  @Override
+  public void doGet(final HttpServletRequest req, final HttpServletResponse resp)
+      throws ServletException, IOException {
+    handleRequest(req, resp);
```
jakhani, line 62: Here, session id is not validated for any GET or POST call. How will the call be authenticated?

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    try {
+      this.application.cancelFlow(execid, user);
+      respMap.put(ConnectorParams.STATUS_PARAM, ConnectorParams.RESPONSE_SUCCESS);
+    } catch (final ExecutorManagerException e) {
+      logger.error(e.getMessage(), e);
+      respMap.put(ConnectorParams.RESPONSE_ERROR, e.getMessage());
+    }
+  }
+
+  private void handleFetchLogEvent(final int execId, final HttpServletRequest req,
```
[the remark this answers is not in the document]
arvindpruthi: I understand. But for now extracting it out seems more work than necessary as the ExecutorServlet is burried deep into execapp code. I did take a few things out which were doable, but have stayed away from going too far for now.

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
arvindpruthi: This should be taken up separately, I think. This is the same as existing ExecutorServlet

PR #2703, opened 2021-01-06
Added Config for the Azkaban Executor Config

dgarg20:
Added a Configuration for the executor so that executor hostname Ip can be used as host. By Default, it picks up the Canonical Hostname


dgarg20:
Can somebody please merge this

PR #2704, opened 2021-01-06
[Dispatch] Allow case insensitive image type names while getting version from versionSet and rampup manager (Jira: BDP-4551).

abhisheknath2011:
[Dispatch] Allow case insensitive image type names while getting version from versionSet and rampup manager (Jira: BDP-4551). This would help getting image version for an image type in case insensitive manner in KubernetesContainerizedImpl.java as well.


azkaban-common/src/main/java/azkaban/imagemgmt/utils/CaseInsensitiveKeyDeserializers.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This KeyDeserializers implementation class handles case insensitive deserialization of the keys.
+ * This implementation only considers key of type either string or object.
+ */
+public class CaseInsensitiveKeyDeserializers implements KeyDeserializers {
+
+  public static final CaseInsensitiveKeyDeserializer DESERIALIZER = new CaseInsensitiveKeyDeserializer();
+
+  @Override
```
jakhani, line 37: Please add javadoc for public methods.

jakhani (approved):
Thank you very much for taking care of this change.

azkaban-common/src/main/java/azkaban/imagemgmt/utils/CaseInsensitiveKeyDeserializers.java
abhisheknath2011: Done.

PR #2706, opened 2021-01-07
Fetch the azkaban base image and azkaban config version using rampup manager and version set is populated with the same. Dispatch uses these versions while creating POD (Jira: BDP-4549, BDP-4550).

abhisheknath2011:
Fetch the azkaban base image and azkaban config version using rampup manager and version set is populated with the same. Dispatch uses these versions while creating POD (Jira: BDP-4549, BDP-4550).


arvindpruthi (approved):

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -418,8 +425,13 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
     }
-
-    final VersionSet versionSet = fetchVersionSet(executionId, flowParam, jobTypes);
+    // Create all image types by adding azkaban base image, azkaban config and all job types for
+    // the flow.
+    final Set<String> allImageTypes = new TreeSet<>();
+    allImageTypes.add(AZKABAN_BASE_IMAGE);
+    allImageTypes.add(AZKABAN_CONFIG);
+    allImageTypes.addAll(jobTypes);
+    final VersionSet versionSet = fetchVersionSet(executionId, flowParam, allImageTypes);
```
jakhani, line 434: Flow parameter for base and config will work fine right?

abhisheknath2011: Yes flow parameter override for azkaban base and config should work.

PR #2708, opened 2021-01-08
Jetty Server addition for FlowContainer

arvindpruthi:
- Added Jetty Server Module for FlowContainer
- Added FlowContainerTest unit test case. The testcase has todos that will need to be fixed post interface changes being worked in parallel
- Rework of #2694


azkaban-exec-server/src/main/java/azkaban/common/ExecJettyServerModule.java
```diff
@@ -77,4 +84,16 @@ private Context createRootContext(@Named(EXEC_JETTY_SERVER) final Server server)
...
 
+  @Provides
+  @Named(EXEC_CONTAINER_CONTEXT)
+  @Singleton
+  private Context createContainerContext(@Named(EXEC_JETTY_SERVER) final Server server) {
+    final Context container = new Context(server, "/", Context.SESSIONS);
+    container.setMaxFormContentSize(MAX_FORM_CONTENT_SIZE);
+
+    container.addServlet(new ServletHolder(new ContainerServlet()), "/container");
+    //container.setAttribute(Constants.AZKABAN_SERVLET_CONTEXT_KEY, );
```
jakhani, line 95: Nit: Not needed? Can you please remove it?

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,187 @@
...
+import org.slf4j.LoggerFactory;
+import org.codehaus.jackson.map.ObjectMapper;
+
+import static azkaban.common.ServletUtils.writeJSON;
+import static azkaban.server.HttpRequestUtils.hasParam;
+import static azkaban.server.HttpRequestUtils.getParam;
+import static azkaban.server.HttpRequestUtils.getIntParam;
+
+
+public class ContainerServlet extends HttpServlet implements ConnectorParams {
```
jakhani, line 43: Can you please validate whether ContainerServlet can extend LoginAbstractAzkabanServlet? Authentication will come automatically. With that, LoginAbstractAzkabanServlet also extends HttpServlet indirectly so it should work well.

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  private void handleFetchLogEvent(final int execId, final HttpServletRequest req,
+      final HttpServletResponse resp, final Map<String, Object> respMap)
+      throws ServletException {
+    final String type = getParam(req, "type");
+    final int startByte = getIntParam(req, "offset");
+    final int length = getIntParam(req, "length");
+
+    resp.setContentType("text/plain");
```
jakhani, line 139: As text/plain and utf-8 are used at many places, it is good to have constant for it.

azkaban-exec-server/src/test/java/azkaban/container/FlowContainerTestBase.java
```diff
@@ -0,0 +1,13 @@
+package azkaban.container;
+
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+
+
+public class FlowContainerTestBase {
```
jakhani, line 7: What is the usage of this class?

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
arvindpruthi: Tried to do this, but there are a bunch of dependency conflicts that show up. Let's create a separate ticket for this. After merge with Deepak's code, this can be taken up.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -414,7 +459,147 @@ private void setupKeyStore() throws ExecutorManagerException {
...
+   */
+  public LogData readFlowLogs(final int execId, final int startByte, final int length)
+    throws ExecutorManagerException {
+    logger.info("readFlowLogs called");
+    if (this.flowRunner == null) {
+      throw new ExecutorManagerException("The flow has not launched yet!");
+    }
+
+    final File dir = flowRunner.getExecutionDir();
+    if (dir != null && dir.exists()) {
```
djaiswal83, line 501: Can you restructure this to avoid indentation with negative case handled here instead of line 514. This way the try catch block is not inside any condition. Applies to readJobLogs and readJobMetadata too.
Something like,
if (dir == null || !dir.exists() ) {
  throw;
} 

try {}
catch{}

djaiswal83 (changes requested):
Thanks for making this change. Can you please address the comments?

azkaban-exec-server/src/main/java/azkaban/container/ContainerServlet.java
jakhani: Sounds good. Thank you for looking into it.

jakhani (approved):

azkaban-exec-server/src/test/java/azkaban/container/FlowContainerTestBase.java
arvindpruthi: Deleted this class for now.

djaiswal83 (approved):
Thanks for addressing the comments. The code looks good.

PR #2709, opened 2021-01-08
[Dispatch] Skip fetching versions and version set, init container initialization for all the job types included as part of azkaban base image (Jira: BDP-4577).

abhisheknath2011:
[Dispatch] Skip fetching versions and version set, init container initialization for all the job types included as part of azkaban base image (Jira: BDP-4577).


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -192,6 +192,22 @@ public KubernetesContainerizedImpl(final Props azkProps,
...
+    // Add all the job types that are readily available as part of azkaban base image.
+    this.addIncludedJobTypes();
+  }
+
+  /**
+   * Populate the included job types set with all the types that are readily available as part of
+   * azkaban base image.
+   */
+  private void addIncludedJobTypes() {
+    INCLUDED_JOB_TYPES.add("hadoopJava");
```
jakhani, line 204: By any chance this list can come from property with comma separated values and you prepare set out of it? If property is not set then go with default? Just to keep it generic for open source.

abhisheknath2011: Initially I thought about that, but all these job types are integral part of the azkaban code itself, so then why we externalize these job types as part of azkaban configs. Also the configs are cluster specific, so for all the cluster we need to have comma separated list. Hence, decided to keep the the list in the same codebase itself. Moreover, it also does not make sense to accept the list from outside as they these jobs are part of the Azkaban code itself. Also there is a possibility of getting incomplete list if we externalize or allow users to pass the list.

jakhani: Good point.

jakhani (approved):

jakhani (commented):
It will be good to have test case around this.

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
```diff
@@ -298,11 +349,10 @@ private static void addImageRampupEntries(String jsonFile, ObjectMapper objectMa
...
-    versionMap.put("spark", "5.1.5");
-    versionMap.put("command", "3.1.2");
-    versionMap.put("hadoopJava", "4.1.2");
+    versionMap.put(AZKABAN_BASE_IMAGE, "7.0.4");
+    versionMap.put(AZKABAN_CONFIG, "9.1.1");
+    versionMap.put("spark", "8.0");
+    versionMap.put("command", "7.1");
     return versionMap;
   }
 }
```
jakhani, line 358: Nit: New line in the end

jakhani (approved):

PR #2714, opened 2021-01-13
Setting env variable for Kubernetes container for execution id.

jakhani:


sshardool (commented):
Looks good. Thanks for making the change.

sshardool (approved):

PR #2718, opened 2021-01-16
Fix handling of image paths associated with version sets

sshardool:


abhisheknath2011:
@sshardool Thanks for working on this and providing a resolution to image path. Overall the solution looks good, but we can simplify a lot if we incorporate the change in RampupManager API response and VersionSet. Currently RampupManger.getVersionByImageTypes returns Map<String, String> containing imageType as key and imageVersion as value. And based on that versionSet is constructed. But we can change RampupManager.getVersionByImageTypes to return something like Map<String, VersionMetadata>. VersionMetadata can contain imagePath and other metadata. Actually for /status page I am writing overloaded method to return versionMetadata. And based on Map<String, VersionMetadata>, VersionSet can be constructed which will contain Map<String, ImageVersionPath>. With that, VersionSet will contain path information as well. This way we can address this by changing some of the original contracts and with this change we can have some additional info apart from imageVersion. So, what do you think about this approach?

sshardool:
@abhisheknath2011 Thanks for reviewing this and yes I agree the simpler alternatives are possible. My first instinct was to modify `VersionSet` and include the the image-paths there in a manner similar to what you described. But I resisted that urge as I was concerned about any implicit contracts/assumptions that have been built around VersionSets, especially around serde from json strings. I essentially looked for a 'clean' solution that could implemented within the day without modifying VersionSets. 
I didn't realize you were already working on something similar. If you are close to getting it done then I would be happy to not commit this and wait for your PR instead. Alternatively, we can commit this and replace with the simpler/final fix whenever you can get to it.

abhisheknath2011:
Thanks @sshardool for the clarification. Yeah you are perfectly right in your assumption and the changes that you have mentioned is clean, without impacting the existing code. But by extending ImageRampupManager and version set implementation, we can capture more information apart from version and that will be useful going forward. As I am already doing the changes mentioned above, I could extend the solution to capture required information along with version. And you are right this would require some changes specially on the test cases side. As I am already working on this, I should be able take care of the broken test cases as well and raise PR for the same.

sshardool:
Sounds good @abhisheknath2011.
I will close this PR then, as equivalent functionality will be provided as part of your changes. Thanks!

PR #2720, opened 2021-01-19
Rename polling DB lock to include schema name

li-ygerchikov:
...In case multiple Azkaban instances are running (in different schemas) on the same DB instance.


djaiswal83 (approved):
The change is simple and effective!
Thanks for adding the comment for explanation.

jakhani (approved):
Thank you for making this change. Big savior.

sshardool (approved):
Looks good!

djaiswal83 (approved):

PR #2722, opened 2021-01-20
Use environment to fetch execution id and version set id for the flow execution.

djaiswal83:
Use environment to fetch execution id and version set id for the flow execution.
Some cosmetic changes around member variables.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -69,6 +69,8 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.codehaus.jettison.json.JSONException;
+import org.codehaus.jettison.json.JSONObject;
 import org.mortbay.jetty.Connector;
 import org.mortbay.jetty.Server;
 import org.mortbay.jetty.servlet.Context;
@@ -96,8 +98,6 @@ public class FlowContainer {
   private static final String CONF_DIR = "conf";
   private static final String JOB_THREAD_COUNT = "flow.num.job.threads";
   private static final String DEFAULT_LOG_CHUNK_SIZE = "5MB";
-  private static final String FLOW_EXECUTION_ID = "FLOW_EXECUTION_ID";
-  private static final String VERSION_SET_ID = "VERSION_SET_ID";
   private static final int DEFAULT_LOG_NUM_FILES = 4;
   private static final int DEFAULT_JOB_TREAD_COUNT = 10;
   private static final boolean DEFAULT_USE_IN_MEMORY_KEYSTORE = false;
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -96,8 +96,9 @@
   private static final String CONF_DIR = "conf";
   private static final String JOB_THREAD_COUNT = "flow.num.job.threads";
   private static final String DEFAULT_LOG_CHUNK_SIZE = "5MB";
+  private static final String FLOW_EXECUTION_ID = "FLOW_EXECUTION_ID";
```
jakhani, line 99: These variables are already created in constants. Please reuse those.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -209,18 +209,31 @@ public FlowContainer(final Props props,
...
     }
 
-    logger.info("Execution ID : " + execId);
+    // Get the version set id.
+    final String versionSetId = System.getenv(VERSION_SET_ID);
+    if (versionSetId == null) {
+      // Should not happen
+      logger.warn("VersionSet ID is not set!");
+    } else {
+      logger.info("VersionSet ID : {}", versionSetId );
```
jakhani, line 235: Can you please print version set json over here? It will give detail about which image version is used for each type.

sshardool: +1

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: thanks!, did not know.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: Will do.

jakhani: Env variable is only id. You need to fetch version set json from database.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -555,4 +538,57 @@ static void launchCtrlMsgListener(FlowContainer flowContainer) {
...
+   * Log the VersionSet for the flow.
+   * @param versionSetLoader
+   */
+  private void logVersionSet(final VersionSetLoader versionSetLoader) {
+    // Get the version set id.
+    final String versionSetIdStr = System.getenv(
+            Constants.ContainerizedDispatchManagerProperties.ENV_VERSION_SET_ID);
+    if (versionSetIdStr == null) {
+      // Should not happen
+      logger.warn("VersionSet ID is not set!");
```
jakhani, line 575: Warn and return. No need to keep else case.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -555,4 +538,57 @@ static void launchCtrlMsgListener(FlowContainer flowContainer) {
...
+    int execId = -1;
+    try {
+      execId = Integer.parseInt(execIdStr);
+      logger.info("Execution ID : {}", execId);
+    } catch (final NumberFormatException ne) {
+      logger.error("Execution ID set in environment is invalid {}", execIdStr);
+      throw new ExecutorManagerException(ne);
+    }
+    if (execId < 1) {
+      logger.error("Invalid Execution ID : {}", execId);
```
jakhani, line 559: Throw exception if execId < 1.

jakhani (approved):

PR #2723, opened 2021-01-21
Renamed default secret name and secret mountpath

aditya1105:
Renamed default secret name and secret mount path as the same secret volume will be used for other secrets too.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -92,9 +92,9 @@ public class KubernetesContainerizedImpl implements ContainerizedImpl {
   public static final String DEFAULT_NSCD_SOCKET_HOST_PATH = "/var/run/nscd/socket";
   public static final String HOST_PATH_TYPE = "Socket";
   public static final String DEFAULT_NSCD_SOCKET_VOLUME_MOUNT_PATH = "/var/run/nscd/socket";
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -92,9 +92,9 @@
   public static final String DEFAULT_NSCD_SOCKET_HOST_PATH = "/var/run/nscd/socket";
   public static final String HOST_PATH_TYPE = "Socket";
   public static final String DEFAULT_NSCD_SOCKET_VOLUME_MOUNT_PATH = "/var/run/nscd/socket";
-  public static final String DEFAULT_SECRET_NAME = "azkaban-private-properties";
+  public static final String DEFAULT_SECRET_NAME = "azkaban-k8s-secret";
   public static final String DEFAULT_SECRET_VOLUME = DEFAULT_SECRET_NAME;
-  public static final String DEFAULT_SECRET_MOUNTPATH = "/var/azkaban/private/conf";
```
sshardool, line 97: These configurations are currently overwritten in the props we provide in our config files. They need to be overridden there as well after you make this change.

sshardool (approved):
Thanks for making the change.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: Sure Will do

PR #2725, opened 2021-01-22
Add service label to pod spec for routing container API calls

sshardool:
Includes an update to the prefix used in the service template file


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -577,12 +578,20 @@ private String getAzkabanConfigVersion(final VersionSet versionSet) {
...
 
   /**
-   * TODO: Add implementation to get labels for Pod.
+   * Create labels that should be applied to the Pod.
    *
    * @return
    */
-  private ImmutableMap getLabelsForPod() {
-    return ImmutableMap.of("cluster", this.clusterName);
+  private ImmutableMap getLabelsForPod(int executionId) {
```
djaiswal83, line 585: nit : please use final for input param

djaiswal83 (commented):
A minor nit, you can drop it if you want to handle it later.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
sshardool: Fixed.

PR #2726, opened 2021-01-22
Added template based design to add items to the pod-spec

aditya1105:
This template design enables the declaration of static initContainers, volumes, volumeMounts, etc. which are not job-types but are useful. For example: init container which will fetch the required certificates/tokens etc. writes to a volume and then that volume will be mounted to the application-container/flow-container.

Items like InitContainers, Volumes, and VolumeMounts for the application-container/flow-container will be extracted from the POD created from this template file. Extracted Items are later merged with the already created pod-spec(builder based approach).
Merging criterion is such that, if the item in the already created pod-spec has the same name as of the item extracted from the template, then it will be retained.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -560,6 +560,8 @@ public class Constants {
         KUBERNETES_FLOW_CONTAINER_PREFIX + "init.jobtypes.mount.path";
     public static final String KUBERNETES_MOUNT_PATH_FOR_JOBTYPES =
         KUBERNETES_FLOW_CONTAINER_PREFIX + "jobtypes.mount.path";
+    public static final String KUBERNETES_POD_TEMPLATE_PATH =
+        KUBERNETES_POD_PREFIX + "template.path";
 
     // Kubernetes service related properties
     public static final String KUBERNETES_SERVICE_PREFIX = AZKABAN_KUBERNETES_PREFIX + "service.";
--- a/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2021 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.container.models;
+
+import io.kubernetes.client.openapi.models.V1Container;
+import io.kubernetes.client.openapi.models.V1Pod;
+import io.kubernetes.client.openapi.models.V1PodSpec;
+import io.kubernetes.client.openapi.models.V1Volume;
+import io.kubernetes.client.openapi.models.V1VolumeMount;
+import io.kubernetes.client.util.Yaml;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * A singleton class which reads a k8s pod-spec yaml file as template. Items like InitContainers,
+ * Volumes and VolumeMounts for the application-container/flow-container will be extracted from the
+ * POD created from this template file.
+ * <p>
+ * Extracted Items are later merged via {@link azkaban.executor.container.KubernetesContainerizedImpl#mergePodSpec(V1PodSpec,
+ * AzKubernetesV1PodTemplate)} with the pod-spec created using {@link AzKubernetesV1SpecBuilder}
+ * <p>
+ * Merging criteria is such that, if the item in the already created pod-spec has the same name as
+ * of the item extracted from the template, then it will be retained.
+ * <p>
+ * This template design enables declaration of static initContainers, volumes, volumeMounts, etc.
+ * which are not job-types but are useful. For example: init container which will fetch the required
+ * certificates/tokens etc. writes to a volume and then that volume will be mounted to the
+ * application-container/flow-container.
+ *
+ * The template must have only one non-init container. Azkaban k8s design is such that
+ * flow-container will be the only non-init container.
+ */
+public class AzKubernetesV1PodTemplate {
+
+  public static final int FLOW_CONTAINER_INDEX = 0;
+  private V1Pod podFromTemplate;
+  private static AzKubernetesV1PodTemplate instance;
+
+  /**
+   * Private constructor to make this class singleton
+   *
+   * @param templatePath th where the template file is located.
+   * @throws IOException If unable to read the template file.
+   */
+  private AzKubernetesV1PodTemplate(String templatePath) throws IOException {
+    File templateFile = Paths.get(templatePath).toFile();
+    this.podFromTemplate = (V1Pod) Yaml.load(templateFile);
+  }
+
+  /**
+   * @param templatePath Path where the template file is located.
+   * @return Singleton instance of this class.
+   * @throws IOException If unable to read the template file.
+   */
+  public static synchronized AzKubernetesV1PodTemplate getInstance(String templatePath)
+      throws IOException {
+    if (null == instance) {
+      instance = new AzKubernetesV1PodTemplate(templatePath);
+    }
+    return instance;
+  }
+
+  /**
+   * @return the {@link V1Pod} POD generated from the template.
+   */
+  public V1Pod getPodFromTemplate() {
+    return this.podFromTemplate;
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the init containers.
+   * @return The list of filtered init Containers derived from the POD specified in the template.
+   */
+  public List<V1Container> getInitContainers(Predicate<? super V1Container> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    if (null == spec) {
+      return Collections.emptyList();
+    }
+    List<V1Container> initContainers = spec.getInitContainers();
+    return initContainers == null ? Collections.emptyList() :
+        initContainers.stream().filter(filterPredicate).collect(Collectors.toList());
+  }
+
+  /**
+   * @return The list of init Containers derived from the POD specified in the template.
+   */
+  public List<V1Container> getInitContainers() {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    return spec == null ? Collections.emptyList() : spec.getInitContainers();
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the volumes.
+   * @return The list of filtered Volumes derived from the POD specified in the template.
+   */
+  public List<V1Volume> getVolumes(Predicate<? super V1Volume> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    if (null == spec) {
+      return Collections.emptyList();
+    }
+    List<V1Volume> volumes = spec.getVolumes();
+    return volumes == null ?
+        Collections.emptyList() :
+        volumes.stream().filter(filterPredicate).collect(Collectors.toList());
+  }
+
+  /**
+   * @return The list of filtered Volumes derived from the POD specified in the template.
+   */
+  public List<V1Volume> getVolumes() {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    return spec == null ? Collections.emptyList() : spec.getVolumes();
+  }
+
+  /**
+   * This method returns volume mounts only for the first container.
+   *
+   * @param filterPredicate Predicate to filter the Volume Mounts.
+   * @return The list of filtered Volume Mounts to the appContainer derived from the POD specified
+   * in the template.
+   */
+  public List<V1VolumeMount> getContainerVolumeMounts(
+      Predicate<? super V1VolumeMount> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    if (null == spec) {
+      return Collections.emptyList();
+    }
+    List<V1Container> containers = spec.getContainers();
+    if (null == containers || containers.isEmpty()) {
+      return Collections.emptyList();
+    }
+    List<V1VolumeMount> volumeMounts =
+        spec.getContainers().get(FLOW_CONTAINER_INDEX).getVolumeMounts();
+    return null == volumeMounts ? Collections.emptyList()
+        : volumeMounts.stream().filter(filterPredicate).collect(Collectors.toList());
+  }
+
+  /**
+   * This method returns volume mounts only for the first container.
+   *
+   * @return The list of Volume Mounts to the appContainer derived from the POD specified in the
+   * template.
+   */
+  public List<V1VolumeMount> getContainerVolumeMounts() {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    List<V1Container> containers = spec == null ? Collections.emptyList() : spec.getContainers();
+    return containers.isEmpty() ? Collections.emptyList() :
+        containers.get(FLOW_CONTAINER_INDEX).getVolumeMounts();
+  }
+}
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -22,6 +22,7 @@ import azkaban.Constants;
 import azkaban.Constants.ConfigurationKeys;
 import azkaban.Constants.ContainerizedDispatchManagerProperties;
 import azkaban.container.models.AzKubernetesV1PodBuilder;
+import azkaban.container.models.AzKubernetesV1PodTemplate;
 import azkaban.container.models.AzKubernetesV1ServiceBuilder;
 import azkaban.container.models.AzKubernetesV1SpecBuilder;
 import azkaban.container.models.ImagePullPolicy;
@@ -41,11 +42,14 @@ import com.google.common.annotations.VisibleForTesting;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.apis.CoreV1Api;
+import io.kubernetes.client.openapi.models.V1Container;
 import io.kubernetes.client.openapi.models.V1DeleteOptions;
 import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.openapi.models.V1PodSpec;
 import io.kubernetes.client.openapi.models.V1Service;
 import io.kubernetes.client.openapi.models.V1Status;
+import io.kubernetes.client.openapi.models.V1Volume;
+import io.kubernetes.client.openapi.models.V1VolumeMount;
 import io.kubernetes.client.util.ClientBuilder;
 import io.kubernetes.client.util.KubeConfig;
 import io.kubernetes.client.util.Yaml;
@@ -54,6 +58,7 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -62,6 +67,7 @@ import java.util.TreeSet;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -123,6 +129,7 @@ public class KubernetesContainerizedImpl implements ContainerizedImpl {
   private final String secretName;
   private final String secretVolume;
   private final String secretMountpath;
+  private final String podTemplatePath;
 
 
   private static final Logger logger = LoggerFactory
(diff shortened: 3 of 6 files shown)


azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
+package azkaban.container.models;
```
sshardool, line 1: Nit: please add the license header

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+   */
+  public V1Pod getPodFromTemplate() {
+    return this.podFromTemplate;
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the init containers.
+   * @return The list of filtered init Containers derived from the POD specified in the template.
+   */
+  public List<V1Container> getInitContainers(Predicate<? super V1Container> filterPredicate) {
```
sshardool, line 73: Nice use of functionals! Makes this quite flexible for future use.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+   * This method returns volume mounts only for the first container.
+   *
+   * @param filterPredicate Predicate to filter the Volume Mounts.
+   * @return The list of filtered Volume Mounts to the appContainer derived from the POD specified
+   * in the template.
+   */
+  public List<V1VolumeMount> getContainerVolumeMounts(
+      Predicate<? super V1VolumeMount> filterPredicate) {
+    List<V1VolumeMount> volumeMounts =
+        this.podFromTemplate.getSpec().getContainers().get(0).getVolumeMounts();
```
sshardool, line 124: What if the container doesn't have any volumeMounts ?

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * This method returns volume mounts only for the first container.
+   *
+   * @param filterPredicate Predicate to filter the Volume Mounts.
+   * @return The list of filtered Volume Mounts to the appContainer derived from the POD specified
+   * in the template.
+   */
+  public List<V1VolumeMount> getContainerVolumeMounts(
```
sshardool, line 121: This one needs a little more thought. 
There will be times when explicitly don't want to mount volumes present in an init container onto the application (non-init) container. 
For example when one of the init containers is mounting any secrets (certificates for example) that the application-container should not have access to.

One workable but little-less-than-ideal solution would be to dedicate a container in the template, identified by the container name, such that only the volume mounts present on that container will be used for merging within this method.  The name of that container can be either hardcoded or a configurable property. 
What do you think ?

sshardool (commented):
Thanks for taking this up, this is a very neat solution. Have a few comments.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+   * This method returns volume mounts only for the first container.
+   *
+   * @return The list of Volume Mounts to the appContainer derived from the POD specified in the
+   * template.
+   */
+  public List<V1VolumeMount> getContainerVolumeMounts() {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    List<V1Container> containers = spec == null ? Collections.emptyList() : spec.getContainers();
+    return containers.isEmpty() ? Collections.emptyList() :
+        containers.get(0).getVolumeMounts();
```
jakhani, line 138: Do see need of another method which gives volumeMounts for any specific container?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -546,7 +555,17 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
     final VersionSet versionSet = fetchVersionSet(executionId, flowParam, allImageTypes);
     final V1PodSpec podSpec = createPodSpec(executionId, versionSet, jobTypes, flowParam);
-
+    if (StringUtils.isNotEmpty(this.podTemplatePath)) {
+      try {
+        AzKubernetesV1PodTemplate podTemplate = AzKubernetesV1PodTemplate
+            .getInstance(this.podTemplatePath);
+        mergePodSpec(podSpec, podTemplate);
+      } catch (IOException e) {
+        logger.info("ExecId: {}, Failed to create k8s pod from template: {}", executionId,
```
jakhani, line 564: The pod creation should not fail if template is not available.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
     this.executorLoader.updateVersionSetId(executionId, versionSet.getVersionSetId());
   }
 
+  /**
+   * Items extracted from the podTemplate will be merged with the pod-spec. Merging criterion is
```
jakhani, line 587: Nit: merging criteria

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+   * such that, if the item in the already created pod-spec has the same name as of the item
```
jakhani, line 588: Nit: if the item, in the already created pod-spec, has the ....

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+   * extracted from the template, then it will retained.
+   *
+   * @param podSpec     Already created podSpec using the {@link AzKubernetesV1SpecBuilder}
+   * @param podTemplate Instance of the class {@link AzKubernetesV1PodTemplate} to extract items
+   */
+  public static void mergePodSpec(V1PodSpec podSpec, AzKubernetesV1PodTemplate podTemplate) {
+    // Add those volumes which are not already available in the podSpec
+    List<V1Volume> podSpecVolumes = podSpec.getVolumes();
+    if (null != podSpecVolumes) {
```
jakhani, line 597: if podSpecVolumes != null and rest of the places in this PR

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+      List<V1Volume> templateVolumes = podTemplate.getVolumes(
```
jakhani, line 598: It is possible template does not have volume. Please add check for that.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+              .noneMatch(name -> name.equals(tempVol.getName())));
+      for (V1Volume volumeItem : templateVolumes) {
+        podSpec.addVolumesItem(volumeItem);
+      }
+    }
+
+    // Add those initContainers which are not already available in the podSpec
+    List<V1Container> podSpecInitContainers = podSpec.getInitContainers();
+    if (null != podSpecInitContainers) {
+      List<V1Container> templateInitContainers = podTemplate.getInitContainers(
```
jakhani, line 609: It is possible template does not have init container. Please add check for that.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+          tempInitContainer -> podSpecInitContainers.stream().map(V1Container::getName)
+              .noneMatch(name -> name.equals(tempInitContainer.getName())));
+      for (V1Container containerItem : templateInitContainers) {
+        podSpec.addInitContainersItem(containerItem);
+      }
+    }
+
+    List<V1VolumeMount> podContainerVolumeMounts =
+        podSpec.getContainers().get(0).getVolumeMounts();
```
jakhani, line 618: Please create constant for 0 and use it wherever needed.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -564,6 +583,49 @@ private void createPod(final int executionId) throws ExecutorManagerException {
...
+    if (null != podContainerVolumeMounts) {
+      List<V1VolumeMount> templateContainerVolumeMounts = podTemplate.getContainerVolumeMounts(
```
jakhani, line 620: Null check for volumeMounts in template.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: Thanks. Done

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: Thanks.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: Sorry, I forgot. I added similar checks rest of the places in this class. Thanks for pointing out.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: In this one, we are not getting volume mounts for init containers, but only the volume mounts for the flow container which is specified in the template.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -560,6 +560,8 @@
         KUBERNETES_FLOW_CONTAINER_PREFIX + "init.jobtypes.mount.path";
     public static final String KUBERNETES_MOUNT_PATH_FOR_JOBTYPES =
         KUBERNETES_FLOW_CONTAINER_PREFIX + "jobtypes.mount.path";
+    public static final String KUBERNETES_POD_TEMPLATE_PATH =
```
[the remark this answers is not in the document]
aditya1105: Thanks. Done

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+ * <p>
+ * Merging criterion is such that, if the item in the already created pod-spec has the same name as
+ * of the item extracted from the template, then it will be retained.
+ * <p>
+ * This template design enables declaration of static initContainers, volumes, volumeMounts, etc.
+ * which are not job-types but are useful. For example: init container which will fetch the required
+ * certificates/tokens etc. writes to a volume and then that volume will be mounted to the
+ * application-container/flow-container.
+ */
+public class AzKubernetesV1PodTemplate {
```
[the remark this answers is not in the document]
aditya1105: "@singleton" doesn't actually make this class singleton but only instructs guice to reuse the already created instance. We probably will never use injection for this class.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * A singleton class which reads a k8s pod-spec yaml file as template. Items like InitContainers,
+ * Volumes and VolumeMounts for the application-container/flow-container will be extracted from the
+ * POD created from this template file.
+ * <p>
+ * Extracted Items are later merged via {@link azkaban.executor.container.KubernetesContainerizedImpl#mergePodSpec(V1PodSpec,
+ * AzKubernetesV1PodTemplate)} with the pod-spec created using {@link AzKubernetesV1SpecBuilder}
+ * <p>
+ * Merging criterion is such that, if the item in the already created pod-spec has the same name as
```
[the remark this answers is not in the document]
aditya1105: Thanks. Today only I realized criteria is plural for criterion.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * @param templatePath Path where the template file is located.
+   * @return Singleton instance of this class.
+   * @throws IOException If unable to read the template file.
+   */
+  public static synchronized AzKubernetesV1PodTemplate getInstance(String templatePath)
+      throws IOException {
+    if (null == instance) {
```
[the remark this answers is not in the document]
aditya1105: Actually having null first is a better practice. It is less prone to error.
For example, if someone ends up using "=" instead of "=="
if (x = null)
instead of 
if (x == null)

Then the logic will be wrong and the compilation won't throw an error.
However
if (null = x) will not even compile. 

https://stackoverflow.com/questions/8048974/ifvalue-null-vs-ifnull-value

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+    return this.podFromTemplate;
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the init containers.
+   * @return The list of filtered init Containers derived from the POD specified in the template.
+   */
+  public List<V1Container> getInitContainers(Predicate<? super V1Container> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    if (null == spec) {
```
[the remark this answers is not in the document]
aditya1105: Added the reasoning in the earlier comments.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: Added the reasoning in the earlier comments.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: This check is done in AzKubernetesV1PodTemplate

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: AzKubernetesV1PodTemplate

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: AzKubernetesV1PodTemplate

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: This check is done in AzKubernetesV1PodTemplate

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
sshardool: That makes more sense, I didn't realize the call for getting 'container' and 'init-containers' were separate.
In this case it will good enough to document the assumption that there is only one 'non-init-container' present in the template, otherwise someone trying to add more than one container might find the behavior unusual.

sshardool (approved):
Looks good. You can add a comment for this subsequently if you so choose. https://github.com/azkaban/azkaban/pull/2726#discussion_r563024726

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,166 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
```
jakhani, line 2: Nit: 2021

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
jakhani: +1

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: We already have a check for the existence of the property "azkaban.kubernetes.pod.template.path". If this is not specified, then only we will try to use a template. So, if we fail to read the template after the property is specified, then we should fail as we are expecting something else.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
aditya1105: Volume mounts are needed only for the application container(non-init container). As of the current design, flow-container will be the only non-init container.
Init containers from the template are taken completely and hence their volume mounts are already included.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+    V1PodSpec spec = this.podFromTemplate.getSpec();
+    return spec == null ? Collections.emptyList() : spec.getInitContainers();
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the volumes.
+   * @return The list of filtered Volumes derived from the POD specified in the template.
+   */
+  public List<V1Volume> getVolumes(Predicate<? super V1Volume> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
```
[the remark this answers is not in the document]
aditya1105: We have the AzKubernetesV1PodTemplate as a singleton. When instantiated, podFromTemplate will always be non-null as we are using "Yaml.load()" which will not return null but will fail in case it is unable to read or even if the file is empty. Hence it will never be null.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+  public V1Pod getPodFromTemplate() {
+    return this.podFromTemplate;
+  }
+
+  /**
+   * @param filterPredicate Predicate to filter the init containers.
+   * @return The list of filtered init Containers derived from the POD specified in the template.
+   */
+  public List<V1Container> getInitContainers(Predicate<? super V1Container> filterPredicate) {
+    V1PodSpec spec = this.podFromTemplate.getSpec();
```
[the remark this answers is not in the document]
aditya1105: We have the AzKubernetesV1PodTemplate as a singleton. When instantiated, podFromTemplate will always be non-null as we are using "Yaml.load()" which will not return null but will fail in case it is unable to read or even if the file is empty. Hence it will never be null.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * @param templatePath Path where the template file is located.
+   * @return Singleton instance of this class.
+   * @throws IOException If unable to read the template file.
+   */
+  public static synchronized AzKubernetesV1PodTemplate getInstance(String templatePath)
+      throws IOException {
+    if (null == instance) {
+      instance = new AzKubernetesV1PodTemplate(templatePath);
```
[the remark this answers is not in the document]
aditya1105: We do want to fail because we try instantiation only when "azkaban.kubernetes.pod.template.path" is specified. Hence expectation is that we must use the template. If we are unable to, then we should let it fail.
If we don't want to use the template, then "azkaban.kubernetes.pod.template.path" shouldn't be specified in the properties file.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -0,0 +1,140 @@
...
+  private static AzKubernetesV1PodTemplate instance;
+
+  /**
+   * Private constructor to make this class singleton
+   *
+   * @param templatePath th where the template file is located.
+   * @throws IOException If unable to read the template file.
+   */
+  private AzKubernetesV1PodTemplate(String templatePath) throws IOException {
+    File templateFile = Paths.get(templatePath).toFile();
```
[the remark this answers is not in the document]
aditya1105: If we don't want to use the template, then "azkaban.kubernetes.pod.template.path" shouldn't be specified in the properties file, otherwise we should fail if we are unable to read the template.

jakhani:
Thank you for answering all my comments offline.

PR #2730, opened 2021-01-25
FlowContainer implements EventListener. Added shutdown hook.

djaiswal83:


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +621,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+    }
+  }
+
+  /**
+   * Shutdown the Container
+   */
+  private void shutdown() {
+    logger.info("Shutting down the pod");
+    try {
+      this.executorService.shutdown();
```
abhisheknath2011, line 651: Can we add a completion check like awaitTermination or isDone before calling executorService shutdown?

djaiswal83: Good point! will do.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +621,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+      final FlowRunner eventFlowRunner = (FlowRunner)event.getRunner();
+      // Sanity check, assert that the event is for same flow execution
+      if (eventFlowRunner.getExecutionId() != this.flowRunner.getExecutionId()) {
+        logger.warn("Execution id mismatch! Expected execution id: {}, got execution id: {}",
+                eventFlowRunner.getExecutionId(), this.flowRunner.getExecutionId());
+      } else {
+        logger.info("Flow execution finished for execution id: {}",
+                eventFlowRunner.getExecutionId());
+      }
+      shutdown();
```
jakhani, line 641: If flow finished event is not for same exec, even then shutting down?
Actually, I understand this situation should not occur on flow container. But just wondering.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +621,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+    }
+  }
+
+  /**
+   * Shutdown the Container
+   */
```
jakhani, line 647: Nit: Can you please add what all things it is taking care of in comment?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +622,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+  /**
+   * Shutdown the Container. This shuts down the ExecutorService which runs the flow execution
+   * as well as JettyServer.
+   */
+  private void shutdown() {
+    logger.info("Shutting down the pod");
+    while (!this.flowFuture.isDone()) {
+      // This should not happen immediately as submitFlowRunner is a blocking call.
+      try {
+        Thread.sleep(100);
```
jakhani, line 635: Why sleep as future.get is blocking call?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: Does not apply anymore, resolving.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: done

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: It is a good practice to add a sleep otherwise due to some bug, we may end up shutting down before flow finishes.
This is more like a sanity check. In my test runs, this code never exercised.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +622,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+          result = this.executorService.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (final InterruptedException e) {
+          logger.error(e.getMessage());
+        }
+      }
+      this.jettyServer.stop();
+      this.jettyServer.destroy();
+    } catch (final Exception e) {
+      // Eat up the exception
+      logger.error("Error shutting down JettyServer while winding down the FlowContainer", e);
```
sshardool, line 657: The `try{}` block is not specific to JettyServer, please either make the message more generic or alternatively break it into separate `try...catch` blocks (e.g for jettrserver shutdown, executorservice shutdown etc)

jakhani (approved):
Thank you for taking care of this task and answering to my comments.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +622,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+    }
```
sshardool, line 658: Maybe terminate with a non-zero status code in case of exceptions ?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -617,4 +622,41 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+      }
+    }
+
+    try {
+      this.executorService.shutdown();
+      boolean result = false;
+      while (!result) {
+        logger.info("Awaiting Shutdown of executing flow");
+        try {
+          result = this.executorService.awaitTermination(5, TimeUnit.SECONDS);
```
sshardool, line 648: Will this block indefinitely due to the `while` loop if the the executorService is stuck for some reason ?
Consider removing the while loop and log an error if the executorService is still not shutdown, making sure the rest of the procedure can still complete and cleanup any resources (system.exit should do that, worth confirming though)

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: Good call. Earlier shutdown was out of the block. Let me take care of this.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: I do not think it will matter as the pod will just wind down.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: Thanks for pointing this out. I intended to limit the while loop.
Eventually, all this is not necessary as System.exit will just wipe everything out and pod completion will cleanup the pod too.
Nonetheless, as a good practice, I will limit the loop.

sshardool (approved):

abhisheknath2011 (approved):
Thanks for taking care of the comment and testing the same.

PR #2739, opened 2021-01-27
Changed the datatype of activatePlan and forceActivatePlan from primitive boolean to Boolean in order to properly handle deactivation of plan during rampup plan update.

abhisheknath2011:
Changed the datatype of activatePlan and forceActivatePlan from primitive boolean to Boolean in order to properly handle deactivation of plan during rampup plan update. Boolean fields are more troublesome specially when boolean user input is taken. If primitive type boolean is used, null cannot be considered specially when user did not specify input. By default it is considered as false. Non primitive Boolean comes handy for handling such scenarios. Hence, changed the data type from primitive boolean to Boolean.


djaiswal83 (approved):
Looks good to me.

PR #2746, opened 2021-01-29
Fetching all the hadoop tokens from single method and moving call to specific method as responsibility of each implementation

jakhani:
This will provide us a way to make single call to fetch all the hadoop tokens together if we have different implementation for AbstractHadoopSecurityManager.


djaiswal83 (changes requested):
Thanks for refactoring the code.
I have one comment. If lambda conversion is not straightforward, lets just change the name which better suits the functionality.

djaiswal83 (approved):
Thanks for addressing the comments. The code looks good to me.

PR #2748, opened 2021-01-30
Changes for putting image management module behind kill switch and will be enabled when dispatch method is set to CONTAINERIZED.

abhisheknath2011:
Changes for putting image management module behind kill switch and will be enabled when dispatch method is set to CONTAINERIZED.


jakhani (dismissed):

djaiswal83 (changes requested):
This PR is not created the right way, as a result it shows incorrect commits in the messages.
Please create the PR again.

PR #2749, opened 2021-01-31
Changes for putting image management module behind kill switch and will be enabled when dispatch method is set to CONTAINERIZED.

abhisheknath2011:
Changes for putting image management module behind kill switch and will be enabled when dispatch method is set to CONTAINERIZED.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
+++ b/azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
@@ -50,6 +50,7 @@ import azkaban.utils.OsCpuUtil;
 import azkaban.utils.Props;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
+import com.google.inject.Scopes;
 import com.google.inject.name.Names;
 import org.apache.commons.dbutils.QueryRunner;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
```diff
@@ -185,4 +176,26 @@ public AzkabanEventReporter createAzkabanEventReporter() {
     }
     return null;
   }
+
+  private void bindImageManagementDependencies() {
+    if(isContainerizedDispatchMethodEnabled()) {
```
djaiswal83, line 181: Please use the negative case first.
if (!isContainer.... ) {
 return;
} 

Rest of the coede 

.....

This way, there is less indentation. Applies to both modules.

djaiswal83 (changes requested):
The code looks good. I have a suggestion to restructure the code a little.

azkaban-common/src/main/java/azkaban/AzkabanCommonModule.java
abhisheknath2011: I believe both are one and the same in this code as we don't have much conditions to handle. So the current code is checking if containerization is enabled then only do the bindings. You are right about the negative usecase and I agree it would be useful if we have more validations or usecases.  But we have only one use case here if containerization is enable then initialize the bindings. So either of the change works. Hence, retaining the same code.

djaiswal83: We may not have them now and likely never but we never know. However, we can do it later. Approving.

djaiswal83 (approved):

PR #2751, opened 2021-02-02
[Status page] Show non active latest version for image types without active rampup plan and active version.

abhisheknath2011:


azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -251,6 +258,54 @@ public int createImageVersion(final ImageVersion imageVersion) {
...
+      final StrSubstitutor strSubstitutor = new StrSubstitutor(valueMap);
+      final String query = strSubstitutor.replace(SELECT_LATEST_IMAGE_VERSION_QUERY);
+      log.info("Image version getActiveVersionByImageTypes query : " + query);
+      final List<Object> params = new ArrayList<>();
+      final Set<String> imageTypesInLowerCase =
+          imageTypes.stream().map(String::toLowerCase).collect(Collectors.toSet());
+      log.info("imageTypesInLowerCase: " + imageTypesInLowerCase);
+
+      params.addAll(nonActiveStates);
+      params.addAll(imageTypesInLowerCase);
```
jakhani, line 295: Why these image types are added two time? Shouldn't there be join for status?

azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
```diff
@@ -271,8 +292,35 @@ public VersionInfo getVersionInfo(final String imageType, final String imageVers
...
+   * @param imageTypes
+   * @return Map<String, ImageVersion>
+   */
+  private Map<String, ImageVersion> getLatestNonActiveImageVersion(final Set<String> imageTypes) {
+    final Map<String, ImageVersion> imageTypeLatestNonActiveVersionMap =
+        new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    if (!CollectionUtils.isEmpty(imageTypes)) {
+      final List<ImageVersion> imageVersions =
+          this.imageVersionDao.getLatestNonActiveVersionByImageTypes(imageTypes);
+      log.debug("Non Active image versions fetched: {} ", imageVersions);
```
jakhani, line 307: Should it be info?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -251,6 +258,54 @@ public int createImageVersion(final ImageVersion imageVersion) {
...
 
+  @Override
+  public List<ImageVersion> getLatestNonActiveVersionByImageTypes(final Set<String> imageTypes)
+      throws ImageMgmtException {
+    List<ImageVersion> imageVersions = new ArrayList<>();
+    try {
+      // Add outer select clause
+      final StringBuilder inClauseBuilder = new StringBuilder();
+      for (int i = 0; i < imageTypes.size(); i++) {
+        inClauseBuilder.append("?,");
```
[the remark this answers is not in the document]
abhisheknath2011: String.join works well if we know the strings to be be appended before hand and they are fixed. In this case "?" is appended for each image type to form the in clause. If we use string.join in the loop it will create a new string every time which is expensive. On the other hand StringBuilder will append to the existing one and thus avoiding creation of new string.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -251,6 +258,54 @@ public int createImageVersion(final ImageVersion imageVersion) {
...
+      final Map<String, String> valueMap = new HashMap<>();
+      valueMap.put("image_types", inClauseBuilder.toString());
+
+      Set<String> nonActiveStates = new TreeSet<>();
+      nonActiveStates.add(State.NEW.getStateValue());
+      nonActiveStates.add(State.UNSTABLE.getStateValue());
+      nonActiveStates.add(State.DEPRECATED.getStateValue());
+      final StringBuilder statesInClauseBuilder = new StringBuilder();
+      for (int i = 0; i < nonActiveStates.size(); i++) {
+        statesInClauseBuilder.append("?,");
```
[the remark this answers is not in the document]
abhisheknath2011: Same as above.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -251,6 +258,54 @@ public int createImageVersion(final ImageVersion imageVersion) {
...
+      final StringBuilder inClauseBuilder = new StringBuilder();
+      for (int i = 0; i < imageTypes.size(); i++) {
+        inClauseBuilder.append("?,");
+      }
+      inClauseBuilder.setLength(inClauseBuilder.length() - 1);
+      final Map<String, String> valueMap = new HashMap<>();
+      valueMap.put("image_types", inClauseBuilder.toString());
+
+      Set<String> nonActiveStates = new TreeSet<>();
+      nonActiveStates.add(State.NEW.getStateValue());
```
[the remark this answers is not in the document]
abhisheknath2011: Done

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
abhisheknath2011: Yes image types are added two times because there are two inner queries containing image types in clause.

azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
abhisheknath2011: Changed to info.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
jakhani: Will the join with status column help for two inner joins?

abhisheknath2011: Status and image types both are part of the queries (both inner queries) and they need to be populated in in order to substitute the query parameters. Btw, this is an existing query and is reused for non active states here as well. For more clarity have added one liner comment regarding addition of these parameters.

jakhani: That sounds good. Thank you for answering.

jakhani (approved):

PR #2715, opened 2021-01-14
Add executor api calls for containerized flows

sshardool:
Modifies ExecutorApiClient and ExecutorApiGateway for
 - Handling Reverse-proxy based access to Executors/FlowContainers
 - Containerized dispatch to FlowContainers

Add executor api call invocations from ContainerizedDispatchManager
 - Done indirectly by adding invocation in base class AbstractDispatchManager
 - Adds getExecutableFlowLog, getExecutionJobLog, getExecutionJobStats, cancelFlow

PR #2752, opened 2021-02-02
Fix handling of null executionId in ExecutorApiGateway

sshardool:
Handling of `null` execution-id in ExecutorApiGateway.callWithExecutionId() was broken by commit
https://github.com/azkaban/azkaban/pull/2715/commits/38f3c103b771d1cf4687f9e97a0e0537dea1a1e4

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import javax.inject.Singleton;
 import org.codehaus.jackson.map.ObjectMapper;
@@ -91,11 +92,16 @@ public class ExecutorApiGateway {
   }
 
   @VisibleForTesting
-  String createExecutionPath(int execId) {
+  String createExecutionPath(Optional<Integer> executionId) throws ExecutorManagerException {
     if (!isReverseProxyEnabled) {
       return "/" + executionResourceName;
     }
-    return "/" + executionResourceNameModifier.apply(execId, executionResourceName);
+
+    if(!executionId.isPresent()) {
+      throw new ExecutorManagerException(
+          "Execution Id must be provided when reverse-proxy is enabled");
+    }
+    return "/" + executionResourceNameModifier.apply(executionId.get(), executionResourceName);
   }
 
   Map<String, Object> callWithExecutionId(final String host, final int port,
@@ -116,7 +122,7 @@ public class ExecutorApiGateway {
 
       // Ideally we should throw an exception if executionId is null but some existing code
       // (updateExecutions()) expects to call this method with a null executionId.
-      String executionPath = (executionId == null) ? null : createExecutionPath(executionId);
+      String executionPath = createExecutionPath(Optional.ofNullable(executionId));
       return callForJsonObjectMap(host, port, executionPath, paramList);
     } catch (final IOException e) {
       throw new ExecutorManagerException(e.getMessage(), e);
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
@@ -34,9 +34,12 @@ import java.util.Optional;
 import java.util.function.BiFunction;
 import javax.inject.Singleton;
 import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Singleton
 public class ExecutorApiGateway {
+  private final static Logger logger = LoggerFactory.getLogger(ExecutorApiGateway.class);
   public final static String DEFAULT_EXECUTION_RESOURCE = "executor";
   public final static String CONTAINERIZED_EXECUTION_RESOURCE = "container";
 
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -116,7 +122,7 @@ String createExecutionPath(int execId) {
 
       // Ideally we should throw an exception if executionId is null but some existing code
       // (updateExecutions()) expects to call this method with a null executionId.
-      String executionPath = (executionId == null) ? null : createExecutionPath(executionId);
+      String executionPath = createExecutionPath(Optional.ofNullable(executionId));
```
li-ygerchikov, line 125: I hope it's ok to exit with an unhandled (and unreported, at least not yet) exception here if `createExecutionPath` throws.

li-ygerchikov (approved):
Thanks for improving the unit tests, too!

azkaban-common/src/test/java/azkaban/executor/ExecutorApiGatewayTest.java
```diff
@@ -85,23 +90,100 @@ public void updateExecutions() throws Exception {
...
   @Test
-  public void testPathWithContainerizedConfigs() {
-    Props containerizedConfigs = new Props();
-    containerizedConfigs.put(ConfigurationKeys.AZKABAN_EXECUTOR_REVERSE_PROXY_ENABLED, "true");
-    containerizedConfigs.put(ConfigurationKeys.AZKABAN_EXECUTION_DISPATCH_METHOD,
-        DispatchMethod.CONTAINERIZED.name());
-    ExecutorApiGateway gateway = new ExecutorApiGateway(client, containerizedConfigs);
+  public void testPathWithContainerizedConfigs() throws Exception {
+    final ExecutorApiGateway gateway = gatewayWithConfigs(this.client, containerizationEnabledProps());
     int executionId = 12345;
```
djaiswal83, line 102: not used. See comment above

azkaban-common/src/test/java/azkaban/executor/ExecutorApiGatewayTest.java
```diff
@@ -85,23 +90,100 @@ public void updateExecutions() throws Exception {
...
+  @Test
+  public void testPingWithDefaultConfigs() throws Exception {
+    Props props = new Props();
+    final ExecutorApiClient clientSpy = Mockito.spy(new SendDisabledExecutorApiClient(props));
+
+    final ExecutorApiGateway gateway = gatewayWithConfigs(clientSpy, props);
+    final String apiAction = ConnectorParams.PING_ACTION;
+    final String apiHost = "host1";
+    final int apiPort = 1234;
+    final Map<String, Object> response = gateway.callWithExecutionId(apiHost, apiPort, apiAction,
```
djaiswal83, line 132: Please write a comment here, that this call is not expected to throw an exception something like comment at line 171

djaiswal83 (changes requested):
Thanks for fixing the issue and adding testcases.
Please address the comments below.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
sshardool: It should be ok as `callWithExecutionId` already throws that exception, however for good measure I have updated `createExecutionPath` to log the exception. 
Please let me know if I misunderstood your comment.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -91,11 +92,16 @@ public ExecutorApiGateway(final ExecutorApiClient apiClient, Props azkProps) {
   }
 
   @VisibleForTesting
-  String createExecutionPath(int execId) {
+  String createExecutionPath(Optional<Integer> executionId) throws ExecutorManagerException {
```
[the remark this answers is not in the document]
sshardool: Fixed.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiGatewayTest.java
```diff
@@ -85,23 +90,100 @@ public void updateExecutions() throws Exception {
...
+    final ExecutorApiGateway gateway = gatewayWithConfigs(this.client, containerizationEnabledProps());
     int executionId = 12345;
-    String path = gateway.createExecutionPath(12345);
+    String path = gateway.createExecutionPath(Optional.of(12345));
     Assert.assertEquals("/" + executionId + "/" +
             ExecutorApiGateway.CONTAINERIZED_EXECUTION_RESOURCE,
         path);
   }
+
+  private Props containerizationEnabledProps() {
```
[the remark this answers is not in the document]
sshardool: Done.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiGatewayTest.java
sshardool: Fixed.

azkaban-common/src/test/java/azkaban/executor/ExecutorApiGatewayTest.java
```diff
@@ -85,23 +90,100 @@ public void updateExecutions() throws Exception {
   }
 
   @Test
-  public void testPathWithDefaultConfigs() {
+  public void testPathWithDefaultConfigs() throws Exception {
     int executionId = 12345;
```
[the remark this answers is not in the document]
sshardool: I would prefer to keep it in the method as in several newer test cases we are assuming this value for port (and other literals are handled similarly) when creating the expected output, which keeps the test easier to follow. 
Let me know if you feel strongly about it.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -92,14 +95,15 @@ public ExecutorApiGateway(final ExecutorApiClient apiClient, Props azkProps) {
...
+  String createExecutionPath(final Optional<Integer> executionId) throws ExecutorManagerException {
     if (!isReverseProxyEnabled) {
       return "/" + executionResourceName;
     }
 
     if(!executionId.isPresent()) {
-      throw new ExecutorManagerException(
-          "Execution Id must be provided when reverse-proxy is enabled");
+      final String errorMessage = "Execution Id must be provided when reverse-proxy is enabled";
+      logger.error(errorMessage);
```
li-ygerchikov, line 105: This is totally fine, no need to change again, but one trick that you may want to consider in cases like this is along the lines of:
```suggestion
final ExecutorManagerException eme = new ExecutorManagerException("...");
logger.error(eme);
throw eme;
```
This way you get a stack trace -- that is, the precise location and call chain and maybe the "caused by" if there were other exceptions -- in the log, compared to just a message.

li-ygerchikov (approved):

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
sshardool: Very useful! Thanks for sharing.

djaiswal83 (approved):
Thanks for addressing the comments. The PR looks good.

PR #2754, opened 2021-02-03
Allow multiple copies of the same dependency file in Thin Zips.

ypadron-in:
Sometimes, to isolate dependencies and avoid conflicts the same dependency jar file is required in different paths in a project. For example:
```
common-libs
 |-jackson-annotations-2.6.0.jar
 |-….jar
 |-….jar
 |-other-libs
     |-jackson-annotations-2.6.0.jar
     |-….jar
```
To allow this we need to take into consideration dependency `destination` when building the _Set_ of dependencies to download.


djaiswal83 (approved):
Thanks for fixing this bug.

PR #2756, opened 2021-02-03
Fix flow cancellation for containerized dispatch

sshardool:


azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -320,6 +323,25 @@ public void run() {
...
+      final Map<Integer, Pair<ExecutionReference, ExecutableFlow>> unfinishedFlows = this.executorLoader
+          .fetchUnfinishedFlows();
+      if (unfinishedFlows.containsKey(exFlow.getExecutionId())) {
+        final Pair<ExecutionReference, ExecutableFlow> pair = unfinishedFlows
+            .get(exFlow.getExecutionId());
+        // Note that ExecutionReference may have the 'executor' as null. ApiGateway call is expected
+        // to handle this scenario.
+        this.apiGateway.callWithReferenceByUser(pair.getFirst(), ConnectorParams.CANCEL_ACTION, userId);
+      } else {
+        throw new ExecutorManagerException("Execution "
```
jakhani, line 339: Can you please add error log over here? I also miss this alot :-)

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -320,6 +323,25 @@ public void run() {
     }
   }
 
+  @Override
+  public void cancelFlow(ExecutableFlow exFlow, String userId)
```
jakhani, line 327: What's the difference between method in AbstractExecutorManagerAdapter and this?

sshardool: The implementation in `AbstractExecutorManagerAdapter` explicitly relies on the existence of a non-null `executor` in the ExecutionReference which result in the failure of cancelFlow. For ContainerizedDispatch a valid executor is not required and will typically not be present. (If it does exist for any reason, ApiGateway calls have already been modified to handle it correctly)

jakhani (approved):

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
sshardool: Fixed.

PR #2761, opened 2021-02-11
Add ping to the ContainerServlet

aditya1105:
Added ping method to Servlet
Refactored code to use Switch instead of if/else
Added tests
Corrected Azkaban checkstyle


sshardool (approved):
Thanks for taking care of this!

PR #2762, opened 2021-02-12
API for deleting image version and associated metadata

abhisheknath2011:
For deleting an image type version is following use cases are considered. 
Do not proceed with delete for the below scenarios.
1. If the image type version is in ACTIVE/NEW state, throw appropriate exception with the version detail and 
2. If the image type version is associated of rampup plan, throw appropriate exception and display the rampup plans and the rampup details.
3. forceDelete flag is set to false.

Proceed with delete for the below scenarios.
1. If the image type version is not associated with any rampup plan and version state is NOT ACTIVE/NEW.
2. forceDelete flag is set to true.

Image version metadata to be deleted:
1.  Image rampup details for an associated image type version (if exists).
2. Image rampup plan for an associated image type version (if exists).
3. Image version details.


azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageMgmtCommonDaoImpl.java
```diff
@@ -0,0 +1,230 @@
...
+    final List<ImageRampupPlan> imageRampupPlans = this.imageRampupDao
+        .getImageRampupPlans(imageTypeName,
+            versionId);
+    // Collect the plan ids and delete the image rampup and rampup plan using that.
+    final List<Integer> planIds = new ArrayList<>();
+    if (!CollectionUtils.isEmpty(imageRampupPlans)) {
+      for (final ImageRampupPlan imageRampupPlan : imageRampupPlans) {
+        planIds.add(imageRampupPlan.getId());
+      }
+      // If the version is used in any of the rampup plans throw error with the rampup plan details.
```
jakhani, line 107: Here it means only active rampup plans right? If so, then there should be only single plan id coming. If not then what is the point of deleting inactive plans? Another thing to think about is shall we just do soft delete for image version? If version is marked for soft delete then it should not be allowed to be picked up by any ramp up plan.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageMgmtCommonDao.java
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Method to delete image version and associated metadata.
+   *
+   * @param imageTypeName
+   * @param versionId
+   * @param forceDelete
+   * @return
+   * @throws ImageMgmtException
+   */
+  public DeleteResponse deleteImageVersion(final String imageTypeName, final int versionId,
```
jakhani, line 36: Why is this method not part of ImageVersionDao?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageRampupDaoImpl.java
```diff
@@ -104,6 +104,12 @@
       + "irp.id = ir.plan_id and iv.id = ir.version_id and irp.type_id = it.id and irp.active = ? "
       + "and iv.type_id = it.id and it.active = ?";
 
+  private static final String SELECT_RAMPUP_PLAN_CONTAINING_IMAGE_VERSION_QUERY = "select irp.id, "
+      + "irp.name, irp.description, irp.active, it.name image_type_name, irp.created_on, "
+      + "irp.created_by, irp.modified_on, irp.modified_by from image_rampup_plan irp, "
+      + "image_types it where irp.type_id = it.id and lower(it.name) = ? and irp.id in (select "
+      + "plan_id from image_rampup where version_id = ?)";
```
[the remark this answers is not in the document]
abhisheknath2011: Good point. Changed the query to use inner joins.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageMgmtCommonDaoImpl.java
abhisheknath2011: The idea to completely delete image version and associated metadata which includes image rampup plan and rampup details. Inactive rampup plans are of no use as we allow update only on active rampup plan. Hence, while deleting rampup plans inacitve plans are also considered. 

Regarding the soft delete of image versions we have to put many if and checks in the code. For example, if the version is marked as soft delete, what will happen to the rampup plan and rampup details. They also need to be marked as soft delete. Also we need to put some conditions in the code as well. Also we have foreign key constraints on all these tables. Moreover, the version cleanup can be leveraged to clean up of unused data as well.

The delete image version API ensures that if the version is in NEW/ACTIVE state or there are associated rampup plans, the delete will fail and display the usage details. Only by specifying forceDelete flag to "true" we should be able to forcefully delete the image version metadata.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageMgmtCommonDao.java
abhisheknath2011: The purpose of delete image version is to delete complete metadata. Delete image version touches multiple tables such as image_versions, image_types, image_rampup_plan and image_rampup and hence required to use some of the existing daos. That is the reason why this method is part of the common dao.

jakhani (approved):

PR #2768, opened 2021-02-17
create-all-sql-*.sql is missing in some distributions

li-ygerchikov:
Note how `distributions` below collects from "$buildDir/sql" -- *not* from 'build/sql'.
Turns out that $buildDir can be azkaban/build/azkaban-db *or* azkaban/azkaban-db/build,
depending on the build configuration. So it's important to be consistent about (1) where
the create-all-sql-*.sql is created and (2) where it is picked up from into distributions.

### DIFF
--- a/azkaban-db/build.gradle
+++ b/azkaban-db/build.gradle
@@ -48,7 +48,7 @@ dependencies {
  */
 task concat() {
     doLast {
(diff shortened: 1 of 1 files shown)


djaiswal83 (approved):

PR #2773, opened 2021-02-23
Changed start status to READY instead of PREPARING for poll based dispatch

aditya1105:
This change makes sure that poll-based dispatch starts with the flow status READY and once the execution is polled, it will be updated to PREPARING. 
Added the test for the same.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -382,7 +382,7 @@ public class ExecutionFlowDao {
 
   public int selectAndUpdateExecution(final int executorId, final boolean isActive)
       throws ExecutorManagerException {
-    final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ?, update_time = ? "
+    final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ?, update_time = ?, status=? "
         + "where exec_id = ?";
     final String selectExecutionForUpdate = isActive ?
         SelectFromExecutionFlows.SELECT_EXECUTION_FOR_UPDATE_ACTIVE :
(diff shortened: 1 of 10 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -163,6 +163,11 @@ private ExecutorService createExecutorInfoRefresherService() {
         new ThreadFactoryBuilder().setNameFormat("azk-refresher-pool-%d").build());
   }
 
+  @Override
+  public Status getStartStatus() {
+    return Status.PREPARING;
```
jakhani, line 168: Shall we mark start status as READY for PUSH based method as well?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -426,10 +427,11 @@ public int selectAndUpdateExecutionWithLocking(final int executorId, final boole
       if (hasLocked) {
         try {
           final List<Integer> execIds = transOperator.query(selectExecutionForUpdate,
-              new SelectFromExecutionFlows(), Status.PREPARING.getNumVal(), executorId);
+              new SelectFromExecutionFlows(), Status.READY.getNumVal(), executorId);
           if (CollectionUtils.isNotEmpty(execIds)) {
             execId = execIds.get(0);
-            transOperator.update(UPDATE_EXECUTION, executorId, System.currentTimeMillis(), execId);
+            transOperator.update(UPDATE_EXECUTION, executorId, System.currentTimeMillis(),
```
jakhani, line 433: Can you please make a change in selectAndUpdateExecutionWithLocking(final boolean batchEnabled,
      final int limit, final Status updatedStatus) method? It should not expect status anymore. It will be consistent to this change that you are making.

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
aditya1105: Poll and containerization has steps where ready is changed to preparation. Push doesn't read from the db so there is no point in adding the ready status and then updating to prepare. Hence prepare there should suffice.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
aditya1105: Let's take it in separate PR because couple of test files also will be changed.

jakhani: Like discussed offline, I will take care of this change in separate PR.

jakhani (approved):

jakhani:
It is against cf-engine-k8s branch. Please raise it for master.

PR #2777, opened 2021-02-26
Add probes from the podTemplate to the podSpec

aditya1105:
Probes are static in nature and hence if they are present in the pod template, they should be added to the actual pod spec during dispatch.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -27,6 +27,7 @@ import azkaban.container.models.AzKubernetesV1PodTemplate;
 import azkaban.container.models.AzKubernetesV1ServiceBuilder;
 import azkaban.container.models.AzKubernetesV1SpecBuilder;
 import azkaban.container.models.ImagePullPolicy;
+import azkaban.container.models.PodTemplateMergeUtils;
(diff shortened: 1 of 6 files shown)


sshardool (approved):
The static pod-template-merge you added earlier is quite handy. Nice work!

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -642,6 +643,23 @@ public static void mergePodSpec(V1PodSpec podSpec, AzKubernetesV1PodTemplate pod
             .addVolumeMountsItem(volumeMountItem);
       }
     }
+
```
jakhani, line 646: Is it possible to take this part in another method? 
mergePodSpec (...) {
  mergeVolumes(..);
  mergeInitContainers(..);
  mergeVolumeMounts(..);
  mergeProbes(..);
}

Something like this?

jakhani (commented):
You have designed this templatization part very nicely. It is very neat and clean approach. Thank you!

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: Good suggestion. I added a separate Utils class for this. Please review.

jakhani (approved):
So neat! Thank you for fixing it.

PR #2779, opened 2021-02-26
Added mechanism to rampup Containerized from POLL dispatch

aditya1105:
This ramp-up mechanism requires that containerization be enabled and ramp-up percentages are set appropriately. Setting it to 100 will make all the executions use Containerization and setting it to 0 implies they will be polled by the executor server.

This allows ContainerizedDispatchManager to submit/upload some flows to the execution_flows table based on the ramp percentage with an additional column value dispatch_method. Whenever executions are pulled from the execution_flows table, respective dispatch managers will query by filtering the dispatch_method. Any future changes to the ContainerizedDispatchManager will automatically be honored as there is no new DispatchManager.

If this mechanism needs to removed then the only change will be required in the class azkaban.executor.container.ContainerizedDispatchManager where getDispatchMethod will be changed to always return DispatchMethod.CONTAINERIZED

Following is the summary of the change:

1) Added a new column dispatch_method in the table exectuion_flows 
2) Modified the getDispatchMethod() in the azkaban.executor.container.ContainerizedDispatchManager to determine the dispatch based on the ramp-up value set in the constructor.
3) getDispatchMethod() is used while the flow is submitted by azkaban.executor.AbstractExecutorManagerAdapter#uploadExecutableFlow
4) Modified the azkaban.executor.ExecutionFlowDao to make sure that dispatch_method value is also added to the execution_flows table while invoking uploadExecutableFlow()
5) Modified the azkaban.executor.ExecutionFlowDao to make sure that all selectAndUpdateExecution methods will require DispatchMethod as a parameter for querying.
6) Whenever selectAndUpdateExecution methods are used respective DispatchManagers will pass the appropriate DispatchMethod and hence only the corresponding executions are polled.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/DispatchMethod.java
+++ b/azkaban-common/src/main/java/azkaban/DispatchMethod.java
@@ -22,10 +22,19 @@ import org.slf4j.LoggerFactory;
  * This enum contains list of dispatch types implemented in Azkaban.
  */
 public enum DispatchMethod {
-  PUSH,
-  POLL,
-  CONTAINERIZED;
+  PUSH(0),
+  POLL(1),
+  CONTAINERIZED(2);
   private static final Logger logger = LoggerFactory.getLogger(DispatchMethod.class);
+  private final int numVal;
+
+  DispatchMethod(final int numVal) {
+    this.numVal = numVal;
+  }
+
+  public int getNumVal() {
+    return this.numVal;
+  }
 
   public static DispatchMethod getDispatchMethod(String value) {
     try {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
@@ -16,6 +16,7 @@
 
 package azkaban.executor;
 
+import azkaban.DispatchMethod;
 import azkaban.db.DatabaseOperator;
 import azkaban.db.EncodingType;
 import azkaban.db.SQLTransaction;
@@ -71,9 +72,11 @@ public class ExecutionFlowDao {
 
     final String INSERT_EXECUTABLE_FLOW = "INSERT INTO execution_flows "
         + "(project_id, flow_id, version, status, submit_time, submit_user, update_time, "
-        + "use_executor, flow_priority, execution_source) values (?,?,?,?,?,?,?,?,?,?)";
+        + "use_executor, flow_priority, execution_source, dispatch_method) values (?,?,?,?,?,?,?,"
+        + "?,?,?,?)";
     final long submitTime = flow.getSubmitTime();
     final String executionSource = flow.getExecutionSource();
+    final DispatchMethod dispatchMethod = flow.getDispatchMethod();
 
     /**
      * Why we need a transaction to get last insert ID?
@@ -84,7 +87,8 @@ public class ExecutionFlowDao {
     final SQLTransaction<Long> insertAndGetLastID = transOperator -> {
       transOperator.update(INSERT_EXECUTABLE_FLOW, flow.getProjectId(),
           flow.getFlowId(), flow.getVersion(), flow.getStatus().getNumVal(),
-          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority, executionSource);
+          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority, executionSource
+          , dispatchMethod.getNumVal());
       transOperator.getConnection().commit();
       return transOperator.getLastInsertId();
     };
@@ -380,7 +384,8 @@ public class ExecutionFlowDao {
     }
   }
 
-  public int selectAndUpdateExecution(final int executorId, final boolean isActive)
+  public int selectAndUpdateExecution(final int executorId, final boolean isActive,
+      final DispatchMethod dispatchMethod)
       throws ExecutorManagerException {
     final String UPDATE_EXECUTION = "UPDATE execution_flows SET executor_id = ?, update_time = ?, status=? "
         + "where exec_id = ?";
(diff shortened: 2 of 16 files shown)


sshardool (commented):
Thanks a lot for looking into this. 
As part of containerized dispatch we mode modifications to both `ExecutorApiGateway` and `ExecutorApiClient`. Most of them related to the reverse proxy but fairly orthogonal  - including, TLS client support, generating prefix and making calls agains the reverse-proxy (ambassador) uri.
For executing any API calls against both the bare-metal executors as well as the FlowContainers we will likely need to have two separate instances of the `ExecutorApiGateway` with different configs. 
Wanted to confirm our plan for addressing it.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -123,7 +138,18 @@ public long getQueuedFlowSize() {
    */
   @Override
   public DispatchMethod getDispatchMethod() {
-    return DispatchMethod.CONTAINERIZED;
+    if (this.rampUp == 0) {
```
abhisheknath2011, line 141: DispatchMethod is very useful and it controlling the behaviour at the moment by turning off and turning on feature. DispatchMethod is currently specified through configs. Hence, dispatch method is initialized during web server startup. The containerization feature is turned off if dispath method is POLL. If dispatch method is set to CONTAINERIZED then only all the guava juice binding for containerization are initialized. As we are dynamically toggling between  POLL and CONTAINERIZED based on rampup we need to reload the bindings as well. In other words, we need to always intialized the bindings or look at different way of handling this.

azkaban-common/src/main/java/azkaban/DispatchMethod.java
```diff
@@ -22,10 +22,19 @@
...
+  POLL(1),
+  CONTAINERIZED(2);
   private static final Logger logger = LoggerFactory.getLogger(DispatchMethod.class);
+  private final int numVal;
+
+  DispatchMethod(final int numVal) {
+    this.numVal = numVal;
+  }
+
+  public int getNumVal() {
```
jakhani, line 35: Is there any need to add method which will return enum for given num value?
What is the significance of this num value? Can you please add it in comment?

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -84,7 +87,8 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
     final SQLTransaction<Long> insertAndGetLastID = transOperator -> {
       transOperator.update(INSERT_EXECUTABLE_FLOW, flow.getProjectId(),
           flow.getFlowId(), flow.getVersion(), flow.getStatus().getNumVal(),
-          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority, executionSource);
+          submitTime, flow.getSubmitUser(), submitTime, executorId, flowPriority, executionSource
+          , dispatchMethod.getNumVal());
```
jakhani, line 91: I assume you are keeping num value for the optimization that int will provide from database side instead of enum name itself.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
jakhani: +1. I didn't see any changes related to injection over here. How will implementation for POLL method be injected in case of ramp up?

azkaban-db/src/main/sql/create.containerization-tables-all.sql
```diff
@@ -107,3 +107,8 @@ CREATE TABLE IF NOT EXISTS version_set (
 -- TODO: Add the alter table script in the specific release
 -- Adding version_set_id column in execution_flows
 -- alter table execution_flows add column version_set_id INT default null;
+
+-- TODO: Add the alter table script in the specific release
+-- Adding dispatch_method column in execution_flows
+-- alter table execution_flows add column dispatch_method TINYINT default 1;
```
jakhani, line 113: Can you please add index for this field? It is in the where clause in select query so it will be useful to have it.

azkaban-db/src/main/sql/create.execution_flows.sql
```diff
@@ -15,6 +15,7 @@ CREATE TABLE execution_flows (
   use_executor INT                  DEFAULT NULL,
   flow_priority TINYINT    NOT NULL DEFAULT 5,
   execution_source VARCHAR(32)        DEFAULT NULL,
+  dispatch_method TINYINT        DEFAULT 1,
```
jakhani, line 18: Can you please add index for this field?

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
jakhani: Got it. And specifying dispatch method as POLL from executor side will take care of polling executions. Thank you for clarifying this.

azkaban-db/src/main/sql/create.containerization-tables-all.sql
aditya1105: done

azkaban-db/src/main/sql/create.execution_flows.sql
aditya1105: done

jakhani (approved):
Thank you for taking care of all the comments.

azkaban-db/src/main/sql/create.execution_flows.sql
juhoautio-rovio: Upgrade sql was not added for this change, but it should be (under https://github.com/azkaban/azkaban/tree/master/azkaban-db/src/main/sql).

Same for the added CREATE INDEX below.

PR #2781, opened 2021-03-02
Introduce a callback driven framework for tracking flow-containers

sshardool:
The commit introduces the following:
- Kubernetes watch API invocation for pods in a given namespace. Ref:
  https://kubernetes.io/docs/reference/using-api/api-concepts/
- State machine for generic singe-use pods (including flow-containers)
  based on the `Watch` events received from the kubernetes api-server
  Ref: Classes `AzPodStatus` and `AzPodStatusExtractor`
- Event based framework which allows multiple listeners to receive
  notification of state transitions.  Ref: Classes
  `AzPodStatusListener`, `RawPodWatchEventListener`, `AzPodStatusDriver`
- Addition of execution-id as a label within flow-container pods spec.
- Unit test which rely on a stream of `Watch` events stored in a file.

The framework will be used for implementing a number of features which
are directly tied to the status of the flow-container. These are some of
the upcoming changes:
- Listener for updating flow-status based on the flow-container pod
  state.
- Listener for calculating metrics, like time spend in each pod state.


azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
```diff
@@ -0,0 +1,347 @@
+package azkaban.executor.container.watch;
```
aditya1105, line 1: Nit: Copyright header

jakhani (commented):
Can we please have a call for this PR? It will be very useful to discuss it and review.

azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatus.java
```diff
@@ -0,0 +1,103 @@
...
+ * Note that the lifecycle state-machine currently assumes that PODs are single-use, such as for
+ * FlowContainers. It's possible to extend it to PODs more generally but that seems unnecessary,
+ * as in most of those cases monitoring Kubernetes ReplicaSets would be more useful.
+ *
+ */
+public enum AzPodStatus {
+  /**
+   * Default state used for the flow pods whose actual state is not known.
+   */
+  AZ_POD_UNSET,
```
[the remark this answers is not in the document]
sshardool: Yes, we will be receiving events for all states (except `POD_UNSET`). However, there is no single 'state' within the events that maps to the states introduced here. The rules for converting the state from a kubernetes watch event `AzPodStatus` is described are defined in `AzPodStatusExtractor`.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is intended for maintaining any relevant data from a Pod Watch event that was derived
+ * during the inference of {@link AzPodStatus}. The data here can be consumed by any downstream
+ * callbacks.
+ */
+public class AzPodStatusMetadata {
```
jakhani, line 34: Shall we rename this class to AzPodEventMetadata?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(AzPodStatusMetadata.class);
+
+  private final AzPodStatus azPodStatus;
+  private final String podName;
+  private final Watch.Response<V1Pod> podWatchEvent;
```
jakhani, line 39: We should revisit the java process memory for Web server when we are planning to ramp up containerization completely to make sure we are not bringing web server down.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+
+  /**
+   * Initialize and start the Pod watch procedure. The processing is expected to continue until a
+   * shutdown request is received. The method will handle any exceptions received during both (1)
+   * the execution of kubernetes API calls (2) processing of events received by the watch.
+   * This also includes re-initialization of the watch using Kuberentes watch API, as needed.
+   */
+  public void initializeAndStartPodWatch() {
```
jakhani, line 158: Shall it be thread safe?

azkaban-common/src/main/java/azkaban/executor/container/watch/WatchException.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package azkaban.executor.container.watch;
+
+/**
+ * Intended as a container for all the exceptions that can be thrown withing the kubernetes watch
+ * implementation.
+ * This can be used for 'wrapping' any checked exceptions used by the Watch listeners and is
+ * open for extension in case there is a need for a more specific exceptions.
+ */
+public class WatchException extends RuntimeException {
```
jakhani, line 24: Shall we rename it to AzkabanPodWatchException?

azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
```diff
@@ -0,0 +1,387 @@
...
+  private static class AzPodStausExtractingListener implements RawPodWatchEventListener {
+    @Override
+    public void onEvent(Response<V1Pod> watchEvent) {
+      logger.debug(String.format("%s : %s, %s, %s", watchEvent.type, watchEvent.object.getMetadata().getName(),
+          watchEvent.object.getStatus().getMessage(), watchEvent.object.getStatus().getPhase()));
+      AzPodStatus azPodStatus = AzPodStatusExtractor.azPodStatusFromEvent(watchEvent).getAzPodStatus();
+      logger.debug("AZ_POD_STATUS: " + azPodStatus);
+    }
+  }
+}
```
jakhani, line 387: Nit: New line in the end.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -0,0 +1,527 @@
...
+    this.podPhase = PodPhase.valueOf(v1PodStatus.getPhase());
+  }
+
+  /**
+   * Evaluate if event can be classified as status {@link AzPodStatus.AZ_POD_REQUESTED}
+   * @return true if the classification succeeds, false otherwise.
+   */
+  private boolean checkForAzPodRequested() {
+    // Scheduled conditions should either not be present or be false
+    if (scheduledConditionStatus.isPresent() &&
```
djaiswal83, line 232: nitpick : please use "this" keyword when using member variables.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -0,0 +1,527 @@
...
+    if (scheduledConditionStatus.get() != PodConditionStatus.True) {
+      logger.debug("PodScheduled false as scheduled condition is not true");
+      return false;
+    }
+    // Initialized condition is not present
+    if (!initializedCondition.isPresent()) {
+      logger.debug("PodScheduled true as initialized condition is not present");
+      return true;
+    }
+    // Initialized condition is not true
```
djaiswal83, line 260: initialized condition is true, right?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -0,0 +1,527 @@
...
+    }
+    logger.debug("AppFailed is true");
+    return true;
+  }
+
+  /**
+   * Return the {@link AzPodStatus} derived from given Pod watch event.
+   * @return
+   */
+  public AzPodStatus createAzPodStatus() {
```
djaiswal83, line 453: This is looking a bit inefficient to me. I dont have a solution but inorder to run checkForAzkPodAppFailure, in the worst case, all the conditions must be checked before.
Is there a way to avoid that?
Perhaps as a TODO for future?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+  public static final int SHUTDOW_TERMINATION_TIMEOUT_SECONDS = 5;
+  private final ExecutorService executor;
+  private final ImmutableMap<AzPodStatus, List<Consumer<AzPodStatusMetadata>>> listenerMap;
+
+  /**
+   * Create a new instance of {@link AzPodStatusDriver}.
+   */
+  @Inject
+  public AzPodStatusDriver() {
+    // Worth considering if we should get the thread-pool size from a config property instead.
```
djaiswal83, line 60: The thread-pool size must be configurable.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Register a new {@link AzPodStatusListener}.
+   * Note that the callbacks are submitted to the {@link ExecutorService} in the order interfaces
+   * were registered, but that does not make any guarantees about the sequence in which all the
+   * callbacks for a given event are invoked.
+   *
+   * @param listener
+   */
+  public void registerAzPodStatusListener(AzPodStatusListener listener) {
```
djaiswal83, line 85: using annotation @NonNull will be more helpful for input params as IntelliJ can help find a coding error beforehand compared to requireNoNull. Applies to any subsequent use for input params.

djaiswal83 (changes requested):
Thanks for implementing this feature.
Very well written code.
I have some comments, can you please address those?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Process all the callbacks for a given event. Lambdas corresponding to the callback method from
+   * every registered {@link AzPodStatusListener} are scheduled for execution on the {@code
+   * executor}
+   *
+   * @param podStatusMetadata
+   */
+  private void deliverCallbacksForEvent(AzPodStatusMetadata podStatusMetadata) {
+    listenerMap.get(podStatusMetadata.getAzPodStatus()).stream()
```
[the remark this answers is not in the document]
sshardool: The method `logPodWatchEvent` is already logging information for every incoming event, I am switching the log verbosity for it to `info`. 
In addition, as discussed, a more accurate metric for this will be the time spent by the callback tasks in the executor-service queue when a sufficient number of threads is not available. This will be added subsequently.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
sshardool: The name `AzPodStatusMetadata` was intended to convey the fact that this contains 'more' data associated with the simple enums defined in `AzPodStatus`. 
Let me know if you have a strong reason for the rename (I do like the sound of of `AzPodEventMetadata` as well).

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
sshardool: Good point. The additional memory usage has been considered as part of the this imlementation - as the individual Watch<V1Pod> events can be a couple of kbs in size. However, as part of this PR we will not be holding on to references to these events and even with few 100s (more than expected) of events 'in-flight' the additional usage should not exceed a couple of MBs. 
A subsequent PR will introduce logic for capturing at least parts of the events for every flow in progress, that will require a more careful analysis of the additional memory usage.

azkaban-common/src/main/java/azkaban/executor/container/watch/WatchException.java
sshardool: Makes sense, I will rename it to `AzkabanWatchException`, just in case we start watching objects other than pods.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -0,0 +1,247 @@
...
+   */
+  protected void initializePodWatch() throws ApiException {
+    try {
+      this.podWatch = Watch.createWatch(this.apiClient,
+          coreV1Api.listNamespacedPodCall(podWatchParams.getNamespace(),
+              "true",
+              false,
+              null,
+              null,
+              podWatchParams.getLabelSelector(),
```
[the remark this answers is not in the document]
sshardool: Good point, the label creation will be accounted for as part of adding the bindings and initializing the watch in a subsequent PR. 
However, the rest of the logic, including implementations of the `AzPodListener` will still be resilient enough to ignore any events which are are not from a 'flow-container-pod'

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatus.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  AZ_POD_INIT_CONTAINERS_RUNNING,
+
+  /**
+   * Application container execution started (for at least 1 app container)
+   */
+  AZ_POD_APP_CONTAINERS_STARTING,
+
+  /**
+   * All application containers have started, pod status is 'ready'.
```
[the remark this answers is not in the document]
sshardool: 'Ready' in this case is referring to an actual state (in the status array) returned by kubernetes and is more accurate.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
sshardool: The comment effectively asserts that 'initialized condition' should be false. Which is what we want in this case, otherwise on the next line if the initialized-condition check return true, the function will return early with the result false.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
sshardool: Actually, I had the opposite in mind. The `checkAz*` methods are currently implemented in a manner where for every raw `Watch<V1Pod>` event exactly one of the `checkAz...` methods should return true. 
Running the raw event through all the methods will serve as a good validation for this logic (e.g we can add a warning in case more than 1 method returns true).
All of these methods will run very quickly, as they include simple point lookup of variables and in some cases traversals of arrays with 5-10 elements.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
sshardool: I did not find a completion for @nonull, were you referring to `javax.annotation.Nonnull` perhaps ?
Good point, however I would prefer to retain the explicit call to `requireNonNull` as that allows for a custom exception message (and also conforms better to the rest of the code).

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+ * Events received by this directly map the states in {@link AzPodStatus} and for every event all
+ * the callbacks registered for the particular event type are 'invoked'
+ *
+ * Invocation of the callbacks is done via pool of threads different from the one invoking this
+ * {@code onEvent} interface method of this class. This is done as a typical implementation of the
+ * {@link AzPodStatusListener} interface can include several blocking calls, for example for
+ * database status updates.
+ */
+@Singleton
+public class AzPodStatusDriver implements RawPodWatchEventListener {
```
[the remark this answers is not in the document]
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+        .forEach(callback -> executor.execute(() -> callback.accept(podStatusMetadata)));
+  }
+
+  @Override
+  public void onEvent(Watch.Response<V1Pod> watchEvent) {
+    // Technically, logging the pod watch event can also be performed as part of a callback.
+    // For now the logging is inline to ensure the the event is logged before any corresponding
+    // callbacks and are invoked.
+    try {
+      AzPodStatusMetadata azPodStatusMetadata = AzPodStatusExtractor.azPodStatusFromEvent(watchEvent);
```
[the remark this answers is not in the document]
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
```diff
@@ -0,0 +1,143 @@
...
+ * {@code onEvent} interface method of this class. This is done as a typical implementation of the
+ * {@link AzPodStatusListener} interface can include several blocking calls, for example for
+ * database status updates.
+ */
+@Singleton
+public class AzPodStatusDriver implements RawPodWatchEventListener {
+  private static final Logger logger = LoggerFactory.getLogger(AzPodStatusDriver.class);
+
+  private static final int THREAD_POOL_SIZE = 4;
+  public static final int SHUTDOW_TERMINATION_TIMEOUT_SECONDS = 5;
```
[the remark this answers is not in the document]
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
sshardool: Fixed. This method was made private and synchronization with additional checks added in `launchPodWatch`

azkaban-common/src/main/java/azkaban/executor/container/watch/WatchException.java
sshardool: Done.

azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatus.java
```diff
@@ -0,0 +1,103 @@
...
+ * The Enum represents the different stages of the pod lifecycle.
+ * While the Kubernetes API can provide very granular information about the current states
+ * of various aspects of a POD, it doesn't quite provide any state-machine like representation of
+ * POD life-cycle at a good enough granularity for use within Azkaban. <br>
+ *   https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+ *
+ * For example, a Pod Phase is very coarse and doesn't convey information regarding
+ * pod-scheduling and init-container completions.
+ * On the other hand Pod Conditions array along with the container statuses provides detailed
+ * information down the states and timestamps on individual containers but there is no direct
```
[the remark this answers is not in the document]
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -0,0 +1,247 @@
...
+    logger.info("Requesting shutdown for kubernetes watch");
+    watchRunner.interrupt();
+    return notAlreadyRequested;
+  }
+
+  /**
+   * Parameters used for setting up the Pod watch.
+   */
+  public static class PodWatchParams {
+    private final static int DEFAULT_RESET_DELAY_MILLIS = 10 * 1000; //30 seconds
```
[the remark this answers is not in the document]
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
djaiswal83: In that case, how about just precompute the results in constructor and cache them for access?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
djaiswal83: Yes, the javax one is what I meant. The annotation is superior in this case. It finds the problem at compile time rather than execution time thus prevents bugs preemptively in most cases.
requireNonNull will find the bugs only after the code is deployed by throwing NPE.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDrivingListener.java
```diff
@@ -44,21 +46,26 @@
...
 
   /**
-   * Create a new instance of {@link AzPodStatusDriver}.
+   * Create a new instance of {@link AzPodStatusDrivingListener}.
    */
   @Inject
-  public AzPodStatusDriver() {
-    // Worth considering if we should get the thread-pool size from a config property instead.
-    executor = Executors.newFixedThreadPool(THREAD_POOL_SIZE,
+  public AzPodStatusDrivingListener(Props azkProps) {
```
djaiswal83, line 62: Using annotation @NonNull would be better.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -186,13 +187,21 @@ public void initializeAndStartPodWatch() {
    *
    * @return
    */
-  public Thread launchPodWatch() {
+  public synchronized Thread launchPodWatch() {
```
djaiswal83, line 190: Can you please add a comment that the caller must handle the exception otherwise it could cause unexpected failures.

djaiswal83 (changes requested):
Thanks for addressing the comments. I have couple of minor comments.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
sshardool: Dug a little more into it and it appears replacing `Objects.requireNonNull` with `javax.annotation.Nonnull` would be incorrect.
`requireNonNull` will force a runtime nullability check whereas the annotation `javax.annotation.Nonnull` will only introduce a compile time check that doesn't necessarily extend to runtime. 
An alternative for this is the annotation from project Lombok `lombok.NonNull` which does add a runtime nullability check as well: https://projectlombok.org/features/NonNull
However, I don't think introducing Lombok just for the nullability checks is a good idea. 
Additional information: https://stackoverflow.com/questions/48103792/what-is-the-difference-between-nonnull-and-objects-requirenonnull
Please let me know if you disagree.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
sshardool: That's definitely worth considering. How about I revisit this part in a subsequent PR? (will create a task for this).

djaiswal83: Sure, can you please create a ticket to track this internally and resolve this issue?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDriver.java
djaiswal83: Thanks for digging this up. The same article says that they are rather complementary. I would recommend using both in this case to avoid programming errors.
Something, I also intend to do in my code in the future.

sshardool: The utility of the `javax.annotation.Nonnull` annotation (which will only do a static check) in addition to the `requireNonNull` check is a little questionable. 
Discussed offline and decided that introducing another `Nonnull` check needs more thought and should be done with a more comprehensive review of the available options and modifying the coding guidelines accordingly.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusDrivingListener.java
sshardool: Discussed in a previous comment.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
sshardool: Added an internal tracking ticket.

djaiswal83 (approved):
Thanks for addressing comments. The code looks good.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
jakhani: For future extensibility, I feel event metadata will be little generic. No hard core reason though.

jakhani (approved):

jakhani:
Thank you for looking into all the comments.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
sshardool: Retaining the current name in that case, will revisit in a subsequent PR if this needs to be changed. Thanks!

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
hqin2016: @sshardool if >1 check methods return true, which status should this method return?

PR #2782, opened 2021-03-05
Resolve symbolic links for jar files added to HadoopSecurityManagerClassloader

haibchen:
Jar files added to HadoopSecurityManagerClassloader as search paths could be located under symbolic links.

For example, let's say there are two jars, `A-0.0.1.jar` and `B-0.0.1.jar` under a symbolic link `/export/apps/hadoop/latest`
When these jars are added, effectively two paths are added to HadoopSecurityManagerClassloader,
`/export/apps/hadoop/latest/A-0.0.1.jar` and `/export/apps/hadoop/latest/B-0.0.1.jar`

When the symbolic link moves, HadoopSecurityManagerClassloader breaks. Let's say the symbolic link `/export/apps/hadoop/latest` is updated to point to a newer version of hadoop that contains newer version of A and B, `A-0.0.2.jar` and `B-0.0.2.jar` respectively. Once the symlink is updated, there are no longer `/export/apps/hadoop/latest/A-0.0.1.jar` and `/export/apps/hadoop/latest/B-0.0.1.jar`, but existing HadoopSecurityManagerClassloader instances continue to load classes from them, therefore may fail all jobs launch post the Hadoop upgrade.

Note an instance of HadoopSecurityManagerClassloader is created per Hadoop cluster, and shared by every job that's submitted to that cluster.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/cluster/Cluster.java
+++ b/azkaban-common/src/main/java/azkaban/cluster/Cluster.java
@@ -106,7 +106,7 @@ public class Cluster {
           final List<URL> clusterUrls;
           try {
             clusterUrls = getClusterComponentURLs(hadoopSecurityManagerDependencyComponents);
-          } catch (IOException e) {
+          } catch (final IOException e) {
             throw new IllegalArgumentException(
                 String.format("Invalid dependency components for " +
                     HadoopSecurityManagerClassLoader.class.getName() + " of cluster %s", clusterId));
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/cluster/Cluster.java
```diff
@@ -176,23 +177,24 @@ public int hashCode() {
...
+      throws IOException {
     final List<URL> resources = new ArrayList<>();
 
     if (Strings.isNullOrEmpty(clusterLibraryPath)) {
       return resources;
     }
 
     for (String path : clusterLibraryPath.split(PATH_DELIMITER)) {
-      final File file = new File(path);
+      // strip the trailing * character from the path, if any
```
djaiswal83, line 188: Can you please add a few examples of path in comment?

djaiswal83 (changes requested):
The logic goods to me. Please address the comments below.

djaiswal83 (approved):

hungj (approved):
LGTM. Thanks for addressing this!

PR #2785, opened 2021-03-10
ApiGateway support for ramp-up mechanism from POLL to CONTANERIZED dispatch

aditya1105:
Ramp-up mechanism is introduced in https://github.com/azkaban/azkaban/pull/2779 allowing some flows to be dispatched as Containerized and others as POLL to have a mechanism of ramping the containerization on the already running POLL based Azkaban cluster.

Hence web server, depending on the execution, may need to send a request to the Container via reverse proxy or to the executor server. Column "flow_data" in the execution_flows is a blob containing information regarding flow execution and it will also contain the dispatch_method for that execution. Based on the dispatch_method derived from "flow_data", we can send the request to the appropriate server.

Key changes:
azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java

This is to make sure that both ExecutorApiGateway and ExecutorApiClient take into account the dispatch_method for every call, and creates & use the appropriate Http/Https client.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/DispatchMethod.java
+++ b/azkaban-common/src/main/java/azkaban/DispatchMethod.java
@@ -36,6 +36,15 @@ public enum DispatchMethod {
     return this.numVal;
   }
 
+  public static DispatchMethod fromNumVal(final int numVal) {
+    for (DispatchMethod dispatchMethod : DispatchMethod.values()) {
+      if (dispatchMethod.getNumVal() == numVal) {
+        return dispatchMethod;
+      }
+    }
+    throw new IllegalArgumentException("No Dispatch Method corresponding to num value " + numVal);
+  }
+
   public static DispatchMethod getDispatchMethod(String value) {
     try {
       logger.info("Value of dispatch method is : " + value);
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -52,6 +52,7 @@ public class ExecutableFlow extends ExecutableFlowBase {
   public static final String IS_LOCKED_PARAM = "isLocked";
   public static final String FLOW_LOCK_ERROR_MESSAGE_PARAM = "flowLockErrorMessage";
   public static final String EXECUTION_SOURCE = "executionSource";
+  public static final String FLOW_DISPATCH_METHOD = "dispatch_method";
 
   private final HashSet<String> proxyUsers = new HashSet<>();
   private int executionId = -1;
@@ -306,6 +307,7 @@ public class ExecutableFlow extends ExecutableFlowBase {
 
     flowObj.put(IS_LOCKED_PARAM, this.isLocked);
     flowObj.put(FLOW_LOCK_ERROR_MESSAGE_PARAM, this.flowLockErrorMessage);
+    flowObj.put(FLOW_DISPATCH_METHOD, getDispatchMethod().getNumVal());
 
     return flowObj;
   }
@@ -359,6 +361,9 @@ public class ExecutableFlow extends ExecutableFlowBase {
 
     this.setLocked(flowObj.getBool(IS_LOCKED_PARAM, false));
     this.setFlowLockErrorMessage(flowObj.getString(FLOW_LOCK_ERROR_MESSAGE_PARAM, null));
+    // Dispatch Method default is POLL
+    this.setDispatchMethod(DispatchMethod.fromNumVal(flowObj.getInt(FLOW_DISPATCH_METHOD,
+        DispatchMethod.POLL.getNumVal())));
   }
 
   @Override
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
@@ -20,6 +20,7 @@ import static azkaban.Constants.ConfigurationKeys.EXECUTOR_CLIENT_TLS_ENABLED;
 import static com.google.common.base.Preconditions.checkState;
 
 import azkaban.Constants.ConfigurationKeys;
+import azkaban.DispatchMethod;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.RestfulApiClient;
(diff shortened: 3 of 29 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -123,14 +128,15 @@ public ExecutorApiClient(final Props azkProps) {
    * @return the URI built from the inputs.
    */
   public URI buildExecutorUri(final String host, final int port, final String path,
-      final boolean isHttp, final Pair<String, String>... params) throws IOException {
-    if (!isReverseProxyEnabled) {
+                              final boolean isHttp, final DispatchMethod dispatchMethod, final Pair<String, String>... params) throws IOException {
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/DispatchMethod.java
```diff
@@ -36,6 +36,18 @@ public int getNumVal() {
     return this.numVal;
   }
 
+  public static DispatchMethod fromNumVal(final int numVal) {
+    if (numVal == 0) {
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -55,61 +55,50 @@
...
       final Executor executor, final String action) throws ExecutorManagerException {
     return callWithExecutionId(executor.getHost(), executor.getPort(), action,
-        exflow.getExecutionId(), null, (Pair<String, String>[]) null);
+        exflow.getExecutionId(), null, exflow.getDispatchMethod(), (Pair<String, String>[]) null);
   }
 
   Map<String, Object> callWithReference(final ExecutionReference ref, final String action,
       final Pair<String, String>... params) throws ExecutorManagerException {
-    final Executor executor = (isReverseProxyEnabled ? defaultEmptyExecutor : ref.getExecutor().get());
+    final Executor executor = (ref.getDispatchMethod().getNumVal() == 2 ? defaultEmptyExecutor : ref.getExecutor().get());
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/ExecutorApiGateway.java
```diff
@@ -55,61 +55,50 @@
...
-        ref.getExecId(), user, params);
+        ref.getExecId(), user, ref.getDispatchMethod(), params);
   }
 
   @VisibleForTesting
-  String createExecutionPath(final Optional<Integer> executionId) throws ExecutorManagerException {
-    if (!isReverseProxyEnabled) {
-      return "/" + executionResourceName;
+  String createExecutionPath(final Optional<Integer> executionId, DispatchMethod dispatchMethod) throws ExecutorManagerException {
+    if (null == dispatchMethod || dispatchMethod.getNumVal() != 2) {
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/DispatchMethod.java
```diff
@@ -36,6 +36,15 @@ public int getNumVal() {
     return this.numVal;
   }
 
+  public static DispatchMethod fromNumVal(final int numVal) {
+    for (DispatchMethod dispatchMethod : DispatchMethod.values()) {
+      if (dispatchMethod.getNumVal() == numVal) {
+        return dispatchMethod;
+      }
+    }
+    throw new IllegalArgumentException("No Dispatch Method corresponding to int value " + numVal);
```
jakhani, line 45: Nit: corresponding to num value.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -359,6 +361,8 @@ public void fillExecutableFromMapObject(
 
     this.setLocked(flowObj.getBool(IS_LOCKED_PARAM, false));
     this.setFlowLockErrorMessage(flowObj.getString(FLOW_LOCK_ERROR_MESSAGE_PARAM, null));
+    // Dispatch Method default is POLL i.e. 1
+    this.setDispatchMethod(DispatchMethod.fromNumVal(flowObj.getInt(FLOW_DISPATCH_METHOD, 1)));
```
jakhani, line 365: For default, you DispatchMedhod.POLL.getNumValue or something. You should not use number any where other than enum. Better to use get method. It will keep it clean.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -174,16 +182,53 @@ public boolean isReverseProxyEnabled() {
...
-      return super.createHttpClient();
-    }
+  protected CloseableHttpClient createHttpsClient() {
     final HttpClientBuilder httpClientBuilder = HttpClients.custom()
         .setSSLSocketFactory(this.tlsSocketFactory);
     return httpClientBuilder.build();
   }
 
+  /**
+   * function to perform a Post https request.
```
jakhani, line 192: Nit: Method

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -174,16 +182,53 @@ public boolean isReverseProxyEnabled() {
...
+
+    final HttpPost post = new HttpPost(uri);
+    return this.sendAndReturnHttps(completeRequest(post, params));
+  }
+
+  public String doPost(final URI uri, final DispatchMethod dispatchMethod, final List<Pair<String, String>> params)
+          throws IOException {
+    // If in future tls support is added for POLL based model, then following condition
+    // can be simplified
+    if (isTlsEnabled && null != dispatchMethod && dispatchMethod == DispatchMethod.CONTAINERIZED) {
```
jakhani, line 215: For any dispatch method, if Tls is enabled then use https and if not then http. Shouldn't that be the only check?

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
aditya1105: done

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
aditya1105: So we have ramp-up when containerization is enabled which makes different executions have different dispatch methods. TLS is enabled only for containerized execution while the property to enable TLS will continue to be true, we still won't be using TLS for POLL-based executions until the TLS is also supported for POLL-based dispatch. Hence we needed extra check.

jakhani: Oh got it now. Thank you for clarifying that.

jakhani (approved):

PR #2787, opened 2021-03-11
Adding props instance variable to AbstractHadoopSecurityManager and logging the Exception to logger if any...

hchatriki-linkedin:
Adding props instance variable to AbstractHadoopSecurityManager and logging the Exception to logger if any...

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
@@ -89,6 +89,7 @@ public abstract class AbstractHadoopSecurityManager extends HadoopSecurityManage
   protected final ConcurrentMap<String, UserGroupInformation> userUgiMap;
   protected boolean shouldProxy;
   protected boolean securityEnabled;
+  protected final Props props;
   public static final String CHOWN = "chown";
   public static final String CHMOD = "chmod";
 
@@ -97,6 +98,7 @@ public abstract class AbstractHadoopSecurityManager extends HadoopSecurityManage
       .getLogger(AbstractHadoopSecurityManager.class);
 
   public AbstractHadoopSecurityManager(final Props props) {
+    this.props = props;
     this.executeAsUser = new ExecuteAsUser(props.getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
     this.conf = new Configuration();
     // Disable yyFileSystem Cache for HadoopSecurityManager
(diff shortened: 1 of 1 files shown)


jakhani (approved):

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -89,6 +89,7 @@
   protected final ConcurrentMap<String, UserGroupInformation> userUgiMap;
   protected boolean shouldProxy;
   protected boolean securityEnabled;
+  protected final Props props;
```
li-ygerchikov, line 92: Please remind me: why did we decide to put it here and not in either `KSudoHadoopSecurityManager` (bottom of the hierarchy, would not require any changes to OSS Azkaban) *or* `HadoopSecurityManager` (all the way to the top of the class hierarchy)? Why is this in the middle?

li-ygerchikov: @jakhani any thoughts on whether we can add props to `KSudoHadoopSecurityManager` itself instead? Since it's not really needed or used anywhere in `HadoopSecurityManager` or `AbstractHadoopSecurityManager`...

jakhani: I thought that too. But then I realized that it is not used anywhere because the props instance variable was not there. Now if it will be there in AbstractHadoopSecurityManager, we can optimize couple of things in future and use it in other methods. Please let me know if you disagree.

li-ygerchikov: @jakhani I see. Thanks for the explanation, the intent makes sense.
While I agree in general that we could refactor `AbstractHadoopSecurityManager` to streamline things, I feel that it may be a separate effort. (Will it not also impact the base class, `HadoopSecurityManager`, and potentially any other implementation of HSM "out there in the wild" that we may not be even aware of?) 
If we are not refactoring it now or soon, then maybe we should not add properties there, but keep changes to a minimum for now, and perhaps file a ticket for that future work. 
I can go either way, it just bothers me a bit that the code won't make sense in the interim... and the interim may be indefinitely long :) .

jakhani: Sounds good. Let's not do it as a part of this change then. We can take it separately. 
Can you please get Harsha onboarded with your thought and appropriate change as well?

li-ygerchikov: Thank you! All clear now. 
Will do!

li-ygerchikov:
Rejected, per the conversation with @jakhani above.

PR #2792, opened 2021-03-16
Adding dependency for javax.el as it needs to be upgraded to 3.0 for hibernateValidator dependency

jakhani:


abhisheknath2011 (approved):
Thanks for working on this. I have verified with these changes sometime back. Hopefully the build should go through this time.

PR #2793, opened 2021-03-17
Flow deterministic ramp-ups for imagetype and containerization

aditya1105:
Current ramp-ups use the random generator to determine the version, which can lead to different executions of the same flow to have different versions.
This implementation uses flow name, compute murmur hash, map it to an integer value between 1 to 100, and determine the ramp-up version.
Implementation of hash is in ContainerImplUtils.java

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
@@ -19,8 +19,10 @@ import azkaban.executor.ExecutableFlow;
 import azkaban.executor.ExecutableFlowBase;
 import azkaban.executor.ExecutableNode;
 import azkaban.executor.ExecutorManagerException;
+import java.nio.charset.StandardCharsets;
 import java.util.Set;
 import java.util.TreeSet;
+import org.apache.commons.codec.digest.MurmurHash3;
 
 /**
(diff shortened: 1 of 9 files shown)


jakhani (approved):

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -61,4 +63,18 @@ private static void populateJobTypeForFlow(final ExecutableNode node, Set<String
...
+   * @param flow
+   * @return
+   */
+  public static int getFlowNameHashValMapping(ExecutableFlow flow) {
+    // Flow name is <projectName>.<flowId>
+    String flowName = flow.getFlowName();
+    byte[] flowNameBytes = flowName.getBytes(StandardCharsets.UTF_8);
+    // To be utilized for flow deterministic ramp-up
+    int flowNameHashVal = MurmurHash3.hash32x86(flowNameBytes);
+    return flowNameHashVal % 100 + 1;
```
sshardool, line 78: Should it be something like `(double)flowNameHashVal / MAX_HASH_VALUE` 
where MAX_HASH_VALUE will be UINT32_MAX for 32 bit hash. 
This will have the advantage of using the entire hashed value without relying on only some of the least-significant-bits and should not have any extra cost.

PR #2797, opened 2021-03-18
Making ExecJettyServer TLS aware

aditya1105:
Refactored the getSocketConnector() and getSslSocketConnector() from WebServerProvider class to JettyServerUtils class so that it can be reused in ExecJettyServerModule

Added the ExecJettyServerModuleTest to check if the correct props are set-up when SSL is enabled/disabled.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2021 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package azkaban.server;
+
+import azkaban.utils.Props;
+import java.util.List;
+import org.mortbay.jetty.bio.SocketConnector;
+import org.mortbay.jetty.security.SslSocketConnector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility class for adding creating connectors, configurations, etc. to the Jetty Server.
+ */
+public class JettyServerUtils {
+
+  private static final int DEFAULT_HEADER_BUFFER_SIZE = 10 * 1024 * 1024;
+  private static final Logger logger = LoggerFactory.getLogger(JettyServerUtils.class);
+  private static final String JETTY_KEYSTORE = "jetty.keystore";
+  private static final String JETTY_PASSWORD = "jetty.password";
+  private static final String JETTY_KEYPASSWORD = "jetty.keypassword";
+  private static final String JETTY_TRUSTSTORE = "jetty.truststore";
+  private static final String JETTY_TRUSTPASSWORD = "jetty.trustpassword";
+
+  private JettyServerUtils() {
+    // Not to be instantiated
+  }
+
+  /**
+   * @param sslPortNumber Port number to bind for the socket
+   * @param props         Azkaban properties containing configurations for Jetty Server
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package azkaban.server;
+
+import azkaban.utils.Props;
+import java.util.List;
+import org.apache.log4j.Logger;
+import org.mortbay.jetty.bio.SocketConnector;
+import org.mortbay.jetty.security.SslSocketConnector;
+
+public class JettyServerUtils {
```
jakhani, line 24: Can you please add class documentation?

azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private static final int DEFAULT_HEADER_BUFFER_SIZE = 10 * 1024 * 1024;
+  private static final Logger logger = Logger.getLogger(JettyServerUtils.class);
```
jakhani, line 27: Will we be able to use slf4j logger here?

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
```diff
@@ -273,7 +273,13 @@ private void startReportingExecMetrics() {
 
   private void initActive() throws ExecutorManagerException {
     final Executor executor;
-    final int port = this.props.getInt(ConfigurationKeys.EXECUTOR_PORT, -1);
+    final int port;
+    final boolean useSsl = props.getBoolean(ConfigurationKeys.JETTY_USE_SSL, true);
+    if (useSsl) {
+      port = this.props.getInt(ConfigurationKeys.EXECUTOR_SSL_PORT, -1);
```
jakhani, line 279: In above class, default port value is 0. Here it is -1. Can you please make it consistent?

azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
aditya1105: Done. Thanks.

azkaban-common/src/main/java/azkaban/server/JettyServerUtils.java
aditya1105: Done. Utilized org.slf4j.LoggerFactory#getLogger(java.lang.Class<?>)

azkaban-exec-server/src/main/java/azkaban/execapp/AzkabanExecutorServer.java
aditya1105: So the default port value of 0 is from the legacy code, where port 0 implies random selection of a port, which is what is used in the present executor-server. 
As -1 can never be a port, here it is to imply that the port is randomly selected and hence it can't be activated from within the code. Hence executor is activated using the ajax APIs. This is what we currently do in the executor-server deployment.

jakhani (approved):

azkaban-exec-server/src/test/java/azkaban/common/ExecJettyServerModuleTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    final ExecutorApiClient tlsEnabledClient = new ExecutorApiClient(clientProps);
+    final String postResponse = tlsEnabledClient
+        .doPost(new URI(SimpleServlet.TLS_ENABLED_URI), DispatchMethod.CONTAINERIZED,
+            null);
+    Assert.assertEquals(ExecJettyServerModuleTest.SimpleServlet.POST_RESPONSE_STRING, postResponse);
+    jettyServer.stop();
+  }
+
+  private static class SimpleServlet extends HttpServlet {
```
sshardool, line 109: Maybe be make this class public in `ExecutorApiClient` and use it directly ?

sshardool (commented):
Thanks for taking this up, changes look good apart from a minor comment.

azkaban-exec-server/src/test/java/azkaban/common/ExecJettyServerModuleTest.java
aditya1105: Done.

sshardool (approved):
Thanks for taking care of the comments.

jakhani (approved):

PR #2801, opened 2021-03-19
Adding a way to setup env variables for a pod which can be honored from pod side.

jakhani:
If there are any flow parameters provided with prefix of "_pod.env.var._" then those will be set as environment variable for the Pod.


aditya1105 (approved):

djaiswal83 (approved):
Code looks good to me.
Thanks for adding Javadocs and tests.

abhisheknath2011 (approved):

PR #2803, opened 2021-03-19
Allow Hadoop to be configured with FileSystem implementations not using the 'hdfs://' scheme.

ypadron-in:
Previously, Azkaban always expected the HDFS scheme to be available in a Hadoop installation, so it would choose a default HDFS configuration when it didn't find the corresponding (HDFS specific) properties in the Hadoop configuration files. This is OK as long as a HDFS file system is available (possibly alongside others), but causes problems when it's not.

To fix this I'm removing the default Hadoop configurations provided by Azkaban, since, in practice, HDFS is usually and should be configured in Hadoop's config files. Although this is not a backward compatible change and issues may arise for those with Hadoop configurations not defining a file system properly this should be very rare.

Also, this change includes refactoring of the HadoopModule class (use of constants in the named bindings
 and removal of Props object from constructor method) but no changes in functionality are expected from that.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/HadoopModule.java
+++ b/azkaban-common/src/main/java/azkaban/HadoopModule.java
@@ -24,7 +24,6 @@ import static java.util.Objects.requireNonNull;
 import azkaban.cachedhttpfilesystem.CachedHttpFileSystem;
 import azkaban.spi.AzkabanException;
 import azkaban.storage.AbstractHdfsAuth;
-import azkaban.utils.Props;
 import azkaban.utils.Utils;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/HadoopModule.java
```diff
@@ -47,79 +46,72 @@
  * needed.
  */
 public class HadoopModule extends AbstractModule {
-  private static final String CHTTP_SCHEME = "chttp";
-  private static final String LOCAL_SCHEME = "file";
-  private static final String HDFS_SCHEME = "hdfs";
 
-  private static final String HDFS_FS_IMPL_CONFIG = "fs.hdfs.impl";
```
li-ygerchikov, line 54: Do these two become externally configured instead of hardcoded -- is this the idea? Maybe a quick comment somewhere describing the essence of the change would help. Otherwise there are a lot of formatting changes, and it's not easy to see what's actually going on. Thanks!

li-ygerchikov (approved):

azkaban-common/src/main/java/azkaban/HadoopModule.java
ypadron-in: Sorry, I forgot to add a description after pushing the code changes.

li-ygerchikov (approved):
Thanks for the description!

PR #2805, opened 2021-03-22
Implemented callback mechanism in FlowContainer.

djaiswal83:
Implemented callback mechanism in FlowContainer.
Use log4j instead of slf4j in FlowContainer.
Added ServerUtils class to refactor common code across AzkabanExecutorServer and FlowContainer.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
+++ b/azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
@@ -17,9 +17,12 @@
 
 package azkaban.common;
 
+import static java.util.Objects.requireNonNull;
+
 import azkaban.Constants;
 import azkaban.execapp.event.JobCallbackManager;
 import azkaban.utils.Props;
+import javax.annotation.Nonnull;
 import org.apache.log4j.Logger;
 
 /**
@@ -32,7 +35,9 @@ public class ServerUtils {
    * @param logger : the logger object of calling class.
    * @param props : Azkaban properties
    */
-  public static void configureJobCallback(final Logger logger, final Props props) {
+  public static void configureJobCallback(@Nonnull final Logger logger, @Nonnull final Props props) {
+    requireNonNull(logger, "Logger must not be null");
+    requireNonNull(props, "Properties can't be null");
     final boolean jobCallbackEnabled =
             props.getBoolean(Constants.ConfigurationKeys.AZKABAN_EXECUTOR_JOBCALLBACK_ENABLED, true);
 
--- a/azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
+++ b/azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2021 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+
+package azkaban.common;
+
+import azkaban.Constants;
+import azkaban.execapp.event.JobCallbackManager;
+import azkaban.utils.Props;
+import org.apache.log4j.Logger;
+
+/**
+ * Common utility methods for Azkaban Executor and Flow Container
+ */
+public class ServerUtils {
+
+  /**
+   * Method to initialize jobcallback manager if it is enabled.
+   * @param logger : the logger object of calling class.
+   * @param props : Azkaban properties
+   */
+  public static void configureJobCallback(final Logger logger, final Props props) {
+    final boolean jobCallbackEnabled =
+            props.getBoolean(Constants.ConfigurationKeys.AZKABAN_EXECUTOR_JOBCALLBACK_ENABLED, true);
+
+    logger.info("Job callback enabled? " + jobCallbackEnabled);
+
+    if (jobCallbackEnabled) {
+      JobCallbackManager.initialize(props);
+    }
+  }
+}
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -690,10 +690,19 @@ public class FlowContainer implements IMBeanRegistrable {
   }
 
   /**
+   * Close the MBeans
+   */
+  @VisibleForTesting
+  void closeMBeans() {
+    this.mBeanRegistrationManager.closeMBeans();
+  }
+
+  /**
    * Shutdown the Container. This shuts down the ExecutorService which runs the flow execution as
    * well as JettyServer.
    */
-  private void shutdown() {
+  @VisibleForTesting
+  void shutdown() {
     logger.info("Shutting down the pod");
     final int execId = this.flowRunner.getExecutionId();
     while (!this.flowFuture.isDone()) {
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -25,11 +25,13 @@ import azkaban.AzkabanCommonModule;
 import azkaban.Constants;
 import azkaban.Constants.PluginManager;
 import azkaban.common.ExecJettyServerModule;
+import azkaban.common.ServerUtils;
 import azkaban.execapp.AbstractFlowPreparer;
 import azkaban.execapp.AzkabanExecutorServer;
 import azkaban.execapp.ExecMetrics;
 import azkaban.execapp.FlowRunner;
 import azkaban.execapp.event.FlowWatcher;
+import azkaban.execapp.event.JobCallbackManager;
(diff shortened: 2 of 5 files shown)


azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
```diff
@@ -0,0 +1,45 @@
...
+
+import azkaban.Constants;
+import azkaban.execapp.event.JobCallbackManager;
+import azkaban.utils.Props;
+import org.apache.log4j.Logger;
+
+/**
+ * Common utility methods for Azkaban Executor and Flow Container
+ */
+public class ServerUtils {
```
jakhani, line 28: Shall we keep this class in azkaban-common module?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -437,12 +451,12 @@ public LogData readFlowLogs(final int execId, final int startByte, final int len
...
         return FileIOUtils.readUtf8File(logFile, startByte, length);
       } else {
-        logger.warn("Flow log file does not exist for flow execId: {}", execId);
+        logger.warn(String.format("Flow log file does not exist for flow execId: %d", execId));
         throw new ExecutorManagerException("Flow log file does not exist.");
       }
     } catch (final IOException e) {
-      logger.warn("IOException while trying to read flow log file for flow execId: {}",
-          execId);
+      logger.warn(String.format("IOException while trying to read flow log file for flow execId: %d",
```
jakhani, line 458: logger.error
For every place where logger.warn is there in catch.

jakhani (commented):
I am glad to know that you were able to test it. Can you please add unit test for this?

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
djaiswal83: Since this class is intended for execution side server utils, I put it in common of exec-server. I suppose we can make it a common class for web,exec and FlowContainer.
I do not have any preference, if you think that it should be in azkaban-common, I can move it there.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: These were just modified to use log4j for the purpose of this PR. @sshardool did this part, would like to get his input before I make the change.
Perhaps, can be done in separate PR?

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
abhisheknath2011, line 28: Shall we rename this class to CallbackUtils as the main purpose of this class is to configure callbacks?

abhisheknath2011 (commented):
One general question regarding callback. Whenever flow container finishes or in case of job success or failure how this callback mechanism will be invoked.

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
jakhani: Sure. We can address in separate PR.
^^ @sshardool

djaiswal83: No need to do it in separate PR, I can do it in this PR only.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: I think this is the comment you intended to reply :)

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
djaiswal83: I feel that in future it could hold many common functionalities that is why kept it generic enough.

djaiswal83:
> One general question regarding callback. Whenever flow container finishes or in case of job success or failure how this callback mechanism will be invoked.

Thanks for asking the question.
These callback gets added to FlowRunner and JobRunner as listeners. Every time there is a status change, an event is emitted, the listeners act based on those events.

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
```diff
@@ -0,0 +1,45 @@
...
+ * the License.
+ */
+
+
+package azkaban.common;
+
+import azkaban.Constants;
+import azkaban.execapp.event.JobCallbackManager;
+import azkaban.utils.Props;
+import org.apache.log4j.Logger;
```
sshardool, line 23: Should it be slf4j ?

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
```diff
@@ -0,0 +1,45 @@
...
+ * Common utility methods for Azkaban Executor and Flow Container
+ */
+public class ServerUtils {
+
+  /**
+   * Method to initialize jobcallback manager if it is enabled.
+   * @param logger : the logger object of calling class.
+   * @param props : Azkaban properties
+   */
+  public static void configureJobCallback(final Logger logger, final Props props) {
```
sshardool, line 35: Do we need to check for null values for incoming params?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -79,11 +84,10 @@
 import javax.inject.Named;
 import javax.inject.Singleton;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.log4j.Logger;
```
sshardool, line 87: Slf4j ?

sshardool (commented):
Thanks a lot for making this change. Have a few questions ?

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
djaiswal83: We are eventually going to move to log4j2 and with that slf4j is not useful. slf4j is anyway a facade and it relies on underlying implementation which is log4j for us.
For basic logging using slf4j would work, however, we make extensive use of appenders and while working on migration, it became really hard to work with slf4j. There were similar views I found on stackoverflow as well.

azkaban-exec-server/src/main/java/azkaban/common/ServerUtils.java
djaiswal83: Given that it is a static method, yes I agree. Will do.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: It was slf4j but the logger is passed into classes which use log4j and share common code with ExecutorServer.
Also, my comment above about slf4j in general.

azkaban-exec-server/src/test/java/azkaban/container/FlowContainerTest.java
```diff
@@ -99,23 +100,47 @@ public void setup() throws Exception {
...
+
+  /**
+   * Don't like the name? Me neither - DJ
+   * Test if Callback Manager is not created if not enabled.
+   * This test must be the first as Callback Manager is a static object which once created
+   * remains there for the remainder of the test suite. The name and ordering ensures that
+   * this test always runs first.
+   */
+  @Test(expected = IllegalStateException.class)
+  public void aaa() throws Exception {
```
jakhani, line 135: Why don't you do it in startup? There is already a method in this class for the same called setUp. Why don't you use that?
For e.g. https://junit.org/junit4/javadoc/4.12/org/junit/BeforeClass.html

djaiswal83: We need to do it once for this negative test only. How will putting it in setUp help? It is called everyday.

azkaban-exec-server/src/test/java/azkaban/container/FlowContainerTest.java
```diff
@@ -99,23 +100,47 @@ public void setup() throws Exception {
...
-    flowContainer.start(props);
-    FlowContainer.launchCtrlMsgListener(flowContainer);
+    this.flowContainer = SERVICE_PROVIDER.getInstance(FlowContainer.class);
+    this.flowContainer.start(props);
+  }
+
+  /**
+   * Don't like the name? Me neither - DJ
+   * Test if Callback Manager is not created if not enabled.
+   * This test must be the first as Callback Manager is a static object which once created
```
sshardool, line 130: Looks like `JobCallbackManager` is creating a singleton the old-fashioned way (with a static `instance` variable), with some fairly obvious drawbacks, especially for testability.
One approach for resolving this would be to make it a Guice based `@Singleton`, that you can freely instantiate for testing.
Another simple (but less preferred) alternative, make this consturctor `private JobCallbackManager(final Props props)` either package-private or `public`, annotate it with `@VisibleForTesting`  and use it directly to instantiate more objects of type `JobCallbackManager`.
I think both of the above approaches would be simpler (and better) than having to jump through some of these hoops for ensuring an ordering of test cases.

sshardool (commented):
Thanks a lot for adding tests for this.

azkaban-exec-server/src/test/java/azkaban/container/FlowContainerTest.java
djaiswal83: If the callback is not enabled then JobCallbackManager should not be instantiated, I think that is why it is not Guice based.
The jobCallbackManager is created from within the Azkaban code and not directly from the test code. Unless we change the Azkaban code to make it happen we cant utilize the 2nd approach.

That is too much code change to make the tests work.
The current approach has only condition which is much easier to meet and avoids touching the existing code.

jakhani (approved):

PR #2820, opened 2021-04-02
Set start status for Containerized exectuions based on the dispatch method

aditya1105:
In containerization, When the dispatch method is containerized based on the ramp and jobtype filter, then start status should be READY otherwise PREPARING

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -251,6 +251,14 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
   }
 
   /**
+   * @param execFlow {@link ExecutableFlow} containing all the information for a flow execution.
+   * @return Return the start status based on the {@link ExecutableFlow}.
+   */
+  public Status getStartStatus(ExecutableFlow execFlow) {
+    return getStartStatus();
+  }
+
+  /**
    * When a flow is submitted, insert a new execution into the database queue. {@inheritDoc}
    */
   @Override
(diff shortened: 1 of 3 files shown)


sshardool (dismissed):
Thanks for addressing this.

jakhani (dismissed):

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -250,6 +250,13 @@ public Status getStartStatus() {
     return Status.PREPARING;
   }
 
+  /**
+   * @return Return the start status based on the {@link ExecutableFlow}
+   */
+  public Status getStartStatus(ExecutableFlow execFlow) {
```
djaiswal83, line 256: execFlow is never used here, is it supposed to call the new API?

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -294,9 +301,9 @@ protected String uploadExecutableFlow(
       String message) throws ExecutorManagerException {
     final int projectId = exflow.getProjectId();
     exflow.setSubmitUser(userId);
-    exflow.setStatus(getStartStatus());
-    exflow.setSubmitTime(System.currentTimeMillis());
     exflow.setDispatchMethod(getDispatchMethod(exflow));
+    exflow.setStatus(getStartStatus(exflow));
```
djaiswal83, line 305: How does changing the order of calls help?

djaiswal83 (changes requested):

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
aditya1105: It is not used in the AbstractExecutorManagerAdapter, but for ContainerizedDispatchManager which overrides it implements it. Since the call to this made will be made on the class AbstractExecutorManagerAdapter, we need it here

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
aditya1105: because dispatch method needs to be set first. Then getStartStatus will use the dispatch method set in the exflow to determine the start status

djaiswal83: Thanks

djaiswal83 (approved):
Thanks for addressing comments. Code looks good.

PR #2823, opened 2021-04-04
Adding a way to log pod spec yaml if pod is marked as dev pod.

jakhani:


djaiswal83 (approved):
Looks good to me. Really handy functionality.

sshardool (approved):
Thanks for making this change!

PR #2825, opened 2021-04-12
[Azkaban containerization CICD] Introduce a TEST state for image version and allow flow override of a test version while executing flow.

abhisheknath2011:
[Azkaban containerization CICD] Introduce a TEST state for image version and allow flow override of a test version while executing flow

Note: This TEST version will be registered to specified Azkaban cluster as part of Azkaban containerization CICD (Orca Job) using image management API. Allow this TEST version to be passed as flow parameter override so that acceptance test flows can executed using this TEST version.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -145,7 +145,8 @@ public class Constants {
   public static final String AZ_HOME = "AZ_HOME";
 
   // Azkaban event reporter constants
-  public static class EventReporterConstants{
+  public static class EventReporterConstants {
+
     public static final String FLOW_NAME = "flowName";
     public static final String AZ_HOST = "azkabanHost";
     public static final String AZ_WEBSERVER = "azkabanWebserver";
@@ -166,7 +167,7 @@ public class Constants {
     public static final String VERSION_SET = "versionSet";
     public static final String EXECUTOR_TYPE = "executorType";
     public static final String PROJECT_FILE_UPLOAD_USER = "projectFileUploadUser";
-    public static final String PROJECT_FILE_UPLOADER_IP_ADDR= "projectFileUploaderIpAddr";
+    public static final String PROJECT_FILE_UPLOADER_IP_ADDR = "projectFileUploaderIpAddr";
     public static final String PROJECT_FILE_NAME = "projectFileName";
     public static final String PROJECT_FILE_UPLOAD_TIME = "projectFileUploadTime";
     public static final String JOB_ID = "jobId";
@@ -476,7 +477,7 @@ public class Constants {
 
     // Job callback
     public static final String AZKABAN_EXECUTOR_JOBCALLBACK_ENABLED =
-            "azkaban.executor.jobcallback.enabled";
+        "azkaban.executor.jobcallback.enabled";
   }
 
   public static class FlowProperties {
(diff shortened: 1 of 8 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -688,5 +690,9 @@
         + ".request";
 
     public static final String FLOW_PARAM_POD_ENV_VAR = "pod.env.var.";
+
+    // Constant to allow test version to be passed as flow parameter. Passing test version will be
+    // allowed for Azkaban ADMIN role only
+    public static final String FLOW_PARAM_ALLOW_TEST_VERSION = "allow.test.version";
```
jakhani, line 696: Shall we rename it to allow.pod.test.version?

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageVersion.java
```diff
@@ -165,13 +168,31 @@ public static State fromStateValue(final String stateValue) {
             .filter(state -> (state.equals(State.NEW) || state.equals(State.ACTIVE)))
             .collect(Collectors.toSet());
 
+    /**
+     * Create set of new, active and test state
+     */
+    private static final Set<State> newActiveAndTestState =
+        Arrays.stream(State.values())
+            .filter(state -> (state.equals(State.NEW) || state.equals(State.ACTIVE)
+                || state.equals(State.TEST)))
```
jakhani, line 177: Are we inserting new azkaban-base version from CICD so that it can be tested?

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageVersion.java
```diff
@@ -165,13 +168,31 @@ public static State fromStateValue(final String stateValue) {
             .filter(state -> (state.equals(State.NEW) || state.equals(State.ACTIVE)))
             .collect(Collectors.toSet());
 
+    /**
+     * Create set of new, active and test state
+     */
+    private static final Set<State> newActiveAndTestState =
+        Arrays.stream(State.values())
+            .filter(state -> (state.equals(State.NEW) || state.equals(State.ACTIVE)
```
jakhani, line 176: Instead of streaming through whole enum set, how about
Set<State> set = new HashSet<>();
    Collections.addAll(set, State.NEW, State.ACTIVE, State.TEST);
    ?

abhisheknath2011: The idea behind this code is to precompute the set of states and keep them in the static variable. Whenever we need to use the filter simply return the state filter. As this list is fixed and part of the enum it is also behaving like how enums are defined and initialized only once. As the states are only a selected few, there won't be any issue if we stream and filter the required states (done only once). On the other hand with the suggested code we have to compute every time whenever the the filter is required (this filter is frequently used during flow dispatch).

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageVersion.java
abhisheknath2011: Yes we are inserting azkaban-base image test version from the Azkaban CICD orca job.

az-core/src/main/java/azkaban/Constants.java
abhisheknath2011: With allow.pod.test.version does not meaning someting like pod test version. How about allow.image.test.version? I think this name provides more context of the usage.

jakhani: That sounds better! Thanks

jakhani (approved):

Issue #1917, opened 2018-08-12
Azkaban unable to get scheduled flow triggers NPE


shangwen:
version: azkaban-3.51.1
```
2018/08/12 19:18:07.067 +0800 ERROR [FlowTriggerScheduler] [Azkaban] unable to get scheduled flow triggers
java.lang.NullPointerException
        at azkaban.flowtrigger.quartz.FlowTriggerScheduler.getScheduledFlowTriggerJobs(FlowTriggerScheduler.java:132)
        at azkaban.webapp.servlet.FlowTriggerServlet.handlePage(FlowTriggerServlet.java:140)
        at azkaban.webapp.servlet.FlowTriggerServlet.handleGet(FlowTriggerServlet.java:55)
        at azkaban.webapp.servlet.LoginAbstractAzkabanServlet.doGet(LoginAbstractAzkabanServlet.java:123)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:668)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:770)
        at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
        at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:401)
        at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:182)
        at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:766)
        at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:152)
        at org.mortbay.jetty.Server.handle(Server.java:326)
        at org.mortbay.jetty.HttpConnection.handleRequest(HttpConnection.java:542)
        at org.mortbay.jetty.HttpConnection$RequestHandler.headerComplete(HttpConnection.java:928)
        at org.mortbay.jetty.HttpParser.parseNext(HttpParser.java:549)
        at org.mortbay.jetty.HttpParser.parseAvailable(HttpParser.java:212)
        at org.mortbay.jetty.HttpConnection.handle(HttpConnection.java:404)
        at org.mortbay.jetty.bio.SocketConnector$Connection.run(SocketConnector.java:228)
        at org.mortbay.thread.QueuedThreadPool$PoolThread.run(QueuedThreadPool.java:582)
```

chengren311:
[...]

PR #2827, opened 2021-04-17
Adding null check to avoid NPE in logs (Fixes issue #1917)

acharneski:
A small fix for https://github.com/azkaban/azkaban/issues/1917

This issue is originally caused by the initialization function at https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/scheduler/QuartzScheduler.java#L58 leaving this field null.

Adding a null check, which returns with the same value as the exception would result in. (Empty list)

This change does not change the behavior of anything except the log output.


acharneski:
Failure appears to be due to a transient failure in unit tests. I cannot reproduce the failure locally.

djaiswal83:
@acharneski Travis CI failed. Can you please check?

acharneski:
@djaiswal83 As I noted, I cannot reproduce this unit test failure, I suspect it is some sort of transient. I do not know how to trigger a retry build.

djaiswal83 (approved):
Thanks for fixing this potential NPE.

PR #2828, opened 2021-04-20
Adding after class method to stop the jetty server.

jakhani:


sshardool (approved):
Thanks for addressing this!

PR #2829, opened 2021-04-20
Log versionSet in FlowContainer.

djaiswal83:


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -302,6 +297,15 @@ void submitFlow(final int execId)
           ". Terminating flow container launch");
     }
 
+    // Log the versionSet for this flow execution
```
jakhani, line 300: Thank you for taking care of this PR.
Minor Nit: Can you please take this logging code in separate method called logVersionSet and call it from here?

jakhani (approved):

PR #2831, opened 2021-04-21
Adds Granular Breadcrumbs to FlowPage and JobPage

chidoneto:
# Issue Description

## No Intermediate Breadcrumbs in Job Page
When you are visiting Job Pages, there is no way to go back to the main flow, not to mention the intermediate sub-flows. The only way you have is either going to the sub-flow where the job belongs, or first going all the way back to the project, and then to the main flow and sub-flows from there:
job-no-link-back

## No Intermediate Breadcrumbs in Flow Page
When you are visiting a Flow Page, there is no way to go back to the main flow, not to mention the intermediate sub-flows. The only way you have is first going all the way back to the project, and then to the main flow or other sub-flows from there:
flow-no-link-back

# Solution
Nested flows are now directly accessible via detailed breadcrumbs in FlowPages and JobPages.
breadcrumbs2


chidoneto:
Newer PR encompasses this one.

PR #2833, opened 2021-04-22
 Add listener for pod monitoring and deletion

sshardool:
This will ensure that flow-pods and corresponding clusterIp-services in
the 'completed' state will be deleted after flow execution finishes.
A subsequent commit will add this automated listener based deletion for
pods in 'errored' states as well.

Also, includes guice bindings for pod event listeners and a minor change
in ContainerizedDispatchManager for startup and shutdown of watch.


azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -174,13 +178,33 @@ public DispatchMethod getDispatchMethod(final ExecutableFlow flow) {
...
+  }
+
+  // Start the event watch if configured.
+  private void startWatch() {
+    if (!containerizedWatch.isPresent()) {
+      logger.info("Containerized watch was not provided and will not be started");
+      return;
+    }
+    if (this.azkProps.getBoolean(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_ENABLED,
+        false) == true) {
```
jakhani, line 201: It is boolean. No need to compare with true.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -174,13 +178,33 @@ public DispatchMethod getDispatchMethod(final ExecutableFlow flow) {
...
   @Override
   public void start() {
     this.queueProcessor = setupQueueProcessor();
     this.queueProcessor.start();
+    this.startWatch();
+  }
+
+  // Start the event watch if configured.
+  private void startWatch() {
+    if (!containerizedWatch.isPresent()) {
```
jakhani, line 196: if (!containerizedWatch.isPresent() || !this.azkProps.getBoolean(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_ENABLED,
        false)) {
log
return;
}

Do you want to reduce cases by adding false check for property in if case?

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -174,13 +178,33 @@ public DispatchMethod getDispatchMethod(final ExecutableFlow flow) {
...
+      logger.info("Containerized watch was not provided and will not be started");
+      return;
+    }
+    if (this.azkProps.getBoolean(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_ENABLED,
+        false) == true) {
+      logger.info("Starting containerized watch.");
+      this.containerizedWatch.get().launchWatch();
```
jakhani, line 203: No exception handling is required here?

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -174,13 +178,33 @@ public DispatchMethod getDispatchMethod(final ExecutableFlow flow) {
...
+   * Currently this includes:
+   * (1) This method is used to start queue processor thread. The queue processor thread will pick
+   *   execution from queue maintained in database and create a container for it.
+   * (2) Starting container event watch, if present.
    */
   @Override
   public void start() {
     this.queueProcessor = setupQueueProcessor();
     this.queueProcessor.start();
+    this.startWatch();
```
djaiswal83, line 191: nitpick: We dont use this to refer to member methods. Feel free to ignore.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -224,6 +248,11 @@ public void shutdown() {
     if (null != this.queueProcessor) {
       this.queueProcessor.shutdown();
     }
+
+    if (containerizedWatch.isPresent()) {
```
jakhani, line 252: Is there any need to check property here?

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedWatch.java
```diff
@@ -0,0 +1,18 @@
+package azkaban.executor.container;
+
+public interface ContainerizedWatch {
```
jakhani, line 3: Please add javadoc comment for this interface.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Contains routines for validation of {@link AzPodStatus} transitions and any other processing
+ * required for the statuses.
+ */
+public class AzPodStatusUtils {
```
jakhani, line 26: Let's rename this class if you want to keep the valid graph walk for statuses here or move it to enum.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
```diff
@@ -0,0 +1,81 @@
...
+  // corresponding to these transitions should never be received.
+  private static final Map<AzPodStatus, Set<AzPodStatus>> invalidSourceStatusMap =
+      buildInvalidStatusMap();
+
+  // Build the invalid status static map.
+  private static Map<AzPodStatus, Set<AzPodStatus>> buildInvalidStatusMap() {
+    ImmutableMap.Builder builder = ImmutableMap.builder()
+        .put(AZ_POD_READY, finalStatuses);
+
+    Set<AzPodStatus> mutableInvalidStatuses = new HashSet<>(finalStatuses);
```
jakhani, line 46: Can you please move builder for each status in separate method or something? 
getInvalidStatusesForAZPodReady and it will return set of invalid statuses for it. It will be useful to maintain this map as well as in future.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Provides callback implementations of {@link AzPodStatusListener} for
+ * (1) Updating WebServer and database state of flows based on Pod status
+ * (2) Driving pod lifecycle actions, such as deleting flow pods in a final state.
+ */
+public class FlowStatusUpdatingListener implements AzPodStatusListener {
```
jakhani, line 32: How about FlowStatusManagerListener? Update is one action. Better to keep it generic.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+    requireNonNull(azkProps, "azkaban properties must not be null");
+    requireNonNull(containerizedImpl, "container implementation must not be null");
+    requireNonNull(executorLoader, "executor loader must not be null");
+    requireNonNull(alerterHolder, "alerter holder must not be null");
+    this.containerizedImpl = containerizedImpl;
+    this.executorLoader = executorLoader;
+    this.alerterHolder = alerterHolder;
+
+    maxCacheEntries =
+        azkProps.getInt(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_DRIVER_THREAD_POOL_SIZE,
```
jakhani, line 66: The property and default value names are confusing. One is for thread pool and other is for cache max entries?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  /**
+   * Validate the the transition from the current state to the new state in the {@code event}
+   * is supported. Note that the validation is on best-effort basis as the cache may not always
+   * contain an event for all running flows. This could happen, for example, during webserver
+   * restarts.
+   * @param event pod watch event
+   */
+  private void validateTransition(AzPodStatusMetadata event) {
```
jakhani, line 92: Generally validate methods are expected to return boolean. Can you please do the same here?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+
+  /**
+   * Includes any validations needed on event that are common to all {@link AzPodStatus}
+   *
+   * @param event pod watch event
+   */
+  private void validateAndPreProcess(AzPodStatusMetadata event) {
+    // Confirm that event is indeed a for a FlowContainer and has the corresponding metadata
+    if (!event.getFlowPodMetadata().isPresent()) {
+      IllegalStateException ise = new IllegalStateException(
```
jakhani, line 121: Just wondering shall we use Watch exception here?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+    ExecutableFlow executableFlow = null;
+    try {
+      executableFlow = executorLoader.fetchExecutableFlow(executionId);
+    } catch (ExecutorManagerException e) {
+      String message = format("Exception while fetching executable flow for pod %s",
+          event.getPodName());
+      logger.error(message, e);
+      throw new AzkabanWatchException(message, e);
+    }
+    checkState(executableFlow != null, "executable flow must not be null");
```
jakhani, line 217: Shall we log it silently?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+
+  /**
+   * Delete the the flow pod and any other related objects (such as services).
+   *
+   * @param event pod watch event
+   */
+  private void deleteFlowContainer(AzPodStatusMetadata event) {
+    logger.info("Deleting Flow Pod: " + event.getPodName());
+    try {
+      containerizedImpl.deleteContainer(
```
jakhani, line 239: Just wondering if deleteContainer should take care of validating flow status and set appropriately if it is not end state.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+    } catch (NumberFormatException ne) {
+      String message = format("Flow metadata execution id is not a valid integer %s",
+          event.getFlowPodMetadata().get().getExecutionId());
+      throw new AzkabanWatchException(message, ne);
+    }
+  }
+
+  @Override
+  public void onPodCompleted(AzPodStatusMetadata event) {
+    requireNonNull(event, "event must not be null");
```
jakhani, line 255: Log error silently?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final Set<AzPodStatus> finalStatuses = EnumSet.of(
+      AZ_POD_COMPLETED,
+      AZ_POD_INIT_FAILURE,
+      AZ_POD_APP_FAILURE,
+      AZ_POD_DELETED,
+      AZ_POD_UNEXPECTED
+  );
+
+  // A key in this map is the 'target' status and the value is the set of 'source statuses' from
+  // which transition the the 'target' (key) status is not permitted and and event sequence
```
arvindpruthi, line 36: "the" is repeated

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
sshardool: The logs in the two separate `if` blocks convey slightly different information. In one case the watch is not initialized and in the other it's explicitly disabled because of the property. I would prefer to keep the separate as it will make it easier to debug future issues from the log file. 
On a sidenote, this kind of a layout would have been avoidable with the use of `OptionalBinder` in guice, but it wasn't clear if the version we use fully supported it. We should consider updating guice in the near future.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
sshardool: `ContainerizedWatch` and the underlying implementation in `KubernetesWatch` are designed to not throw any checked exceptions. They can throw unchecked exceptions though, however at this point in the code it seemed correct for those exceptions to be propagated to the callers of this method.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
sshardool: Good catch. I will make that change.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
sshardool: I thought about it as well but the current exception seemed a lot more direct of a match, which in general is the recommended practice as well. Let me know if there is still a concern about this.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
sshardool: I do see the value in logging this silently, however for a valid flow container we `fetchExecutableFlow` should not return null.
I have refactored parts of this code for the next PR, which should address any concerns about this. We can revisit this as part of that PR.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
sshardool: Wanted keep this function both small and specific to the objective of deleting a pod. Any additional checks for flows status etc can be composed in a higher level method like onPodCompleted

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
sshardool: We need to exit early if the argument is null in this case. The call-stack components like AzPodDrivingListener and KubernetesWatch and designed to be resilient to any exceptions reported from this code.

djaiswal83 (changes requested):
Thanks for implementing this feature. The code is of high quality!
Can you please use "final" keyword wherever applicable as per the coding guidelines?
I have few comments, can you please address them?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
sshardool: I have generally seen the convention that methods of the form `is*` return a boolean whereas `validate*` will usually throw exceptions (somewhat like an assertion). In the existing Azkaban code base there are ample examples of `validate*` methods with return types of both void and boolean  (int and other types are also there for that matter). 
Let me know if this is still a concern.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -174,13 +178,33 @@ public DispatchMethod getDispatchMethod(final ExecutableFlow flow) {
...
+    this.startWatch();
+  }
+
+  // Start the event watch if configured.
+  private void startWatch() {
+    if (!containerizedWatch.isPresent()) {
+      logger.info("Containerized watch was not provided and will not be started");
+      return;
+    }
+    if (this.azkProps.getBoolean(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_ENABLED,
```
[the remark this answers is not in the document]
sshardool: Good point, I agree with it overall however the code within the block is quite small so is fairly readable even in the current format.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
```diff
@@ -0,0 +1,81 @@
...
+  // corresponding to these transitions should never be received.
+  private static final Map<AzPodStatus, Set<AzPodStatus>> invalidSourceStatusMap =
```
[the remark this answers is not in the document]
sshardool: That was my original thought too, however the final pod statuses, reflected in the set `finalStatuses` is the largest common subset of the states from which the transitions are not allowed to almost all of the other states. And starting from those states started making a lot more sense and is also more reflective of the reality that intermediate transitions could be missing but we know for sure out-of-order (i.e invalid) transitions events should not occur. 
I am refactoring this code a bit which should help clarify further.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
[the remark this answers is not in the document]
sshardool: Great catch! Missed it.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
```diff
@@ -0,0 +1,317 @@
...
+  // Convenience member for referring to the Cache through ConcurrentMap interface.
+  private final ConcurrentMap<String, AzPodStatusMetadata> podStatusMap;
+  private final AtomicLong flowContainerEventCount = new AtomicLong(0);
+
+  @Inject
+  public FlowStatusUpdatingListener(Props azkProps,
+      ContainerizedImpl containerizedImpl,
+      ExecutorLoader executorLoader,
+      AlerterHolder alerterHolder) {
+    requireNonNull(azkProps, "azkaban properties must not be null");
```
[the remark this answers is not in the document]
sshardool: It's still good to have these as sanity checks as not all invocations may be from guice. E.g test-code and future production code may have valid cases for not instantiating this through guice.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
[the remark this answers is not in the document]
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedWatch.java
sshardool: Done.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
sshardool: Fixed.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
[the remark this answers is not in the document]
sshardool: Fixed.

sshardool:
> Thanks for implementing this feature. The code is of high quality!
> Can you please use "final" keyword wherever applicable as per the coding guidelines?
> I have few comments, can you please address them?

I believe all the comments other than the one for 'final' keyword and log4j have been addressed. I would prefer to address these in a separate PR which will update this for other files in the module as well, if that's ok. Separately, I have concerns about the requirement of using 'final' for local variables and arguments but we can discuss those in a different forum.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
jakhani: Sounds good.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusUpdatingListener.java
jakhani: No concern. Sounds good.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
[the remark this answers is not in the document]
djaiswal83: In the future it may become large so making that change now would be better.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusUtils.java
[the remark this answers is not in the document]
djaiswal83: thanks

djaiswal83 (approved):
Thanks for addressing the comments. The change looks good.

PR #2836, opened 2021-04-27
Added indexes for the containerization tables

aditya1105:


azkaban-db/src/main/sql/create.containerization-tables-all.sql
```diff
@@ -71,9 +72,14 @@ CREATE TABLE IF NOT EXISTS image_rampup_plan (
 --  One option is to move each table create scripts to separate file. But all the  containerization
 --  tables are placed in this file so that it easier to manage.
 
--- Index on image_rampup_plan table
--- create index active_rampup_plan
--- on image_rampup_plan (active);
+CREATE INDEX image_rampup_plan_name
+  ON image_rampup_plan (name);
```
jakhani, line 76: Are we using this in where clause?

azkaban-db/src/main/sql/create.containerization-tables-all.sql
```diff
@@ -39,6 +33,9 @@ CREATE TABLE IF NOT EXISTS image_versions (
   UNIQUE (type_id, version)
 );
 
+CREATE INDEX image_versions_type_id
+  ON image_versions (type_id);
```
jakhani, line 37: Do we need index on state column as well for image_versions table?

azkaban-db/src/main/sql/create.containerization-tables-all.sql
aditya1105: Actually this is not required.

azkaban-db/src/main/sql/create.containerization-tables-all.sql
aditya1105: so the type id is a foreign key. Hence it will be useful to have indexed for it. For the state, the index is not required.

jakhani: If there is state column in where clause then it will be useful.

aditya1105: the state is not there in the where clause for any query

jakhani: Sounds good. Thank you for clarifying the doubt.

jakhani (approved):

PR #2839, opened 2021-04-28
Introducing userToProxyFQN in getProxyUser method.

jakhani:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -169,35 +170,51 @@ private String verifySecureProperty(final Props props, final String s)
...
+  private UserGroupInformation getProxiedUser(final String userToProxy)
+      throws HadoopSecurityManagerException {
+    return getProxiedUser(userToProxy, userToProxy);
+  }
+
   /**
    * Create a proxied user based on the explicit user name, taking other parameters necessary from
    * properties file.
+   *
+   * @param userToProxyFQN
```
li-ygerchikov, line 189: This might be a good place to explain the need and the meaning of the additional parameter. Otherwise, to someone who does not see the entire class hierarchy, only one of the two parameters is used here, and it's not clear why we need this change.
BTW, this is also a sign that structurally this is not optimal. We are changing this place *because* we know something else about somewhere else in the class hierarchy. So this class does not stand on its own. I'd have to think about a better way, and I guess we don't have time for it.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -169,35 +170,51 @@ private String verifySecureProperty(final Props props, final String s)
...
    */
   @Override
-  public synchronized UserGroupInformation getProxiedUser(final String userToProxy)
+  public synchronized UserGroupInformation getProxiedUser(final String userToProxyFQN,
+      final String userToProxy)
       throws HadoopSecurityManagerException {
 
-    if (userToProxy == null) {
+    if (userToProxyFQN == null) {
       throw new HadoopSecurityManagerException("userToProxy can't be null");
```
li-ygerchikov, line 198: Should we refer here to the parameter we actually use and check?

li-ygerchikov (approved):

djaiswal83 (approved):
Looks good to me.
I will followup this with cleaner resolution to @li-ygerchikov 's concerns.

PR #2841, opened 2021-04-29
Allow longer project events to be recorded.

ypadron-in:
User actions on projects are recorded in the DB and available for inspection.
Examples of those actions are uploading a new zip file, granting permissions to users and adding/modifying/deleting job properties in a flow through the UI.
When users update multiple job properties at a time or the value of one of them is very long it can happen (we often see it in the web server logs) that the event fails to be saved in the DB because the log message is too long.
To account for those scenarios the size of the MySQL column storing the log message is increased.

### DIFF
--- a/azkaban-db/src/main/sql/create.project_events.sql
+++ b/azkaban-db/src/main/sql/create.project_events.sql
@@ -3,7 +3,7 @@ CREATE TABLE project_events (
   event_type TINYINT NOT NULL,
   event_time BIGINT  NOT NULL,
   username   VARCHAR(64),
-  message    TEXT
+  message    VARCHAR(65400)
 );
 
 CREATE INDEX log
--- a/azkaban-db/src/main/sql/create.project_events.sql
+++ b/azkaban-db/src/main/sql/create.project_events.sql
@@ -3,7 +3,7 @@ CREATE TABLE project_events (
   event_type TINYINT NOT NULL,
   event_time BIGINT  NOT NULL,
   username   VARCHAR(64),
-  message    VARCHAR(512)
+  message    TEXT
 );
 
 CREATE INDEX log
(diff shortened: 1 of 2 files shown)


azkaban-db/src/main/sql/create.project_events.sql
```diff
@@ -3,7 +3,7 @@ CREATE TABLE project_events (
   event_type TINYINT NOT NULL,
   event_time BIGINT  NOT NULL,
   username   VARCHAR(64),
-  message    VARCHAR(512)
+  message    VARCHAR(65400)
```
li-ygerchikov, line 6: Not 65535? Or 65000? Or 65500? Seems a bit arbitrary, but either is fine, practically speaking.

li-ygerchikov (approved):

azkaban-db/src/main/sql/create.project_events.sql
ypadron-in: Tried with 65,535 but got an error:
```
ERROR 1118 (42000): Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
```
Based on https://stackoverflow.com/questions/13283381/row-size-too-large-error-in-mysql-create-table-query/13283574 looks like the entire row size should not exceed 65,535 bytes so I subtracted the size of the other columns and rounded down ~65400.

The table `project_events` currently have below columns:
project_id   ->  int (4 bytes)
event_type ->  tinyint (1 byte)
event_time -> bigint (8 bytes) 
username   -> varchar(64) (64 + 2 for size? bytes)

But maybe I should leave it at 65000 in case more columns are added.

li-ygerchikov: TIL. Thank you.
I like 65000.

li-ygerchikov (approved):

PR #2843, opened 2021-04-29
Adds Granular Breadcrumbs to Flow, Job, Job Properties and Execution Pages

chidoneto:
# Issue Description

## No Intermediate Breadcrumbs in Job Page
When you are visiting Job Pages, there is no way to go back to the main flow, not to mention the intermediate sub-flows. The only way you have is either going to the sub-flow where the job belongs, or first going all the way back to the project, and then to the main flow and sub-flows from there:
job-no-link-back

## No Intermediate Breadcrumbs in Flow Page
When you are visiting a Flow Page, there is no way to go back to the main flow, not to mention the intermediate sub-flows. The only way you have is first going all the way back to the project, and then to the main flow or other sub-flows from there:
flow-no-link-back

## No Intermediate Breadcrumbs in Properties Page
properties-page-no-links

## No Intermediate Breadcrumbs in Flow Execution Page
flow-execution-no-links

# Solution
Nested flows are now directly accessible via detailed breadcrumbs in FlowPages and JobPages.
breadcrumbs2

## Job Properties w/Granular Breadcrumbs
properties-page-with-links2

## Flow Execution w/Granular Breadcrumbs
exec-page-with-links

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -379,7 +379,8 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
 
     page.add("projectName", project.getName());
     page.add("flowid", flow.getId());
-    page.add("flowlist", flow.getId().split(":", 0));
+    page.add("flowlist", flow.getId().split(Constants.PATH_DELIMITER, 0));
+    page.add("pathDelimiter", Constants.PATH_DELIMITER);
     page.add("parentflowid", node.getParentFlow().getFlowId());
     page.add("jobname", node.getId());
     page.add("jobType", node.getType());
@@ -446,7 +447,8 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
     page.add("projectId", project.getId());
     page.add("projectName", project.getName());
     page.add("flowid", triggerInst.getFlowId());
-    page.add("flowlist", triggerInst.getFlowId().split(":", 0));
+    page.add("flowlist", triggerInst.getFlowId().split(Constants.PATH_DELIMITER, 0));
+    page.add("pathDelimiter", Constants.PATH_DELIMITER);
 
     page.render();
   }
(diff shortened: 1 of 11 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -379,6 +379,7 @@ private void handleExecutionJobDetailsPage(final HttpServletRequest req,
 
     page.add("projectName", project.getName());
     page.add("flowid", flow.getId());
+    page.add("flowlist", flow.getId().split(":", 0));
```
ypadron-in, line 382: Can we use the constant `azkaban.Constants.PATH_DELIMITER` to refer to the separator?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/executingflowpage.vm
```diff
@@ -97,13 +97,25 @@
...
+            #set($flowlabel = "Flow")
+            #set($linkclass = "")
+          #else
+            #set($flowlabel = "")
+            #set($linkclass = "flowlink")
+          #end
+
+          <li class="${linkclass}"><a href="${ref}"><strong>$flowlabel</strong> $flow</a></li>
+
+          #set($ref = $ref + ":")
```
ypadron-in, line 116: To avoid hardcoding the path delimiter and reduce repeating this logic across templates what do you think about creating a map of (node name for label, fully qualified path of node for uri) in the servlets and passing that to the velocity template instead of a list of node names only (`flowlist`).

azkaban-web-server/src/main/less/header.less
```diff
@@ -67,5 +67,9 @@
     margin: 0px;
     padding: 0px;
     background: transparent;
+
+    .flowlink:before {
+      content: ":";
+    }
```
ypadron-in, line 73: To make sure all clients apply these style please update `v=...` to current timestamp in style.vm file.
`<link href="/css/azkaban.css?v=<current-timestamp>" rel="stylesheet">`

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
chidoneto: Done.

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/executingflowpage.vm
chidoneto: Rather than repeating the code on the servlet side, and altering the logic, I am passing the constant down to the form.

azkaban-web-server/src/main/less/header.less
chidoneto: Done.

ypadron-in (approved):

PR #2845, opened 2021-04-30
Make the rampup logic more deterministic by sorting the result images by versionID.

djaiswal83:
Make the rampup logic more deterministic by sorting the result images by versionID.

Updated the input JSONs for tests in sorted order as it would be if it were queried from the db.

### DIFF
--- a/azkaban-common/src/test/resources/image_management/image_type_rampups.json
+++ b/azkaban-common/src/test/resources/image_management/image_type_rampups.json
@@ -16,7 +16,11 @@
     },
     {
       "imageVersion": "2.1.2",
-      "rampupPercentage": "70"
+      "rampupPercentage": "20"
+    },
+    {
+      "imageVersion": "2.1.3",
+      "rampupPercentage": "50"
     }
   ],
   "azkaban_core": [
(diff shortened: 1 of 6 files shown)


azkaban-common/src/test/resources/image_management/image_type_rampups.json
```diff
@@ -1,42 +1,42 @@
...
-    {
-      "imageVersion": "1.1.2",
-      "rampupPercentage": "70"
-    },
     {
       "imageVersion": "1.1.1",
       "rampupPercentage": "30"
+    },
+    {
+      "imageVersion": "1.1.2",
```
jakhani, line 8: Can you please add three or more versions in one of the test cases? Add different order like
1.1.3
1.1.1
1.1.2
This will cover that ordering is not required.

djaiswal83: This JSON is what is created after the query, that is why it needs to be sorted to mimic the query result. There is no change other than sort order of image version.
I can add 1 more version in a few cases.

jakhani: That sounds good. Can you please create ramp up plan with three and order that I mentioned and then verify this order?

djaiswal83: Done.

hqin2016 (approved):
Thanks for fixing this issue.

jakhani (approved):

PR #2847, opened 2021-05-03
Add all callback methods for the FlowStatusManagerListener

sshardool:
This also:
- Simplifies event processing and moves the executor for it within `FlowStatusManager`.
- Fixes part of the logic for identifying init-container failures.


azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -217,18 +221,21 @@ private void updatePodStatus(AzPodStatusMetadata event) {
...
+      logger.error("Unable to find executable flow for execution: " + executionId);
+      return Optional.empty();
+    }
+    Status originalStatus = executableFlow.getStatus();
 
-    if (!expectedStatusMatcher.apply(originalStatus)) {
+    if (!Status.isStatusFinished(originalStatus)) {
       logger.info(format(
-          "Flow for pod %s does not have the expected status in database and will be finalized.",
+          "Flow for pod %s does not have a final status in database and will be finalized.",
```
jakhani, line 232: It will be good to log execution id.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -253,68 +260,95 @@ private void deleteFlowContainer(AzPodStatusMetadata event) {
...
-    logger.warn("Unsupported method.", createUnsupportedException(event));
+    executor.execute(() -> processFinalState(event));
+  }
+
+  public void shutdown() {
+    this.executor.shutdown();
+    try {
+      this.executor.awaitTermination(SHUTDOWN_TERMINATION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      logger.warn("Executor service shutdown was interrupted.", e);
```
jakhani, line 351: Can you please add that which executor service it is in log message?

jakhani (commented):
Thank you for taking care of this task. I have added minor two comments.
With that, can you please add final for all the local or method variables wherever it is applicable?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
sshardool: Done.

jakhani (approved):

PR #2848, opened 2021-05-03
Allow hotfix version in image versions.

djaiswal83:


jakhani (commented):
Do we have any unit test around it that you can modify to cover this use case?

djaiswal83:
> Do we have any unit test around it that you can modify to cover this use case?

I looked up and there is none. Let me see if I can add one. I did test this out on a test cluster.

jakhani (approved):

jakhani (approved):
Superb! Thank you so much!

PR #2850, opened 2021-05-06
Fixed job life cycle event listener exception on log

hqin2016:
Tested in holdem 6 Kubernetes pods, the exception is not appearing after the fix.


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -699,12 +702,16 @@ private void uploadLogFile(final int execId) {
    */
   @Override
   public void handleEvent(final Event event) {
-    final FlowRunner flowRunner = (FlowRunner) event.getRunner();
-    final ExecutableFlow flow = flowRunner.getExecutableFlow();
-    // Set Flow level SLA options for containerized executions
-    this.triggerManager
-        .addTrigger(flow.getExecutionId(), SlaOption.getFlowLevelSLAOptions(flow
-            .getExecutionOptions().getSlaOptions()));
+    final List<EventType> flowEventTypeList = Arrays.asList(EventType.FLOW_STARTED,
```
jakhani, line 705: Why don't you take this in enum? Introduce method like isFlowEvent(eventType)

djaiswal83 (approved):
Thanks for addressing this quickly.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
hqin2016: good idea

hqin2016:
tested on holdem 4 for sanity check, and holdem 6 Kubernetes pod

djaiswal83 (approved):
This is much cleaner code. Excellent suggestion @jakhani

jakhani (approved):

PR #2853, opened 2021-05-08
Fixed excessive pod event reporting, and incorrect executor type in the event metadata

hqin2016:
Tested on holdem 6, verified the event data by consuming the Kafka message. Tested on holdem 4 for sanity check.


azkaban-common/src/main/java/azkaban/executor/ExecutionController.java
```diff
@@ -49,7 +49,7 @@ public ExecutionController(final Props azkProps, final ExecutorLoader executorLo
       final CommonMetrics commonMetrics,
       final ExecutorApiGateway apiGateway, final AlerterHolder alerterHolder, final
   ExecutorHealthChecker executorHealthChecker) {
-    super(azkProps, executorLoader, commonMetrics, apiGateway, alerterHolder);
+    super(azkProps, executorLoader, commonMetrics, apiGateway, alerterHolder, new DummyEventListener());
```
jakhani, line 52: Won't this break the READY or PREPARING event that you are sending from Abstract class?

azkaban-common/src/main/java/azkaban/executor/container/DummyEventListener.java
```diff
@@ -0,0 +1,28 @@
...
+import azkaban.event.EventListener;
+
+/**
+ * This class is a no-op implementation of {@link EventListener}
+ */
+public class DummyEventListener implements EventListener {
+
+  @Override
+  public void handleEvent(Object event) {
+  }
```
jakhani, line 27: Please log here that doing nothing kind of message.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -151,13 +152,15 @@ public KubernetesContainerizedImpl(final Props azkProps,
...
-      final KubernetesWatch kubernetesWatch)
+      final KubernetesWatch kubernetesWatch, EventListener eventListener)
       throws ExecutorManagerException {
     this.azkProps = azkProps;
     this.executorLoader = executorLoader;
     this.versionSetLoader = versionSetLoader;
     this.imageRampupManager = imageRampupManager;
     this.kubernetesWatch = kubernetesWatch;
+    this.eventListener = eventListener;
+    this.addListener(this.eventListener);
```
jakhani, line 163: Don't add the listener here again. It will fire two events. It should be added only once which is happening from ContainerizedDispatchManager.

hqin2016: removed listener

azkaban-common/src/main/java/azkaban/executor/ExecutorManager.java
```diff
@@ -88,8 +89,8 @@ public ExecutorManager(final Props azkProps, final ExecutorLoader executorLoader
       final ActiveExecutors activeExecutors,
       final ExecutorManagerUpdaterStage updaterStage,
       final ExecutionFinalizer executionFinalizer,
-      final RunningExecutionsUpdaterThread updaterThread) {
-    super(azkProps, executorLoader, commonMetrics, apiGateway, null);
+      final RunningExecutionsUpdaterThread updaterThread, EventListener eventListener) {
```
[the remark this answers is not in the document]
hqin2016: It's set to null.

azkaban-common/src/main/java/azkaban/executor/container/DummyEventListener.java
li-ygerchikov: I appreciate the logging idea for debug purposes, but would this not be a bit excessive and not terribly useful in web server logs? Could we maybe log with DEBUG or TRACE level instead of INFO?

hqin2016: Good idea, used log.trace

azkaban-common/src/main/java/azkaban/executor/container/DummyEventListener.java
```diff
@@ -16,17 +16,16 @@
...
-  private static final Logger logger = LoggerFactory.getLogger(EventListener.class);
+  private static final Logger logger = Logger.getLogger(FlowStatusChangeEventListener.class);
 
   @Override
   public void handleEvent(Object event) {
-    logger.info("Event listener has not been configured");
+    logger.debug("Event listener is not been implemented");
   }
-}
+}
```
jakhani, line 31: Nit: New line

azkaban-common/src/main/java/azkaban/executor/container/DummyEventListener.java
```diff
@@ -16,17 +16,16 @@
...
 import azkaban.event.EventListener;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.apache.log4j.Logger;
 
 /**
- * This class is a no-op implementation of {@link EventListener}
+ * This class is a no-op implementation of {@link FlowStatusChangeEventListener}
  */
 public class DummyEventListener implements EventListener {
```
jakhani, line 24: This can be generic class. You can move it to the package where EventListener is and also no need to use logger of FlowStatusChangeEventListener class. Use its own class name.

azkaban-common/src/main/java/azkaban/executor/container/FlowStatusChangeEventListener.java
```diff
@@ -31,20 +31,20 @@
...
 
 /**
- * This class is the event listener for flow executed in a Kubernetes pod
- * It handles event, gets event metadata, and sends metadata to event reporter plugin
+ * This class reports FLOW_STATUS_CHANGED events in web server, flow statuses include
+ * e.g. READY, DISPATCHING, PREPARING
+ * It handles event, prepares event metadata, and sends metadata to event reporter plugin
  */
-public class PodEventListener implements EventListener<Event> {
+public class FlowStatusChangeEventListener implements EventListener<Event> {
```
jakhani, line 41: You don't need to keep this class in container package as it is generic for all types of dispatch methods.

hqin2016: moved to azkaban/executor

azkaban-common/src/main/java/azkaban/executor/container/DummyEventListener.java
hqin2016: moved to azkaban/executor

jakhani (approved):

djaiswal83 (approved):
Thanks for addressing comments.

PR #2855, opened 2021-05-14
Completion of html tags

Ansen:
When the HTML tags is incomplete,  [Mail and Calendar](https://www.microsoft.com/en-us/p/mail-and-calendar/9wzdncrfhvqm) App cannot render the mail content.

image


djaiswal83:
@Ansen thanks for making this change. Can you please check?

PR #2857, opened 2021-05-18
Disable automounting of service account token in kubernetes pod

sshardool:


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -692,6 +701,9 @@ private void createPod(final int executionId) throws ExecutorManagerException {
     allImageTypes.addAll(jobTypes);
     final VersionSet versionSet = fetchVersionSet(executionId, flowParam, allImageTypes, flow);
     final V1PodSpec podSpec = createPodSpec(executionId, versionSet, jobTypes, flowParam);
+    disableSATokenAutomount(podSpec);
```
jakhani, line 704: Shouldn't this be coming from property? Default value can be false.

sshardool: Do you mean from `azkaban.properties` ?
(1) It should not be coming from a flow property (will be a security concern)
(2) There is a slightly stronger case to add this in `azkaban.properties` but still seemed like something that should not even be made configurable. This will also force us to make any future changes in a manner that does not assume presence of sa-tokens on the pod.

jakhani: In that case, shall we directly get pod spec with disabled service account mounting? Shall we just add 
v1PodSpec.automountServiceAccountToken(false); after
https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodBuilder.java#L81

jakhani (approved):

PR #2858, opened 2021-05-19
Marking nscd mount as read only

jakhani:
For some volume mounts like nscd, it needs to be marked as read only for security reason. If it doesn't need to be kept read only then that can be set from property as well
azkaban.kubernetes.pod.nscd.mount.read.only=false


djaiswal83 (approved):
Code looks good to me.

PR #2860, opened 2021-05-20
Stop router-specific admin properties from being overridden by jobs

haibchen:
"azkaban.job.cluster.classpaths" and "azkaban.job.cluster.nativelib" are two critical job properties that ensure jobs are executed properly when a cluster router is enabled.

However, jobs can accidentally override these properties, resulting in job failures. For example, some HadoopJava jobs may output every one of its job properties as its output property, including these two.
When a downstream job starts, it gets them as its own job properties (nodes gets output properties from their in-nodes), which may be incorrect.

These two are meant to be generated by the cluster router per job and stay private within a job.
Here we essentially always allow the router to generate "azkaban.job.cluster.classpaths" and "azkaban.job.cluster.nativelib"

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -325,6 +325,13 @@ public class JobTypeManager {
     return createJob(jobId, jobParams, logger);
   }
 
+  /**
+   * Create job parameters that can be used to create a job instance.
+   * @param jobId job id
+   * @param jobProps job properties
+   * @param logger logger
+   * @return job parameters that can be used to create a job instance
+   */
   public JobParams createJobParams(final String jobId, Props jobProps, final Logger logger) {
     // This is final because during build phase, you should never need to swap
     // the pluginSet for safety reasons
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -399,6 +403,15 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
     }
   }
```
djaiswal83, line 405: Can you please add JavaDoc?

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -381,7 +381,11 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
       // inject cluster jars and native libraries into jobs through properties
       // User's job props should take precedence over cluster props
       Props finalProps = getClusterSpecificJobProps(targetCluster, jobProps, pluginLoadProps);
+      Props nonOverriddableClusterProps = getClusterSpecificNonOverridableJobProps(finalProps);
       finalProps.putAll(jobProps);
+      // add router-specific properties that are critical for job execution as the last step
```
djaiswal83, line 386: Please add cautionary comment, perhaps something in all caps to make sure no one adds props after that code.

djaiswal83 (changes requested):
This is a good catch!
I have couple of minor comments.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
haibchen: added

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
haibchen: addressed

djaiswal83 (approved):

PR #2868, opened 2021-06-02
Made 'Finish all possible' as default failure-option

svnarumugam:
This is a specific use case in conditional flows (https://azkaban.readthedocs.io/en/latest/conditionalFlow.html). If we consider the following flow, the user wants to run JobC regardless of failure in any node. Currently, when jobX fails, the remaining nodes get canceled. 

JobX
    |
JobA 
   |
JobB -> all_success (default) 
   |
JobC -> conditioned all_done 

**Current behaviour:** if a node fails, we decide to run subsequent nodes based on the flag "failure option". The default value of this is finish currently running nodes. 
**Expected behaviour:** In the case of conditional flows, we should evaluate all nodes ideally as the condition can be placed on any node and the condition influence the nodes to run regardless of any previous nodes' failure or success. 
**Fix:** Modifying the default failure option to "finish all" will make sure that all nodes will run for conditional flows. This will not affect any existing schedules. The non-conditional flows shouldn't have any functional problem because of this. cc: @ypadron-in 

**Testing:** Tested this change by starting an ad-hoc flow/scheduling a flow in a local machine.

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/flowexecutionpanel.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/flowexecutionpanel.vm
@@ -126,7 +126,7 @@
                       class="form-control form-control-auto">
                 <option value="finishCurrent">Finish Current Running</option>
                 <option value="cancelImmediately">Cancel All</option>
-                <option value="finishPossible">Finish All Possible</option>
+                <option value="finishPossible" selected>Finish All Possible</option>
               </select>
             </div>
 


anish-mal:
The description for the PR (and the included URL https://jira01.corp.linkedin.com:8443/browse/LIHADOOP-57817) will be visible externally.

However, the URL itself will not be accessible to everyone. Let's pull in the required information and make the description self contained.

anish-mal (changes requested):
Thanks for the change.The description (#2868 (comment)) for the PR (and the included URL https://jira01.corp.linkedin.com:8443/browse/LIHADOOP-57817) will be visible externally.

However, the URL itself will not be accessible to everyone. Let's pull in the required information and make the description self contained.

svnarumugam:
Thanks for pointing that out @anish-mal. Updated the description.

anish-mal (approved):
Thanks for additional testing and for improving the description.

ypadron-in (changes requested):
As mentioned `FINISH_CURRENTLY_RUNNING` has been the default failure action everywhere (UI, APIs, data triggers) for a long time, so it's probable that a lot of users just skip setting up this option when executing flows because they know what to expect and rely on that. For those users things might break with this change as the failure option to use depends on the behavior and the definition choices of a specific user DAG.

From Azkaban perspective `FINISH_ALL_POSSIBLE` means executing more jobs, so the resource consumption should be higher with that option than with `FINISH_CURRENTLY_RUNNING`. I don't feel `FINISH_ALL_POSSIBLE` is the appropriate default option, specially because it only benefits flows using conditional flows, which I don't expect majority (>50%) of total flows to be using. Do we know what's the percentage of flows using Flow 2.0 and (out of that) the percentage of flows using Conditional Flows?

Based on the description I feel the less disruptive and more performant solution to this issue is to improve our internal and official documentations and educate the user regarding the meaning of the different failure options and their effect when using Conditional Flows.

svnarumugam:
I agree that resource consumption will be more (although not significant imo) and the percentage of the conditional flow would be low in the overall flows we run. More than anything else, I do agree that it will change the default behaviour for the majority of the users, and we will only help a small percentage of users who use conditional flows. I will update the document accordingly. closing this PR. Thank you for your input @ypadron-in .

PR #2873, opened 2021-06-10
[UI] Load all inherited properties in Job Page for flows using Flow 2.0.

ypadron-in:
In the Job page, job inherited properties are not generated properly for flows using Flow 2.0. 
Here is what's currently rendered for an embedded job:
Screen Shot 2021-06-10 at 11 45 49 AM

**Issues**:
-The `longLongs.flow` link suggests that root flow properties will be displayed when clicked but in fact the properties of direct parent (emb_5) of job jobE5B are shown instead.
-Properties inherited from parents emb_3, emb_2 and root flow itself are not available for inspection.

With changes in this PR inherited properties from all parents are listed and loaded correctly:
Screen Shot 2021-06-10 at 11 52 04 AM
Example Properties Page:
Screen Shot 2021-06-10 at 12 37 35 PM

**Test** the changes locally. I verified that inherited properties continue to load correctly for flows using Flow 1.0.

**Follow up**:
-Populate the sections `Inherited From` and `Source of` in the Property Page for flows written with Flow 2.0 syntax.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/flow/Flow.java
+++ b/azkaban-common/src/main/java/azkaban/flow/Flow.java
@@ -296,6 +296,8 @@ public class Flow {
 
   /**
    * Return all flow parents of this flow. From each parent its name and node path are returned.
+   *
+   * @return the list of parents
    */
   public List<Pair<String, String>> getParents() {
     final List<Pair<String, String>> parents = new ArrayList<>(); // Tuple (flow name, node path)
--- a/azkaban-common/src/main/java/azkaban/flow/Flow.java
+++ b/azkaban-common/src/main/java/azkaban/flow/Flow.java
@@ -18,6 +18,7 @@ package azkaban.flow;
 
 import azkaban.Constants;
 import azkaban.executor.mail.DefaultMailCreator;
+import azkaban.utils.Pair;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -47,7 +48,7 @@ public class Flow {
   private static final String IS_LOCKED_PROPERTY = "isLocked";
   private static final String FLOW_LOCK_ERROR_MESSAGE_PROPERTY = "flowLockErrorMessage";
 
-  private final String id;  // This is actually the flow name
+  private final String id;  // This is actually the node path. I.e. rootFlow:subFlow1:subFlow2
   private int projectId;
   private int version = -1; // This is actually the project version
   private final HashMap<String, Node> nodes = new HashMap<>();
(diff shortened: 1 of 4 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1584,99 +1579,110 @@ private void handleJobPage(final HttpServletRequest req, final HttpServletRespon
...
     final String projectName = getParam(req, "project");
-    final String flowName = getParam(req, "flow");
+
+    // flow and job parameters are mainly used to build the breadcrumb.
+    final String flowNodePath = getParam(req, "flow");
     final String jobName = getParam(req, "job");
-    final String propSource = getParam(req, "prop");
+
+    // The name of the file where properties are located
+    final String propsSource = getParam(req, "prop");
```
amielkelinkedin, line 1590: Is there any particular reason that we renamed this? Just curious.

ypadron-in: nope, I guess I assumed that there are multiple properties.

azkaban-common/src/main/java/azkaban/flow/Flow.java
```diff
@@ -293,6 +294,22 @@ public String getId() {
     return this.id;
   }
 
+  /**
+   * Return all flow parents of this flow. From each parent its name and node path are returned.
```
djaiswal83, line 298: Please use javadoc format.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1493,89 +1490,85 @@ private void handleJobPage(final HttpServletRequest req, final HttpServletRespon
...
       if (!hasPermission(project, user, Type.READ)) {
-        throw new AccessControlException("No permission to view project "
-            + projectName + ".");
+        throw new AccessControlException("No permission to view project " + projectName + ".");
       }
-
       page.add("project", project);
-      flow = project.getFlow(flowName);
+
+      final Flow flow = project.getFlow(flowNodePath);
```
djaiswal83, line 1497: It appears that each code block fetches some info and then populates the page object. Can you add 1 liner comment before each of them?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1493,89 +1490,85 @@ private void handleJobPage(final HttpServletRequest req, final HttpServletRespon
...
-        while (parent.getInheritedSource() != null) {
-          source.add(parent.getInheritedSource());
-          parent = flow.getFlowProps(parent.getInheritedSource());
+      page.add("dependents", dependents);
+
+      // Resolve inherited properties
+      final List<Pair<String, String>> allParentFlows = flow.getParents();
+      // List of triplets of NAME and NODE PATH of flows from which properties are
+      // inherited as well as the FILE NAME where they are to be found
+      final List<String[]> inheritedProperties = new ArrayList<>();
```
djaiswal83, line 1554: This is a bug prone logic. Can you please make a private class with such triplets?

djaiswal83 (changes requested):
Thanks for working on this issue. The logic looks solid to me. I have few comments. Can you please address them?

djaiswal83 (approved):
Thanks for addressing the review comments. Looks good to me.

PR #2883, opened 2021-06-22
Intermittent testSetupFlowByMultipleThreads failures

li-ygerchikov:
Turns out that System.currentTimeMillis() is not a safe way to create temporary files.
Also we sometimes fail to delete temp dirs that we create for project downloads.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
@@ -121,9 +121,10 @@ public class FlowPreparer extends AbstractFlowPreparer {
           (flowPrepCompletionTime - criticalSectionStartTime) / 1000,
           flow.getExecutionId(), execDir.getPath());
     } catch (final Exception ex) {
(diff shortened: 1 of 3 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowPreparer.java
```diff
@@ -163,11 +164,10 @@ private String generateProjectDirName(final ProjectDirectoryMetadata proj) {
     return String.valueOf(proj.getProjectId()) + "." + String.valueOf(proj.getVersion());
   }
 
-  private File createTempDir(final ProjectDirectoryMetadata proj) {
+  private File createTempDir(final ProjectDirectoryMetadata proj) throws IOException {
     final String projectDir = generateProjectDirName(proj);
-    final File tempDir = new File(this.projectCacheDir,
-        "_temp." + projectDir + "." + System.currentTimeMillis());
```
li-ygerchikov, line 169: Please, please please don't reinvent filesystem primitives. You'll probably get it wrong. Like we did here.

jakhani (approved):

PR #2885, opened 2021-06-23
Add flow finalization and container cleaup invalid transtions

sshardool:
We observed cases where invalid `AzPodStatus` transitions would
result in corresponding Flows getting stuck in a non-final state
like Running.
This change expicitly adds finalization of flows for such cases
as well as deletion of kubernetes resources for such flows.
As an additional change, we now log the entire watch event for
cases where a flow had to be finalized, this will help determine
the root-cause and allow further analysis of failure scenarios.


azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -130,7 +121,23 @@ private void validateAndPreProcess(AzPodStatusMetadata event) {
...
+    boolean isTransitionValid = validateTransition(event, currentStatus);
+    if (!isTransitionValid) {
+      IllegalStateException transitionException = new IllegalStateException(
+          format("Pod status transition is not supported %s -> %s, for pod %s",
+              currentStatus,
+              event.getAzPodStatus(),
+              event.getPodName()));
+      logger.error("Unsupported state transition.", transitionException);
+      try {
+        finalizeFlowAndDeleteContainer(event);
```
jakhani, line 135: Can you please delete container only for set of invalid transition? We might get pod down due to some statuses missing in state machine we have missed but not be deleting container for that.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -96,23 +96,14 @@ public ExecutorLoader getExecutorLoader() {
    * contain an event for all running flows. This could happen, for example, during webserver
    * restarts.
    * @param event pod watch event
+   * @returns true if transition is valid else false
    */
-  private void validateTransition(AzPodStatusMetadata event) {
-    AzPodStatus currentStatus = getCurrentAzPodStatus(event.getPodName());
+  private boolean validateTransition(AzPodStatusMetadata event, AzPodStatus currentStatus) {
```
hqin2016, line 101: nit: I thought we're put final before the method params, this applies to all methods in this class.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
sshardool: Yes, the deletion  will only happen for the invalid transitions. The logic underlying the `validateTransition` method will take care of ensuring this.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
sshardool: There is an internal ticket for adding `final` for all the params in this (sub)package. I would prefer to defer this change for that ticket as this will also minimize unrelated changes showing up in the interim reviews.

hqin2016 (approved):

jakhani (approved):

PR #654, opened 2016-05-17
Fixes Custom context path

bmsq:
Fixes #653

Exposes jetty.contextPath configuration setting and updates Web Server to serve resources relative to the defined context path setting.

Using the pull request, it is possible to host azkaban at mydomain:8081/azkaban and then use a reverse proxy (such as Apache HTTPD) to server azkaban at mydomain/azkaban along with other web apps on the same server (eg. mydomain/jira).

PR #2888, opened 2021-06-25
Context path in jetty working : )

astrapotro:
Context path in jetty as "jetty.contextPath=/azkaban" .
This changes helps with proxified environments, fixes some problems with that situations (beacause context path is non exisistent until now) , so makes azkaban's exposition in a web server front so much easier.
Based on changes in [this](https://github.com/azkaban/azkaban/pull/654/)


astrapotro (commented):
minor fix

PR #2890, opened 2021-06-30
Fixing intermittently failing ContainerizedDispatchManagerTest

aditya1105:
Various tests in azkaban.executor.ContainerizedDispatchManagerTest were failing with the error:

    org.mockito.exceptions.misusing.WrongTypeOfReturnValue: 
    RegularImmutableMap cannot be returned by selectAndUpdateExecutionWithLocking()
    selectAndUpdateExecutionWithLocking() should return Set
    ***
    If you're unsure why you're getting above error read on.
    Due to the nature of the syntax above problem might occur because:
    1. This exception *might* occur in wrongly written multi-threaded tests.
       Please refer to Mockito FAQ on limitations of concurrency testing.
    2. A spy is stubbed using when(spy.foo()).then() syntax. It is safer to stub spies - 
       - with doReturn|Throw() family of methods. More in javadocs for Mockito.spy() method.

**This failure occurs because:**

the mocked azkaban.executor.ExecutorLoader 
is used to initialize azkaban.executor.container.ContainerizedDispatchManager. 
This ContainerizedDispatchManager is started in the method azkaban.executor.ContainerizedDispatchManagerTest#initializeContainerizedDispatchImpl 
which starts the 
azkaban.executor.container.ContainerizedDispatchManager.QueueProcessorThread

The method QueueProcessorThread#run  calls QueueProcessorThread#processQueuedFlows which tries to execute
executorLoader.selectAndUpdateExecutionWithLocking()

The method selectAndUpdateExecutionWithLocking is not stubbed for the mocked executorLoader. Also as mentioned in the wiki https://github.com/mockito/mockito/wiki/FAQ#is-mockito-thread-safe, mocks are not fully thread-safe.

**Fix**

After checking various test cases in the ContainerizedDispatchManagerTest, the conclusion is that, it is not necessary to "start" the initialized ContainerizedDispatchManager, and hence the processor thread won't be initialized. Only the methods which require testing around QueueProcessorThread, can "start" the ContainerizedDispatchManager instance.

### DIFF
--- a/azkaban-common/src/test/java/azkaban/executor/ContainerizedDispatchManagerTest.java
+++ b/azkaban-common/src/test/java/azkaban/executor/ContainerizedDispatchManagerTest.java
@@ -312,7 +312,6 @@ public class ContainerizedDispatchManagerTest {
   @Test
   public void testSubmitFlowsExceedingMaxConcurrentRuns() throws Exception {
     initializeContainerizedDispatchImpl();
-    this.containerizedDispatchManager.disableQueueProcessorThread();
     this.props.put(ConfigurationKeys.CONCURRENT_RUNS_ONEFLOW_WHITELIST, "exectest1,"
         + "exec2,3");
     submitFlow(this.flow2, this.ref2);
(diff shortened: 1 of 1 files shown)


azkaban-common/src/test/java/azkaban/executor/ContainerizedDispatchManagerTest.java
```diff
@@ -369,6 +367,7 @@ public void testSetFlowLock() throws Exception {
   @Test
   public void testDisablingQueueProcessThread() throws Exception {
     initializeContainerizedDispatchImpl();
+    this.containerizedDispatchManager.start();
```
jakhani, line 370: In disabled, shouldn't we stopping it?

aditya1105: It is initialized in every test, and JUnit takes care of all the objects created in the test, including their threads.

jakhani (approved):
Big thank you for fixing this issue!

PR #2891, opened 2021-07-01
JobRunner may see dangling references in jobProps post PR-2786

haibchen:
Post https://github.com/azkaban/azkaban/pull/2786, after JobRunner.createJobParams() is called, the JobRunner may have dangling references in jobProps that it passed in to createJobParams().

This is because in PR-2786, we added the cluster-specific properties to a new instance of Props, `finalProps`, while still adding properties that may reference cluster-specific ones to `jobProps`. The `finalProps` is passed to Job constructors, which is fine. But `jobProps` might be left with dangling references to cluster-specific properties that can never be resolved.

JobCallbackManager is one place where `jobProps` of the jobRunner is consumed after JobRunner.createJobParams(). When it tries to resolve `jobProps`, it fails and drops the callback on the floor. 

In this change, we  reimplement PR-2786 by always writing to `jobProps` and only add cluster-specific properties if they are not already defined in 'jobProps' to main job properties' precedence.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -387,10 +387,10 @@ public class JobTypeManager {
 
       // inject cluster jars and native libraries into jobs through properties
       Props clusterSpecificProps = getClusterSpecificJobProps(targetCluster, jobProps, pluginLoadProps);
-      for (final String k : clusterSpecificProps.getKeySet()) {
+      for (final String key : clusterSpecificProps.getKeySet()) {
         // User's job props should take precedence over cluster props
-        if (!jobProps.containsKey(k)) {
-          jobProps.put(k, clusterSpecificProps.get(k));
+        if (!jobProps.containsKey(key)) {
+          jobProps.put(key, clusterSpecificProps.get(key));
         }
       }
       Props nonOverriddableClusterProps = getClusterSpecificNonOverridableJobProps(clusterSpecificProps);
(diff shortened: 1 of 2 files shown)


li-ygerchikov (commented):
@haibchen Thank you for coming up with a fix so quickly.
Any chance you could add a unit test for the condition that PR-2786 introduced, and that this change addresses? In other words, reproduce the PR-2786 bug in a unit test? If it makes sense... 
Properties in general is a convoluted and fragile area IMO, so could use more tests in general.

hqin2016:
Is Travis CI test disabled?

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -386,16 +386,20 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
       }
 
       // inject cluster jars and native libraries into jobs through properties
-      // User's job props should take precedence over cluster props
-      Props finalProps = getClusterSpecificJobProps(targetCluster, jobProps, pluginLoadProps);
-      Props nonOverriddableClusterProps = getClusterSpecificNonOverridableJobProps(finalProps);
-      finalProps.putAll(jobProps);
+      Props clusterSpecificProps = getClusterSpecificJobProps(targetCluster, jobProps, pluginLoadProps);
+      for (final String k : clusterSpecificProps.getKeySet()) {
```
djaiswal83, line 390: please use full name "key" instead of k

djaiswal83 (changes requested):
Thanks for fixing the bug. I have one comment. Please address it.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
haibchen: addressed. and also added an explicit unit test

djaiswal83 (approved):
Thank for addressing the comments. The change looks good to me.

PR #2882, opened 2021-06-22
Add metrics for time to dispatch a Kubernetes pod, number of flows dispatched to containers and to bare metal executors

hqin2016:
This is an effort in continuation to https://github.com/azkaban/azkaban/pull/2856 (Added meter metrics for Azkaban pod statuses)

PR #2892, opened 2021-07-02
added flow status for pod failure

hqin2016:
A new state will help identify flow failures due to pod (k8s) related failures.
Needs to account for state showing up correctly on UI
Tested on local dev box solo server and h6 web server.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -249,7 +249,9 @@ public class ExecutionControllerUtils {
       exFlow.setEndTime(time);
     }
 
-    exFlow.setStatus(Status.FAILED);
+    if (exFlow.getStatus() != Status.POD_FAILED) {
+      exFlow.setStatus(Status.FAILED);
+    }
   }
 
   /**
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -249,7 +249,7 @@ public class ExecutionControllerUtils {
       exFlow.setEndTime(time);
     }
 
-    if (exFlow.getStatus() != Status.POD_FAILED) {
+    if (!Status.isStatusFinished(exFlow.getStatus())) {
       exFlow.setStatus(Status.FAILED);
     }
   }
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -85,7 +85,6 @@ public enum Status {
 
   public static boolean isStatusFailed(final Status status) {
     switch (status) {
-      case POD_FAILED:
       case FAILED:
       case KILLED:
       case CANCELLED:
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -30,7 +30,7 @@ public enum Status {
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
-  // EXECUTION_STOPPED refers to a terminating flow status due to crashed executor/container
+  // EXECUTION_STOPPED refers to a terminal flow status due to crashed executor/container
   EXECUTION_STOPPED(65),
   FAILED(70),
   FAILED_FINISHING(80),
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -30,8 +30,8 @@ public enum Status {
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
-  // POD_FAILED refers to a failed containerized flow due to pod failure
-  POD_FAILED(65),
+  // EXECUTION_STOPPED refers to a terminating flow status due to crashed executor/container
+  EXECUTION_STOPPED(65),
   FAILED(70),
   FAILED_FINISHING(80),
   SKIPPED(90),
@@ -59,7 +59,7 @@ public enum Status {
 
   public static boolean isStatusFinished(final Status status) {
     switch (status) {
-      case POD_FAILED:
+      case EXECUTION_STOPPED:
       case FAILED:
       case KILLED:
       case SUCCEEDED:
--- a/azkaban-common/src/main/java/azkaban/executor/Status.java
+++ b/azkaban-common/src/main/java/azkaban/executor/Status.java
@@ -30,6 +30,8 @@ public enum Status {
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
+  // POD_FAILED refers to a failed containerized flow due to pod failure
+  POD_FAILED(65),
   FAILED(70),
   FAILED_FINISHING(80),
   SKIPPED(90),
@@ -57,6 +59,7 @@ public enum Status {
 
   public static boolean isStatusFinished(final Status status) {
     switch (status) {
+      case POD_FAILED:
       case FAILED:
       case KILLED:
       case SUCCEEDED:
@@ -82,6 +85,7 @@ public enum Status {
 
   public static boolean isStatusFailed(final Status status) {
     switch (status) {
+      case POD_FAILED:
       case FAILED:
       case KILLED:
       case CANCELLED:
--- a/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
@@ -98,7 +98,8 @@ public class FlowStatusManagerListener implements AzPodStatusListener {
    * @param event pod watch event
    * @returns true if transition is valid else false
    */
-  private boolean validateTransition(AzPodStatusMetadata event, AzPodStatus currentStatus) {
+  private boolean validateTransition(AzPodStatusMetadata event) {
+    final AzPodStatus currentStatus = getCurrentAzPodStatus(event.getPodName());
     logger.debug(format("Transition requested from %s -> %s, for pod %s",
         currentStatus,
         event.getAzPodStatus(),
@@ -122,12 +123,11 @@ public class FlowStatusManagerListener implements AzPodStatusListener {
     long eventCount = flowContainerEventCount.incrementAndGet();
     logEventCacheStats(eventCount);
 
-    AzPodStatus currentStatus = getCurrentAzPodStatus(event.getPodName());
-    boolean isTransitionValid = validateTransition(event, currentStatus);
+    boolean isTransitionValid = validateTransition(event);
     if (!isTransitionValid) {
       IllegalStateException transitionException = new IllegalStateException(
           format("Pod status transition is not supported %s -> %s, for pod %s",
-              currentStatus,
+              getCurrentAzPodStatus(event.getPodName()),
               event.getAzPodStatus(),
               event.getPodName()));
       logger.error("Unsupported state transition.", transitionException);
@@ -239,6 +239,9 @@ public class FlowStatusManagerListener implements AzPodStatusListener {
           "Flow execution-id %d for pod %s does not have a final status in database and will be "
               + "finalized.", executionId, event.getPodName()));
       final String reason = "Flow Pod execution was completed.";
+      if (!validateTransition(event)) {
+        executableFlow.setStatus(Status.POD_FAILED);
+      }
       ExecutionControllerUtils.finalizeFlow(executorLoader, alerterHolder, executableFlow, reason,
           null);
       // Log event for cases where the flow was not already in a final state
--- a/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
@@ -132,7 +132,7 @@ public class FlowStatusManagerListener implements AzPodStatusListener {
               event.getPodName()));
       logger.error("Unsupported state transition.", transitionException);
       try {
-        finalizeFlowAndDeleteContainer(event, Status.EXECUTION_STOPPED);
+        finalizeFlowAndDeleteContainer(event, Optional.of(Status.EXECUTION_STOPPED));
       } catch (Exception deletionException) {
         transitionException.addSuppressed(deletionException);
       }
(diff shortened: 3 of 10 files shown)


azkaban-common/src/main/java/azkaban/metrics/ContainerizationMetricsImpl.java
```diff
@@ -53,6 +51,9 @@ public void setUp() {
     this.podReady = this.metricsManager.addMeter("Pod-Ready-Meter");
     this.podInitFailure = this.metricsManager.addMeter("Pod-Init-Failure-Meter");
     this.podAppFailure = this.metricsManager.addMeter("Pod-App-Failure-Meter");
+    this.flowSubmitToContainer = this.metricsManager.addMeter("Flow-Submit-to-Container-Meter");
```
aditya1105, line 54: Where are these meters incremented or utilized?

hqin2016: They should have been added in https://github.com/azkaban/azkaban/pull/2882.

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -82,6 +85,7 @@ public static boolean isStatusRunning(final Status status) {
 
   public static boolean isStatusFailed(final Status status) {
     switch (status) {
+      case POD_FAILED:
```
svnarumugam, line 88: In conditional flows, this method is used to check if we need to run other nodes. when we combine the node failures with system-related failure (in this case pod), this will likely affect that feature. It's worth double-checking that conditional flow use case.

aditya1105 (approved):

svnarumugam (dismissed):
LGTM. Thank a lot for addressing the comments, Hao!

azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -93,6 +93,8 @@ public static String formatStatus(final Status status) {
         return "Killing";
       case DISPATCHING:
         return "Dispatching";
+      case POD_FAILED:
```
jakhani, line 96: Shall we rename this status to CONTAINER_FAILED? 
^^ cc: @sshardool

azkaban-web-server/src/main/less/flow.less
```diff
@@ -364,6 +376,10 @@ li.tree-list-item {
       background-position: 48px 0px;
     }
 
+    &.POD_FAILED .icon {
```
jakhani, line 379: Can you please add new status in historypage.vm as well?

azkaban-web-server/src/web/js/azkaban/util/job-status.js
```diff
@@ -16,7 +16,7 @@
 
 var statusList = ["FAILED", "FAILED_FINISHING", "SUCCEEDED", "RUNNING",
   "WAITING", "KILLED", "DISABLED", "READY", "CANCELLED", "UNKNOWN", "PAUSED",
-  "SKIPPED", "QUEUED", "FAILED_SUCCEEDED", "DISPATCHING"];
+  "SKIPPED", "QUEUED", "FAILED_SUCCEEDED", "DISPATCHING", "POD_FAILED"];
```
jakhani, line 19: Can you please validate that Kill link is not coming on Flow execution page and prepare execution link is coming instead? The behavior should be same as any other final state.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -249,7 +249,9 @@ public static void failEverything(final ExecutableFlow exFlow) {
       exFlow.setEndTime(time);
     }
 
-    exFlow.setStatus(Status.FAILED);
+    if (exFlow.getStatus() != Status.POD_FAILED) {
```
sshardool, line 252: Perhaps make this check more generic and switch the status to FAILED if the previous status is already a final one ?
Something like `if (!isStatusFinised(exFlow.getStatus())`

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -30,6 +30,8 @@
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
+  // POD_FAILED refers to a failed containerized flow due to pod failure
+  POD_FAILED(65),
```
sshardool, line 34: We should consider giving this state a more generic name which could be used for flows in similar state in non-containerized Azkban. For instance in the event of a bare-metal executor crashing we could the switch any flows assigned to that executor to this state. 
Something like `EXECUTE_INFRA_FAILED`, better alternatives should be possible though.

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -122,12 +123,11 @@ private void validateAndPreProcess(AzPodStatusMetadata event) {
...
     logEventCacheStats(eventCount);
 
-    AzPodStatus currentStatus = getCurrentAzPodStatus(event.getPodName());
-    boolean isTransitionValid = validateTransition(event, currentStatus);
+    boolean isTransitionValid = validateTransition(event);
     if (!isTransitionValid) {
       IllegalStateException transitionException = new IllegalStateException(
           format("Pod status transition is not supported %s -> %s, for pod %s",
-              currentStatus,
+              getCurrentAzPodStatus(event.getPodName()),
```
sshardool, line 130: Why is this change needed ?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -239,6 +239,9 @@ private void updatePodStatus(AzPodStatusMetadata event) {
           "Flow execution-id %d for pod %s does not have a final status in database and will be "
               + "finalized.", executionId, event.getPodName()));
       final String reason = "Flow Pod execution was completed.";
+      if (!validateTransition(event)) {
```
sshardool, line 242: The method `compareAndFinalizeFlowStatus` should not be invoking `validateTransition` directly, would be better to let this method be narrow in scope which and thereby retain its reusability.  Also note that the check for `validateTransition` has already occurred before this method will be invoked and calling it again in this method could be a little confusing.

I recommend,

1. Add an additional argument `Status finalStatus` to all of these methods - `compareAndFinalizeFlow`, `finalizeFlowAndDeleteContainer`,  `ExecutionControllerUtils.finalizeFlow`. The default value for this argument can be `FAILED` if needed (using overloaded methods with this argument missing)
2. In the method `validateAndPreProcess`, once we have found that the transition is invalid then invoke `finalizeFlowAndDeleteContainer ` with the `finalStatus` argument set to `POD_FAILED`

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -239,6 +239,9 @@ private void updatePodStatus(AzPodStatusMetadata event) {
...
+        executableFlow.setStatus(Status.POD_FAILED);
```
sshardool, line 243: Setting the flow status directly will bypass a number of additional tasks performed by `ExecutionControllerUtils.finalizeFlow`, including alerting users about flow failures. 
As mentioned in the previous comment, please see if the `finalizeFlow` can be updated to accept and additional param for `finalStatus`. We can discuss if that seem problematic for any reason.

sshardool (commented):
Thanks a lot for making this change, this will be very useful.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
sshardool: `CONTAINER_FAILED` certainly seems better than `POD_FAILED` but I was hoping for name that could be use to reflect similar failures for bare-metal executors, e.g when the executor-server crashes. 
Any suggestions @jakhani ?  (or should we not try that at all)

azkaban-common/src/main/java/azkaban/executor/Status.java
ypadron-in: +1 on using a generic name not indicative or coupled to a specific implementation.

ypadron-in (changes requested):
@hqin2016 Could you please enumerate some example scenarios where an execution will finish with the new status?

I may be missing some context here but I feel we're mixing 2 things with the new status: `execution status` and `reason for failure`.

Reason for failure can be anything from network failures to Azkaban server failures to user code failures, we can't possibly list them all.

Why do we think we should create a status that indicates an infrastructure failure? 
What will prevent us in the future to keep adding failed status with associated reason for failure? 
Wouldn't execution status become unreliable and a maintenance nightmare for users who depend on them to do X or Y? For example: today users only need to monitor FAILED status to determine if an execution failed (for whatever reason). With this new status users could miss failed executions if they don't update their code to start monitoring the new status as well.

hqin2016 (commented):
> @hqin2016 Could you please enumerate some example scenarios where an execution will finish with the new status?
> 
> I may be missing some context here but I feel we're mixing 2 things with the new status: `execution status` and `reason for failure`.
> 
> Reason for failure can be anything from network failures to Azkaban server failures to user code failures, we can't possibly list them all.
> 
> Why do we think we should create a status that indicates an infrastructure failure?
> What will prevent us in the future to keep adding failed status with associated reason for failure?
> Wouldn't execution status become unreliable and a maintenance nightmare for users who depend on them to do X or Y? For example: today users only need to monitor FAILED status to determine if an execution failed (for whatever reason). With this new status users could miss failed executions if they don't update their code to start monitoring the new status as well.

The purpose of creating such PR and status is described in BDP-6898. If you have time, would you like to join the next focus group meeting?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -238,7 +240,10 @@ private void updatePodStatus(AzPodStatusMetadata event) {
       logger.info(format(
           "Flow execution-id %d for pod %s does not have a final status in database and will be "
               + "finalized.", executionId, event.getPodName()));
-      final String reason = "Flow Pod execution was completed.";
+      final String reason = "Flow pod execution was completed.";
+      if (finalStatus == Status.EXECUTION_STOPPED) {
```
sshardool, line 244: Looks like the argument `finalStatus` will be meaningful only when its value is `EXECUTION_STOPPED` and will be completely ignored otherwise. 
How about one of the following alternatives:
(1) Don't ignore the argument and if the flow is being finalized then always use the given `finalStatus` 
(2) Change the argument type to `Optional<Status> finalStatus` and use the argument only when `finalStatus.isPresent() == true`

sshardool (dismissed):
Thanks for incorporating the comments!

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -30,6 +30,8 @@
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
+  // EXECUTION_STOPPED refers to a terminating flow status due to crashed executor/container
```
djaiswal83, line 33: terminal flow status?

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -283,11 +288,12 @@ private void deleteFlowContainer(AzPodStatusMetadata event) {
 
   /**
    * Finalize the flow for the given event and delete its container.
-   *
-   * @param event
+   *  @param event
```
djaiswal83, line 291: nit: indentation

djaiswal83 (changes requested):
I have couple of minor comments. Feel free to ignore them if time sensitive.

djaiswal83 (dismissed):

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -30,6 +30,8 @@
   SUCCEEDED(50),
   KILLING(55),
   KILLED(60),
+  // EXECUTION_STOPPED refers to a terminal flow status due to crashed executor/container
+  EXECUTION_STOPPED(65),
```
ypadron-in, line 34: Could the name of the status be confusing to users? It says the execution stopped but is it a temporary stop? Will it continue to run at some point? To me the name does not suggest a final status but a transition one. WDYT @sshardool @jakhani @djaiswal83 @aditya1105

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -132,7 +132,7 @@ private void validateAndPreProcess(AzPodStatusMetadata event) {
               event.getPodName()));
       logger.error("Unsupported state transition.", transitionException);
       try {
-        finalizeFlowAndDeleteContainer(event);
+        finalizeFlowAndDeleteContainer(event, Optional.of(Status.EXECUTION_STOPPED));
```
ypadron-in, line 135: In the case of Pod going down due to OOM caused by user code. Will the execution be marked with status FAILED or EXECUTION_STOPPED? @sshardool @jakhani @djaiswal83 @aditya1105

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -283,11 +288,12 @@ private void deleteFlowContainer(AzPodStatusMetadata event) {
 
   /**
    * Finalize the flow for the given event and delete its container.
-   *
    * @param event
+   * @param finalStatus
    */
-  private void finalizeFlowAndDeleteContainer(AzPodStatusMetadata event) {
-    Optional<Status> originalFlowStatus = compareAndFinalizeFlowStatus(event);
+  private void finalizeFlowAndDeleteContainer(AzPodStatusMetadata event,
```
ypadron-in, line 294: This PR implements the new status for containerized executions only but bare metal executors also crash so we should be consistent and implement it for bare metal executors too. Could you please make the necessary changes?
So far execution failures due to executor crashes has been marked as FAILED. When this PR is merged their final status should change to EXECUTION_STOPPED which affects ALL Azkaban users and potentially break their code. Users will need to modify their code and logic to account for the new status. Are we ok with that? @sshardool @jakhani @djaiswal83 @aditya1105

azkaban-web-server/src/main/java/azkaban/webapp/servlet/WebUtils.java
```diff
@@ -93,6 +93,8 @@ public static String formatStatus(final Status status) {
         return "Killing";
       case DISPATCHING:
         return "Dispatching";
+      case EXECUTION_STOPPED:
+        return "Execution stopped, crashed executor/container";
```
ypadron-in, line 97: How is this message used? Can we remove the "crashed executor/container" part which is an implementation detail?

PR #2893, opened 2021-07-02
Introducing the defaultProxyUser feature for the jobTypes

aditya1105:
### Background
In the current setup, Azkaban allows users to define a “user.to.proxy” property for each Azkaban Job irrespective of the Azkaban JobType. If such a property is defined, Azkaban will ensure that the Azkaban Job process will get executed as the “user.to.proxy” UID.
Azkaban will honour the “user.to.proxy” only when it is also added as part of the “Proxy Users” section under Azkaban project permissions.

### ProblemStatement
Azkaban Job Type owners/developers require their Azkaban Job to always be executed as a default “user.to.proxy”, irrespective of the property defined in the Azkaban Job Properties or the submitting user.

### Solution
The solution is to provide the optional ability to define a default proxy-user mapping for the Job Type. If such a mapping is defined then,

- Azkaban won’t check if the proxy user defined in the mapping is part of the “Proxy Users” in the Azkaban Project Permissions for such Azkaban Job Type.
- Azkaban won’t honor the “user.to.proxy” set by the users in the Azkaban job properties for such Azkaban Job Type.

Constraints,
- Configurable list of users who are not allowed as the default proxy users. Example: root, azkaban, etc.
- Configurable list of job type classes that are allowed for the default proxy users.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -17,6 +17,7 @@
 package azkaban.jobtype;
 
 import azkaban.Constants;
+import azkaban.Constants.PluginManager;
 import azkaban.cluster.Cluster;
 import azkaban.cluster.ClusterRouter;
 import azkaban.cluster.DisabledClusterRouter;
@@ -42,8 +43,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
 public class JobTypeManager {
@@ -159,6 +162,38 @@ public class JobTypeManager {
 
     plugins.setCommonPluginJobProps(commonPluginJobProps);
     plugins.setCommonPluginLoadProps(commonPluginLoadProps);
+
+    if (commonPluginLoadProps
+        .containsKey(Constants.PluginManager.DEFAULT_PROXY_USERS_JOBTYPE_CLASSES)) {
+      plugins.addDefaultProxyUsersJobTypeClasses(commonPluginLoadProps
+          .getStringList(Constants.PluginManager.DEFAULT_PROXY_USERS_JOBTYPE_CLASSES));
+    }
+
+    if (commonPluginLoadProps.containsKey(PluginManager.DEFAULT_PROXY_USERS_FILTER)) {
+      plugins.addDefaultProxyUsersFilter(
+          commonPluginLoadProps.getStringList(PluginManager.DEFAULT_PROXY_USERS_FILTER));
+    }
+
+    // Loads the default-proxy-users mappings for all job types.
+    final File defaultProxyUsersFile = new File(jobPluginsDir,
+        Constants.PluginManager.DEFAULT_PROXY_USERS_FILE);
+    if (!defaultProxyUsersFile.exists()) {
+      LOGGER.info("Default proxy users file " + defaultProxyUsersFile
+          + " not found.");
+    } else {
+      LOGGER.info("Default proxy users file " + defaultProxyUsersFile
+          + " found. Attempting to load.");
+      try {
+        final Props defaultProxyUsers = new Props(null, defaultProxyUsersFile);
+        for (String jobType : defaultProxyUsers.getKeySet()) {
+          plugins.addDefaultProxyUser(jobType, defaultProxyUsers.getString(jobType,
+              StringUtils.EMPTY));
+        }
+      } catch (final IOException e) {
+        throw new JobTypeManagerException(
+            "Failed to load common plugin loader properties" + e.getCause());
+      }
+    }
 
     // Loading job types
     for (final File dir : jobPluginsDir.listFiles()) {
@@ -338,12 +373,12 @@ public class JobTypeManager {
     final JobTypePluginSet pluginSet = getJobTypePluginSet();
 
     try {
-      final String jobType = jobProps.getString("type");
-      if (jobType == null || jobType.length() == 0) {
-        /* throw an exception when job name is null or empty */
+      final Optional<String> jobTypeOptional = getJobType(jobProps);
+      if (!jobTypeOptional.isPresent()) {
         throw new JobExecutionException(String.format(
-            "The 'type' parameter for job[%s] is null or empty", jobProps));
+            "The 'type' parameter for job[%s] is missing or null or empty", jobProps));
       }
+      final String jobType = jobTypeOptional.get();
 
       logger.info("Building " + jobType + " job executor. ");
 
@@ -414,6 +449,15 @@ public class JobTypeManager {
     }
   }
 
+  /**
+   * @param jobProps Properties for an Azkaban Job
+   * @return The {@link Optional} jobType for the Azkaban Job.
+   */
+  public static Optional<String> getJobType(Props jobProps) {
+    final String jobType = jobProps.getString("type", StringUtils.EMPTY);
+    return StringUtils.isNotBlank(jobType) ? Optional.of(jobType) : Optional.empty();
+  }
+
   private static Props getClusterSpecificNonOverridableJobProps(final Props clusterSpecificJobProp) {
     final Props props = new Props();
     final String clusterClasspath =
@@ -452,7 +496,7 @@ public class JobTypeManager {
     return jobProps;
   }
 
-  private static Props getPluginLoadProps(JobTypePluginSet pluginSet, String jobType) {
+  static Props getPluginLoadProps(JobTypePluginSet pluginSet, String jobType) {
     Props pluginLoadProps = pluginSet.getPluginLoaderProps(jobType);
     if (pluginLoadProps != null) {
       pluginLoadProps = PropsUtils.resolveProps(pluginLoadProps);
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
@@ -19,7 +19,12 @@ import azkaban.utils.Props;
 
 import java.net.URL;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Container for job type plugins
@@ -37,7 +42,14 @@ public class JobTypePluginSet {
   private final Map<String, Props> pluginPrivatePropsMap;
   private final Map<String, JobPropsProcessor> pluginJobPropsProcessor;
   private final Map<String, String> jobToClassName;
+  // Mapping for jobType to defaultProxyUser
+  private final Map<String, String> jobToDefaultProxyUser;
   private final Map<String, URL[]> jobToClassLoaderURLs;
+
+  // List of jobType classes which are allowed for the defaultProxyUsers feature
+  private final Set<String> defaultProxyUsersJobTypeClasses;
+  // List of users which are not allowed as defaultProxyUsers
+  private final Set<String> defaultProxyUsersFilter;
 
   private Props commonJobProps;
   private Props commonLoadProps;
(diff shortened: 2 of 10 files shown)


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -160,6 +163,39 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
...
+      plugins.addDefaultProxyUsersFilter(
+          commonPluginJobProps.getStringList(PluginManager.DEFAULT_PROXY_USERS_FILTER));
+    }
+
+    // Loads the default-proxy-users mappings for all job types.
+    final File defaultProxyUsersFile = new File(jobPluginsDir,
+        Constants.PluginManager.DEFAULT_PROXY_USERS_FILE);
+    if (defaultProxyUsersFile.exists()) {
+      LOGGER.info("Default proxy users file " + defaultProxyUsersFile
+          + " found. Attempt to load.");
```
amielkelinkedin, line 182: Can we change "Attempt to load" to "Attempting to load"?

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
```diff
@@ -171,4 +188,88 @@ public void addPluginJobPropsProcessor(final String jobTypeName,
       JobPropsProcessor jobPropsProcessor) {
     this.pluginJobPropsProcessor.put(jobTypeName, jobPropsProcessor);
   }
+
+  /**
+   * @return The list of users which are not allowed as defaultProxyUsers.
+   */
+  public List<String> getDefaultProxyUsersFilter() {
+    return this.defaultProxyUsersFilter;
```
amielkelinkedin, line 196: Can we be consistent with our `return` use?

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerTest.java
```diff
@@ -347,4 +352,31 @@ public void testResetPlugins() throws Exception {
     assertEquals("3", ntjobProps.get("pluginprops3"));
     assertEquals("pluginprops", ntjobProps.get("commonprop3"));
   }
+
+  @Test
```
amielkelinkedin, line 356: Thank you for adding tests :)

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -190,8 +191,50 @@ public Props getProps() {
...
+    if (jobType.isPresent()) {
+      // JobTypePluginSet is never NULL
+      Optional<String> defaultProxyUser = this.jobtypeManager.getJobTypePluginSet()
+          .getDefaultProxyUser(jobType.get());
+      if (defaultProxyUser.isPresent()) {
+        this.logger.info(String.format("Found default proxy user %s for the job type %s",
+            defaultProxyUser.get(), jobType.get()));
+        return defaultProxyUser;
+      }
+    }
```
amielkelinkedin, line 218: Should this be factored out into its own method?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -741,23 +784,13 @@ private Status prepareJob() throws RuntimeException {
...
-        }
-      } else {
-        final String submitUser = this.getNode().getExecutableFlow().getSubmitUser();
-        this.props.put(JobProperties.USER_TO_PROXY, submitUser);
-        this.logger.info("user.to.proxy property was not set, defaulting to submit user " +
-            submitUser);
+      Optional<String> effectiveUserOptional = determineEffectiveUser();
+      // Fail if the effective user is not present
+      if (!effectiveUserOptional.isPresent()) {
+        return null;
```
amielkelinkedin, line 790: It seems strange to me to do so much work with Optionals only to return null anyway :).

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -160,6 +163,39 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
...
+
+    if (commonPluginJobProps.containsKey(PluginManager.DEFAULT_PROXY_USERS_FILTER)) {
+      plugins.addDefaultProxyUsersFilter(
+          commonPluginJobProps.getStringList(PluginManager.DEFAULT_PROXY_USERS_FILTER));
+    }
+
+    // Loads the default-proxy-users mappings for all job types.
+    final File defaultProxyUsersFile = new File(jobPluginsDir,
+        Constants.PluginManager.DEFAULT_PROXY_USERS_FILE);
+    if (defaultProxyUsersFile.exists()) {
```
djaiswal83, line 180: suggestion : Since the negative case is small, switch the if else block with if containing negative condition, that way while reading code, it is out of the way and logic is more readable.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -160,6 +163,39 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
...
+      LOGGER.info("Default proxy users file " + defaultProxyUsersFile
+          + " found. Attempt to load.");
+      try {
+        Props defaultProxyUsers = new Props(null, defaultProxyUsersFile);
```
djaiswal83, line 184: Use final keyword wherever applicable.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -410,6 +446,18 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
     }
   }
 
+  /**
+   * @param jobProps Properties for an Azkaban Job
+   * @return The {@link Optional} jobType for the Azkaban Job.
+   */
+  public static Optional<String> getJobType(Props jobProps) {
+    final String jobType = jobProps.getString("type", StringUtils.EMPTY);
+    if (StringUtils.isNotBlank(jobType)) {
```
djaiswal83, line 455: Can you make it a single line statement "return StringUtils.isNotBlank(jobType) ? Optional.of(jobType) : empty()" ?

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
```diff
@@ -37,8 +41,15 @@
...
   private final Map<String, JobPropsProcessor> pluginJobPropsProcessor;
   private final Map<String, String> jobToClassName;
+  // Mapping for jobType to defaultProxyUser
+  private final Map<String, String> jobToDefaultProxyUser;
   private final Map<String, URL[]> jobToClassLoaderURLs;
 
+  // List of jobType classes which are allowed for the defaultProxyUsers feature
+  private final List<String> defaultProxyUsersJobTypeClasses;
+  // List of users which are not allowed as defaultProxyUsers
+  private final List<String> defaultProxyUsersFilter;
```
djaiswal83, line 51: the name is a bit confusing, perhaps, defaultProxyUserDenyList? This will also mean changes everywhere it is used, provided you keep it list.
More importantly, List is not appropriate data structure, since it is meant to filter out certain users, in such case a Set is more appropriate, then perhaps, the name could be kept as it is.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
djaiswal83: +1

djaiswal83 (changes requested):
The logic is good overall. I believe you should be storing these properties in private.properties instead of plugin.properties.
I have few comments. Please address them.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
aditya1105: done

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
aditya1105: Sorry I didn't get it. Can you explain?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
aditya1105: Thanks for this. The usage of optional instead of null is to have safety again the possible NPE. 
However, this method prepareJob() is either returned null or as Status.RUNNING. Hence I have changed it to have boolean as return type, and updated the place where it is called.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -160,6 +163,39 @@ private void loadPluginJobTypes(final JobTypePluginSet plugins)
     plugins.setCommonPluginJobProps(commonPluginJobProps);
     plugins.setCommonPluginLoadProps(commonPluginLoadProps);
 
+    if (commonPluginJobProps
```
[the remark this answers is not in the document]
aditya1105: The place where we have used is not the place where it can be overridden. It gets overridden when passed to the JobType class.
However, commonPluginLoadProps seems more appropriate in this case as these are indeed load-time props. Hence changed it to read from private properties.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
aditya1105: done

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
aditya1105: done

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
aditya1105: done

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
aditya1105: Make sense. I will utilize set everywhere in this PR where List is a bad data structure.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
aditya1105: There is not much calculation to it apart from returning the value and logging it, hence refactoring may not be needed here. The actual calculation of defaultProxyUser is already part of the method azkaban.jobtype.JobTypePluginSet#getDefaultProxyUser

amielkelinkedin: Fair enough.

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
amielkelinkedin: I might be mistaken, but it looks like we have some methods which are explicitly returning like this one, and some which are not like addDefaultProxyUsersFilter. I didn't even know you didn't have to explicitly return! I thought that was only a groovy thing, not a Java thing.

aditya1105: Those which are return type void like "addDefaultProxyUsersFilter", need not have explicit return to be mentioned. And unlike scala where the last line is automatically returned, java needs an explicit return if the return type is not void.

amielkelinkedin: Sorry, missed that it was void, thanks.

amielkelinkedin (approved):

azkaban-common/src/main/java/azkaban/jobtype/JobTypePluginSet.java
```diff
@@ -37,8 +42,15 @@
   private final Map<String, Props> pluginPrivatePropsMap;
   private final Map<String, JobPropsProcessor> pluginJobPropsProcessor;
   private final Map<String, String> jobToClassName;
+  // Mapping for jobType to defaultProxyUser
+  private final Map<String, String> jobToDefaultProxyUser;
   private final Map<String, URL[]> jobToClassLoaderURLs;
 
+  // List of jobType classes which are allowed for the defaultProxyUsers feature
```
djaiswal83, line 49: Nit : Comment still says List. It should be Set.

djaiswal83 (approved):
Overall looks good. I see that the change was force committed making it harder to figure out.
The new changes in JobRunner look good to me to.
I have a minor nit.

PR #2896, opened 2021-07-12
Adding Image Type Get Request

amielkelinkedin:
This is the change necessary for adding a GET request for ImageTypes which will be used for proxy authentication in Jenkins when automating Image Management.

Tested on Server6. Added an azkaban user to my local web-server iteration to test Admin calls.

curl <server>/imageTypes?session.id=<session.id>
curl <server>/imageTypes/{imageTypeName}?session.id=<session.id>
curl <server>/imageTypes/{id}?session.id=<session.id>
curl <server>/imageTypes?imageType={imageTypeName}&session.id=<session.id>


azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
       throws ServletException, IOException {
     try {
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+      ImageTypeDTO imageTypeDTO = null;
+      if (StringUtils.isNotBlank(req.getPathInfo())) {
```
djaiswal83, line 86: Will there be a scenario when the pathInfo can be whitespaces? if not, then we do not need to use this API and instead use String's isEmpty API.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
       throws ServletException, IOException {
     try {
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+      ImageTypeDTO imageTypeDTO = null;
```
djaiswal83, line 85: Can you please add the format of path name with an example for easier visualization.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
```diff
@@ -60,6 +61,17 @@ public ImageTypeServiceImpl(final ImageTypeDao imageTypeDao,
...
   }
 
+  @Override
+  public ImageTypeDTO findImageTypeWithOwnersByName(String imageTypeName) throws ImageMgmtException {
+    Optional<ImageType> it = this.imageTypeDao.getImageTypeWithOwnershipsByName(imageTypeName);
+    if (it.isPresent()) {
+      return this.converter.convertToApiResponseDTO(it.get());
+    }
+    else {
+      throw new ImageMgmtException("No Image Type Exists for ImageType name " + imageTypeName);
```
jakhani, line 71: Can you please log this exception?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
jakhani, line 86: Why request URI check is not needed for this case?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
...
+      if (StringUtils.isNotBlank(req.getPathInfo())) {
+        String imageTypeName = req.getPathInfo().substring(INDEX_AFTER_LEADING_SLASH);
+        if (imageTypeName.contains(PATH_SEPARATOR)) {
+          throw new ImageMgmtInvalidInputException(
+              "The path provided is not supported by this API. Please check the documentation.");
+        }
+        if (hasImageManagementPermission(imageTypeName, session.getUser(), Type.GET)) {
```
jakhani, line 92: Cases should be like:
1. Proxy user is provided: Validate the logged in user has permission for the image type and proxy user also has permission for the image type (Proxy user can be either Azkaban admin or image type owner.)
2. Proxy user is not provided: Check logged in user has permission to get the image type. Validate from Azkaban admin side as well as image type ownership.
If proxy user is specified and logged in user has permission then it will not go in else if case which is wrong.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
```diff
@@ -75,6 +78,20 @@ public void testCreateImageType() throws Exception {
...
+    final ImageType imageType = new ImageType();
+    when(this.imageTypeDao.getImageTypeWithOwnershipsByName(any(String.class))).thenReturn
+   ((Optional<ImageType>) Optional.of(imageType));
+    this.imageTypeService.findImageTypeWithOwnersByName("anyString");
+  }
+
+  @Test(expected = ImageMgmtException.class)
+  public void testFindImageTypeWithOwnersByNameFailsWithImageMgmtException() throws Exception {
+    when(this.imageTypeDao.getImageTypeWithOwnershipsByName(any(String.class))).thenReturn(Optional.empty());
+    this.imageTypeService.findImageTypeWithOwnersByName("anyString");
```
jakhani, line 92: This test is not useful. Is it possible to create some image types with owners and validate if you can get image type?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
       throws ServletException, IOException {
     try {
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+      ImageTypeDTO imageTypeDTO = null;
+      if (StringUtils.isNotBlank(req.getPathInfo())) {
+        String imageTypeName = req.getPathInfo().substring(INDEX_AFTER_LEADING_SLASH);
```
djaiswal83, line 87: Please add comments here.

djaiswal83 (changes requested):
Thanks for this PR. The logic is looking good. Thanks for adding unit tests.
I have few comments, please address those.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -44,15 +47,19 @@
  * This servlet exposes the REST APIs such as create, get etc. for image type. Below are the
  * supported APIs. Create Image Type API: POST /imageTypes?session.id=? --data @payload.json Get
  * Image Type API: GET /imageTypes?session.id=?&imageType=? GET /imageTypes/{id}?session.id=?
+ * GET /imageTypes/{imageType}?sessionid=? GET /imageTypes/{imageType}?proxyUser=?&session.id=?
```
sshardool, line 50: What prevents someone from specifying an arbitrary `proxyUser` or will there be some additional checks ?

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
```diff
@@ -60,6 +61,17 @@ public ImageTypeServiceImpl(final ImageTypeDao imageTypeDao,
     this.converter = converter;
   }
 
+  @Override
+  public ImageTypeDTO findImageTypeWithOwnersByName(String imageTypeName) throws ImageMgmtException {
+    Optional<ImageType> it = this.imageTypeDao.getImageTypeWithOwnershipsByName(imageTypeName);
```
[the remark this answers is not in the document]
amielkelinkedin: I think variables with small scopes should have small names while variables with large scopes should have longer, more descriptive names. I will change it if you still disagree. Thanks!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
       throws ServletException, IOException {
     try {
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+      ImageTypeDTO imageTypeDTO = null;
+      if (StringUtils.isNotBlank(req.getPathInfo())) {
+        String imageTypeName = req.getPathInfo().substring(INDEX_AFTER_LEADING_SLASH);
+        if (imageTypeName.contains(PATH_SEPARATOR)) {
```
sshardool, line 88: Why is an explicit check needed for the path separator in the imageTypeName and are imageTypeNames even allowed to have a separator ?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: Good Point! Thank you!

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
amielkelinkedin: My understanding is that we'd then be testing the logic behind the imageTypeDao.getImageTypeWithOwnershipsByName(String s) method. Does that sound right to you? I can certainly add those tests to the right package if you'd like! Thank you.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +82,36 @@ protected void handleGet(
...
+          throw new ImageMgmtInvalidInputException(
+              "The path provided is not supported by this API. Please check the documentation.");
+        }
+        if (hasImageManagementPermission(imageTypeName, session.getUser(), Type.GET)) {
+          imageTypeDTO = this.imageTypeService.findImageTypeWithOwnersByName(imageTypeName);
+        } else if (hasParam(req, "proxyUser")) {
```
sshardool, line 94: Is a similar support for proxyUsers required for other parts of the ImageManagement APIs as well or is current set of changes enough for Jenkins/CRT needs ?

sshardool (commented):
Thanks a lot for the PR, had a few questions.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: My understanding is security through obscurity for native calls to the API itself. Through Jenkins, we will be constructing the API call from the User who clicks the button (as the User is passed from Jenkins to the Python job). I hope that answers your question?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: I don't believe this is required for anything else, but I think there is hope that this is value added for ImageManagement Users.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: req.getPathInfo() will return whatever is after the imageTypes/ and before the first parameter, IE the "?". So if someone calls /imageType/a/imageType?proxyUser=proxyUser, we will get /a/imageType for req.getPathInfo(). This is clearly not a valid imageType. So we can either have execution continue, and have it fail gracefully do to that image type not being in the database, or we can check directly. Do you have a preference?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
sshardool: 'the User who clicks the button' will be the user specified in the 'proxyUser' parameter, is that correct ?
I would recommend making the security/privileges for this more robust as otherwise this could be open to accidental abuse and would be a more serious concern if this paradigm was extended to more APIs. 
One way of doing this could be to have an allow-list (perhaps in config files) that grants these proxy/impersonation privileges to a specific set of users (which would be the user Jenkins uses and the one we fetch the session-id for). 

In either case I think this warrants a broader discussion with the team (we can do this offline)

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
sshardool: Got it, thanks for the explanation. For this case I would recommend having the check refactored into a utility method, something like `validatedImageTypeName(String name)` 
That will allow (1) easy updates/additions to the checks (2) reusability of the same from other parts of the code.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
ypadron-in: Forgive my lack of context but could you provide an example scenario where the additional proxy user parameter needs to be supported? What purpose is it serving?  
Why can't we use headless accounts to make the API call? In that case the logged in user is the headless account and we just need to verify permissions on one user, as opposed to logging in as ME providing proxy-user=XYZ where 2 user permissions would need to be checked.

ypadron-in (changes requested):

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
sshardool: Wondering if something similar is needed for updating rampup plans ? (Please ignore, if that's not in the scope for this work)

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: > 'the User who clicks the button' will be the user specified in the 'proxyUser' parameter, is that correct ?

Yes, that is correct.

The code as it is written now, will make a request from Jenkins using the az_test user, which I believe is an admin on all servers. Therefore, at least for Jenkins, I don't think the proxyUser logic will be utilized, and another check will be necessary to make sure the user passed from Jenkins is authorized to make changes. I believe the proxyUser logic is solely for calls into the API by users. @jakhani, can you confirm?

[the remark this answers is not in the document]
amielkelinkedin: I think the answer to your question is that we don't necessarily know the ID of the imageType when we're making calls from Jenkins, so we need to find by name. So we can ask /imageTypes?imageName=<imageName>, which might fit better with patterns that exist already, or we can ask /imageTypes/nameOfImageType. I think the issue with /imageTypes?imageName=imageName is that you're left with a top level object. So if you need to know more about that object in the future, you have to do strange things. In terms of the difference between imageTypes/{id} and imageTypes/imageTypeName, the difference is finding by id in the database which requires knowledge of the id in the database, verses knowing the image name itself, which doesn't require knowledge of the database. Thanks!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -44,15 +47,19 @@
  * This servlet exposes the REST APIs such as create, get etc. for image type. Below are the
  * supported APIs. Create Image Type API: POST /imageTypes?session.id=? --data @payload.json Get
  * Image Type API: GET /imageTypes?session.id=?&imageType=? GET /imageTypes/{id}?session.id=?
+ * GET /imageTypes/{imageType}?sessionid=? GET /imageTypes/{imageType}?proxyUser=?&session.id=?
  */
```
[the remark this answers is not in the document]
amielkelinkedin: Thank you for the comment! My understanding is that this request will be sent via a python script invoked by Jenkins, and therefore will not go through a browser, and perhaps therefore not be cached? I guess it is bad to have that kind of optionality on a GET request, but for what we're trying to do Jenkins side, I'm not sure how we get around using a session id given the patterns we're currently using. I will try to avoid this kind of thing going forward.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
ypadron-in: Forgive my lack of context but could you provide an example scenario where the additional proxy user parameter needs to be supported? What purpose is it serving?
Why can't we use headless accounts to make the API call? In that case the logged in user is the headless account and we just need to verify permissions on one user, as opposed to logging in as ME providing proxy-user=XYZ where 2 user permissions would need to be checked.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: My understanding is that req.getRequestURI() will match the entire path, which in this case is variable. IE it would return /imageTypes/imageTypeNameA for GET <server>/imageTypes/imageTypeNameA and /imageType/ImageTypeNameB for GET <server>/imageTypes/imageTypeNameB. I will confirm when I test again after making changes.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
ypadron-in: I see, thank you for the quick response.
Is the ID an alphanumeric string or the image type name?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
ypadron-in: A browser is just one stop in the communication chain, there are DNS servers, proxies, etc. which can also cache the requests. 
Please remember Azkaban is open source, LinkedIn use cases should not be the only ones considered when implementing a change.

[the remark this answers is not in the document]
ypadron-in: I'm not suggesting to drop the use of a session id, I'm just saying that it should not be sent as a query parameter but as a cookie.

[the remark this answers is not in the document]
amielkelinkedin: Thank you for the clarification, thank makes sense!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
amielkelinkedin: My understanding was that this is supposed to be an alphanumeric string, but I could be wrong.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
```diff
@@ -60,6 +61,17 @@ public ImageTypeServiceImpl(final ImageTypeDao imageTypeDao,
...
+    if (it.isPresent()) {
+      return this.converter.convertToApiResponseDTO(it.get());
+    }
+    else {
```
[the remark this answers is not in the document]
jakhani: Please apply Azkaban code style and save actions.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
jakhani: @sshardool @ypadron-in ,
Logic will be like:
Case 1: Proxy user param is not passed:
Check that logged in user has permission on Image type. If yes, then return image type dto. If not then throw exception.
Case 2: Proxy user param is passed:
Check that logged in user has permission on Image type. If yes, then check that proxy user has permission on image type. If yes then return image type DTO. If not then throw exception.

Logged in user can be Azkaban admin and willing to verify whether proxy user can access any image type or not. Here proxy user can be either Azkaban admin or image type owner.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
jakhani: Agree with @ypadron-in . Let's have separate ticket to remove session id from query param. It might require us to inform internal API users as well as currently they might be using session id in query param for various APIs.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
jakhani: @sshardool We are adding this API to validate the proxyUser has permission on any image type. We are validating that using this API from our CICD and then allowing to call any other APIs like create/update/delete ramp up plan or create image version.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
jakhani: @ypadron-in I have shared answer for this question in another thread with "Logic will be like:" as starting. Please let me know if you have any other concerns.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
sshardool: Thanks, that clarifies things. Looking at the PR I got the impression that we are attempting to add impersonation as part of the API itself, hence the multiple questions. I believe a more accurate description is that its updating the existing `imageType` API endpoint to return information about users other than the currently logged in user (i.e the owner of session-id). Which seems fine as long as the executing user (with session-id) has privileges for doing this.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
amielkelinkedin: I have added the tests I believe you wanted in ImageTypeDaoTest. Let me know if that works for you.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -86,31 +82,22 @@ protected void handleGet(
...
-          String imageTypeName = req.getPathInfo().substring(INDEX_AFTER_LEADING_SLASH);
-          if (imageTypeName.contains(PATH_SEPARATOR)) {
-            throw new ImageMgmtInvalidInputException(PATH_NOT_SUPPORTED);
-          }
-          if(!hasImageManagementPermission(imageTypeName, session.getUser(), Type.GET)){
+      ImageTypeDTO imageTypeDTO;
+      if (IMAGE_TYPE_WITH_NAME_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+        String imageTypeName = templateVariableToValue.get(IMAGE_TYPE_NAME_KEY);
+        if (!hasImageManagementPermission(imageTypeName, session.getUser(), Type.GET)) {
+          throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, FORBIDDEN_USER);
```
djaiswal83, line 89: Log line

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -43,16 +45,21 @@
...
+ * /imageTypes/{imageType}?owner=?
  */
 public class ImageTypeServlet extends LoginAbstractAzkabanServlet {
 
-  private static final String GET_IMAGE_TYPE_URI = "/imageTypes";
-  private static final String IMAGE_TYPE_ID_KEY = "id";
-  private static final UriTemplate IMAGE_TYPE_WITH_ID_URI_TEMPLATE = new UriTemplate(
-      String.format("/imageTypes/{%s}", IMAGE_TYPE_ID_KEY));
+  private static final String IMAGE_TYPE_NAME_KEY = "name";
+  private static final String FORBIDDEN_USER = "The user does not have appropriate permissions to"
```
djaiswal83, line 54: nit : FORBIDDEN_USER_ERR_MSG?

djaiswal83 (changes requested):
Thanks for making the changes! The tests are much better.
I have few cosmetic comments related to logging.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -153,6 +153,10 @@ public int createImageType(final ImageType imageType) {
       imageTypes = this.databaseOperator
           .query(FetchImageTypeHandler.FETCH_IMAGE_TYPE_BY_NAME, fetchImageTypeHandler,
               name.toLowerCase());
+      if (imageTypes == null) {
+        throw new ImageMgmtDaoException(ErrorCode.NOT_FOUND,
+            "Unable to fetch image type metadata from image type : " + name);
```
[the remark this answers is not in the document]
amielkelinkedin: That is correct, thank you :)

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -153,6 +153,10 @@ public int createImageType(final ImageType imageType) {
       imageTypes = this.databaseOperator
           .query(FetchImageTypeHandler.FETCH_IMAGE_TYPE_BY_NAME, fetchImageTypeHandler,
               name.toLowerCase());
+      if (imageTypes == null) {
+        throw new ImageMgmtDaoException(ErrorCode.NOT_FOUND,
```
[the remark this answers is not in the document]
amielkelinkedin: Will do, thanks!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: Thank you!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: Sure! Thanks!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: I feel like this is no longer necessary, as the code has been modified to follow the regular format.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -43,16 +45,20 @@
 /**
  * This servlet exposes the REST APIs such as create, get etc. for image type. Below are the
  * supported APIs. Create Image Type API: POST /imageTypes?session.id=? --data @payload.json Get
- * Image Type API: GET /imageTypes?session.id=?&imageType=? GET /imageTypes/{id}?session.id=?
+ * Image Type API: GET /imageTypes?&imageType=? GET /imageTypes/{imageType}
```
jakhani, line 48: Nit: Extra & in GET /imageTypes?&imageType=?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +81,18 @@ protected void handleGet(
...
-        // TODO: Implementation will be provided in the future PR
+      ImageTypeDTO imageTypeDTO;
+      if (IMAGE_TYPE_WITH_NAME_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+        String imageTypeName = templateVariableToValue.get(IMAGE_TYPE_NAME_KEY);
+        if (!hasImageManagementPermission(imageTypeName, session.getUser(), Type.GET)) {
+          log.info(FORBIDDEN_USER_ERR_MSG);
+          throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, FORBIDDEN_USER_ERR_MSG);
+        }
+        imageTypeDTO = this.imageTypeService.findImageTypeWithOwnersByName(imageTypeName);
+        sendResponse(resp, HttpServletResponse.SC_OK, imageTypeDTO);
```
jakhani, line 92: Just want to double check, will you be validating owner from python side of API calls?

amielkelinkedin: Yes, the code for that is ready, but I haven't created the RB, because I wanted this merged first. I can create that and show you!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: I believe one is a path parameter and one is a filter parameter. Unless we don't want to do that?

ypadron-in: Is the name of an image type unique? 
If not then the URI `/imageTypes/{imageType}` can represent 2 or more resources at the same time which is ambiguous and incorrect.
If yes then supporting the URI `/imageTypes/{imageType}` means that the name cannot be changed probably not even to correct typos because that would change the URI and break any code using the old one. How are we planning to support renames in that case?
Why `GET /imageTypes?&imageType=?` isn't enough for this use case?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +81,18 @@ protected void handleGet(
       throws ServletException, IOException {
     try {
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
-        // TODO: Implementation will be provided in the future PR
-      } else if (GET_IMAGE_TYPE_URI.equals(req.getRequestURI())) {
-        // TODO: Implementation will be provided in the future PR
+      ImageTypeDTO imageTypeDTO;
+      if (IMAGE_TYPE_WITH_NAME_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
```
ypadron-in, line 85: The comment in the class indicates that the endpoint `GET /imageTypes?&imageType=?` is supported could you point to its implementation?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -116,7 +131,7 @@ protected void handlePost(
       // prepare to send response
       resp.setStatus(HttpStatus.SC_CREATED);
       resp.setHeader("Location",
-          IMAGE_TYPE_WITH_ID_URI_TEMPLATE.createURI(imageTypeId.toString()));
+          IMAGE_TYPE_WITH_NAME_URI_TEMPLATE.createURI(imageType));
```
ypadron-in, line 134: Same comment as above, using the name of an image type instead of its ID in the URI to obtain resource details forces the name to be unique. How are we planning to support renames if a user made a typo while creating the image type or the name of the jobtype changed.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
```diff
@@ -75,9 +77,25 @@ public void testCreateImageType() throws Exception {
     Assert.assertEquals(100, imageTypeId);
   }
 
+  @Test
+  public void testFindImageTypeWithOwnersByName() throws Exception {
+    final ImageType imageType = new ImageType();
+    when(this.imageTypeDao.getImageTypeWithOwnershipsByName(any(String.class))).thenReturn
+        ((Optional<ImageType>) Optional.of(imageType));
+    this.imageTypeService.findImageTypeWithOwnersByName("anyString");
+  }
```
ypadron-in, line 86: Are we missing an assert here somewhere?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,10 +81,18 @@ protected void handleGet(
...
+          log.info(FORBIDDEN_USER_ERR_MSG);
+          throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, FORBIDDEN_USER_ERR_MSG);
+        }
+        imageTypeDTO = this.imageTypeService.findImageTypeWithOwnersByName(imageTypeName);
+        sendResponse(resp, HttpServletResponse.SC_OK, imageTypeDTO);
+      } else {
+        log.info(PATH_NOT_SUPPORTED);
+        throw new ImageMgmtInvalidInputException(PATH_NOT_SUPPORTED);
       }
     } catch (final Exception e) {
```
ypadron-in, line 97: If an authorization error occurred we should return status code 403, for internal server errors the correct code is 500. However, the current implementation will always return 404 regardless of the nature of the error which is misleading. Could you please fix this?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: The name needs to be unique. Renames should be possible with a post/put request.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: Image type name should be unique. Updates should be possible for renaming with put/post. In regards to the URI, my understanding is that renaming for a filter parameter and a path parameter would break things equally. The users don't have access to IDs, so using an ID is unrealistic.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: It is not currently supported. I will remove it. Thank you.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
amielkelinkedin: I can make this test a little better, thank you!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: That's a very good point, yes.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,17 +88,71 @@ protected void handleGet(
...
       final Map<String, String> templateVariableToValue = new HashMap<>();
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
-        // TODO: Implementation will be provided in the future PR
+      if (IMAGE_TYPE_NAME_OR_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+        getImageTypeDTOByIdOrImageTypeName(resp, session, templateVariableToValue);
       } else if (GET_IMAGE_TYPE_URI.equals(req.getRequestURI())) {
-        // TODO: Implementation will be provided in the future PR
+        ImageTypeDTO imageTypeDTO;
+        if (!StringUtils.isEmpty(req.getQueryString()) && req.getQueryString()
+            .contains(IMAGE_TYPE)) {
```
ypadron-in, line 96: Is the `StringUtils.isEmpty()` check needed? Can `req.getQueryString()` return null?

amielkelinkedin: No, I don't believe it is. I will remove it. Thanks.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
```diff
@@ -60,6 +62,34 @@ public ImageTypeServiceImpl(final ImageTypeDao imageTypeDao,
...
+  public ImageTypeDTO findImageTypeWithOwnershipsByName(String imageTypeName)
+      throws ImageMgmtException {
+    Optional<ImageType> imageType =
+        this.imageTypeDao.getImageTypeWithOwnershipsByName(imageTypeName);
+    if (imageType.isPresent()) {
+      return this.converter.convertToApiResponseDTO(imageType.get());
+    } else {
+      log.info("No Image Type Exists for ImageType name " + imageTypeName);
+      throw new ImageMgmtException("No Image Type Exists for ImageType name " + imageTypeName);
+    }
```
ypadron-in, line 90: I think `throw new ImageMgmtException("No Image Type Exists for ImageType name " + imageTypeName);` will generate a Bad Request response instead of a Not Found one. Could you check?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -122,6 +122,32 @@ public int createImageType(final ImageType imageType) {
...
   }
 
+  @Override
+  public ImageType getImageTypeWithOwnershipsById(final String id) {
+    final FetchImageTypeHandler fetchImageTypeHandler = new FetchImageTypeHandler();
+    List<ImageType> imageTypes = new ArrayList<>();
+    try {
+      imageTypes = this.databaseOperator.query(FetchImageTypeHandler.FETCH_IMAGE_TYPE_BY_ID,
+          fetchImageTypeHandler, id);
+      if (imageTypes == null || imageTypes.isEmpty()) {
```
ypadron-in, line 132: Is the null check needed? I think we should guarantee a non null result.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -122,6 +122,32 @@ public int createImageType(final ImageType imageType) {
...
+      }
+      if (imageTypes != null && imageTypes.size() > 1) {
+        throw new ImageMgmtDaoException(ErrorCode.UNPROCESSABLE_ENTITY, "The request for image "
+            + "type with id " + id + " unexpectedly returned more than one result!");
+      }
+    } catch (final SQLException ex) {
+      log.error(FetchImageTypeHandler.FETCH_IMAGE_TYPE_BY_ID + " failed.", ex);
+      throw new ImageMgmtDaoException(ErrorCode.NOT_FOUND,
+          "Unable to get image type for : " + id);
+    }
```
ypadron-in, line 144: If a SQLException occurred then I think we should return an Internal Server (500) error response to the user.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -153,8 +179,16 @@ public int createImageType(final ImageType imageType) {
       imageTypes = this.databaseOperator
           .query(FetchImageTypeHandler.FETCH_IMAGE_TYPE_BY_NAME, fetchImageTypeHandler,
               name.toLowerCase());
+      if (imageTypes == null) {
```
ypadron-in, line 182: A non-null result is not guaranteed?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -153,8 +179,16 @@ public int createImageType(final ImageType imageType) {
...
+            "Unable to fetch image type metadata for image type : " + name);
+      }
       // Check if there are more then one image types for a given name. If so throw exception
       if (imageTypes != null && imageTypes.size() > 1) {
+        log.error(
+            "Failed to get image type with ownerships by name. Can't have more that one image type"
+                + " record for a given type with name : " + name);
         throw new ImageMgmtDaoException(ErrorCode.NOT_FOUND, "Failed to get image type with "
             + "ownerships by name. Can't have more that one image type record for a given type "
             + "with name : " + name);
```
ypadron-in, line 194: In `getImageTypeWithOwnershipsById` we returned `ErrorCode.UNPROCESSABLE_ENTITY` in this case. Can we agree on one status and be consistent?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,17 +88,70 @@ protected void handleGet(
...
+      throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, FORBIDDEN_USER_ERR_MSG);
+    }
+    sendResponse(resp, HttpServletResponse.SC_OK, imageTypeDTO);
+  }
+
+  private void getAllImageTypeDTOs(final HttpServletResponse resp, final Session session)
+      throws IOException {
+    if (!isAzkabanAdmin(session.getUser())) {
+      throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, FORBIDDEN_USER_ERR_MSG);
+    }
```
ypadron-in, line 149: Why only Azkaban ADMINs can use this endpoint if non admins users can still make calls to `/imageTypes?&imageType=?`. I would expect that calls to `/imageTypes` by non admin users would return a list of imageTypes  that particular user has access to.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
```diff
@@ -75,9 +83,48 @@ public void testCreateImageType() throws Exception {
...
+    ImageTypeDTO expected = getImageTypeDTO();
+    when(this.imageTypeDao.getAllImageTypesWithOwnerships()).thenReturn(imageTypes);
+    List<ImageTypeDTO> imageTypeDTOs = this.imageTypeService.getAllImageTypesWithOwnerships();
+    assert imageTypeDTOs.size() == 1;
+    assertObjectsAreEqual(imageTypeDTOs.get(0), expected);
+  }
+
+  @Test
+  public void testFindImageTypeWithOwnershipsById() throws Exception {
+    String id = "1";
```
ypadron-in, line 98: Have you Installed the Intellij's Save Actions plugin mentioned in https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md#style-guides in your IDE?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,17 +88,70 @@ protected void handleGet(
...
+    ImageTypeDTO imageTypeDTO;
+    String imageTypeIdOrName = templateVariableToValue.get(IMAGE_TYPE_NAME_OR_ID);
+    if (!StringUtils.isNumeric(imageTypeIdOrName)) {
+
+      imageTypeDTO = this.imageTypeService.findImageTypeWithOwnershipsByName(imageTypeIdOrName);
+    } else {
+      imageTypeDTO = this.imageTypeService.findImageTypeWithOwnershipsById(imageTypeIdOrName);
+    }
+    if (!hasImageManagementPermission(imageTypeDTO.getName(), session.getUser(), Type.GET)) {
+      log.info(FORBIDDEN_USER_ERR_MSG);
```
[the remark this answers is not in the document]
amielkelinkedin: Not that I know of, we need the imageTypeName to get permissions - the whole code base is very image type name centric, so we need to get the ImageType name before proceeding which means for an ID, we need to ask the DB first.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: I think this is a question for @jakhani. As written it makes sense to me, but I can see it both ways.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: Thanks. I think we should just go with ErrorCode.UNPROCESSABLE_ENTITY. I will change it.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: For the code, I think yes, but when we mock things with our tests, I think not.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: I agree. Have changed them all for all calls in the class.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: Again, yes and no.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageTypeServiceImpl.java
amielkelinkedin: Thank you, I will change it.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,17 +88,70 @@ protected void handleGet(
...
-      if (IMAGE_TYPE_WITH_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
-        // TODO: Implementation will be provided in the future PR
+      if (IMAGE_TYPE_NAME_OR_ID_URI_TEMPLATE.match(req.getRequestURI(), templateVariableToValue)) {
+        getImageTypeDTOByIdOrImageTypeName(resp, session, templateVariableToValue);
       } else if (GET_IMAGE_TYPE_URI.equals(req.getRequestURI())) {
-        // TODO: Implementation will be provided in the future PR
+        ImageTypeDTO imageTypeDTO;
+        if (req.getQueryString().contains(IMAGE_TYPE)) {
+          final String imageTypeName = HttpRequestUtils
+              .getParam(req, ImageMgmtConstants.IMAGE_TYPE);
```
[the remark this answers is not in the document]
amielkelinkedin: I don't think I understand the concern here. Can you rephrase it differently?

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageTypeServiceImplTest.java
amielkelinkedin: It is installed and enabled, yes.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,17 +88,70 @@ protected void handleGet(
...
+          sendResponse(resp, HttpServletResponse.SC_OK, imageTypeDTO);
+        }
+        getAllImageTypeDTOs(resp, session);
+      } else {
+        log.info(PATH_NOT_SUPPORTED);
+        throw new ImageMgmtInvalidInputException(ErrorCode.NOT_FOUND, PATH_NOT_SUPPORTED);
+      }
+    } catch (final ImageMgmtInvalidPermissionException e) {
+      log.error("The user provided does not have permissions to access the resource");
+      resp.setStatus(HttpStatus.SC_FORBIDDEN);
```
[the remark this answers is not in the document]
amielkelinkedin: Done, thanks!

djaiswal83 (dismissed):
Thanks for addressing the comments. Looks good to me.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
ypadron-in: I see, thanks, in the future maybe we can consider adding "get image type ownership by id"

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
amielkelinkedin: Thank you for clarifying, fixed!

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
ypadron-in: Shouldn't we fix the tests then?

ypadron-in: For example this conditional block here is not needed as it will never be executed, why not simplify and optimize the code?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,15 +87,74 @@ protected void handleGet(
...
+      sendErrorResponse(resp, HttpServletResponse.SC_FORBIDDEN,
+          "User does not have permissions. Error Message: " + e.getMessage());
+    } catch (final ImageMgmtException e) {
+      if (e.getErrorCode() != null) {
+        log.error("An error has occurred");
+        resp.setStatus(e.getErrorCode().getCode());
+      } else {
+        resp.setStatus(HttpStatus.SC_BAD_REQUEST);
+      }
+      sendErrorResponse(resp, HttpServletResponse.SC_BAD_REQUEST,
```
ypadron-in, line 122: This call overwrites the status set in line 118 above right?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -74,15 +87,74 @@ protected void handleGet(
...
+          "Exception on GET call to /imageTypes. Reason: " + e.getMessage());
     } catch (final Exception e) {
       log.error("Content is likely not present " + e);
       resp.setStatus(HttpStatus.SC_NOT_FOUND);
+      sendErrorResponse(resp, HttpServletResponse.SC_NOT_FOUND,
+          "Exception on GET call to /imageTypes. Reason: " + e.getMessage());
```
ypadron-in, line 128: Maybe we should revisit the Error Handling logic in a separate PR. In my opinion Internal Server Error (500) is more suitable than Not Found (404) when unexpected errors happen.

jakhani (approved):

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: Thank you for making me double check this. I was wrong.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: You are correct!

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
amielkelinkedin: Fixed.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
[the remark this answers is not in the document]
amielkelinkedin: Perhaps in future work, definitely!

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
amielkelinkedin: Perhaps in future work, definitely!

ypadron-in (dismissed):
Thanks for implementing these API endpoints and addressing all the comments!

djaiswal83 (approved):
Looks good to me.

PR #2897, opened 2021-07-14
Users are needlessly spooked by ominous cryptic line in job logs:

li-ygerchikov:
`Failed with 6 inputs with exception e = java.lang.NoSuchMethodException: azkaban.jobtype.HadoopShell.<init>(java.lang.String, azkaban.utils.Props, azkaban.utils.Props, azkaban.utils.Props, org.apache.log4j.Logger)`

It is benign but misleading and is sometimes blamed if job fails elsewhere. I don't think we need that line in the logs if it is fully expected.
Also, minor cosmetic refactoring to avoid nested try{} and to eliminate duplicate catch(){}.


djaiswal83 (approved):
You did it so  much better :D

li-ygerchikov:
> You did it so much better :D

Why thank you! Glad you like it. Thanks for the quick review!

PR #2900, opened 2021-07-18
Better CSRF tokens

artem-smotrakov:
Unfortunately, I couldn't find a security contact to discuss it privately before opening a pull request. It might be good to add a [security policy](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) with such info.

I'd suggest to update `CSRFTokenUtility`:

- Use `SecureRandom` to generate a key for MAC. That strengthens the key, and makes it more difficult to forte a CSRF token.
- Use `MessageDigest.isEqual()` when validating tokens. `String.equals()` uses a non-constant-time algorithm to compare inputs. As a result, it might allow an attacker to run a timing attack. Instead, it is better to validate tokens with `MessageDigest.isEqual()` that uses a constant-time algorithm whose time doesn't depend on the content of the inputs.
- Use `StandardCharsets.UTF_8` when converting strings to byte arrays. That allows to get rid of a try-catch block for `UnsupportedEncodingException`.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/CSRFTokenUtility.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/CSRFTokenUtility.java
@@ -19,13 +19,15 @@ import azkaban.server.session.Session;
 import azkaban.utils.StringUtils;
 import java.io.UnsupportedEncodingException;
 import java.security.InvalidKeyException;
+import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


aditya1105:
@artem-smotrakov Good suggestions.

artem-smotrakov:
> @artem-smotrakov Good suggestions.

Thanks for the feedback @aditya1105 It would be great if someone from the project maintainers reviewed the patch.

PR #2903, opened 2021-07-21
Add param for disabling pod cleanup

sshardool:
Introduces a flow parameter `disable.pod.cleanup`, if it's set
the flow-pod will not be auto-cleaned after flow execution ends.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -115,6 +115,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
   public static final String APP_LABEL_NAME = "app";
   public static final String EXECUTION_ID_LABEL_NAME = "execution-id";
   public static final String EXECUTION_ID_LABEL_PREFIX = "execid-";
+  public static final String DISABLE_CLEANUP_LABEL_NAME = "cleanup-disabled";
 
   private final String namespace;
   private final ApiClient client;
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -798,6 +799,13 @@ private ImmutableMap getLabelsForPod(final int executionId) {
       mapBuilder.put("service", String.join("-", SERVICE_SELECTOR_PREFIX,
           clusterQualifiedExecId(this.clusterName, executionId)));
     }
+
+    // Set the label for disabling pod-cleanup.
+    if (flowParam != null && !flowParam.isEmpty() && flowParam
+        .containsKey(FlowParameters.FLOW_PARAM_DISABLE_POD_CLEANUP)) {
```
jakhani, line 805: Shouldn't you add value check as well? With that, shall it be admin only flow param?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
```diff
@@ -110,7 +122,21 @@ public String getClusterName() {
         return Optional.empty();
       }
       executionId = executionLabel.substring(EXECUTION_ID_LABEL_PREFIX.length());
-      return Optional.of(new FlowPodMetadata(executionId, clusterName));
+
+      String cleanupDisabledLabel =
```
jakhani, line 126: It will be good to clean up this pod at sometime. Otherwise it will keep hanging there. Whether pod is 10 days old or some check. Please share your thoughts around this.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
sshardool: You're right, it should be an admin only param, adding that and a boolean value check as well.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusMetadata.java
sshardool: Yes, it will be cleaned up as part of the 'stale-flow' check. Will add that to the ticket if it's not explicitly there.

jakhani (approved):
LGTM!

jakhani (approved):

PR #2904, opened 2021-07-21
Use Webserver host name instead of Executor host name for Grid Attribution.

djaiswal83:
This works better when containerization is enabled.

### DIFF
--- a/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
+++ b/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
@@ -523,7 +523,11 @@ public abstract class AbstractHadoopSecurityManager extends HadoopSecurityManage
 
   /*
    * Create a suffix for Kerberos principal, the format is,
-   * az_<host name>_<execution id><DOMAIN_NAME>
+   * az_<webserver_host name>_<execution id><DOMAIN_NAME>
+   * The UGI with executor host name is less useful in containerized world
+   * where each flow runs in its own container with unique host name.
+   * For meaningful analytics using the UGI data, it is better to use webserver
+   * hostname.
    */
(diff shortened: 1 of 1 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -528,7 +528,7 @@ private void assignPermissions(final String user, final File tokenFile, final St
   protected String kerberosSuffix(final Props props) {
     // AZKABAN_SERVER_HOST_NAME is not set in Props here, get it from another instance of Props.
     final String host = ServiceProvider.SERVICE_PROVIDER.getInstance(Props.class)
-        .getString(AZKABAN_SERVER_HOST_NAME, "unknown");
+        .getString(AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME, "unknown");
```
jakhani, line 531: This will change things for both on prem and containerization. Is that ok?

djaiswal83: Yes. This way, the grid attribution pipeline does not have to handle things differently keeping it simple

sshardool: The comment above this line is still referring to `AZKABAN_SERVER_HOST_NAME`. Please fix that and update to include a note about why we are choosing the webserver name as part of the ugi.

djaiswal83: Done

hqin2016 (approved):
Thanks for fixing the grid attribution issue.

jakhani (approved):
Thank you for taking care of this issue!

sshardool (approved):

PR #2905, opened 2021-07-23
Avoid NPE in FlowContainer.shutdown() if flowRunner is NULL. 

djaiswal83:
Avoid NPE in FlowContainer.shutdown() if flowRunner is NULL. It happens when for some reason, the flow never starts.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -724,9 +724,8 @@ public class FlowContainer implements IMBeanRegistrable, EventListener<Event> {
   @VisibleForTesting
   void shutdown() {
     logger.info("Shutting down the container");
-    int execId = -1;
     if (this.flowRunner != null) {
-      execId = this.flowRunner.getExecutionId();
+      final int execId = this.flowRunner.getExecutionId();
       while (!this.flowFuture.isDone()) {
         // This should not happen immediately as submitFlowRunner is a blocking call.
         try {
@@ -764,7 +763,7 @@ public class FlowContainer implements IMBeanRegistrable, EventListener<Event> {
     logger.info("Sayonara!");
     if (this.flowRunner != null) {
       // If the flowRunner is not created, the execId would be invalid.
-      uploadLogFile(execId);
+      uploadLogFile(this.flowRunner.getExecutionId());
     }
     System.exit(0);
   }
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -723,16 +723,20 @@ public class FlowContainer implements IMBeanRegistrable, EventListener<Event> {
    */
   @VisibleForTesting
   void shutdown() {
-    logger.info("Shutting down the pod");
-    final int execId = this.flowRunner.getExecutionId();
-    while (!this.flowFuture.isDone()) {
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -723,16 +723,20 @@ void closeMBeans() {
...
-    final int execId = this.flowRunner.getExecutionId();
-    while (!this.flowFuture.isDone()) {
-      // This should not happen immediately as submitFlowRunner is a blocking call.
-      try {
-        Thread.sleep(100);
-      } catch (final InterruptedException e) {
-        logger.error(String.format("The sleep while waiting for execution : %d to finish was interrupted",
-            execId));
+    logger.info("Shutting down the container");
+    int execId = -1;
```
aditya1105, line 727: The value -1 is never utilized in the code. All the places where execId is used, are the places where there is a check for flowRunner != null. Hence this initialization of execId can be done inside the if condition. 
For the uploadLogFile, it can be uploadLogFile(this.flowRunner.getExecutionId())

svnarumugam:
Thanks Deepak for this change. while this fix will mitigate the current problem, we should track and find the cause of this problem.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -723,16 +723,19 @@ void closeMBeans() {
...
-      } catch (final InterruptedException e) {
-        logger.error(String.format("The sleep while waiting for execution : %d to finish was interrupted",
-            execId));
+    logger.info("Shutting down the container");
+    if (this.flowRunner != null) {
+      final int execId = this.flowRunner.getExecutionId();
+      while (!this.flowFuture.isDone()) {
+        // This should not happen immediately as submitFlowRunner is a blocking call.
+        try {
+          Thread.sleep(100);
```
amielkelinkedin, line 732: Perhaps in the future we should try a wait() implementation that is more precise.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -758,7 +761,10 @@ void shutdown() {
       logger.error("Error shutting down JettyServer while winding down the FlowContainer", e);
     }
     logger.info("Sayonara!");
-    uploadLogFile(execId);
+    if (this.flowRunner != null) {
```
amielkelinkedin, line 764: I don't like this. If this flowRunner is null wouldn't the .getExecutionId() call above give an NPE?

amielkelinkedin, line 764: My understanding (which might be wrong), is that this is a change that is needed to quickly mitigate an issue. Additionally, my understanding of the Azkaban code is not the best. However, it seems like we should strive for a state where the flowRunner cannot be null, especially because we control the flowRunner object internally, right? Would it be wise to work toward such a state in the future?

amielkelinkedin (approved):

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: It is possible. What is happening is for some reason, the FlowRunner object is not created and shutdown is called (being investigated separately) which is causing this problem.
If it is identified that it is possible to not have FlowRunner object in some cases, we may have to keep this logic.

aditya1105 (approved):

PR #2907, opened 2021-07-28
set job status to terminal state when pod goes down

hqin2016:
In containerized executions, when a pod crashes, the jobs need to be in a terminal state. This PR aims to fix a bug that when a flow was finalized from RUNNING to FAILED/EXECUTION_STOPPED due to pod failure, the job nodes showed KILLING status.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -220,33 +220,45 @@ public class ExecutionControllerUtils {
    */
   public static void failEverything(final ExecutableFlow exFlow) {
     final long time = System.currentTimeMillis();
-    for (final ExecutableNode node : exFlow.getExecutableNodes()) {
-      switch (node.getStatus()) {
-        case SUCCEEDED:
-        case FAILED:
-        case KILLED:
-        case SKIPPED:
-        case DISABLED:
-          continue;
-          // case UNKNOWN:
-        case READY:
-          // if flow status is EXECUTION_STOPPED due to e.g. pod failure, set sub node to KILLED
-          if (exFlow.getStatus()==Status.EXECUTION_STOPPED) {
-            node.setStatus(Status.KILLED);
-          } else {
-            node.setStatus(Status.KILLING);
-          }
-          break;
-        default:
-          node.setStatus(Status.FAILED);
-          break;
+    final LinkedList<ExecutableNode> queue = new LinkedList<>();
+    queue.add(exFlow);
+    // Traverse the DAG and fail every node that's not in a terminal state
+    while (!queue.isEmpty()) {
+      final ExecutableNode node = queue.poll();
+      if (node instanceof ExecutableFlowBase) {
+        final ExecutableFlowBase base = (ExecutableFlowBase) node;
+        for (final ExecutableNode subNode : base.getExecutableNodes()) {
+          queue.add(subNode);
+        }
       }
+      if (node != exFlow) {
+        switch (node.getStatus()) {
+          case SUCCEEDED:
+          case FAILED:
+          case KILLED:
+          case SKIPPED:
+          case DISABLED:
+            continue;
+            // case UNKNOWN:
+          case READY:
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -230,7 +230,12 @@ public static void failEverything(final ExecutableFlow exFlow) {
           continue;
           // case UNKNOWN:
         case READY:
-          node.setStatus(Status.KILLING);
+          // if flow status is EXECUTION_STOPPED due to e.g. pod failure, set sub node to KILLED
+          if (exFlow.getStatus()==Status.EXECUTION_STOPPED) {
+            node.setStatus(Status.KILLED);
```
jakhani, line 235: Just to confirm, this will take care of all the nodes in graph. right?

sshardool: +1

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -230,7 +230,12 @@ public static void failEverything(final ExecutableFlow exFlow) {
           continue;
           // case UNKNOWN:
         case READY:
-          node.setStatus(Status.KILLING);
+          // if flow status is EXECUTION_STOPPED due to e.g. pod failure, set sub node to KILLED
+          if (exFlow.getStatus()==Status.EXECUTION_STOPPED) {
```
sshardool, line 234: Wondering if there were any issues around around jobs showing up in 'running' (something not at 'Ready' state) and will those also be handled by this change ?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
hqin2016: No, the current code is only taking care of first level of nodes, ignoring the sub-nodes within an embedded flow. I'll refactor the code to traverse the DAG.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
hqin2016: It's already handled by the default option, setting these nodes (e.g. RUNNING) to FAILED.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -220,28 +220,45 @@ public static void alertUserOnFirstError(final ExecutableFlow flow,
...
-        case DISABLED:
-          continue;
-          // case UNKNOWN:
-        case READY:
-          node.setStatus(Status.KILLING);
-          break;
-        default:
-          node.setStatus(Status.FAILED);
-          break;
+    final LinkedList<ExecutableNode> queue = new LinkedList<>();
```
jakhani, line 223: Where are you using this queue? In this while loop, you have prepared it. Then traversing it and marking status is missing.

djaiswal83: This is incorrect way of declaring it. Always use the base class, in this case, List

djaiswal83: @jakhani the entire logic to mark the status is inside the while loop of the BFS.

djaiswal83 (changes requested):
Thanks for addressing this. I have one comment.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
hqin2016: @djaiswal83 changed to Queue.
Discussed with @jakhani offline, the logic is correct.

aditya1105 (approved):

djaiswal83 (approved):

jakhani (approved):
Thank you so much for taking care of it thoroughly!

PR #2909, opened 2021-08-03
Version selection based on max rampup percentage when flow is not specified.

aditya1105:
After deterministic ramp up, a version is selected based on the ramp-up percentages of job types and the hash value derived from the flow name.
However when the flow is not specified, currently the first element in the ramp-up plan will be selected. Flow is not specified when loading the status page. This creates an abnormal state on the status UI, when a ramp-up of 100 is specified, but is not the first element.

This change, makes sure that in such a scenario, the version with the maximum ramp-up is selected and hence the status UI will display that version.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
@@ -296,11 +296,16 @@ public class ImageRampupManagerImpl implements ImageRampupManager {
       if (null == flow) {
         log.info("Flow object is null, so continue");
         final ImageRampup firstImageRampup = imageRampupList.get(0);
+
+        // Find the imageVersion in the Rampup list with maximum rampup percentage.
+        final ImageRampup maxImageRampup = imageRampupList.stream()
+            .max(Comparator.comparing(ImageRampup::getRampupPercentage))
+            .orElseGet(() -> firstImageRampup);
         imageTypeRampupVersionMap.put(imageTypeName,
-            this.fetchImageVersion(imageTypeName, firstImageRampup.getImageVersion())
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
```diff
@@ -296,11 +296,16 @@ public VersionInfo getVersionInfo(final String imageType, final String imageVers
...
         log.info("Flow object is null, so continue");
         final ImageRampup firstImageRampup = imageRampupList.get(0);
+
+        // Find the imageVersion in the Rampup list with maximum rampup percentage.
+        final ImageRampup maxImageRampup = imageRampupList.stream()
+            .max(Comparator.comparing(ImageRampup::getRampupPercentage))
+            .orElseGet(() -> firstImageRampup);
         imageTypeRampupVersionMap.put(imageTypeName,
-            this.fetchImageVersion(imageTypeName, firstImageRampup.getImageVersion())
+            this.fetchImageVersion(imageTypeName, maxImageRampup.getImageVersion())
```
djaiswal83, line 305: I thought we decided on using the latest active version and not the one which is max ramped up.

djaiswal83 (changes requested):
Can you please address my concern? Thanks for adding a unit test.

azkaban-common/src/main/java/azkaban/imagemgmt/rampup/ImageRampupManagerImpl.java
aditya1105: In the case when there is no ramp-up strategy, then the latest active version will be displayed as that will be the one to be utilized. However, when the ramp plan is specified, then the one with the max percentage of it will be displayed. This will be consistent when a version is 100% in the ramp up plan.

PR #2913, opened 2021-08-10
Merge containers from pod-spec template and dynamically generated pod-spec

aditya1105:
The current merge method for init containers from the pod-spec template and dynamically generated pod-spec is a simple append to the list, i.e. take all the init containers from the dynamically generated pod-spec and take only those init containers from the pod-spec template which are not part of generated pod-spec.

However, the init containers from generated pod-spec are relatively simple with only a couple of env variables, one fixed volume which gets mounted to the init container as well as flow container. There is a need for more complex init containers where the complex part can come from the pod-spec template.

This change updates the merge method for init containers from the pod-spec template and dynamically generated pod-spec, such that:
1) Add all the init containers which are only part of podTemplate. 
2) Add all the init containers which are only part of pod-spec. 
3) Add the init containers which are part of both pod-spec and pod-template by merging them such that: 
    a) Skeleton of templateInitContainer is utilized. 
    b) Environment variables from podSpecInitContainer are added to the corresponding  templateInitContainer. 
    c) ImagePullPolicy, Image, and VolumeMounts are overridden from podSpecInitContainer to templateInitContainer.

This change is also leveraged to merge the flow-container from the pod template and the dynamically generated pod-spec. Hence removed the unnecessary methods.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
@@ -81,6 +81,22 @@ public class AzKubernetesV1PodTemplate {
   }
 
   /**
+   * @return The Flow Container which must be the first container among all app-containers
+   */
+  public V1Container getFlowContainer() {
+    List<V1Container> containers = this.podFromTemplate.getSpec().getContainers();
+    return containers.isEmpty() ? null : containers.get(FLOW_CONTAINER_INDEX);
+  }
+
+  /**
+   * @return The list of all app-containers.
+   */
+  public List<V1Container> getAllContainers() {
+    return this.podFromTemplate.getSpec().getContainers();
+  }
+
+
+  /**
    * @return the {@link V1Pod} POD generated from the template.
    */
   public V1Pod getPodFromTemplate() {
--- a/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
@@ -84,7 +84,7 @@ public class AzKubernetesV1PodTemplate {
    * @return The Flow Container which must be the first container among all app-containers
    */
   public V1Container getFlowContainer() {
-    List<V1Container> containers = this.podFromTemplate.getSpec().getContainers();
+    final List<V1Container> containers = this.podFromTemplate.getSpec().getContainers();
     return containers.isEmpty() ? null : containers.get(FLOW_CONTAINER_INDEX);
   }
 
--- a/azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
@@ -16,11 +16,14 @@
 package azkaban.container.models;
 
 import io.kubernetes.client.openapi.models.V1Container;
+import io.kubernetes.client.openapi.models.V1EnvVar;
 import io.kubernetes.client.openapi.models.V1PodSpec;
-import io.kubernetes.client.openapi.models.V1Probe;
 import io.kubernetes.client.openapi.models.V1Volume;
 import io.kubernetes.client.openapi.models.V1VolumeMount;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * This class consists of various methods to merge pod-spec with the pod template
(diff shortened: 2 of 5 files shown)


azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
```diff
@@ -42,95 +45,180 @@ private PodTemplateMergeUtils() {
...
   /**
-   * Add those container volumes which are not already available in the podSpec
+   * This method combines the templateContainer and podSpecContainer such that: 1) Skeleton of
+   * templateContainer is utilized. 2) Environment variables from podSpecContainer are added to
+   * corresponding templateContainer. 3) ImagePullPolicy, Image, and VolumeMounts are overridden
+   * from podSpecContainer to templateContainer.
    *
-   * @param podSpec     Already created podSpec using the {@link AzKubernetesV1SpecBuilder}
-   * @param podTemplate Instance of the class {@link AzKubernetesV1PodTemplate} to extract items
+   * @param templateContainer Container form the pod-spec template
```
djaiswal83, line 148: from, applies to both lines

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
```diff
@@ -42,95 +45,180 @@ private PodTemplateMergeUtils() {
...
-    List<V1Container> podSpecInitContainers = podSpec.getInitContainers();
-    if (null != podSpecInitContainers) {
-      List<V1Container> templateInitContainers = podTemplate.getInitContainers(
-          tempInitContainer -> podSpecInitContainers.stream().map(V1Container::getName)
-              .noneMatch(name -> name.equals(tempInitContainer.getName())));
-      for (V1Container containerItem : templateInitContainers) {
-        podSpec.addInitContainersItem(containerItem);
-      }
+  private static List<V1VolumeMount> getMergedContainerVolumeMounts(
+      List<V1VolumeMount> templateContainerVolumeMounts,
```
djaiswal83, line 203: Please use final for input params not meant to be out params. Applies to all such methods.

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
```diff
@@ -42,95 +45,180 @@ private PodTemplateMergeUtils() {
...
+      List<V1VolumeMount> podSpecContainerVolumeMounts) {
+    final List<V1VolumeMount> allContainerVolumeMounts = new ArrayList<>();
+    if (null != templateContainerVolumeMounts && null != podSpecContainerVolumeMounts) {
+      List<V1VolumeMount> templateOnlyInitContainerVolumeMounts = templateContainerVolumeMounts
```
djaiswal83, line 207: This is very inefficient way to merge. Can you create a set from templateContainerVolumeMounts and iterate over podSpecContainerVolumentMounts to create union which would have anything common overridden from podSpec.
Convert the set into a list. Unless ofc, I am not able to understand this logic.

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
```diff
@@ -42,95 +45,180 @@ private PodTemplateMergeUtils() {
...
    */
-  private static void mergeLivelinessProbe(V1PodSpec podSpec,
+  private static void mergeInitContainers(V1PodSpec podSpec,
       AzKubernetesV1PodTemplate podTemplate) {
-    V1Probe containerLivelinessProbe = podTemplate.getContainerLivelinessProbe();
-    if (null != containerLivelinessProbe) {
-      podSpec.getContainers().get(AzKubernetesV1PodTemplate.FLOW_CONTAINER_INDEX)
-          .setLivenessProbe(containerLivelinessProbe);
+    List<V1Container> podSpecInitContainers = podSpec.getInitContainers();
+    if (null != podSpecInitContainers) {
```
djaiswal83, line 74: if (null == podSpecInitContainers) return;

Keep the rest of code with less indentation.

djaiswal83 (changes requested):
Thanks for making this change. The code is lot more simplified, generic and extensible.

I have few comments mostly nitpicks.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -80,6 +80,22 @@ public static synchronized AzKubernetesV1PodTemplate getInstance(String template
     return instance;
   }
 
+  /**
+   * @return The Flow Container which must be the first container among all app-containers
+   */
+  public V1Container getFlowContainer() {
+    List<V1Container> containers = this.podFromTemplate.getSpec().getContainers();
```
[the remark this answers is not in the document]
aditya1105: done

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
aditya1105: done

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
aditya1105: done

azkaban-common/src/main/java/azkaban/container/models/PodTemplateMergeUtils.java
aditya1105: This method is doing exactly the same. It is finding all the volume mounts in templateContainerVolumeMounts that are not part of podSpecContainerVolumentMounts. And then adding them all. 
It is however not necessary to create a list, hence updated it to utilize foreach loop.

djaiswal83 (approved):
Thanks for addressing comments. Looks good.

PR #2915, opened 2021-08-11
Setup environment variables for RPMs based on flow-params

aditya1105:
This change allows setting up a list of RPMs via flow params as ENV variables to the flow container. 

The motivation is that the startup scripts for the flow containers can utilize this ENV variable to install the RPMs for various testing purposes.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -669,6 +669,7 @@ public class Constants {
     public static final String ENV_FLOW_EXECUTION_ID = "FLOW_EXECUTION_ID";
     public static final String ENV_JAVA_ENABLE_DEBUG = "JAVA_ENABLE_DEBUG";
     public static final String ENV_ENABLE_DEV_POD = "ENABLE_DEV_POD";
+    public static final String ENV_RPM_PREFIX = "RPM_";
   }
 
   public static class ImageMgmtConstants {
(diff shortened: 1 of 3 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -606,6 +607,30 @@ private void setupDevPod(final Map<String, String> envVariables,
...
+   * This method is used to setup environment variable for RPMs. Based on this environment
+   * variable startup scripts for the flow container can download/install them.
+   *
+   * @param envVariables
+   * @param flowParam
+   */
+  @VisibleForTesting
+  static void setupRPMs(final Map<String, String> envVariables,
+      final Map<String, String> flowParam) {
+    if (flowParam == null || flowParam.isEmpty()) {
```
svnarumugam, line 620: Nit: MapUtils.isEmpty(flowParam) will be concie.

aditya1105: Thanks. I have updated it to use MapUtils.isEmpty()

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -693,6 +694,10 @@
     // Constant to enable pod for developer testing
     public static final String FLOW_PARAM_ENABLE_DEV_POD = "enable.dev.pod";
 
+    // Params with this prefix will be set as environment variables to the flow container and can
+    // be utilized by any script for download/installation.
+    public static final String FLOW_PARAM_RPM_PREFIX = "flow.container.rpm.";
```
jakhani, line 699: Can't we utilize pod.env.var. flow param? We are using it for Azkaban distribution already.

aditya1105:
Based on Janki's comment, it seems like pod.env.var can be utilized and hence no code changes are required. Hence closing this PR.

PR #2918, opened 2021-08-16
Update the log message to info from error when a project is not found in cache during server startup.

djaiswal83:


sshardool (approved):

azkaban-common/src/main/java/azkaban/project/InMemoryProjectCache.java
```diff
@@ -115,7 +115,7 @@ public void putProject(final Project project) {
   public Optional<Project> getProjectById(final Integer key) throws ProjectManagerException {
     Project project = this.projectsById.get(key);
     if (project == null) {
-      logger.error("Project not found in cache, fetching from DB");
+      logger.info("Project with ID " + key + " not found in cache, fetching from DB");
```
jakhani, line 118: Can key be null?

PR #2921, opened 2021-08-20
Add Hadoop env vars to list of non-overridable properties

hungj:
Similar to https://github.com/azkaban/azkaban/pull/2860. Prevent Hadoop env vars from being passed from upstream jobs.


djaiswal83 (approved):
Looks good to me.

PR #2922, opened 2021-08-24
Using a new instance of V1PodSpec from the template to avoid corruption from previous merging of pod-specs

aditya1105:
After the PR https://github.com/azkaban/azkaban/pull/2913 
azkaban.container.models.PodTemplateMergeUtils#mergePodSpec 
makes changes in the Pod generated from the Template which is a singleton and has only one instance of Pod stored internally. Hence multiple executions lead to corruption of the same instance of POD in the singleton class.

This can be fixed by making sure that a new instance of POD is generated from the template every time during the course of pod spec creation. Since this instance will be local to the method, multiple threads won't affect it.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
+++ b/azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
@@ -16,55 +16,41 @@
 
 package azkaban.container.models;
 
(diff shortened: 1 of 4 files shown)


jakhani (approved):

djaiswal83 (approved):
Thanks for fixing this quickly. The code looks good to me.

PR #2924, opened 2021-08-25
🐛  Delete button bug fix

amielkelinkedin:
Users have requested that Azkaban be changed so that Runtime Properties are easier to delete.

Before: The delete button is only available on mouse over, and is a child element of the Value cell.

Screen Shot 2021-08-25 at 1 08 18 AM

After: The delete button is available on all rows and is an element of the row itself.

Screen Shot 2021-08-25 at 1 06 23 AM

Tested locally and on in-house servers.


juhoautio:
Thank you for this! I have also stumbled on this but am not so skilled with the front end so never got to doing it myself.

Could the layout be improved a bit though? Maybe create a column of its own for the buttons so that they would be displayed with the same varying background color (white/grey) and be more clearly bound to their data row?

amielkelinkedin:
Closing in preference to #2928

PR #2928, opened 2021-08-27
Always show delete button of runtime properties

juhoautio:
Proposing an alternative fix for the issue of https://github.com/azkaban/azkaban/pull/2924:

> Users have requested that Azkaban be changed so that Runtime Properties are easier to delete.

> Before: The delete button is only available on mouse over, and is a child element of the Value cell.

After: The delete button is always visible on all rows and has its own "column" in the table:

Näyttökuva 2021-8-27 kello 22 39 42


amielkelinkedin (approved):
Tested local and on in house servers. Thank you Juho!

juhoautio:
Nice. I don't have permission to merge, so need someone with that permission to do it.

amicks:
Thanks @amielkelinkedin  @juhoautio !

PR #2929, opened 2021-08-28
Document Runtime Properties

juhoautio:
Updating the documentation to cover the changes & new features that were introduced in PR https://github.com/azkaban/azkaban/pull/2597.

Rendered doc from local:
Näyttökuva 2021-8-28 kello 18 21 02

(For comparison, current public doc site is here:
https://azkaban.readthedocs.io/en/latest/useAzkaban.html?highlight=override#flow-parameters)

----

NOTE: the new example screenshot assumes that https://github.com/azkaban/azkaban/pull/2928 (or an equivalent PR) is merged to make the delete button visible without mouseover.


docs/useAzkaban.rst
```diff
@@ -245,14 +245,23 @@ executing, several options can be set.
...
+.. image:: figures/executeflowruntimeproperties.png
+
+The scope of each runtime property is defined by the Node selection. Rows with `ROOT` Node apply to all jobs in the flow. Properties can be scoped per sub-flow as well as individual jobs. Most specific node path takes precedence.
+
+Overrides are applied depending on the Azkaban server configuration property `executor.props.resolve.overrideExisting.enabled`:
+
++ **If overriding of existing properties is not enabled (default):**
+   + Runtime properties override the global properties for a job, but not the properties of the job itself.
++ **If overriding of existing properties is enabled:**
+   + Runtime properties override also existing job props
```
ypadron-in, line 264: I understand the documentation already referred to "global properties for a job" but I'm not sure what that means  because there are properties defined a flow/rootnode level which apply to all jobs in a DAG but there are also "global" properties at Application(Azkaban) level and plugin level that could impact all jobs as well.

Here is an alternative explanation. Does it sound clearer?
`executor.props.resolve.overrideExisting.enabled=false` (default)
Predefined job properties (props in .job, .properties (Flow 1.0)  or .flow (Flow 2.0) files) have higher precedence than runtime properties overrides specified in any parent node.
`executor.props.resolve.overrideExisting.enabled=true`
Runtime properties overrides in parent nodes have higher precedence than predefined job properties.

docs/useAzkaban.rst
```diff
@@ -245,14 +245,23 @@ executing, several options can be set.
...
 
-Allows users to override flow parameters. The flow parameters override
-the global properties for a job, but not the properties of the job
-itself.
+Runtime Properties allow users to override properties temporarily.
 
-.. image:: figures/executeflowparameters.png
+Note: This feature was previously known as Flow Parameters.
+
+.. image:: figures/executeflowruntimeproperties.png
```
ypadron-in, line 255: Does the `example-flow` node in the picture represent a job or a subflow? Is it possible to use a more representative name to avoid confusions between the ROOT node and rest of nodes?

ypadron-in (commented):
Thanks a lot for adding documentation for this new feature! I have added a few comments below for your consideration.

docs/useAzkaban.rst
```diff
@@ -245,14 +245,23 @@ executing, several options can be set.
...
+
+The scope of each runtime property is defined by the Node selection. Rows with `ROOT` Node apply to all jobs in the flow. Properties can be scoped per sub-flow as well as individual jobs. Most specific node path takes precedence.
+
+Overrides are applied depending on the Azkaban server configuration property `executor.props.resolve.overrideExisting.enabled`:
```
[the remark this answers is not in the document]
juhoautio: I can see the possibility for confusion. How about this: `executor.runtimeProps.override.eager`? The default would be `false`. Or did you have some name in mind that you'd like to be used?

docs/useAzkaban.rst
juhoautio: ## enabled=false

> Predefined job properties (props in .job, .properties (Flow 1.0) or .flow (Flow 2.0) files) have higher precedence than runtime properties overrides specified in any parent node.

**I'm afraid that's incorrect**. It's only the job level properties that have higher precedence than runtime properties. An example of that here:

https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java#L62-L67

The flow under test is loaded from https://github.com/azkaban/azkaban/tree/master/test/execution-test-data/execpropstest and especially there is this `.properties` file with the following property:

https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/test/execution-test-data/execpropstest/moo.properties#L3

Now, let's jump into that `assertProperties()`.

For example this ROOT level runtime property is set:

https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java#L120

It's asserted that the runtime property was used (instead of using the value defined in `moo.properties`):

https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java#L159

The order of precedence is also written out as a comment in the same test method:
https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerPropertyResolutionTest.java#L152-L153

## enabled=true

> Runtime properties overrides in parent nodes have higher precedence than predefined job properties.

I feel this somehow lacking because it sounds like it only concerns the job properties, although it's also .properties (or any predefined properties) that are overridden in this case.

How about this:

> Runtime properties (also those that are set on a parent node) have higher precedence than predefined properties.

?

docs/useAzkaban.rst
juhoautio: Sure thing, I was also being too lazy here.

juhoautio: How about this? (also committed & pushed)

image

This is missing the DELETE buttons though. I can rebase & generate this again if otherwise ok.

tools/deploy_example_flow.py
```diff
@@ -0,0 +1,83 @@
+"""
+Uploads the "execpropstest" folder as a project to Azkaban at localhost.
+Then executes the flow "job3" with some runtime property overrides.
```
juhoautio, line 3: Added this script to make it easier to do manual testing locally. I'm not aware of anything similar existing previously.

docs/useAzkaban.rst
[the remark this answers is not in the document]
ypadron-in: I think `executor.runtimeProps.override.eager` is perfect (with `azkaban.` prefix even more :) ) because it hints at changing the behavior of _overrides_ instead of disabling them and is ambiguous enough that users will have to read the documentation instead of making assumptions. Thanks!

docs/useAzkaban.rst
ypadron-in: > Runtime properties (also those that are set on a parent node) have higher precedence than predefined properties.

Sounds good!

Regarding:
> Predefined job properties (props in .job, .properties (Flow 1.0) or .flow (Flow 2.0) files) have higher precedence than runtime properties overrides specified in any parent node.

I think in Flow 2.0 properties defined in _.properties_ files are interpreted as parent node props instead of job specific props (those defined in _.job_ files). That was my understanding as well for Flow 1.0 DAG definitions but I may be wrong. 
Based on that reasoning, `props5=moo5` would not count as a "predefined job property" so setting it to the runtime specified value it's ok and doesn't invalidate the sentence right?

But let's scratch that, what do you think of:
```
Runtime properties don't override predefined properties from a more specific node path. In particular
runtime properties at flow level don't override predefined properties at job level.
```
instead?

docs/useAzkaban.rst
ypadron-in: Great, thanks! 
I think it's ok to ignore the style of the DELETE button, that doesn't change the behavior of Runtime props.

docs/useAzkaban.rst
juhoautio: That's perfect. Using that.

juhoautio:
All review comments so far should be now resolved.

I also updated the PR description with a new screenshot.

PR #2930, opened 2021-08-28
Document python venv for docs building

juhoautio:
IMHO it's better to offer readymade commands for at least some way of creating the venv. Otherwise the docs building can be painful to setup for developers that are not so familiar with Python.


juhoautio:
Could this be reviewed, please? I'm also already referring to this in https://github.com/azkaban/azkaban/pull/2929 regarding the python env.

PR #2939, opened 2021-09-02
Improve logging of email send error

juhoautio:
If email sending fails, log the message subject & recipient to Azkaban server log


juhoautio:
Thanks! Would someone like to merge this? Someone that has permission :)

PR #2940, opened 2021-09-03
Overwrite any common property between sys props and cluster props with cluster property.

djaiswal83:
Overwrite any common property between sys props and cluster props with cluster property.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -432,6 +432,10 @@ public class JobTypeManager {
         if (!jobProps.containsKey(key)) {
           jobProps.put(key, clusterSpecificProps.get(key));
         }
+        // If there is a cluster specific property found in pluginLoadProps, then overwrite it in pluginLoadProps
+        if (pluginLoadProps.containsKey(key)) {
+          pluginLoadProps.put(key, clusterSpecificProps.get(key));
+        }
       }
       Props nonOverriddableClusterProps = getClusterSpecificNonOverridableJobProps(clusterSpecificProps);
       // CAUTION: ADD ROUTER-SPECIFIC PROPERTIES THAT ARE CRITICAL FOR JOB EXECUTION AS THE LAST


hqin2016 (approved):
Thanks for fixing this.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -432,6 +432,10 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
         if (!jobProps.containsKey(key)) {
           jobProps.put(key, clusterSpecificProps.get(key));
         }
+        // If there is a cluster specific property found in pluginLoadProps, then overwrite it in pluginLoadProps
+        if (pluginLoadProps.containsKey(key)) {
+          pluginLoadProps.put(key, clusterSpecificProps.get(key));
```
haibchen, line 437: Overall idea LGTM.

One thing to note is that pluginLoadProps is shared between instances of built-in job types.
See https://github.com/azkaban/azkaban/blob/8df583c350f050c097780cdc8be69c973cea861f/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java#L507.

My understanding is that jobs of types such as `commonad` `javaprocess` would go through the `else` block, share the object returned by `pluginSet.getCommonPluginLoadProps()`.

With this new behavior, if multiple jobs of built-in types are running, they could be modifying  the same object at the same time.

I think we can just create a copy of `pluginSet.getCommonPluginLoadProps()` for each job, that will do. That is 

```
static Props getPluginLoadProps(JobTypePluginSet pluginSet, String jobType) {
    Props pluginLoadProps = pluginSet.getPluginLoaderProps(jobType);
    if (pluginLoadProps != null) {
      pluginLoadProps = PropsUtils.resolveProps(pluginLoadProps);
    } else {
      // pluginSet.getCommonPluginLoadProps() will return null if there is no plugins directory.
      // hence assigning default Props() if that's the case
      pluginLoadProps = pluginSet.getCommonPluginLoadProps();
      if (pluginLoadProps == null) {
        pluginLoadProps = new Props();
      } else {
        
        // create a copy for individual jobs so that they can modify independently
        pluginLoadProps = new Props(pluginLoadProps);
      }
    }
    return pluginLoadProps;
  }
```

PR #2941, opened 2021-09-06
Set flow effectiveUsers and resource utilization to FLOW_FINISHED_EVENT

aditya1105:
This change adds the comma-separated list of effective users across all the jobs to the FLOW_FINISHED_EVENT. This enables us to determine at one place all the effective users for a flow.

This change also exposes the method to set CPU and MEMORY Utilization as part of FLOW_FINISHED_EVENT. However, this method is currently called only by the FlowContainer, i.e., for containerized executions. It reads the data of CPU and Memory requested from the Container ENV variables.


hqin2016 (commented):
Nit: perhaps it's better to add final key words to some of the fields in the method body and arguments.

aditya1105:
> Nit: perhaps it's better to add final key words to some of the fields in the method body and arguments.

Thank you. I have added the final keyword at many places.

hqin2016 (approved):
LGTM

PR #2949, opened 2021-09-09
When using cluster routing method to execute jobs, some properties need to be overridden.

djaiswal83:
When using cluster routing method to execute jobs, some properties required to fetch token from HadoopSecurityManager

are overridden. It happens because the cluster routing properties which contain the specific property become part of jobProps,
whereas the default cluster routing property is part of the job's sysProps.
These properties are merged in AbstractProcessJob.getAllProps() where the default property in sysProps overrides the correct
property from jobProps.
This patch uses azkaban property AZ_PLUGIN_LOAD_OVERRIDE_PROPS to fetch list of properties which must be overridden
using cluster router properties into sysProps.
However, the sysProps object is unique per jobtype so it is cloned and the cloned copy is used thereafter thus preserving the
original copy of pluginLoadProps.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -439,17 +439,18 @@ public class JobTypeManager {
       }
 
       // Override any plugin load props if specified.
-      Props pluginLoadPropsCopy = pluginLoadProps;
+      // Make a clone of pluginLoadProps to ensure the original object is not corrupted.
+      // Use the cloned object from here on.
+      final Props pluginLoadPropsCopy = Props.clone(pluginLoadProps);
       if (pluginLoadOverrideProps != null) {
-        // Make a clone of pluginLoadProps
-        pluginLoadPropsCopy = Props.clone(pluginLoadProps);
         final String[] propsList = pluginLoadOverrideProps.split(",");
         for (final String prop : propsList) {
           final String value = clusterSpecificProps.getString(prop, null);
           if (value == null) {
             // The property must be present in cluster specific props
-            throw new JobExecutionException(String.format("Required cluster property %s "
-            + " is not present in ClusterSpecific Properties", prop));
+            logger.warn(String.format("Expected override property %s is not "
+            + " present in ClusterSpecific Properties, ignoring it.", prop));
+            continue;
           }
           pluginLoadPropsCopy.put(prop, value);
         }
--- a/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
+++ b/azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
@@ -64,19 +64,23 @@ public class JobTypeManager {
   private JobTypePluginSet pluginSet;
   // Only used to load keyStore.
   private Props cachedCommonPluginLoadProps;
+  // Overridable plugin load properties
+  private final String pluginLoadOverrideProps;
 
   @VisibleForTesting
   public JobTypeManager(final String jobtypePluginDir, final Props globalProperties,
       final ClassLoader parentClassLoader) {
-    this(jobtypePluginDir, globalProperties, parentClassLoader, new DisabledClusterRouter());
+    this(jobtypePluginDir, globalProperties, parentClassLoader, new DisabledClusterRouter(), null);
   }
 
   public JobTypeManager(final String jobtypePluginDir, final Props globalProperties,
-    final ClassLoader parentClassLoader, ClusterRouter clusterRouter) {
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -433,14 +437,32 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
...
+      Props pluginLoadPropsCopy = pluginLoadProps;
+      if (pluginLoadOverrideProps != null) {
+        // Make a clone of pluginLoadProps
+        pluginLoadPropsCopy = Props.clone(pluginLoadProps);
+        final String[] propsList = pluginLoadOverrideProps.split(",");
+        for (final String prop : propsList) {
+          final String value = clusterSpecificProps.getString(prop, null);
+          if (value == null) {
+            // The property must be present in cluster specific props
+            throw new JobExecutionException(String.format("Required cluster property %s "
```
jakhani, line 451: If property isn't present then we should ignore. Shouldn't we?

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
```diff
@@ -433,14 +437,32 @@ public JobParams createJobParams(final String jobId, Props jobProps, final Logge
           jobProps.put(key, clusterSpecificProps.get(key));
         }
       }
+
+      // Override any plugin load props if specified.
+      Props pluginLoadPropsCopy = pluginLoadProps;
```
jakhani, line 442: Why not create clone here itself? 
Props pluginLoadPropsCopy = Props.clone(pluginLoadProps);

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerWithDynamicClusterTest.java
```diff
@@ -267,4 +269,60 @@ public void testJobRunnerObservesClusterSpecificParams() {
...
+    final JobParams jobParams = manager.createJobParams("testjob", jobProps, LOG);
+    Assert.assertTrue(
+        jobParams.contextClassLoader instanceof HadoopSecurityManagerClassLoader);
+
+    final Job job = JobTypeManager.createJob("testjob", jobParams, LOG);
+    Props sysProps = ((FakeJavaJob2) job).getSysProps();
+    Assert.assertEquals(null, sysProps.get("PropA"));
+
+    // Create JobTypeManager again but with pluginOverrideProps this time.
+    final String pluginLoadOverrideProps = "PropC,PropB";
```
jakhani, line 295: What was the value in system props for this? Sorry I am missing something.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
djaiswal83: The set of properties must be there as we should list only the mandatory ones.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
djaiswal83: Why clone unnecessarily when it is not needed? That is why cloning is done when if condition is true.

azkaban-common/src/test/java/azkaban/jobtype/JobTypeManagerWithDynamicClusterTest.java
djaiswal83: System props does not have it. However, it is supposed to be populated as the two props exist in cluster props.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
jakhani: I think we should ignore rather than failing. Log error and go ahead.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
jakhani: For safety. Later on by mistake if anyone adds anything in Copy before you clone in future.

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
djaiswal83: Ack

azkaban-common/src/main/java/azkaban/jobtype/JobTypeManager.java
djaiswal83: Sure, I was trying to keep it efficient but safety trumps it.

jakhani (approved):

PR #2951, opened 2021-09-13
Fixing a bug to accept false value for dispatch.execution.to.container and move execution to bare metal in that case. Doesn't matter what ramp up % is set for dispatch.

jakhani:
Currently if dispatch.execution.to.container flow param is set to true then it is accepted and execution is moved to containers instead of bare metal execution. There is a bug that if this flow param is set to false explicitly then execution is not going to bare metal which is wrong. This PR is taking care of fixing it.


djaiswal83 (approved):
Thanks for fixing the bug and adding useful log lines.

sshardool (approved):
Thanks for fixing and increasing the logging around this.

PR #2952, opened 2021-09-20
Allow only one ACTIVE image version for a given image type.

djaiswal83:
The Image Management APIs allow multiple image versions of same type to remain in ACTIVE state.

This leads to unexpected behavior of getting incorrect version among the ACTIVE ones for executing flows.
Introducing new image version state STABLE.
If an image is set as ACTIVE using the ACTIVATE API, all the other image versions which are currently
set to ACTIVE will be set to STABLE state.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
@@ -354,8 +354,27 @@ public class ImageVersionDaoImpl implements ImageVersionDao {
       queryBuilder.append(" where iv.id = ? and iv.type_id = it.id and it.name = ?");
       params.add(imageVersionRequest.getId());
       params.add(imageVersionRequest.getName());
-      final int updateCount = this.databaseOperator.update(queryBuilder.toString(),
-          Iterables.toArray(params, Object.class));
+      int updateCount = 0;
+      if (imageVersionRequest.getState() != State.ACTIVE) {
+        updateCount = this.databaseOperator.update(queryBuilder.toString(),
+            Iterables.toArray(params, Object.class));
+      } else {
+        // If the state is being set to ACTIVE then only the current image version can be set
+        // to ACTIVE and the rest of the image versions with ACTIVE state must be set to STABLE.
+        final SQLTransaction<Integer> setCurrentVersionActive = transOperator -> {
+          // Set all the current active versions to STABLE
+          final String setStable = String.format("update image_versions iv, image_types it " +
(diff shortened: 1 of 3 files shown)


hqin2016 (approved):

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -354,8 +354,27 @@ public void updateImageVersion(final ImageVersion imageVersionRequest)
       queryBuilder.append(" where iv.id = ? and iv.type_id = it.id and it.name = ?");
       params.add(imageVersionRequest.getId());
       params.add(imageVersionRequest.getName());
-      final int updateCount = this.databaseOperator.update(queryBuilder.toString(),
-          Iterables.toArray(params, Object.class));
+      int updateCount = 0;
+      if (imageVersionRequest.getState() != State.ACTIVE) {
```
jakhani, line 358: Shall we also put a check for state machine that from which all state transition is possible to ACTIVE?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
```diff
@@ -354,8 +354,27 @@ public void updateImageVersion(final ImageVersion imageVersionRequest)
...
+        updateCount = this.databaseOperator.update(queryBuilder.toString(),
+            Iterables.toArray(params, Object.class));
+      } else {
+        // If the state is being set to ACTIVE then only the current image version can be set
+        // to ACTIVE and the rest of the image versions with ACTIVE state must be set to STABLE.
+        final SQLTransaction<Integer> setCurrentVersionActive = transOperator -> {
+          // Set all the current active versions to STABLE
+          final String setStable = String.format("update image_versions iv, image_types it " +
+              " set iv.state = 'STABLE' where iv.type_id = it.id and it.name = '%s' "
```
jakhani, line 367: Please use enum name instead of string for STABLE state

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
djaiswal83: That is out of scope for this change, let me create a ticket to track it.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageVersionDaoImpl.java
djaiswal83: Will create a followup PR to make the change.

PR #2953, opened 2021-09-20
Removing method doRampActions from ContainerizedDispatchManager and utilize super method

aditya1105:
Parent class AbstractExecutorManagerAdapter does implement the doRampActions method which allows making rampup changes for the executors. 

Once containerization is enabled, ContainerizedDispatchManager is utilized. Hence even if the rampup is not 100%, we see UnsupportedOperationException which shouldn't be the case.


djaiswal83 (approved):
LGTM

PR #2954, opened 2021-09-24
In KubernetesContainerizedImpl, add ability to use flow properties to fetch execution params.

djaiswal83:
In KubernetesContainerizedImpl, add ability to use flow properties to fetch execution params.

The flow properties are merged with flow parameters on top to ensure the precedence is maintained.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -45,6 +45,7 @@ import azkaban.imagemgmt.version.VersionSet;
 import azkaban.imagemgmt.version.VersionSetBuilder;
 import azkaban.imagemgmt.version.VersionSetLoader;
 import azkaban.metrics.ContainerizationMetrics;
+import azkaban.project.ProjectLoader;
 import azkaban.spi.EventType;
 import azkaban.utils.Props;
 import com.google.common.annotations.VisibleForTesting;
@@ -157,6 +158,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
   private final ContainerizationMetrics containerizationMetrics;
   private final String azkabanBaseImageName;
   private final String azkabanConfigImageName;
+  private final ProjectLoader projectLoader;
 
 
   private static final Logger logger = LoggerFactory
@@ -169,7 +171,8 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
       final ImageRampupManager imageRampupManager,
       final KubernetesWatch kubernetesWatch,
       final EventListener eventListener,
-      final ContainerizationMetrics containerizationMetrics)
+      final ContainerizationMetrics containerizationMetrics,
+      final ProjectLoader projectLoader)
       throws ExecutorManagerException {
     this.azkProps = azkProps;
     this.executorLoader = executorLoader;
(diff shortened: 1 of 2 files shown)


aditya1105:
This change is very specific to one use case and hence any other requirement of using flow props will require similar changes across. There will be more similar requirements and all of them need not go through the containerized dispatch manager. 

It is better to keep it at one common place, i.e., in the class azkaban.executor.ExecutableFlow
This class has a constructor which accepts Flow object and we can utilize the method azkaban.flow.Flow#getAllFlowProps and add the relevant properties from there to the ExecutionOptions. Throughout the codebase, ExecutionOptions are utilized and hence there will be consistency.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -716,6 +720,25 @@ V1Pod createPodFromSpec(final int executionId, final V1PodSpec podSpec, Map<Stri
...
+   * @param flow the executable flow
+   * @return flattened Props object with all the flow properties.
+   */
+  private Props getFlowProps(final ExecutableFlow flow) {
+    Props commonProps = new Props();
+    final Map<String, Props> propsMap =
+        projectLoader.fetchProjectProperties(flow.getProjectId(), flow.getVersion());
+    if (null != propsMap) {
+      // Flatten the props as long list of props objects may cause stack overflow.
+      for (final Props props : propsMap.values()) {
```
jakhani, line 735: Good to use propsMap.forEach and lambda.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -716,6 +720,25 @@ V1Pod createPodFromSpec(final int executionId, final V1PodSpec podSpec, Map<Stri
...
+  /**
+   *  This method fetches all the flow properties from the db and flattens
+   *  them into a single Props object.
+   * @param flow the executable flow
+   * @return flattened Props object with all the flow properties.
+   */
+  private Props getFlowProps(final ExecutableFlow flow) {
+    Props commonProps = new Props();
+    final Map<String, Props> propsMap =
+        projectLoader.fetchProjectProperties(flow.getProjectId(), flow.getVersion());
```
jakhani, line 732: Will this cover both Flow1.0 and Flow2.0?

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
```diff
@@ -262,7 +262,7 @@ public void testPodConstruction() throws Exception {
     allImageTypes.addAll(jobTypes);
     allImageTypes.addAll(dependencyTypes);
     final VersionSet versionSet = this.kubernetesContainerizedImpl
-        .fetchVersionSet(flow.getExecutionId(), flowParam, allImageTypes, flow);
+        .fetchVersionSet(flow.getExecutionId(), new Props(null, flowParam), allImageTypes, flow);
```
jakhani, line 265: Can you please add valid test case which will test project props and precedence as well?

djaiswal83: I created this PR for reference only. Will work on production code once we agree on the approach.

djaiswal83:
> This change is very specific to one use case and hence any other requirement of using flow props will require similar changes across. There will be more similar requirements and all of them need not go through the containerized dispatch manager.
> 
> It is better to keep it at one common place, i.e., in the class azkaban.executor.ExecutableFlow
> This class has a constructor which accepts Flow object and we can utilize the method azkaban.flow.Flow#getAllFlowProps and add the relevant properties from there to the ExecutionOptions. Throughout the codebase, ExecutionOptions are utilized and hence there will be consistency.

The idea was to showcase the solution. A PR with proper code change and test cases will come once we agree on the approach.

PR #2955, opened 2021-09-24
Remove the nscd setup by default from code

hqin2016:
Tested alternative approach on containerized cluster by adding nscd volume on pod-spec-template.


djaiswal83 (approved):
The change looks good to me.

jakhani (approved):
We don't have any unit test around this? No change needed there?

hqin2016:
> We don't have any unit test around this? No change needed there?

Yes we do. It's in AzKubernetesV1PodBuilderTest.java. However it's called as a hard-coded form             ```
.addHostPathVolume("nscd-socket", "/var/run/nscd/socket", "Socket",
                "/var/run/nscd/socket", true) 
```
not the method addNscdSocketInVolume deleted here.

aditya1105 (approved):

PR #2958, opened 2021-10-01
Utilize azkaban.cluster.env property to set the AZ_CLUSTER ENV variable for containerization

aditya1105:
clusterEnv is the class variable that will be set using property azkaban.cluster.env, which defaults to class variable clusterName for backward compatibility.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -125,6 +125,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
   private final String podPrefix;
   private final String servicePrefix;
   private final String clusterName;
+  private final String clusterEnv;
   private final String flowContainerName;
   private final String cpuLimit;
   private final String cpuRequest;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -189,6 +190,9 @@ public KubernetesContainerizedImpl(final Props azkProps,
             DEFAULT_SERVICE_NAME_PREFIX);
     this.clusterName = this.azkProps.getString(ConfigurationKeys.AZKABAN_CLUSTER_NAME,
         DEFAULT_CLUSTER_NAME);
+    // This is utilized to set AZ_CLUSTER ENV variable to the POD containers.
+    this.clusterEnv = this.azkProps.getString(ConfigurationKeys.AZKABAN_CLUSTER_ENV,
```
djaiswal83, line 194: In what scenarios can the cluster name be different from the env?

aditya1105: Environment variables are utilized within the containers, mostly by the scripts running in the container/init-container. 
Cluster Name is utilized at more than one place like pod names, pod labels, API gateway for reverse proxy, etc.

Hence the value of cluster name has restrictions like "underscore" can't be part of it. However, environment variables have no such restriction.
Moreover, the environment variable name has no dependency with pod names, pod labels, API gateway, etc. hence it can be set independently of cluster name. The default value is set to clusterName for backward compatibility.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -535,7 +539,7 @@ V1PodSpec createPodSpec(final int executionId, final VersionSet versionSet,
     final String flowContainerCPURequest = getFlowContainerCPURequest(flowParam);
     final String flowContainerMemoryRequest = getFlowContainerMemoryRequest(flowParam);
     final AzKubernetesV1SpecBuilder v1SpecBuilder =
-        new AzKubernetesV1SpecBuilder(this.clusterName, Optional.empty())
+        new AzKubernetesV1SpecBuilder(this.clusterEnv, Optional.empty())
```
sshardool, line 542: Please check if a unit test output needs to be updated for this.

sshardool (approved):
Thanks for addressing this.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
aditya1105: Out unit test relies on the variable name passed to the spec builder rather than calling the KubernetesContainerizedImpl. So the unit tests are ok with this.

PR #2959, opened 2021-10-05
Load project white list .xml file in FlowRunnerManager and FlowContainer

hqin2016:
Users reported that setting flow.num.job.threads = 128 in the runtime properties does not spin up a flow with  128 concurrently running jobs, even the project is added to the white list .xml in executor  server conf/ directory. 
This bug is caused by project white list file not loaded, because ProjectManager (which is responsible for loading this file) may not instantiate in executor server. 
Therefore, this bug fix aims to load project white list file in FlowRunnerManager (bare metal execution) or FlowContainer (containerized execution).

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -54,6 +54,8 @@ import azkaban.jobtype.JobTypeManager;
 import azkaban.metrics.CommonMetrics;
 import azkaban.metrics.MetricsManager;
 import azkaban.project.ProjectLoader;
+import azkaban.project.ProjectWhitelist;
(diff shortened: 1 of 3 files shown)


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -378,6 +381,29 @@ private void createFlowRunner(final ExecutableFlow flow) throws ExecutorManagerE
       watcher = new RemoteFlowWatcher(pipelinedExecId, this.executorLoader);
     }
 
+    int numJobThreads = this.numJobThreadPerFlow;
```
aditya1105, line 384: This code looks very similar to what we already have. Is it possible to refactor it?

aditya1105 (approved):

PR #2979, opened 2021-11-10
Use fetchProperty instead of fetchProperties which is compute heavy and has memory leaks.

djaiswal83:
This also adds requirement for users to create param_override.properties which will contain all override properties.
Updated the tests accordingly.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -153,6 +153,9 @@ public class Constants {
   // Overridable plugin load properties
   public static final String AZ_PLUGIN_LOAD_OVERRIDE_PROPS = "azkaban.plugin.load.override.props";
 
+  // File containing param override configs
+  public static final String PARAM_OVERRIDE_FILE = "param_override.properties";
+
   // Azkaban event reporter constants
   public static class EventReporterConstants {
 
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -15,6 +15,7 @@
  */
 package azkaban.executor;
 
+import azkaban.Constants;
 import azkaban.DispatchMethod;
 import azkaban.flow.Flow;
 import azkaban.imagemgmt.version.VersionSet;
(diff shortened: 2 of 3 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -153,6 +153,9 @@
   // Overridable plugin load properties
   public static final String AZ_PLUGIN_LOAD_OVERRIDE_PROPS = "azkaban.plugin.load.override.props";
 
+  // File containing param override configs
```
anish-mal, line 156: It will be helpful to specify the intended priority as well for the properties within this file.

anish-mal (commented):
Since this is a alternative implementation of an already existing feature, I guess its testcases already exist. Is there a one that checks the 'override' from the file?

djaiswal83:
> Since this is a alternative implementation of an already existing feature, I guess its testcases already exist. Is there a one that checks the 'override' from the file?

Yes. Check the test file modification which was updated.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -486,26 +487,16 @@ public void setVersionSet(final VersionSet versionSet) {
...
-      }
-    }
+    Props props = FlowLoaderUtils.isAzkabanFlowVersion20(this.azkabanFlowVersion) ?
+        FlowLoaderUtils.loadPropsFromYamlFile(projectLoader, this, null) :
+        projectLoader.fetchProjectProperty(projectId, version, Constants.PARAM_OVERRIDE_FILE);
 
-    // Filter out the properties to keep only the override ones.
-    Map<String, String> flowOverridePropsMap = props.getMapByPrefix(PARAM_OVERRIDE);
+    // Clone the props object and filter out the properties to keep only the override ones.
+    Map<String, String> flowOverridePropsMap = Props.clone(props).getMapByPrefix(PARAM_OVERRIDE);
```
jakhani, line 495: getMapByPrefix is creating map. So we don't need to clone props here. Please double check it. Why to waster memory by cloning.

az-core/src/main/java/azkaban/Constants.java
djaiswal83: I dont understand. It is pretty evident from the name and comment.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
djaiswal83: This is to keep a distance from original Props object which was probable cause of memory leak.
Let me know if you still want me to remove it.

az-core/src/main/java/azkaban/Constants.java
anish-mal: 1.
For a directory structure, property files in Proj_Dir used to have lower precedence than A. For the newly introduced file, this will no longer be true

Proj_Dir
  basic.properties
  param_override.properties
  A/
    foo_a.properties
    foo.job
    
i.e. 
(a). param_override.properties precedence will be higher than foo_a.properties. 
(b). foo_a.properties precedence will be higher than that of basic.properties.

Hence this should be very well documented.


2. 
Furthermore, so far UI specified parameters take the highest precedence. I believe for this file, this won't be the case - which can be very counter intuitive for a user making changes in the UI.

I would say the UI should always have higher precedence. Any reason why that would be undesired in this case? Assuming, it's a valid reason, that reason also ought to be documented.

djaiswal83: Resolving as we discussed it offline

jakhani (approved):

az-core/src/main/java/azkaban/Constants.java
anish-mal: In addition to add this to the documentation (https://azkaban.github.io/azkaban/docs/latest/), a comment over here will be very helpful when someone else reads this code. It will be good to add intent as well as the rationale as to why we wanted a file based property to override a UI-specified parameter.

anish-mal:
testFlowPropertyAndParamsMergeWithOverwrite() looks good.

flowProps and Param code is already convoluted and is getting more convoluted. Comments will be helpful - as suggested in https://github.com/azkaban/azkaban/pull/2979#discussion_r747782117

az-core/src/main/java/azkaban/Constants.java
djaiswal83: I will raise a followup PR with comments.

PR #2980, opened 2021-11-11
Move the project upload version update to the very end after all the project related files are uploaded in the db.

djaiswal83:
If the version is updated before property files are uploaded there can be a race condition where a flow comes in and uses the updated version and webserver may need to use one of the project properties which are not yet uploaded.


jakhani (commented):
Can you please add/modify some unit tests around this?

jakhani (approved):
Let's cover extensive unit tests for project upload in next PR.

PR #2981, opened 2021-11-12
Fix the NPE by adding a null check before dereferencing flow object.

djaiswal83:


anish-mal (approved):
Thanks!

jakhani (approved):

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -86,7 +86,9 @@ public void updateExecutableFlow(final ExecutableFlow flow)
   public ExecutableFlow fetchExecutableFlow(final int id)
       throws ExecutorManagerException {
     final ExecutableFlow flow = this.executionFlowDao.fetchExecutableFlow(id);
-    flow.setFlowPropsAndParams(this.projectLoader);
+    if (null != flow) {
+      flow.setFlowPropsAndParams(this.projectLoader);
```
sshardool, line 90: Maybe log this as a warning when the flow is not `null` ?

PR #2983, opened 2021-11-16
Make metrics reporting more open source friendly.

ypadron-in:
Azkaban uses the Dropwizard Metrics library to collect metrics to gauge how the app is doing. To report those metrics for proper storage and plotting for example, it allows custom reporters to be hooked in.
Before, to connect a custom reporter the fully qualified name of its class had to be provided in a config property and the class should have a constructor with the following signature: `(String reporterName, MetricRegistry registry, String metricsServerURL, Boolean someFlag)`

This approach leaks LinkedIn specific settings for reporting and forces users to implement a constructor with parameters that might not make sense to them. In addition, it's not extensible as to accommodate new configs, in the past the constructor signature has been modified making it a backward incompatible change.

To make the configuration of custom reporters more open source friendly (able to support reporter specific settings and can withstand changes) a constructor with the following signature can be used instead: `(MetricRegistry registry, String reporterConfigsPath)`
- A MetricRegistry should always be required as it contains all the app metrics.
- Optionally, the path to a file containing custom reporter specific settings can be specified. The file format and how to read it is up to the reporter implementation.

### DIFF
--- a/az-core/src/main/java/azkaban/metrics/MetricsManager.java
+++ b/az-core/src/main/java/azkaban/metrics/MetricsManager.java
@@ -17,7 +17,7 @@
 package azkaban.metrics;
 
 import static azkaban.Constants.ConfigurationKeys.CUSTOM_METRICS_REPORTER_CLASS_NAME;
-import static azkaban.Constants.ConfigurationKeys.METRICS_SERVER_URL;
+import static azkaban.Constants.ConfigurationKeys.CUSTOM_METRICS_REPORTER_CONFIG_PATH;
 
 import azkaban.utils.Props;
 import com.codahale.metrics.Counter;
@@ -33,13 +33,14 @@ import java.lang.reflect.Constructor;
 import java.util.function.Supplier;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * The singleton class, MetricsManager, is the place to have MetricRegistry and ConsoleReporter in
- * this class. Also, web servers and executors can call {@link #startReporting(String, Props)} to
- * start reporting AZ metrics to remote metrics server.
+ * this class. Also, web servers and executors can call {@link #startReporting(Props)} to start
+ * reporting AZ metrics to remote metrics server.
  */
 @Singleton
 public class MetricsManager {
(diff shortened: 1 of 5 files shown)


az-core/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -114,27 +115,27 @@ public Timer addTimer(final String name) {
    * reporting metrics to remote metrics collector. Note: this method must be synchronized, since
    * both web server and executor will call it during initialization.
    */
-  public synchronized void startReporting(final String reporterName, final Props props) {
+  public synchronized void startReporting(final Props props) {
     final String metricsReporterClassName = props.get(CUSTOM_METRICS_REPORTER_CLASS_NAME);
-    final String metricsServerURL = props.get(METRICS_SERVER_URL);
-    if (metricsReporterClassName != null && metricsServerURL != null) {
+    final String metricsReporterConfigPath =
```
jakhani, line 120: Can you please move this to inner if case? It is not needed outside.

az-core/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -114,27 +115,27 @@ public Timer addTimer(final String name) {
...
+        final Constructor<?> ctor = metricsClass.getConstructor(this.registry.getClass(),
+            metricsReporterConfigPath.getClass());
+        ctor.newInstance(this.registry, metricsReporterConfigPath);
 
       } catch (final Exception e) {
-        log.error("Encountered error while loading and instantiating "
-            + metricsReporterClassName, e);
-        throw new IllegalStateException("Encountered error while loading and instantiating "
-            + metricsReporterClassName, e);
+        final String errMsg = "Encountered an error while loading or instantiating the custom "
```
jakhani, line 132: If possible, try to keep complete text message in one line. It will be easy to search.

jakhani (commented):
Thank you for adding this fix!

az-core/src/main/java/azkaban/metrics/MetricsManager.java
ypadron-in: done, thanks!

az-core/src/main/java/azkaban/metrics/MetricsManager.java
ypadron-in: you're right, done.

jakhani (approved):

PR #2985, opened 2021-11-18
set flow status to EXECUTION_STOPPED when Infra failure (Containerized) occurs

hqin2016:


azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -182,6 +182,7 @@ public FetchActiveFlowDao(final DatabaseOperator dbOperator) {
             + " LEFT JOIN "
             + " executors et ON ex.executor_id = et.id"
             + " WHERE ex.status NOT IN ("
+            + Status.EXECUTION_STOPPED.getNumVal() + ", "
```
jakhani, line 185: This method can cause bugs. Why don't you use list of finalize statuses and then use join by comma and that will give you correct value.

Use Apache Commons Lang library

List<Integer> list = Arrays.asList(1,2,3,4,5,6,7);
String str = org.apache.commons.lang.StringUtils.join(list, ",");

You can use it at multiple places shown in this diff.

hqin2016: Thanks for the suggestion. The entire code block is error prone. I'll refactor it.

jakhani (approved):

djaiswal83 (approved):
Looks good to me.

PR #2986, opened 2021-11-18
Fix test ContainerizedDispatchManagerTest#testRestartFlow to use thread.join instead of wait

aditya1105:


djaiswal83 (approved):

azkaban-common/src/test/java/azkaban/executor/ContainerizedDispatchManagerTest.java
```diff
@@ -549,7 +549,7 @@ public void testRestartFlow() throws Exception {
         this.containerizedDispatchManager.getExecutionDispatcher(this.flow1.getExecutionId()));
     thread.start();
     synchronized (thread) {
```
ypadron-in, line 551: I think we don't need synchronization here.

PR #2991, opened 2021-11-24
Adjust build settings for Gradle 5.

ypadron-in:


hqin2016 (approved):
Thanks a lot for the fix! You saved the world.

PR #3010, opened 2021-12-06
Merge metadata in template pod spec.

ypadron-in:
Add support for combining pod metadata generated by the Azkaban code and metadata in the template pod spec.
Only merging labels and annotations for now.
Metadata from the template pod spec has precedence when resolving collisions.


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -589,6 +591,16 @@ V1PodSpec createPodSpec(final int executionId, final VersionSet versionSet,
...
   }
 
+  @VisibleForTesting
+  V1ObjectMeta createPodMetadata(final int executionId, Map<String, String> flowParam) {
+    return new V1ObjectMetaBuilder()
+        .withName(getPodName(executionId))
+        .withNamespace(this.namespace)
+        .addToLabels(getLabelsForPod(executionId, flowParam))
+        .addToAnnotations(getAnnotationsForPod())
+        .build();
```
ypadron-in, line 601: Creating a `AzKubernetesV1MetadataBuilder` class similar to `AzKubernetesV1SpecBuilder` seems unnecessary for now as no additional processing is required. Methods of `AzKubernetesV1MetadataBuilder` would only call corresponding methods of `V1ObjectMetaBuilder` like this:
```
public class AzKubernetesV1MetadataBuilder {
    private final V1ObjectMetaBuilder v1ObjectMetaBuilder = new V1ObjectMetaBuilder();

    public AzKubernetesV1MetadataBuilder() {
    }

    public AzKubernetesV1MetadataBuilder addLabels(final ImmutableMap<String, String> labels) {
        this.v1ObjectMetaBuilder.addToLabels(labels);
        return this;
    }

    public AzKubernetesV1MetadataBuilder addAnnotations(final ImmutableMap<String, String> annotations) {
        this.v1ObjectMetaBuilder.addToAnnotations(annotations);
        return this;
    }

    public AzKubernetesV1MetadataBuilder withPodName(final String podName) {
        this.v1ObjectMetaBuilder.withName(podName);
        return this;
    }

    public AzKubernetesV1MetadataBuilder withNameSpace(final String nameSpace) {
        this.v1ObjectMetaBuilder.withNamespace(nameSpace);
        return this;
    }

    public V1ObjectMeta build() {
        return this.v1ObjectMetaBuilder.build();
    }
}
```

azkaban-common/src/test/resources/azkaban/executor/container/v1PodTestTemplate1.yaml
```diff
@@ -47,15 +55,13 @@ spec:
         initialDelaySeconds: 5
         periodSeconds: 5
       volumeMounts:
+        - mountPath: /export/some/path
+          name: jobtype-volume-spark
```
ypadron-in, line 59: Values of container "mountPath" (`/export/some/path`) and init container "mountPath" (`/data/jobtypes/spark2`) above does not match those in the pod spec generated by Azkaban code however, the final pod spec is resolved with the latter values (`/export/apps/azkaban/azkaban-exec-server/current/plugins/jobtypes/spark` and `/data/jobtypes/spark` respectively). 
@aditya1105 Is that expected? Why the values in the template pod spec don't overwrite values in the pod spec generated by Azkaban?

azkaban-common/src/test/resources/azkaban/executor/container/v1PodTestTemplate1.yaml
```diff
@@ -47,15 +55,13 @@ spec:
         initialDelaySeconds: 5
         periodSeconds: 5
       volumeMounts:
+        - mountPath: /export/some/path
```
ypadron-in, line 58: If I change the container name from `az-flow-container` in the template pod spec to `template-az-flow-container` for example, below spec portion still gets applied to container `az-flow-container` in the final pod spec
```
livenessProbe:
  exec:
    command:
      - cat
      - /tmp/liveProbe
  initialDelaySeconds: 5
  periodSeconds: 5
readinessProbe:
  exec:
    command:
      - cat
      - /tmp/readyProbe
```
Why is that? I would expect above yaml portion to be missing from final pod spec since container names didn't match. Or perhaps an error saying that the template pod spec can only contain 1 non init container with name `az-flow-container`.

aditya1105: The template and the dymanic one which gets created by webserver are merged together if they have the same container names . For app/flow container, we need exactly one container, so if the names are different in the pod-spec and the one passed to the builder, this can occur. We need to change the name of "az-flow-container" at both places in pod-spec-template and in the builder test.

azkaban-common/src/test/resources/azkaban/executor/container/v1PodTestTemplate1.yaml
aditya1105: Nearly at every place, the higher priority is with the generated pod-spec rather than the template one. This is the case at many places like image path, image version, pull policy, mount path, etc.
If the volume name is the same, then the final path is overridden by the generated one.

aditya1105 (approved):

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
```diff
@@ -71,7 +73,17 @@ public static synchronized AzKubernetesV1PodTemplate getInstance(String template
    * @return the {@link V1PodSpec} POD spec generated from the template.
    * @throws IOException If unable to read the template file.
    */
-  public synchronized V1PodSpec getPodSpecFromTemplate() throws IOException {
+  public V1PodSpec getPodSpecFromTemplate() throws IOException {
```
jakhani, line 76: This method doesn't need to be synchronized anymore?

jakhani (commented):
Can you please share why did you delete AzKubernetesV1PodBuilderTest.java class?

ypadron-in:
> Can you please share why did you delete AzKubernetesV1PodBuilderTest.java class?

Please check my comment above (https://github.com/azkaban/azkaban/pull/3010#discussion_r763252939):
> Test in AzKubernetesV1PodBuilderTest.java and KubernetesContainerizedImplTest.testPodConstruction was doing similar things to so I merged them.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodTemplate.java
ypadron-in: I believe this code isn't called from different threads and we're just reading the data so synchronized isn't needed.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1PodBuilder.java
```diff
@@ -30,53 +30,25 @@
  * @see io.kubernetes.client.openapi.models to create kubernetes "Pod" resource.
  */
 public class AzKubernetesV1PodBuilder {
-    private static final Logger LOGGER =
```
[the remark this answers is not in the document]
ypadron-in: The code that was using the logger is no longer needed so the logger has no usages in the class. We can add it again when needed.

jakhani (approved):

PR #3011, opened 2021-12-06
Implemented new REST API hpFlows to add high priority flows to Azkaban.

djaiswal83:
These high priority flows when run in containeirzed dispatch mode, will only use ACTIVE version of the docker image of all the jobtypes required.
This ensures that these flows always run on a stable version rather than any version which is being ramped up and is potentially unstable.
There API can be only called by either Azkaban admins or the designated owners.
Currently the ability to add owners is a manual process. An ADMIN only API to add such owners will come in the future.
API to remove a flow or list of flows from the high priority flows will also come in future.

Added unit tests as and where needed.


ypadron-in:
I'm lacking some context here but I wonder why can't we use a flow/job level property to instruct Azkaban to use a stable jobtype version when creating the pod spec? That shouldn't require a lot of code changes, no need to add a new table in the DB either, no follow up or maintenance work from our side like this approach will require. 

I have a vague memory that users can specify the Azkaban bin version to use for their executions. Isn't that available? Can't we extend that to support generic values like "stable" or "latest".

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/HPFlowServlet.java
```diff
@@ -0,0 +1,129 @@
...
+   * @return
+   */
+  private boolean hasHPFlowManagementPermission(final User user) {
+    /**
+     * Azkaban ADMIN role must have full permission to access image management APIs. Hence, no
+     * further permission check is required.
+     */
+    if (isAzkabanAdmin(user)) {
+      return true;
+    }
```
ypadron-in, line 123: The new api endpoint adds maintenance overhead on the Azkaban team as it involves putting in place a process to add high priority flows. Why is this feature available to high priority flows only? What's a high priority flow?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/HPFlowDaoImpl.java
```diff
@@ -0,0 +1,167 @@
...
+  private final DatabaseOperator databaseOperator;
+
+  private static String INSERT_HP_FLOW_OWNER =
+      "INSERT into hp_flow_owners (name, created_on, created_by, "
+      + "modified_on, modified_by ) "
+      + "values (?, ?, ?, ?, ?)";
+
+  private static String INSERT_HP_FLOWS =
+      "INSERT into hp_flows (flow_id, created_on, created_by) "
+      + "values (?, ?, ?)";
```
ypadron-in, line 37: What happens if a project containing high priority flows is deleted or a high priority flow is removed in a certain project version? How do we make sure this table is up to date? How do we prevent this table from growing indefinitely?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/HPFlowDaoImpl.java
```diff
@@ -0,0 +1,167 @@
...
+
+  private final DatabaseOperator databaseOperator;
+
+  private static String INSERT_HP_FLOW_OWNER =
+      "INSERT into hp_flow_owners (name, created_on, created_by, "
+      + "modified_on, modified_by ) "
+      + "values (?, ?, ?, ?, ?)";
+
+  private static String INSERT_HP_FLOWS =
+      "INSERT into hp_flows (flow_id, created_on, created_by) "
```
ypadron-in, line 36: What's an example of `flow_id`? If I remember correctly Azkaban doesn't generate numeric flow ids like it does for executions. It uses the flow name as the id but the same name can be used in different projects. How are we resolving name conflicts?

ypadron-in (changes requested):

djaiswal83:
> 

@ypadron-in  this was extensively reviewed by our partners and this is the approach they preferred. The idea is the jobtype owners want to control which flow is marked high priority and which one is not.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/HPFlowServlet.java
djaiswal83: As the name suggests a flow which has high business value is a high priority flow. With containerization and image rampup feature, the jobtype owners want to make sure that a flow which has high priority runs on ACTIVE version rather than any version which is being ramped up and could be potentially unstable.

djaiswal83: The API will be eventually self served a list of owners.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/HPFlowDaoImpl.java
djaiswal83: As of now the check is not added, it will be added in the future as we add more APIs around the feature.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/HPFlowDaoImpl.java
djaiswal83: Good point, I will add a comment. The flow id here is,
project_name.flow_name

PR #3012, opened 2021-12-07
Instantiate OnExecutionEventListener for Containerization

aditya1105:
The onExecutionEventListener was not getting initialized and hence when trying to restart the flow we were facing NPE.
This change fixes the issue by instantiating it at the start of the webserver.


hqin2016 (approved):
Thanks for the fix!

djaiswal83 (approved):
Thanks for making this change on such short notice.

PR #3017, opened 2021-12-08
Added a way to support custom hostname to emit metrics

svnarumugam:
This change will enable us to pass custom hostname (usually a fixed one) to emit and refer metrics in containerized Azkaban web server.

Tested this change in conjunction with AmfReporter library update, and observed that metrics got emitted.

### DIFF
--- a/az-core/src/main/java/azkaban/metrics/MetricsManager.java
+++ b/az-core/src/main/java/azkaban/metrics/MetricsManager.java
@@ -16,9 +16,6 @@
 
 package azkaban.metrics;
 
-import static azkaban.Constants.ConfigurationKeys.CUSTOM_METRICS_REPORTER_CLASS_NAME;
-import static azkaban.Constants.ConfigurationKeys.METRICS_SERVER_URL;
-
 import azkaban.utils.Props;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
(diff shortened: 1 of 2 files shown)


az-core/src/main/java/azkaban/metrics/MetricsManager.java
```diff
@@ -117,14 +117,19 @@ public Timer addTimer(final String name) {
   public synchronized void startReporting(final String reporterName, final Props props) {
     final String metricsReporterClassName = props.get(CUSTOM_METRICS_REPORTER_CLASS_NAME);
     final String metricsServerURL = props.get(METRICS_SERVER_URL);
+    final String customHostNameToEmitMetrics = props.get(CUSTOM_METRICS_HOSTNAME);
```
sshardool, line 120: What's the behavior when the property `CUSTOM_METRICS_HOSTNAME` is not set ?
Ideally we would wan't the default behavior to stay unchanged, if that's the case please mention that in a comment. Please let me know if you think otherwise.

sshardool (commented):
Thanks for taking this up, overall change looks good except one comment below.

az-core/src/main/java/azkaban/metrics/MetricsManager.java
svnarumugam: this property is an optional one. If we don't set it, the default behaviour stays as it is. Let me update the PR description too. Thanks for taking a look!

PR #3018, opened 2021-12-09
Fixing restart by ContainerCleanupManager

aditya1105:
When the cancel throws an exception, then the flow is not restarted. Even if the flow is not cancelled properly, it is still finalized. Hence restart should happen.
The below PR tries to restart the flow even if the cancel throws an exception.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -135,7 +135,9 @@ public class ContainerCleanupManager {
       if (shouldIgnore(flow, status)) {
         continue;
       }
-      cancelFlowQuietly(flow);
+      Status originalStatus = flow.getStatus();
+      cancelFlowQuietly(flow, originalStatus);
+      retryFlowQuietly(flow, originalStatus);
       deleteContainerQuietly(flow.getExecutionId());
     }
   }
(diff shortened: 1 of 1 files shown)


djaiswal83 (approved):
Thanks for making this change on a short notice. Looks good to me.
Lets keep the PR around for sometime.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -170,21 +172,28 @@ private boolean shouldIgnore(final ExecutableFlow flow, final Status status) {
...
       logger.error("ExecutorManagerException while cancelling flow.", eme);
     } catch (RuntimeException re) {
       logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
     }
   }
 
+  private void retryFlowQuietly(ExecutableFlow flow, Status originalStatus) {
+    try {
+      logger.info("Restarting cleaned up flow " + flow.getExecutionId());
+      ExecutionControllerUtils.restartFlow(flow, originalStatus);
```
jakhani, line 191: Validation of status to whether it is restartable or not is taken care of inside restartFlow so no validation needed here right?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -135,7 +135,9 @@ public void cleanUpStaleFlows(final Status status) {
       if (shouldIgnore(flow, status)) {
         continue;
       }
-      cancelFlowQuietly(flow);
+      Status originalStatus = flow.getStatus();
+      cancelFlowQuietly(flow, originalStatus);
+      retryFlowQuietly(flow, originalStatus);
```
jakhani, line 140: I don't understand how moving ExecutionControllerUtils.restartFlow(flow, originalStatus); from cancelFlowQuietly to retryFlowQuietly help. Can you please share detail about it as well as cover it in comments?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
aditya1105: Yes, that logic is already part of ExecutionControllerUtils.restartFlow

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
aditya1105: The problem is that cancelFlowQuietly executes "containerizedDispatchManager.cancelFlow" first and then tries to execute ExecutionControllerUtils.restartFlow
If the containerizedDispatchManager.cancelFlow throws an exception, then the ExecutionControllerUtils.restartFlow wouldn't have been called.
As part of the design containerizedDispatchManager.cancelFlow, is supposed to throw exceptions if the flow is unreachable but the flow will still be finalized by it. Hence we should execute ExecutionControllerUtils.restartFlow.

This change makes sure that both of them are called separately.

PR #3019, opened 2021-12-10
Emit FLOW_FINISHED events for cancelled flow

hqin2016:
From the events, especially EXECUTION_STOPPED status, analytics can be done on containerized executions.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
@@ -19,6 +19,10 @@ import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 import azkaban.Constants.ContainerizedDispatchManagerProperties;
+import azkaban.event.Event;
+import azkaban.event.EventData;
+import azkaban.event.EventHandler;
+import azkaban.event.EventListener;
 import azkaban.executor.AlerterHolder;
 import azkaban.executor.ExecutableFlow;
 import azkaban.executor.ExecutionControllerUtils;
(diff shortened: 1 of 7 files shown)


azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
```diff
@@ -297,6 +302,12 @@ public void testFlowManagerListenerInitContainerFailure() throws Exception {
     // Setup a FlowUpdatingListener
     Props azkProps = new Props();
     FlowStatusManagerListener updatingListener = flowStatusUpdatingListener(azkProps);
+    // Verify EXECUTION_STOPPED flow life cycle event is emitted
+    updatingListener.addListener((event) -> {
```
djaiswal83, line 306: The code is duplicated at 3 places. Put it in a common method and call it here.

djaiswal83 (changes requested):
Thanks for making this change. This is a useful event.
I have a minor comment for the test code.

azkaban-common/src/test/java/azkaban/executor/container/watch/KubernetesWatchTest.java
hqin2016: addressed the comment

djaiswal83 (approved):

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -50,7 +55,7 @@
  * (2) Driving pod lifecycle actions, such as deleting flow pods in a final state.
  */
 @Singleton
-public class FlowStatusManagerListener implements AzPodStatusListener {
+public class FlowStatusManagerListener extends EventHandler implements AzPodStatusListener {
```
sshardool, line 58: Why does this need to extend the `EventHandler`?
Please see if equivalent functionality can be provided by:
- Passing an EventHandler or EventListener object to this class and retaining that as a member variable
- Using that handler directly whenever we need to emit an event, instead of extending `FlowStatusManagerListener` directly.
 As a general principle composition should be favored over inheritance: https://en.wikipedia.org/wiki/Composition_over_inheritance (unless the class hierarchy arises as a more natural representation of the use case)

sshardool (commented):
Thanks for adding this event, will be very useful. Have a few comments below.

PR #3021, opened 2021-12-13
Add an option to add java options to Java process jobs mentioned in commonprivate.properties

jakhani:
Property defined for this is: azkaban.jobs.java.opts

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -770,6 +770,7 @@ public class JobRunner extends EventHandler implements Runnable {
 
       insertJobMetadata();
       insertJVMAargs();
+
       this.props.put(CommonJobProperties.JOB_ID, this.jobId);
       this.props.put(CommonJobProperties.JOB_ATTEMPT, this.node.getAttempt());
       this.props.put(CommonJobProperties.JOB_METADATA_FILE,
@@ -893,7 +894,7 @@ public class JobRunner extends EventHandler implements Runnable {
     jobJVMArgs += (previousJVMArgs == null) ? "" : " " + previousJVMArgs;
 
     // Add useful Java options for java jobs which are provided through properties
-    String javaOpts = insertJavaOptions();
+    final String javaOpts = insertJavaOptions();
     jobJVMArgs += (javaOpts == null) ? "" : " " + javaOpts;
     this.logger.info("job JVM args: " + jobJVMArgs);
     this.props.put(JavaProcessJob.JVM_PARAMS, jobJVMArgs);
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -770,7 +770,6 @@ public class JobRunner extends EventHandler implements Runnable {
 
       insertJobMetadata();
       insertJVMAargs();
-
       this.props.put(CommonJobProperties.JOB_ID, this.jobId);
       this.props.put(CommonJobProperties.JOB_ATTEMPT, this.node.getAttempt());
       this.props.put(CommonJobProperties.JOB_METADATA_FILE,
@@ -876,6 +875,7 @@ public class JobRunner extends EventHandler implements Runnable {
     return projectPermissionsURL;
   }
 
+
   /**
    * Add useful JVM arguments so it is easier to map a running Java process to a flow, execution id
    * and job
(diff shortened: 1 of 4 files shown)


anish-mal (dismissed):

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -892,10 +892,27 @@ private void insertJVMAargs() {
     final String previousJVMArgs = this.props.get(JavaProcessJob.JVM_PARAMS);
     jobJVMArgs += (previousJVMArgs == null) ? "" : " " + previousJVMArgs;
 
+    // Add useful Java options for java jobs which are provided through properties
+    String javaOpts = insertJavaOptions();
```
sshardool, line 896: Could we repurpose `JavaProcessJob.JVM_PARAMS` to add this instead of creating a new property ?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -892,10 +892,27 @@ private void insertJVMAargs() {
...
+    jobJVMArgs += (javaOpts == null) ? "" : " " + javaOpts;
     this.logger.info("job JVM args: " + jobJVMArgs);
     this.props.put(JavaProcessJob.JVM_PARAMS, jobJVMArgs);
   }
 
+  /**
+   * Add useful Java options for java jobs provided through properties.
+   */
+  private String insertJavaOptions() {
+    if (this.jobtypeManager.getCommonPluginLoadProps() == null) {
```
sshardool, line 906: Which property file is expected to contain the new prop values ?

sshardool (commented):
Thanks for the quick changes. Looks good apart from a few minor comments.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
djaiswal83, line 896: nit : final

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -770,7 +770,6 @@ private boolean prepareJob() throws RuntimeException {
 
       insertJobMetadata();
       insertJVMAargs();
-
```
djaiswal83, line 773: extreme nit : restore this blank line, it maintains readability.

djaiswal83 (changes requested):
Minor comments.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
djaiswal83: It is in commonprivate.properties

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: The reason to keep it separate is we eventually want to remove this fix. That is why keeping the effect of code as minimal as possible.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: Done

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: Done

djaiswal83 (approved):
Thanks for making this change in such short notice!

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -892,10 +893,27 @@ private void insertJVMAargs() {
...
+    final String javaOpts = insertJavaOptions();
+    jobJVMArgs += (javaOpts == null) ? "" : " " + javaOpts;
     this.logger.info("job JVM args: " + jobJVMArgs);
     this.props.put(JavaProcessJob.JVM_PARAMS, jobJVMArgs);
   }
 
+  /**
+   * Add useful Java options for java jobs provided through properties.
+   */
+  private String insertJavaOptions() {
```
amielkelinkedin, line 906: I see no reason not to have the method be final if we are having the calling variable be final, but seeing as we intend to delete this, that's a-ok.

amielkelinkedin (approved):

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
sshardool: Will discuss this offline.

sshardool (approved):

PR #3023, opened 2021-12-16
Unclosed Stream Causes Memory Leak

XinhhD:
Recently I tested azkaban with a very high concurrency.(tag 3.90)
Among the testing, azkaban server was always killed by OOM-Reaper which is a Linux memory protection mechanism.
By debuging with some perf-tools, I finally find it was the reason by unclosed Gzip stream.
After fixing it, server will not be killed by OOM-reaper again.


XinhhD:
It still an unfixed issue in lastest code which can be a serious problem in bussiness enviroment.

XinhhD:
#3026 this issue
@rjurney

ismailsimsek (approved):

ypadron-in (approved):
Good catch! Thanks a lot @XinhhD!

I'm curious about how you debugged this issue. Could you share the tools you used and if possible, perhaps elaborate a bit on how you detected this bugt?

XinhhD:
> Good catch! Thanks a lot @XinhhD!
> 
> I'm curious about how you debugged this issue. Could you share the tools you used and if possible, perhaps elaborate a bit on how you detected this bugt?

First, I built 10,000 flows to test azkaban. These flows will be triggered at the same time ( eg. once half hour).

PR #3025, opened 2021-12-17
Adding UserPermissions and GroupPermissions to the Permission Change Event

ShubhamGupta29:
This change is done to pass on the additional information related to updated UserPermissions and GroupPermissions after a change in permission is made. Since the metadata that is provided to KafkaEvent reporter is of type String so tried to convert the Hashmap's value to a string which can be then split back into String by event reporter later.

**How change is tested:**
Added respective test cases.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
+++ b/azkaban-common/src/main/java/azkaban/project/ProjectManager.java
@@ -35,6 +35,7 @@ import azkaban.user.Permission.Type;
 import azkaban.user.User;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
+import com.google.common.base.Joiner;
 import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -433,6 +434,19 @@ public void updateProjectPermission(final Project project, final String name,
       eventData.put("updatedGroup", "null");
     }
 
+    //Getting updated user and Group permissions as String
+    Map<String, String> updatedUserPermissionMap = new HashMap<>();
```
djaiswal83, line 438: Use final keyword wherever applicable.
Initialize using the size of respective structures.

djaiswal83 (changes requested):
Looks good. Left a minor comment.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
ShubhamGupta29: Ack.
Made the change, kindly have a look

djaiswal83 (approved):
LGTM!

PR #3028, opened 2022-01-06
Add additional checks during AzPodStatus event extraction

hqin2016:
This is a follow-up task for https://github.com/azkaban/azkaban/pull/2781#discussion_r601904885 
In the pull request:
1) all the checkAz* methods are executed 
2) the first determined pod status is returned (refactored code does not alter the outcome of derived pod status)
3) in case more than one of them returns true, warnings are reported

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
@@ -568,33 +568,60 @@ public class AzPodStatusExtractor {
    * @return
    */
   public AzPodStatus createAzPodStatus() {
+    // All pod statuses are checked for any status extracted from the pod event. If multiple
+    // statuses are detected, the first derived pod status will be returned, and the conflict
+    // will be logged
+    int statusCount = 0;
+    AzPodStatus azPodStatus = AzPodStatus.AZ_POD_UNEXPECTED;
     if (checkForAzPodRequested()) {
-      return AzPodStatus.AZ_POD_REQUESTED;
+      azPodStatus = AzPodStatus.AZ_POD_REQUESTED;
+      statusCount++;
     }
-    logInitContainerStatuses();
     if (checkForAzPodScheduled()) {
-      return AzPodStatus.AZ_POD_SCHEDULED;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus, AzPodStatus.AZ_POD_SCHEDULED);
     }
-    logAppContainerStatuses();
+    if (AzPodStatus.AZ_POD_REQUESTED != azPodStatus) {
+      logInitContainerStatuses();
+      if (AzPodStatus.AZ_POD_SCHEDULED != azPodStatus) {
+        logAppContainerStatuses();
+      }
+    }
     if (checkForAzPodInitContainersRunning()) {
-      return AzPodStatus.AZ_POD_INIT_CONTAINERS_RUNNING;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus,
+          AzPodStatus.AZ_POD_INIT_CONTAINERS_RUNNING);
     }
     if (checkForAzPodAppContainerStarting()) {
-      return AzPodStatus.AZ_POD_APP_CONTAINERS_STARTING;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus,
+          AzPodStatus.AZ_POD_APP_CONTAINERS_STARTING);
     }
     if (checkForAzPodReady()) {
-      return AzPodStatus.AZ_POD_READY;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus, AzPodStatus.AZ_POD_READY);
     }
     if (checkForAzPodCompleted()) {
-      return AzPodStatus.AZ_POD_COMPLETED;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus, AzPodStatus.AZ_POD_COMPLETED);
     }
     if (checkForAzPodInitFailure()) {
-      return AzPodStatus.AZ_POD_INIT_FAILURE;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus, AzPodStatus.AZ_POD_INIT_FAILURE);
     }
     if (checkForAzPodAppFailure()) {
-      return AzPodStatus.AZ_POD_APP_FAILURE;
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -568,33 +568,60 @@ private void logAppContainerStatuses() {
...
-      return AzPodStatus.AZ_POD_REQUESTED;
+      azPodStatus = AzPodStatus.AZ_POD_REQUESTED;
+      statusCount++;
     }
-    logInitContainerStatuses();
     if (checkForAzPodScheduled()) {
-      return AzPodStatus.AZ_POD_SCHEDULED;
+      azPodStatus = updateAzPodStatus(statusCount++, azPodStatus, AzPodStatus.AZ_POD_SCHEDULED);
+    }
+    if (AzPodStatus.AZ_POD_REQUESTED != azPodStatus) {
```
sshardool, line 583: Why are these nested if's needed ?

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -568,33 +568,60 @@ private void logAppContainerStatuses() {
...
+  }
+
+  /**
+   * Current status will be returned if no prior status is derived
+   * @param statusCount
+   * @param original
+   * @param current
+   * @return
+   */
+  private AzPodStatus updateAzPodStatus(int statusCount, AzPodStatus original,
```
sshardool, line 622: The change overall should get us close the what's needed for the task. 
However, please do consider alternatives that may be more succinct and readable. 
For example the problem overall translates to -> Given N boolean functions in a sequence, run an input against all N functions. Note the first function that returns true, for all subsequent methods returning true log a warning. 

We can consider utilizing functional aspects of java for this. Happy to chat about it in more detail.

sshardool (commented):
Thanks a lot for taking this up. 
Looks good apart from 1 question and and 1 suggestion for structuring existing code and change differently.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
djaiswal83, line 583: Please add comments for any condition evaluation.

hqin2016: removed nested ifs.

hqin2016: removed unnecessary if checks

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
hqin2016: Thanks for the suggestions. It seems we can use a list to store any statuses that are derived from the pod event, and return the first derived status, and log an error if multiple statuses are derived.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -568,33 +569,42 @@ private void logAppContainerStatuses() {
...
-      return AzPodStatus.AZ_POD_INIT_FAILURE;
+      derivedStatuses.add(AzPodStatus.AZ_POD_INIT_FAILURE);
     }
     if (checkForAzPodAppFailure()) {
-      return AzPodStatus.AZ_POD_APP_FAILURE;
+      derivedStatuses.add(AzPodStatus.AZ_POD_APP_FAILURE);
+    }
+    if (derivedStatuses.size() > 1) {
+      logger.error("%d pod statuses are derived from current pod watch event",
+          derivedStatuses.size());
```
sshardool, line 605: Looks good, apart from these minor points. 
(1) Replace ther error with `log.warn`
(2) Print all the derived statuses as part of the log message, that will help with debugging future issues.

hqin2016: thanks, changed to logger.warn, and logged all statuses extracted from the pod event.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -568,33 +569,43 @@ private void logAppContainerStatuses() {
...
+      derivedStatuses.add(AzPodStatus.AZ_POD_INIT_FAILURE);
     }
     if (checkForAzPodAppFailure()) {
-      return AzPodStatus.AZ_POD_APP_FAILURE;
+      derivedStatuses.add(AzPodStatus.AZ_POD_APP_FAILURE);
+    }
+    if (derivedStatuses.size() > 1) {
+      logger.warn(format("%d pod statuses are derived from current pod watch event: ",
+          derivedStatuses.size()));
+      derivedStatuses.stream().forEach((status) -> logger.warn(status + " "));
```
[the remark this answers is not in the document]
hqin2016: concatenate the statuses to a string and append the string to the warning message.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -568,33 +571,43 @@ private void logAppContainerStatuses() {
...
+      derivedStatuses.add(AzPodStatus.AZ_POD_INIT_FAILURE);
     }
     if (checkForAzPodAppFailure()) {
-      return AzPodStatus.AZ_POD_APP_FAILURE;
+      derivedStatuses.add(AzPodStatus.AZ_POD_APP_FAILURE);
+    }
+    if (derivedStatuses.size() > 1) {
+      logger.warn(format("%d pod statuses are derived from current pod watch event: "
+          + derivedStatuses.stream().map(s -> s.toString()).collect(Collectors.joining(","))),
+          derivedStatuses.size());
```
sshardool, line 608: Recommend rewriting this as:
```
logger.warn(format("%d pod statuses are derived from current pod watch event: %s ",
          derivedStatuses.size(),
          derivedStatuses.stream().map(s -> s.toString()).collect(Collectors.joining(","))));
```
This avoids mixing the 2 styles of composing log message (format string versus string-concat) and makes a little easier to read the statement.

sshardool (dismissed):
Approved with a minor comment.

sshardool (approved):

djaiswal83 (approved):
Thanks for addressing my comments.

PR #3029, opened 2022-01-06
Refactoring UI Pages to reduce Object Load

amielkelinkedin:
Refactoring UI Pages to reduce the amount of information passed to the front end.

Tested locally on SoloServer.

### DIFF
--- a/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/flowexecutionpanel.vm
+++ b/azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/flowexecutionpanel.vm
@@ -215,13 +215,6 @@
             </div>
           #end
 
-          #*
-          #if ($triggerPlugins.size() > 0)
-            #foreach ($triggerPlugin in $triggerPlugins)
-                        <button type="button" class="btn btn-default" id=set-$triggerPlugin.pluginName>$triggerPlugin.pluginName</button>
-            #end
-          #end
-          *#
         <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
         <button type="button" class="btn btn-primary" id="execute-btn">Execute</button>
       </div><!-- /modal-footer -->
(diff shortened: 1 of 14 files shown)


ypadron-in:
I noticed that velocity templates related to the Reportal plugin were left out. Is that intentional? How are we going to handle Reportal?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/jobpage.vm
```diff
@@ -49,14 +49,14 @@
       <div class="row">
         <div class="header-title">
           <h1>
-            <a href="${context}/manager?project=${project.name}&flow=${flowid}&job=${jobid}">
+            <a href="${context}/manager?project=${name}&flow=${flowid}&job=${jobid}">
```
ypadron-in, line 52: Given that this and above are the job and flow pages respectively I wonder is `projectName` instead of `name` is more readable. I also understand that using the same variable names for all pages makes the code more predictable and easy to read, so if a different variable name makes sense here can we use the new one in remaining .vm too?

amielkelinkedin: I think it makes sense to use projectName instead of name. I will change it.

amielkelinkedin:
@ypadron-in In terms of the Reportal, are you talking about the ReportalServlet which also uses Pages in the same way? I was thinking that perhaps it would be good to change them in separate PRs, to reduce the surface area in case there is an error, though I can do them together if you'd prefer?

azkaban-web-server/src/main/resources/azkaban/webapp/servlet/velocity/flowexecutionpanel.vm
```diff
@@ -215,13 +215,6 @@
             </div>
           #end
 
-          #*
-          #if ($triggerPlugins.size() > 0)
-            #foreach ($triggerPlugin in $triggerPlugins)
-                        <button type="button" class="btn btn-default" id=set-$triggerPlugin.pluginName>$triggerPlugin.pluginName</button>
-            #end
-          #end
-          *#
```
ypadron-in, line 224: Thank you for removing unused code!

ypadron-in (approved):
LGTM

jakhani (approved):

PR #3035, opened 2022-01-18
Adding a way to add group and user in case environment doesn't provide it by default. It is controlled by property.

jakhani:
Where this change will help? 
In case of containers running in Kubernetes, we might not have linux users and groups available. In that case, Azkaban jobs will start failing. With the help of this change, it will add user and group for effective user before any job starts.
How to enable group and user addition for any flow execution?
The property introduced to enable group and user addition is: azkaban.add.group.and.user.for.effective.user. Set this property in commonprivate.properties to true and it will enable group and user addition prior to start of any user job process.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -563,7 +563,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
     // Get CPU and memory requested for a flow container
     final String flowContainerCPURequest = getFlowContainerCPURequest(flowParam);
     final String flowContainerMemoryRequest = getFlowContainerMemoryRequest(flowParam);
-
+    logger.info("Creating pod for : " + executionId);
     final AzKubernetesV1SpecBuilder v1SpecBuilder =
         new AzKubernetesV1SpecBuilder(this.clusterEnv, Optional.empty())
             .addFlowContainer(this.flowContainerName,
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -563,7 +563,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
     // Get CPU and memory requested for a flow container
     final String flowContainerCPURequest = getFlowContainerCPURequest(flowParam);
     final String flowContainerMemoryRequest = getFlowContainerMemoryRequest(flowParam);
-    logger.info("Creating pod for : " + executionId);
+    logger.info("Creating pod for execution-id: " + executionId);
     final AzKubernetesV1SpecBuilder v1SpecBuilder =
         new AzKubernetesV1SpecBuilder(this.clusterEnv, Optional.empty())
             .addFlowContainer(this.flowContainerName,
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -15,6 +15,7 @@
  */
 package azkaban.jobExecutor;
 
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_ADD_GROUP_AND_USER_FOR_EFFECTIVE_USER;
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_GROUP_NAME;
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_NATIVE_LIB_FOLDER;
 import static azkaban.ServiceProvider.SERVICE_PROVIDER;
@@ -26,6 +27,7 @@ import azkaban.jobExecutor.utils.process.AzkabanProcess;
 import azkaban.jobExecutor.utils.process.AzkabanProcessBuilder;
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.ExecuteAsUser;
+import azkaban.utils.ExecuteAsUserUtils;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.SystemMemoryInfo;
@@ -248,8 +250,20 @@ public class ProcessJob extends AbstractProcessJob {
             String.format("Not permitted to proxy as '%s' through Azkaban", effectiveUser)
         );
       }
+
+      final boolean addGroupAndUserForEffectiveUser =
+          this.getSysProps().getBoolean(AZKABAN_ADD_GROUP_AND_USER_FOR_EFFECTIVE_USER, false);
+      if (addGroupAndUserForEffectiveUser) {
+        // If linux group and user needs to be added for effectiveUser before job process starts
+        // then set it up before effective user is used to set permission for any file.
+        info("Adding group and user for effective user: " + effectiveUser);
+        final ExecuteAsUser executeAsUser = new ExecuteAsUser(
+            this.getSysProps().getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
+        ExecuteAsUserUtils.addGroupAndUserForEffectiveUser(executeAsUser, effectiveUser);
+      }
+
       // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
-      // if the directory is not permissioned correctly already (should happen once per execution)
+      // if the directory is not permission correctly already (should happen once per execution)
       if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
         info("Changing current working directory ownership");
         assignUserFileOwnership(effectiveUser, getWorkingDirectory());
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -15,7 +15,6 @@
  */
 package azkaban.jobExecutor;
 
-import static azkaban.Constants.ConfigurationKeys.AZKABAN_ADD_GROUP_AND_USER_FOR_EFFECTIVE_USER;
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_GROUP_NAME;
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_SERVER_NATIVE_LIB_FOLDER;
 import static azkaban.ServiceProvider.SERVICE_PROVIDER;
@@ -27,7 +26,6 @@ import azkaban.jobExecutor.utils.process.AzkabanProcess;
 import azkaban.jobExecutor.utils.process.AzkabanProcessBuilder;
 import azkaban.metrics.CommonMetrics;
 import azkaban.utils.ExecuteAsUser;
-import azkaban.utils.ExecuteAsUserUtils;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.SystemMemoryInfo;
@@ -251,19 +249,9 @@ public class ProcessJob extends AbstractProcessJob {
         );
       }
 
-      final boolean addGroupAndUserForEffectiveUser =
-          this.getSysProps().getBoolean(AZKABAN_ADD_GROUP_AND_USER_FOR_EFFECTIVE_USER, false);
-      if (addGroupAndUserForEffectiveUser) {
-        // If linux group and user needs to be added for effectiveUser before job process starts
-        // then set it up before effective user is used to set permission for any file.
-        info("Adding group and user for effective user: " + effectiveUser);
-        final ExecuteAsUser executeAsUser = new ExecuteAsUser(
-            this.getSysProps().getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
-        ExecuteAsUserUtils.addGroupAndUserForEffectiveUser(executeAsUser, effectiveUser);
-      }
-
       // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
-      // if the directory is not permission correctly already (should happen once per execution)
+      // if the directory does not have correct permission already (should happen once per
+      // execution)
       if (!canWriteInCurrentWorkingDirectory(effectiveUser)) {
         info("Changing current working directory ownership");
         assignUserFileOwnership(effectiveUser, getWorkingDirectory());
--- a/azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
@@ -26,32 +26,22 @@ public class ExecuteAsUserUtils {
       final String effectiveUser) throws Exception {
     List<String> commands = Arrays.asList(ADD_GROUP_PATH, effectiveUser);
     int result = executeAsUser.execute("root", commands);
-    if (result == 0) {
+    if (result == 0 || result == 9) {
       logger.info("Group is added successfully.");
       commands = Arrays
           .asList(ADD_USER_PATH, "-l", "-m", effectiveUser, "-g", effectiveUser);
       result = executeAsUser.execute("root", commands);
-      if (result == 0) {
+      if (result == 0 || result == 9) {
         logger.info("User is added successfully.");
       } else {
-        if (result == 9) {
-          logger.info("User already exist for " + effectiveUser);
-        } else {
-          final String errorMessage = "Failed to add user: " + result;
-          logger.error(errorMessage);
-          throw new JobExecutionException(errorMessage);
-        }
-      }
-    } else {
-      //If group already exist then it will return error code 9. Log the info and don't raise any
-      // exception in that case.
-      if (result == 9) {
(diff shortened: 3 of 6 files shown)


ypadron-in:
Could you please elaborate on the property's use case? On what scenarios should be set? How is it set? What user is supposed to set it?

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    int result = executeAsUser.execute("root", commands);
+    if (result != 0) {
+      final String errorMessage = "Failed to add group: " + result;
+      //Log the error message and don't throw exception as this method is called for every job.
+      //If all the jobs have same effective user then first time, adding group and user will pass
+      //but for second time, it will fail. So fail silently.
+      logger.error(errorMessage);
+    }
+    commands =Arrays
+        .asList(ADD_USER_COMMAND, "-l", "-m", effectiveUser, "-g", effectiveUser);
```
ypadron-in, line 35: Does it make sense to execute add user if add group failed? The man page states that the group must exist when using `-g` option.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,42 @@
...
+  public static void addGroupAndUserForEffectiveUser(final ExecuteAsUser executeAsUser,
+      final String effectiveUser) throws Exception {
+    List<String> commands = Arrays.asList(ADD_GROUP_COMMAND, effectiveUser);
+    int result = executeAsUser.execute("root", commands);
+    if (result != 0) {
+      final String errorMessage = "Failed to add group: " + result;
+      //Log the error message and don't throw exception as this method is called for every job.
+      //If all the jobs have same effective user then first time, adding group and user will pass
+      //but for second time, it will fail. So fail silently.
+      logger.error(errorMessage);
```
ypadron-in, line 32: how can we discern between legit errors and errors due to already existing group/user? Is it possible to check existence before trying to add group/user? Here looks like we're logging errors which might not be real errors.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is used as utility class for ExecuteAsUser script.
+ */
+public class ExecuteAsUserUtils {
+
+  private static final Logger logger = LoggerFactory.getLogger(ExecuteAsUserUtils.class);
+  private static final String ADD_GROUP_COMMAND = "/usr/sbin/groupadd";
```
amielkelinkedin, line 14: Should we rename this to GROUP_PATH?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -392,6 +394,20 @@ public void prefetchToken(final File tokenFile, final Props props, final Logger
...
   protected void doPrefetch(final File tokenFile, final Props props, final Logger logger,
       final String userToProxy) throws HadoopSecurityManagerException {
+    final boolean addGroupAndUserForEffectiveUser =
+        props.getBoolean(AZKABAN_ADD_GROUP_AND_USER_FOR_EFFECTIVE_USER, false);
+    if (addGroupAndUserForEffectiveUser) {
+      // If linux group and user needs to be added for effectiveUser before job process starts
+      // then set it up before effective user is used to set permission for any file.
+      logger.info("Adding group and user for effective user: " + userToProxy);
+      try {
+        ExecuteAsUserUtils.addGroupAndUserForEffectiveUser(executeAsUser, userToProxy);
```
ypadron-in, line 404: Isn't the effective user same as user to proxy? Why trying to do add the same user from 2 different places?

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -248,8 +250,20 @@ public void run() throws Exception {
...
+        // then set it up before effective user is used to set permission for any file.
+        info("Adding group and user for effective user: " + effectiveUser);
+        final ExecuteAsUser executeAsUser = new ExecuteAsUser(
+            this.getSysProps().getString(AZKABAN_SERVER_NATIVE_LIB_FOLDER));
+        ExecuteAsUserUtils.addGroupAndUserForEffectiveUser(executeAsUser, effectiveUser);
+      }
+
       // Set parent directory permissions to <uid>:azkaban so user can write in their execution directory
-      // if the directory is not permissioned correctly already (should happen once per execution)
+      // if the directory is not permission correctly already (should happen once per execution)
```
sshardool, line 266: Nit: replace with "if the directory does not have correct permissions already"...

sshardool (commented):
Thanks for making the change.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Good point. Covered it.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Let me try to find out error code in case group and user exist.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Good point. Changed it to ADD_GROUP_PATH and ADD_USER_PATH

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
jakhani: It is same as user to proxy. Moving this logic to JobRunner so that we won't need it at two places to cover various jobtypes.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -392,6 +394,20 @@ public void prefetchToken(final File tokenFile, final Props props, final Logger
    */
   protected void doPrefetch(final File tokenFile, final Props props, final Logger logger,
       final String userToProxy) throws HadoopSecurityManagerException {
+    final boolean addGroupAndUserForEffectiveUser =
```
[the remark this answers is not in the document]
jakhani: Totally agree. Found a place in JobRunner for this. Thanks.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -563,7 +563,7 @@ V1PodSpec createPodSpec(final int executionId, final VersionSet versionSet,
     // Get CPU and memory requested for a flow container
     final String flowContainerCPURequest = getFlowContainerCPURequest(flowParam);
     final String flowContainerMemoryRequest = getFlowContainerMemoryRequest(flowParam);
-
+    logger.info("Creating pod for : " + executionId);
```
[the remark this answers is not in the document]
jakhani: Covered.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
jakhani: Good point. Covered.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Add group and user on environment where job will run before job process starts.
+   *
+   * @param effectiveUser
+   * @throws Exception
+   */
+  public static void addGroupAndUserForEffectiveUser(final ExecuteAsUser executeAsUser,
+      final String effectiveUser) throws Exception {
+    List<String> commands = Arrays.asList(ADD_GROUP_COMMAND, effectiveUser);
+    int result = executeAsUser.execute("root", commands);
```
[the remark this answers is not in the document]
jakhani: Good point. Noting it down in comments. Btw it is the same case for many other commands to give permissions to proxy users to token files and directory etc.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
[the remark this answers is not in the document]
jakhani: Added more checks for the same. Can you please review it again?

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,57 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is used as utility class for ExecuteAsUser script.
+ */
+public class ExecuteAsUserUtils {
+
+  private static final Logger logger = LoggerFactory.getLogger(ExecuteAsUserUtils.class);
+  private static final String ADD_GROUP_PATH = "/usr/sbin/groupadd";
```
amielkelinkedin, line 15: Perhaps this should be configurable as well, but I'm happy with that being in a future PR.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,57 @@
...
+        } else {
+          final String errorMessage = "Failed to add user: " + result;
+          logger.error(errorMessage);
+          throw new JobExecutionException(errorMessage);
+        }
+      }
+    } else {
+      //If group already exist then it will return error code 9. Log the info and don't raise any
+      // exception in that case.
+      if (result == 9) {
```
amielkelinkedin, line 48: Since this is a failure case where nothing else is required, shouldn't it be the first thing we check for at the top of the stack? I don't really think it needs to be changed, just an opinion.

amielkelinkedin (approved):

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Plus one. If we do it, better we do it for all the bin paths like CHOWN and CHMOD. Let's take it later.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Reduced the if-else cases. Can you please take a look at it?

amielkelinkedin: Thank you @jakhani

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -845,6 +856,21 @@ private boolean prepareJob() throws RuntimeException {
     return true;
   }
 
+  // If linux group and user needs to be added for effectiveUser before job process starts
+  // then set it up before effective user is used to set permission for any file.
+  private void addGroupAndUser(final Props pluginPrivateProps) throws Exception {
```
sshardool, line 861: Another concern:
With this mechanism in place it's possible that different instances of jobs will see different user-id for the same user, which isn't the case when using LDAP + nscd. 
If for any reason a job has made assumptions about the UID for a user being same across all invocations (or matching LDAP)  then it could run into trouble. 
I don't think there should be many (if any) jobs making this assumption but still wanted to highlight the possibility.

sshardool (commented):
Thanks for addressing the comments, have a few minor questions.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,47 @@
...
+      final String effectiveUser) throws Exception {
+    List<String> commands = Arrays.asList(ADD_GROUP_PATH, effectiveUser);
+    int result = executeAsUser.execute("root", commands);
+    if (result == 0 || result == 9) {
+      logger.info("Group is added successfully.");
+      commands = Arrays
+          .asList(ADD_USER_PATH, "-l", "-m", effectiveUser, "-g", effectiveUser);
+      result = executeAsUser.execute("root", commands);
+      if (result == 0 || result == 9) {
+        logger.info("User is added successfully.");
```
ypadron-in, line 35: To make logs more useful, is it possible to add the specific group/user that was added in the successful/failure logs?
For example: 
"Group **XYZ** is added added successfully."
Failed to add user: **XYZ**. Return value was: " + result;
"User **XYZ** is added successfully."

ypadron-in: ```"Failed to add group:" +  effectiveUser +  "Return value was: " + result;```

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -694,6 +699,12 @@ private void doRun() {
         writeStatus();
         fireEvent(Event.create(this, EventType.JOB_STATUS_CHANGED,
             new EventData(Status.RUNNING, this.node.getNestedId())));
+        try {
+          addGroupAndUser(this.jobtypeManager.getCommonPluginLoadProps());
+        } catch (Exception e) {
+          finalStatus = changeStatus(Status.FAILED);
+          logError("Job run failed while adding group and user.");
```
ypadron-in, line 706: just curious: why is this not considered and infrastructure failure and therefore the status picked was FAILED instead of EXECUTION_STOPPED.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -694,6 +699,12 @@ private void doRun() {
         writeStatus();
         fireEvent(Event.create(this, EventType.JOB_STATUS_CHANGED,
             new EventData(Status.RUNNING, this.node.getNestedId())));
+        try {
+          addGroupAndUser(this.jobtypeManager.getCommonPluginLoadProps());
```
ypadron-in, line 703: Wondering if this logic classifies under "job preparation" and therefore doing it inside `prepareJob()` would be more appropriate/expected.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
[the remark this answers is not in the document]
jakhani: It's not part of job logs. It is part of flowcontainer log so it appears as a part of flow log.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: Oh yes. I tested couple of flows and did not see any assumption related to same uid. Actually that was the motivation behind adding this as a part of azkaban code rather than finding generic solution.

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
```diff
@@ -0,0 +1,47 @@
...
+  public static void addGroupAndUserForEffectiveUser(final ExecuteAsUser executeAsUser,
+      final String effectiveUser) throws Exception {
+    List<String> commands = Arrays.asList(ADD_GROUP_PATH, effectiveUser);
+    int result = executeAsUser.execute("root", commands);
+    if (result == 0 || result == 9) {
+      logger.info("Group is added successfully.");
+      commands = Arrays
+          .asList(ADD_USER_PATH, "-l", "-m", effectiveUser, "-g", effectiveUser);
+      result = executeAsUser.execute("root", commands);
+      if (result == 0 || result == 9) {
```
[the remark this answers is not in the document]
jakhani: Done

azkaban-common/src/main/java/azkaban/utils/ExecuteAsUserUtils.java
jakhani: Done

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: This is Job status. So that should be marked as Failed. Isn't it?

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
jakhani: prepareJob is more like setting all the necessary properties. That is why I kept it out. Please let me know if you strongly feel about it. I will move it.

sshardool (approved):
Thanks for taking care of the comments!

ypadron-in (approved):
Thanks for addressing all the comments!

PR #3037, opened 2022-01-19
Adding update and delete APIs for imageType image management API

utk-12:
Adding image management APIs for updating and deleting.

**_Query Params for POST Request:_**
session.id = "Insert session id from browser here"
updateImageOwners = "removeImageOwners OR addImageOwners"

**_Body for POST Request (Required Parameters):_**

```
{
"imageType": "utktest",
"ownerships":[{"owner": "jakhani", "role": "ADMIN"},
{"owner": "djaiswal", "role": "ADMIN"}]
}
```

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
@@ -86,4 +86,21 @@ public interface ImageTypeDao {
    * @return ImageType
    */
   public ImageType getImageTypeWithOwnershipsById(String id);
+
+  /**
+   * Add an owner for a given image type.
+   *
+   * @param imageType - ImageType to be updated
+   * @return int - Id representing image type metadata
+   */
+  public int addOwnerOfImageType(ImageType imageType) throws ImageMgmtException;
+
+  /**
+   * Remove an owner for a given image type.
+   *
+   * @param imageType - ImageType to be updated
+   * @return int - Id representing image type metadata
+   */
+  public int removeOwnerOfImageType(ImageType imageType) throws ImageMgmtException;
+
 }
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
@@ -93,7 +93,7 @@ public interface ImageTypeDao {
    * @param imageType - ImageType to be updated
    * @return int - Id representing image type metadata
    */
-  public int addOwnerOfImageType(ImageType imageType) throws ImageMgmtException;
+  public int addImageTypeOwner(ImageType imageType) throws ImageMgmtException;
 
   /**
    * Remove an owner for a given image type.
@@ -101,6 +101,6 @@ public interface ImageTypeDao {
    * @param imageType - ImageType to be updated
    * @return int - Id representing image type metadata
    */
-  public int removeOwnerOfImageType(ImageType imageType) throws ImageMgmtException;
+  public int removeImageTypeOwner(ImageType imageType) throws ImageMgmtException;
 
 }
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
@@ -130,13 +130,14 @@ public class ImageTypeDaoImpl implements ImageTypeDao {
   }
 
   @Override
-  public int addOwnerOfImageType(final ImageType imageType) {
+  public int addImageTypeOwner(final ImageType imageType) {
     final Set<ImageOwnership> currentOwners =
         new HashSet<>(getImageTypeOwnership(imageType.getName()));
     final Set<ImageOwnership> newOwners = new HashSet<>(imageType.getOwnerships());
     newOwners.removeAll(currentOwners);
     //Check if there are actually new owners added.
     if (newOwners.isEmpty()) {
+      log.info("No new owners were added to image type " + imageType.getName());
       throw new ImageMgmtDaoException(ErrorCode.BAD_REQUEST, "Exception while updating image, "
           + "specified no new owners");
     }
@@ -152,13 +153,13 @@ public class ImageTypeDaoImpl implements ImageTypeDao {
       transOperator.getConnection().commit();
       return 1;
     };
+    if (imageTypeId < 1) {
+      log.error(String.format("Exception while updating image type due to invalid "
+          + "imageTypeId: %d.", imageTypeId));
+      throw new ImageMgmtDaoException(ErrorCode.BAD_REQUEST, "Exception while updating image ");
+    }
     try {
       this.databaseOperator.transaction(update);
-      if (imageTypeId < 1) {
-        log.error(String.format("Exception while updating image type due to invalid "
-            + "imageTypeId: %d.", imageTypeId));
-        throw new ImageMgmtDaoException(ErrorCode.BAD_REQUEST, "Exception while updating image ");
-      }
       log.info("Successfully Updated image ownerships for :" + imageType.getName());
     } catch (final SQLException e) {
       log.error("Unable to update the image type metadata", e);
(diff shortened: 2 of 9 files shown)


azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -106,22 +124,82 @@ public int createImageType(final ImageType imageType) {
...
   }
 
+  @Override
+  public int removeOwnerOfImageType(ImageType imageType) throws ImageMgmtException {
+    final Set<ImageOwnership> currentOwners =
+        new HashSet<>(getImageTypeOwnership(imageType.getName()));
+    final Set<ImageOwnership> ownersToRemove = new HashSet<>(imageType.getOwnerships());
+    ownersToRemove.retainAll(currentOwners);
+      //Check if it's removing all owners
+    if (ownersToRemove.size() >= currentOwners.size() - 1){
```
amielkelinkedin, line 177: Is this actually checking if there are two members? It seems like it is doing a size comparison.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -175,29 +175,41 @@ protected void handlePost(
...
+      genericImageType.setCreatedBy(session.getUser().getUserId());
+      if (!HttpRequestUtils.hasParam(req, "updateImageOwners")){
+        if (!hasImageManagementPermission(imageType, session.getUser(), Type.CREATE)) {
+          log.debug(String.format("Invalid permission to create image type for "
+              + "user: %s, image type: %s.", session.getUser().getUserId(), imageType));
+          throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, "Invalid permission to "
+              + "create image type");
+        }
+        // Create image type /Update  and get image type id
+        imageTypeId = this.imageTypeService.createImageType(genericImageType);
```
amielkelinkedin, line 189: Can we change the name of this method from create to update?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
```diff
@@ -175,29 +175,41 @@ protected void handlePost(
...
             + " null. Must provide valid imageType to create/register image type.");
       }
-      if (!hasImageManagementPermission(imageType, session.getUser(), Type.CREATE)) {
-        log.debug(String.format("Invalid permission to create image type for "
-            + "user: %s, image type: %s.", session.getUser().getUserId(), imageType));
-        throw new ImageMgmtInvalidPermissionException(ErrorCode.FORBIDDEN, "Invalid permission to "
-            + "create image type");
+      final Integer imageTypeId;
+      genericImageType.setModifiedBy(session.getUser().getUserId());
+      genericImageType.setCreatedBy(session.getUser().getUserId());
```
amielkelinkedin, line 180: Wouldn't this overwrite the creator in the DB with the updater?

amielkelinkedin:
Thank you for making this change! This is a really interesting approach!

Do we know how we want to structure this change? Here we are assuming a user will want to add and remove users as separate operations rather than just POSTing over an existing ImageOwnership record with a new request. I see pluses and minuses to both approaches. Here if the request is largely the same the user doesn't have to include a lot of the same information, which is quite nice. One the other hand, by simply POSTing over the object we could add and remove at the same time. It would also make updates backwards compatible to our current Jenkins job. If we're planning on moving towards LDAP, we might have smaller requests in the future with fewer users.

I'm fine with either approach!

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -106,22 +124,82 @@ public int createImageType(final ImageType imageType) {
...
+            imageType.getModifiedBy(), currentTimestamp);
       }
-      if (e.getErrorCode() == SQL_ERROR_CODE_DATA_TOO_LONG) {
-        errorMessage = "Reason: Data too long for one or more column(s).";
+      transOperator.getConnection().commit();
+      return 1;
+    };
+    try {
+      this.databaseOperator.transaction(update);
+      if (imageTypeId < 1) {
```
djaiswal83, line 157: Why imagetypeId check matters here? the value is set at line 153. The update does not affect its value.

djaiswal83: The imagetype check for validity must happen before the update SQL is executed.

djaiswal83: The try catch block needs rearrangement by moving the transaction call after the if block and adding its own return value check.

hqin2016:
> Thank you for making this change! This is a really interesting approach!
> 
> Do we know how we want to structure this change? Here we are assuming a user will want to add and remove users as separate operations rather than just POSTing over an existing ImageOwnership record with a new request. I see pluses and minuses to both approaches. Here if the request is largely the same the user doesn't have to include a lot of the same information, which is quite nice. One the other hand, by simply POSTing over the object we could add and remove at the same time. It would also make updates backwards compatible to our current Jenkins job. If we're planning on moving towards LDAP, we might have smaller requests in the future with fewer users.
> 
> I'm fine with either approach!

@utk-12 could you sync up with @amielkelinkedin offline regarding his suggestion?

hqin2016 (commented):
Thanks for taking this story. This is indeed a very useful function.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
```diff
@@ -86,4 +86,21 @@
...
    */
   public ImageType getImageTypeWithOwnershipsById(String id);
+
+  /**
+   * Add an owner for a given image type.
+   *
+   * @param imageType - ImageType to be updated
+   * @return int - Id representing image type metadata
+   */
+  public int addOwnerOfImageType(ImageType imageType) throws ImageMgmtException;
```
hqin2016, line 96: how about using addImageTypeOwner as the function name? the same for the method below.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -106,22 +124,82 @@ public int createImageType(final ImageType imageType) {
...
+  }
+
+  @Override
+  public int addOwnerOfImageType(final ImageType imageType) {
+    final Set<ImageOwnership> currentOwners =
+        new HashSet<>(getImageTypeOwnership(imageType.getName()));
+    final Set<ImageOwnership> newOwners = new HashSet<>(imageType.getOwnerships());
+    newOwners.removeAll(currentOwners);
+    //Check if there are actually new owners added.
+    if (newOwners.isEmpty()) {
```
hqin2016, line 139: can we add a log here?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -106,22 +124,82 @@ public int createImageType(final ImageType imageType) {
...
+      //Check if it's removing all owners
+    if (ownersToRemove.size() >= currentOwners.size() - 1){
+        throw new ImageMgmtDaoException(ErrorCode.BAD_REQUEST, "Exception while deleting "
+            + " image owners, need greater than two owners to be present");
+      }
+      int imageTypeId =
+          getImageTypeByName(imageType.getName()).map(BaseModel::getId).orElse(0);
+
+    try {
+      if (imageTypeId < 1) {
```
hqin2016, line 185: @djaiswal83 's comment above applies here too.

azkaban-common/src/main/java/azkaban/imagemgmt/dto/ImageTypeDTO.java
```diff
@@ -65,6 +65,7 @@ public void setDeployable(final Deployable deployable) {
   }
 
   public void setOwnerships(final List<ImageOwnershipDTO> ownerships) {
-    this.ownerships = ownerships;
+            ownerships.stream().forEach(o -> o.setName(getName()));
```
hqin2016, line 68: Why is this needed?

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
hqin2016, line 180: I thought createdBy and modifiedBy are different people. You might only need to set modifiedBy.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
utk-12: I can do that. I shall change it to get the final owner list size and see if it's greater than 1.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDao.java
utk-12: ack

azkaban-common/src/main/java/azkaban/imagemgmt/dto/ImageTypeDTO.java
utk-12: This is needed becuase the imageOwnershipDTO doesn't get its name field populated. It shows as "null" in the request.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageTypeServlet.java
utk-12: in the case of creating the image_types table maybe, but here either the person a create is happening where both will be the same. or a new entry in the image_ownership table takes place for which both created and modified entry will be the user logged in.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
utk-12: ack. Will change this.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
utk-12: It is re-arranged here right ? I presume this is the correct way ? where I check if the imageType id before the sql update.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -106,22 +124,85 @@ public int createImageType(final ImageType imageType) {
...
 
+  @Override
+  public int removeImageTypeOwner(ImageType imageType) throws ImageMgmtException {
+    final Set<ImageOwnership> currentOwners =
+        new HashSet<>(getImageTypeOwnership(imageType.getName()));
+    final Set<ImageOwnership> ownersToRemove = new HashSet<>(imageType.getOwnerships());
+    ownersToRemove.retainAll(currentOwners);
+    currentOwners.removeAll(ownersToRemove);
+      //Check if it's removing all owners
+    if (!(currentOwners.size() > 1)){
```
hqin2016, line 179: nit: isEmpty() may be better

hqin2016 (approved):

djaiswal83 (changes requested):
Please address the comments

azkaban-common/src/main/java/azkaban/imagemgmt/daos/ImageTypeDaoImpl.java
```diff
@@ -152,13 +153,13 @@ public int addOwnerOfImageType(final ImageType imageType) {
       transOperator.getConnection().commit();
       return 1;
     };
+    if (imageTypeId < 1) {
```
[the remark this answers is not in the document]
utk-12: Addressed

djaiswal83 (approved):

PR #3040, opened 2022-01-25
Add a method to check whether a user belongs to a user group

hqin2016:


djaiswal83 (approved):
Thanks for adding this interface!

azkaban-common/src/main/java/azkaban/user/UserManager.java
```diff
@@ -49,4 +49,12 @@ public User getUser(String username, String password)
...
 
   public boolean validateProxyUser(String proxyUser, User realUser);
+
+  /**
+   * @param username e.g. user alias
+   * @param groupName e.g. name of hadoop headless group / LDAP group
+   * @return Returns true if the user belongs to a group. This is used when verifying user
+   * permission by checking its group membership
+   */
+  public boolean validateUserGroupMembership(String username, String groupName);
```
ypadron-in, line 59: How is this new interface different from existing `validateProxyUser` method? If I understand correctly the `proxyUser` parameter is essentially a group and the method should verify if `realUser` belongs to it.

hqin2016: I think validateProxyUser only checks if the realUser is part of proxyUser or one of realUser's headless group has admin permission. So it is not exactly what this new interface is for.

PR #3041, opened 2022-01-27
Add missing email properties

vhermecz:
Based on: https://github.com/azkaban/azkaban/blob/8b1385ceaafd1e1c2bef6e768cd7a336118bb2c4/azkaban-common/src/main/java/azkaban/utils/EmailMessageCreator.java#L38-L44


docs/configuration.rst
```diff
@@ -232,17 +232,23 @@ Executor Manager Properties
...
-| mail.password | The email password user name.                       |         |
-+---------------+-----------------------------------------------------+---------+
++---------------+-----------------------------------------------------+-----------+
+| Parameter     | Description                                         | Default   |
++===============+=====================================================+===========+
+| mail.sender   | The email address that azkaban uses to send emails. |           |
++---------------+-----------------------------------------------------+-----------+
+| mail.host     | The email server host machine.                      | localhost |
++---------------+-----------------------------------------------------+-----------+
+| mail.host     | The email server port.                              | 25        |
```
ypadron-in, line 242: Could you please fix the typo here? It should be `mail.port` right?

ypadron-in (changes requested):
Thanks for improving the documentation!

docs/configuration.rst
vhermecz: Shoot, my bad, thanks for catching it. Fixed.

ypadron-in (approved):

PR #3042, opened 2022-02-04
Set finalized flow finalFlowStatus so that Azkaban event reporter can report correct flow status

hqin2016:
Verified flow status reported in the flow life cycle event.


djaiswal83 (approved):
LGTM

PR #3043, opened 2022-02-09
ContainerCleanupManager logic should only process containerized executions

ypadron-in:
Skip executions dispatched to bare metal executors when searching for stale containerized executions.


djaiswal83 (approved):
This is a good catch!

jakhani (approved):

PR #3044, opened 2022-02-09
Fix bug in canWriteInCurrentWorkingDirectory, delete TEMP_FILE user_can_write in current working directory if it exists

hqin2016:
There's bug when traversing a special DAG like A->A->B->A, where A and B are different effective users. The CWD ownership would be A->A->B->B instead of A->A->B->A. This is because the TEMP_FILE created in the second job would persist (which A always has write permission to this temp file) and violates the logic to check whether current effective user has write permission in CWD.
Therefore, if TEMP_FILE is created, it should be destroyed at the end of the function.

A test has conducted with a flow DAG of job1(proxy user A)->job2(proxy user A)->job3(proxy user B)->job4(proxy user A).
In job4's log:
08-02-2022 23:27:35 PST job4 INFO - Changing current working directory ownership
08-02-2022 23:27:35 PST job4 INFO - Change ownership of /export/apps/azkaban/azkaban-exec-server/A/executions/1103257 to A:azkaban.
which indicates the fix is working.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
@@ -33,6 +33,8 @@ import azkaban.utils.Utils;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -392,6 +394,17 @@ public class ProcessJob extends AbstractProcessJob {
     final List<String> checkIfUserCanWriteCommand = Arrays
         .asList(CREATE_FILE, getWorkingDirectory() + "/" + TEMP_FILE_NAME);
     final int result = executeAsUser.execute(effectiveUser, checkIfUserCanWriteCommand);
+    // There's bug when traversing a special DAG like A->A->B->A, where A and B are different
+    // effective users. The CWD ownership would be A->A->B->B instead of A->A->B->A.
+    // This is Because the TEMP_FILE created in the second job would persist (which A always has
+    // write permission of this temp file) and violates the logic to check whether current effective
+    // user has write permission in CWD.
+    // Therefore, if TEMP_FILE is created, it should be destroyed at the end of the function
+    try {
+      Files.deleteIfExists(Paths.get(getWorkingDirectory() + "/" + TEMP_FILE_NAME));
+    } catch (Exception e) {
+      info(String.format("Failed to delete %s in current working directory", TEMP_FILE_NAME));
+    }
     return result == SUCCESSFUL_EXECUTION;
   }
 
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -392,6 +394,12 @@ private boolean canWriteInCurrentWorkingDirectory(final String effectiveUser)
     final List<String> checkIfUserCanWriteCommand = Arrays
         .asList(CREATE_FILE, getWorkingDirectory() + "/" + TEMP_FILE_NAME);
     final int result = executeAsUser.execute(effectiveUser, checkIfUserCanWriteCommand);
+    // If TEMP_FILE user_can_write is created, it should be deleted at the end of the function
+    try {
+      Files.deleteIfExists(Paths.get(getWorkingDirectory() + "/" + TEMP_FILE_NAME));
```
xkrogen, line 399: Minor nit: save the filename instead of recomputing it twice:
```
final String tmpFilePath = getWorkingDirectory() + "/" + TEMP_FILE_NAME;
```
You might also consider using the logic within `Paths` to construct the hierarchical path, instead of string manipulation:
```
final Path tmpFilePath = Paths.get(getWorkingDirectory(), TEMP_FILE_NAME);
final List<String> checkIfUserCanWriteCommand = Arrays.asList(CREATE_FILE, tmpFilePath.toString());
...
```

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
```diff
@@ -392,6 +394,12 @@ private boolean canWriteInCurrentWorkingDirectory(final String effectiveUser)
...
+    } catch (Exception e) {
+      info(String.format("Failed to delete %s in current working directory", TEMP_FILE_NAME));
```
xkrogen, line 401: Include `e` in the log? So that we can tell why it failed.

xkrogen (commented):
In the current PR, the user running the Azkaban executor process will be used to delete the temp file. Are we confident that this user will always have the correct permissions to do so? It seems like it will based on the test you described, just want to be sure this will work in all cases.

djaiswal83 (commented):
Thanks @xkrogen  for finding the issue and @hqin2016 for working on it. Will review again once @xkrogen 's comments are addressed.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
hqin2016: @xkrogen thank for the review. I have modified the code as suggested.

azkaban-common/src/main/java/azkaban/jobExecutor/ProcessJob.java
hqin2016: Included e in the log.

hqin2016:
> 

I tried another test, using  a flow DAG of job1(sparktest)->job2(sparktest)->job3(azktest)->job4(sparktest).
In job4's log:
09-02-2022 11:10:45 PST job4 INFO - effective user is: sparktest
09-02-2022 11:10:45 PST job4 INFO - effective user is: sparktest
09-02-2022 11:10:45 PST job4 INFO - Changing current working directory ownership
09-02-2022 11:10:45 PST job4 INFO - Change ownership of /export/apps/azkaban/azkaban-exec-server/haqin/executions/1104686 to sparktest:azkaban.

djaiswal83 (approved):
Looks good to me.

xkrogen:
LGTM thanks @hqin2016 !

PR #3048, opened 2022-02-14
Change groupName to groupSet in UserManager::validateUserGroupMembership function signature

hqin2016:
This allows a more efficient validation of user's group membership to be implemented in plugins


djaiswal83 (approved):
LGTM

PR #3049, opened 2022-02-16
Add disk resource for flow container pod spec.

djaiswal83:
Also added gobblin jobtype to included list.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -670,6 +670,10 @@ public class Constants {
         KUBERNETES_FLOW_CONTAINER_PREFIX + "max.allowed.memory";
     public static final String KUBERNETES_FLOW_CONTAINER_MEMORY_REQUEST =
         KUBERNETES_FLOW_CONTAINER_PREFIX + "memory.request";
+    public static final String KUBERNETES_FLOW_CONTAINER_DISK_REQUEST =
+        KUBERNETES_FLOW_CONTAINER_PREFIX + "disk.request";
+    public static final String KUBERNETES_FLOW_CONTAINER_MAX_ALLOWED_DISK =
+        KUBERNETES_FLOW_CONTAINER_PREFIX + "max.allowed.disk";
     public static final String KUBERNETES_FLOW_CONTAINER_SECRET_NAME =
         KUBERNETES_FLOW_CONTAINER_PREFIX + "secret.name";
     public static final String KUBERNETES_FLOW_CONTAINER_SECRET_VOLUME =
@@ -754,6 +758,10 @@ public class Constants {
     // Constant for memory request for flow container
     public static final String FLOW_PARAM_FLOW_CONTAINER_MEMORY_REQUEST =
         "flow.container.memory.request";
+
+    // Constant for disk request for flow container
+    public static final String FLOW_PARAM_FLOW_CONTAINER_DISK_REQUEST =
+        "flow.container.disk.request";
 
     public static final String FLOW_PARAM_POD_ENV_VAR = "pod.env.var.";
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -688,6 +688,9 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
         flowParam.get(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_DISK_REQUEST);
     int resourceCompare = compareResources(this.maxAllowedDisk, userDiskRequest);
     if (resourceCompare < 0) {
+      // User request disk capacity exceeds max allowed disk
+      userDiskRequest = this.maxAllowedDisk;
+    } else if (resourceCompare == 0) {
       // If comparison fails for any reason, use the value from the config.
       userDiskRequest = this.diskRequest;
     }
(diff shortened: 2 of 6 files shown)


hqin2016 (dismissed):
LGTM

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -670,6 +670,10 @@
         KUBERNETES_FLOW_CONTAINER_PREFIX + "max.allowed.memory";
     public static final String KUBERNETES_FLOW_CONTAINER_MEMORY_REQUEST =
         KUBERNETES_FLOW_CONTAINER_PREFIX + "memory.request";
+    public static final String KUBERNETES_FLOW_CONTAINER_DISK_REQUEST =
+        KUBERNETES_FLOW_CONTAINER_PREFIX + "disk.request";
+    public static final String KUBERNETES_FLOW_CONTAINER_MAX_ALLOWED_DISK =
+        KUBERNETES_FLOW_CONTAINER_PREFIX + "max.allowed.disk";
```
ypadron-in, line 676: Wondering if we should refer to ephemeral storage specifically in the name of the property to avoid any confusion.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -563,13 +574,14 @@ V1PodSpec createPodSpec(final int executionId, final VersionSet versionSet,
...
     final String flowContainerMemoryRequest = getFlowContainerMemoryRequest(flowParam);
+    final String flowContainerDiskRequest = getFlowContainerDiskRequest(flowParam);
     logger.info("Creating pod for execution-id: " + executionId);
     final AzKubernetesV1SpecBuilder v1SpecBuilder =
         new AzKubernetesV1SpecBuilder(this.clusterEnv, Optional.empty())
             .addFlowContainer(this.flowContainerName,
                 azkabanBaseImageFullPath, ImagePullPolicy.IF_NOT_PRESENT, azkabanConfigVersion)
             .withResources(this.cpuLimit, flowContainerCPURequest, this.memoryLimit,
-                flowContainerMemoryRequest);
+                flowContainerMemoryRequest, flowContainerDiskRequest);
```
ypadron-in, line 584: curious about why there is "limit" and "request" values for CPU and memory but not for ephemeral storage.

az-core/src/main/java/azkaban/Constants.java
svnarumugam: I had the same question about the `flow.container.disk.request` param. I believe users know that this disk is a temporary space.

jakhani (commented):
Can you please add unit test for disk request the way there are unit tests for CPU and memory?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -657,6 +669,31 @@ String getFlowContainerMemoryRequest(final Map<String, String> flowParam) {
...
+  String getFlowContainerDiskRequest(final Map<String, String> flowParam) {
+    if (flowParam == null || flowParam.isEmpty() || !flowParam
+        .containsKey(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_DISK_REQUEST)) {
+      return this.diskRequest;
+    }
+
+    String userDiskRequest =
+        flowParam.get(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_DISK_REQUEST);
+    int resourceCompare = compareResources(this.maxAllowedDisk, userDiskRequest);
+    if (resourceCompare < 0) {
```
ypadron-in, line 690: Based on the documentation of `compareResources` method it looks like zero will be return if there is an issue with the comparison. Shouldn't we use `<= 0` in that case?
Also if the requested storage is bigger that max allowed shouldn't we default to max instead of min to user match request as much as possible?

az-core/src/main/java/azkaban/Constants.java
djaiswal83: From flow container's  perspective that is the disk it has.

djaiswal83: Besides, its a short name. let me know if you want me to change it.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
djaiswal83: These are legacy params which will be removed later on. We do not want to have them separate that is why the multiplier is now 1. With storage we are not having any multiplier at all.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
djaiswal83: Let me fix the logic. Good catch.

ypadron-in (approved):
Thanks for addressing the comments! 
There is one comment about adding unit tests which I'm not sure is going to be addressed in this PR but otherwise changes look good to me.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
ypadron-in: got it, thanks!

az-core/src/main/java/azkaban/Constants.java
ypadron-in: let's keep current name then, thanks for clarifying!

PR #3053, opened 2022-02-23
Adding /admin GET method which will return GOOD to denote the health of Azkaban web server

jakhani:
/admin call will return GOOD. It can be used to integrate with health and monitoring system for web server.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/AdminServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/AdminServlet.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


svnarumugam (commented):
the changes look good. I am wondering
1. why don't we use `/status` itself to solve the same problem?
2. can we consider returning a JSON response? The clients who are using can parse this easily and if we choose to add more fields in the response it will be convenient.

jakhani:
> 

@svnarumugam , we have been using /status but it is not just for health check. It queries lot of things from database and sometimes slow when it comes to pinging /status for every hit coming to Azkaban web server. We need light API.
Reason why we are not returning json is for internal LinkedIn integration for any services' status, they expect response this way (plain GOOD).

amielkelinkedin:
Can we change the name of this endpoint? /health? /monitor? /ping? /heartbeat?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/AdminServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      resp.setContentType(JSON_MIME_TYPE);
+      resp.getOutputStream().println("GOOD");
+      resp.setStatus(HttpServletResponse.SC_OK);
+    } catch (final Exception e) {
+      logger.error("Error!! while reporting status: ", e);
+      resp.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage());
+    } finally {
+      resp.getOutputStream().close();
+    }
```
ypadron-in, line 58: Can we use the `sendResponse` method defined in the `AbstractAzkabanServlet` class?

ypadron-in:
Could you please an example response in the PR description?

PR #3055, opened 2022-02-24
Enable periodic bulk cleaning up stale containers in current Azkaban cluster and namespace via ContainerCleanupManager

hqin2016:
Verified in test cluster:
```
2022/02/24 07:05:10.161 +0000  INFO [KubernetesContainerizedImpl] [azk-container-cleanup] [Azkaban] Cleaning up containers older than 10 days
2022/02/24 07:05:10.226 +0000  INFO [KubernetesContainerizedImpl] [azk-container-cleanup] [Azkaban] Deleting stale pod fc-dep-xxxxxxx-439102
2022/02/24 07:05:10.226 +0000  INFO [KubernetesContainerizedImpl] [azk-container-cleanup] [Azkaban] Getting Pod Events for execution Id 439102
2022/02/24 07:05:11.439 +0000  INFO [KubernetesContainerizedImpl] [azk-container-cleanup] [Azkaban] ExecId: 439102, Action: Pod Deletion, Pod Name: fc-dep-xxxxxxx-439102
2022/02/24 07:05:11.440 +0000  INFO [AzPodStatusDrivingListener] [Thread-16] [Azkaban] WatchEvent for pod fc-dep-xxxxxxx-439102 : AZ_POD_READY
2022/02/24 07:05:11.440 +0000  INFO [ContainerStatusMetricsListener] [azk-container-metrics-pool-0] [Azkaban] Event pod status is same as current AZ_POD_READY, for pod fc-dep-xxxxxxx-439102. Any updates will be skipped.
2022/02/24 07:05:11.453 +0000  INFO [KubernetesContainerizedImpl] [azk-container-cleanup] [Azkaban] ExecId: 439102, Action: Service Deletion, Service Name: fc-svc-xxxxxxx-439102, code: null, message: null
```

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -204,7 +204,7 @@ public class ContainerCleanupManager {
       logger.info("Restarting cleaned up flow " + flow.getExecutionId());
       ExecutionControllerUtils.restartFlow(flow, originalStatus);
     } catch (RuntimeException re) {
-      logger.error("Unexpected RuntimeException while restarting flow during clean up." + re);
+      logger.error("Unexpected Exception while restarting flow during clean up." + re);
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -83,8 +83,8 @@ public class ContainerCleanupManager {
             ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_EXECUTION_CLEANUP_INTERVAL_MIN,
             DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL.toMinutes());
     this.containerCleanupIntervalMin = azkProps.getLong(
-        ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_CONTAINER_CLEANUP_INTERVAL_MIN
-        , DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
+        ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_CONTAINER_CLEANUP_INTERVAL_MIN,
+        DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
     this.cleanupService = Executors.newSingleThreadScheduledExecutor(
         new ThreadFactoryBuilder().setNameFormat("azk-container-cleanup").build());
     this.executorLoader = executorLoader;
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -127,7 +127,7 @@ public class ContainerCleanupManager {
   public void cleanUpStaleContainers() {
     try {
       this.containerizedImpl.deleteAgedContainers(validityMap.get(Status.RUNNING).getFirst());
-    } catch (Exception e) {
+    } catch (final Exception e) {
       logger.error("Exception occurred while cleaning up stale pods and services." + e);
     }
   }
@@ -194,7 +194,7 @@ public class ContainerCleanupManager {
           "Cleaning up stale flow " + flow.getExecutionId() + " in state " + originalStatus
               .name());
       this.containerizedDispatchManager.cancelFlow(flow, flow.getSubmitUser());
-    } catch (Exception e) {
+    } catch (final Exception e) {
       logger.error("Unexpected Exception while canceling and finalizing flow during clean up." + e);
     }
   }
@@ -203,8 +203,8 @@ public class ContainerCleanupManager {
     try {
       logger.info("Restarting cleaned up flow " + flow.getExecutionId());
       ExecutionControllerUtils.restartFlow(flow, originalStatus);
-    } catch (RuntimeException re) {
-      logger.error("Unexpected Exception while restarting flow during clean up." + re);
+    } catch (final Exception e) {
+      logger.error("Unexpected Exception while restarting flow during clean up." + e);
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -126,9 +126,11 @@ public class ContainerCleanupManager {
    */
   public void cleanUpStaleContainers() {
     try {
-      this.containerizedImpl.deleteContainers(validityMap.get(Status.RUNNING).getFirst().toMillis());
-    } catch (ExecutorManagerException e) {
-      logger.error("Exception occurred while deleting stale pods and services." + e);
+      this.containerizedImpl.deleteAgedContainers(validityMap.get(Status.RUNNING).getFirst());
+    } catch (ExecutorManagerException eme) {
+      logger.error("ExecutorManagerException occurred while deleting stale pods and services." + eme);
+    } catch (RuntimeException re) {
+      logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -217,7 +217,7 @@ public class ContainerCleanupManager {
     try {
       this.containerizedImpl.deleteContainer(executionId);
     } catch (final Exception e) {
-      logger.error("Unexpected RuntimeException while deleting container.", e);
+      logger.error("Unexpected exception while deleting container.", e);
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -127,10 +127,8 @@ public class ContainerCleanupManager {
   public void cleanUpStaleContainers() {
     try {
       this.containerizedImpl.deleteAgedContainers(validityMap.get(Status.RUNNING).getFirst());
-    } catch (ExecutorManagerException eme) {
-      logger.error("ExecutorManagerException occurred while deleting stale pods and services." + eme);
-    } catch (RuntimeException re) {
-      logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
+    } catch (Exception e) {
+      logger.error("Exception occurred while cleaning up stale pods and services." + e);
     }
   }
 
@@ -196,10 +194,8 @@ public class ContainerCleanupManager {
           "Cleaning up stale flow " + flow.getExecutionId() + " in state " + originalStatus
               .name());
       this.containerizedDispatchManager.cancelFlow(flow, flow.getSubmitUser());
-    } catch (ExecutorManagerException eme) {
-      logger.error("ExecutorManagerException while cancelling flow.", eme);
-    } catch (RuntimeException re) {
-      logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
+    } catch (Exception e) {
+      logger.error("Unexpected Exception while canceling and finalizing flow during clean up." + e);
     }
   }
 
@@ -220,10 +216,8 @@ public class ContainerCleanupManager {
   private void deleteContainerQuietly(final int executionId) {
     try {
       this.containerizedImpl.deleteContainer(executionId);
-    } catch (final ExecutorManagerException eme) {
-      logger.error("ExecutorManagerException while deleting container.", eme);
-    } catch (final RuntimeException re) {
-      logger.error("Unexpected RuntimeException while deleting container.", re);
+    } catch (final Exception e) {
+      logger.error("Unexpected RuntimeException while deleting container.", e);
     }
   }
 
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -54,8 +54,10 @@ public class ContainerCleanupManager {
 
   public static final int DEFAULT_AZKABAN_MAX_FLOW_RUNNING_MINS = -1;
   private static final Logger logger = LoggerFactory.getLogger(ContainerCleanupManager.class);
-  private static final Duration DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL = Duration.ofMinutes(10);
-  private final long cleanupIntervalMin;
+  private static final Duration DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL = Duration.ofMinutes(10);
+  private static final Duration DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL = Duration.ofMinutes(60);
+  private final long executionCleanupIntervalMin;
+  private final long containerCleanupIntervalMin;
   private final ScheduledExecutorService cleanupService;
   private final ExecutorLoader executorLoader;
   private final ContainerizedImpl containerizedImpl;
@@ -76,10 +78,13 @@ public class ContainerCleanupManager {
   public ContainerCleanupManager(final Props azkProps, final ExecutorLoader executorLoader,
       final ContainerizedImpl containerizedImpl,
       final ContainerizedDispatchManager containerizedDispatchManager) {
-    this.cleanupIntervalMin = azkProps
+    this.executionCleanupIntervalMin = azkProps
         .getLong(
             ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_EXECUTION_CLEANUP_INTERVAL_MIN,
-            DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
+            DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL.toMinutes());
+    this.containerCleanupIntervalMin = azkProps.getLong(
+        ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_CONTAINER_CLEANUP_INTERVAL_MIN
+        , DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
     this.cleanupService = Executors.newSingleThreadScheduledExecutor(
         new ThreadFactoryBuilder().setNameFormat("azk-container-cleanup").build());
     this.executorLoader = executorLoader;
@@ -114,6 +119,17 @@ public class ContainerCleanupManager {
   public void cleanUpStaleFlows() {
     this.validityMap.entrySet().stream().filter(e -> !e.getValue().getFirst().isNegative()).map(
         Entry::getKey).forEach(this::cleanUpStaleFlows);
+  }
+
+  /**
+   * Try to clean the stale containers that are older than maximum azkaban flow running time
+   */
+  public void cleanUpStaleContainers() {
+    try {
+      this.containerizedImpl.deleteContainers(validityMap.get(Status.RUNNING).getFirst().toMillis());
+    } catch (ExecutorManagerException e) {
+      logger.error("Exception occurred while deleting stale pods and services." + e);
+    }
   }
 
   /**
@@ -215,8 +231,11 @@ public class ContainerCleanupManager {
   @SuppressWarnings("FutureReturnValueIgnored")
   public void start() {
     logger.info("Start container cleanup service");
+    // Default execution clean up interval is 10 min, container clean up interval is 60 min
     this.cleanupService.scheduleAtFixedRate(this::cleanUpStaleFlows, 0L,
-        this.cleanupIntervalMin, TimeUnit.MINUTES);
+        this.executionCleanupIntervalMin, TimeUnit.MINUTES);
+    this.cleanupService.scheduleAtFixedRate(this::cleanUpStaleContainers, 0L,
+        this.containerCleanupIntervalMin, TimeUnit.MINUTES);
   }
 
   /**
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
@@ -20,4 +20,5 @@ import azkaban.executor.ExecutorManagerException;
 public interface ContainerizedImpl {
   void createContainer(final int executionId) throws ExecutorManagerException;
   void deleteContainer(final int executionId) throws ExecutorManagerException;
+  void deleteContainers(final long containerValidity) throws ExecutorManagerException;
 }
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
@@ -16,9 +16,10 @@
 package azkaban.executor.container;
 
 import azkaban.executor.ExecutorManagerException;
+import java.time.Duration;
 
 public interface ContainerizedImpl {
   void createContainer(final int executionId) throws ExecutorManagerException;
   void deleteContainer(final int executionId) throws ExecutorManagerException;
-  void deleteContainers(final long containerValidity) throws ExecutorManagerException;
+  void deleteAgedContainers(final Duration containerValidity) throws ExecutorManagerException;
 }
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -414,15 +414,15 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
 
   /**
    * This method is used to delete invalid containers in batch
-   * @param containerValidity pod validity in milliseconds
+   * @param containerValidity container valid duration
    * @throws ExecutorManagerException
    */
   @Override
-  public void deleteAgedContainers(Duration containerValidity) throws ExecutorManagerException {
+  public void deleteAgedContainers(final Duration containerValidity) throws ExecutorManagerException {
     logger.info(String.format("Cleaning up containers older than %d days",
         containerValidity.toDays()));
     final Set<Integer> containersToDelete = getStaleContainers(containerValidity);
-    for (int execId : containersToDelete) {
+    for (final int execId : containersToDelete) {
       if (execId > 0) {
         logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
             getServiceName(execId)));
@@ -438,20 +438,20 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
 
   /**
    * This method is used to fetch all stale pods in the current az cluster and namespace
-   * @param containerValidity container validity in milliseconds
+   * @param containerValidity container valid duration
    * @return Set of stale container's execution ids
    */
   private Set<Integer> getStaleContainers(final Duration containerValidity) throws ExecutorManagerException {
     try {
       // Select pods only from current Azkaban cluster and namespace
       final String label = CLUSTER_LABEL_NAME + "=" + this.clusterName + "," + LABEL_POSTFIX;
-      V1PodList items= this.coreV1Api.listNamespacedPod(this.namespace, null,
+      final V1PodList items= this.coreV1Api.listNamespacedPod(this.namespace, null,
           null, null, null, label,
           null, null, null, null);
 
       // Get all execution ids of the pods whose age is older than Azkaban max flow running time
       // (e.g. 10 days)
-      final Long validStartTimeStamp = System.currentTimeMillis() - containerValidity.toMillis();
+      final long validStartTimeStamp = System.currentTimeMillis() - containerValidity.toMillis();
       return getExecutionIdFromPodList(items, validStartTimeStamp);
     } catch (ApiException ae) {
       logger.error(String.format("Unable to fetch stale pods in %s.", this.clusterName),
@@ -468,8 +468,8 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
    */
   @VisibleForTesting
   Set<Integer> getExecutionIdFromPodList (final V1PodList podList, final long validStartTimeStamp) {
-    Set<Integer> staleContainerExecIdSet = new HashSet<>();
-    for (V1Pod pod: podList.getItems()) {
+    final Set<Integer> staleContainerExecIdSet = new HashSet<>();
+    for (final V1Pod pod: podList.getItems()) {
       if (pod.getMetadata().getCreationTimestamp().getMillis() < validStartTimeStamp) {
         final String execIdLabel =
             pod.getMetadata().getLabels().getOrDefault(EXECUTION_ID_LABEL_NAME,
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -422,15 +422,13 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
         containerValidity.toDays()));
     final Set<Integer> containersToDelete = getStaleContainers(containerValidity);
     for (final int execId : containersToDelete) {
-      if (execId > 0) {
-        logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
-            getServiceName(execId)));
-        try {
-          deleteContainer(execId);
-        } catch (final ExecutorManagerException e) {
-          logger.error(String.format("Unable to delete stale pod and service of exec-id %d",
-              execId), e.getMessage());
-        }
+      logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
+          getServiceName(execId)));
+      try {
+        deleteContainer(execId);
+      } catch (final ExecutorManagerException e) {
+        logger.error(String.format("Unable to delete stale pod and service of exec-id %d",
+            execId), e.getMessage());
       }
     }
   }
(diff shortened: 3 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+  @Override
+  public void deleteContainers(long containerValidity) throws ExecutorManagerException {
+    logger.info(String.format("Cleaning up containers older than %d days",
+        containerValidity / DAY_TO_MILLISECOND_CONVERSION_FACTOR));
+    final List<String> podsToDelete = getStalePods(containerValidity);
+    for (String podName : podsToDelete) {
+      logger.info(String.format("Deleting stale pod %s", podName));
+      final int execId = getExecutionId(podName);
+      if (execId > 0) {
+        deleteContainer(execId);
```
sshardool, line 430: Seems like every webserver will be deleting all the stale pods including those for other Azkaban clusters if they happen to be sharing the same kubernetes namespace. 
Is this intentional?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+
+  /**
+   * This method is used to extract execution id from pod mame
+   * @param podName
+   * @return execution id
+   */
+  private int getExecutionId(final String podName) {
+    int execId = -1;
+    try {
+      execId = Integer.valueOf(podName.replace(this.podPrefix + "-", "")
```
sshardool, line 469: Flow container pods are currently launched with the following label associated with them, e.g
`app=azkaban-exec-server,cluster=lasso,execution-id=execid-8606092,service=flow-lasso-8606092`
Checking these labels can serve as a good mechanism for:

1. Ensuring we are indeed deleting a flow-container pod and not a pod whose name accidentally happens to match the naming scheme `fc-xxx-yyy` 
2. Extracting the execution-id more directly as part of the label `execution-id`. 
3. Validating that the pod is for the current webserver with the label `cluster`. 

I would recommend matching based on the labels, let me know your thoughts on this.

sshardool (commented):
Thanks a lot for addressing this!

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
```diff
@@ -20,4 +20,5 @@
 public interface ContainerizedImpl {
   void createContainer(final int executionId) throws ExecutorManagerException;
   void deleteContainer(final int executionId) throws ExecutorManagerException;
+  void deleteContainers(final long containerValidity) throws ExecutorManagerException;
```
svnarumugam, line 23: 1. would it be helpful if the signature is 
deleteContainers(10, TimeUnit.Days)? 
the method can then do `timeUnit.toMillis(duration)`
2. would it make sense if we add the aged deletion part to the method name? for example, `deleteAgedContainers`?

amielkelinkedin:
Thanks for making this change, @hqin2016! Is there a better way to do this? For instance, implementing some sort of expiration logic through the liveness probe in kubernetes?  It seems to me like we should keep as much kuberentes logic in that domain as possible. Thoughts, @sshardool?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Thank you for reviewing. In ln 443, the label indicates that only pods belong to current az cluster is queried (final String label = "cluster=" + this.clusterName;).

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -116,6 +121,17 @@ public void cleanUpStaleFlows() {
         Entry::getKey).forEach(this::cleanUpStaleFlows);
   }
 
+  /**
+   * Try to clean the stale containers that are older than maximum azkaban flow running time
+   */
+  public void cleanUpStaleContainers() {
```
sshardool, line 127: Note that any unchecked exceptions thrown by this method will potentially block subsequent executions of this method through in the cleanupService (quirk of the ScheduledExecutorService). 
Please catch and log any RuntimeException (do not rethrow). Look at `deleteContainerQuietly` for reference.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: 1  We are querying the pods under this web server and namespace, which should prevent irrelevant pods to be deleted
2. We can fetch execution-id string from label map, and extract the execid out of it. 
3. explained in 1

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
svnarumugam: @sshardool The scheduled executor instance we are using is a single-threaded one. we are adding two clean-up tasks in that (current one and earlier clean-up task). I believe it's okay given that our calls are not that heavy. Thinking out loud here.

hqin2016: Thanks for pointing it out. I added RuntimeException catch block.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedImpl.java
hqin2016: Thanks for reviewing. 
1. Changed to Duration from long
2. Renamed the method to deleteAgedContainers
3. This is much more convenient.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
     }
   }
 
+  /**
+   * This method is used to delete invalid containers in batch
+   * @param containerValidity pod validity in milliseconds
+   * @throws ExecutorManagerException
+   */
+  @Override
+  public void deleteContainers(long containerValidity) throws ExecutorManagerException {
```
ypadron-in: Wondering if we should send an email to Azkaban admins about the list of containers that could not be deleted?

ypadron-in (commented):
Thanks for working on this.

azkaban-common/src/test/java/azkaban/executor/container/ContainerCleanupManagerTest.java
```diff
@@ -46,12 +47,13 @@
   private ContainerizedImpl containerImpl;
   private ContainerizedDispatchManager containerizedDispatchManager;
   private ContainerCleanupManager cleaner;
+  private static long MAX_FLOW_RUNNING_MIN = 14400;
```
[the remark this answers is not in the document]
hqin2016: Changed to what is suggested.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+      // Get all pods whose age is older than Azkaban max flow running time (e.g. 10 days)
+      final Long validStartTimeStamp = System.currentTimeMillis() - containerValidity;
+      stalePodList =
+          items.getItems().stream()
+              .filter((pod) -> pod.getMetadata().getCreationTimestamp().getMillis() < validStartTimeStamp)
+              .map((pod) -> pod.getMetadata().getName()).collect(Collectors.toList());
+    } catch (ApiException e) {
+      logger.error(String.format("Unable to fetch stale pods in %s.", this.clusterName),
+          e.getResponseBody());
+    }
```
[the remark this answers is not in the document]
hqin2016: Added throwing an ExecutorManagerException.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -116,6 +121,19 @@ public void cleanUpStaleFlows() {
...
+    try {
+      this.containerizedImpl.deleteAgedContainers(validityMap.get(Status.RUNNING).getFirst());
+    } catch (ExecutorManagerException eme) {
+      logger.error("ExecutorManagerException occurred while deleting stale pods and services." + eme);
+    } catch (RuntimeException re) {
+      logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
+    }
```
ypadron-in, line 134: Do we need to treat these exceptions differently? Why not
`catch (ExecutorManagerException | RuntimeException e ) {..}`
or just 
`catch (Exception e) {..}`

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+
+  /**
+   * This method is used to fetch all stale pods in the current az cluster and namespace
+   * @param containerValidity pod validity in milliseconds
+   * @return List of stale pods
+   */
+  private List<String> getStalePods(final long containerValidity) {
+    List<String> stalePodList = new ArrayList<>();
+    try {
+      final String label = "cluster=" + this.clusterName;
```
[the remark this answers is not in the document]
hqin2016: Changed to use a constant to replace hard-coded "cluster=". 
The reason I'm using clusterName is that I hope to select pods that only belong to this az cluster.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,12 +402,78 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
    */
   @Override
   public void deleteContainer(final int executionId) throws ExecutorManagerException {
-    deletePod(executionId);
+    try { // if pod deletion is not successful, the service deletion can still be handled
+      deletePod(executionId);
+    } catch (ExecutorManagerException e) {
+      logger.error("ExecId: {}, Unable to delete Pod in Kubernetes: {}", executionId,
+          e.getMessage());
+    }
```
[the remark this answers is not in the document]
hqin2016: Removed catch block, moved deleteService to finally block.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: I'll address this in a separate PR. Right now logging the error should be sufficient.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: This makes sense. Made changes to catch general Exception.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -76,10 +78,13 @@
...
       final ContainerizedDispatchManager containerizedDispatchManager) {
-    this.cleanupIntervalMin = azkProps
+    this.executionCleanupIntervalMin = azkProps
         .getLong(
             ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_EXECUTION_CLEANUP_INTERVAL_MIN,
-            DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
+            DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL.toMinutes());
+    this.containerCleanupIntervalMin = azkProps.getLong(
+        ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_CONTAINER_CLEANUP_INTERVAL_MIN
+        , DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
```
djaiswal83, line 87: extreme nit: keep the comma in line above.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -178,10 +194,8 @@ private void cancelFlowQuietly(ExecutableFlow flow, Status originalStatus) {
           "Cleaning up stale flow " + flow.getExecutionId() + " in state " + originalStatus
               .name());
       this.containerizedDispatchManager.cancelFlow(flow, flow.getSubmitUser());
-    } catch (ExecutorManagerException eme) {
-      logger.error("ExecutorManagerException while cancelling flow.", eme);
-    } catch (RuntimeException re) {
-      logger.error("Unexpected RuntimeException while finalizing flow during clean up." + re);
+    } catch (Exception e) {
```
djaiswal83, line 197: If the idea is to catch everything, catch "Throwable" instead of Exception.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -202,10 +216,8 @@ private void retryFlowQuietly(ExecutableFlow flow, Status originalStatus) {
   private void deleteContainerQuietly(final int executionId) {
     try {
       this.containerizedImpl.deleteContainer(executionId);
-    } catch (final ExecutorManagerException eme) {
-      logger.error("ExecutorManagerException while deleting container.", eme);
-    } catch (final RuntimeException re) {
-      logger.error("Unexpected RuntimeException while deleting container.", re);
+    } catch (final Exception e) {
```
djaiswal83, line 219: catch Throwable.

djaiswal83 (changes requested):
I have added few comments. Please address those.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +404,70 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+
+      // Get all execution ids of the pods whose age is older than Azkaban max flow running time
+      // (e.g. 10 days)
+      final Long validStartTimeStamp = System.currentTimeMillis() - containerValidity.toMillis();
+      staleContainerExecIdSet =
+          items.getItems().stream()
+              .filter((pod) -> pod.getMetadata().getCreationTimestamp().getMillis() < validStartTimeStamp)
+              .map((pod) -> pod.getMetadata().getLabels().getOrDefault(EXECUTION_ID_LABEL_NAME,
+                  EXECUTION_ID_LABEL_PREFIX + DEFAULT_EXECUTION_ID)
+                  .replace(EXECUTION_ID_LABEL_PREFIX, ""))
```
djaiswal83, line 456: Please use subString instead of replace method.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: Made the change

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: It seems Exception should be sufficient. The cleanup action is schedule in a frequent manner. A failure does not affect the performance.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: It seems Exception should be sufficient. The cleanup action is schedule in a frequent manner. A failure does not affect the performance.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Used substring instead of replace.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -130,7 +129,7 @@
   public static final String DEFAULT_AZKABAN_BASE_IMAGE_NAME = "azkaban-base";
   public static final String DEFAULT_AZKABAN_CONFIG_IMAGE_NAME = "azkaban-config";
   private static final int DEFAULT_EXECUTION_ID = -1;
-  private static final String EQUAL_SIGN = "=";
+  private static final String EQUAL_TO = "=";
```
djaiswal83, line 132: EQUALS_TO

djaiswal83 (changes requested):
One minor change. Rest looks good.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Changed to EQUALS_TO

djaiswal83 (dismissed):

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,71 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+  @Override
+  public void deleteAgedContainers(Duration containerValidity) throws ExecutorManagerException {
+    logger.info(String.format("Cleaning up containers older than %d days",
+        containerValidity.toDays()));
+    final Set<Integer> containersToDelete = getStaleContainers(containerValidity);
+    for (int execId : containersToDelete) {
+      if (execId > 0) {
+        logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
+            getServiceName(execId)));
+        deleteContainer(execId);
```
ypadron-in, line 429: I think we should handle exceptions thrown by `deleteContainer` method here. For example, lets say there are 3 containers to delete but first deletion attempt failed, because an exception was thrown we're not going to attempt deletion of the remaining 2. Moreover, the cleanup routine will run periodically and if the list of containers to delete is sorted the same every time only we'll probably never pass the first attempt.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -202,10 +216,8 @@ private void retryFlowQuietly(ExecutableFlow flow, Status originalStatus) {
...
+      logger.error("Unexpected RuntimeException while deleting container.", e);
```
ypadron-in, line 220: can we adjust the message to log `Unexpected exception` instead of `Unexpected RuntimeException`?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,71 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+      final Long validStartTimeStamp = System.currentTimeMillis() - containerValidity.toMillis();
+      staleContainerExecIdSet =
+          items.getItems().stream()
+              .filter((pod) -> pod.getMetadata().getCreationTimestamp().getMillis() < validStartTimeStamp)
+              .map((pod) -> pod.getMetadata().getLabels().getOrDefault(EXECUTION_ID_LABEL_NAME,
+                  EXECUTION_ID_LABEL_PREFIX + DEFAULT_EXECUTION_ID))
+              .map((str) -> str.substring(str.indexOf(
+                  EXECUTION_ID_LABEL_PREFIX) + EXECUTION_ID_LABEL_NAME.length()))
+              .map((execId) -> Integer.valueOf(execId))
+              .collect(Collectors.toSet());
```
ypadron-in, line 458: I believe the part 
```
.map((str) -> str.substring(str.indexOf(
   EXECUTION_ID_LABEL_PREFIX) + EXECUTION_ID_LABEL_NAME.length()))
```
is not correct as we're not adding correct length. If I understand correctly it should be `EXECUTION_ID_LABEL_PREFIX.length`. Also why doing indexOf? At this point shouldn't we have a string like "execid-12345" and therefore we know index is zero?

Maybe it's just me but I find this code difficult to understand and error-prone. Is it possible to encapsulate the logic in a method unit test it to make sure labels are processed correctly and make it a bit more readable?

hqin2016: You're right. Let me correct it. I used string.replace to replace the "execid-" with empty string initially. Later @djaiswal83 suggested using substring instead. I'll add a unit test.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,71 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+      // Get all execution ids of the pods whose age is older than Azkaban max flow running time
+      // (e.g. 10 days)
+      final Long validStartTimeStamp = System.currentTimeMillis() - containerValidity.toMillis();
+      staleContainerExecIdSet =
+          items.getItems().stream()
+              .filter((pod) -> pod.getMetadata().getCreationTimestamp().getMillis() < validStartTimeStamp)
+              .map((pod) -> pod.getMetadata().getLabels().getOrDefault(EXECUTION_ID_LABEL_NAME,
+                  EXECUTION_ID_LABEL_PREFIX + DEFAULT_EXECUTION_ID))
+              .map((str) -> str.substring(str.indexOf(
+                  EXECUTION_ID_LABEL_PREFIX) + EXECUTION_ID_LABEL_PREFIX.length()))
```
svnarumugam, line 456: Sorry if this comment was too late! I was wondering if we can add this label also as part of the label selector we are supplying at line 442. I believe we can pass multiple labels. I find this piece of code error-prone and slightly hard to read too.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Added a try catch block to handle exception thrown by deleteContainer.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: Changed the log message as requested.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,88 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
    */
   @Override
   public void deleteContainer(final int executionId) throws ExecutorManagerException {
-    deletePod(executionId);
-    if (isServiceRequired()) {
-      deleteService(executionId);
+    try { // if pod deletion is not successful, the service deletion can still be handled
+      deletePod(executionId);
+    } finally {
```
sshardool, line 408: Include `catch` that will log any exceptions during pod and service deletions.

sshardool (commented):
Was examining the code in `KubernetesContainerizedImpl` and noticed some larger issues related to thread-safety.
We create a singleton of the class which is used within different threads (including in CleanupManager) but the class itself is not inherently thread-safe. 
At the very least parallel access to the member field kubernetes `ApiClient` could be problematic (among others). cc: @jakhani 
However, addressing this should be out of the scope for this PR and can be followed up subsequently. 

Thanks for addressing other comments on the PR, left 1 more minor comment for logging of exceptions in the try-catch.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: In `deletePod ` we're catching and logging exception, and throwing ExecutorManagerException. I'm not sure if we want to catch the exception and log the error again in in `deleteContainer` level.

svnarumugam (dismissed):
Thanks a lot Hao for addressing all the comments! LGTM.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,88 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+        try {
+          final String execId = execIdLabel.substring(execIdLabel.indexOf(
+              EXECUTION_ID_LABEL_PREFIX) + EXECUTION_ID_LABEL_PREFIX.length());
+          staleContainerExecIdSet.add(Integer.valueOf(execId));
+        } catch (Exception e) {
+          logger.error(String.format("Unable to retrieve execution id from pod %s",
+              pod.getMetadata().getName()), e.getMessage());
+        }
+      }
     }
```
ypadron-in, line 486: Thanks for encapsulating and testing this logic!
A few additional suggestions:
-`pod.getMetadata()` is called 3 times, consider calling once and reusing the result.

-I don't think `execIdLabel.indexOf()` call is needed as index should always be zero. 
`execIdLabel.substring(EXECUTION_ID_LABEL_PREFIX.length());` should be sufficient.

-this method skips executions that have started recently, please also consider exclusion of pods without an execution id (excluding execid--1 from the resulting list). 

-consider adding `s` in the method name: getExecutionId_**s**_FromPodList

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
ypadron-in: Thank you! Looks like `RuntimeException` is still there, perhaps the change was not `git add`?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +403,88 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+
+  /**
+   * This method is used to fetch all stale pods in the current az cluster and namespace
+   * @param containerValidity container valid duration
+   * @return Set of stale container's execution ids
+   */
+  private Set<Integer> getStaleContainers(final Duration containerValidity) throws ExecutorManagerException {
+    try {
+      // Select pods only from current Azkaban cluster and namespace
+      final String label = CLUSTER_LABEL_NAME + "=" + this.clusterName + "," + LABEL_POSTFIX;
```
ypadron-in, line 447: please, consider using `APP_LABEL_NAME + "=" + POD_APPLICATION_TAG` instead of defining constant `LABEL_POSTFIX`

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Used podMetadata instead of calling `pod.getMetadata()` 3 times. Changed the function name to `getExecutionIdsFromPodList`
exec-id == -1 is excluded in the `deleteAgedContainers`

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Used APP_LABEL_NAME + "=" + POD_APPLICATION_TAG instead of LABEL_POSTFIX

svnarumugam (approved):
LGTM! Thanks again! :)

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +402,90 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+    final Set<Integer> containersToDelete = getStaleContainers(containerValidity);
+    for (final int execId : containersToDelete) {
+      if (execId > 0) {
+        logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
+            getServiceName(execId)));
+        try {
+          deleteContainer(execId);
+        } catch (ExecutorManagerException e) {
+          logger.error(String.format("Unable to delete stale pod and service of exec-id %d",
+              execId), e.getMessage());
```
djaiswal83, line 432: This method is throwing the ExecutorManagerException, then shouldn't it be thrown again here?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -437,35 +442,47 @@ public void deleteAgedContainers(Duration containerValidity) throws ExecutorMana
...
-      logger.error("Error in parsing execution Id from pod label", ne.getMessage());
-      throw new ExecutorManagerException(ne);
+    }
+  }
+
+  /**
+   * Obtain a set of execution ids from stale pod list
+   * @param podList
+   * @param validStartTimeStamp
+   * @return
```
djaiswal83, line 467: Fill java doc for return value

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +402,90 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+      final V1ObjectMeta podMetadata = pod.getMetadata();
+      if (podMetadata.getCreationTimestamp().getMillis() < validStartTimeStamp) {
+        final String execIdLabel =
+            podMetadata.getLabels().getOrDefault(EXECUTION_ID_LABEL_NAME,
+            EXECUTION_ID_LABEL_PREFIX + DEFAULT_EXECUTION_ID);
+        try {
+          final String execId = execIdLabel.substring(execIdLabel.indexOf(
+              EXECUTION_ID_LABEL_PREFIX) + EXECUTION_ID_LABEL_PREFIX.length());
+          staleContainerExecIdSet.add(Integer.valueOf(execId));
+        } catch (Exception e) {
```
djaiswal83, line 482: nit  : final

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: I'd prefer not to throw the exception here to break the for loop

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Added.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Added final

djaiswal83 (approved):
Looks good to me.  Thanks for addressing the comments.

byjiang1996 (approved):

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -116,6 +121,17 @@ public void cleanUpStaleFlows() {
         Entry::getKey).forEach(this::cleanUpStaleFlows);
   }
 
+  /**
+   * Try to clean the stale containers that are older than maximum azkaban flow running time
+   */
+  public void cleanUpStaleContainers() {
+    try {
+      this.containerizedImpl.deleteAgedContainers(validityMap.get(Status.RUNNING).getFirst());
```
sshardool, line 129: Looks like the way it's currently implemented, any unchecked exception while deleting a container will prevent the deletion of other 'stale' containers. 
Please handle all exceptions (including checked) within `deleteAgedContainers` and/or other underlying methods. 
Note that this is also good use-case for suppressed exceptions: https://www.baeldung.com/java-suppressed-exceptions
(Not necessary in our case as long as we log all the unchecked exceptions)

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
sshardool: Good point @svnarumugam, I was curious about it too and checked, seems like multiple schedules within a single ScheduledExecutor are supported, though not very obvious from the javadoc. Single thread is ok as the operations are exepected to be infrequent and quick.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -76,10 +78,13 @@
...
       final ContainerizedImpl containerizedImpl,
       final ContainerizedDispatchManager containerizedDispatchManager) {
-    this.cleanupIntervalMin = azkProps
+    this.executionCleanupIntervalMin = azkProps
         .getLong(
             ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_EXECUTION_CLEANUP_INTERVAL_MIN,
-            DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL.toMinutes());
+            DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL.toMinutes());
+    this.containerCleanupIntervalMin = azkProps.getLong(
+        ContainerizedDispatchManagerProperties.CONTAINERIZED_STALE_CONTAINER_CLEANUP_INTERVAL_MIN,
```
sshardool, line 86: Would be good to sanity test this change and config on a test cluster by launching multiple fllow-pods which take a predictable amount of time to run and setting the CLEANUP_INTERVAL config to a few minutes (say 15 or so). 
Also, please explicitly test for cases the flow-pod has been deleted but the corresponding service still exists. (should be auto-deleted after these changes).

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -398,10 +402,91 @@ public void createContainer(final int executionId) throws ExecutorManagerExcepti
...
+    final Set<Integer> containersToDelete = getStaleContainers(containerValidity);
+    for (final int execId : containersToDelete) {
+      logger.info(String.format("Deleting stale pod %s and service %s", getPodName(execId),
+          getServiceName(execId)));
+      try {
+        deleteContainer(execId);
+      } catch (final ExecutorManagerException e) {
+        logger.error(String.format("Unable to delete stale pod and service of exec-id %d",
+            execId), e.getMessage());
+      }
```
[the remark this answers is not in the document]
hqin2016: I'll address this in a separate pr. Thanks for bringing it up.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: @sshardool I'm addressing this in https://github.com/azkaban/azkaban/pull/3058. Thanks for the suggestion.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
hqin2016: It's now handled in deleteAgedContainers

PR #3056, opened 2022-02-24
Get event data distinguishable for flow retries

byjiang1996:
In containerized azkaban executions, auto restart is enabled for flow failed due to Kubernetes failures in PREPARING stage. Right now the Azkaban flow life cycle events emitted for original and retry are the same. There needs to be a way to distinguish between retries and originally failed flow execution in the azkaban events.

Validation: 
Picture1

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -16,6 +16,8 @@
 
 package azkaban.executor;
 
+import static azkaban.Constants.EventReporterConstants.EXECUTION_RETRY_BY_AZKABAN;
+
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.sla.SlaOption;
 import azkaban.utils.TypedMapWrapper;
@@ -63,8 +65,6 @@ public class ExecutionOptions {
   public static final String FAILURE_ACTION_OVERRIDE = "failureActionOverride";
   private static final String MAIL_CREATOR = "mailCreator";
   private static final String MEMORY_CHECK = "memoryCheck";
-  // Flow restartability check
-  private static final String EXECUTION_RETRY_BY_AZKABAN = "executionRetryByAzkaban";
   private boolean isExecutionRetried = false;
 
   private boolean notifyOnFirstFailure = true;
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -16,7 +16,7 @@
 
 package azkaban.executor;
 
-import static azkaban.Constants.EventReporterConstants.EXECUTION_RETRY_BY_AZKABAN;
+import static azkaban.Constants.EventReporterConstants.EXECUTION_RETRIED_BY_AZKABAN;
 import static azkaban.Constants.EventReporterConstants.ORIGINAL_FLOW_EXECUTION_ID_BEFORE_RETRY;
 
 import azkaban.executor.mail.DefaultMailCreator;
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -64,7 +64,7 @@
   private static final String MAIL_CREATOR = "mailCreator";
   private static final String MEMORY_CHECK = "memoryCheck";
   // Flow restartability check
-  private static final String EXECUTION_RETRY = "executionRetryByAzkaban";
+  private static final String EXECUTION_RETRY_BY_AZKABAN = "executionRetryByAzkaban";
```
hqin2016, line 67: You can remove this field and use the one defined in Constants.

byjiang1996: Sounds good! Done.

djaiswal83 (changes requested):
As discussed earlier, lets add the original flow execution id in the event.

byjiang1996:
Original flow execution id is added in the event.
Validation:
Picture1

hqin2016 (approved):
LGTM

hqin2016 (approved):

svnarumugam (commented):
Thanks Binyao for working on this change! The screenshot you have attached is quite helpful!

changes look good. I have added some questions to appreciate this change better!
1. if we have retry count as 2, we will have E1 (failed)-> E2 (failed, retry-1) -> E3 (retry-2). Does it mean original execution id of E2 and E3 will refer to E1?
2. where do we specify these properties should go under propagated properties?
3. nit: names like `executionRetriedByAzkaban` or `retried` or `isExecutionRetried` represent the better. what do you think?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -67,6 +67,7 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
     options.setMailCreator(flow.getMailCreator());
     // Update the flow options so that the flow will be not retried again by Azkaban
     options.setExecutionRetried(true);
+    options.setOriginalFlowExecutionIdBeforeRetry(exFlow.getExecutionId());
```
ypadron-in, line 70: If an execution is retried 3 times how is the original execution id reported? Based on current code it seems only the first retry would report the original execution id, the other 2 attempts would report a retry as original execution.

byjiang1996: Yes, other two will report the retry as original execution because they are retrying a retried execution:
1. #4 will report #3 as original flow
2. #3 will report #2 as original flow
3. #2 will report #1 as original flow

Add @svnarumugam into the thread for the explanation of original execution id.

Do we want to use #1's id as both #4's and #3's original execution id, or just want to continue the current behavior: cascaded original execution id? I think either makes sense to me.

byjiang1996:
@svnarumugam , sorry that I lost the other two comments you left:
1. In Linkedin's internal EventReporter plugin developed by Hao, any non-basic metadata (e.g. flowName and projectName are the basic event metadata and thus will not go into propagated properties field) will be passed through and added into propagated properties field.
2. retry -> retried: done.

byjiang1996:
@svnarumugam and @ypadron-in , the retry logic has been updated as follows:
if we have retry count as 2, we will have E1 (failed)-> E2 (failed, retry-1) -> E3 (retry-2). Original execution id of E2 and E3 will refer to E1.

ypadron-in (approved):
Thanks for addressing the comments! LGTM

djaiswal83 (approved):
Looks good to me. Thanks for enhancing the events.

PR #3057, opened 2022-02-26
Throw exception only if no metastore toekn is fetched from EXTRA_HCAT_LOCATION urls

hqin2016:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -424,14 +424,23 @@ private void fetchMetaStoreToken(final String userToProxyFQN,
...
+                hcatToken =
+                    fetchHcatToken(userToProxyFQN, hiveConf, thriftUrl, logger);
+                cred.addToken(hcatToken.getService(), hcatToken);
+                ++extraHcatTokenCount;
+              } catch (Exception e) {
+                logger.error("Failed to fetch extra metastore tokens from : " + thriftUrl, e);
+              }
+            }
+            if (0 == extraHcatTokenCount) {
+              throw new HadoopSecurityManagerException("No extra metastore token can be fetched.");
```
jakhani, line 443: Is this fix only required for EXTRA_HCAT_LOCATION? We don't need similar fix for EXTRA_HCAT_CLUSTERS?

jakhani (approved):

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
hqin2016: I added fix for EXTRA_HCAT_CLUSTERS as well.

jakhani (approved):

PR #3058, opened 2022-02-26
Reset KubernetesContainerizedImpl class instances cpuLimit and memoryLimit to config values in case they were overridden by user requested resources in the flow parameters

hqin2016:


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -231,6 +234,7 @@ public KubernetesContainerizedImpl(final Props azkProps,
             DEFAULT_MEMORY_LIMIT_MULTIPLIER);
     this.memoryLimit = this.getResourceLimitFromResourceRequest(this.memoryRequest, this.memoryRequest,
         memoryLimitMultiplier);
+    this.memoryLimitByConf = this.memoryLimit;
```
amielkelinkedin, line 237: can we change this variable name to defaultMemoryLimit?

amielkelinkedin:
@hqin2016, Thank you for making this change! I believe this should work!

svnarumugam (commented):
Thanks Hao for the fix and clarifying on slack! I was wondering about the motivation for this change - adding it in the PR description will help appreciate this change more. 

Added a comment, rest looks good to me.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -710,6 +714,7 @@ V1ObjectMeta createPodMetadata(final int executionId, Map<String, String> flowPa
   String getFlowContainerCPURequest(final Map<String, String> flowParam) {
     if (flowParam == null || flowParam.isEmpty() || !flowParam
         .containsKey(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_CPU_REQUEST)) {
+      this.cpuLimit = this.cpuLimitByConf;
```
svnarumugam, line 717: for my understanding, why do we have variables cpuLimit, memorylimit or any resourcelimits as instance variables? I was wondering about the implication of this because we use this `KubernatesContainerizedImpl` instance to create pods for all the executions. Please correct me if I am missing something obvious!

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: I'm eliminating all cpu and memory limit instances in the class. The limit should always be determined dynamically from resource request.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
hqin2016: Thanks for the suggestion. I'm eliminating all cpu and memory limit instances in the class.

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
```diff
@@ -288,6 +288,56 @@ public void testCPUAndMemoryRequestedFromProperties() throws Exception {
...
+
+    // container 3: when user again does not specify resource request, the limit should again be
+    // the same as config defined limit
+    flowParam.remove(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_CPU_REQUEST);
+    flowParam.remove(FlowParameters.FLOW_PARAM_FLOW_CONTAINER_MEMORY_REQUEST);
+    this.kubernetesContainerizedImpl.getFlowContainerCPURequest(flowParam);
+    this.kubernetesContainerizedImpl.getFlowContainerMemoryRequest(flowParam);
+
+    assert (this.kubernetesContainerizedImpl.getCpuLimit()).equals(cpuLimitByConf);
+    assert (this.kubernetesContainerizedImpl.getMemoryLimit()).equals(memoryLimitByConf);
```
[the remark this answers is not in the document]
hqin2016: Changing to Assert.assertTrue.

djaiswal83 (approved):
Thanks for fixing this bug. With the changes made by other reviewers, the code looks good to me.

svnarumugam (approved):
thanks for addressing the comments Hao. Looks good to me.

PR #3061, opened 2022-03-02
Thread safe kubernetes containerized impl

byjiang1996:
Currently public methods of the class `KubernetesContainerizedImpl` do not gurantee any thread-safety. Method `deleteContainer` in particular is executed from several different threads. Arguably 'createContainer` can also be executed from different threads (depending on the exact dispatch logic)

Among synchronized, BlockingQueue, atomic variables and stateless design, synchronized keyword seems to be the best way to go considering that:
1. It is recommended to use coreV1Api class per thread: coreV1Api class is not thread-safe from this github [page](https://github.com/kubernetes-client/java/issues/100).
2. In CreatePod method, we access sql db multiple times and do read+write operations. Guarantee the read-write ordering is much safer (no need to think through the side-effect of out-of-order read-writes in multi-threading) using synchronized method.


hqin2016 (approved):
Thanks for the fix. LGTM.

PR #3062, opened 2022-03-02
Logging time for executor ping

utk-12:
Adding a log to the executor health checker class to capture ping time /success/ time-out of ping.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
@@ -136,14 +136,17 @@ public class ExecutorHealthChecker {
       Optional<ExecutorManagerException> healthcheckException = Optional.empty();
       Map<String, Object> results = null;
       try {
-        // Todo jamiesjc: add metrics to monitor the http call return time
+        long pingTime = System.currentTimeMillis();
         results = this.apiGateway
             .callWithExecutionId(executor.getHost(), executor.getPort(),
                 ConnectorParams.PING_ACTION, null, null, null);
+        pingTime = System.currentTimeMillis() - pingTime;
+        logger.info("Got ping response from " + executorDetailString(executor)
+            + " in " + pingTime + "ms");
       } catch (final ExecutorManagerException e) {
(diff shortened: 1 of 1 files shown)


jakhani (approved):

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -136,14 +136,17 @@ void checkExecutorHealth() {
       Optional<ExecutorManagerException> healthcheckException = Optional.empty();
       Map<String, Object> results = null;
       try {
-        // Todo jamiesjc: add metrics to monitor the http call return time
+        long pingTime = System.currentTimeMillis();
         results = this.apiGateway
             .callWithExecutionId(executor.getHost(), executor.getPort(),
                 ConnectorParams.PING_ACTION, null, null, null);
+        pingTime = System.currentTimeMillis() - pingTime;
+        logger.info("Got ping response from " + executorDetailString(executor)
```
sshardool, line 144: Maybe log this in a `finally` block (line 151) so that we can can get this info for failed cases as well ?

sshardool, line 144: Any thoughts around how frequently these log messages will be printed? 
(Just to make sure the logs are not excessively filled up with these,  AFAIR the healtcheck was not happen too frequently for this to be a concern though)

sshardool (commented):
Thanks for making the change!

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
utk-12: Ack. I shall do that.

utk-12: Its every 5 mins from what I see. Hope thats okay.

ypadron-in: Is it possible to log action and executor details before the call? That way if the call hangs we can figure out the bad executor.

PR #3063, opened 2022-03-04
Implement execution SLA checks for jobs in containerized executions.

ypadron-in:
This PR implements execution SLA checks for jobs in containerized executions.
It also makes some refactoring to the executions SLA feature.

Tested locally in solo server, testing clusters and Kubernes pods. 
Verified that both Flow and Job SLA checks are setup and alert/kill actions are taken upon SLA misses.

Follow up: work on adding tests for the feature overall which do not exist today.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
@@ -16,8 +16,6 @@
 
 package azkaban.executor;
 
-import azkaban.utils.FileIOUtils.LogData;
-
 /**
  * Defines operations common to all flow execution controllers.
  */
--- a/azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package azkaban.executor;
+
+import azkaban.utils.FileIOUtils.LogData;
+
+/**
+ * Defines operations common to all flow execution controllers.
+ */
+public interface IFlowRunnerManager {
+
+  void pauseFlow(int execId, String user) throws ExecutorManagerException;
+
+  void resumeFlow(final int execId, String user) throws ExecutorManagerException;
+
+  void cancelFlow(int execId, String user) throws ExecutorManagerException;
+
+  void cancelJobBySLA(int execId, String jobId) throws ExecutorManagerException;
+
+  /**
+   * Attempts to retry the failed jobs in a running execution.
+   */
+  void retryFailures(int execId, String user) throws ExecutorManagerException;
+}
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package azkaban.executor;
+
+import azkaban.utils.FileIOUtils.LogData;
+
+/**
+ * Defines operations common to all flow execution controllers.
+ */
+public interface IFlowRunnerManager {
```
djaiswal83, line 24: This interface is not needed as FlowContainer does not implement half of the APIs and does not require execId at all.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -114,7 +117,7 @@
  *  RUNNING from FlowRunner. The rest of the state machine is handled by FlowRunner.
  */
 @Singleton
-public class FlowContainer implements IMBeanRegistrable, EventListener<Event> {
+public class FlowContainer implements IFlowRunnerManager, IMBeanRegistrable, EventListener<Event> {
```
djaiswal83, line 120: Please remove the IFlowRunnerManager

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -509,13 +525,26 @@ public void cancelFlow(final int execId, final String user)
...
+  public void cancelJobBySLA(final int execId, final String jobId)
+      throws ExecutorManagerException {
+    if (this.flowRunner == null) {
+      throw new ExecutorManagerException("Execution " + execId + " is not running.");
+    }
+
+    this.flowRunner.getExecutableFlow().setModifiedBy("SLA");
+
+    for (final JobRunner jobRunner : this.flowRunner.getActiveJobRunners()) {
+      if (jobRunner.getJobId().equals(jobId)) {
```
djaiswal83, line 538: Please put negative case here and put the positive case after that.
if (!jobRunner.getJobId().equals(jobId)) { 
  continue};
}
<rest of the code>

djaiswal83 (changes requested):
Thanks for implementing the missing feature in FlowContainer.
The idea of the interface is not very neat here and not required.
Also, passing a top level object such as FlowRunnerManager to Kill*Action classes does not seem to be the right way to do it. 
Thanks for cleaning up the unused APIs.

azkaban-common/src/main/java/azkaban/executor/IFlowRunnerManager.java
ypadron-in: Could you please elaborate on why the interface isn't needed? 
-As I see it FlowContainer and FlowRunnerManager have the same main responsibility: to control a flow execution.
-Trigger action objects from `KillExecutionAction` and `KillJobAction` depend on such flow execution controllers for their functionality. Without the interface then we need to go back to if-else of current DispatchMethod and using  `SERVICE_PROVIDER.getInstance` of Guice which I believe is an anti-pattern and not so much extensible/maintainable code.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
ypadron-in: Do you mean like this?
```
{ // start for loop
  if (!jobRunner.getJobId().equals(jobId)) {
    continue
  };
  logger.info("Killing job " + jobId + " in execution " + execId + " by SLA");
  jobRunner.killBySLA();
  break;
} // end for
```
I understand that asking for the negative case is better overall but given the simplicity of the logic (log and call kill action) existing code looks more readable and simple to me than above block. I can definitely make the change if you feel strongly about it though.

jakhani (approved):
Can you please check if you can add unit tests for SLA kill? Thank you for working on this PR.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
djaiswal83: This code may become larger if further development happens here for some reason in the future. It is better to keep it the way suggested for such cases.

djaiswal83 (changes requested):
I resolved my interface related comments. I just have one comment, please address that. Overall the code looks much cleaner due to this PR. Thanks!

djaiswal83 (approved):
Thanks for addressing the comments. The code looks good.

ypadron-in:
> Can you please check if you can add unit tests for SLA kill? Thank you for working on this PR.

I'll add tests in a follow up PR

PR #3067, opened 2022-03-14
Adding a timeout parameters for all HTTP Clients

utk-12:
Using parameterized approach for adding HTTP timeouts.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -209,7 +209,7 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
 
     final String[] hostPortSplit = hostPort.split(":");
     return this.apiGateway.callForJsonObjectMap(hostPortSplit[0],
-        Integer.valueOf(hostPortSplit[1]), "/jmx", null, Optional.of(-1), paramList);
+        Integer.valueOf(hostPortSplit[1]), "/jmx", null, Optional.empty(), paramList);
   }
 
   /**
@@ -235,7 +235,7 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
         .add(new Pair<>(ConnectorParams.ACTION_PARAM, action));
 
     return this.apiGateway.callForJsonObjectMap(executor.getHost(), executor.getPort(),
-        "/stats", null, Optional.of(-1), paramList);
+        "/stats", null, Optional.empty(), paramList);
   }
 
   @Override
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -209,7 +209,7 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
 
     final String[] hostPortSplit = hostPort.split(":");
     return this.apiGateway.callForJsonObjectMap(hostPortSplit[0],
-        Integer.valueOf(hostPortSplit[1]), "/jmx", null, -1, paramList);
+        Integer.valueOf(hostPortSplit[1]), "/jmx", null, Optional.of(-1), paramList);
   }
 
   /**
@@ -235,7 +235,7 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
         .add(new Pair<>(ConnectorParams.ACTION_PARAM, action));
 
     return this.apiGateway.callForJsonObjectMap(executor.getHost(), executor.getPort(),
-        "/stats", null, -1, paramList);
+        "/stats", null, Optional.of(-1), paramList);
   }
 
   @Override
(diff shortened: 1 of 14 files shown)


azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -209,7 +209,7 @@ public int getExecutableFlows(final int projectId, final String flowId, final in
 
     final String[] hostPortSplit = hostPort.split(":");
     return this.apiGateway.callForJsonObjectMap(hostPortSplit[0],
-        Integer.valueOf(hostPortSplit[1]), "/jmx", null, paramList);
+        Integer.valueOf(hostPortSplit[1]), "/jmx", null, Optional.of(-1), paramList);
```
jakhani, line 212: Can you please use Optional.empty()? It will return false for Optional.isPresent(). Same for other places.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -182,10 +183,19 @@ public boolean isReverseProxyEnabled() {
...
-    return httpClientBuilder.build();
+    if (httpTimeout.isPresent()) {
+      int timeout = httpTimeout.get();
+      final RequestConfig requestConfig = RequestConfig.custom()
+          .setConnectTimeout(timeout)
+          .setSocketTimeout(timeout)
+          .build();
+      return httpClientBuilder.setDefaultRequestConfig(requestConfig).build();
+    } else {
+      return httpClientBuilder.create().build();
```
ypadron-in, line 197: `create()` method was called by the `custom()` invocation a few lines above. Please don't call `create()` again here.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
ypadron-in: Alternatively, we can use the default timeout value (`-1`) in the httpClient library and pass that when no timeout is defined.

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -196,7 +206,8 @@ protected CloseableHttpClient createHttpsClient() {
    * @return the response object type of which is specified by user.
    * @throws UnsupportedEncodingException, IOException.
    */
-  public String httpsPost(final URI uri, final List<Pair<String, String>> params)
+  public String httpsPost(final URI uri,
+      final Optional<Integer> httpTimeout, final List<Pair<String, String>> params)
```
ypadron-in, line 210: Please consider passing a regular Integer (as opposed to Optional\<Integer\>) with value -1 when there is no timeout.

ypadron-in, line 210: Please consider passing a regular Integer (as opposed to Optional<Integer>) with value -1 when there is no timeout.

ypadron-in (changes requested):
Thank you for working on this. 
Left a few comments below.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
utk-12: -1 is the default  by the library actually. So passing empty would indirectly make it -1.

azkaban-common/src/main/java/azkaban/executor/ExecutorHealthChecker.java
```diff
@@ -70,6 +71,8 @@ public ExecutorHealthChecker(final Props azkProps, final ExecutorLoader executor
     this.executorLoader = executorLoader;
     this.apiGateway = apiGateway;
     this.alerterHolder = alerterHolder;
+    this.executorPingTimeout =
+        azkProps.getInt(ConfigurationKeys.AZKABAN_EXECUTOR_PING_TIMEOUT, 5000);
```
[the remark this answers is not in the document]
utk-12: ack

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
```diff
@@ -182,10 +183,19 @@ public boolean isReverseProxyEnabled() {
    *
    * @return http client
    */
-  protected CloseableHttpClient createHttpsClient() {
+  protected CloseableHttpClient createHttpsClient(final Optional<Integer> httpTimeout) {
     final HttpClientBuilder httpClientBuilder = HttpClients.custom()
         .setSSLSocketFactory(this.tlsSocketFactory);
-    return httpClientBuilder.build();
+    if (httpTimeout.isPresent()) {
+      int timeout = httpTimeout.get();
```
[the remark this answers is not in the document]
utk-12: ack

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
utk-12: ack

azkaban-common/src/main/java/azkaban/executor/ExecutorApiClient.java
utk-12: Have changed it to optional.empty() ins case of no timeout being provided. basically all the places where Optional.of(-1) is present right now.

jakhani (approved):

ypadron-in (approved):
Thank you addressing the comments!

svnarumugam:
Thank you so much for working on this change. Left two comments to appreciate this change better.

[not to block shipping] 
1. Multiple methods accept timeout value as Optional. This is done better with a method overloading than accepting an Optional object. any challenges we faced in method overloading in this context?

`Reference`: https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments

2. I certainly see the usefulness of this change. I am curious to know what triggered this change. will this change be applicable in containerized executions too @utk-12?

PR #3071, opened 2022-03-31
Use local path instead of name for adding hadoop configuration

byjiang1996:
In the past, routerConfPath should be relative to classpath but it is not what we want when we first look at the code -- error-prune. Changing this to a local file path instead: i.e. relative to the current program execution path.

Screen Shot 2022-03-30 at 6 37 12 PM


djaiswal83 (approved):
Thanks for working on the fix. LGTM

PR #3072, opened 2022-04-01
Retry EXECUTION_STOPPED executions when allow.restart.on.execution.stopped:true is added to flow parameters/runtime properties

hqin2016:
Users can control whether to auto retry the EXECUTION_STOPPED flows.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -138,7 +138,8 @@ public class ExecutionControllerUtils {
       return;
     }
     final ExecutionOptions options = flow.getExecutionOptions();
-    if (options == null || options.isExecutionRetried()) { // flow can only be retried once
+    // flow can only be retried once
+    if (options == null || options.isExecutionRetried()) {
       return;
     }
     // If the original execution status is not EXECUTION_STOPPED, it can be retried
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -124,26 +124,32 @@ public class ExecutionControllerUtils {
 
   /**
    * This method tries to restart the flow for certain statuses otherwise simply return.
-   * If the flow is already retried once, then it won't be retried again.
-   * If the status is EXECUTION_STOPPED, then it will be retried only if
+   * There are three scenarios that this method is called: 1. a flow is cleaned up by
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
```diff
@@ -278,7 +278,7 @@ private void updatePodStatus(AzPodStatusMetadata event) {
       // Emit EXECUTION_STOPPED flow event
       this.fireEventListeners(Event.create(executableFlow,
           EventType.FLOW_FINISHED, new EventData(executableFlow)));
-      ExecutionControllerUtils.restartFlow(executableFlow, originalStatus);
+      ExecutionControllerUtils.restartFlow(executableFlow, Status.EXECUTION_STOPPED);
```
svnarumugam, line 281: The context of why `execution_stopped` status is passed is not easy to understand here. If the intention was to make sure we get to the flow param part in the restartFlow method, can we just have a separate method in `ExecutionControllerUtils` and call that method?

svnarumugam (dismissed):
Thanks for making this change @hqin2016 .

azkaban-common/src/main/java/azkaban/executor/container/watch/FlowStatusManagerListener.java
byjiang1996: Instead of passing a Status.EXECUTION_STOPPED here, can we have a high-priority if in ExecutionControllerUtils.restartFlow checking:
1. currentStatus == EXECUTION_STOPPED && allow.restart.on.execution.stopped:true (should be placed after "options.isExecutionRetried()" check because we don't want to restart a flow twice)

So the flow can be still restarted there.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -124,26 +124,32 @@ public static void finalizeFlow(final ExecutorLoader executorLoader, @Nullable f
...
    * @param originalStatus
    */
   public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
-    if (!RESTARTABLE_STATUSES.contains(originalStatus) && flow.getStatus() != EXECUTION_STOPPED) {
+    if (!RESTARTABLE_STATUSES.contains(originalStatus)) {
       return;
     }
     final ExecutionOptions options = flow.getExecutionOptions();
-    if (options == null || options.isExecutionRetried()) {
+    if (options == null || options.isExecutionRetried()) { // flow can only be retried once
```
djaiswal83, line 141: nit : move the comment in new line.

djaiswal83 (approved):

PR #3074, opened 2022-04-13
Do not add job's jobtype image if a job is disabled in containerized …

byjiang1996:
…Azkaban

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
@@ -61,6 +61,8 @@ public class ContainerImplUtils {
         populateJobTypeForFlow(subNode, jobTypes);
       }
     } else {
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -60,7 +61,9 @@ private static void populateJobTypeForFlow(final ExecutableNode node, Set<String
         populateJobTypeForFlow(subNode, jobTypes);
       }
     } else {
-      jobTypes.add(node.getType());
+      if (node.getStatus() != Status.DISABLED) {
```
jakhani, line 64: Can you please add comment? Also log for which is disabled and not added? It might help later on for debugging.

byjiang1996: Good point! Done.

djaiswal83 (approved):

PR #3075, opened 2022-04-13
K8s Java client upgrade

utk-12:
Upgrading K8s Java client:

**_Why this change was required?_** 
AKS and On-Prem K8s cluster upgrades are imminent and we will have breaking API compatibility as mentioned here,
https://github.com/kubernetes-client/java/wiki/2.-Versioning-and-Compatibility

**_Issues that triggered this investigation_**
K8s watch events were not being received after a few days on AKS. 
After looking through the following issues https://github.com/kubernetes-client/java/issues/1370 we realized our client version might become incompatible soon. 

We still plan to keep the call/read/connect timeouts to the watch server.

**_High-level Changes_** :

1. We now use offsetDateTime type for getting pod creation timestamps instead of getting the timestamp using System.currentTimeMillis()
2. We now use  Enums for RestartPolicy 

**_Gotchas:_**:
We are now using the GenericKubernetesApi instead of CoreV1APi for service deletion due to :
https://github.com/kubernetes-client/java/wiki/6.-Known-Issues


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -437,13 +439,24 @@ private Set<Integer> getStaleContainers(final Duration containerValidity) throws
...
       final String label =
           CLUSTER_LABEL_NAME + "=" + this.clusterName + "," + APP_LABEL_NAME + "=" + POD_APPLICATION_TAG;
-      final V1PodList podList= this.coreV1Api.listNamespacedPod(this.namespace, null,
-          null, null, null, label,
-          null, null, null, null);
+      final V1PodList podList= this.coreV1Api.listNamespacedPod(
+          this.namespace,
+          null,
+          null,
+          null,
```
jakhani, line 446: Why are you removing label?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1159,7 +1178,13 @@ private void createService(final int executionId) throws ExecutorManagerExceptio
           .withPort(String.valueOf(this.servicePort))
           .withTimeoutMs(String.valueOf(this.serviceTimeout))
           .build();
-      this.coreV1Api.createNamespacedService(this.namespace, serviceObject, null, null, null);
+      this.coreV1Api.createNamespacedService(
+          this.namespace,
+          serviceObject,
```
jakhani, line 1183: Can you please follow stying that Azkaban has provided in repo?

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -77,7 +79,12 @@ public KubernetesWatch(ApiClient apiClient,
     this.apiClient = apiClient;
     // no timeout for request completion
     OkHttpClient httpClient =
-        this.apiClient.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build();
+        this.apiClient.getHttpClient().newBuilder()
+            .protocols(Arrays.asList(Protocol.HTTP_2,Protocol.HTTP_1_1))
+            .connectTimeout(60, TimeUnit.SECONDS)
+            .callTimeout(1200, TimeUnit.SECONDS)
```
jakhani, line 85: Shall we introduce these timeouts from config? Just thinking out loud.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -114,7 +122,8 @@ protected void initializePodWatch() throws ApiException {
               null,
               null,
               null,
-              true,
+              null,
```
jakhani, line 125: What was this true parameter?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: ack

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
utk-12: This is ideally not even needed since this newer client handles this for us. However if we do decide to place it , I shall include it via a config.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
utk-12: It was for watch. I shall fix this.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: That looks like a mistake. Fixed

svnarumugam (commented):
Thanks for making this change Utkarsh! 
1. Could you please add more context as we discussed offline?
2. Can we test the functional use cases manually, and call out the use cases tested?
3. Please call out the contract changes like RestartPolicyEnum, and the need for OffsetDateTime to inform the readers of this PR.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -75,9 +85,23 @@ public KubernetesWatch(ApiClient apiClient,
...
+            DEFAULT_KUBERNETES_WATCH_CONNECT_TIMEOUT);
+    this.readTimeout =
+        azkProps.getInt(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_READ_TIMEOUT,
+            DEFAULT_KUBERNETES_WATCH_READ_TIMEOUT);
+    // Setting read and call timeouts. So pod watch will be reinitialized.
     OkHttpClient httpClient =
-        this.apiClient.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build();
+        this.apiClient.getHttpClient().newBuilder()
+            .protocols(Arrays.asList(Protocol.HTTP_2,Protocol.HTTP_1_1))
+            .connectTimeout(this.connectTimeout, TimeUnit.SECONDS)
```
sshardool, line 101: Is there a reason whey the client version upgrade needs to be clubbed with the changes for adding these timeouts ?
Ideally the timeouts can be added first (as they seem more urgent) and followed up with the version upgrade, the version upgrade changes will take a little more time to review as there are quite a few of those.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -723,6 +723,12 @@ public static class ContainerizedDispatchManagerProperties {
     public static final String KUBERNETES_WATCH_ENABLED = KUBERNETES_WATCH_PREFIX + "enabled";
     public static final String KUBERNETES_WATCH_EVENT_CACHE_MAX_ENTRIES =
         KUBERNETES_WATCH_PREFIX + "cache.max.entries";
+    public static final String KUBERNETES_WATCH_CALL_TIMEOUT = KUBERNETES_WATCH_PREFIX + "call"
```
djaiswal83, line 726: nitpick : Please put the suffixes in one line. It adds to code readability and search. Applies to line 731 as well.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1SpecBuilder.java
```diff
@@ -50,7 +52,7 @@ public class AzKubernetesV1SpecBuilder {
             LoggerFactory.getLogger(AzKubernetesV1SpecBuilder.class);
     private static final String AZ_CLUSTER_KEY = "AZ_CLUSTER";
     private static final String AZ_CONF_VERSION_KEY = "AZ_CONF_VERSION";
-    private static final String DEFAULT_RESTART_POLICY = "Never";
+    private static final String DEFAULT_RESTART_POLICY = "NEVER";
```
djaiswal83, line 55: Please remove this as it is not used anywhere.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1212,20 +1233,23 @@ private void deletePod(final int executionId) throws ExecutorManagerException {
...
-          new V1DeleteOptions());
-      logger.info("ExecId: {}, Action: Service Deletion, Service Name: {}, code: {}, message: {}",
-          executionId,
-          serviceName,
-          deleteResult.getCode(),
-          deleteResult.getMessage());
+      // Using GenericKubernetesApi due to a Known issue in K8s Java client and OpenAPIv2:
+      // See more here: https://github.com/kubernetes-client/java/wiki/6.-Known-Issues
+
+      GenericKubernetesApi<V1Service, V1ServiceList> serviceClient =
```
djaiswal83, line 1239: nit : final

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1212,20 +1233,23 @@ private void deletePod(final int executionId) throws ExecutorManagerException {
...
+          new GenericKubernetesApi<>(V1Service.class, V1ServiceList.class, "",
+              "v1", "services", this.client);
+      V1Service deletedService = serviceClient.delete(
```
djaiswal83, line 1242: nit : final

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1212,20 +1233,23 @@ private void deletePod(final int executionId) throws ExecutorManagerException {
...
+          this.namespace, serviceName).throwsApiException().getObject();
+      if (deletedService != null) {
+        logger.info(
```
djaiswal83, line 1245: Is this the more common case? If not, then swap the if else blocks.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -180,7 +180,7 @@ private void extractConditions() {
     Map<String, V1PodCondition> conditionMap = new HashMap<>();
     this.podConditions.stream().forEach(
         condition ->
-            conditionMap.put(condition.getType(), condition));
+            conditionMap.put(String.valueOf(condition.getType()), condition));
```
djaiswal83, line 183: Instead of converting to string, use the TypeEnum as key instead of String. This is likely more efficient and saves the conversion both here and at line 228.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -48,6 +52,12 @@ public class KubernetesWatch implements ContainerizedWatch {
   private final CoreV1Api coreV1Api;
   private final PodWatchParams podWatchParams;
   private final Thread watchRunner;
+  private final int DEFAULT_KUBERNETES_WATCH_CALL_TIMEOUT = 86400;
```
djaiswal83, line 55: Please add a comment stating the units of the timeouts.

djaiswal83 (changes requested):
Thanks for working on the client upgrade. Overall the code looks good. I have left a few comments. Please address those.

az-core/src/main/java/azkaban/Constants.java
sshardool: +1

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -723,6 +723,12 @@ public static class ContainerizedDispatchManagerProperties {
...
+        + ".timeout";
+    public static final String KUBERNETES_WATCH_CONNECT_TIMEOUT = KUBERNETES_WATCH_PREFIX +
```
sshardool, line 728: For all of these constants and literals please add the units in the suffix. For example
```
KUBERNETES_WATCH_CONNECT_TIMEOUT_SECONDS = KUBERNETES_WATCH_PREFIX +  "connect.timeout.seconds";
```

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1212,20 +1233,23 @@ private void deletePod(final int executionId) throws ExecutorManagerException {
...
-          null,
-          null,
-          new V1DeleteOptions());
-      logger.info("ExecId: {}, Action: Service Deletion, Service Name: {}, code: {}, message: {}",
-          executionId,
-          serviceName,
-          deleteResult.getCode(),
-          deleteResult.getMessage());
+      // Using GenericKubernetesApi due to a Known issue in K8s Java client and OpenAPIv2:
+      // See more here: https://github.com/kubernetes-client/java/wiki/6.-Known-Issues
```
sshardool, line 1237: Refer to this github issue directly in the comment: https://github.com/kubernetes-client/java/issues/86

The currently linked page will likely keep changing with different versions of the kubernetes client (for example if the issue actually gets fixed)

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -437,13 +440,24 @@ private Set<Integer> getStaleContainers(final Duration containerValidity) throws
       // Select pods only from current Azkaban cluster and namespace
       final String label =
           CLUSTER_LABEL_NAME + "=" + this.clusterName + "," + APP_LABEL_NAME + "=" + POD_APPLICATION_TAG;
-      final V1PodList podList= this.coreV1Api.listNamespacedPod(this.namespace, null,
-          null, null, null, label,
-          null, null, null, null);
+      final V1PodList podList= this.coreV1Api.listNamespacedPod(
```
sshardool, line 443: What's the new argument added to this call as part of the api version change ?

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
sshardool: Or preferably suffix the constant name with "_SECONDS"

sshardool: Also, write the constant values as `days * hours * mins * seconds` to make it more readable, something like `24 * 60 * 60`

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -75,9 +85,23 @@ public KubernetesWatch(ApiClient apiClient,
...
+    this.callTimeout =
+        azkProps.getInt(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_CALL_TIMEOUT,
+        DEFAULT_KUBERNETES_WATCH_CALL_TIMEOUT);
+    this.connectTimeout =
+        azkProps.getInt(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_CONNECT_TIMEOUT,
+            DEFAULT_KUBERNETES_WATCH_CONNECT_TIMEOUT);
+    this.readTimeout =
+        azkProps.getInt(ContainerizedDispatchManagerProperties.KUBERNETES_WATCH_READ_TIMEOUT,
+            DEFAULT_KUBERNETES_WATCH_READ_TIMEOUT);
+    // Setting read and call timeouts. So pod watch will be reinitialized.
```
sshardool, line 97: Expand this comment elaborate the needs to timeouts a little more. I.e reflect that there are cluster specific setups where we would like to force the reinitialization to handle silent connections drops during the blocking reads.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
```diff
@@ -114,7 +139,8 @@ protected void initializePodWatch() throws ApiException {
               null,
               null,
               null,
-              true,
+              null,
```
sshardool, line 142: what's the change in this method signature? please double check if the parameter values are aligning with the method in the older client version.

sshardool (commented):
Thanks a lot for taking this up!

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
utk-12: There's a new parameter called resourseVersionMatch which would also be null since the resourceVersion is null. Since we're passing one more parameter here, the true for the watch boolean is moved ahead.

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
utk-12: I've changed it to .getname() which is essentially pulling it from an Enum type.

utk-12: Changing the type of the entire map, could be part of another change ? Since it's not relevant to this upgrade.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: ack

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: resourseVersionMatch

djaiswal83 (approved):
Looks good to me.

azkaban-common/src/main/java/azkaban/executor/container/watch/KubernetesWatch.java
utk-12: Note:
To ensure our client's watch connection is not stale and is properly timed out/ refreshed, this upgrade isn't necessary. However, while we are upgrading the client to ensure we don't break compatibility in future K8s version upgrades, this timeout is introduced at the same time. We could possibly separate this change.

sshardool (approved):
Thanks for taking care of these. Due to the nature of the changes, please give them extra soak-time on test/dev setups.

jakhani (approved):
Thank you for all the hard work to fix this issue.

PR #3078, opened 2022-04-18
Check validity negative before cleaning up stale containers

byjiang1996:
Bug fix. Validity should be positive when used to determine if a container is stale or not.


jakhani (approved):

djaiswal83 (approved):

hqin2016 (approved):
Thanks for fixing this bug.

PR #3081, opened 2022-04-21
Removing EXECUTION_STOPPED from container cleanup manager states as i…

utk-12:
…t is a final state. Reverting 32637fdbecf81 due to this change as it's not required.
We have all the flows in cleanUpStaleFlows under EXECUTION_STOPPED being attempted to be deleted on K8s. This is inaccurate as EXECUTION_STOPPED is a final state. We have FlowStatusManagerListener.lambda$onPodCompleted  & cleanUpStaleContainers to take care of completed Pods. 

Having this in place floods our logs with 404 Not Found APiexceptions from K8s as these pods don't exist.


hqin2016 (dismissed):
Thanks for making the change!

hqin2016 (approved):
LGTM

PR #3092, opened 2022-05-11
Update TriggerManager.java

haolijun:
If the wait time is zero, and not wake up via notify(),  the thread will to wait forever.


haolijun (commented):
Due to this problem, the task is not distributed

PR #3094, opened 2022-05-16
Adding extra logging and retry logic for fetching hadoop security tokens

utk-12:


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/HadoopSecurityManager_H_2_0.java
```diff
@@ -203,10 +212,14 @@ private void fetchJobTrackerToken(final String userToProxyFQN,
       final JobConf jobConf = new JobConf();
       final JobClient jobClient = new JobClient(jobConf);
       logger.info("Pre-fetching JT token from JobTracker");
-      Token<DelegationTokenIdentifier> mrdt;
+      Token<DelegationTokenIdentifier> mrdt = null;
```
djaiswal83, line 215: mrDelegationToken instead of mrdt

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -130,6 +130,12 @@ public class Constants {
   // Used by UserManager for password validation (to tell apart real passwords from auth codes).
   // Empirically, passwords are shorter than this, and ACs are longer:
   public static final int OAUTH_MIN_AUTHCODE_LENGTH = 80;
+  // Hadoop Security Manager retry policy constants
+  public static final int HSM_MAX_RETRY_ATTEMPTS = 5;
```
djaiswal83, line 134: Can you please move these constants to HSM2 code it self? They are only used in that code only. Having them declared within the class should be good enough.

djaiswal83 (changes requested):
Thanks for working on this change. The code looks good to me.
I have one comment about placement of constants, please address it.

djaiswal83 (approved):
Thanks for addressing the comments. LGTM

PR #3095, opened 2022-05-16
Bug fix: when canceling flows, we should send out Killed events corre…

byjiang1996:
…ctly

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -526,9 +526,6 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
         final Pair<ExecutionReference, ExecutableFlow> pair = unfinishedFlows
             .get(exFlow.getExecutionId());
         handleCancelFlow(pair.getFirst(), exFlow, userId);
-        // Emit FLOW_FINISHED event
-        this.fireEventListeners(Event.create(exFlow,
-            EventType.FLOW_FINISHED, new EventData(exFlow)));
       } else {
         final ExecutorManagerException eme = new ExecutorManagerException("Execution "
             + exFlow.getExecutionId() + " of flow " + exFlow.getFlowId() + " isn't running.");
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -560,6 +557,8 @@ protected void handleCancelFlow(ExecutionReference executionReference,
       ExecutionControllerUtils.finalizeFlow(this.executorLoader, this.alerterHolder,
           executableFlow, "Cancel action has been called but the flow is unreachable.", null,
           finalizingStatus);
+      this.fireEventListeners(Event.create(executableFlow,
```
hqin2016, line 560: can you add comment to the code changes?

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -574,6 +573,8 @@ protected void handleCancelFlow(ExecutionReference executionReference,
       final String finalizingReason = "Unable to gracefully kill the flow execution.";
       ExecutionControllerUtils.finalizeFlow(this.executorLoader, this.alerterHolder,
           executableFlow, finalizingReason, e, finalizingStatus);
+      this.fireEventListeners(Event.create(executableFlow,
```
hqin2016, line 576: same as above, comment is needed to describe the code change.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: Done. Thx!

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: Done. Thx!

hqin2016 (approved):
Thanks for the bug fix.

PR #3096, opened 2022-05-17
Fix the pod status mismatch when pod is running but container is not ready

NancyXie2022:
Since the pod goes into Running status very quick in AKS, so when the pod is running, the container is not ready yet, which cause https://github.com/azkaban/azkaban/blob/master/azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java#L578 fall into AZ_POD_UNEXPECTED.
In this case,  it make more sense to let it move to AZ_POD_APP_CONTAINERS_STARTING status.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
@@ -366,6 +366,7 @@ public class AzPodStatusExtractor {
       logger.debug("ContainerStarting false as container status is null or empty");
       return false;
     }
+    // All containers shouldn't be waiting and container should be ready
     boolean allContainersWaiting =
         containerStatuses.stream().allMatch(status -> status.getStarted() == false);
     if (!allContainersWaiting && this.containersReadyConditionStatus.isPresent() &&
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -368,8 +368,9 @@ private boolean checkForAzPodAppContainerStarting() {
     }
     boolean allContainersWaiting =
         containerStatuses.stream().allMatch(status -> status.getStarted() == false);
-    if (!allContainersWaiting) {
-      logger.debug("ContainerStarting false as all containers are not waiting");
+    if (!allContainersWaiting && this.containersReadyConditionStatus.isPresent() &&
```
hqin2016, line 371: Could you add some comments?

djaiswal83 (approved):

azkaban-common/src/main/java/azkaban/executor/container/watch/AzPodStatusExtractor.java
```diff
@@ -366,10 +366,12 @@ private boolean checkForAzPodAppContainerStarting() {
       logger.debug("ContainerStarting false as container status is null or empty");
       return false;
     }
+    // All containers shouldn't be waiting and container should be ready
```
sshardool, line 369: This should probably say:
`All containers should be waiting and ContainersReady condition should not be true`

Otherwise the change to the condition itself seems reasonable. Would be good to know more about the sequence of events which led to a problem.

sshardool (commented):
The changes look good. 

Please add a unit test for this within `KubernetesWatchTests` to avoid future regression, that can be done with a subsequent PR.

PR #3027, opened 2022-01-05
Log pod details

amielkelinkedin:
This change allows the logging of Events and Pod Logs for specifics types of failures to the UI before pod deletion.

PR #3097, opened 2022-05-19
Pod logging feature toggle changes

amielkelinkedin:
Please reference #3027, this is the same story which was reverted due to errors discovered in testing, resolved in #3058.

This story aims to log pod details from a K8s pod that does not complete successfully, giving the user useful information and insights into what might have gone wrong.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
+++ b/azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
@@ -23,15 +23,23 @@ import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import com.google.common.collect.ImmutableMap;
 import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.log4j.Logger;
 
 @Singleton
 public class JdbcExecutorLoader implements ExecutorLoader {
+
+  private static final Logger logger = Logger.getLogger(JdbcExecutorLoader.class);
 
   private final ExecutionFlowDao executionFlowDao;
   private final ExecutorDao executorDao;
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ notifications:
 env:
   global:
     # Reduce the memory pressure on Travis CI to reduce build failures.
-    - GRADLE_OPTS="-Xms128m"
```
hqin2016, line 32: I think this is not going to help. Try to run rebuild multiple times.

amielkelinkedin: Thank you! I'll try that!

djaiswal83: lets not change this. The tests have some flakiness due to which they fail sometimes.

azkaban-common/src/main/java/azkaban/executor/JdbcExecutorLoader.java
```diff
@@ -299,6 +307,35 @@ public void uploadLogFile(final int execId, final String name, final int attempt
...
 
+  @Override
+  public void appendLogs(final int execId, final String name, final String logs) throws
+      ExecutorManagerException {
+    if (StringUtils.isBlank(logs)) {
+      return;
+    }
+    File tempFile = null;
+    try {
+      tempFile = File.createTempFile("appendLogs-" + execId + "-" + name, ".tmp");
```
djaiswal83, line 318: nit : use Final wherever applicable.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -172,7 +172,13 @@ public void cleanUpContainersInTerminalStatuses() {
       if (!activeFlows.contains(executionId)) {
         logger.info("Cleaning up the stale pod and service for finished execution: {}",
             executionId);
-        deleteContainerQuietly(executionId);
+        try {
+          this.containerizedImpl.deletePod(executionId);
```
djaiswal83, line 176: Why change this? Why not keep the try catch but keep the deleteContainerQuietly API?

djaiswal83 (changes requested):
Left a few minor comments.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
amielkelinkedin: I changed deleteContainerQuietly to accept a Flow Object instead of an executionId. This is so that we can interrogate the status inside the flow object down the stack. In general we should be passing as much info down the stack as we can, IMHO. Here we are relying on our own naming convention and essentially making a hard contract on something that we impose. So we have some options on the implementation:

1. We duplicate some code with a path that takes an executionId and a path that takes a Flow object. This will look a little funny, but it is workable. As part of this change we are already duplicating deletePod.
2. We change the logic here to fetch a flow for each executionId. The issue with that is then we are making a lot of calls to the DB essentially on startup of our application. This will make it so it takes longer for the WS to come up.
3. We do something like I've done here.

I think the best option might be the first one, in all honesty. The issue with what we've done here is that if an exception is always thrown for the pod then the service will never be cleaned up. In structuring the code this way I am kind of assuming that an error will not be thrown for a delete call immediately after kubernetes tells us something is there. Also, pods that are deleted from this part of the workflow won't have logs show up in the UI! 

Ultimately, as we've discussed before, I think we're making a mistake by trying to manage an orchestrator like kubernetes. I think it would make the most sense to give k8s a deployment definition and have it manage the pod/service lifecycle. There is a lot of code written by teams of people to manage edge cases we haven't even thought of. By taking on all of these responsibilities ourselves we open ourselves up to a whole  host of bugs/issues. I believe we are beginning to see some of those issues come to a head here.

.travis.yml
amielkelinkedin: I was looking into how we might be able to fix this, but I didn't get very far. We're running into many different issues be it OOM or otherwise. Will leave it as it currently is for now. Thank you!

PR #3098, opened 2022-05-19
Remove ConsoleAppender from HadoopJavaJobRunnerMain 

djaiswal83:
Remove ConsoleAppender from HadoopJavaJobRunnerMain  as it is not needed anymore.

It breaks compatibility with applications using log4j2.


byjiang1996:
I also see ConsoleAppender in JavaJobRunnerMain.java. Do we need to remove it as well?

byjiang1996 (approved):
LGTM. thx!

PR #3099, opened 2022-05-24
Migrate job callback from executor module to common module

byjiang1996:
Prerequisite for triggering job callback in web server


azkaban-common/src/main/java/azkaban/jobcallback/JobCallbackManager.java
```diff
@@ -166,76 +172,79 @@ private void processJobCallOnFinish(final Event event) {
...
     }
   }
 
   private String getAzkabanHostName(final Props props) {
-    final String baseURL = props.get(JobRunner.AZKABAN_WEBSERVER_URL);
+    final String baseURL = props.get(AZKABAN_WEBSERVER_URL);
     try {
-      String hostName = InetAddress.getLocalHost().getHostName();
+      String hostName =
+          props.getString("jetty.hostname", "localhost") + ":" + (props.getBoolean(JETTY_USE_SSL
```
hqin2016, line 245: Can you use constants instead of strings?

byjiang1996: Good point. Done.

hqin2016 (approved):
LGTM

PR #3100, opened 2022-05-24
Use ".override" instead of "param.override" to remove c14n properties from pig properties

djaiswal83:
Use ".override" instead of "param.override" to remove c14n properties from pig properties


hqin2016 (commented):
LGTM

hqin2016 (approved):

PR #3101, opened 2022-05-25
Trigger Job Callback in web server if flow is killed/execution_stoppe…

byjiang1996:
…d before FlowContainer is up in pod

Otherwise, we will miss Job Callback events.

1. failEverything is the common util function to change node status when killing / execution_stopping a flow
2. When we change a node status, we should fire JOB_STARTED or JOB_FINISHED events accordingly. In order to fire those events, we should have EventHandler to fire events -> pass EventHandler to failEverything function.
2.1. If a node status is changed to Killed when it has not started running, we should fire JOB_STARTED event
2.2. If a node status is changed to a FINISHED status, we should fire JOB_FINISHED event
3. When events like JOB_STARTED or JOB_FINISHED are fired, we should have JobCallManager's event listener associated with the EventHandler. In addition, we should have node props set up appropriately before the events go to JobCallManager's listener. ProjectManager is the class to load props as shown in ProjectManagerServlet -> pass ProjectManager to failEverything function call.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -368,7 +368,7 @@ public class ExecutionControllerUtils {
       final ProjectManager projectManager, final ExecutableFlow exFlow,
       final Status finalFlowStatus) {
     final long time = System.currentTimeMillis();
-    final Map<ExecutableNode, Status> node2origStatus = new HashMap<>();
+    final Map<ExecutableNode, Status> nodeToOrigStatus = new HashMap<>();
     final Queue<ExecutableNode> queue = new LinkedList<>();
     queue.add(exFlow);
     // Traverse the DAG and fail every node that's not in a terminal state
@@ -390,7 +390,7 @@ public class ExecutionControllerUtils {
             continue;
             // case UNKNOWN:
           case READY:
-            node2origStatus.put(node, node.getStatus());
+            nodeToOrigStatus.put(node, node.getStatus());
             if (finalFlowStatus == Status.KILLED) {
               // if the finalFlowStatus is KILLED, then set the sub node status to KILLED.
               node.setStatus(Status.KILLED);
@@ -403,7 +403,7 @@ public class ExecutionControllerUtils {
             }
             break;
           default:
-            node2origStatus.put(node, node.getStatus());
+            nodeToOrigStatus.put(node, node.getStatus());
             // Set the default job/node status as the finalFlowStatus
             node.setStatus(finalFlowStatus);
             break;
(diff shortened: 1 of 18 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -318,8 +350,25 @@ public static String[] getFinalizeFlowReasons(final String reason, final Throwab
...
+   * @param eventHandler the event handler to fire job events
+   * @param projectManager the project manager to load job props from DB
+   * @param exFlow the executable flow
+   * @param finalFlowStatus Status to be set if the flow is not in one of the "finished" statues.
+   */
+  public static void failEverything(final EventHandler eventHandler,
+      final ProjectManager projectManager, final ExecutableFlow exFlow,
+      final Status finalFlowStatus) {
     final long time = System.currentTimeMillis();
+    final Map<ExecutableNode, Status> node2origStatus = new HashMap<>();
```
djaiswal83, line 371: nit: consider naming : nodeToOrigStatus

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -367,6 +418,39 @@ public static void failEverything(final ExecutableFlow exFlow, final Status fina
       }
     }
 
+    // Fire correct JOB_STARTED and JOB_FINISHED events.
+    if (eventHandler != null && projectManager != null) {
+      Project project = projectManager.getProject(exFlow.getProjectId());
```
djaiswal83, line 423: nit : Use final wherever applicable.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -367,6 +418,39 @@ public static void failEverything(final ExecutableFlow exFlow, final Status fina
...
+      Flow flow = project.getFlow(exFlow.getFlowId());
+      for (Entry<ExecutableNode, Status> entry: node2origStatus.entrySet()) {
+        ExecutableNode node = entry.getKey();
+        Status origStatus = entry.getValue();
+
+        // Fill in job props by following the logic in ProjectManagerServlet.
+        if (node.getInputProps() == null) {
+          node.setInputProps(projectManager.getJobOverrideProperty(project, flow, node.getId(),
```
djaiswal83, line 431: Why reading from file here? As discussed earlier, this is already in ImmutableFlowProps

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -367,6 +418,39 @@ public static void failEverything(final ExecutableFlow exFlow, final Status fina
...
+              node.getJobSource()));
+        }
+        if (node.getInputProps() == null) {
+          node.setInputProps(projectManager.getProperties(project, flow, node.getId(),
```
djaiswal83, line 435: Same as above.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -367,6 +418,39 @@ public static void failEverything(final ExecutableFlow exFlow, final Status fina
       }
     }
 
+    // Fire correct JOB_STARTED and JOB_FINISHED events.
```
djaiswal83, line 421: Why are we doing it here? Isnt it supposed to happen in JobRunner/FlowRunner?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
byjiang1996: ImmutableFlowProps is not persisted into memory but gets GC after uploading a project zip (in transient DirectoryYamlFlowLoader and DirectoryFlowLoader objects). And it cannot cover “job override properties” overridden via api call.

I think my current implementation is fine. It is referring to the design as in this page: https://<clusterName>/manager?project=<projectName>&flow=<flowName>&job=<jobName>. (specifically ProjectManagerServlet.java)

djaiswal83 (approved):
Thanks for addressing the comments. The code looks good to me.

PR #3102, opened 2022-05-26
Enable kafka log4j appender for both Azkaban flowLogs and jobLogs and…

byjiang1996:
… modernize the old kafka appender design.

Refer to docs/howTo.rst file for more detailed explanations.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -291,9 +291,14 @@ public class Constants {
     public static final String HISTORY_SERVER_JOB_URL = "azkaban.server.external.history_server_job_url";
     public static final String SPARK_HISTORY_SERVER_JOB_URL = "azkaban.server.external.spark_history_server_job_url";
 
-    // Configures the Kafka appender for logging user jobs, specified for the exec server
-    public static final String AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST = "azkaban.server.logging.kafka.brokerList";
-    public static final String AZKABAN_SERVER_LOGGING_KAFKA_TOPIC = "azkaban.server.logging.kafka.topic";
+    // Configures the Kafka appender for logging user jobs and flows.
+    public static final String AZKABAN_LOGGING_KAFKA_ENABLED = "azkaban.logging.kafka.enabled";
+    public static final String AZKABAN_LOGGING_KAFKA_CLASS_PARAM = "azkaban.logging.kafka.class";
+    public static final String AZKABAN_LOGGING_KAFKA_BROKERS = "azkaban.logging.kafka.brokers";
+    public static final String AZKABAN_JOB_LOGGING_KAFKA_TOPIC = "azkaban.job.logging.kafka.topic";
+    public static final String AZKABAN_FLOW_LOGGING_KAFKA_TOPIC = "azkaban.flow.logging.kafka.topic";
+    public static final String AZKABAN_LOGGING_KAFKA_SCHEMA_REGISTRY_URL =
+        "azkaban.logging.kafka.schema.registry.url";
 
     public static final String IS_METRICS_ENABLED = "azkaban.is.metrics.enabled";
 
@@ -545,9 +550,6 @@ public class Constants {
   }
 
   public static class JobProperties {
-
-    // Job property that enables/disables using Kafka logging of user job logs
-    public static final String AZKABAN_JOB_LOGGING_KAFKA_ENABLE = "azkaban.job.logging.kafka.enable";
 
     /*
      * this parameter is used to replace EXTRA_HCAT_LOCATION that could fail when one of the uris is not available.
--- a/azkaban-common/src/main/java/azkaban/utils/KafkaLog4jUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/KafkaLog4jUtils.java
@@ -24,7 +24,7 @@ public class KafkaLog4jUtils {
   }
 
   private static KafkaLog4jAppender getAzkabanKafkaLog4jAppender(final Props props,
-      final String execId, final String flowId, final String jobId, final String jobAttempt,
+      final String execId, final String name, final String jobAttempt,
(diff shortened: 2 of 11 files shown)


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -291,9 +291,14 @@ public static class ConfigurationKeys {
     public static final String HISTORY_SERVER_JOB_URL = "azkaban.server.external.history_server_job_url";
     public static final String SPARK_HISTORY_SERVER_JOB_URL = "azkaban.server.external.spark_history_server_job_url";
 
-    // Configures the Kafka appender for logging user jobs, specified for the exec server
-    public static final String AZKABAN_SERVER_LOGGING_KAFKA_BROKERLIST = "azkaban.server.logging.kafka.brokerList";
-    public static final String AZKABAN_SERVER_LOGGING_KAFKA_TOPIC = "azkaban.server.logging.kafka.topic";
+    // Configures the Kafka appender for logging user jobs and flows.
+    public static final String AZKABAN_LOGGING_KAFKA_ENABLED = "azkaban.logging.kafka.enabled";
+    public static final String AZKABAN_LOGGING_KAFKA_CLASS_PARAM = "azkaban.logging.kafka.class";
+    public static final String AZKABAN_LOGGING_KAFKA_BROKERS = "azkaban.logging.kafka.brokers";
```
hqin2016, line 297: Is this broker different from event reporter broker?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -291,9 +291,14 @@ public static class ConfigurationKeys {
...
+    public static final String AZKABAN_JOB_LOGGING_KAFKA_TOPIC = "azkaban.job.logging.kafka.topic";
+    public static final String AZKABAN_FLOW_LOGGING_KAFKA_TOPIC = "azkaban.flow.logging.kafka.topic";
+    public static final String AZKABAN_LOGGING_KAFKA_SCHEMA_REGISTRY_URL =
```
hqin2016, line 300: same as the above.

azkaban-common/src/main/java/azkaban/utils/KafkaLog4jUtils.java
```diff
@@ -0,0 +1,94 @@
...
+
+      final Constructor<?> kafkaLog4jAppenderClassConstructor =
+          kafkaLog4jAppenderClass.getConstructor();
+      KafkaLog4jAppender kafkaLog4jAppender =
+          (KafkaLog4jAppender) kafkaLog4jAppenderClassConstructor.newInstance();
+
+      String[] origIds = new String[] {execId, flowId, jobId, jobAttempt};
+      String[] nonNullIds = Arrays.stream(origIds).filter(Objects::nonNull).toArray(String[]::new);
+
+      kafkaLog4jAppender.setBrokerList(props
```
hqin2016, line 50: It might be better to provide a default value.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -365,33 +363,19 @@ private void createLogger() {
...
-            removeAppender(this.kafkaAppender);
-            this.flowLogger.error("Failed to create Kafka appender for job " + this.jobId, e);
-          }
-        } else {
-          this.flowLogger.info(
-              "Kafka appender not created as brokerlist or topic not provided by executor server");
+      if (this.azkabanProps.getBoolean(Constants.ConfigurationKeys.AZKABAN_LOGGING_KAFKA_ENABLED, false)) {
+        this.kafkaLog4jAppender =
+            KafkaLog4jUtils.getAzkabanJobKafkaLog4jAppender(this.azkabanProps,
+                String.valueOf(this.executionId),  this.node.getParentFlow().getFlowId(),
```
hqin2016, line 369: Is 'this.node.getParentFlow().getFlowId()' valid for nested flow?

az-core/src/main/java/azkaban/Constants.java
byjiang1996: The meaning is the same but values can be the same or different. (in our case, values are different)

az-core/src/main/java/azkaban/Constants.java
byjiang1996: The meaning is the same but values can be the same or different. (in our case, value is the same)

azkaban-common/src/main/java/azkaban/utils/KafkaLog4jUtils.java
byjiang1996: We cannot give a default value b/c brokerList is the key to kafka (can be treated as the kafka cluster name; we shouldn't give an arbitrary value here).
If no value is given in config, it will fallback to the catch block and return null which will be handled nicely in FlowRunner and JobRunner.

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
byjiang1996: It is an excellent question! Thank you so much for bringing this up!

I find when we upload or query logs from mysql, we are using this.node.getNestedId() only for the query key. 'this.node.getParentFlow().getFlowId()' is not enough for nested flows but this.node.getNestedId() is the best representative for a node.

Also, I changed the getAzkabanJobKafkaLog4jAppender interface accordingly.

az-core/src/main/java/azkaban/Constants.java
hqin2016: Thanks, just a little suggestion, we could consider reusing the broker and schema registry in the azkaban config.

hqin2016 (approved):
LGTM, thanks for enabling the log streaming to Kakfa.

az-core/src/main/java/azkaban/Constants.java
byjiang1996: Yep, that makes sense ^. But actually we are using different broker for event reporter and for logging

PR #3103, opened 2022-06-01
Use Azkaban external (user facing) web server url if configured

ypadron-in:
when generating execution related urls in runtime properties.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
@@ -977,10 +977,13 @@ public class JobRunner extends JobRunnerBase implements Runnable {
   }
 
   /**
-   * Build Azkaban frontend base url. Use external configs
-   * ({@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME},
-   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_SSL_PORT}) if set
-   * else {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_URL}
+   * Build Azkaban frontend base url. Use external or user facing configs (for the cases where the
+   * Web server is behind a proxy for example) if set else use property
+   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_URL}.
+   * User facing access is configured with properties:
+   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME},
+   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_SSL_PORT} and
+   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_PORT}
    */
   private String getAzkabanWebServerURL() {
     String azkabanWebServerURL = this.azkabanProps.get(AZKABAN_WEBSERVER_URL);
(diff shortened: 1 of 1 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -969,6 +976,43 @@ private void insertJobMetadata() {
...
   }
 
+  /**
+   * Build Azkaban frontend base url. Use external configs
+   * ({@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_HOSTNAME},
+   * {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_EXTERNAL_SSL_PORT}) if set
+   * else {@link azkaban.Constants.ConfigurationKeys#AZKABAN_WEBSERVER_URL}
+   */
+  private String getAzkabanWebServerURL() {
+    String azkabanWebServerURL = this.azkabanProps.get(AZKABAN_WEBSERVER_URL);;
```
byjiang1996, line 986: Remove extra ;

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
```diff
@@ -969,6 +976,43 @@ private void insertJobMetadata() {
         StringUtils.join2(this.node.getInNodes(), ","));
   }
 
+  /**
+   * Build Azkaban frontend base url. Use external configs
```
byjiang1996, line 980: Could you please elaborate more on why AZKABAN_WEBSERVER_URL does not work by default here?

ypadron-in: I think there is some redundancy of properties in the code to refer to the web server host. For example there is: `azkaban.server.hostname`, `jetty.hostname` and `azkaban.webserver.url`. I'm not sure what's the difference between them honestly and maybe we can get rid of `azkaban.webserver.url` at least since that one was only used in JobRunner class, the 2 other are more widely used throughout the code but it would be good to reduce usage to one only.

For user facing server url (which might be different from hostname and above properties if the web server sits behind a proxy for example) the code suggests to use `azkaban.webserver.external_hostname`, `azkaban.webserver.external_ssl_port` and `azkaban.webserver.external_port`, that is why we're using them here.

byjiang1996: From what I know:
1. jetty.hostname is the de-facto web server name but this config only exists in web server but not exec/pod
2. azkaban.server.hostname: it is either webserver/execserver hostname or pod name (fc-dep-<clustername>-<execid>) which is used to uniquely identity the physical machine.

Yep. Sorry I didn't make it clear. Could you please add 
`For user facing server url (which might be different from hostname and above properties if the web server sits behind a proxy for example) the code suggests to use azkaban.webserver.external_hostname, azkaban.webserver.external_ssl_port and azkaban.webserver.external_port, that is why we're using them here.`
into the code comment?

ypadron-in: `jetty.hostname` should be valid for both Web and Executor servers as there is a jetty web server running in the Executor.
IMO `azkaban.server.hostname` should not be the pod name as that is not a hostname but a logical application related name right? Based on code `azkaban.server.hostname` might be equivalent to Web server prop `jetty.hostname` in Executor server.
Both properties has `hostname` in their name which makes it difficult to distinguish their meaning and usage.

byjiang1996 (approved):

azkaban-exec-server/src/main/java/azkaban/execapp/JobRunner.java
byjiang1996: Yeah. I understand the azkaban.server.hostname part. But that's how we did there LOL.
"Both properties has hostname in their name which makes it difficult to distinguish their meaning and usage.": can't agree more. Those props are pretty confusing!!!

ypadron-in:
Closing as this is no longer needed.

PR #3104, opened 2022-06-03
Set loglevel to INFO for HadoopJavaJobRunnerMain and JavaJobRunnerMain.

djaiswal83:


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJobRunnerMain.java
```diff
@@ -94,6 +94,7 @@ public void run() {
     // Separate try catch for logger
     try {
       _logger = Logger.getRootLogger();
+      _logger.setLevel(Level.INFO);
```
amielkelinkedin, line 97: Do we want to set this from configuration such that it can be changed without a code change? We have a lot of things in configs already, not sure if it would be useful to add this as well, but thought I'd ask the question.

jakhani: Plus one. This should come from configuration.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/JavaJobRunnerMain.java
```diff
@@ -88,6 +89,7 @@ public void run() {
       String propsFile = System.getenv(ProcessJob.JOB_PROP_ENV);
 
       _logger = Logger.getRootLogger();
+      _logger.setLevel(Level.INFO);
```
jakhani, line 92: Same as above. This info should come from config.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJavaJobRunnerMain.java
amielkelinkedin: Agreed to perhaps address this in a future change as this is blocking and our logging strategy might change in the near future anyway, rendering this obsolete.

amielkelinkedin (approved):
Approving without injecting from configs as this is blocking.

djaiswal83:
> Approving without injecting from configs as this is blocking.

thank you.

PR #3106, opened 2022-06-07
Split ExecutorLoader into ExecutorLoader and ExecutionLogsLoader, and pluginize ExecutionLogsLoader

byjiang1996:
New log read logic in AbstractExecutorManagerAdapter.java:
1. Read from local file if execution is running
2. Read from MySQL/NearlineExecutionLogsLoader for nearline
3. Read from external logs loader / OfflineExecutionLogsLoader emitted by Kafka log4j appender if pod/execution is crashed OR no logs available in NearlineExecutionLogsLoader/MySQL
3.1. Inject dependencies NearlineExecutionLogsLoader and OfflineExecutionLogsLoader in AzkabanCommonModule
4. Introduce new API endpoint to turn on/off OfflineExecutionLogsLoader in web server at runtime: /executor?ajax=updateProp&propType=general&subType= enableOfflineLogsLoader&val=true

Since we have OfflineExecutionLogsLoader ready, we will reduce the default MySQL retention period from 12 weeks to 30 days.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -15,7 +15,7 @@
  */
 package azkaban.executor;
 
-import static azkaban.Constants.LogConstants.NEARLINE_LOGS;
+import static azkaban.Constants.ConfigurationKeys.AZKABAN_OFFLINE_LOGS_LOADER_ENABLED;
 
 import azkaban.Constants;
 import azkaban.Constants.ConfigurationKeys;
@@ -49,7 +49,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import javax.inject.Named;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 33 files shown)


azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -438,8 +460,18 @@ protected LogData getFlowLogData(final ExecutableFlow exFlow, final int offset,
...
     } else {
-      return this.executorLoader.fetchLogs(exFlow.getExecutionId(), "", 0, offset,
-          length);
+      LogData logData = this.nearlineExecutionLogsLoader.fetchLogs(exFlow.getExecutionId(), "", 0,
+          offset, length);
+
+      // Return offline logs if nearline logs are empty or the flow in kubernetes pod crashed
+      if (offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() &&
+          (logData.getLength() == 0 ||
+              (exFlow.getDispatchMethod() == DispatchMethod.CONTAINERIZED &&
```
hqin2016, line 469: do execution_stopped and killed flows have logdata length > 0? If not, the second condition only needs to check the logdata length.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -463,8 +495,22 @@ protected LogData getJobLogData(final ExecutableFlow exFlow, final String jobId,
...
     } else {
-      return this.executorLoader.fetchLogs(exFlow.getExecutionId(), jobId, attempt,
-          offset, length);
+      LogData logData = this.nearlineExecutionLogsLoader.fetchLogs(exFlow.getExecutionId(), jobId,
+          attempt, offset, length);
+
+      // Return offline logs if nearline logs are empty or the flow and job in kubernetes pod
+      // crashed
+      if (offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() &&
+          (logData.getLength() == 0 ||
```
hqin2016, line 504: same comment as above flow log data.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: do execution_stopped and killed flows have logdata length > 0?: yes. But Adam's LogPodDetails PR will write additional info into mysql if crashed. For safety, I just add the second condition here. If we finally decide not to add the LogPodDetails, I can remove the second condition later.

hqin2016 (approved):
LGTM

PR #3110, opened 2022-06-08
Refactor loggings in FlowContainer and add Kafka log4j appender support

byjiang1996:
Before my change, FlowContainer assumes log4j.properties file is using "logs/azkaban-execserver.log" as the log file and further uploads that local file into MySQL. The assumption can easily break if we upgrade our log4j.properties or when open source users start using containerized azkaban.

After my change, following things will be done:
1. Check if log4j rootLogger has FileAppender integrated, if yes, use FileAppender.getFile() as the log file and uploads to MySQL when process ends.
2. If not, create FileAppender internally and attach it into log4j rootLogger. Use newly created FileAppender.getFile() as the log file and uploads to MySQL when process ends.
3. Add Kafka log4j appender support into FlowContainer b/c FlowRunner's kafka log4j appender support is disabled when flow is running in containerization.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
+++ b/azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
@@ -29,6 +29,7 @@ import azkaban.cluster.ClusterModule;
 import azkaban.cluster.ClusterRouter;
 import azkaban.common.ExecJettyServerModule;
 import azkaban.logs.ExecutionLogsLoader;
+import azkaban.utils.KafkaLog4jUtils;
 import azkaban.utils.ServerUtils;
 import azkaban.event.Event;
 import azkaban.event.EventListener;
@@ -86,6 +87,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.KeyStore;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -190,9 +200,14 @@ public FlowContainer(final Props props,
...
-    this.azKabanProps = props;
+    this.azkabanProps = props;
+    // Get the execution ID from the environment
+    this.execId = getExecutionId();
+    // Set up RootLogger
+    createLogger();
+
     // Setup global props if applicable
-    final String globalPropsPath = this.azKabanProps.getString("executor.global.properties", null);
+    final String globalPropsPath = this.azkabanProps.getString("executor.global.properties", null);
```
hqin2016, line 210: how about making "executor.global.properties" constant?

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -805,12 +813,60 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+          logger.info("Attached new Kafka appender for flow");
+        }
+      }
+    } catch (final Exception e) {
+      logger.error("Could not open log file: " + logFile.getAbsolutePath(), e);
+    }
+  }
+
+  private static void removeAppender(final Appender appender) {
+    Logger rootLogger = Logger.getRootLogger();
```
hqin2016, line 856: final

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
```diff
@@ -805,12 +813,60 @@ public static void deleteSymlinkedFile(final Path symlinkedFilePath)
...
+    if (appender != null) {
+      rootLogger.removeAppender(appender);
+      appender.close();
```
hqin2016, line 859: do we need to handle exception here?

byjiang1996: Good catch! It should be handled well otherwise uploadLogFile will not be called if exception occurs.
I will address it in next follow-up PR tomorrow b/c FlowContainer&JobRunner&FlowRunner all have the removeAppender logic. It is out of the scope of this PR if touching JobRunner&FlowRunner.

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
byjiang1996: done

azkaban-exec-server/src/main/java/azkaban/container/FlowContainer.java
byjiang1996: done. Thanks for bringing this up!

PR #3112, opened 2022-06-13
Use correct yarn resource config file in the cancel job

chiragagrawal19:
Cancel flow is currently sending the cancel request to __default_ yarn cluster_. In case **cluster-router** is enabled, it might send the job cancel request to a different yarn cluster, causing the cancel request to fail. _This issue is being seen for Spark Job Type._
With this change, the correct yarn resource configuration file is used to initialize the yarn client.(The yarn resource config path is available in job properties)


djaiswal83 (approved):
Thanks for contributing Chirag.
This code looks good to me.

PR #3114, opened 2022-06-16
Revert "Revert "Use Apache's Parquet-Avro which uses Avro 1.9.2 (#2810)"" and fix Avro HDFS file viewer bugs

byjiang1996:
1. The changes are needed for potential avro upgrades to 1.9.2 because _EncoderFactory.get().jsonEncoder(schema, g);_ interface will be no longer available there.
2. The issue mentioned in https://github.com/azkaban/azkaban/pull/2861 is fixed by reverting the parquetAvro package changes. Reproduced successfully and verified it can work after my fix.

### DIFF
--- a/az-hdfs-viewer/src/main/java/azkaban/viewer/hdfs/ParquetFileViewer.java
+++ b/az-hdfs-viewer/src/main/java/azkaban/viewer/hdfs/ParquetFileViewer.java
@@ -30,8 +30,8 @@ import org.apache.avro.io.EncoderFactory;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.log4j.Logger;
-import org.apache.parquet.avro.AvroParquetReader;
-import org.apache.parquet.avro.AvroSchemaConverter;
+import parquet.avro.AvroParquetReader;
+import parquet.avro.AvroSchemaConverter;
 
 /**
  * This class implements a viewer for Parquet files.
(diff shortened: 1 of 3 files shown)


hqin2016 (approved):
Thank you so much for the bug fix.

djaiswal83 (approved):
Looks good to me. Please merge it.

az-hdfs-viewer/src/main/java/azkaban/viewer/hdfs/ParquetFileViewer.java
```diff
@@ -58,15 +52,14 @@ public String getName() {
...
-      throws AccessControlException {
+  public Set<Capability> getCapabilities(FileSystem fs, Path path) {
     if (logger.isDebugEnabled()) {
       logger.debug("Parquet file path: " + path.toUri().getPath());
     }
 
     AvroParquetReader<GenericRecord> parquetReader = null;
     try {
-      parquetReader = new AvroParquetReader<GenericRecord>(path);
+      parquetReader = new AvroParquetReader<>(path);
```
ypadron-in, line 62: I found a [commit](https://github.com/ypadron-in/azkaban/commit/d8532ea229fbe8132e485dbf5832d15d97f74ff3) (not yet finished apparently) in my local repo that was supposed to fix below issue:
https://github.com/azkaban/azkaban/pull/2861
> 
The issue is: Files that should be handled by the TextFileViewer (like JSON files) are now incorrectly handled by the ParquetFileViewer.

Could you please take a look at the commit? There are a couple of tests that you can leverage perhaps and more importantly there is a comment that suggests that we need to read the first record of the file to actually prevent the issue.

PR #3115, opened 2022-06-25
apple silicon arm

jeffsteinmetz:
Here is a working example of the Azkaban source / gradel / node working with Apple Silicon M1 (ARM architecture)

# prerequisite
You must install the ARM versions of the Java JDK as well as the Java JDX FX.
This was tested by installing JDK 1.8  and JDK FX 1.8 for MacOS ARM from Azul: 

https://www.azul.com/downloads/?package=jdk

- uses gradle 6.3
- uses latest node plugin and node V16
- adds an architecture target hack to `az-exec-util`
- commented out the unsupported build scan plugin

### DIFF
--- a/az-exec-util/build.gradle
+++ b/az-exec-util/build.gradle
@@ -17,6 +17,7 @@ apply plugin: 'c'
 model {
     components {
         main(NativeExecutableSpec) {
+            targetPlatform "osx_x86-64"
             sources {
                 c {
                     source {
--- a/azkaban-web-server/build.gradle
+++ b/azkaban-web-server/build.gradle
@@ -20,7 +20,7 @@ plugins {
 
 node {
     // Version of node to use.
-    version = '8.10.0'
+    version = '16.15.1'
 
     // Version of npm to use.
     npmVersion = ''
@@ -30,7 +30,7 @@ node {
 
     // If true, it will download node using above parameters.
     // If false, it will try to use globally installed node.
-    download = false
+    download = true
 
     // Set the work directory for unpacking node
     workDir = file("${project.buildDir}/nodejs")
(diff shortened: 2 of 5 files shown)


az-exec-util/build.gradle
```diff
@@ -17,6 +17,7 @@ apply plugin: 'c'
 model {
     components {
         main(NativeExecutableSpec) {
+            targetPlatform "osx_x86-64"
```
juhoautio, line 20: Is this also a TODO? I'm not expert with Gradle, but does this sett the target platform regardless of what the current platform is?

jeffsteinmetz: I don't know much about grade.  all the Gradle docs state is "Specifies the names of one or more [Flavor] that this component should be built for."

I think this forces the build to x86 vs ARM so that it runs Azkaban in rosetta mode on ARM / apple silicon.

jeffsteinmetz: I now see it only applies to a "c" utility in `az-exec-util`.  I think it just makes the utility written in C build for x86, so the util will run in rosetta mode on ARM.
There may be a way to get that utility written in C to work on ARM, but that was beyond the scope of what I was able to pull together for this PR.

juhoautio: Thanks. I was wondering what happens on other than osx platforms. Seems that this doesn't ruin it for them, at least the Travis build seems to have succeeded past the point of building `az-exec-util`.

jeffsteinmetz: you are welcome to fork this and create a different MR if you can get everything else to work.  I just put this MR here for posterity for anyone else having apple silicon / ARM issues.

Issue #1897, opened 2018-08-07
ERROR happens while using condition parameter 


huaxuehuaxue:
@jamiesjc  any idea about this? 

07-08-2018 09:23:06 CST conditional_flow1 ERROR - Invalid expression.
javax.script.ScriptException: &lt;eval&gt;:1:1 Expected ; but found {
${jobA:azkaban.server.name} == 'bar'
 ^ in &lt;eval&gt; at line number 1 at column number 1
	at jdk.nashorn.api.scripting.NashornScriptEngine.throwAsScriptException(NashornScriptEngine.java:470)
	at jdk.nashorn.api.scripting.NashornScriptEngine.compileImpl(NashornScriptEngine.java:537)
	at jdk.nashorn.api.scripting.NashornScriptEngine.compileImpl(NashornScriptEngine.java:524)
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:402)
	at jdk.nashorn.api.scripting.NashornScriptEngine.eval(NashornScriptEngine.java:155)
	at javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:264)
	at azkaban.execapp.FlowRunner.evaluateExpression(FlowRunner.java:989)
	at azkaban.execapp.FlowRunner.isConditionOnRuntimeVariableMet(FlowRunner.java:948)
	at azkaban.execapp.FlowRunner.getImpliedStatus(FlowRunner.java:905)
	at azkaban.execapp.FlowRunner.runReadyJob(FlowRunner.java:572)
	at azkaban.execapp.FlowRunner.progressGraph(FlowRunner.java:524)
	at azkaban.execapp.FlowRunner.runFlow(FlowRunner.java:422)
	at azkaban.execapp.FlowRunner.run(FlowRunner.java:249)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: jdk.nashorn.internal.runtime.ParserException: &lt;eval&gt;:1:1 Expected ; but found {
${jobA:azkaban.server.name} == 'bar'
 ^
[...]

PR #3118, opened 2022-06-28
Fail azkaban flow when conditional variable cannot be resolved

cbrentharris:
Currently the conditional azkaban workflow will return null when
attempting to resolve a variable in the job properties. However,
this leads to strange exceptions such as ones seen here:

https://github.com/azkaban/azkaban/issues/1897

"javax.script.ScriptException: <eval>:1:1 Expected ; but found {"

This is not very helpful and indicitive of what the problem is.
This changeset fails hard if required variables cannot be found,
the job name cannot be found, or the properties don't exist.
It also provides helpful other information like the available properties
that have variables to choose from.

### DIFF
--- a/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalJobsTest.java
+++ b/azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalJobsTest.java
@@ -73,6 +73,7 @@ public class FlowRunnerConditionalJobsTest extends FlowRunnerTestBase {
     final Props generatedProperties = new Props();
     generatedProperties.put("key1", "value1");
     generatedProperties.put("key2", "value2");
+    generatedProperties.put("key3", "value4");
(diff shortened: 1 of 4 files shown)


azkaban-exec-server/src/test/java/azkaban/execapp/FlowRunnerConditionalJobsTest.java
```diff
@@ -144,7 +145,7 @@ public void runFlowOnJobStatusConditionNull() throws Exception {
    * JobB has defined "condition: var fImport = new JavaImporter(java.io.File); with(fImport) { var
    * f = new File('new'); f.createNewFile(); }"
    * Null ProtectionDomain will restrict this arbitrary code from creating a new file.
-   * However it will not kick in when the change for condition whitelisting is implemented.
+   * However it will not kick in when the change for condition allow-listing is implemented.
```
svnarumugam, line 148: Thanks for making this change!

svnarumugam (approved):
Changes look good to me. Thanks for working on this change, appreciate the context you have added in the PR.

djaiswal83 (approved):
Thanks for contributing to Azkaban. The change looks good to me.

PR #3119, opened 2022-06-29
Set queued flows as executions in PREPARING state via containerized dispatch method

hqin2016:
so that metric of NumQueuedFlows can represent actual number of queued flows; Increase ContainerDispatchFail meter when pod is cleaned up by CleanupManager when stuck in PREPARING and DISPATCHING

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
@@ -149,7 +149,7 @@ public class ContainerizedDispatchManager extends AbstractExecutorManagerAdapter
   public List<Integer> getQueuedFlowIds() {
     final List<Integer> allIds = new ArrayList<>();
     try {
-      getExecutionIdsHelper(allIds, this.executorLoader.fetchQueuedFlows(Status.READY));
+      getExecutionIdsHelper(allIds, this.executorLoader.fetchQueuedFlows(Status.PREPARING));
     } catch (final ExecutorManagerException e) {
       logger.error("Failed to get queued flow ids.", e);
     }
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -149,15 +149,15 @@ public List<Integer> getRunningFlowIds() {
   public List<Integer> getQueuedFlowIds() {
     final List<Integer> allIds = new ArrayList<>();
     try {
-      getExecutionIdsHelper(allIds, this.executorLoader.fetchQueuedFlows(Status.READY));
+      getExecutionIdsHelper(allIds, this.executorLoader.fetchQueuedFlows(Status.PREPARING));
```
byjiang1996, line 152: I think DISPATCHING should also be included in getQueuedFlowIds function.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -149,15 +149,15 @@ public List<Integer> getRunningFlowIds() {
   public List<Integer> getQueuedFlowIds() {
```
byjiang1996, line 149: Remember to update the comment for this function

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
hqin2016: Added dispatching.

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
hqin2016: Thanks, updated the comment.

byjiang1996 (approved):
LGTM. Thanks for the fix

PR #3127, opened 2022-07-20
Do not use offline logs when getting application ids

byjiang1996:
There are reported slowness issues when clicking job logs button in UI when offline logs is enabled. 
The RC is: getApplicationIds function synchronously loads all logs from offline logs batch by batch (default batch size: 50000 Bytes), it can be very slow when log size is huge. Note: Most offline log fetching throughput is great but latency is worse than fetching from Nearline log MySQL.

As short-term mitigation, we will not use offline logs when getApplicationIds.
For long-term, we will improve getApplicationIds soon in these months to remove its dependency over reading job logs so this PR can be reverted after that.


NancyXie2022 (approved):
ship it

PR #3130, opened 2022-07-21
Add flowEndTime to fetchLogs interface

byjiang1996:
Since we introduced offline logging fetching recently into Azkaban and every offline logging fetching platform has decent amount of delay from log-is-sent to log-is-available, we need flowEndTime here in order to tell if the offline logs is complete or not, something like:
If flowEndTime + offlineLogAvailableSLA >= currentTime: offline logs are incomplete
If flowEndTime + offlineLogAvailableSLA < currentTime: offline logs are complete


NancyXie2022 (approved):
ship it

PR #3131, opened 2022-07-27
Return offline logs if nearline logs are empty

byjiang1996:
The original logic assumes nearline logs will not be empty if flow/job is crashed (what logPodDetails PR will do). There is no need to guarantee that assumption now as logPodDetails PR is abandoned. Instead, if mysql logs is empty, we can fetch offline logs directly without any additional checking.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -462,11 +462,8 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
     } else {
       LogData logData = this.nearlineExecutionLogsLoader.fetchLogs(exFlow.getExecutionId(), "", 0,
           offset, length, exFlow.getEndTime());
-      // Return offline logs if nearline logs are empty or the flow in kubernetes pod crashed
-      if (offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() &&
-          (logData == null ||
-              (exFlow.getDispatchMethod() == DispatchMethod.CONTAINERIZED &&
-                  (exFlow.getStatus() == Status.KILLED || exFlow.getStatus() == Status.EXECUTION_STOPPED)))) {
+      // Return offline logs if nearline logs are empty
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -462,11 +462,8 @@ protected LogData getFlowLogData(final ExecutableFlow exFlow, final int offset,
     } else {
       LogData logData = this.nearlineExecutionLogsLoader.fetchLogs(exFlow.getExecutionId(), "", 0,
           offset, length, exFlow.getEndTime());
-      // Return offline logs if nearline logs are empty or the flow in kubernetes pod crashed
-      if (offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() &&
-          (logData == null ||
-              (exFlow.getDispatchMethod() == DispatchMethod.CONTAINERIZED &&
-                  (exFlow.getStatus() == Status.KILLED || exFlow.getStatus() == Status.EXECUTION_STOPPED)))) {
+      // Return offline logs if nearline logs are empty
+      if (offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() && logData == null) {
```
amielkelinkedin, line 466: Since this logic is duplicated below would it make sense to refactor into a methd?

byjiang1996: I feel like it is okay to duplicate here. One condition is 
`offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() && logData == null`
Another is 
`!nearlineOnly && offlineLogsLoaderEnabled && offlineExecutionLogsLoader.isPresent() && logData == null` (so slightly different).

If we want to refactor the last three ANDs into a method, the method name itself is hard to figure out and it only combines three AND operations which seems to be an overkill.

amielkelinkedin: This makes sense, thank you

amielkelinkedin (approved):
Thank you so much, @byjiang1996

PR #3132, opened 2022-07-28
Implement a flow name based filter. This filter functions as follows.

djaiswal83:
- If the flow name is in the filter, then the flow runs on bare metal,
- otherwise, the flow runs on container.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
@@ -27,7 +27,7 @@ public class ContainerFlowCriteria {
   // Flows are stored in a map where key is project name and value is set of flows.
   private Map<String, Set<String>> flows = new HashMap<>(1);
   private final String fileLocation;
-  private final org.slf4j.Logger logger;
+  private final Logger logger;
 
   /**
    *
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
@@ -21,7 +21,12 @@ import org.slf4j.Logger;
 
 
 /**
- * Class for determining {@link azkaban.DispatchMethod} based on flow name filter.
+ * Class for determining {@link azkaban.DispatchMethod} based on project or flow name filter.
+ * A sample file can look like this:
+ * proj1:flow1
+ * proj1:flow2
+ * proj2
+ * proj3:flow1
  */
 public class ContainerFlowCriteria {
   // Flows are stored in a map where key is project name and value is set of flows.
@@ -78,7 +83,7 @@ public class ContainerFlowCriteria {
       for (String line; (line = reader.readLine()) != null;) {
         line = line.trim();
         List<String> flowFQN = Arrays.asList(line.split(":"));
-        validateAndAddFlow(flowFQN);
+        validateAndAdd(flowFQN);
       }
     } catch (final IOException e) {
       // Log and ignore
@@ -86,18 +91,28 @@ public class ContainerFlowCriteria {
   }
 
   /** Validate flowFQN and add it to flows map.
-   * @param flowFQN stored in the file is of format,
-   * <project name>:<flow name>. Each flowFQN must be of size two.
+   * @param flowFQN stored in the file is of format:
+   * <project name> when entire project needs to be in the filter and
+   * <project name>:<flow name>.
    */
-  private void validateAndAddFlow(final List<String> flowFQN) {
-    if (flowFQN.size() != 2) {
+  private void validateAndAdd(final List<String> flowFQN) {
+    if (!(flowFQN.size() == 1 || flowFQN.size() == 2)) {
       return;
     }
+
     final String project = flowFQN.get(0);
+    if (flowFQN.size() == 1) {
+      // Entire project is in the filter.
+      this.flows.put(project, null);
+      return;
+    }
+
+    // Handle the case when the FQN also contains flow name
     if (!this.flows.containsKey(project)) {
       // Make an entry for the project
       this.flows.put(project, new HashSet<>(1));
     }
+
     this.flows.get(project).add(flowFQN.get(1));
   }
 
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
```diff
@@ -0,0 +1,120 @@
...
+  private void validateAndAddFlow(final List<String> flowFQN) {
+    if (flowFQN.size() != 2) {
+      return;
+    }
+    final String project = flowFQN.get(0);
+    if (!this.flows.containsKey(project)) {
+      // Make an entry for the project
+      this.flows.put(project, new HashSet<>(1));
+    }
+    this.flows.get(project).add(flowFQN.get(1));
```
byjiang1996, line 101: Azkaban uses ":" to build nested flow names: e.g. basicFlows:embededFlow.
Can we use line.split(":")[0] as project name and the string after the first ":" as flow name here?

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
```diff
@@ -0,0 +1,120 @@
...
+   * <project name>:<flow name>. Each flowFQN must be of size two.
+   */
+  private void validateAndAddFlow(final List<String> flowFQN) {
+    if (flowFQN.size() != 2) {
+      return;
+    }
+    final String project = flowFQN.get(0);
+    if (!this.flows.containsKey(project)) {
+      // Make an entry for the project
+      this.flows.put(project, new HashSet<>(1));
```
byjiang1996, line 99: How about using "this.flows.getOrDefault(project, new HashSet<>(1)).add(flowFQN.get(1))" instead of an additional put operation?

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+/**
+ * Class for determining {@link azkaban.DispatchMethod} based on flow name filter.
+ */
+public class ContainerFlowCriteria {
+  // Flows are stored in a map where key is project name and value is set of flows.
+  private Map<String, Set<String>> flows = new HashMap<>(1);
+  private final String fileLocation;
+  private final org.slf4j.Logger logger;
```
byjiang1996, line 30: Consider replacing "org.slf4j.Logger logger" to "Logger logger"

azkaban-common/src/test/resources/flow_filter.txt
```diff
@@ -0,0 +1,3 @@
+proj1:flow1
```
svnarumugam, line 1: I am curious how often these configurations get updated. This is a well-structured configuration and we have provided a way to reload the configuration as necessary in this PR. Did we consider keeping this config in our database?

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Reloads the in-memory flows map with the flow filter file.
+   */
+  public void reloadFlowFilter() {
+    loadFlowFilter();
+  }
+
+  @VisibleForTesting
+  public void reloadFlowFilter(final String fileLocation) {
+    loadFlowFilter(fileLocation);
```
svnarumugam, line 53: [thinking out loud] this change can be re-used in other cases too. If we have decided where to store, the format of the data, and APIs to reload.

azkaban-common/src/test/resources/flow_filter2.txt
```diff
@@ -0,0 +1,4 @@
+proj1:flow1
+proj2:flow2
```
svnarumugam, line 2: did we consider using a well-defined format like json/xml? It could be simple in this case but I am wondering about the custom code to read and parse this correctly.

svnarumugam (commented):
The change looks fine to solve the current problem. Left a few comments.

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
djaiswal83: The subflows are logical entities which do not come into picture at this level.

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
djaiswal83: Logically its the same code, I prefer to keep it this way. Call me old school, but this is more readable.

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
djaiswal83: Ack

dhilsand (changes requested):
Could we add a feature to also filter on the basis of project name along with flow name? Project level filters would be more concise when setting up the filter configuration. It will also cover the following scenario: when new flows are added to a filtered project, those new flows automatically get filtered out as well. This is also the common case - generally we want to filter everything in the project, its a rare case otherwise.

azkaban-common/src/test/resources/flow_filter.txt
djaiswal83: We expect to update the config by directly modifying the file either using some script or manually but not frequently. Once modified, we call the reload API which updates the filter in-memory.

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
djaiswal83: If future case comes up, we can refactor the code.

azkaban-common/src/test/resources/flow_filter2.txt
djaiswal83: This is a simple use case. I would use JSON for a complex structure.

djaiswal83:
> Could we add a feature to also filter on the basis of project name along with flow name? Project level filters would be more concise when setting up the filter configuration. It will also cover the following scenario: when new flows are added to a filtered project, those new flows automatically get filtered out as well. This is also the common case - generally we want to filter everything in the project, its a rare case otherwise.

Working on it.

byjiang1996 (approved):
LGTM. Thanks for the change!

PR #3133, opened 2022-07-28
Fixed a bug in recent flow filter logic where the reloading of filter…

djaiswal83:
… must reset the map.


NancyXie2022 (commented):
Will HashMap.clear() be better? You don't need to create too many objects

byjiang1996 (approved):

PR #3136, opened 2022-08-02
Only allow fetching one batch of flow log at a time

byjiang1996:
There is a bug to load flow logs in UI: if the user click on "Refresh" button multiple times without waiting for the previous log batch to be loaded. Same log batch will be loaded together via Azkaban API and rendered in the UI.

Instead, we should do **wait** for the previous log batch to be loaded and then fetch the next batch.


amielkelinkedin (approved):
Thanks so much for the change!

PR #3138, opened 2022-08-04
Log the time taken to download all thin archive dependencies

byjiang1996:
We can possibly leverage this metrics in the future to improve our system.


svnarumugam (approved):
Thanks for explaining offline that this metric will be useful to compare with the spark's user-level cache effort.

PR #3140, opened 2022-08-16
When activating 1 image_version, correctly set other ACTIVE ones to STABLE state and update their modified_on timestamp

tsangz2013:
**Bug to fix:** 
This is a follow up of #2952
When activating 1 image_version, other previously ACTIVE image_versions (if ant) would be set to a **wrong** 'STABLE' state and their 'modified_on' timestamp is **NOT** updated

**Changes made:**
Modify and correct the SQL execution logic

**Testing done:**
1. Push the built jar to an cluster machine
2. Restart the azkaban-web-server with the new jar
3. Call /imageVersions API to reproduce the image_version activating scenario


tsangz2013:
Tested in an instance and validated that the issue is being fixed

djaiswal83 (approved):
Thanks for making this change. Looks good to me.

PR #3144, opened 2022-08-30
Implement CreateRule for RampManager

ShellyWEI:
Support a Rest request to specify a rule_id, image_name, and image_version, in order to fill in denyList for the rule.
Exception would be thrown on:
- Missing requirements rule_id, image_name, image_version
- Not Authorized if creator is not the owner of image_type
- Duplicate rule_id

End2End tests including above failure cases have been operated and validated on Holdem6.
Success request would return:
< HTTP/1.1 201 Created
Otherwise would return:
HTTP/1.1 <ErrorCode> <Error Message>
for example:
HTTP/1.1 404 Unable to fetch image version metadata. Invalid image version: 0.1.169.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package azkaban.imagemgmt.daos;
 
 import azkaban.imagemgmt.models.ImageRampRule;
@@ -5,19 +20,42 @@ import azkaban.imagemgmt.models.ImageRampRule;
 
 /**
 * Data access object (DAO) for accessing image ramp rule that deny an image ramping version for flows.
-* This interface defines add/remove/modify an image ramp rule.
+* This interface defines add/remove/modify/get an image ramp rule.
 * */
 public interface RampRuleDao {
 
-  boolean isHPRule(String ruleId);
+  /**
+   * Query table ramp_rules to check whether a HPFlowRule .
+   *
+   * @param ruleName - ruleName in {@see ImageRampRule}
+   * @return true - if ramp rule is setup for HP Flows, denying any ramp up plan;
+   *         false - if ramp rule is normal rule, regulate only on certain image version.
+   */
+  boolean isHPFlowRule(final String ruleName);
 
-  /*
-  * Insert a new row into DB for ramp rule
-  * */
-  int addRampRule(ImageRampRule rule);
+  /**
+   * Insert Image Ramp Rule metadata into DB.
+   *
+   * @param rule - RampRule model created in {@see ImageRampRuleServiceImpl}
+   * @return int - id of the DB entry
+   * @throws azkaban.imagemgmt.exception.ImageMgmtDaoException
+   */
+  int addRampRule(final ImageRampRule rule);
 
-  String getOwners(String ruleId);
+  /**
+   * Query table ramp_rules to get owners of Ramp rule.
+   *
+   * @param ruleName - ruleName in {@see ImageRampRule}
+   * @return owners of the ramp rule.
+   */
+  String getOwners(final String ruleName);
 
-  void deleteRampRule(String ruleId);
+  /**
+   * delete the ramp_rules to get owners of Ramp rule.
+   *
+   * @param ruleName - ruleName in {@see ImageRampRule}
+   * @throws azkaban.imagemgmt.exception.ImageMgmtDaoException
+   */
+  void deleteRampRule(final String ruleName);
 
 }
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
@@ -1,0 +1,23 @@
+package azkaban.imagemgmt.daos;
+
+import azkaban.imagemgmt.models.ImageRampRule;
+
+
+/**
+* Data access object (DAO) for accessing image ramp rule that deny an image ramping version for flows.
+* This interface defines add/remove/modify an image ramp rule.
+* */
+public interface RampRuleDao {
+
+  boolean isHPRule(String ruleId);
+
+  /*
+  * Insert a new row into DB for ramp rule
+  * */
+  int addRampRule(ImageRampRule rule);
+
+  String getOwners(String ruleId);
+
+  void deleteRampRule(String ruleId);
+
+}
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -1,12 +1,24 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package azkaban.imagemgmt.daos;
 
 import azkaban.db.DatabaseOperator;
 import azkaban.imagemgmt.exception.ErrorCode;
 import azkaban.imagemgmt.exception.ImageMgmtDaoException;
-import azkaban.imagemgmt.exception.ImageMgmtException;
-import azkaban.imagemgmt.exception.ImageMgmtInvalidInputException;
 import azkaban.imagemgmt.models.ImageRampRule;
-import azkaban.imagemgmt.models.ImageType;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
@@ -23,14 +35,18 @@ import org.apache.commons.dbutils.ResultSetHandler;
 import org.apache.log4j.Logger;
 import org.bouncycastle.util.Strings;
 
-
+/**
+ * Dao Implementation for access DB of table ramp_rules.
+ * Create/update/delete/get RampRule or partial RampRule metadata. {@link ImageRampRule}
+ * */
 @Singleton
 public class RampRuleDaoImpl implements RampRuleDao {
   private static final Logger LOG = Logger.getLogger(RampRuleDaoImpl.class);
   private final DatabaseOperator databaseOperator;
 
   private static String INSERT_RAMP_RULE = "INSERT into ramp_rules "
-      + "(rule_id, image_name, image_version, owners, is_HP, created_by, created_on) values (?, ?, ?, ?, ?, ?, ?)";
+      + "(rule_name, image_name, image_version, owners, is_HP, created_by, created_on, modified_by, modified_on)"
+      + " values (?, ?, ?, ?, ?, ?, ?, ?, ?)";
 
   @Inject
   public RampRuleDaoImpl(DatabaseOperator databaseOperator) {
@@ -38,50 +54,60 @@ public class RampRuleDaoImpl implements RampRuleDao {
   }
 
   @Override
-  public boolean isHPRule(String ruleId) {
+  public boolean isHPFlowRule(final String ruleName) {
     return false;
   }
 
+  /**
+   * Insert new ramp rule into DB, check if duplicate ruleName exists first
+   *
+   * @param rampRule
+   * @throws ImageMgmtDaoException
+   * */
   @Override
-  public int addRampRule(ImageRampRule rampRule) {
+  public int addRampRule(final ImageRampRule rampRule) {
     try {
+      // duplicated rule should be forbidden as client side error, which needs a separate check to
+      // avoid it falls into normal SQL Exception as server error.
       List<ImageRampRule> rampRules = databaseOperator.query(
-          FetchRampRuleHandler.FETCH_RAMP_RULE_BY_ID, new FetchRampRuleHandler(), rampRule.getRuleId());
+          FetchRampRuleHandler.FETCH_RAMP_RULE_BY_ID, new FetchRampRuleHandler(), rampRule.getRuleName());
       if (!rampRules.isEmpty()) {
-        LOG.error("Error in create ramp rule on duplicate ruleId: " + rampRule.getRuleId());
-        throw new ImageMgmtException(ErrorCode.BAD_REQUEST,
-            "Error in create ramp rule on duplicate ruleId: " + rampRule.getRuleId());
+        LOG.error("Error in create ramp rule on duplicate ruleName: " + rampRule.getRuleName());
+        throw new ImageMgmtDaoException(ErrorCode.BAD_REQUEST,
+            "Error in create ramp rule on duplicate ruleName: " + rampRule.getRuleName());
       }
       return this.databaseOperator.update(INSERT_RAMP_RULE,
-          rampRule.getRuleId(),
+          rampRule.getRuleName(),
           rampRule.getImageName(),
           rampRule.getImageVersion(),
           rampRule.getOwners(),
           rampRule.isHPRule(),
           rampRule.getCreatedBy(),
+          Timestamp.valueOf(LocalDateTime.now()),
+          rampRule.getCreatedBy(),
           Timestamp.valueOf(LocalDateTime.now()));
     } catch (SQLException e) {
       LOG.error("Error in create ramp rule on DB" + rampRule);
       throw new ImageMgmtDaoException(ErrorCode.INTERNAL_SERVER_ERROR,
-          "Error creating ramp rule " + rampRule.getRuleId() + "with " + e.getMessage());
+          "Error creating ramp rule " + rampRule.getRuleName() + "with " + e.getMessage());
     }
   }
 
   @Override
-  public String getOwners(String ruleId) {
+  public String getOwners(final String ruleName) {
     return null;
   }
 
   @Override
-  public void deleteRampRule(String ruleId) {
+  public void deleteRampRule(final String ruleName) {
 
   }
 
   public static class FetchRampRuleHandler implements ResultSetHandler<List<ImageRampRule>> {
 
     private static final String FETCH_RAMP_RULE_BY_ID =
-        "SELECT rule_id, image_name, image_version, owners, is_HP"
-            + " FROM ramp_rules WHERE rule_id = ?";
+        "SELECT rule_name, image_name, image_version, owners, is_HP"
+            + " FROM ramp_rules WHERE rule_name = ?";
 
     @Override
     public List<ImageRampRule> handle(final ResultSet rs) throws SQLException {
@@ -90,13 +116,19 @@ public class RampRuleDaoImpl implements RampRuleDao {
       }
       final List<ImageRampRule> imageRampRules = new ArrayList<>();
       do {
-        final String ruleId = rs.getString("rule_id");
+        final String ruleName = rs.getString("rule_name");
         final String imageName = rs.getString("image_name");
         final String imageVersion = rs.getString("image_version");
         final String owners = rs.getString("owners");
         final boolean isHP = rs.getBoolean("is_HP");
         Set<String> ownerLists = new HashSet<>(Arrays.asList(Strings.split(owners, ',')));
-        final ImageRampRule imageRampRule = new ImageRampRule(ruleId, imageName, imageVersion, ownerLists, isHP);
+        final ImageRampRule imageRampRule = new ImageRampRule.Builder()
+            .setRuleName(ruleName)
+            .setImageName(imageName)
+            .setImageVersion(imageVersion)
+            .setOwners(ownerLists)
+            .setHPRule(isHP)
+            .build();
         imageRampRules.add(imageRampRule);
       } while (rs.next());
       return imageRampRules;
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -49,7 +49,7 @@ public class RampRuleDaoImpl implements RampRuleDao {
       + " values (?, ?, ?, ?, ?, ?, ?, ?, ?)";
 
   @Inject
-  public RampRuleDaoImpl(DatabaseOperator databaseOperator) {
+  public RampRuleDaoImpl(final DatabaseOperator databaseOperator) {
     this.databaseOperator = databaseOperator;
   }
 
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -1,0 +1,105 @@
+package azkaban.imagemgmt.daos;
+
+import azkaban.db.DatabaseOperator;
+import azkaban.imagemgmt.exception.ErrorCode;
+import azkaban.imagemgmt.exception.ImageMgmtDaoException;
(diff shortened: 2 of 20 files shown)


azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
```diff
@@ -0,0 +1,23 @@
...
+import azkaban.imagemgmt.models.ImageRampRule;
+
+
+/**
+* Data access object (DAO) for accessing image ramp rule that deny an image ramping version for flows.
+* This interface defines add/remove/modify an image ramp rule.
+* */
+public interface RampRuleDao {
+
+  boolean isHPRule(String ruleId);
```
byjiang1996, line 12: Add comments for each method here?
Specially the full name of "HP".

byjiang1996: Also add "throws ImageMgmtException" if the method will throw. Refer to ImageRampupDao.java.

byjiang1996:
General, it is better to add class comment for all classes: e.g.
`/**
 * Data access object (DAO) for accessing image rampup plan and rampups for an image type. This
 * interface defines method such as create, get, update etc. for image rampups.
 */
public interface ImageRampupDao {`

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    // validate image_name and image_version
+    final ImageType imageType = imageTypeDao
+      .getImageTypeByName(rampRuleRequest.getImageName())
+        .orElseThrow(() -> new ImageMgmtInvalidInputException(ErrorCode.NOT_FOUND, String.format("Unable to"
+            + " fetch image type metadata. Invalid image type: %s.", rampRuleRequest.getImageName())));
+    if (!this.imageVersionDao.isInvalidVersion(rampRuleRequest.getImageName(), rampRuleRequest.getImageVersion())) {
+      throw new ImageMgmtInvalidInputException(ErrorCode.NOT_FOUND, String.format(
+          "Unable to fetch image version metadata. Invalid image version: %s.", rampRuleRequest.getImageVersion()));
+    }
+    boolean isAuthorized = isAzkabanAdmin;
```
byjiang1996, line 53: Can you reuse hasImageManagementPermission() method to tell if a user is authorized or not?
Also put this authorization into *Servlet: e.g. see how we do it in ImageRampupServlet.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    if (!isAuthorized) {
+      throw new ImageMgmtInvalidPermissionException(ErrorCode.UNAUTHORIZED,
+          "Only image type owners would be authorized to create ramp rules, "
+              + "unauthorized user " + ldapUser + " is not in " + imageOwnerships);
+    }
+    // convert ImageRampRule and insert new ramp rule into DB
+    ImageRampRule rampRule = new ImageRampRule(rampRuleRequest.getRuleId(), rampRuleRequest.getImageName(),
+        rampRuleRequest.getImageVersion(), imageOwnerships, false);
+    rampRule.setCreatedBy(ldapUser);
```
byjiang1996, line 70: Can you pass "createdBy" as a param in the ImageRampRule constructor?

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -0,0 +1,50 @@
+package azkaban.imagemgmt.models;
+
+import java.util.List;
+import java.util.Set;
+
+
+public class ImageRampRule extends BaseModel {
+  // ramp rule name
+  String ruleId;
```
byjiang1996, line 9: Minor: can we rename the ruleId into ruleName? It does not look correct to me b/c I think *Id should be an integer?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Override
+  public boolean isHPRule(String ruleId) {
+    return false;
+  }
+
+  @Override
+  public int addRampRule(ImageRampRule rampRule) {
+    try {
+      List<ImageRampRule> rampRules = databaseOperator.query(
```
byjiang1996, line 48: I don't think we need to manually check the duplicate in java level. Since you defined rule_id as PRIMARY KEY, it is by default UNIQUE at db level and thus if duplicate, db will throw the error back directly.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -0,0 +1,50 @@
...
+  // image name of ramp rule for
+  String imageName;
+  // image version of ramp rule for
+  String imageVersion;
+  // rule owners
+  Set<String> owners;
```
byjiang1996, line 15: From the code logic, owners here should always == image owner? Introducing too many and redundant owner names will make future maintenance too complicated. Can we just discard this field but rely on image_ownership table to tell the ownership?
What's your thought?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
```diff
@@ -0,0 +1,23 @@
+package azkaban.imagemgmt.daos;
```
djaiswal83, line 1: Please add open source and copyright header. Make sure it has correct year. Can be picked from any of the existing code.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -0,0 +1,105 @@
+package azkaban.imagemgmt.daos;
```
djaiswal83, line 1: Header missing

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
```diff
@@ -0,0 +1,23 @@
...
+* This interface defines add/remove/modify an image ramp rule.
+* */
+public interface RampRuleDao {
+
+  boolean isHPRule(String ruleId);
+
+  /*
+  * Insert a new row into DB for ramp rule
+  * */
+  int addRampRule(ImageRampRule rule);
```
djaiswal83, line 17: Please use keyword final wherever applicable. Please follow the coding guidelines of Azkaban.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
djaiswal83: I agree here. We should handle the error and return to user if insert fails.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.List;
+import java.util.Set;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.commons.dbutils.ResultSetHandler;
+import org.apache.log4j.Logger;
+import org.bouncycastle.util.Strings;
+
+
+@Singleton
```
djaiswal83, line 27: Please add comments in javadocs format everywhere

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    } catch (SQLException e) {
+      LOG.error("Error in create ramp rule on DB" + rampRule);
+      throw new ImageMgmtDaoException(ErrorCode.INTERNAL_SERVER_ERROR,
+          "Error creating ramp rule " + rampRule.getRuleId() + "with " + e.getMessage());
+    }
+  }
+
+  @Override
+  public String getOwners(String ruleId) {
+    return null;
```
djaiswal83, line 72: Why is this not implemented?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public void deleteRampRule(String ruleId) {
+
```
djaiswal83, line 77: Why is this not implemented?

azkaban-common/src/main/java/azkaban/imagemgmt/dto/ImageRampRuleRequestDTO.java
```diff
@@ -0,0 +1,49 @@
+package azkaban.imagemgmt.dto;
```
djaiswal83, line 1: Missing header

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
djaiswal83: +1

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
djaiswal83: This is likely a subset of owners, however, I agree that image_ownership is better bet. It keeps single source of truth.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -0,0 +1,50 @@
...
+  // Ramp Rule for HP flows or not
```
djaiswal83, line 16: Change the comment to suggest what the value implies.
Ramp Rule for HP flows. True if the flow is HP, false otherwise.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -0,0 +1,50 @@
+package azkaban.imagemgmt.models;
```
djaiswal83, line 1: Missing header

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleService.java
```diff
@@ -0,0 +1,27 @@
+package azkaban.imagemgmt.services;
```
djaiswal83, line 1: Missing header

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
+package azkaban.imagemgmt.services;
```
djaiswal83, line 1: Missing header

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+import azkaban.imagemgmt.models.ImageType;
+import azkaban.user.UserManager;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.apache.log4j.Logger;
+
+@Singleton
```
djaiswal83, line 21: Add comments wherever applicable. Describe the functionality of the class.
Add comments for all the API definitions below as well.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageRampRuleServlet.java
```diff
@@ -0,0 +1,85 @@
+package azkaban.imagemgmt.servlets;
```
djaiswal83, line 1: Missing header

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageRampRuleServlet.java
```diff
@@ -0,0 +1,85 @@
...
+      String requestURI = req.getRequestURI();
+      String userId = session.getUser().getUserId();
+      boolean isAzkabanAdmin = isAzkabanAdmin(session.getUser());
+      if (CREATE_RULE_URI.equals(requestURI)) {
+        LOG.info("handle request from post uri: " + requestURI);
+        handleCreateRampRule(req, resp, userId, isAzkabanAdmin);
+      }
+  }
+
+  // create an exclusive rule for a certain version of the image,
```
djaiswal83, line 61: Please use javadocs style comment here.

azkaban-web-server/src/test/java/azkaban/imagemgmt/services/ImageRampRuleServiceImplTest.java
```diff
@@ -0,0 +1,96 @@
+package azkaban.imagemgmt.services;
```
djaiswal83, line 1: Missing header

djaiswal83 (changes requested):
Good craftsmanship overall. Appreciate that you added unit tests for both positive and negative test cases.

I have few comments,

The code is not following the coding guidelines such as adding header with copyright and licensing at the top of new files, javadocs style comments.
Use of keyword "final" wherever applicable.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
ShellyWEI: those will be implemented in different API implementation. For this one, it will be implemented in AddFlowstoRule API.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
ShellyWEI: It will be implemented in DeleteRule API. Don't want to squeeze everything into createRule PR. It would be more easier to review too

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
ShellyWEI: I saw that code. The reason I am not using that is I think this is duplicated code. I need to getOwnerships too instead of simply check if this is authorized. I setup ownership of the rule instead image ownership directly to handle the different cases, for example hp flows; and later API would only rely on rule owner to authorize instead of looking into image owner.
If you still think reusing PermissionManager is needed. I will probably add validatedPermissionAndGetImageOwnerships
validatedPermissionAndGetHPFlowOwnerships
hasPermissionOnRule
as new method,
change hasPermission to hasPermissionOnImage

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
ShellyWEI: I tested this will be returned as any other SQLException and, I think internal SQL error would be marked as 500 server error, while this duplicated request would be categorized as client error, and falls in to 400 BadRequest.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
ShellyWEI: Sorry to make the confusion here but I meant to add HP flow owners as rule owners too. The owner source of truth would come from image owners for regular rule and hp flow owners for HP flow rule.
This PR is only for regular rule creation, So I didn't add HP flow part in this one. The next one would add HP as owers

byjiang1996: I see. It makes sense.
But I feel like owners field for non-hp rule is still not good. We had so many asks before to add some member to the owner list. If we want to store owner list also in this table, it means whenever an ask comes, we need to add owner in two places (or maybe more places b/c one imageName may have multiple ImageRampRules): higher maintenance cost.

ShellyWEI: Initially I do added such logic to only add group owners into the owner list. But I wonder if this is too harsh? Because many teams are missing the group as owner, and we likely get little group as owner, we can ask them to update the image ownerships with their group, but then it means we need to open such API to change. Because all the other APIs are supposed to rely on this owner column to operate the permission.
I can add it back if you think we should do that. Let me know how you feel.

ShellyWEI: Or maybe a safe second check would just work. If update/delete API failed to authorize, then just do another permission check.

byjiang1996: The current logic makes sense. Thanks for discussing with me in offline!

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
byjiang1996: It makes sense to me. Thanks for the clarification!

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
byjiang1996: Sounds good to me

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
```diff
@@ -77,4 +78,34 @@ public boolean hasPermission(final UserManager userManager, final String imageTy
...
+   * @param user
+   * @return boolean
+   */
+  @Override
+  public Set<String> validatePermissionAndGetOwnerships(String imageTypeName, User user)
+      throws ImageMgmtException {
+    String userId = user.getUserId();
+    // validate azkaban admin
+    boolean isAzkabanAdmin = user.getRoles().stream()
+        .anyMatch(role -> userManager.getRole(role).getPermission().isPermissionSet(Type.ADMIN));
```
byjiang1996, line 97: Can you move LoginAbstractAzkabanServlet's isAzkabanAdmin into this class and reuse that method here to prevent the code redundancy?

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
```diff
@@ -42,25 +44,24 @@ public class PermissionManagerImpl implements PermissionManager {
...
    * @param imageTypeName
    * @param userId
    * @param type
    * @return boolean
    */
   @Override
-  public boolean hasPermission(final UserManager userManager, final String imageTypeName,
-      final String userId,
-      final Type type) {
+  public boolean hasPermission(final String imageTypeName, final String userId, final Type type) {
```
byjiang1996, line 64: Can you run validatePermissionAndGetOwnerships inside of hasPermission method?
Most of the code here are redundant.

ShellyWEI: They're not exactly the same. So hasPermission is based on authorizing image by Type, the type means imageCreate/imageGet/ImageDelete. But for ramp manager, it does not need detailed image operation level authorization. I tried to use Type.Admin but unfortunately, in the testing I didn't didn't find Admin is in User's Permission (even my ldap for az-platform-image) so we would get no imageOwnerships if we select ADMIN and for non azkaban admin it will fail.

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
ShellyWEI: ok, this makes sense.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
ShellyWEI: done

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
ShellyWEI: done

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
ShellyWEI: Done and I would add more details related to HP flow rule in CreateHPFlow API implementation. This is only a marker for them.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -1,35 +1,62 @@
...
-  public ImageRampRule(String ruleId,
+  public ImageRampRule(String ruleName,
                 String imageName,
                 String imageVersion,
                 Set<String> owners,
-                boolean isHP) {
-    this.ruleId = ruleId;
+                boolean isHP,
+                String createdBy,
+                @Nullable String createdOn ,
```
djaiswal83, line 44: Why is this field Nullable?

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
```diff
@@ -1,35 +1,62 @@
...
   String imageVersion;
   // rule owners
   Set<String> owners;
-  // Ramp Rule for HP flows or not
+  // Ramp Rule for HP flows
+  // ture if rule is for HP flow; false otherwise
   boolean isHPRule;
 
-  public ImageRampRule(String ruleId,
+  public ImageRampRule(String ruleName,
```
djaiswal83, line 38: All input params should be "final"

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
```diff
@@ -77,4 +78,34 @@ public boolean hasPermission(final UserManager userManager, final String imageTy
...
+   * @return boolean
+   */
+  @Override
+  public Set<String> validatePermissionAndGetOwnerships(String imageTypeName, User user)
+      throws ImageMgmtException {
+    String userId = user.getUserId();
+    // validate azkaban admin
+    boolean isAzkabanAdmin = user.getRoles().stream()
+        .anyMatch(role -> userManager.getRole(role).getPermission().isPermissionSet(Type.ADMIN));
+    // fetch owners from image_ownerships with matching permission set (i.g. ADMIN)
```
djaiswal83, line 98: e.g instead of i.g

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
djaiswal83: I still dont see code dedup or am I missing something?

djaiswal83 (changes requested):
Thanks for addressing comments. We are almost there. I have a few minor comments.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
ShellyWEI: updated.

azkaban-web-server/src/main/java/azkaban/imagemgmt/servlets/ImageRampRuleServlet.java
ShellyWEI: updated

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
ShellyWEI: check the latest code, not this version.
For review purpose I didn't squash all the commits. But every time the changes based on the review would be a separate commit.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
ShellyWEI: By mistake, initially I thought a different constructor then later I abandoned it and forgot to remove the annotation. It's fixed.

azkaban-common/src/main/java/azkaban/imagemgmt/models/ImageRampRule.java
ShellyWEI: fixed

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
ShellyWEI: fixed

byjiang1996 (approved):

djaiswal83 (approved):
Thanks for addressing the comments. Code looks good to me.

PR #3148, opened 2022-09-08
Implement CreateHPFlowRule and updateOwnerships

ShellyWEI:
- Add implementation for creating ramp rule for high priority flows /createHPFlowRule Introducing "ownerships" directly in the rule creation for easy managing RampRule and serves as source of truth. For /createRule, missing ownerships would use default ImageType Owners; For /createHPFlowRule, ownerships parameter is madantory since current azkaban system does not maintain concept of high priotity flows ownership.
- Add two APIs to manage ownership management: /addOwners and /removeOwners. update ownership operation only can be done by existing owners or azkaban admin.

End to End test has been performed in the following cases success cases:
createRampRule w/ and w/o designated owners 201 Created createHPFlowRule 201 Created
add/removeOwners appropriate permission 200 OK
Failure cases:
Invalid ownerId (user and group) 400
updateOwners without being existing owner 401
updateOwners with non existing rule 400

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
@@ -16,6 +16,7 @@
 package azkaban.imagemgmt.daos;
 
 import azkaban.imagemgmt.models.ImageRampRule;
+import java.util.Set;
 
 
 /**
@@ -43,12 +44,24 @@ public interface RampRuleDao {
   int addRampRule(final ImageRampRule rule);
 
   /**
+   * Update owners metadata into table ramp_rules.
+   *
+   * @param newOwners
+   * @param ruleName
+   * @param modifiedBy
+   * @return int - id of the DB entry
+   * @throws azkaban.imagemgmt.exception.ImageMgmtDaoException
+   */
+  int updateOwnerships(final String newOwners, final String ruleName, final String modifiedBy);
+
+  /**
    * Query table ramp_rules to get owners of Ramp rule.
+   * Serves as the source of truth for rule ownership management.
    *
    * @param ruleName - ruleName in {@see ImageRampRule}
    * @return owners of the ramp rule.
    */
-  String getOwners(final String ruleName);
+  Set<String> getOwners(final String ruleName);
 
   /**
    * delete the ramp_rules to get owners of Ramp rule.
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -44,9 +44,12 @@ public class RampRuleDaoImpl implements RampRuleDao {
   private static final Logger LOG = Logger.getLogger(RampRuleDaoImpl.class);
   private final DatabaseOperator databaseOperator;
 
-  private static String INSERT_RAMP_RULE = "INSERT into ramp_rules "
+  private static final String INSERT_RAMP_RULE = "INSERT into ramp_rules "
       + "(rule_name, image_name, image_version, owners, is_HP, created_by, created_on, modified_by, modified_on)"
       + " values (?, ?, ?, ?, ?, ?, ?, ?, ?)";
+
+  private static final String UPDATE_RULE_OWNERSHIP = "UPDATE ramp_rules "
+      + "SET owners=?, modified_by=?, modified_on=? WHERE rule_name=?";
 
   @Inject
   public RampRuleDaoImpl(final DatabaseOperator databaseOperator) {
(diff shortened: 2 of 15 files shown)


azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
```diff
@@ -121,4 +117,29 @@ public boolean isAzkabanAdmin(final User user) {
...
+  /**
+   * Validate each userId's identity using userManager. userId can be group or individual user account.
+   *
+   * @param ids
+   * @throws ImageMgmtValidationException
+   */
+  @Override
+  public void validateIdentity(List<String> ids) {
+    Set<String> invalids = ids.stream()
+        .filter(id -> !userManager.validateGroup(id) && !userManager.validateUser(id))
```
byjiang1996, line 130: IIRC, the group can be validated by "validateGroup" here is merely a subset of valid groups used in validateUserGroupMembership

azkaban-web-server/src/test/resources/image_management/image_ramp_rule.json
```diff
@@ -1,5 +1,6 @@
 {
   "ruleName": "daliSparkExclude",
```
byjiang1996, line 2: Can you add one more test for a list of ownerships?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
```diff
@@ -42,13 +43,25 @@ public interface RampRuleDao {
...
+
   /**
    * Query table ramp_rules to get owners of Ramp rule.
+   * Serves as the source of truth for rule ownership management.
    *
    * @param ruleName - ruleName in {@see ImageRampRule}
    * @return owners of the ramp rule.
    */
-  String getOwners(final String ruleName);
+  Set<String> getOwners(final String ruleName);
```
tsangz2013, line 64: Minor: since these changes also included in PR, maybe also mention them in title, or consider the next time have 2 separate commits / PRs

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -93,9 +96,50 @@ public int addRampRule(final ImageRampRule rampRule) {
...
+   * Query table ramp_rules to get owners of Ramp rule.
+   * Serves as the source of truth for rule ownership management.
+   *
+   * @param ruleName - ruleName in {@see ImageRampRule}
+   * @return owners of the ramp rule.
+   */
   @Override
-  public String getOwners(final String ruleName) {
-    return null;
+  public Set<String> getOwners(final String ruleName) {
```
tsangz2013, line 129: the above method may not need test, but this method worth add some simple tests

azkaban-db/src/main/sql/create.ramp_rules.sql
```diff
@@ -1,13 +1,13 @@
...
+    image_name VARCHAR(200),
+    image_version VARCHAR (100),
     owners VARCHAR (1000) NOT NULL,
     is_HP BIT NOT NULL,
     created_by VARCHAR(20) NOT NULL,
     created_on DATE NOT NULL,
-    modified_by VARCHAR(20),
-    modified_on DATE,
+    modified_by VARCHAR(20) NOT NULL,
+    modified_on DATE NOT NULL,
```
tsangz2013, line 10: Consider this

```suggestion
    modified_on DATE NOT NULL ON UPDATE CURRENT_TIMESTAMP,
```

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDao.java
ShellyWEI: long story short: 
the reason I added ownershipManagement part into this PR is now the ownership management is different than the original design.
In the previous design doc, there is a plan to maintaining an independent hp_flow_ownership table just as image_ownerships. I abandoned this for two reasons:
- hard to fetch ownership during HP rule creation as lack of flow information;
- introducing high priority flow service into the system is not necessary, we only want hp_owners so let's just keep this ownership within rampRule to make the ownership more straightforward and easy managable.

So in order to address the difference to change the ownership source of truth from maintaining a table to a column in ramp rules. I added support to modify ownerships too, comparing to the previous design that all the ownership would need to fetch from hp_flow_ownership and image_ownership tables.

azkaban-common/src/main/java/azkaban/imagemgmt/permission/PermissionManagerImpl.java
ShellyWEI: let's have an offline discussion since this is Linkedin specific.

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -117,7 +118,11 @@ public void createRule(final ImageRampRuleRequestDTO rampRuleRequest, final User
    * @throws ImageMgmtValidationException when user does not have permission
    * */
   @Override
-  public void createHpFlowRule(final RampRuleOwnershipRequestDTO hpFlowRuleRequestDTO, final User user) {
+  public void createHpFlowRule(final RampRuleOwnershipDTO hpFlowRuleRequestDTO, final User user) {
```
byjiang1996, line 121: Can you rename hpFlowRuleRequestDTO to hpFlowRuleOwnershipDTO?

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleService.java
```diff
@@ -41,25 +41,25 @@ public interface ImageRampRuleService {
...
    * Create an HP Flow exclusive Rule to denying all image Types,
-   * validation performed based on {@link RampRuleOwnershipRequestDTO} and convert to {@link ImageRampRule}
+   * validation performed based on {@link RampRuleOwnershipDTO} and convert to {@link ImageRampRule}
    *
    * @param hpFlowRuleRequestDTO
    * @param user
    * @throws ImageMgmtException
    * */
-  void createHpFlowRule(final RampRuleOwnershipRequestDTO hpFlowRuleRequestDTO, final User user);
+  void createHpFlowRule(final RampRuleOwnershipDTO hpFlowRuleRequestDTO, final User user);
```
byjiang1996, line 50: Can you rename hpFlowRuleRequestDTO to hpFlowRuleOwnershipDTO?

azkaban-db/src/main/sql/create.ramp_rules.sql
ShellyWEI: our mysql version is old, it does not support this syntax.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
ShellyWEI: I changed the service implementation code to cover updateOwners which would be using this Dao.getOwner.
Unless we setup test DB in for integration test otherwise it's useless only mock the behavior in unit test.

byjiang1996 (approved):
LGTM. Thanks for adding this feature!

PR #3149, opened 2022-09-09
Upload newer version project property files before incrementing proje…

byjiang1996:
…ct version

Otherwise there will be some edge cases: when a flow is executed, its project version is already incremented but project property files haven't been uploaded yet and thus take no effect to that execution.


djaiswal83 (approved):
Thanks for fixing the bug.

PR #3150, opened 2022-09-09
Add getApplications API call and use it for Yarn app killing

tsangz2013:


azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -30,6 +38,38 @@ public class YarnUtils {
...
+  /**
+   * Use the yarnClient to query the unfinished yarn applications, then use the yarnClient to kill
+   * them sequentially
+   *
+   * @param yarnClient the yarnClient already connects to the cluster
+   * @param flowExecID the azkaban flow execution id whose yarn applications needs to be killed
+   * @return the set of all to-be-killed (alive) yarn applications' IDs
+   * @throws IOException   for RPC issue
+   * @throws YarnException for YARN server issue
+   */
```
tsangz2013, line 58: 1: call yarn cluster to get app-report and transform to app-ids

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -373,11 +376,40 @@ public Void run() throws Exception {
...
 
+  /**
+   * Get the yarn applications' ids that needs to be killed (the ones alive / spawned),
+   * first use yarn client to call the cluster, if it fails, fallback to read the job log file
+   * content looking for application ids
+   * @param yarnClient the started client
+   * @param jobProps should contain flow execution id, and the job log file's path
+   * @param log logger
+   * @return the set of application ids
+   */
```
tsangz2013, line 392: 2: helper method first call yarnClient, with fallback to scan log

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -30,6 +38,38 @@ public class YarnUtils {
...
+  public static Set<String> getAllAliveAppIDsByExecID(final YarnClient yarnClient,
```
NancyXie2022, line 59: You would like to return the ApplicationReport from this method and create another method for application id. Since we need to get the user from ApplicationReport as well, if we write it in the current way, we need to have another call to get the user.

PR #3151, opened 2022-09-10
Move PermissionManager bindings from Common to WebServer

ShellyWEI:
[Bug Fix]Move PermissionManager bindings from CommonModule to WebServerModule

This is caused by missing UserManager initializtion config in flow container and fail in pods. Webserver is configured to correct userManager so let's keep authorization in web server only.


djaiswal83 (approved):
Thanks for fixing this bug.

PR #3152, opened 2022-09-12
Add logic to reload flow filter when a project is uploaded to ensure …

djaiswal83:
…the flow filter is always current.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
@@ -43,7 +43,7 @@ public class ContainerFlowCriteria {
     this.fileLocation = azkProps.getString(Constants.
         ContainerizedDispatchManagerProperties.CONTAINERIZED_FLOW_FILTER_FILE, null);
     this.logger = logger;
-    loadFlowFilter();
+    loadFlowFilter(new HashMap<>());
   }
 
   /**
@@ -51,19 +51,17 @@ public class ContainerFlowCriteria {
    */
   public void reloadFlowFilter() {
     // Reset the filter map
-    flows = new HashMap<>(1);
-    loadFlowFilter();
+    loadFlowFilter(new HashMap<>());
   }
 
   @VisibleForTesting
   public void reloadFlowFilter(final String fileLocation) {
     // Reset the filter map
-    flows = new HashMap<>(1);
-    loadFlowFilter(fileLocation);
+    loadFlowFilter(new HashMap<>(), fileLocation);
   }
 
-  private void loadFlowFilter() {
-    loadFlowFilter(this.fileLocation);
+  private void loadFlowFilter(final Map<String, Set<String >> flowMap) {
+    loadFlowFilter(flowMap, this.fileLocation);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1954,6 +1955,10 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
       });
 
       registerErrorsAndWarningsFromValidationReport(resp, ret, reports);
+      // Reload the flow_filter to ensure if this project is added to the file, then the filter is current.
+      ContainerizedDispatchManager containerizedDispatchManager =
```
byjiang1996, line 1959: Add a if before this reload.
`if (this.executorManagerAdapter instanceof ContainerizedDispatchManager) {`

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1954,6 +1955,10 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+          (ContainerizedDispatchManager) this.executorManagerAdapter;
+      containerizedDispatchManager.getContainerFlowCriteria().reloadFlowFilter();
```
byjiang1996, line 1961: In current implementation of reloadFlowFilter(), we will do follows:
1. Reset "Map<String, Set<String>> flows"
2. Parse flow filter file and add flow into flows map.

This can cause similar ordering issue when:
1. Project is being uploaded
2. Flow is being scheduled/executed at the same time.

As a result, flow will be scheduled to containerization b/c flow map is almost empty / slowly added during the reloadFlowFilter(). As a result, I think we should change the logic inside of reloadFlowFilter:
1. Parse flow filter file and add flow into flows map.
2. Replace the current "Map<String, Set<String>> flows"

djaiswal83: Good suggestion. This avoids race condition. Will update the API.

byjiang1996: BTW, I feel like the fix I proposed is incorrect. Instead, the ideal way is to add a lock in getDispatchMethod() and reloadFlowFilter() so that only one can be executed at a time. I'm not pretty sure re-pointing the flows map (the previous design) will still cause race condition or not.

azkaban-common/src/main/java/azkaban/executor/container/ContainerFlowCriteria.java
```diff
@@ -43,35 +43,33 @@ public ContainerFlowCriteria(final Props azkProps, final Logger logger) {
     this.fileLocation = azkProps.getString(Constants.
         ContainerizedDispatchManagerProperties.CONTAINERIZED_FLOW_FILTER_FILE, null);
     this.logger = logger;
-    loadFlowFilter();
+    loadFlowFilter(new HashMap<>());
```
NancyXie2022, line 46: Shall we use this https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#emptyMap()? for emptyMap

djaiswal83: It is immutable. We are intending to populate that map and replace "this.flows" with it.

byjiang1996 (approved):

PR #3153, opened 2022-09-12
Return null instead of exception when job output prop does not exist …

byjiang1996:
…and show READY jobs in UI.

In Azkaban partial execution for a conditional flow/job, if parent job's output prop that is depended by some child job is missing, the flow will be marked as FAILED directly even though some other children jobs are okay to proceed if they do not have dependency over the missing output prop.

As the fix, if output prop is missing, we will still continue processing the jobs that do not depend on the missing output prop instead of marking the flow as FAILED immediately. Such output prop missing info will still be printed out in flow logs to notify users.

### DIFF
--- a/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
+++ b/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
@@ -1201,9 +1201,7 @@ public class FlowRunner extends EventHandler<Event> implements Runnable {
     while (variableMatcher.find()) {
       final String value = findValueForJobVariable(node, variableMatcher.group(1),
           variableMatcher.group(2));
-      if (value != null) {
-        replaced = replaced.replace(variableMatcher.group(), "'" + value + "'");
-      }
+      replaced = replaced.replace(variableMatcher.group(), "'" + value + "'");
       this.logger.info("Resolved condition of " + node.getId() + " is " + replaced);
     }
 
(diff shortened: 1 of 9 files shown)


azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1224,10 +1221,12 @@ private String findValueForJobVariable(final ExecutableNode node, final String j
 
     final Props outputProps = target.getOutputProps();
     if (outputProps == null) {
-      throw new IllegalStateException("No output properties were generated for job " + jobName + ".");
+      this.logger.warn("No output properties were generated for job " + jobName + ".");
+      return null;
     }
     if (!outputProps.containsKey(variable)) {
-      throw new IllegalStateException("No variable named " + variable + " found in job " + jobName + " output parameters. Properties: " + outputProps);
+      this.logger.warn("No variable named " + variable + " found in job " + jobName + " output parameters. Properties: " + outputProps);
```
svnarumugam, line 1228: Clarification: what is the implication of this change? Currently, when the user code hasn't passed a property we throw an exception and it surfaces in the logs. How are we planning to have the same behavior after this fix?

byjiang1996: Good point!
After the fix, user can still see that message in the flow log but it will not KILL the flow directly but fail the condition check for the conditional job.

svnarumugam (approved):
Looks good to me. Thanks for fixing this bug. for future reference, could you please consider adding more context about the bug fix in the description? I tried to answer the questions
1. If this bug is a regression of any recent change, it looks like yes
2. how did we find this problem? Is it a user-reported issue? what symptoms made us look into this?
3. what will happen if we don't fix this problem? My guess is that the subsequent jobs that depend on the job property will fail or behave erratically.

PR #3145, opened 2022-09-06
Extract all yarn related methods to azkaban-common as shared util library

tsangz2013:
**Issue to solve**: A part of the fixing / upgrade of the yarn application killing logic: make it easier to add and reuse "call yarn cluster for application ids" method later on

**Changes made**: Some simple refactor of the code, extract all the yarn-client related methods to an azkaban-common util s class

**Testing done**: 

- Added unit tests for killAppOnCluster() method;
- Tested in the bare metal exec-server on a spark flow:

1. upload both jar for az-hadoop-jobtype-plugin.jar and azkaban-common.jar to a cluster machine
2. restarted a exec-server
3. trigger the sparkFlow execution and kill it after the yarn application is created
4. verify the job log that the yarn app is killed

PR #3154, opened 2022-09-15
Revert "Extract all yarn related methods to azkaban-common as shared …

tsangz2013:
…util library (#3145)"

This reverts commit a31c30dcdb4f89bb32afd5e63d05d15576ba7878.

**Reason:** this change doesn't work with containerization: when try to call yarnClient API will encounter below error:
```
java.io.IOException: Failed on local exception: java.io.IOException: javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)];
```


svnarumugam (approved):
Thank you Zhi for investigating the problem, and working on the fix. LGTM.

PR #3157, opened 2022-09-22
Wait for all JobRunners to be closed before closing FlowRunner

byjiang1996:
We should follow what we did in FlowContainer: wait for all JobRunners/FlowRunner thread to be completed first and then close the FlowRunner/FlowContainer. Otherwise, some final cleanup operations happened in JobRunners/FlowRunner will be bypassed.


djaiswal83 (approved):
Thanks for fixing the bug. Looks good to me.

PR #3159, opened 2022-09-23
UpdateVersion and deleteRule API for RampManager

ShellyWEI:
Update Version for existing RampRule will update version metadata on DB directly; while delete rule will delete all metadata related to the rule. Both API would not introduce new insert but only update existing data. Failure would occur when:
- no existing rule
- input failed validation
- no user permission

### DIFF
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -244,9 +244,8 @@ public class RampRuleDaoImpl implements RampRuleDao {
       return transOperator.getLastInsertId();
     };
     // end SQL transaction operator
-    int batchInsertId;
     try {
-      batchInsertId = this.databaseOperator.transaction(fetchRampRuleAndUpdateDenyList).intValue();
+      int batchInsertId = this.databaseOperator.transaction(fetchRampRuleAndUpdateDenyList).intValue();
       if (batchInsertId == 0) {
         LOG.warn(String.format("creating no new flow deny list based on rule: %s, "
                 + "flowList: %s. Might due to deny rule already exists", ruleName, flowIds));
--- a/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
+++ b/azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
@@ -20,6 +20,7 @@ import azkaban.db.SQLTransaction;
 import azkaban.imagemgmt.dto.RampRuleFlowsDTO.ProjectFlow;
 import azkaban.imagemgmt.exception.ErrorCode;
 import azkaban.imagemgmt.exception.ImageMgmtDaoException;
+import azkaban.imagemgmt.exception.ImageMgmtInvalidInputException;
 import azkaban.imagemgmt.models.ImageRampRule;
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -221,7 +244,7 @@ public int addFlowDenyInfo(final List<ProjectFlow> flowIds, final String ruleNam
       return transOperator.getLastInsertId();
     };
     // end SQL transaction operator
-    int batchInsertId = 0;
+    int batchInsertId;
     try {
       batchInsertId = this.databaseOperator.transaction(fetchRampRuleAndUpdateDenyList).intValue();
```
djaiswal83, line 249: Move bactchInsertId declaration from line 247 to here.

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
```diff
@@ -245,6 +268,38 @@ public int addFlowDenyInfo(final List<ProjectFlow> flowIds, final String ruleNam
...
+   * @param user
+   * @throws azkaban.imagemgmt.exception.ImageMgmtDaoException
+   */
+  @Override
+  public void updateVersionOnRule(final String newVersion, final String ruleName, final String user) {
+    final SQLTransaction<Long> fetchRampRuleAndUpdateDenyList = transOperator -> {
+      final ImageRampRule imageRampRule =
+          transOperator.query(FetchRampRuleHandler.FETCH_RAMP_RULE_BY_ID, new FetchRampRuleHandler(), ruleName);
+      final String newDenyVersion = String.join(":", imageRampRule.getImageName(), newVersion);
+      if (!newVersion.equals(imageRampRule.getImageVersion())) {
```
djaiswal83, line 286: In which case the newVersion will be same as existing version?

azkaban-web-server/src/main/java/azkaban/imagemgmt/services/ImageRampRuleServiceImpl.java
```diff
@@ -184,9 +185,22 @@ public String updateRuleOwnership(final RampRuleOwnershipDTO ruleOwnershipDTO, f
...
+   * @param ruleName - ruleName in {@link ImageRampRule}
+   * @param user - user to operate
+   * */
   @Override
-  public void deleteRule(final String ruleName) {
-
+  public void deleteRule(final String ruleName, final User user) {
+    // validate permission
+    final Set<String> owners = rampRuleDao.getOwners(ruleName);
+    if (!permissionManager.hasPermission(user, owners)) {
```
djaiswal83, line 198: This and code block at line 255 are duplicate. Please dedupe.

djaiswal83 (changes requested):
The code overall looks good. I have few comments, can you please address them?

azkaban-common/src/main/java/azkaban/imagemgmt/daos/RampRuleDaoImpl.java
ShellyWEI: It might be, if someone makes a mistake to put old version. Or others in the team don't know someone had already update the version to the target one, they might be update the same version.

djaiswal83 (approved):
Thanks for addressing the comments. The code looks good now.

PR #3161, opened 2022-09-27
Fix TriggerScanner to prevent a blocking object.wait()

sshardool:
Currently the Trigger Scanner thread can enter a blocking wait when invoked with a timeout value of 0ms. The change prevents this indefinite blocking and adds additional logging around this.


djaiswal83 (approved):
Thanks for fixing this years old bug!

PR #3162, opened 2022-09-28
Adding exclude rule into Ramp Manager during image version fetch

ShellyWEI:
Currently Ramp Manager selects ramp image version based on ramp up plan and the remaining images using current active version. Apply Ramp rule during image version fetch stage would prevent the flows to run versions to be ruled out by ramp rules. Two case for denying selected ramp version:
- flow is inside a HP flow Rule, this is the most highest priority, means the flow would not using any ramping version;
- flow is inside a normal ramp rule, this is only targets certain ramp versions that defined in the ramp rule;


svnarumugam (approved):
Changes look good to me. Thanks for working on this change @ShellyWEI !

PR #3165, opened 2022-09-29
Load tokens from HADOOP_TOKEN_FILE_LOCATION file and add to proxy user UGI for permission

tsangz2013:
**What need to be fixed:**
Yarn Client killApplication() call never succeeds in in containerization, as the proxyUser UGI never have proper token loaded.
And this RP in combination to #3164, the yarn application kill function works correctly in containerization.

**What is being done:**
When creating the proxyUser, use HadoopSecurityManager to get tokens from the previously stored tokenFile from file system, and grand them into the proxyUser.

**Tests done:**

- Tested with multiple different flows across different clusters, 
- kill yarn application works correctly, 
- verified in yarnUI that the call is being made from the pod as the proxyUser

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
@@ -16,12 +16,15 @@
 package azkaban.jobtype;
 
 import azkaban.Constants;
+import azkaban.security.commons.HadoopSecurityManagerException;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Properties;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
 import org.apache.log4j.Logger;
@@ -47,9 +50,11 @@ public class HadoopSecureWrapperUtils {
    * @return a UserGroupInformation object for the specified userToProxy, which will also contain
    * the logged in user's tokens
    */
-  private static UserGroupInformation createSecurityEnabledProxyUser(String userToProxy,
+  private static UserGroupInformation createSecurityEnabledProxyUser(
+      HadoopSecurityManager hadoopSecurityManager, String userToProxy,
       String fileLocation, Logger log
-  ) throws IOException {
+  ) throws IOException, HadoopSecurityManagerException {
+    log.info("createSecurityEnabledProxyUser starts");
 
     if (!new File(fileLocation).exists()) {
       throw new RuntimeException("hadoop token file doesn't exist.");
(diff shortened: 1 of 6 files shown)


azkaban-hadoop-security-plugin/src/main/java/azkaban/security/commons/HadoopSecurityManager.java
```diff
@@ -101,6 +102,9 @@ public abstract FileSystem getFSAsUser(String realIdentity, String proxyUser)
   public abstract void cancelTokens(File tokenFile, String userToProxy,
       Logger logger) throws HadoopSecurityManagerException;
 
+  public abstract Credentials getTokens(File tokenFile, Logger logger)
+      throws HadoopSecurityManagerException;
+
```
tsangz2013, line 107: HSM expose a new method to get tokens from file, and implemented in AbstractHSM, which will be inherent by KsudoHSM.

tsangz2013 (commented):
For better understanding

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
```diff
@@ -59,32 +66,61 @@ private static UserGroupInformation createSecurityEnabledProxyUser(String userTo
...
+          proxyUser));
+      for (Token<?> token : creds.getAllTokens()) {
+        proxyUser.addToken(token);
+        log.info(String.format("Token = %s, %s, %s ", token.getKind(), token.getService(),
+            Arrays.toString(token.getIdentifier())));
+      }
+      proxyUser.addCredentials(creds);
+    }
+
+    log.info("token copy finished for " + loginUser.getUserName());
```
svnarumugam, line 96: The login user is `azdata`. proxy user is any proxy user P. wouldn't it be sufficient if we do this alone? I am trying to figure out why we are doing the copy from line 76 to line 81. am i missing something?

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -401,12 +401,22 @@ protected void doPrefetch(final File tokenFile, final Props props, final Logger
...
 
     try {
+      // cred is being populated
       fetchAllHadoopTokens(userToProxyFQN, userToProxy, props, logger, cred);
       getProxiedUser(userToProxyFQN).doAs((PrivilegedExceptionAction<Void>) () -> {
         registerAllCustomCredentials(userToProxy, props, cred, logger);
         return null;
       });
+
+      logger.info("fetched cred = " + cred);
```
svnarumugam, line 411: [Logging] any strong reasons we shouldn't keep this as a `debug` level? I have the same question for other logging too.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
```diff
@@ -618,6 +628,16 @@ public void cancelTokens(final File tokenFile, final String userToProxy, final L
     }
   }
 
+  @Override
+  public Credentials getTokens(File tokenFile, Logger logger)
+      throws HadoopSecurityManagerException {
+    try {
+      return Credentials.readTokenStorageFile(new Path(tokenFile.toURI()), this.conf);
```
svnarumugam, line 635: we seem to be not using the logger instance here. did we intent to log something?

svnarumugam (changes requested):
Functionally the changes look good to me. Thank you so much @tsangz2013 for working on this. I have left some comments.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
tsangz2013: I don't think using any user out of nowhere is good, plus this is the intend of `proxy`User, right?
Duplicate code but these are 2 sources of tokens, first is copy from login user, second is from token-file.

azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java
tsangz2013: Good call, just follow the pattern for other HSM methods

svnarumugam (approved):
The changes look good to me. I think we can revisit the token copy part afterwards. The logging should be fine.

svnarumugam (approved):
Looks good to me.

PR #3167, opened 2022-10-03
Emit events for subflows and add subflow names for job life cycle events

byjiang1996:
It is a feature request.

1. If flow life cycle event is coming from subflows,  event object will have `isRootFlow=false` field
2. `subflowName` field will be attached to job life cycle events to represent the parent flow of the job
3. Use `event.getData().getNode() instanceof ExecutableFlow` to tell if an event is coming from root flow or subflow in azkaban main logic.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/event/EventData.java
+++ b/azkaban-common/src/main/java/azkaban/event/EventData.java
@@ -8,6 +8,7 @@ import azkaban.executor.Status;
  */
 public class EventData {
 
+  private final ExecutableNode node;
   private final Status status;
   private final String nestedId;
   private final String jobId;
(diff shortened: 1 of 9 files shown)


azkaban-common/src/main/java/azkaban/event/EventData.java
```diff
@@ -32,6 +34,10 @@ public EventData(final ExecutableNode node) {
     this.executionId = (node.getParentFlow() == null) ? -1: node.getParentFlow().getExecutionId();
   }
 
+  public ExecutableNode getNode() {
```
svnarumugam, line 37: Non blocker: Consider exposing a method that tells if the event is meant for executableFlow. we can use the method instead of the expression`getNode() instanceof ExecutableFlow`.

svnarumugam (approved):
Changes look good to me. Added a non-blocker comment.

svnarumugam (approved):
Thanks for clarifying questions offline.

azkaban-common/src/main/java/azkaban/event/EventData.java
byjiang1996: done

PR #3168, opened 2022-10-04
YarnUtils: Add method to call yarn getApplication api to get alive app IDs

tsangz2013:
query applications with execution-id tag and un-finished application states

**Tests done**
Tested together with #3169, in the c15n pod by killing the flow after yarn apps created.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
@@ -15,9 +15,17 @@
  */
 package azkaban.utils;
 
+import static azkaban.Constants.FlowProperties.AZKABAN_FLOW_EXEC_ID;
+
+import com.google.common.collect.ImmutableSet;
 import java.io.IOException;
+import java.util.EnumSet;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.yarn.api.records.ApplicationReport;
+import org.apache.hadoop.yarn.api.records.YarnApplicationState;
 import org.apache.log4j.Logger;
 import org.apache.hadoop.yarn.api.records.ApplicationId;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -30,6 +38,44 @@ public class YarnUtils {
...
 
+  public static final EnumSet<YarnApplicationState> YARN_APPLICATION_ALIVE_STATES = EnumSet.of(
+      YarnApplicationState.NEW,
+      YarnApplicationState.NEW_SAVING,
+      YarnApplicationState.SUBMITTED,
+      YarnApplicationState.ACCEPTED,
+      YarnApplicationState.RUNNING
+  );
+
+  public static Set<String> getAllAliveAppIDsByExecID(final YarnClient yarnClient,
```
svnarumugam, line 49: Clarification: any specific reason we are exposing two public methods one that returns List<AppReport> and the other that returns Set<String>? would it be any problem if just combine these methods into one?

svnarumugam (changes requested):
Added a clarifying question. Rest LGTM. Thanks Zhi!

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
tsangz2013: Good call, the one with `List<ApplicationReport>` gives full info, but the `Set<String>` is to be compatible with current kill method being used in the read-log-and-kill logic

tsangz2013: Refactoring is valuable, but only make sense to include after we got all the functions on the plate I think

svnarumugam: Thanks for the clarification.

svnarumugam (approved):
LGTM.

PR #3169, opened 2022-10-04
Use the getApplication IDs API from YarnUtils, and use scan ID from log as fallback logic

tsangz2013:
Use the new API in #3168 to get yarn app IDs, and use the old log-scanning method as a fallback

**Tested done**
Tested in c15n pod by killing the flow after yarn apps created.


svnarumugam (approved):
looks good to me. Thanks for adding the connection between the previous PR and this one.

PR #3170, opened 2022-10-05
Don't delete the token file so as it can be used when killing yarn apps

tsangz2013:
If we delete the token file before `killAllSpawnedHadoopJobs` is called, the proxyUser cannot get the tokens from the file, and thus the kill application call will not work.

This PR is to fix this.


djaiswal83 (approved):
Thanks for fixing the bug.

svnarumugam (approved):
LGTM

PR #3172, opened 2022-10-11
Add sanity check in FlowRunner to skip finished flows and kill pod wh…

byjiang1996:
…en canceling flows

There is a bug introduced by containerized azkaban when killing/canceling an execution. At that time, execution will have negative elapse time. The gap is:
1. Before FlowContainer java process is up, Azkaban web server kills the execution and do NOP to the underlining pod b/c its FlowContainer process is not up yet. — end time is marked for the execution
2. When FlowContainer in the pod starts, it does not check the flow status == KILLED or not, but directly continue all the flow related operations (for job node, it is fine b/c we have job status == KILLED check). — start time is marked for the execution which is later than end time

The fix is to:
1. Kill pod when execution is killed/canceled
2. Have sanity check in FlowRunner to skip processing root flows if flow status is not valid in the first place.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -24,6 +24,7 @@ import azkaban.event.Event;
 import azkaban.event.EventData;
 import azkaban.event.EventHandler;
 import azkaban.event.EventListener;
+import azkaban.executor.container.ContainerizedImpl;
 import azkaban.flow.FlowUtils;
 import azkaban.jobcallback.JobCallbackManager;
 import azkaban.logs.ExecutionLogsLoader;
@@ -49,6 +50,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -505,6 +501,17 @@ public void run() {
...
   }
 
+  @Override
+  protected void finalizeFlow(final ExecutableFlow flow, final String reason,
+      @Nullable final Throwable originalError, final Status finalFlowStatus) {
+    super.finalizeFlow(flow, reason, originalError, finalFlowStatus);
+    try {
+      this.containerizedImpl.deleteContainer(flow.getExecutionId());
+    } catch (ExecutorManagerException e) {
+      logger.info("Failed to delete container when finalizing flow", e);
```
ShellyWEI, line 511: can we add exec_id here. It might help to get more logs associated with the execution for debug.
Also should the log be marked as ERROR?

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -599,40 +601,45 @@ protected void handleCancelFlow(ExecutionReference executionReference,
...
-          EventType.FLOW_FINISHED, new EventData(executableFlow)));
-      // Throwing exception to make the reason appear on the UI.
-      throw new ExecutorManagerException(finalizingReason
-          + " Finalizing the flow.");
+      handleCancelFlowManually(executionReference, executableFlow, finalizingReason, e,
+          finalizingStatus);
     }
   }
 
+  private void handleCancelFlowManually(ExecutionReference executionReference,
```
ShellyWEI, line 624: Let's have comments for each extracted method. 
explanation what this method does and in which case these method should be used.
for example cancelFlowManually should be called when some exception hits and flow can not be gracefully killed. 
(BTW should we be consistent to use Kill or Cancel? it confused me a bit at first)

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -599,40 +601,45 @@ protected void handleCancelFlow(ExecutionReference executionReference,
...
+    finalizeFlow(executableFlow, reason, originalError, finalizingStatus);
+    // Killed flow events can only be sent out if callWithReferenceByUser completed successfully
+    // so we need to manually send one here.
+    this.fireEventListeners(Event.create(executableFlow,
+        EventType.FLOW_FINISHED, new EventData(executableFlow)));
+    // Throwing exception to make the reason appear on the UI.
+    throw new ExecutorManagerException(reason + " Finalizing the flow.");
+  }
+
+  protected void finalizeFlow(final ExecutableFlow flow, final String reason,
```
ShellyWEI, line 637: Same here. I had to look into low level to understand what is finalizeFlow means. it helps if there is a comment saying "mark the flow finish status in DB".

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
```diff
@@ -505,6 +501,17 @@ public void run() {
     }
   }
 
+  @Override
+  protected void finalizeFlow(final ExecutableFlow flow, final String reason,
```
ShellyWEI, line 505: Same here.
This one is "killing the corrresponding pod and mark the flow as finished in DB".
Also do you know why the original throwable can be null?

ShellyWEI (commented):
Thanks for patching flow process to make it to be more elegant clean up. It looks good to me overall.

ShellyWEI (approved):
lgtm. Thanks for addressing the comments

azkaban-common/src/main/java/azkaban/executor/container/ContainerizedDispatchManager.java
byjiang1996: It seems WARN is more appropriate b/c maybe the container is not up indeed.

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: Good point! Done.

PR #3173, opened 2022-10-11
Added a heartbeat metric in cron scheduler

svnarumugam:
This change adds a meter that acts like a heartbeat for the cron scheduler. verified that metrics are emitted. This metric can be one of the signals to check the health of the scheduler.


utk-12 (approved):
LGTM.

PR #3174, opened 2022-10-11
Read config for feature flagging of Yarn App Killing module

tsangz2013:
**Why we need this**: 
In future roll out, this can help fasten mitigation if we need to fallback to the old approach. 
We only need to change the `.properties` file config and ping the new config version to switch the feature.

**What's in this**:
Bring back the old logic of yarn kill, and use a config flag from `System Props` to control which approach to use

**Tests done**:
Tested in containerization, setting runtime property in UI, and the value can be passed to the jobProps for feature flagging

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -15,6 +15,9 @@
  */
 package azkaban.jobtype;
 
+import static azkaban.utils.YarnUtils.YARN_CONF_DIRECTORY_PROPERTY;
+import static azkaban.utils.YarnUtils.YARN_CONF_FILENAME;
+
 import azkaban.flow.CommonJobProperties;
 import azkaban.security.commons.HadoopSecurityManager;
 import azkaban.security.commons.HadoopSecurityManagerException;
@@ -46,7 +49,6 @@ import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.log4j.Logger;
 import org.apache.hadoop.fs.Path;
-import scala.util.parsing.combinator.testing.Str;
 
 
 /**
@@ -88,16 +90,13 @@ public class HadoopJobUtils {
   public static final String YARN_KILL_VERSION = "yarn.kill.version";
   // values of the yarn kill version flag
   public static final String YARN_KILL_LEGACY = "legacy";
-  public static final String YARN_KILL_TOKEN_FIX = "token_copy";
+  public static final String YARN_KILL_USE_API_WITH_TOKEN = "api_with_token";
 
   protected static final int APPLICATION_TAG_MAX_LENGTH = 100;
   // Root of folder in storage containing startup dependencies
   public static final String DEPENDENCY_STORAGE_ROOT_PATH_PROP = "dependency.storage.path.prefix";
   // Azkaban property for listing additional namenodes for delegation tokens
   private static final String OTHER_NAMENODES_PROPERTY = "other_namenodes";
-  //Yarn resource configuration directory for the cluster where the job is scheduled by the cluster router
-  private static final String YARN_CONF_DIRECTORY_PROPERTY = "env.YARN_CONF_DIR";
-  private static final String YARN_CONF_FILENAME = "yarn-site.xml";
 
   private HadoopJobUtils() {
   }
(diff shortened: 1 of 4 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -352,24 +365,50 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
...
-          public Void run() throws Exception {
-            findAndKillYarnApps(jobProps, log);
-            return null;
-          }
-        });
-      } else {
-        findAndKillYarnApps(jobProps, log);
+    // todo: use feature flag
+    String yarnKillVersion = jobProps.getString(YARN_KILL_VERSION, YARN_KILL_LEGACY);
+    if (YARN_KILL_LEGACY.equals(yarnKillVersion)) {
```
byjiang1996, line 370: How about leaving this to "else" instead of using "if & elseif"?
I'm afraid yarnKillVersion != YARN_KILL_LEGACY && yarnKillVersion != YARN_KILL_TOKEN_FIX by manual error.

svnarumugam (commented):
As we discussed offline, keeping this configuration in azkaban.properties would be easy to operate. If we keep it as an env variable, we would have to roll out a new version of azkaban-core to update the env variable.  Thanks for making this change very quickly @tsangz2013 .

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -80,11 +84,20 @@ public class HadoopJobUtils {
   public static final String MAPREDUCE_JOB_OTHER_NAMENODES = "mapreduce.job.hdfs-servers";
   // MapReduce config for mapreduce job tags
   public static final String MAPREDUCE_JOB_TAGS = "mapreduce.job.tags";
+
+  public static final String YARN_KILL_VERSION = "yarn.kill.version";
+  // values of the yarn kill version flag
+  public static final String YARN_KILL_LEGACY = "legacy";
+  public static final String YARN_KILL_TOKEN_FIX = "token_copy";
```
djaiswal83, line 91: lets rename to YARN_KILL_USING_API

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
djaiswal83: Agreed. The behavior should be "default is legacy". So something like,

if (YARN_KILL_TOKEN_FIX) { 
  new logic;
} else {
  old logic;
}

djaiswal83 (changes requested):
Looks good to me. I have couple of suggestions.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
djaiswal83: @tsangz2013  please address this. token copy does not sound like the right way.

tsangz2013: updated the naming to `api_with_token`

byjiang1996 (approved):

djaiswal83 (approved):
LGTM

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -80,6 +86,12 @@ public class HadoopJobUtils {
   public static final String MAPREDUCE_JOB_OTHER_NAMENODES = "mapreduce.job.hdfs-servers";
   // MapReduce config for mapreduce job tags
   public static final String MAPREDUCE_JOB_TAGS = "mapreduce.job.tags";
+
+  public static final String YARN_KILL_VERSION = "yarn.kill.version";
+  // values of the yarn kill version flag
+  public static final String YARN_KILL_LEGACY = "legacy";
+  public static final String YARN_KILL_USE_API_WITH_TOKEN = "api_with_token";
```
sshardool, line 93: For most features an appropriately named boolean flag is adequate. For example in this case it could be something like the property `enable.advanced.yarn.app.kill` or `enable.yarn.kill.with.token` with true/false values.
That also makes a bit easier for someone to reason about the code in future and eliminate the flag when not needed anymore. 

For now, we can retain this property as-is as it will be removed in the near future.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -352,24 +364,50 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
...
+            }
+          });
+        } else {
+          findAndKillYarnApps(jobProps, log);
+        }
+      } catch (final Throwable t) {
+        log.warn("something happened while trying to kill all spawned jobs", t);
+      }
+    } else {
+      final String logFilePath = jobProps.getString(CommonJobProperties.JOB_LOG_FILE);
```
sshardool, line 389: Is the entire `else {}` block legacy code from before the current yarn-kill feature was added ?

sshardool (commented):
Thanks a lot for taking this up. Left a few comments.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: I am using string because it could be non-binary, e.g. when we substitute getAppFromLog with getAppWithApi, we can have a 3rd option.

good advice on naming

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: Affirmative, legacy code before my change.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -462,6 +523,38 @@ public static Set<String> findApplicationIdFromLog(final String logFilePath, fin
...
+   *   If the spark job is complete, it will return immediately, with a job not found on job tracker
+   * </pre>
+   */
+  public static void killJobOnCluster(final String applicationId, final Logger log,
+      final Props jobProps)
+      throws YarnException,
+      IOException {
+
+    final YarnConfiguration yarnConf = new YarnConfiguration();
+    final YarnClient yarnClient = YarnClient.createYarnClient();
```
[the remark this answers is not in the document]
tsangz2013: Yes direct copy from legacy code, so no optimization at all.

The suggestion is reflected in new code btw.

PR #3180, opened 2022-10-24
Allowing images versions in all statuses to run for dev pods

amielkelinkedin:
In order to expedite testing and development, any version of an image should be able to launch a pod, when enable.dev.pod is true.


utk-12:
Hi @amielkelinkedin 
Can we add why we need all the statuses to be returned here in the description or rather the motivation for this change ?

amielkelinkedin:
> Hi @amielkelinkedin Can we add why we need all the statuses to be returned here in the description or rather the motivation for this change ?
I believe I had update the header, @utk-12

byjiang1996 (dismissed):

ShellyWEI (approved):
LGTM! thanks for fixing this dev pod issue!

PR #3181, opened 2022-10-25
Add validateLdapGroup in UserManager interface

ShellyWEI:


byjiang1996 (approved):
lgtm

azkaban-common/src/main/java/azkaban/user/UserManager.java
```diff
@@ -44,6 +44,14 @@ public User getUser(String username, String password)
...
    */
   public boolean validateGroup(String group);
+  /**
+   * Check whether a given group is a valid ldap group account.
+   *
+   * @param group Ldap group account name
+   * @return True, if given group is a valid ldap group.
+   *         False, otherwise
+   * */
+  public boolean validateLdapGroup(String group);
```
amielkelinkedin, line 54: This is just searching a different LDAP domain than the query above, right?

PR #3182, opened 2022-10-25
Creating SQL tables and wrapper for project_flow_resource_recommendations

byjiang1996:
We will use project_flow_resource_recommendations table to cache and store resource recommendation for each azkaban flow.

Recommendation logic will come in future PRs.


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -802,6 +804,89 @@ public void uploadProjectProperties(final Project project,
...
+    }
+  }
+
+  @Override
+  public FlowRecommendation fetchFlowRecommendation(final int projectId, final String flowId)
+      throws ProjectManagerException {
+    logger.info("Fetching flow recommendation. ProjectId: " + projectId + ", FlowId: " + flowId);
+    try {
+      return this.dbOperator
+          .query(ProjectFlowRecommendationsResultHandler.SELECT_PROJECT_FLOW_RECOMMENDATION,
```
ShellyWEI, line 856: if this query does not find a match result, handler would return an empty list as you set in ProjectFlowRecommendationsResultHandler. So better have a check above get(index), otherwise, it would have IndexOutOfBoundsException.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -802,6 +804,89 @@ public void uploadProjectProperties(final Project project,
...
+              new ProjectFlowRecommendationsResultHandler(), projectId, flowId).get(0);
+    } catch (final SQLException e) {
+      logger.error("Error fetching flow recommendation", e);
+      throw new ProjectManagerException("Error fetching flow recommendation. ProjectId: " + projectId + ", FlowId: " + flowId, e);
+    }
+  }
+
+  @Override
+  public List<FlowRecommendation> fetchAllProjectFlowRecommendations(final Project project) throws ProjectManagerException {
```
ShellyWEI, line 865: I think actually fetchAllProjectFlowRecommendations and fetchAllFlowRecommendationsForProjects could have consist behavior. This fetchAllProjectFlowRecommendations could have return null, if there isn't a match for given project; while fetchAllFlowRecommendationsForProjects always return notNull even if not a match result. Either both null or both empty so it easier to use them.
Also adding comments for each method with returning values with empty/null/list of results in the interface, and with Exceptions.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
```diff
@@ -161,6 +162,37 @@ List<Flow> fetchAllProjectFlows(final Project project)
...
+  /**
+   * Fetches one particular flow recommendation.
+   */
+  FlowRecommendation fetchFlowRecommendation(int projectId, String flowId)
+      throws ProjectManagerException;
+
+  /**
+   * Fetches all flow recommendations for a given project
+   */
+  List<FlowRecommendation> fetchAllProjectFlowRecommendations(Project project)
```
ShellyWEI, line 187: Also I feel like passing through the entire Project all the way here could have the chance to increase the Java garbage collection burden so it might result in frequent gc to our fragile web server, as Project itslef contains too many informations. (Not sure about the numbers but I've already feel like our current preparing/loading process is too long)
If the purpose is only a project identifier, can we use project_id only perhaps?

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -802,6 +804,89 @@ public void uploadProjectProperties(final Project project,
     }
   }
 
+  @Override
+  public synchronized FlowRecommendation createFlowRecommendation(final int projectId, final String flowId)
```
ShellyWEI, line 808: And since createPod in K8sImpl also could have call createFlowRecommendation for existing project (if I understand it correctly), better to add duplicate check here, i,e. fetch by project_id, flow_id, if exists, update instead of insert.
If you plan something else, please ignore this one.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -802,6 +804,89 @@ public void uploadProjectProperties(final Project project,
...
+      final int numRowsInserted = this.dbOperator.transaction(insertFlowRecommendation);
+      if (numRowsInserted == 0) {
+        throw new ProjectManagerException("No flow recommendations have been inserted.");
+      }
+    } catch (final SQLException ex) {
+      logger.error(INSERT_FLOW_RECOMMENDATION + " failed.", ex);
+      throw new ProjectManagerException("Insert flow recommendation " + flowId + " for existing "
+          + "project failed. ", ex);
+    }
+    return fetchFlowRecommendation(projectId, flowId);
```
ShellyWEI, line 828: fetch also involve get DB. I think better to keep them (write and read) in a single transaction to avoid potential race condition.

azkaban-db/src/main/sql/create.project_flow_recommendations.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE project_flow_recommendations (
+  id                    INT         NOT NULL PRIMARY KEY AUTO_INCREMENT,
+  project_id            INT         NOT NULL,
+  flow_id               VARCHAR(128),
```
ShellyWEI, line 4: Why flow_id can be null?

ShellyWEI (commented):
Overall looks good to me. I have few comments.
And I think flowRecommendation is not specific, it's better to name with flowResourceRecommendation/flowPodsParamRecommendation, something like that.

azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -152,6 +155,22 @@ public void setFlows(final Map<String, Flow> flows) {
     this.flows = ImmutableMap.copyOf(flows);
   }
 
+  public FlowRecommendation getFlowRecommendation(final String flowId) {
+    if (this.flowRecommendations == null) {
```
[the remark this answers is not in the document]
byjiang1996: Good point!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: Agree!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: It will throw duplicate SQLException here so duplicate check is not necessary.
But ideally no duplication will occur b/c I will lock and read in-memory project object whenever operating flow recommendations.

azkaban-db/src/main/sql/create.project_flow_recommendations.sql
byjiang1996: Good point! Copied from our project_flows table. Let me mark mine as not null.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: It should be fine
1. That write and read logic is copied from createNewProject function in the same file.
2. Every flow recommendation table read and write will be guarded by a lock in outer java classes.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
byjiang1996: 1. "Passing entire Project" is coming from the existing fetchAllFlowsForProjects logic and creating one more list of projectIds in addition to existing list of projects in InMemoryProjectCache seems an overkill.
2. "Our current preparing/loading process is too long": currently it is b/c we have too many projects and do not get retentioned/deleted since most of them are really old. I don't think using projectIds here will reduce load speed a lot b/c projectIds should be generated in outer level as well from projects. It is the same effort: difference is where to get id from project: in outer class or ProjectLoader class but it will not affect performance a lot.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: Hmmm, I found fetchAllProjectFlowRecommendations is not used at all: removed.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
ShellyWEI: Agree with extra mapping is an overkill. My opinion is from garbage collection point of view. I used to work on a project, we pass in a complex object into multiple layer and this is on critical path, the memory usage increased exponentially and the service crashed frequently since then. Because in each method of that call tree, such huge object would occupy a lot memory (java is copy-by-value instead of copy-by-reference) and such method is being called so frequently. It then ends up triggering heavy JGC. I am sensitive to see such and would think ahead if we really need whole object to pass through.
Anyway I agree it might not be a big deal now. However, when the projects keep growing, the headache would become more serious as I think we're already in a situation to clean up unused projects sometime or so. So what is already existing, might not be a best way to follow. Let always keep in critical mind for things: )

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
ShellyWEI: sounds good!

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
ShellyWEI: I am under the impression that only duplicate on Primary key would throw that exception? Correct me if I am wrong.
I am not seeing the logic now but I trust you will handle this well.

ShellyWEI (approved):
Thank you for addressing my comments! This is looking good to me.

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: I'm capturing SQLException (base sql related exception). It should be thrown and caught for most bad sql issues.

azkaban-common/src/main/java/azkaban/project/ProjectLoader.java
byjiang1996: After offline discussion, copy-by-value in java does not mean deep copy the object but means copy the object pointer/reference for non-primitive objects so it is lightweight no matter how the huge the object itself is.
https://www.infoworld.com/article/3512039/does-java-pass-by-reference-or-pass-by-value.html

PR #3183, opened 2022-10-25
Add KubernetesVPARecommender class to interact with k8s VPA recommender on providing flow resource recommendations

byjiang1996:
Comments in order:
1. Auto generate VPA java client code via k8s CRD generator (No need to go through this commit in detail).
2. Add k8s label selector support to VPA client so that each VPA object can map to one azkaban flow: e.g. VPA object set labelSelector as "flow-vpa=fc-vpa-azkaban-1", each azkaban pod launched by one azkaban flow set its pod label as "flow-vpa=fc-vpa-azkaban-1"; after that, VPA object is able to know which azkaban flow each k8s pod maps to.
3. Add KubernetesVPARecommender class to interact with k8s VPA recommender: QueryOrCreate VPA object for the Azkaban flow and set up labelSelector, max&min resource limit appropriately; apply multiplier to VPA resource recommendation to guarantee a safety margin; apply timeout in case if any operation in getFlowContainerRecommendedRequests function is stuck


az-core/src/main/java/azkaban/Constants.java
```diff
@@ -727,6 +727,13 @@ public static class ContainerizedDispatchManagerProperties {
     public static final String KUBERNETES_SERVICE_CREATION_TIMEOUT_MS = KUBERNETES_SERVICE_PREFIX +
         "creation.timeout.ms";
 
+    // Kubernetes vpa related properties
+    public static final String KUBERNETES_VPA_PREFIX = AZKABAN_KUBERNETES_PREFIX + "vpa.";
+    public static final String KUBERNETES_VPA_MAX_ALLOWED_NO_RECOMMENDATION_SINCE_CREATION_S =
+        KUBERNETES_VPA_PREFIX + "max.allowed.no.recommendation.since.creation.s";
```
amielkelinkedin, line 733: Any reason for the extra s at the end, rather than just sec?

byjiang1996: It represents "second". I think what you said makes sense... Let me rename it as "sec"

azkaban-common/src/main/java/io/kubernetes/autoscaling/models/V1beta2VerticalPodAutoscalerStatusRecommendationContainerRecommendations.java
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  public V1beta2VerticalPodAutoscalerStatusRecommendationContainerRecommendations putTargetItem(String key, Object targetItem) {
+    this.target.put(key, targetItem);
+    return this;
+  }
+
+   /**
+   * Recommended amount of resources. Observes ContainerResourcePolicy.
+   * @return target
```
amielkelinkedin, line 117: Are the comments neccesary if the value off ApiModel is the same as the comment?

azkaban-common/src/main/java/io/kubernetes/autoscaling/models/V1beta2VerticalPodAutoscalerStatusRecommendation.java
```diff
@@ -0,0 +1,104 @@
...
+    sb.append("    containerRecommendations: ").append(toIndentedString(containerRecommendations)).append("\n");
+    sb.append("}");
+    return sb.toString();
+  }
+
+  /**
+   * Convert the given object to string with each line indented by 4 spaces
+   * (except the first line).
+   */
+  private String toIndentedString(Object o) {
```
amielkelinkedin, line 96: If we're going to use these methods several places, can we put them in a utility?

byjiang1996: See the description. The first commit can be ignored. Those VPA client code are auto-generated .

azkaban-common/src/main/java/azkaban/executor/container/KubernetesVPARecommender.java
```diff
@@ -0,0 +1,191 @@
...
+            return null;
+          }
+
+          // Unknown errors
+          throw e;
+        }
+
+        try {
+          // If VPA object is newly created: no recommendation yet and should return null.
+          if (vpaObject.getMetadata() == null || vpaObject.getMetadata().getCreationTimestamp() == null ||
```
amielkelinkedin, line 135: Can we put this outside the try?

byjiang1996: Makes sense. Will do.

azkaban-common/src/main/java/io/kubernetes/autoscaling/models/V1beta2VerticalPodAutoscalerStatusRecommendationContainerRecommendations.java
byjiang1996: See the description. The first commit can be ignored. Those VPA client code are auto-generated .

az-core/src/main/java/azkaban/Constants.java
byjiang1996: Done

azkaban-common/src/main/java/azkaban/executor/container/KubernetesVPARecommender.java
byjiang1996: Done

amielkelinkedin (approved):

amielkelinkedin (approved):

PR #3186, opened 2022-10-28
Add timeout for yarnClient to avoid the api call hang

NancyXie2022:
Testing done:
Try to kill a yarn app which is already killed, the api call returns quickly instead of hang there for 15 mins(which is the current behavior)


amielkelinkedin (approved):
Thank you so much @NancyXie2022

tsangz2013 (approved):
lgtm

svnarumugam (approved):
Change looks good to me. Thank you @NancyXie2022 !

Issue #3087, opened 2022-05-04
Property name contains two dots


xandris:
https://github.com/azkaban/azkaban/blob/767240e8b0db82c3b558b583b6d7de47e23c0783/az-core/src/main/java/azkaban/Constants.java#L672

Minor nit; effective constant value is `azkaban.kubernetes.flow.container..name`

PR #3187, opened 2022-10-31
#3087 Removed duplicate dot from constant value

jaybob007:
Removed extra dot from the KUBERNETES_FLOW_CONTAINER_NAME constant.


svnarumugam (approved):
Thank you for making this change @jaybob007. Could you please add notes on how you identified this bug?

PR #3189, opened 2022-10-31
Yarn App Killing: default to scan log for application ID and read tokens and kill

tsangz2013:
**Why is this needed**: as communicated, the `getApplication()` API should not be used in a potential high QPS scenario. Thus we default to scan log for application ID, and then load tokens and kill the application.

**What's done**: update the feature branching, let `log-scanning + token loading` be the default option.

**Tests done**: tested in containerization pod, the flow and the yarn app both got killed successfully.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -366,48 +366,25 @@ public class HadoopJobUtils {
     properties.putAll(jobProps.getFlattened());
 
     // todo: use feature flag, default to use legacy mode
-    String yarnKillVersion = jobProps.getString(YARN_KILL_VERSION, YARN_KILL_LEGACY);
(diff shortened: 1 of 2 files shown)


NancyXie2022 (commented):
Why don't you just change this single line?
https://github.com/azkaban/azkaban/blob/master/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java#L418

to 

Set<String> jobAppIDsToKill = findApplicationIdFromLog(logFilePath, log);;

tsangz2013:
> Why don't you just change this single line? https://github.com/azkaban/azkaban/blob/master/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java#L418
> 
> to
> 
> Set jobAppIDsToKill = findApplicationIdFromLog(logFilePath, log);;

Changing this line will not suffice, it will need us to set `YARN_KILL_VERSION` property to `YARN_KILL_USE_API_WITH_TOKEN` for all the jobs (or set it in `common.properties` file), which is more complicate.

Using this way, the default behavior is clean and clear.

NancyXie2022 (approved):
LGTM

PR #3190, opened 2022-11-01
Allowing stable state images to be run

amielkelinkedin:
Allows Stable state images to be run. Images reach a stable state when they have been in active state and a new image is Activated.


svnarumugam (approved):
Thanks @amielkelinkedin for improving the dev pod testing experience with this change

PR #3194, opened 2022-11-02
CleanupManager - find the flow executions whose yarn apps needs to be killed

tsangz2013:
**Why need this**: we want to find out the flow executions whose yarn apps needs to be killed.

**What's in this**: creates a method to find EXECUTION_STOPPED executions, and extras a method to find dangling pods of the terminated executions.

**Tests done**: tested in the test web-server instance, the functions works well

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -17,6 +17,7 @@ package azkaban.executor.container;
 
 import static azkaban.Constants.ConfigurationKeys.*;
 import static azkaban.Constants.ContainerizedDispatchManagerProperties;
+import static azkaban.utils.YarnUtils.YARN_CONF_DIRECTORY_PROPERTY;
 
 import azkaban.Constants.FlowParameters;
 import azkaban.cluster.Cluster;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -194,17 +221,70 @@ public void cleanUpContainersInTerminalStatuses() {
...
+        iii. kill these yarn applications
+     */
+    logger.info("cleanUpDanglingYarnApplications start ");
+
+    Set<Integer> executionStoppedFlows = getExecutionStoppedFlows();
+    logger.info("Get executionStoppedFlows: " +
+        executionStoppedFlows.stream().map(Object::toString).collect(Collectors.joining(",")));
+
+    // get those flows terminated but containers are still alive (failed to properly killed)
+    Set<Integer> toBeCleanedContainers = getContainersOfTerminatedFlows();
```
tsangz2013, line 278: This part of finding `flows terminated but containers are still alive` is originally used in the other schedule method `cleanUpContainersInTerminalStatuses`.
Theoretically whenever we found these `containers mistakenly still alive` we should kill these containers, while in this PR I am focusing on cleaning up the yarn apps, and don't want to interfere with others.
We can circle back to discuss whether can improve this part later.

PR #3196, opened 2022-11-03
CleanupManager - get to-be-killed yarn applications from execution IDs

tsangz2013:
**Why need this**: we want to find out the yarn applications to be killed for those flow executions need to be killed.

**What's in this**: use a method in `YarnUtils` call getApplciations() API with tags of the set of execution-IDs to find these applications

**Tests done**: tested in the test web-server instance, the functions works well

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -17,7 +17,10 @@ package azkaban.executor.container;
 
 import static azkaban.Constants.ConfigurationKeys.*;
 import static azkaban.Constants.ContainerizedDispatchManagerProperties;
+import static azkaban.Constants.FlowProperties.AZKABAN_FLOW_EXEC_ID;
 import static azkaban.utils.YarnUtils.YARN_CONF_DIRECTORY_PROPERTY;
+import static azkaban.utils.YarnUtils.createYarnClient;
+import static azkaban.utils.YarnUtils.getAllAliveAppReportsByExecIDs;
 
 import azkaban.Constants.FlowParameters;
 import azkaban.cluster.Cluster;
@@ -34,6 +37,7 @@ import azkaban.utils.Props;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.security.PrivilegedExceptionAction;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,7 +293,39 @@ public void cleanUpDanglingYarnApplications() {
...
     // TODO: For each of yarn clusters: get applications and kill those in the above set
+    for (Entry<String, Cluster> entry : this.allClusters.entrySet()) {
+      logger.info("clean up yarn applications in cluster:" + entry.getValue().clusterId);
+      cleanUpYarnApplicationsInCluster(toBeCleanedContainers, entry.getValue());
+    }
+  }
+
+  private void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers,
+      Cluster cluster) {
+    org.apache.log4j.Logger apacheLogger =
```
NancyXie2022, line 310: why not import this class to make the code clean?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,7 +293,39 @@ public void cleanUpDanglingYarnApplications() {
...
+    }
+
     // TODO: For each of yarn clusters: get applications and kill those in the above set
+    for (Entry<String, Cluster> entry : this.allClusters.entrySet()) {
+      logger.info("clean up yarn applications in cluster:" + entry.getValue().clusterId);
+      cleanUpYarnApplicationsInCluster(toBeCleanedContainers, entry.getValue());
+    }
+  }
+
+  private void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers,
```
NancyXie2022, line 308: I think we need to make this whole class in the async path and gate it by feature flag

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: because the `Logger` class collides with existing `org.slf4j.Logger`

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: Async logic is in the following PRs

NancyXie2022: will this impact CleanupManager if you don't put it in async? or at least we can use feature flag to gate it.

NancyXie2022: I mean if you merge this PR

tsangz2013: this method is not the one we do `schedule` on, the `cleanUpDanglingYarnApplications` will be scheduled asynchronously at a cadence

NancyXie2022 (approved):

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: and at this moment (this PR) these logics are not invoked at all, will start calling the logic at the final PR

PR #3197, opened 2022-11-07
Add OOMKilled metrics and VPARecommenderFail metrics

byjiang1996:
1. Record number of flow executions stopped due to app container OOM killed
2. Record number of vpa recommender get recommendation failure


azkaban-common/src/main/java/azkaban/metrics/ContainerizationMetricsImpl.java
```diff
@@ -127,6 +129,12 @@ public void markPodAppFailure() {
   @Override
   public void markExecutionStopped() { executionStopped.mark(); }
 
+  @Override
+  public void markOOMKilled() { executionStopped.mark(); }
```
tsangz2013, line 133: These 2 are identical now, any plan to improve granularity? e.g. for OOMKilled, mark exec-stop as well as a new metrics `oomKilled.mark()`?

byjiang1996: No plan yet. There will be a huge amount of reasons why a pod moves to Execution_stopped. It is beyond the scope of this pr to improve granularity here

tsangz2013 (dismissed):
lgtm

tsangz2013 (approved):
lgtm

PR #3198, opened 2022-11-07
Use ConcurrentHashMap to store in-memory project flowresourcerecommen…

byjiang1996:
…dation map

After discussing internally, there is no need to use "synchronize(project)" to ensure the locking which is quite heavy operation (may block other thread for a long time: project upload will take a long time for huge project zips). Using ConcurrentHashMap is more lightweight and better.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -807,7 +807,7 @@ public class JdbcProjectImpl implements ProjectLoader {
   @Override
   public synchronized FlowResourceRecommendation createFlowResourceRecommendation(final int projectId, final String flowId)
       throws ProjectManagerException {
-    logger.info("Creating flow resource recommendation " + flowId);
+    logger.info("Creating flow resource recommendation. ProjectId: " + projectId + ", FlowId: " + flowId);
     final String INSERT_FLOW_RESOURCE_RECOMMENDATION =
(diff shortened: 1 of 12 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -821,9 +821,10 @@ public synchronized FlowResourceRecommendation createFlowResourceRecommendation(
         throw new ProjectManagerException("No flow resource recommendations have been inserted.");
       }
     } catch (final SQLException ex) {
-      logger.error(INSERT_FLOW_RESOURCE_RECOMMENDATION + " failed.", ex);
-      throw new ProjectManagerException("Insert flow resource recommendation " + flowId + " for existing "
-          + "project failed. ", ex);
+      // Possibly failed due to duplicate key. If not, fetchFlowResourceRecommendation will
+      // return another exception back.
+      logger.warn("Insert flow resource recommendation " + flowId + " for existing "
```
amielkelinkedin, line 826: Thank you for adding flow ID. Is there anything else you could add here to help the person who would need to debug this in the future?

amielkelinkedin (approved):

amielkelinkedin (approved):

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
byjiang1996: Makes sense. Thanks for bringing this up!

PR #3200, opened 2022-11-08
CleanupManager - asynchronously proxy as user and kill the yarn applications

tsangz2013:
**Why need this**: After get the list of Yarn Applications, we want to kill them in parallel, and report the summarized killing result.

**What's in this**: use a `FixedThreadPool` to handle all the applications' killings: in each thread create proxy user using yarn-user's name, doAs() to create yarnClient and call killApplication() API.

**Tests done**: tested in the test web-server instance with some flows having yarn applications running, the killing and reporting works well

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -323,7 +323,7 @@ public class ContainerCleanupManager {
     }
   }
 
-  private void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers,
+  void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers,
       Cluster cluster) {
     org.apache.log4j.Logger apacheLogger =
         org.apache.log4j.Logger.getLogger(ContainerCleanupManager.class);
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -17,7 +17,6 @@ package azkaban.executor.container;
 
 import static azkaban.Constants.ConfigurationKeys.*;
 import static azkaban.Constants.ContainerizedDispatchManagerProperties;
-import static azkaban.Constants.FlowProperties.AZKABAN_FLOW_EXEC_ID;
 import static azkaban.utils.YarnUtils.YARN_CONF_DIRECTORY_PROPERTY;
 import static azkaban.utils.YarnUtils.createYarnClient;
 import static azkaban.utils.YarnUtils.getAllAliveAppReportsByExecIDs;
@@ -34,10 +33,10 @@ import azkaban.executor.Status;
 import azkaban.metrics.ContainerizationMetrics;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import azkaban.utils.YarnUtils;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import java.security.PrivilegedExceptionAction;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -45,15 +44,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
-import org.apache.hadoop.security.UserGroupInformation;
-import org.apache.hadoop.security.token.Token;
-import org.apache.hadoop.yarn.api.records.ApplicationId;
 import org.apache.hadoop.yarn.api.records.ApplicationReport;
 import org.apache.hadoop.yarn.client.api.YarnClient;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -325,9 +343,45 @@ private void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers
...
     }
 
-    // todo: kill the apps asynchronously
+    // Use a fix thread pool to concurrently kill yarn apps
+    Map<String, Boolean> appsSuccessfulKilled = new ConcurrentHashMap<>();
+    aliveApplications.forEach(
+        app -> appsSuccessfulKilled.put(app.getApplicationId().toString(), false));
+
+    ExecutorService yarnKillThreadPool = Executors.newFixedThreadPool(this.yarnAppKillParallelism);
+    aliveApplications.forEach(app ->
```
NancyXie2022, line 352: I think getApplication also should be in the async path

tsangz2013: The flow is this: main thread get all to-kill yarn app, then use a thread poll to kill each in async manner.
I cannot image what can we do making get-yarn-app async

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -325,9 +343,45 @@ private void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers
       return;
     }
 
-    // todo: kill the apps asynchronously
+    // Use a fix thread pool to concurrently kill yarn apps
+    Map<String, Boolean> appsSuccessfulKilled = new ConcurrentHashMap<>();
+    aliveApplications.forEach(
+        app -> appsSuccessfulKilled.put(app.getApplicationId().toString(), false));
+
+    ExecutorService yarnKillThreadPool = Executors.newFixedThreadPool(this.yarnAppKillParallelism);
```
byjiang1996, line 351: Two questions:
1. Will there be two cleanUpYarnApplicationsInCluster instances running in parallel? If yes, we will spin up too many threads for yarn kill purpose: this.yarnAppKillParallelism * number_of_cleanUpYarnApplicationsInCluster_parallelism.
2. Can you move yarnKillThreadPool to a global var instead of a local var? I feel like creating a thread pool is heavy.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -438,10 +493,8 @@ public void start() {
         this.executionCleanupIntervalMin, TimeUnit.MINUTES);
     this.cleanupService.scheduleAtFixedRate(this::cleanUpContainersInTerminalStatuses, 0L,
         this.containerCleanupIntervalMin, TimeUnit.MINUTES);
-
-    // TODO: currently run-once for development; change to a fixed schedule after development
-    //  complete
-    this.cleanupService.schedule(this::cleanUpDanglingYarnApplications, 0L, TimeUnit.MINUTES);
+    this.cleanupService.scheduleAtFixedRate(this::cleanUpDanglingYarnApplications, 0L,
```
byjiang1996, line 496: Could you please write a unit test for cleanUpDanglingYarnApplications function?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: 1. No it won't, looking at line 322, `cleanUpYarnApplicationsInCluster` will be called sequentially for multiple clusters.
2. good point, will do

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: sure, will do

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
byjiang1996: Ah, sorry. Let me elaborate more on the first question.
Will two cleanUpDanglingYarnApplications instances be executed in parallel (if yes, cleanUpYarnApplicationsInCluster will be executed in parallel as well)? cleanUpDanglingYarnApplications is called by scheduleAtFixedRate. Suppose the first run of cleanUpDanglingYarnApplications lasts for 1 day, and then in the middle of that 1d, will scheduleAtFixedRate call cleanUpDanglingYarnApplications again or simply wait for the 1st run to be completed?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -74,14 +72,21 @@ public class ContainerCleanupManager {
...
   private static final Duration DEFAULT_STALE_EXECUTION_CLEANUP_INTERVAL = Duration.ofMinutes(10);
   private static final Duration DEFAULT_STALE_CONTAINER_CLEANUP_INTERVAL = Duration.ofMinutes(60);
+  private static final Duration DEFAULT_YARN_APP_CLEANUP_INTERVAL = Duration.ofMinutes(10);
   private static final int DEFAULT_AZKABAN_MAX_FLOW_DISPATCHING_MINS = 10;
   private static final int DEFAULT_AZKABAN_MAX_FLOW_PREPARINGING_MINS = 15;
   private static final int DEFAULT_AZKABAN_MAX_FLOW_RUNNING_MINS = 10 * 24 * 60; // 10 days
   private static final int DEFAULT_AZKABAN_MAX_FLOW_KILLING_MINS = 15;
   private static final int DEFAULT_AZKABAN_MAX_FLOW_EXEC_STOPPED_MINS = 15;
+  private static final int DEFAULT_AZKABAN_YARN_BATCH_KILL_TIMEOUT_IN_MINUTE = 10;
+  private static final int DEFAULT_AZKABAN_YARN_BATCH_KILL_PARALLELISM = 5;
```
NancyXie2022, line 82: why 5? cannot we kill all of the them in one thread?

tsangz2013: adding parallelism killing a large number of jobs brings efficiency + avoid 1 blocking failure

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: Yes as we discussed this is not the CM "main" thread, this thread is spawned every time CM schedules this method at fixed rate, so it won't jam the others or the "main" thread

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: From the method signature itself:
`If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute.`

azkaban-common/src/test/java/azkaban/executor/container/ContainerCleanupManagerTest.java
```diff
@@ -222,4 +236,64 @@ public void testGetExecutionStoppedFlowsFail() throws ExecutorManagerException {
...
+    YarnClient mockClient = mock(YarnClient.class);
+    when(YarnUtils.createYarnClient(any(), any())).thenReturn(mockClient);
+    when(YarnUtils.getAllAliveAppReportsByExecIDs(any(), any(), any()))
+        .thenThrow(new IOException("ops"));
+//    // this will never be called
+//    PowerMockito.doNothing().when(YarnUtils.class,
+//        "killApplicationAsProxyUser", any(), any(), any());
+
+    this.cleaner.cleanUpYarnApplicationsInCluster(
+        ImmutableSet.of(), new Cluster("abc", new Props()));
```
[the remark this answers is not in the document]
tsangz2013: I found it difficult to verify on static class, as this class has multiple static methods being called, and the current `verifyStatic` method can only verify whether 1 class is called, cannot verify on static method level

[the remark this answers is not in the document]
byjiang1996: Oh, it makes sense! Thanks for updating!

byjiang1996 (approved):
LGTM

NancyXie2022 (approved):
LGTM

PR #3201, opened 2022-11-09
CleanupManager - yarn app killing features wrap-ups

tsangz2013:
**What's in this** 
1. when robin not enabled, use default `hadoop-conf-dir-path` for yarn cluster connection
2. use `yarn.application.cleanup.interval.min` as feature flag: non-positive value won't start the yarn-kill schedule
3. handle unexpected throwables for all methods, so as 1 failure won't terminate the following runs

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -178,8 +178,6 @@ public class ContainerCleanupManager {
             new Pair<>(Duration.ofMinutes(maxExecStoppedValidity), UPDATE_TIME))
         .build();
 
-    // get yarn config for instance without robin enabled
-
     // get config in instance using robin, so as can connect to multiple yarn cluster RMs
     for (Entry<String, Cluster> entry : this.clusterRouter.getAllClusters().entrySet()) {
       logger.info("Now printing cluster named: " + entry.getKey());
@@ -196,6 +194,7 @@ public class ContainerCleanupManager {
       }
     }
     if (allClusters.isEmpty()) {
+      // get yarn config for instance without robin enabled
       String hadoopConfDir = azkProps.getString(HADOOP_CONF_DIR_PATH, "");
       if (hadoopConfDir.isEmpty()) {
         logger.warn("No Cluster config or default hadoop-conf-dir-path specified,"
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -493,8 +520,14 @@ public void start() {
         this.executionCleanupIntervalMin, TimeUnit.MINUTES);
     this.cleanupService.scheduleAtFixedRate(this::cleanUpContainersInTerminalStatuses, 0L,
         this.containerCleanupIntervalMin, TimeUnit.MINUTES);
-    this.cleanupService.scheduleAtFixedRate(this::cleanUpDanglingYarnApplications, 0L,
-        this.yarnAppCleanupIntervalMin, TimeUnit.MINUTES);
+    if (this.yarnAppCleanupIntervalMin <= 0) {
```
byjiang1996, line 523: Do we need <= 0 check? I think if we want to disable this feature, simply leave yarnAppCleanupIntervalMin==0 is enough

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -295,31 +314,36 @@ public void cleanUpDanglingYarnApplications() {
...
-    // For each of yarn clusters: find applications of the above executionIDs and kill them
-    for (Entry<String, Cluster> entry : this.allClusters.entrySet()) {
-      logger.info("clean up yarn applications in cluster:" + entry.getValue().clusterId);
-      cleanUpYarnApplicationsInCluster(toBeCleanedContainers, entry.getValue());
+      // For each of yarn clusters: find applications of the above executionIDs and kill them
+      for (Entry<String, Cluster> entry : this.allClusters.entrySet()) {
+        logger.info("clean up yarn applications in cluster:" + entry.getValue().clusterId);
+        cleanUpYarnApplicationsInCluster(toBeCleanedContainers, entry.getValue());
+      }
+    } catch (Throwable t) {
```
byjiang1996, line 344: If we do not add catch here, what will happen? Will it terminate the start() function immediately?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: yes while it's of no harm and tolerant more ;)

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: it will fail this call and will no longer schedule any of the future call of `scheduleAtFixedRate` method

byjiang1996 (approved):
LGTM. Thanks for the change!

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
NancyXie2022, line 523: when it is <= 0, it also means the feature is not enabled, right?

tsangz2013: yes <= 0 is disabled

NancyXie2022: then in this case, your warning message is confused,  "invalid time interval " + this.yarnAppCleanupIntervalMin + ", please correct value "

NancyXie2022 (approved):

PR #3203, opened 2022-11-16
Timeboxing fetchLogs function to accelerate the log fetching

byjiang1996:
FlowStartTime and FlowEndTime timeboxing can help accelerate log search/query especially when logs are DB-indexed/DB-sharded by timestamp: i.e. we can do this after this PR to fetch logs: `select * from logs where execId=<execId> and timestamp > FlowStartTime and timestamp < FlowEndTime` instead of `select * from logs where execId=<execId>` before.


utk-12:
Thank you for adding this! Could you please elaborate on the PR as to how exactly adding both timestamps helps? 
It's evident that the DB is indexed by timestamp. How exactly is having both the start and end helping us?
LGTM otherwise!

NancyXie2022 (approved):

utk-12 (approved):

PR #3204, opened 2022-11-17
CleanupManager Add YarnApplicationKillFail metrics

tsangz2013:
**Why need this**: emit metrics of counter of yarn getApplication failure and counter of failed-to-killed yarn application, so as can create dashboard and alert amongst this

**Test done**: deployed in our test web server, and verified the metrics was emitted from ingraph

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -366,6 +366,7 @@ public class ContainerCleanupManager {
     } catch (Exception e) {
       logger.error("fail to get yarn applications by execution IDs from cluster "
           + cluster.getClusterId() + ", exiting", e);
+      containerizationMetrics.markYarnGetApplicationsFail();
       return;
     }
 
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -363,7 +364,8 @@ void cleanUpYarnApplicationsInCluster(Set<Integer> toBeCleanedContainers,
           aliveApplications.stream().map(app -> app.getApplicationId().toString())
               .collect(Collectors.joining(",")));
     } catch (Exception e) {
-      logger.error("fail to get yarn applications by execution IDs", e);
+      logger.error("fail to get yarn applications by execution IDs from cluster "
```
byjiang1996, line 367: Do we need to emit any similar metrics here?
If thread failed due to some random reason here, markYarnApplicationKillFail will not be called.

byjiang1996 (approved):
LGTM. Left a minor comment.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: good point, adding one now

byjiang1996 (approved):

PR #3205, opened 2022-11-19
Implement MissScheduleManager to support operations when miss schedule happens

ShellyWEI:
Two major tasks inside missScheduleManager:
- send email notification
- back execution This PR only includes the thread pool and task creation. Each tasks would be taken from a task queue to execute on their own.
MissScheudleManager is controled by a feature flag: azkaban.miss.schedule.manager.enabled should help role out the feature in production clusters. Currently No real task would be put into task queue and the implementation would be in the subsequent PRs.


azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+  @VisibleForTesting
+  protected MissedScheduleTaskQueueNode peekFirstTask() {
+    return this.missedScheduleTaskQueue.peek();
+  }
+
+  public void start() {
+    // Start the thread pool
+    for (int i = 0; i < threadPoolSize; i++) {
+      MissedSchedulesOperationTask task = new MissedSchedulesOperationTask(i, sleepIntervalInMs);
+      this.backgroundTasksList.add(task);
```
byjiang1996, line 133: If MISSED_SCHEDULE_MANAGER_ENABLED is false, backgroundTasksList is null. Will throw NPE here.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+      if (threadPoolSize <= 0) {
+        LOG.warn("An invalid thread pool size is passed in: " + threadPoolSize + ", assuming that the service "
+            + "is disabled, will not start any thread to do missed schedules operations.");
+      }
+      this.missedSchedulesExecutor = null;
+      this.backgroundTasksList = null;
+      return;
+    }
+    this.missedSchedulesExecutor =
+        Executors.newFixedThreadPool(threadPoolSize, new DaemonThreadFactory("MissedSchedulesTask_thread"));
```
ypadron-in, line 83: When naming these threads there seem to be convention in the Azkaban code of prefixing them with `azk-` and separating words with `-`. The pattern looks like this: "`azk-`[use-case]`-pool-`[thread-number]". Here are a few examples:
```
"azk-metric-pool-%d"
"azk-callback-pool-%d"
"azk-watch-pool-%d"
```
Can we use the same pattern here?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+
+  @Inject
+  public MissedSchedulesManager(final Props azkProps,
+      final ProjectManager projectManager,
+      final Emailer emailer) {
+    this.projectManager = projectManager;
+    this.emailer = emailer;
+    this.sleepIntervalInMs = azkProps.getLong(Constants.MISSED_SCHEDULE_THREAD_IDLE_TIME, DEFAULT_THREAD_IDLE_TIME_IN_MS);
+    this.threadPoolSize = azkProps.getInt(Constants.MISSED_SCHEDULE_THREAD_POOL_SIZE, DEFAULT_THREAD_POOL_SIZE);
+    if (!azkProps.getBoolean(Constants.MISSED_SCHEDULE_MANAGER_ENABLED, false) || threadPoolSize <= 0) {
```
ypadron-in, line 73: Why having a feature flag property if the same result can be achieved by setting thread pool size to a negative value? Why not fail initialization of MissedSchedulesManager if the thread pool size configured is negative? What's the advantage of allowing negative values?

azkaban-common/src/main/java/azkaban/utils/DaemonThreadFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public DaemonThreadFactory(String threadNamePrefix) {
+    this.threadNumber = new AtomicInteger(1);
+    this.namePrefix = threadNamePrefix;
+  }
+
+  @Override
+  public Thread newThread(Runnable r) {
+    Thread t = new Thread(r, namePrefix + "-t" + threadNumber.getAndIncrement());
+    t.setDaemon(true);
```
ypadron-in, line 38: Could you use `this.threadNumber` and `this.namePrefix`?
Maybe the **Save actions** (https://github.com/azkaban/azkaban/blob/master/CONTRIBUTING.md#style-guides) plugin isn't activated in the IDE. Could you please check and activate if disabled?

ypadron-in, line 38: It seems the name of these threads will look like: `MissedSchedulesTask_thread-t1`. Can we please follow general naming conventions for threads like explained before.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+   * 2. execute back execution if enabled feature flag (TODO)
+   * when a missed schedule happens.
+   * */
+  private class MissedSchedulesOperationTask implements Runnable {
+    private Logger logger;
+    private final String threadId;
+    private boolean isRunning;
+    private long sleepIntervalInMs;
+    public MissedSchedulesOperationTask(int threadId, long sleepIntervalInMs) {
+      this.threadId = "MissedSchedulesTask_thread_" + threadId;
```
ypadron-in, line 169: Can we please use general naming conventions for threads like explained before?

ypadron-in (changes requested):

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public void start() {
+    // Start the thread pool
+    for (int i = 0; i < threadPoolSize; i++) {
+      MissedSchedulesOperationTask task = new MissedSchedulesOperationTask(i, sleepIntervalInMs);
+      this.backgroundTasksList.add(task);
+      // No need to maintain a list of task futures, since the life-cycles of the threads are maintained through
+      // the Executors thread pool
+      Future future = this.missedSchedulesExecutor.submit(task);
```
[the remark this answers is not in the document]
ypadron-in: For example, it is possible to run tasks in the same way flow threads are managed in FlowRunnerManager (https://github.com/azkaban/azkaban/blob/master/azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunnerManager.java#L349)? Of course we would need a metric to detect when the number of threads becomes insufficient and needs to be increased.

azkaban-common/src/test/java/azkaban/scheduler/MissedSchedulesManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    _missedSchedulesManager = new MissedSchedulesManager(azkProps, projectManager, emailer);
+  }
+
+  @After
+  public void cleanup() throws InterruptedException {
+    _missedSchedulesManager.stop();
+  }
+
+  @Test
+  public void testAddSchedule() {
```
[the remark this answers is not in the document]
ShellyWEI: Emailer functionality should be covered in EmailerTest not inside MissedSchedulerManager and same with ExecuteFlowAction. I think the test is only cover the manager would receive and execute task is enough. I will do end2end test to make sure the function when the second part PR (adding tasks) published.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  /**
+   * Put timestamps for Missed executions along with email notifications into Task Queue.
+   *
+   * @param missedScheduleTimesInMs, timestamps of missed schedule
+   * @param action, execute flow action
+   * @param backExecutionEnabled, a schedule config from user, the default is false
+   * */
+  public boolean addMissedSchedule(final List<Long> missedScheduleTimesInMs,
```
sshardool, line 83: Where is this invoked and will the users get an email for every missed schedule?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,216 @@
...
+        Emailer emailer,
+        List<String> emailRecipients,
+        ExecuteFlowAction executeFlowAction) {
+      this.emailer = emailer;
+      this.emailRecipients = emailRecipients;
+      this.executeFlowAction = executeFlowAction;
+      final String missScheduleTimestampToDate = formatTimestamps(missedScheduleTimes);
+      MessageFormat messageFormat = new MessageFormat("This is autogenerated email to notify to flow owners that"
+          + " the flow {0} in project {1} has scheduled the executions on {2} but failed to trigger on time. "
+          + "Back execution will start soon as you enabled the config.");
```
sshardool, line 183: Is this designed to handle the `backExecution` in some form, and what config is this referring to ?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,216 @@
...
+    public MissedSchedulesOperationTask(List<Long> missedScheduleTimes,
+        Emailer emailer,
+        List<String> emailRecipients,
+        String projectName,
+        String flowName) {
+      this.emailer = emailer;
+      this.emailRecipients = emailRecipients;
+      final String missScheduleTimestampToDate = formatTimestamps(missedScheduleTimes);
+      MessageFormat messageFormat = new MessageFormat("This is autogenerated email to notify to flow owners that"
+          + " the flow {0} in project {1} has scheduled the executions on {2} but failed to trigger on time.");
```
sshardool, line 168: The 2 constructors are nearly identical. Please check if they can be merged or at least refactor the email message creation into a single place/method.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
sshardool, line 83: Why is the return type a boolean, shouldn't it be throwing exceptions in case of any errors anyway?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,216 @@
...
+            new MissedSchedulesOperationTask(
+                missedScheduleTimesInMs,
+                this.emailer,
+                emailRecipients,
+                projectId,
+                flowName));
+      }
+      return true;
+    } catch (RejectedExecutionException e) {
+      LOG.error("Failed to add more missed schedules tasks to the thread pool", e);
```
sshardool, line 113: Maybe rethrow the exception ?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,216 @@
...
+          + "Back execution will start soon as you enabled the config.");
+      this.emailMessage = messageFormat.format(
+          new Object[]{executeFlowAction.getFlowName(), executeFlowAction.getProjectName(), missScheduleTimestampToDate});
+    }
+
+    private String formatTimestamps(List<Long> missScheduleTimestamps) {
+      List<String> datesFromTimestamps = missScheduleTimestamps.stream()
+          .map(timestamps -> new Date(timestamps).toString())
+          .collect(Collectors.toList());
+      return String.join(",", datesFromTimestamps);
```
sshardool, line 192: What's the timezone used for the conversion?
Please follow the same time related conventions as for other email notifications.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: where is invoked
- it will be coming from TriggerManager. everytime calculateNextCheckTime updates and override previous value, the missed timestamps would be noticed and send as a task. This will be coming from second PR; 
The flow trigger side miss schedule calculation still need to be validated.

will the users get an email for every missed schedule?
- Yes if they configured failureEmails, this task will send email action as soon as it got executed.

ShellyWEI: I was planning to encapsulate all the failure handling logic inside this manager class, without leaking to other components. Since this service is robust, it also reduces the chances of introducing regression for existing system. The callers of this manager only know whether they successfully put the notification tasks in queue or not.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: Same as above.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: Done.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: back execution config will be added as Schedule Parameter. User would need to update existing schedule to apply this config. It will be turned off by default. TriggerManager would aware of this config and passed into this Manager.
This will also come from later PR (change schedule servlet to support adding this config)

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: apply the flowtrigger email date format. It will be on thread's timezone.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,255 @@
...
+  // parameters to adjust MissScheduler
+  private final long sleepIntervalInMs;
+  private final long DEFAULT_THREAD_IDLE_TIME_IN_MS = 60000;
+  private final int threadPoolSize;
+  private final int DEFAULT_THREAD_POOL_SIZE = 50;
+
+  @Inject
+  public MissedSchedulesManager(final Props azkProps,
+      final ProjectManager projectManager,
+      final Emailer emailer) {
```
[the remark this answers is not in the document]
ypadron-in: sounds good.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: If the Web Server is down for 18 mins and there is a flow scheduled to run every 5 mins so potentially there could be 3 schedules misses. When the Web Server comes back up and TriggerManager executes `calculateNextCheckTime`:
1-Will it detect the 3 missed schedules or just one for example? 
2-If the 3 misses are detected, is it possible to add one notification task for the combined misses instead of 3 separate notification tasks?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,199 @@
...
+  public boolean addMissedSchedule(final List<Long> missedScheduleTimesInMs,
+      final ExecuteFlowAction action,
+      final boolean backExecutionEnabled) {
+    String projectId = action.getProjectName();
+    String flowName = action.getFlowName();
+    Project project = projectManager.getProject(projectId);
+    if (project == null) {
+      return false;
+    }
+    List<String> emailRecipients = project.getFlow(flowName).getFailureEmails();
```
ypadron-in, line 93: With `getFlow().getFailureEmails()` we'll probably only get failure emails defined in the yaml file but we'll miss any overwrites or new definitions added when the schedule was created. We need to check all possible sources.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,199 @@
...
+    try {
+      Future taskFuture = this.missedSchedulesExecutor.submit(
```
ypadron-in, line 95: If `emailRecipients` is an empty list does it still make sense to submit a notification task?

ypadron-in: I guess if the user configured automatic recovery execution, but then the task to queue will be in charge of 2 things (notification + execution-triggering). Wondering if we should break it into 2 tasks. Thoughts?

ypadron-in, line 95: If the feature is disabled missedSchedulesExecutor is set to null then submit call will throw NPE right?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,199 @@
...
+      final String missScheduleTimestampToDate = formatTimestamps(missedScheduleTimes);
+      MessageFormat messageFormat = new MessageFormat("This is autogenerated email to notify to flow owners that"
+          + " the flow {0} in project {1} has scheduled the executions on {2} but failed to trigger on time. "
+          + (backExecutionEnabled ? "Back execution will start soon as you enabled the config." : ""));
+
+      this.emailMessage = messageFormat.format(
+          new Object[]{executeFlowAction.getFlowName(), executeFlowAction.getProjectName(), missScheduleTimestampToDate});
+    }
+
+    private String formatTimestamps(List<Long> missScheduleTimestamps) {
```
ypadron-in, line 170: Can we use a utility method in TimeUtils class for the formatting?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,199 @@
...
+      List<String> datesFromTimestamps = missScheduleTimestamps.stream()
+          .map(timestamps -> sdf.format(new Date(timestamps)))
+          .collect(Collectors.toList());
+      return String.join(",", datesFromTimestamps);
+    }
+
+    @Override
+    public void run() {
+      try {
+        this.emailer.sendEmail(this.emailRecipients, "Missed Azkaban Schedule Notification", this.emailMessage);
```
ypadron-in, line 181: If there is a failure while sending the email then the backfill/recovery execution will not be triggered. Are we ok with that?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,199 @@
...
+        if (this.executeFlowAction != null) {
+          this.executeFlowAction.doAction();
+        }
+      } catch (InterruptedException e) {
+        String warningMsg = "MissedScheduleTask thread is being interrupted, throwing out the exception";
+        LOG.warn(warningMsg, e);
+        throw new RuntimeException(warningMsg, e);
```
ypadron-in, line 188: Should we throw an AzkabanException here?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: 1. I haven't added the triggerManager logic yet but yes it will be capable to detect 3 times miss. But currently it is not. It will skip to 4th.
2. There is only one task for each trigger at a time. So even if there is 3 times, it will be only one. That's why the missedTimestamps is a list.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: What do you mean that? If they change the flow failure emails it can not be overwrites?

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: See the implementation of Emailer. If the emailRecipient List is empty there will be operations. And this sendEmail will not throw any exception if you saw the implementation, it handles and print the error message itself. So the back execution if configured will not be impacted.
But I think it would make sense to add metrics for each operations to track the failure count vs total task count.
I am planning to have separate PR to add all the necessary metrics later

ShellyWEI: Yes it will be NPE but the caller would know if they would send the task or not, because caller itself also would capture the backExecution config.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: Same as other commands, emailer should not affect back execution as it catches Exception there

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: Sure

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: Actually on the second thought, i think it makes sense to add this logic back inside missedScheduleManager itself. Even if the thread pool size has default value > 0 but let's be conscientious.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: Great, thanks!

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: Sorry, I should have described more. I meant checking TimeUtils methods like `formatDateTimeZone` or `formatDateTime` to see if we could use those instead of writing the formatting logic here. Didn't mean to create a utility method for the strings work.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: Failure emails can be defined in the .yaml/.properties files but also at the time of creating the schedule where runtime properties can be provided. I believe here we're only taking into consideration failure emails defined in the yaml but not possible overwrites or additions done through runtime properties and in fact those have higher precedence than the ones in the .yaml/.properties files.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: Never mind sorry, I just noticed below if in line 86 which will stop execution of `addMissedSchedule` and the NPE will not be thrown.

```
if (!this.missedSchedulesManagerEnabled) {
      LOG.warn("missed Schedule manager is not enabled, can not add tasks.");
      return false;
    }
```

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: sure I changed to leverage formatDateTime.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: That's actually good call!! thank you, I didn't aware when defining schedules we provide them a failureEmail override. Adjust it in the code.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
```diff
@@ -0,0 +1,213 @@
...
+    private ExecuteFlowAction executeFlowAction;
+
+    public MissedSchedulesOperationTask(List<Long> missedScheduleTimes,
+        Emailer emailer,
+        List<String> emailRecipients,
+        boolean backExecutionEnabled,
+        ExecuteFlowAction executeFlowAction) {
+      this.emailer = emailer;
+      this.emailRecipients = emailRecipients;
+      this.executeFlowAction = backExecutionEnabled ? executeFlowAction : null;
```
ypadron-in, line 169: I think we'll always need to provide an `executionFlowAction` object regardless of the value of `backExecutionEnabled`. This object cannot be null because we need it to get possible overwrites of failure emails in line 99 right?

ypadron-in (dismissed):

ypadron-in:
Thank you for addressing all the comments!

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ypadron-in: Talked offline, my bad, emails are resolved outside of the task object so the attribute executeFlowAction can be null.

azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
ShellyWEI: applied with formatDateTime instead. still the server local time instead of UTC.

byjiang1996 (approved):

PR #3207, opened 2022-11-23
Remove redundant project.setVersion line which is added by mistake

byjiang1996:
There is already a project.setVersion in the tail of this function.

A redundant project.setVersion prior to `uploading project property files` will introduce the bug described in this PR: https://github.com/azkaban/azkaban/pull/3149


tsangz2013 (approved):
lgtm, thanks for finding and fix this!

PR #3208, opened 2022-11-29
Use flowSubmitTime instead of flowStartTime for log timebox querying

byjiang1996:
In most cases, the timestamp for Azkaban flow logs and job logs is ahead of the flowStartTime (when flow status is set to RUNNING) b/c there are also other logs before we want to mark flow status to RUNNING.


utk-12 (approved):
LGTM! 
Thanks for identifying the gap and ensuring the missed timeframe of logs are present !

PR #3211, opened 2022-12-06
New feature flag to disable Yarn Application Kill

tsangz2013:
**Why it's needed**
We still observe that the Yarn KillApplication() call got stuck for ~ 15 minutes before it throws exception, so we need a way to mitigate this by disabling yarn kill feature, to avoid a failed flow stuck for too long and blocks following retry.

**What's done**
Introduce a new "disable" feature flag, when set, will skip doing any Yarn Application Kill;

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -88,10 +88,15 @@ public class HadoopJobUtils {
   // MapReduce config for mapreduce job tags
   public static final String MAPREDUCE_JOB_TAGS = "mapreduce.job.tags";
 
+  // feature version key, resides in jobProps, also can be config from properties file
   public static final String YARN_KILL_VERSION = "yarn.kill.version";
   // values of the yarn kill version flag
+  // get the application IDs by reading the job log (default value)
   public static final String YARN_KILL_LEGACY = "legacy";
+  // get the application IDs by calling the getApplications() API with tokens loaded in yarnClient
   public static final String YARN_KILL_USE_API_WITH_TOKEN = "api_with_token";
+  // disabling the whole yarn kill feature, skip killing the applications
+  public static final String YARN_KILL_DISABLED = "disabled";
 
   protected static final int APPLICATION_TAG_MAX_LENGTH = 100;
   // Root of folder in storage containing startup dependencies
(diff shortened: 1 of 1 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -90,6 +90,7 @@ public class HadoopJobUtils {
 
   public static final String YARN_KILL_VERSION = "yarn.kill.version";
   // values of the yarn kill version flag
+  public static final String YARN_KILL_DISABLED = "disabled";
```
sshardool, line 93: Where will this value be specified and would it apply to all the Azkaban jobs ?

sshardool (commented):
Thanks for making the change. 
Have a couple of clarifying questions below.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -389,6 +390,11 @@ public Void run() throws Exception {
   }
 
   private static void findAndKillYarnApps(Props jobProps, Logger log) {
+    String yarnKillVersion = jobProps.getString(YARN_KILL_VERSION, YARN_KILL_LEGACY).trim();
+    if (YARN_KILL_DISABLED.equals(yarnKillVersion)) {
```
NancyXie2022, line 394: Please add comment to the change to record the change.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: By the nature of flow container, rolling out a feature flag requires setting the flag in the `common.properties` in az-cluster-config, create and activate a new config image, so as it can be applied.

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -389,6 +390,11 @@ public Void run() throws Exception {
   }
 
   private static void findAndKillYarnApps(Props jobProps, Logger log) {
+    String yarnKillVersion = jobProps.getString(YARN_KILL_VERSION, YARN_KILL_LEGACY).trim();
```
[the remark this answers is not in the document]
tsangz2013: Got it.

NancyXie2022 (approved):

PR #3215, opened 2022-12-09
Use ExecutorService/Future to ensure Yarn KillApplication() call timeout

tsangz2013:
**Why we need this**
We still observe that the Yarn KillApplication() call got stuck for ~ 15 minutes before it throws exception, so we need a way to mitigate to avoid a failed job stuck for too long and thus breaking the SLA, on Azkaban system side.

**What's done**
Use Future to handle each Yarn KillApplication() call, and add a timeout to each Future usage, so as it will only wait for a short period of time.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
+++ b/azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
@@ -26,6 +26,11 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.security.UserGroupInformation;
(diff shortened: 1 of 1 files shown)


azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -107,17 +112,32 @@ public static List<ApplicationReport> getAllAliveAppReportsByExecIDs(final YarnC
...
-   * Uses YarnClient to kill the jobs one by one
+   * Uses YarnClient to kill the jobs one by one, each kill has a timeout
    */
   public static void killAllAppsOnCluster(YarnClient yarnClient, Set<String> applicationIDs,
       Logger log) {
     log.info(String.format("Killing applications: %s", applicationIDs));
+    ExecutorService executor = Executors.newSingleThreadExecutor();
 
     for (final String appId : applicationIDs) {
+      Future<?> future = executor.submit(new Runnable() {
```
NancyXie2022, line 123: Has this been tested?

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -107,17 +112,32 @@ public static List<ApplicationReport> getAllAliveAppReportsByExecIDs(final YarnC
 
 
   /**
-   * Uses YarnClient to kill the jobs one by one
+   * Uses YarnClient to kill the jobs one by one, each kill has a timeout
    */
   public static void killAllAppsOnCluster(YarnClient yarnClient, Set<String> applicationIDs,
       Logger log) {
     log.info(String.format("Killing applications: %s", applicationIDs));
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
svnarumugam, line 120: I am curious to know the reasoning behind single thread executor. would it be helpful if we use a multiple threads?

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
```diff
@@ -107,17 +112,32 @@ public static List<ApplicationReport> getAllAliveAppReportsByExecIDs(final YarnC
...
+            log.warn("something happened while trying to kill this job: " + appId, t);
+          }
+        }
+      });
+      // wait for the kill with a timeout
       try {
-        YarnUtils.killAppOnCluster(yarnClient, appId, log);
-      } catch (final Throwable t) {
-        log.warn("something happened while trying to kill this job: " + appId, t);
+        future.get(YARN_APP_TIMEOUT_IN_MILLIONSECONDS, TimeUnit.MILLISECONDS);
```
svnarumugam, line 135: Can we consider adding a unit test for the killAllAppscluster method?

svnarumugam (commented):
Thanks @tsangz2013 for putting this change together! Left a few comments. Changes look good.

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
tsangz2013: let me add some unit tests

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
tsangz2013: I couldn't find a way to make a unit test work: mock sleeping in unit test is hard, instead I tested in an executor pod, the killing works well and the future timed out after 5 seconds

azkaban-common/src/main/java/azkaban/utils/YarnUtils.java
NancyXie2022: Do you mind put the execution link here? and any log line saying about the timeout?

tsangz2013: I posted in our internal channel

NancyXie2022: 13-12-2022 11:46:54 PST pigWordCount1 WARN - Timed out killing the yarn app application_1670314587930_1089113, cancelling the runnable...

NancyXie2022 (approved):

PR #3217, opened 2022-12-12
Refactor Alerter plugins feature.

ypadron-in:
Changes include:
-allowing multiple alerter plugins to be configured (not just one)
-failing initialization only for specific faulty plugins (not for all)
-simplifying plugin configuration (ie: by eliminating redundant config property `alert.type`) and triggering logic. Now when there is an alert-worthy event, control is passed to each installed alerter for it to decide whether something should be done or not.     
-adding more logging to plugin initialization for better troubleshooting

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
@@ -16,119 +16,118 @@
 package azkaban.executor;
 
 import azkaban.alert.Alerter;
+import azkaban.spi.AzkabanException;
 import azkaban.utils.Emailer;
 import azkaban.utils.FileIOUtils;
 import azkaban.utils.PluginUtils;
 import azkaban.utils.Props;
 import azkaban.utils.PropsUtils;
-import javax.inject.Inject;
-import javax.inject.Singleton;
 import java.io.File;
 import java.lang.reflect.Constructor;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.apache.log4j.Logger;
-
+import java.util.function.BiConsumer;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Singleton
 public class AlerterHolder {
 
-  private static final Logger logger = Logger.getLogger(AlerterHolder.class);
-  private Map<String, Alerter> alerters;
+  private static final Logger logger = LoggerFactory.getLogger(AlerterHolder.class);
+  private final Map<String, Alerter> alerters;
 
   @Inject
   public AlerterHolder(final Props props, final Emailer mailAlerter) {
-    try {
-      this.alerters = loadAlerters(props, mailAlerter);
-    } catch (final Exception ex) {
-      logger.error(ex);
-      this.alerters = new HashMap<>();
-    }
-  }
-
-  private Map<String, Alerter> loadAlerters(final Props props, final Emailer mailAlerter) {
     final Map<String, Alerter> allAlerters = new HashMap<>();
-    // load built-in alerters
+    // Load built-in alerters.
     allAlerters.put("email", mailAlerter);
-    // load all plugin alerters
-    final String pluginDir = props.getString("alerter.plugin.dir", "plugins/alerter");
-    allAlerters.putAll(loadPluginAlerters(pluginDir));
-    return allAlerters;
+    // Load additional alerter plugins if any.
+    final String customAlertersBaseDir = props.getString("alerter.plugin.dir", "plugins/alerter");
+    allAlerters.putAll(loadCustomAlerters(customAlertersBaseDir));
+    this.alerters = allAlerters;
+    logger.info("Alerter plugins loaded: {}", this.alerters.keySet());
   }
 
-  private Map<String, Alerter> loadPluginAlerters(final String pluginPath) {
-    final File alerterPluginPath = new File(pluginPath);
-    if (!alerterPluginPath.exists()) {
+  private Map<String, Alerter> loadCustomAlerters(final String alerterPluginsRootPath) {
+    final File pluginsRootPath = new File(alerterPluginsRootPath);
+    if (!pluginsRootPath.exists()) {
       return Collections.<String, Alerter>emptyMap();
     }
 
     final Map<String, Alerter> installedAlerterPlugins = new HashMap<>();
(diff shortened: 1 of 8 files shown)


azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -16,119 +16,118 @@
...
+    this.alerters = allAlerters;
+    logger.info("Alerter plugins loaded: {}", this.alerters.keySet());
   }
 
-  private Map<String, Alerter> loadPluginAlerters(final String pluginPath) {
-    final File alerterPluginPath = new File(pluginPath);
-    if (!alerterPluginPath.exists()) {
+  private Map<String, Alerter> loadCustomAlerters(final String alerterPluginsRootPath) {
+    final File pluginsRootPath = new File(alerterPluginsRootPath);
+    if (!pluginsRootPath.exists()) {
```
tsangz2013, line 57: minor: maybe add log here

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -16,119 +16,118 @@
...
+    if (pluginProps == null) {
+      throw new AzkabanException("Plugin config properties could not be loaded.");
+    }
 
-      if (alerterClass == null) {
-        continue;
-      }
+    final String pluginName = pluginProps.getString("alerter.name", "").trim();
+    if (pluginName.isEmpty()) {
+      throw new AzkabanException("Alerter name is required.");
```
tsangz2013, line 85: anything else in this `pluginProps` can be logged for more info?

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -16,119 +16,118 @@
...
+  }
 
-      Class<?> alerterClass =
-          PluginUtils.getPluginClass(pluginClass, pluginDir, extLibClassPaths, parentLoader);
+  private void loadCustomAlerter(final Map<String, Alerter> installedAlerterPlugins,
+      final ClassLoader parentLoader, final File pluginDir) {
+    // Load plugin properties.
+    final Props pluginProps = PropsUtils.loadPluginProps(pluginDir);
+    if (pluginProps == null) {
+      throw new AzkabanException("Plugin config properties could not be loaded.");
```
tsangz2013, line 80: maybe also log `pluginDir`?

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -16,119 +16,118 @@
...
-      Constructor<?> constructor = null;
-      try {
-        constructor = alerterClass.getConstructor(Props.class);
-      } catch (final NoSuchMethodException e) {
-        logger.error("Constructor not found in " + pluginClass);
-        continue;
-      }
+    final String pluginClass = pluginProps.getString("alerter.class", "").trim();
+    if (pluginClass.isEmpty()) {
+      throw new AzkabanException("Alerter class is required.");
```
tsangz2013, line 93: also here

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
```diff
@@ -16,119 +16,118 @@
...
-        obj = constructor.newInstance(pluginProps);
-      } catch (final Exception e) {
-        logger.error(e);
-      }
+    final Class<?> alerterClass =
+        PluginUtils.getPluginClass(pluginClass, pluginDir, extLibClassPaths, parentLoader);
+    if (alerterClass == null) {
+      throw new AzkabanException(
+          String.format("Alerter class '%s' could not be loaded.", pluginClass));
+    }
```
tsangz2013, line 101: old logic will `continue` and here it's breaking, is this still good?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -204,112 +203,71 @@ public static void restartFlow(final ExecutableFlow flow, final Status originalS
...
    *
-   * @param flow the execution
+   * @param flow          the execution
    * @param alerterHolder the alerter holder
-   * @param extraReasons the extra reasons for alerting
+   * @param extraReasons  the extra reasons for alerting
    */
-  public static void alertUserOnFlowFinished(final ExecutableFlow flow, final AlerterHolder
-      alerterHolder, final String[] extraReasons) {
-    final ExecutionOptions options = flow.getExecutionOptions();
```
tsangz2013, line 228: So this diff will default to use all alerters, no longer make them optional?

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
ypadron-in: done in line 69 above.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
ypadron-in: other interesting properties are logged in the plugin itself, only logging here Azkaban related properties.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
ypadron-in: all the exceptions thrown inside `loadCustomAlerter()` are logged in line 69.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
ypadron-in: it's the same behavior now and then, current changes encapsulate initialization of a plugin into a method, if something goes wrong an exception is thrown. That exception is then captured and logged before continuing to initialize next configured plugin if any.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
ypadron-in: correct, control is passed to each and every configured plugin, it's for their logic to decide whether to do something or not. Knowledge of when to call a plugin isn't/shouldn't belong in Azkaban.

azkaban-common/src/main/java/azkaban/executor/AlerterHolder.java
ypadron-in: I think we can use log in line 52 to learn if custom alerters are configured or not. As they might not be configured too often adding a log here could be just noise most of the times.

tsangz2013 (approved):
lgtm, please make sure to deploy and test in staging/test cluster before merging

PR #3219, opened 2022-12-13
CleanupManager: check all recent termination status flow to clean yarn apps

tsangz2013:
**Why we need this**
This is because we observe that in addition to "EXECUTION_STOPPED", other termination status flows could also have dangling yarn applications, we need to take care of them all.

**What's done**
Extend the current functions from only check for recent "EXECUTION_STOPPED", to check for recent flows of status in (execution_stopped, killed, killing, failed, failed_finishing, failed_succeeded, canceled).

**Tests done**
Updated & added unit tests;
Deployed and verified in staging cluster, everything works as expected

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -43,6 +43,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -309,7 +310,8 @@ public class ContainerCleanupManager {
         recentlyTerminatedFlows.addAll(
             flows.stream().map(ExecutableFlow::getExecutionId).collect(Collectors.toSet()));
         logger.info("Got recently terminated flows executions of Status " + status + ": " +
-            flows.stream().map(Object::toString).collect(Collectors.joining(",")));
+            flows.stream().map(ExecutableFlow::getExecutionId).map(Objects::toString)
+                .collect(Collectors.joining(",")));
       } catch (final ExecutorManagerException e) {
         logger.error("Unable to obtain current flows executions of Status " + status, e);
       }
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -336,17 +336,26 @@ public class ContainerCleanupManager {
       logger.info("cleanUpDanglingYarnApplications start ");
 
       Set<Integer> recentlyTerminatedFlows = getRecentlyTerminatedFlows();
+      StringBuffer sb1 = new StringBuffer();
+      recentlyTerminatedFlows.forEach(f -> {
+        sb1.append(f);
+        sb1.append(",");
+      });
+      logger.info("Get recentlyTerminatedFlows: " + sb1);
+
       // get those flows terminated but containers are still alive (failed to properly killed)
-      Set<Integer> toBeCleanedContainers = getContainersOfTerminatedFlows();
-      logger.info("Get terminatedContainers: " +
-          toBeCleanedContainers.stream().map(Object::toString).collect(Collectors.joining(",")));
+      Set<Integer> toBeCleanedFlows = getContainersOfTerminatedFlows();
+      StringBuffer sb2 = new StringBuffer();
+      toBeCleanedFlows.forEach(f -> {
+        sb2.append(f);
+        sb2.append(",");
+      });
+      logger.info("Get containersOfTerminatedFlows: " + sb2);
 
       // combine both sets
-      toBeCleanedContainers.addAll(recentlyTerminatedFlows);
-      logger.info("The whole set of all executions to clean yarn apps: " +
-          toBeCleanedContainers.stream().map(Object::toString).collect(Collectors.joining(",")));
+      toBeCleanedFlows.addAll(recentlyTerminatedFlows);
 
-      if (toBeCleanedContainers.isEmpty()) {
+      if (toBeCleanedFlows.isEmpty()) {
         logger.info("No execution needs to kill yarn application, exit");
         return;
       }
@@ -354,7 +363,7 @@ public class ContainerCleanupManager {
       // For each of yarn clusters: find applications of the above executionIDs and kill them
       for (Entry<String, Cluster> entry : this.allClusters.entrySet()) {
         logger.info("clean up yarn applications in cluster:" + entry.getValue().getClusterId());
-        cleanUpYarnApplicationsInCluster(toBeCleanedContainers, entry.getValue());
+        cleanUpYarnApplicationsInCluster(toBeCleanedFlows, entry.getValue());
       }
     } catch (Throwable t) {
       logger.warn("Encounter unexpected throwable during cleanup dangling yarn app, "
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -300,7 +300,7 @@ public class ContainerCleanupManager {
    * killing, failed, failed_finishing, failed_succeeded, canceled)
    */
   @NotNull
-  Set<Integer> getRecentKilledFlows() {
+  Set<Integer> getRecentTerminationFlows() {
     Set<Integer> recentKilledFlows = new HashSet<>();
     this.recentTerminatedStatusMap.forEach((status, value) -> {
       try {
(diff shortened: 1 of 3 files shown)


NancyXie2022 (approved):

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -174,9 +174,21 @@ public ContainerCleanupManager(
             new Pair<>(Duration.ofMinutes(runningFlowValidity), START_TIME))
         .build();
 
-    this.executionStoppedMap = new Builder<Status, Pair<Duration, String>>()
+    this.recentTerminatedStatusMap = new Builder<Status, Pair<Duration, String>>()
         .put(Status.EXECUTION_STOPPED,
-            new Pair<>(Duration.ofMinutes(maxExecStoppedValidity), UPDATE_TIME))
+            new Pair<>(Duration.ofMinutes(recentTerminationValidity), UPDATE_TIME))
```
sshardool, line 179: What's the utility of the second field of the pair, all of them have the value as 'update_time' in the map ?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,22 +296,25 @@ public Set<Integer> getContainersOfTerminatedFlows() {
...
   /**
-   * Get the flows of EXECUTION_STOPPED status
+   * Get the flows recently (last 15 minutes) terminated: status in (execution_stopped, killed,
+   * killing, failed, failed_finishing, failed_succeeded, canceled)
    */
   @NotNull
-  Set<Integer> getExecutionStoppedFlows() {
-    Set<Integer> executionStoppedFlows = new HashSet<>();
-    if (this.executionStoppedMap.containsKey(Status.EXECUTION_STOPPED)) {
+  Set<Integer> getRecentTerminationFlows() {
```
sshardool, line 303: Suggest rename to `getRecentlyTerminatedFlows`

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -318,18 +333,15 @@ public void cleanUpDanglingYarnApplications() {
...
       // get those flows terminated but containers are still alive (failed to properly killed)
       Set<Integer> toBeCleanedContainers = getContainersOfTerminatedFlows();
       logger.info("Get terminatedContainers: " +
           toBeCleanedContainers.stream().map(Object::toString).collect(Collectors.joining(",")));
 
       // combine both sets
-      toBeCleanedContainers.addAll(executionStoppedFlows);
-      logger.info("Get the set of all executions: " +
+      toBeCleanedContainers.addAll(recentTerminationFlows);
+      logger.info("The whole set of all executions to clean yarn apps: " +
```
sshardool, line 344: What's the expected/observed number of flows with 'dangling' yarn applications and is this change likely to have a performance impact on the overall working of cleanup-manager (single thread monitoring for this and dispatching for cleanup) ?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,22 +296,25 @@ public Set<Integer> getContainersOfTerminatedFlows() {
...
+    Set<Integer> recentKilledFlows = new HashSet<>();
+    this.recentTerminatedStatusMap.forEach((status, value) -> {
       try {
         List<ExecutableFlow> flows = this.executorLoader.fetchFreshFlowsForStatus(
-            Status.EXECUTION_STOPPED, this.executionStoppedMap);
-        executionStoppedFlows.addAll(
+            status, this.recentTerminatedStatusMap);
+        recentKilledFlows.addAll(
             flows.stream().map(ExecutableFlow::getExecutionId).collect(Collectors.toSet()));
+        logger.error("Got recent termination flows executions of Status " + status + ": " +
```
sshardool, line 311: Should this be logged as an 'error' or an info/warn ?
What's the expected number of such flows (worst case observation) ?

sshardool (commented):
Thanks for making the changes, had some questions.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: This is to fit the method being called below, it takes in `status` and a map of `<status, <retention, timestamp_tupe>>`

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: sure

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: 1. Yes it should be `info`
2. Will investigate. In our test cluster I think it's <30 in total

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: Will investigate

tsangz2013: Investigated for this question and below are evidences:
1. The number of "recently terminated flows" doesn't increase dramatically - I checked the busy clusters' metric graph, and run mysql queries, each batch of "recently terminated flows" should be < 100, and averagely 1 running yarn app per flow;
2. Run the mysql queries used by cleanupManager directly and the performance is good enough, each finish with 1.5s;
3. In previous test, the GetApplications() api call can easily handle input of ~200 execution_IDs and no performance issue;
4. In previous test, killing ~100 yarn application in a batch works well with the current parallel killing method (using `FixedThreadPool`);

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,22 +297,26 @@ public Set<Integer> getContainersOfTerminatedFlows() {
...
       try {
         List<ExecutableFlow> flows = this.executorLoader.fetchFreshFlowsForStatus(
-            Status.EXECUTION_STOPPED, this.executionStoppedMap);
-        executionStoppedFlows.addAll(
+            status, this.recentTerminatedStatusMap);
+        recentlyTerminatedFlows.addAll(
             flows.stream().map(ExecutableFlow::getExecutionId).collect(Collectors.toSet()));
+        logger.info("Got recently terminated flows executions of Status " + status + ": " +
+            flows.stream().map(ExecutableFlow::getExecutionId).map(Objects::toString)
+                .collect(Collectors.joining(",")));
```
sshardool, line 314: Nit: Consider using a StringBuffer for this, as currently implemented it will result in a larger number of string objects created than needed.

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -318,18 +335,15 @@ public void cleanUpDanglingYarnApplications() {
...
+      Set<Integer> recentlyTerminatedFlows = getRecentlyTerminatedFlows();
       // get those flows terminated but containers are still alive (failed to properly killed)
       Set<Integer> toBeCleanedContainers = getContainersOfTerminatedFlows();
       logger.info("Get terminatedContainers: " +
           toBeCleanedContainers.stream().map(Object::toString).collect(Collectors.joining(",")));
 
       // combine both sets
-      toBeCleanedContainers.addAll(executionStoppedFlows);
-      logger.info("Get the set of all executions: " +
+      toBeCleanedContainers.addAll(recentlyTerminatedFlows);
```
sshardool, line 345: Wondering, why is this called `toBeCleanedContainers`, I don't think the current logic confines this to containers which are pending cleanup, is that correct ?

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
```diff
@@ -284,22 +297,26 @@ public Set<Integer> getContainersOfTerminatedFlows() {
...
-    Set<Integer> executionStoppedFlows = new HashSet<>();
-    if (this.executionStoppedMap.containsKey(Status.EXECUTION_STOPPED)) {
+  Set<Integer> getRecentlyTerminatedFlows() {
+    Set<Integer> recentlyTerminatedFlows = new HashSet<>();
+    this.recentTerminatedStatusMap.forEach((status, value) -> {
       try {
         List<ExecutableFlow> flows = this.executorLoader.fetchFreshFlowsForStatus(
-            Status.EXECUTION_STOPPED, this.executionStoppedMap);
-        executionStoppedFlows.addAll(
+            status, this.recentTerminatedStatusMap);
```
sshardool, line 309: How frequently is this check and cancellation executed and is it possible that some corner case executions are missed between 2 subsequent executions of this method ?

sshardool (commented):
Thanks for the explanations regarding the usage. A few additional questions:

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: good call out, this piece of code was copied from another method and reuse the naming, will rename to `toBeCleanedFlows`

azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
tsangz2013: Sure

sshardool (approved):
Thanks for explanations and addressing review comments.

PR #3220, opened 2022-12-15
Kill flow and its yarn Applications: correct the proxyUser token loading with DEBUG logs added

tsangz2013:
**What is this**
Fix the token related bug that the `yarnClient.killApplication()` jammed for 15 mins and error out with credential exception.

**Tests done**
Tested in multiple flows in the container, the yarn kill application function works well, directly return result and no more blocking.


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
```diff
@@ -76,7 +76,12 @@ private static UserGroupInformation createSecurityEnabledProxyUser(
       log.info(String.format("Token = %s, %s, %s ", token.getKind(), token.getService(),
           Arrays.toString(token.getIdentifier())));
     }
-    proxyUser.addCredentials(loginUser.getCredentials());
```
tsangz2013, line 79: Suspecting these 2 lines causing the issue, but no idea why exactly

PR #3221, opened 2022-12-16
Kill flow and its yarn Applications: correct the proxyUser token loading with DEBUG logs added

tsangz2013:
**What is this**
Fix the token related bug that the `yarnClient.killApplication()` jammed for 15 mins and error out with credential exception.

**Tests done**
Tested in multiple flows in the container, the yarn kill application function works well, directly return result and no more blocking; Also reproduce the scenario that the SLA issue kill action on the flow and the yarn applications all killed at once, no blocking.

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
@@ -34,6 +34,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.security.PrivilegedExceptionAction;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -44,6 +45,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.security.token.Token;
 import org.apache.hadoop.yarn.api.records.ApplicationId;
 import org.apache.hadoop.yarn.client.api.YarnClient;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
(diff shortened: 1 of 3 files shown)


az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
```diff
@@ -76,7 +76,12 @@ private static UserGroupInformation createSecurityEnabledProxyUser(
       log.info(String.format("Token = %s, %s, %s ", token.getKind(), token.getService(),
           Arrays.toString(token.getIdentifier())));
     }
-    proxyUser.addCredentials(loginUser.getCredentials());
```
tsangz2013, line 79: Suspecting this is the cause of the bug

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -376,9 +378,57 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
         final UserGroupInformation proxyUser =
             HadoopSecureWrapperUtils.setupProxyUserWithHSM(hadoopSecurityManager, properties,
                 tokenFile.getAbsolutePath(), log);
+
+        // todo: print tokens in proxyUser
+        log.debug("proxyUserKillAllSpawnedHadoopJobs.proxyUser = " + proxyUser);
+        for (Token<?> token : proxyUser.getCredentials().getAllTokens()) {
```
sshardool, line 384: Is proxyUser.getCredentials() alway non-null, or is there a chance of NPE here ?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -376,9 +378,57 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
...
+          log.debug(String.format("proxyUserKillAllSpawnedHadoopJobs.proxyUser.Token = %s, %s",
```
sshardool, line 385: Similar to above, is `token` always non-null ?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -376,9 +378,57 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
...
+
+            UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+            log.debug("doAs.run.getCurrentUser = " + ugi);
+            for (Token<?> token : ugi.getCredentials().getAllTokens()) {
+              log.debug(String.format("doAs.run.getCurrentUser.Token = %s, %s", token.getKind(),
+                  token.getService()));
+            }
+            log.debug("doAs.run.getCurrentUser.Token --- end");
+
+            ugi = UserGroupInformation.getLoginUser();
```
sshardool, line 424: Is this the line fixing the issue ?
Could you please summarize what the root-cause was and how it's being addressed with this change?

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
```diff
@@ -376,9 +378,57 @@ public static void proxyUserKillAllSpawnedHadoopJobs(
         final UserGroupInformation proxyUser =
             HadoopSecureWrapperUtils.setupProxyUserWithHSM(hadoopSecurityManager, properties,
                 tokenFile.getAbsolutePath(), log);
+
+        // todo: print tokens in proxyUser
```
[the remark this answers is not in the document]
tsangz2013: will remove

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: it's always non-null as we observed

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: it's always non-null as we have observed for a long time

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopJobUtils.java
tsangz2013: no, it's not, let me add comments and ping you there

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/HadoopSecureWrapperUtils.java
sshardool: Could you please elaborate on this a bit, how was this leading to the bug ?

tsangz2013: Suspecting `.addCredentials()` method overwrite the tokens loaded into the UGI in the above for loop, thus causing RM delegation tokens missing

raymondlam12: Have you verified this? It seems like it is just adding the same credentials albeit I haven't deep-dived into the code implementation.

One way of verifying this is by printing the token state before & after removing this line in the case where cancelling fails.

djaiswal83: @tsangz2013  please log the token content to verify this.

tsangz2013:
Not the root cause of this problem

PR #3222, opened 2022-12-19
add latency and trigger count metrics for scheduling system

ShellyWEI:
Latency Metrics would show us how busy "Schedule System" is;
Count Gauge would show us the numbers of triggers being looping over and the trends. 
No Alert necessary to attach.


tsangz2013 (approved):
Very straightforward and clean, lgtm

PR #3224, opened 2023-01-03
Adding upload user information in project object metadata during Pod dispatch phase

utk-12:
Modifying the DB query while reading Project Object metadata to ensure the project upload user information is included within it. 
Currently this is not present in the Project table and one extra JOIN us required. Since this read only once during the active project load phase it should add constant memory overhead.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -1005,6 +1005,7 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
     }
 
     final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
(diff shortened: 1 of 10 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -1005,6 +1005,7 @@ private void ajaxExecuteFlow(final HttpServletRequest req,
     }
 
     final ExecutableFlow exflow = FlowUtils.createExecutableFlow(project, flow);
+    exflow.setUploadUser(project.getUploadUser());
```
byjiang1996, line 1008: Could you please also add this line in ReportalServlet.java?
(Before .setExecutionSource() line)

utk-12: fixed. good catch !

byjiang1996 (approved):

PR #3225, opened 2023-01-06
Fix formatting errors in some tables

vhermecz:
I've noticed that some tables (`Multiple Executor Mode Parameters` and `Executor Server Properties`) are missing from the config page.
Performing some tests with https://rubygems.org/gems/github-markup it turns out, that even a single formatting error within the table causes the whole table to be skipped by the RST2html converter.


amielkelinkedin (approved):
LGTM

PR #3226, opened 2023-01-06
Made the flow file size configurable

svnarumugam:
**Context**: Currently, we limit the size of the flow file to 10 MB, and changing this is possible only through a code change. This PR is to make this property configurable so that we can change the limit in certain clusters. This is needed by ML customers whose flows can be quite huge. Only if we supply the config value, we use that limit. Otherwise, we default to the current limit. ie. No behaviour change if this config is not supplied.

**Testing**: Tested by setting the size to 15MB. I was able to upload the flows of size ~15MB. 

**Additional notes:**
We store the flow object in two places `project_flows`, and `project_flow_files` table as JSON object and as flow file respectively. These two tables have their column type as medium blob and long blob. Because medium blob can store only up to 16MB, we would have to change it to a `long blob` if we decide to set it beyond 16MB.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
+++ b/azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
@@ -78,10 +78,13 @@ public class JdbcProjectImpl implements ProjectLoader {
 
   private static final int CHUCK_SIZE = 1024 * 1024 * 10;
   // Flow yaml files are usually small, set size limitation to 10 MB should be sufficient for now.
-  private static final int MAX_FLOW_FILE_SIZE_IN_BYTES = 1024 * 1024 * 10;
+  private static final int MAX_FLOW_FILE_SIZE_IN_MB_DEFAULT = 10;
+
+  public static final String MAX_FLOW_FILE_SIZE_KEY = "max.flow.file.size.mb";
   private final DatabaseOperator dbOperator;
   private final File tempDir;
   private final EncodingType defaultEncodingType = EncodingType.GZIP;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -95,6 +98,9 @@ public JdbcProjectImpl(final Props props, final DatabaseOperator databaseOperato
         logger.info("project temporary folder already existed.");
       }
     }
+    int maxFlowFileSizeInMB = props.getInt("max.flow.file.size.mb", MAX_FLOW_FILE_SIZE_IN_MB_DEFAULT);
+    maxFlowFileSizeInBytes = maxFlowFileSizeInMB * 1024 * 1024;
```
tsangz2013, line 102: I think you might use the `MAX_FLOW_FILE_SIZE_KEY` variable.

Also, this value will be used by web-server and be honored as long as it's running?

svnarumugam: Thank you for spotting that @tsangz2013. Fixed and updated the PR. Yes, I did want to use the const over there. This property is like most of the properties we have. Loaded when web server starts, and the value is used throughout the lifetime of the process. If we want to change, a web server restart is required.

tsangz2013 (dismissed):
thanks for raising this!

tsangz2013 (approved):
lgtm as only a dummy force push changing nothing.
please avoid force push if possible in the future

PR #3227, opened 2023-01-09
fix missed scheudle tests

ShellyWEI:
Previously 5 minutes can be large to let the actual miss falling into 5 minutes window and results in one more valid miss count with intentional miss in tests. Reducing time window as the schedule system does not support second level granularity.


utk-12:
LGTM. Thank you !

utk-12 (approved):

PR #3228, opened 2023-01-09
CleanupManager: add metrics of heartbeat and timer for all cleanup schedule tasks

tsangz2013:
**Why we need this**
So as we can know whether the cleanup manager is alive and functioning well, and get alerted if not, to prevent outages.

**Tests done**
Launched in a test web-server, able to see all the metrics in the graph dashboard.


byjiang1996 (approved):
LGTM

PR #3231, opened 2023-01-18
remove schedule global lock

ShellyWEI:
Previously Schedule system's lock is pretty strict.
- synchronized keywords ensures only one user would getSchedule, and even if viewing a single schedule would require updates the whole schedules.
- syncObj make sure only one trigger can be updated at a time, and it's conflicting with trigger scanning process. 

These locks makes schedule system unable to handle large load as only one of hundreds would acquire the lock and do the operation while all other requests would need to wait for the lock to release, even if those requests are totally irrelevant with each other. (One should be able to update its schedule, while less rely on other schedule updates) 

This PR removes all sync keywords and the global lock to lower the lock level to individual schedule/triggers instead of keeping all of them in the global synchronized stage. ScheduleManager is the entry point that links schedule and trigger, it will maintain its own checkTime for each individual trigger with its lastModifyTime as evidence to check if a schedule/trigger is getting updated.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
@@ -20,6 +20,7 @@ import azkaban.utils.Pair;
 import azkaban.utils.TimeUtils;
 import azkaban.utils.Utils;
 import java.util.Date;
+import java.util.concurrent.locks.ReentrantLock;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.ReadablePeriod;
@@ -44,6 +45,8 @@ public class Schedule {
   private int scheduleId;
   private long nextExecTime;
   private ExecutionOptions executionOptions;
+
+  private ReentrantLock lock = new ReentrantLock();
 
   public Schedule(final int scheduleId,
       final int projectId,
@@ -242,5 +245,13 @@ public class Schedule {
 
   public long getEndSchedTime() {
     return this.endSchedTime;
+  }
+
+  public void lock() {
+    this.lock.lock();
+  }
+
+  public void unlock() {
+    this.lock.unlock();
   }
 }
--- a/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
@@ -21,11 +21,13 @@ import azkaban.trigger.TriggerAgent;
 import azkaban.trigger.TriggerStatus;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import javax.inject.Inject;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import javax.validation.constraints.NotNull;
 import org.apache.log4j.Logger;
 import org.joda.time.DateTimeZone;
 import org.joda.time.ReadablePeriod;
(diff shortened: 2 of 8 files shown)


azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
```diff
@@ -238,4 +246,30 @@ public void loadTriggerFromProps(final Props props) throws ScheduleManagerExcept
...
+        if (s.getStatus().equals(TriggerStatus.EXPIRED.toString())) {
+          onScheduleExpire(s);
+        } else {
+          internalSchedule(s);
+        }
+        return updatedSchedule.get();
+      }
+      return s;
+    } finally {
+      s.unlock();
```
tsangz2013, line 272: Not super familiar with Java `finally`, I think move `return s;` to the end of method is more easy to read

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -331,19 +322,20 @@ public void run() {
                   "Trigger manager thread " + this.getName() + " is too busy! Remaining idle time in ms: "
                       + TriggerManager.this.runnerThreadIdleTime);
             } else {
-              TriggerManager.this.syncObj.wait(TriggerManager.this.runnerThreadIdleTime);
+              Thread.sleep(TriggerManager.this.runnerThreadIdleTime);
```
tsangz2013, line 325: Can you explain a little about the old dynamic of `this.syncObj` and why it's safe to directly sleep the current Thread instead?

azkaban-common/src/main/java/azkaban/scheduler/Schedule.java
```diff
@@ -243,4 +246,12 @@ public boolean skipPastOccurrences() {
   public long getEndSchedTime() {
     return this.endSchedTime;
   }
+
+  public void lock() {
+    this.lock.lock();
+  }
```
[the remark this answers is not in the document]
ShellyWEI: check the ReentrantLock's java doc. This method does not throw any exceptions. concurrent package usually operated atomically on JVM directly.

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
ShellyWEI: If anything happens before "returnn s", for example an exception happens. this method might exit but this lock would never release. So in java, it's recommended to use "finally" keywords to release lock.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
ShellyWEI: Again it's a good habit to check java doc directly. An obj's wait method is a direct waiting to be invoked operation, anything interruption would awake this obj to continue with the code. -> of course in the original implementation, there are keywords "synchrinozed", which means others need to wait when someone ontain this obj's lock. But I remove them in this PR, when this syncObj only remains the "sleep" functionality.

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
tsangz2013: in current setting, if no exception, it will unlock and then return? or it will return but also unlock?

tsangz2013 (approved):
lgtm, please test carefully before merging

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
ShellyWEI: yes, using "finally" means after the "try" it will whatever unblock. If it successfully executes return it will return.

PR #3232, opened 2023-01-24
Support custom alerters when defining SLAs with alert action.

ypadron-in:
Before, SLA rules with action set to `ALERT` would only support email notifications. 
With this change SLA misses can also be notified via custom notification systems installed as plugins. When defining SLAs through the `setSla` or `executFlow` apis, users can now set configs for custom alerters. For example:

```
curl --location --request POST 'http://localhost:8081/schedule' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'ajax=setSla' \
--data-urlencode 'scheduleId=12345’ \
--data-urlencode 'settings[0]=,FINISH,00:01,true,false' \

--- new params -------------
--data-urlencode 'slaAlerters[myCustomAlerter][prop1]=value1' \ 
--data-urlencode 'slaAlerters[email][recipients]=abc@abc.com'
```

TODO: include the new custom alerter configs in the responses of APIs that retrieve SLA info.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/alert/Alerter.java
+++ b/azkaban-common/src/main/java/azkaban/alert/Alerter.java
@@ -22,6 +22,7 @@ import azkaban.executor.ExecutorManagerException;
 import azkaban.flow.Flow;
 import azkaban.project.Project;
 import azkaban.sla.SlaOption;
+import azkaban.utils.Emailer;
 import java.util.List;
 import java.util.Map;
 
@@ -33,7 +34,13 @@ public interface Alerter {
 
   void alertOnFirstError(ExecutableFlow exflow) throws Exception;
 
-  void alertOnSla(SlaOption slaOption, String slaMessage) throws Exception;
+  @Deprecated
+  default void alertOnSla(SlaOption slaOption, String slaMessage) throws Exception { }
+
+  default void alertOnSla(ExecutableFlow exflow, SlaOption slaOption) throws Exception {
+    final String slaMessage = Emailer.createSlaMessage(exflow, slaOption, getAzkabanURL());
+    alertOnSla(slaOption, slaMessage);
+  }
 
   void alertOnFailedUpdate(Executor executor, List<ExecutableFlow> executions,
       ExecutorManagerException e);
@@ -44,5 +51,8 @@ public interface Alerter {
 
   void alertOnJobPropertyOverridden(Project project, Flow flow, Map<String, Object> metaData);
 
-  String getAzkabanURL();
+  @Deprecated
+  default String getAzkabanURL() {
+    return "";
+  };
 }
--- a/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
@@ -44,11 +44,9 @@ import org.apache.commons.lang.StringUtils;
 
 public class HttpRequestUtils {
 
-  public static final String PARAM_SLA_EMAILS = "slaEmails";
   public static final String PARAM_SLA_SETTINGS = "slaSettings";
 
-  public static ExecutionOptions parseFlowOptions(final HttpServletRequest req,
-      final String flowName)
+  public static ExecutionOptions parseFlowOptions(final HttpServletRequest req, final String flowName)
       throws ServletException {
     final ExecutionOptions execOptions = new ExecutionOptions();
 
@@ -116,11 +114,7 @@ public class HttpRequestUtils {
       execOptions.setMailCreator(mailCreator);
     }
 
-    final Map<String, String> slaSettings = getParamGroup(req, PARAM_SLA_SETTINGS);
-    // emails param is optional
-    final String emailStr = getParam(req, PARAM_SLA_EMAILS, null);
-    final List<SlaOption> slaOptions = SlaRequestUtils.parseSlaOptions(flowName, emailStr,
-        slaSettings);
+    final List<SlaOption> slaOptions = SlaRequestUtils.parseSlaOptions(req, flowName, PARAM_SLA_SETTINGS);
     if (!slaOptions.isEmpty()) {
       execOptions.setSlaOptions(slaOptions);
     }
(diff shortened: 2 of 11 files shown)


azkaban-common/src/main/java/azkaban/alert/Alerter.java
```diff
@@ -33,7 +34,13 @@ public interface Alerter {
 
   void alertOnFirstError(ExecutableFlow exflow) throws Exception;
 
-  void alertOnSla(SlaOption slaOption, String slaMessage) throws Exception;
+  @Deprecated
+  default void alertOnSla(SlaOption slaOption, String slaMessage) throws Exception { }
```
amielkelinkedin, line 38: I know this is in your interface and all, but is there anyway we can delete this? Is it used elsewhere?

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -26,28 +30,52 @@
...
+  public static final String PARAM_SLA_ALERTERS = "slaAlerters";
   public static final String SLA_STATUS_SUCCESS = "SUCCESS";
   private static final Logger logger = Logger.getLogger(SlaRequestUtils.class);
 
-  public static List<SlaOption> parseSlaOptions(final String flowName, final String emailStr,
-      final Map<String, String> settings) throws ServletException {
-    final List<String> slaEmails;
-    if (emailStr == null) {
-      slaEmails = Arrays.asList();
+  public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, final String flowName,
```
svnarumugam, line 44: [non-blocker, improvement] I am curious to know the reason behind sending the servlet request object here. Does it make testing this SlaRequestUtils harder? The parseSlaOptions is doing quite a lot of work. Should we consider adding the reasoning or adding a todo for testing this?

svnarumugam (approved):
Overall, the change looks neat. Added a clarifying question. Thanks a lot for working on this @ypadron-in . Are we making this available only through the API to start with? do we have plans to add it to the UI?

azkaban-common/src/main/java/azkaban/alert/Alerter.java
ypadron-in: After some time we can definitely delete these deprecated APIs to keep the Alerter interface clean. For now we're deprecating and defaulting to old behavior to avoid breaking any existing custom alerters which have to implement the Alerter interface.

ypadron-in:
> Overall, the change looks neat. Added a clarifying question. Thanks a lot for working on this @ypadron-in . Are we making this available only through the API to start with? do we have plans to add it to the UI?

Yes, users will be able to set properties for custom alerters through the "create SLA" UI page as well. I'll follow up with those changes shortly.

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: The parsing of SLA options is done in 2 different places, when SLAs are created and new executions are submitted. We moved common duplicated logic like capturing SLA email param into the `parseSlaOptions`. Based on that we might need to adjust the method name accordingly, you're right there.

I think tests in `HttpRequestUtilsTest` and `ExecutorServletTest` classes check this parsing logic indirectly but please let me know if something is missing and I'll add it in a follow up PR.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -366,21 +364,25 @@ public static Map<String, String> getParamGroup(final HttpServletRequest request
...
     final Enumeration<String> enumerate = request.getParameterNames();
     final String matchString = groupName + "[";
 
     final Map<String, Map<String, String>> groupParam = new HashMap<>();
     while (enumerate.hasMoreElements()) {
       final String str = enumerate.nextElement();
       if (str.startsWith(matchString)) {
-        final int level1KeyEnd = str.indexOf("]");
+        try {
+          final int level1KeyEnd = str.indexOf("]");
```
amielkelinkedin, line 376: Can this be outside the try?

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -26,28 +30,52 @@
...
+      final String settingsParamName) throws ServletException {
+    final Map<String, String> settings = getParamGroup(req, settingsParamName);
+    final Map<String, Map<String, String>> alertersConfigs = getMapParamGroup(req, PARAM_SLA_ALERTERS);
+    String slaEmailsStr = getParam(req, PARAM_SLA_EMAILS, null);
+
+    // Don't allow combining old & new in the same request:
+    // This ensures that there's no need to handle possible conflicts between 'slaEmails' and
+    // 'slaAlerters[email][recipients]'.
+    if (slaEmailsStr != null && !alertersConfigs.isEmpty()) {
```
amielkelinkedin, line 53: Should slaEmailsStr be notEmpty as well?

azkaban-common/src/main/java/azkaban/sla/SlaOption.java
```diff
@@ -30,9 +30,6 @@
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.joda.time.DateTime;
```
amielkelinkedin, line 33: Thanks so much for getting rid of joda! I assume it is used elsewhere as well, but might be good to double check if the import could be removed entirely!

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
svnarumugam: Thanks Yeni for the clarification!

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
ypadron-in: If you mean the `level1KeyEnd` assignment no because then it will be out of scope inside the try block where it's used.

amielkelinkedin: Ah, missed that, thanks!

amielkelinkedin (approved):
Thank you so much, @ypadron-in

PR #3233, opened 2023-02-04
job status update on execution_stop

ShellyWEI:
Update job status in DB when calling finalizing flows. Table execution_jobs will be updated with corresponding status in job history page.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -132,7 +132,7 @@ public class ExecutionControllerUtils {
         // If it's marked finished, we're good. If not, we fail everything and then mark it
         // finished.
         if (!isFinished(dsFlow)) {
-          failEverything(eventHandler, projectManager, dsFlow, finalFlowStatus);
+          failEverything(eventHandler, projectManager, dsFlow, finalFlowStatus, executorLoader);
           executorLoader.updateExecutableFlow(dsFlow);
         }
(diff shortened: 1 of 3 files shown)


tsangz2013 (approved):
LGTM.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -373,6 +373,11 @@ public static void failEverything(final EventHandler eventHandler,
         if (node.getEndTime() == -1) {
           node.setEndTime(time);
         }
+        try {
+          executorLoader.updateExecutableNode(node);
+        } catch (ExecutorManagerException e) {
+          logger.error("fail to update job status in DB, ", e);
+        }
```
ypadron-in, line 380: Can we please add the node embedded id or name to the log message?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -324,7 +324,7 @@ public static void failEverything(final ExecutableFlow exFlow,
    */
   public static void failEverything(final EventHandler eventHandler,
       final ProjectManager projectManager, final ExecutableFlow exFlow,
-      final Status finalFlowStatus) {
+      final Status finalFlowStatus, final ExecutorLoader executorLoader) {
```
ypadron-in, line 327: If we are passing the `executorLoader` as a parameter perhaps we could also include the executable flow update call (`updateExecutableFlow(executableFlow)`) inside this method. What do you think about that refactor?

PR #3218, opened 2022-12-13
add backExecuteOnce checkbox in UI

ShellyWEI:
To support backExecute Option for user, adding a checkbox in UI page with default by false; 
(Some code refactoring by azkaban Style of saveActionsPlugin)
New UI page please refer to design doc: https://docs.google.com/document/d/1rpgWZfN6aogZRGvPFqlC1QXJyyS9VJnUQX2uLPXC-Lw

PR #3234, opened 2023-02-08
add back execute support for miss schedule

ShellyWEI:
Add a back execute flag in the missedScheduleManager as to execute the flow immediately if such flag is marked enabled. Once enabled, every miss detection will trigger an execution of flow, even if there are multiple missed schedule in a single detection. The detection reuses the same logic of previous PR #3209 of email notification.
This PR implements the backend logic of previous PR of #3218 . User can enable this feature by updating their schedule with "Back Execute Once When missed Schedule Detects." on azkaban UI.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/MissedSchedulesManager.java
@@ -61,9 +61,9 @@ public class MissedSchedulesManager {
   // parameters to adjust MissScheduler
   private final int threadPoolSize;
   private final int DEFAULT_THREAD_POOL_SIZE = 5;
-  private final MetricsManager metricsManager;
   private final Counter emailCounter;
   private final Counter missScheduleCounter;
+  private final Counter backExecutionCounter;
 
   @Inject
   public MissedSchedulesManager(final Props azkProps,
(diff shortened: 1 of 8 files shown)


tsangz2013 (approved):
LGTM!
If I am right this PR is adding this flag and the usage of it will be in later PRs, please link this PR number to following PRs.

ypadron-in:
How does this feature work? Will a new execution be triggered for each schedule miss?
Could you please add the explanation to the PR description?

ShellyWEI:
> LGTM! If I am right this PR is adding this flag and the usage of it will be in later PRs, please link this PR number to following PRs.

No this is the final PR:) 
The trigger to execute flow if already implemented in MissedScheduleManager, I left the parameter "backExecutionEnabled" in there before, this PR gives that parameter the real value.

ShellyWEI:
> 

Actually no. Multiple misses will only trigger a single execution by the time web server realize there is a miss. So this feature is not intended for those frequent flows (e.g. scheduled every minute, usually those flows will have the ability to recover on its own, so this is useless feature for them), but mostly for those hourly/daily flows that a miss is really missing one execution for that single slot data.

ShellyWEI:
> > LGTM! If I am right this PR is adding this flag and the usage of it will be in later PRs, please link this PR number to following PRs.
> 
> No this is the final PR:) The trigger to execute flow if already implemented in MissedScheduleManager, I left the parameter "backExecutionEnabled" in there before, this PR gives that parameter the real value.

tsangz2013:
> > LGTM! If I am right this PR is adding this flag and the usage of it will be in later PRs, please link this PR number to following PRs.
> 
> No this is the final PR:) The trigger to execute flow if already implemented in MissedScheduleManager, I left the parameter "backExecutionEnabled" in there before, this PR gives that parameter the real value.

Thanks for explaining! This is great! Good job!

PR #3235, opened 2023-02-09
Introduce VPAFlowCriteria to deny VPA feature based on project or flo…

byjiang1996:
…w name filter

Project and flow name filter file looks like this:

>  proj1:flow1
proj1:flow2
proj2
proj3:flow1

which means VPA feature will not apply to proj1's flow1 flow, proj1's flow2 flow, all proj2's flows and proj3's flow1 flow.

It serves as the safeguard if there is any flow in production does not fit well with VPA's autoscaling algorithm.

One can use azkaban web api to reload flow filter text at runtime

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
@@ -1,0 +1,127 @@
+package azkaban.executor.container;
+
+import azkaban.Constants;
+import azkaban.executor.ExecutableFlow;
+import azkaban.utils.Props;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+
+
+/**
+ * Class for denying VPA feature based on project or flow name filter.
+ * A sample file can look like this:
+ * proj1:flow1
+ * proj1:flow2
+ * proj2
+ * proj3:flow1
+ */
+public class VPAFlowCriteria {
+  // Flows are stored in a map where key is project name and value is set of flows.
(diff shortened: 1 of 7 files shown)


azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
```diff
@@ -0,0 +1,127 @@
+package azkaban.executor.container;
+
```
svnarumugam, line 2: Clarification: if we update the (project, flow) pair, we update the config and then restart the web server to be able to pick that. Is it correct to say? 

This change has a lot in common with a previous change https://github.com/azkaban/azkaban/pull/3152. Is there a possibility to re-use some parts of it for this use case?

svnarumugam (commented):
Thanks for the super fast turnaround on this change @byjiang1996 . Added a clarifying question to better appreciate this change. The changes look good to me.

azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
byjiang1996: Good point. Addressed.

svnarumugam (approved):
If we are in a hurry we can go ahead and push this change. we should consider having this flow filter config in a standard format like `json`, or `hocon`. They are easy to parse and use with less code. what are your thoughts @byjiang1996?

byjiang1996:
> If we are in a hurry we can go ahead and push this change. we should consider having this flow filter config in a standard format like `json`, or `hocon`. They are easy to parse and use with less code. what are your thoughts @byjiang1996?

I don't think we should do it in a json way. This config is not a k-v pair but simply a list/array of flow/project names. JSON is an overkill for it from my perspective. In addition, I feel like this feature can be deprecated in the future once VPA is more stable

svnarumugam:
> > If we are in a hurry we can go ahead and push this change. we should consider having this flow filter config in a standard format like `json`, or `hocon`. They are easy to parse and use with less code. what are your thoughts @byjiang1996?
> 
> I don't think we should do it in a json way. This config is not a k-v pair but simply a list/array of flow/project names. JSON is an overkill for it from my perspective. In addition, I feel like this feature can be deprecated in the future once VPA is more stable

I certainly see the point about this being a short-term use case - we are good to ship this. However, validating, and reading a well-defined format can be done with less code. Right now the code to read from a file, and validate the input data can be done rather easily.

PR #3216, opened 2022-12-12
Adding Principle of Least Priviledge Security model for Keystore Mana…

utk-12:
…ger and for in-memory Security Certificates

PR #3236, opened 2023-02-14
Optimizing how proxy users are determinded prior to pod dispatch for …

utk-12:
…Security changes related to POLP commit https://github.com/azkaban/azkaban/pull/3216 and modified tests.

This commit has the following changes: 
1. Determination of Proxy Users needed for the flow is optimized by taking it in the order of precedence ( Flow Param --> Job Override Property --> Node Property ) 
2. Add this HashMap of proxy users to the security init container.
3. Added a test testPopulatingProxyUsersFromProject() to test this logic. 
4. Added a list of jobtype specific users that might be needed getJobTypeUsersForFlow() and mentioned the format for the configuration.
5. Added a test for getJobTypeUsersForFlow to ensure we get the correct proxy users.
6. Increased Junit heap memory for faster azkaban-common module builds 

Screenshot of local build :
Screenshot 2023-03-04 at 2 08 04 PM


azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +90,57 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+          proxyUsers.add(userToProxyFromJobNode);
+        } else if (userToProxyFromNode != null) {
+          proxyUsers.add(userToProxyFromNode);
+        }
+      }
+    }
+  }
+
+  // Extract the proxy users needed from  PREFETCH_JOBTYPE_PROXY_USER_MAP
+  public static HashSet<String> getJobTypeUsersForFlow(String jobTypePrefetchUserMap,
```
svnarumugam, line 132: It was challenging to read through this code. As discussed, could we consider adding a unit test for this method? after seeing the format `"jobtype1,jobtype1_proxyuser;jobtype2,"
      + "jobtype2_proxyuser";` I was able to better understand this change.

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +90,57 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+    populateProxyUsersForFlow(flow, flow, projectManager, proxyUsers);
+    return proxyUsers;
+  }
+
+  public static void populateProxyUsersForFlow(final ExecutableFlow flow,
+      final ExecutableNode node, final ProjectManager projectManager, HashSet<String> proxyUsers) {
+    if (node instanceof ExecutableFlowBase) {
+      final ExecutableFlowBase base = (ExecutableFlowBase) node;
+      for (ExecutableNode subNode : base.getExecutableNodes()) {
+        populateProxyUsersForFlow(flow, subNode, projectManager, proxyUsers);
```
svnarumugam, line 106: I am trying to understand if we have to walk through the nodes recursively. Though we may have nested flows, Azkaban flattens it at some point and keeps it as a single DAG. @ypadron-in probably knows the context better. could you please comment?

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
```diff
@@ -509,6 +523,66 @@ public void testPodConstruction() throws Exception {
         .getVolumeMounts().size());
   }
 
+  @Test
+  public void testPopulatingProxyUsersFromProject() throws Exception {
```
svnarumugam, line 527: [non-blocker] we are testing three scenarios using this single test. Having three separate tests would make it easy to reason about. Did we consider that @utk-12?

svnarumugam (commented):
Thanks for this optimization. Left a couple of comments. The changes look good overall.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1129,6 +1132,32 @@ private void createPod(final int executionId)
     if (flowParam != null && !flowParam.isEmpty()) {
       logger.info("ExecId: {}, Flow Parameters are: {}", executionId, flowParam);
     }
+
+    // Check if flow params has user.to.proxy If it does, skip the next step since this overrides
+    // all the project and node properties.
+    HashSet<String> proxyUsersMap = new HashSet<>();
+    if (flowParam != null && flowParam.containsKey("user.to.proxy")) {
+      proxyUsersMap.add(flowParam.get("user.to.proxy"));
+    }
```
ypadron-in, line 1141: Overwriting a USER_TO_PROXY value at runtime doesn't mean that's the value that is going to be used in the end, there are other ways this value can change which have higher precedence than runtime properties.

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
ypadron-in: I'm not able to understand what this method is doing. Could you please add a comment explaining? Please consider adding examples as well.

ypadron-in (changes requested):

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -144,9 +145,15 @@ public void addAllProxyUsers(final Collection<String> proxyUsers) {
     this.proxyUsers.addAll(proxyUsers);
   }
 
-  public Set<String> getProxyUsers() {
+  public HashSet<String> getProxyUsers() {
```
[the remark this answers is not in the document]
utk-12: ack. Changing this.

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +90,57 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+        String userToProxyFromNode = currentNodeProps.getString(USER_TO_PROXY, null);
+        // Get the node level override by user from the UI for proxy user.
+        Props currentNodeJobProps = projectManager.getJobOverrideProperty(project, flowObj,
+            node.getId(), node.getJobSource());
+        String userToProxyFromJobNode = currentNodeJobProps.getString(USER_TO_PROXY, null);
+        if (userToProxyFromJobNode != null) {
+          proxyUsers.add(userToProxyFromJobNode);
+        } else if (userToProxyFromNode != null) {
+          proxyUsers.add(userToProxyFromNode);
+        }
```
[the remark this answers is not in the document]
utk-12: hi, thank you for reviewing the PR. We have addressed that changing the user.to.proxy at runtime is not something we are supporting going forward as it is essential for us to prefetch credentials in our new security model ( where we need to obtain all custom credentials in the init container prior to flow start )as discussed internally. This is being done so we are intentionally unable to fetch [custom credentials](https://github.com/azkaban/azkaban/blob/8a65e94d8e9e268564cf2d864d7d7712998417c6/azkaban-hadoop-security-plugin/src/main/java/azkaban/security/AbstractHadoopSecurityManager.java#L284) after the flow starting in any way.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: ack. removing the if condition and adding it to the list of proxy users. thanks for catching this.

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -738,6 +738,8 @@ public static class ContainerizedDispatchManagerProperties {
         KUBERNETES_FLOW_CONTAINER_PREFIX + "init.jobtypes.mount.path";
     public static final String PREFETCH_PROXY_USER_CERTIFICATES =
         AZKABAN_CONTAINERIZED_PREFIX + "prefetch.certificates";
+    public static final String PREFETCH_JOBTYPE_PROXY_USER_MAP =
+        AZKABAN_CONTAINERIZED_PREFIX + "prefetch.jobtype.user.map";
```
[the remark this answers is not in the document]
utk-12: Added examples in the test and in the function getJobTypeUsersForFlow

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
utk-12: Ack. Will use 3 separate tests. thank you !

svnarumugam (dismissed):
Thanks for patiently responding to and incorporating the comments @utk-12 !

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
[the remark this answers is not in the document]
ypadron-in: I understand the problem however, even if USER_TO_PROXY changes at runtime aren't supported I believe this logic still doesn't resolve the property value correctly because upstream parent properties aren't taken into consideration. For example, let's say USER_TO_PROXY is set at flow level only and it's not specifically set by any jobs. What would be the resulting USER_TO_PROXY value for each job based on above code?

Also, It seems to me that 2 DB calls per enabled executable node may result from the `populateProxyUsersForFlow` method. In the case of complex DAGs have we analyzed how many of those do we have, how much time and number of queries will it take and what's the impact?

With regards to not allowing changing the USER_TO_PROXY value at runtime I'm a bit confused as to why are the flow parameters (or flow level runtime overwrites) checked in [KubernetesContainerizedImpl.java line ~1140 in that case?

```
if (flowParam != null && flowParam.containsKey("user.to.proxy")) {
      proxyUsersMap.add(flowParam.get("user.to.proxy"));
    }
```
In addition if USER_TO_PROXY can't be changed at runtime I believe we should add some sort of check to fail early and not the a middle of executions, that would probably result in a bad user experience.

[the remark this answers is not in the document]
utk-12: Sorry, I believe the objective here was unclear.
The goal here is **not** to determine the accurate proxy user required for a job. The goal is to enumerate all possible proxy users for a given flow so we can fetch appropriate custom credentials for them in the init container. 

Which is what we are doing here when we get it from the flow param -> job param --> override param. We add all of them and are not trying to evaluate which one will be used. This will be ultimately decided during execution .

Regarding the DB calls to get the node properties from the DAG, this evaluation has not been performed yet.

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +92,65 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+  }
+
+  // Extract the proxy users needed from  PREFETCH_JOBTYPE_PROXY_USER_MAP
+  // This method is being introduced to be able to assign a specific proxy user that will require
+  // custom credentials for a given job type. This allows the jobtype to perform specific checks
+  // without requiring azkaban executor's credentials to do this, once we enforce POLP defined in
+  // https://github.com/azkaban/azkaban/pull/3216
+  // This will parse the jobTypePrefetchUserMap of the format : "jobtype1,jobtype1_proxyuser;
+  // jobtype2,jobtype2_proxyuser" and add the proxy user for a given job if that jobtype is
+  // present in the flow.
```
ypadron-in, line 140: Thanks for adding the comment!
nit: Could you make it a method doc comment by replacing `//` with `/**...*/`

Also, I'm wondering, will the jobs be able to impersonate the jobtypes proxy users? How is that prevented?

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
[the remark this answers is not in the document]
ypadron-in: My point is that the logic is not gathering all possible proxy users therefore it can lead to flow failure and very bad user experience because I don't think the reason for failure will be clearly stated in the logs

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
utk-12: Will change this thanks !
potentially yes. if they break out of the job process. However, the impact will be limited to access to the other users prefetched, which is an improvement over the current situation of the azkaban executor user.

azkaban-common/src/main/java/azkaban/container/models/AzKubernetesV1SpecBuilder.java
```diff
@@ -32,6 +32,7 @@
 import io.kubernetes.client.openapi.models.V1VolumeMount;
 import io.kubernetes.client.openapi.models.V1VolumeMountBuilder;
 import java.util.ArrayList;
+import java.util.HashSet;
```
ypadron-in, line 35: Could you remove this import since it is not used afterwards?

amielkelinkedin: Anyway to add this to the linter, so builds fail when unused imports are present?

utk-12: removing this for now, however all files in the project have some unused imports. Lets tackle that as a separate task.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+     also provided with a flow parameter.
+     */
+
+    if (flow.getProxyUsers().size() > this.proxyUserPrefetchThreshold || !prefetchAllProxyUserCredentials) {
+        proxyUsersMap.addAll(ContainerImplUtils
+            .getProxyUsersForFlow(this.projectManager, flow));
+      }
+      else{
+        proxyUsersMap.addAll(flow.getProxyUsers());
+      }
```
ypadron-in, line 1176: nit: indentation doesn't seem to be correct.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -354,6 +360,10 @@ public KubernetesContainerizedImpl(final Props azkProps,
     this.azkabanSecurityInitImageName = this.azkProps
         .getString(ContainerizedDispatchManagerProperties.KUBERNETES_POD_AZKABAN_SECURITY_INIT_IMAGE_NAME,
             DEFAULT_AZKABAN_SECURITY_INIT_IMAGE_NAME);
+    this.jobTypePrefetchUserMap =
+        this.azkProps.getString(ContainerizedDispatchManagerProperties.PREFETCH_JOBTYPE_PROXY_USER_MAP, null);
```
ypadron-in, line 364: Please consider default value empty String instead of null to avoid unnecessary null checks.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+      }
+      else{
+        proxyUsersMap.addAll(flow.getProxyUsers());
+      }
+    // If certain jobtypes need specific user credentials we add them to the prefetch list
+    if (this.jobTypePrefetchUserMap != null) {
+      final Set<String> jobTypeUsersForFlow =
+          ContainerImplUtils.getJobTypeUsersForFlow(this.jobTypePrefetchUserMap, jobTypes);
+          proxyUsersMap.addAll(jobTypeUsersForFlow);
+      }
```
ypadron-in, line 1182: Wondering if we should take the number of jobtype related proxy users into consideration when deciding which path to take: use the proxy users in the Permissions page or estimating proxy users for the current flow execution.
For example if there are 4 proxy users declared in the Permissions Page we'll not traverse the DAG to estimate proxy users. However if another 4 proxy users for jobtype usage are added to the prefect list then that might be too much and we would better have done the proxy users estimation.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+    // We add the submit user if, no proxy user is mentioned. The submit user is the proxy user
+    // in this case.
+    proxyUsersMap.add(flow.getSubmitUser());
```
ypadron-in, line 1185: Should we do this only if the proxyUsersMap is empty BEFORE adding jobtype proxy users?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+
+    // Finally, set the collected list of users as a flow object variable to be accessed while
+    // creating pod spec template.
+    flow.setProxyUsersFromFlowObj(proxyUsersMap);
+    logger.info("Proxy Users determined for this flow are " + String.join(", ", proxyUsersMap));
```
ypadron-in, line 1190: Could you please add the execution id to the log?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+     try to check if the proxy users list from the project page is less then a certain threshold.
+     If it is, less than this threshold, we simply add all of them. This configuration will be
+     custom to a given environment. An option to fetch all of them via the project page is
+     also provided with a flow parameter.
+     */
+
+    if (flow.getProxyUsers().size() > this.proxyUserPrefetchThreshold || !prefetchAllProxyUserCredentials) {
+        proxyUsersMap.addAll(ContainerImplUtils
+            .getProxyUsersForFlow(this.projectManager, flow));
+      }
```
ypadron-in, line 1173: Please consider logging when the route of estimating the proxy users for current execution is taken.

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +93,82 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+     flow and not the job. Passing null as the job name is able to get us the top level flow
+     properties mentioned for the flow.*/
+
+    Props flowProps = projectManager.getProperties(project, flowObj,
+        null, flow.getJobSource());
+    if (flowProps != null) {
+      String proxyUserFromFlowProp = flowProps.getString(USER_TO_PROXY, null);
+      if (proxyUserFromFlowProp != null && !proxyUserFromFlowProp.isEmpty()) {
+        proxyUsers.add(proxyUserFromFlowProp);
+      }
```
ypadron-in, line 114: Please consider defaulting to empty string in line 111 to avoid the null check in line 112.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+     As an optimization for not walking through the DAG and making DB queries for large DAGs we
+     try to check if the proxy users list from the project page is less then a certain threshold.
+     If it is, less than this threshold, we simply add all of them. This configuration will be
+     custom to a given environment. An option to fetch all of them via the project page is
+     also provided with a flow parameter.
+     */
+
+    if (flow.getProxyUsers().size() > this.proxyUserPrefetchThreshold || !prefetchAllProxyUserCredentials) {
+        proxyUsersMap.addAll(ContainerImplUtils
+            .getProxyUsersForFlow(this.projectManager, flow));
```
ypadron-in, line 1172: Could you please measure and log the time `getProxyUsersForFlow` takes to run?

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +93,82 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+        populateProxyUsersForFlow(flow, subNode, flowObj, project, projectManager, proxyUsers);
+      }
+    } else {
+      // If a node is disabled, we don't need to initialize its jobType container image when
+      // creating a pod.
+      if (node.getStatus() != Status.DISABLED) {
+        Props currentNodeProps = projectManager.getProperties(project, flowObj,
+            node.getId(), node.getJobSource());
+        // Get the node level property for proxy user.
+        String userToProxyFromNode = (currentNodeProps != null) ? currentNodeProps.getString(USER_TO_PROXY, null) : null;
```
ypadron-in, line 136: Please consider defaulting to empty string here and in line 140 below to avoid unnecessary null checks and improve readability.

ypadron-in:
Thanks a lot for adding plenty of comments!

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: The job type proxy users are independent and will be required regardless of the total count from the Permissions page. They will never be in the permissions page.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: ack

ypadron-in:
In the cases where we traversed the DAG to gather potential proxy users for current execution but let's assume one proxy users was missed and therefore the flow failed at runtime. 
1-What error will the user see? Is it indicative of what happened? 
2-How are we letting the user know how to resolve the issue? 
3-Also, how are we recording the cases where the proxy user was not in the estimated list? 
Will there be a follow up PR covering these?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: What if the one job has a proxy user defined and and other does not ? If there are flow flow parameteres or properties. The expected behaviour is to use the submit user. So adding it to this list makes sense IMO.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1141,54 @@ private void createPod(final int executionId)
...
+     not be able to fetch them at runtime as Azkaban user present in the Azkaban Executor Pod
+     will no longer be able to do so. This prefetching of custom credentials  is performed in
+     the azkaban-security-init container by attaching the proxy users needed to be prefetched as
+     environment variables.
+     */
+    Set<String> proxyUsersMap = new HashSet<>();
+
+    final boolean prefetchAllProxyUserCredentials =
+        flowParam != null && flowParam.containsKey(FlowParameters.PROXY_USER_PREFETCH_ALL) &&
+            Boolean.parseBoolean(flowParam.get(FlowParameters.PROXY_USER_PREFETCH_ALL));
```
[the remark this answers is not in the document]
utk-12: using similar methodology as line 1141( different commit) . Not sure if replacing it with an empty map here will affect other things. Is there any downside of keeping it the way it is ?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
ypadron-in, line 1182: It can be addressed in a separate PR but perhaps the list of jobtype proxy users can fetched from the pod spec template instead of adding them to the web server configuration. The web server probably shouldn't be aware of jobtype details and by leveraging the pod spec template the web server config is more clean. 

The pod spec template can also be leveraged to specify proxy users for which tokens always need to be prefetched (ie: the user to download the project zips). Where are those required proxy users defined right now?

utk-12: _The pod spec template can also be leveraged to specify proxy users for which tokens always need to be prefetched_

That is the plan ^ . 
However for the jobtype proxy users we need to programmatically determine if that jobtype is present. it cannot be present by default in the pod spec template.

ypadron-in: Great!
Maybe I didn't explain it well, I'm just suggesting to move the web server config `prefetch.jobtype.user.map` to the pod spec template. In the code, we would fetch that info from the template instead of the webserver config here:
```
this.jobTypePrefetchUserMap =
      this.azkProps.getString(ContainerizedDispatchManagerProperties.PREFETCH_JOBTYPE_PROXY_USER_MAP, null);
```
Everything else remains the same.

ypadron-in: Of course the jobtype proxy users won't go in the same place as the always-present proxy users.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1131,6 +1142,56 @@ private void createPod(final int executionId)
...
+    /*
+     As an optimization for not walking through the DAG and making DB queries for large DAGs we
+     try to check if the proxy users list from the project page is less then a certain threshold.
+     If it is, less than this threshold, we simply add all of them. This configuration will be
+     custom to a given environment. An option to fetch all of them via the project page is
+     also provided with a flow parameter.
+     */
+
+    if (flow.getProxyUsers().size() > this.proxyUserPrefetchThreshold || prefetchAllProxyUserCredentials) {
```
ypadron-in, line 1171: The comment above indicates that `prefetchAllProxyUserCredentials=true` means 'prefetch all proxy users listed in the project page', however the code says if true to traverse the DAG to gather proxy users for current execution. Could you please clarify what's the expected behavior?

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +93,80 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+    while (st.hasMoreTokens()) {
+      StringTokenizer stInner = new StringTokenizer(st.nextToken(), ",");
+      String jobType = stInner.nextToken();
+      String jobTypeUser = stInner.nextToken();
+      if (jobTypes.contains(jobType)) {
+        jobTypeProxyUserSet.add(jobTypeUser);
+      }
+    }
+    return jobTypeProxyUserSet;
+  }
```
ypadron-in, line 171: Since the `jobTypePrefetchUserMap` will not change from one execution to the next, to avoid parsing this string everytime can we please do it just once in the constructor of KubernetesContainerizedImpl.java? The result will be a map of <jobtype1, jobtype1_proxyuser> which can be iterated later to populate jobTypeProxyUserSet. That should be much faster. 
Also, please trim PREFETCH_JOBTYPE_PROXY_USER_MAP when parsing to account for typos.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1342,7 +1404,11 @@ private void addInitContainers(final ExecutableFlow executableFlow,
       final VersionSet versionSet)
       throws ExecutorManagerException {
     final ExecutableFlow flow = executableFlow;
-    final Set<String> proxyUserList = flow.getProxyUsers();
+    final Set<String> proxyUserList = flow.getProxyUsersFromFlowObj();
+    // Remove all empty strings and null in this set.
+    if (proxyUserList != null ) {
+      proxyUserList.removeAll(Collections.singleton(""));
+    }
```
ypadron-in, line 1411: Can we please move the call to `removeAll` to the method where the `proxyUsersMap` is created? I believe the null check isn't needed as it's guaranteed it will be at least an empty Map.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1372,7 +1438,7 @@ private void addInitContainers(final ExecutableFlow executableFlow,
             dependency + " in versionSet");
       }
     }
-    if (this.prefetchAllCredentials) {
+    if (this.prefetchAllCredentials && proxyUserList.size() > 0) {
```
ypadron-in, line 1441: Why adding the size check? Is it guaranteed that it will always contain at least the submit user?

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
utk-12: I'll need the jobtypes set to make this set in the constructor. Which is a DAG walk. If i do this in the constructor. Its one extra DAG walk. Not sure if there's any other way to achieve this.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
utk-12: sorry, i think i removed the not condition. Adding it back .

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
utk-12: resolved this by creating a new function

azkaban-common/src/test/java/azkaban/executor/container/KubernetesContainerizedImplTest.java
```diff
@@ -617,6 +632,92 @@ public void testVersionSetConstructionWithRampupManager() throws Exception {
...
+
+    Assert.assertTrue(proxyUsers.contains("testUser2"));
+    Assert.assertEquals(2, proxyUsers.size());
+  }
+
+  @Test
+  public void testPopulatingJobTypeUsersForFlow() throws Exception {
+    Set<String> proxyUsers;
+    TreeSet<String> jobTypes = new TreeSet<>();
+    jobTypes.add("jobtype1");
```
ypadron-in, line 708: Can be done in a subsequent PR but perhaps the most useful thing to test here is `parseJobTypeUsersForFlow()`, making sure that spaces, empty string and malformed strings won't break the code.

ypadron-in (dismissed):
Thanks for addressing all the comments!

There are a few lines commented out through the code that we should probably remove in subsequent PRs but otherwise the code looks good overall. 
```
//when(currentNodeProps2.getString("user.to.proxy", "")).thenReturn("testUser2");
```
```
// Remove all empty strings and null in this set.
```

ypadron-in (dismissed):

azkaban-common/src/main/java/azkaban/executor/ExecutableFlowBase.java
```diff
@@ -123,6 +123,11 @@ public String getFlowId() {
     return this.flowId;
   }
 
+  // Top level flow file name
+  public String getFlowFileName(){
+    return this.flowId + ".flow";
+  }
+
```
ypadron-in, line 130: Is there another way to get this info? This would cover flow 2.0 syntax but not flow 1.0 (.job + .properties files). We shouldn't hardcode these extensions here anyways IMO.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -312,7 +312,13 @@ public Props getProperties(final Project project, final Flow flow, final String
...
     if (FlowLoaderUtils.isAzkabanFlowVersion20(flow.getAzkabanFlowVersion())) {
       // Return the properties from the original uploaded flow file.
-      return getPropertiesFromFlowFile(flow, jobName, source, 1);
+      if (jobName != null) {
+        return getPropertiesFromFlowFile(flow, jobName, source, 1);
+      } else {
+        // If job is null we intend to fetch the top level flow properties, and flowID does not
+        // have the .flow extension suffix.
+        return getPropertiesFromFlowFile(flow, null, source + Constants.FLOW_FILE_SUFFIX, 1);
+      }
```
ypadron-in, line 321: I tested this change, it breaks fetching the flow properties page for flow 2.0 dags (we get to that page by clicking in the link under `Inherited Properties` section in the screenshot attached.)

To cover the cases of flow 2.0 in `ContainerImplUtils.getProxyUsersForFlow()` could you try something like this?
```
String flowFileName = null;
if(FlowLoaderUtils.isAzkabanFlowVersion20(flow.getAzkabanFlowVersion())) {
  flowFileName = flow.getFlowId() + Constants.FLOW_FILE_SUFFIX;
}
....
Props flowProps = projectManager.getProperties(project, flowObj, flowFileName, flow.getFlowId());
```
Note that this works IF the flow id matches the name of the dag file **IF there is no such requirement/validation then there is a chance that this code won't work either**.

Screenshot 2023-03-05 at 1 05 43 AM

utk-12: reverted.

ypadron-in (dismissed):
Changes LGTM, thanks!

azkaban-common/src/main/java/azkaban/executor/container/ContainerImplUtils.java
```diff
@@ -82,4 +96,103 @@ public static int getFlowNameHashValMapping(ExecutableFlow flow) {
...
+      flowProps.ifPresent(s -> proxyUsers.add(s.getString(USER_TO_PROXY, "")));
+    } else {
+      // Handle Flow 1.0 Properties fetch.
+      final Map<String, String> flow10Props = new HashMap<>();
+      for (String propsFile : flowObj.getAllFlowProps().keySet()) {
+        flow10Props
+            .putAll(projectManager.getProperties(project, flowObj, null, propsFile).getFlattened());
+      }
+      proxyUsers.add(flow10Props.getOrDefault(USER_TO_PROXY, ""));
+    }
```
ypadron-in, line 126: We might need to revisit this logic in the future depending on how many times `getProxyUsersForFlow` gets called and how long it takes to run for now it should be ok. Issues with it are:
-to work flow id must match the name of the flow file for Flow 2.0.
-DB may be called multiple times to fetch Flow 1.0 props.

utk-12: ack. added TODO.

ypadron-in (approved):

PR #3239, opened 2023-02-24
UpdateProperty endpoint: if not admin, directly return error message and no further operation

tsangz2013:
**What's the problem**
The old code leaves a security hole which allow anyone to hit the UpdateProperty endpoint.

**What's done**
directly return error message, no further operation

**Test done**
Tested in a test cluster, asked some non-admin person to hit the endpoint, and the request is blocked, the property is not changed from status page

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -1065,6 +1065,7 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
       if (!HttpRequestUtils.hasPermission(this.userManager, user, Type.ADMIN)) {
         ret.put("error", String.format("User %s doesn't have ADMIN permission for updating "
             + "property", user));
+        return;
       }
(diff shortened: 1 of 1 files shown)


byjiang1996 (approved):

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -1065,6 +1065,7 @@ private void ajaxUpdateProperty(final HttpServletRequest req,
       if (!HttpRequestUtils.hasPermission(this.userManager, user, Type.ADMIN)) {
         ret.put("error", String.format("User %s doesn't have ADMIN permission for updating "
             + "property", user));
+        return;
```
tsangz2013, line 1068: For your reference, the `ajaxRampActions` method also rely on `hasPermission` and will just ignore all non-admin ppl:
https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java#L1174

tsangz2013:
Travis CI check always fails

PR #3240, opened 2023-02-24
UpdateProperty endpoint: if not admin, directly return error and skip operation

tsangz2013:
**What's the problem**
The old code leaves a security hole which allow anyone to hit the UpdateProperty endpoint.

**What's done**
directly return error message, no further operation

**Test done**
Tested in a test cluster, asked some non-admin person to hit the endpoint, and the request is blocked, the property is not changed from status page

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
@@ -1065,6 +1065,7 @@ public class ExecutorServlet extends LoginAbstractAzkabanServlet {
       if (!HttpRequestUtils.hasPermission(this.userManager, user, Type.ADMIN)) {
         ret.put("error", String.format("User %s doesn't have ADMIN permission for updating "
             + "property", user));
+        return;
       }
       String propType = getParam(req, "propType");
       if (propType.equals("containerDispatch")) {


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -1065,6 +1065,7 @@ private void ajaxUpdateProperty(final HttpServletRequest req,
       if (!HttpRequestUtils.hasPermission(this.userManager, user, Type.ADMIN)) {
         ret.put("error", String.format("User %s doesn't have ADMIN permission for updating "
             + "property", user));
+        return;
```
tsangz2013, line 1068: For your reference, the `ajaxRampActions` method also rely on `hasPermission` to check, and will just ignore all non-admin ppl:
https://github.com/azkaban/azkaban/blob/master/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java#L1174

byjiang1996 (approved):

svnarumugam (approved):
Thanks @tsangz2013 for the fix. should we use this opportunity to double-check other admin APIs to make sure azkaban is doing fine? :)

tsangz2013:
> Thanks @tsangz2013 for the fix. should we use this opportunity to double-check other admin APIs to make sure azkaban is doing fine? :)

Good point, let's close this first and do a short but joint walk-through for all endpoints with the team later

PR #3243, opened 2023-02-28
increase single yarn application tag to 250

ShellyWEI:
To avoid truncated application_tag with insufficient flow/job description.
Local build pass with test successful.
Screenshot 2023-02-27 at 17 00 15


utk-12 (commented):
hey ! Any reason why the test has 350 and the config is set to 250 ?

tsangz2013 (approved):
lgtm

ShellyWEI:
> hey ! Any reason why the test has 350 and the config is set to 250 ? Is it just to test if its going over the set limit of 250 ?

Yes! Just the test purpose, it can be any value over 250

PR #3244, opened 2023-02-28
Introduce VPAFlowCriteria to deny VPA feature based on project or flow name filter

byjiang1996:
Project and flow name filter file looks like this:

proj1:flow1
proj1:flow2
proj2
proj3:flow1

which means VPA feature will not apply to proj1's flow1 flow, proj1's flow2 flow, all proj2's flows and proj3's flow1 flow.

It serves as the safeguard if there is any flow in production does not fit well with VPA's autoscaling algorithm.

One can use azkaban web api to reload flow filter text at runtime


ShellyWEI (approved):
lgtm

PR #3246, opened 2023-03-01
Support configuration of alerter plugins via Web UI.

ypadron-in:
This is a follow up on https://github.com/azkaban/azkaban/pull/3232. It:
-adds custom alerter configurations to the responses of `slaInfo` API.
-allows custom alerters to be configured via UI for SLA miss events. Flow execution failure alerts are configured via properties so no UI changes are required to support those. 

Assuming there is custom alerter plugin named `myCustom` which can be enabled by configuring a parameter called `myCustomAlerterParam` here is how the SLA definition page will be rendered.

 
Screenshot 2023-02-28 at 10 36 33 PM

Tested the changes in a local Azkaban instance. Verified that SLA miss alerts are set through UI successfully and that the feature doesn't break existing SLAs. In addition, I verified that the custom alerters configs are added correctly to the `slaInfo` responses. Here is an example response:

```
{
    "settings": [
        {
            "duration": "1m",
            "rule": "SUCCEED",
            "id": "",
            "actions": [
                "EMAIL",
                "KILL"
            ]
        }
    ],
    "slaEmails": [
        "ypadron@ypadron.com"
    ],
    "slaAlerters": {
        "myCustom": {
            "myCustomAlerterParam": [
                "Some value"
            ]
        },
        "email": {
            "recipients": [
                "ypadron@ypadron.com"
            ]
        }
    },
    "allJobNames": [
        "myFlow",
        "job1"
    ]
}
```

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
@@ -19,6 +19,7 @@ import static azkaban.server.HttpRequestUtils.getMapParamGroup;
 import static azkaban.server.HttpRequestUtils.getParam;
 import static azkaban.server.HttpRequestUtils.getParamGroup;
 
+import azkaban.Constants;
 import azkaban.sla.SlaAction;
 import azkaban.sla.SlaOption;
 import azkaban.sla.SlaOption.SlaOptionBuilder;
@@ -50,8 +51,7 @@ public class SlaRequestUtils {
     final Map<String, String> settings = getParamGroup(req, settingsParamName);
     final Map<String, Map<String, String>> alertersParams =
         getMapParamGroup(req, PARAM_SLA_ALERTERS);
-    final String slaEmailsStr = getParam(req, PARAM_SLA_EMAILS,
-        null);
+    final String slaEmailsStr = getParam(req, PARAM_SLA_EMAILS, null);
 
     // Don't allow combining old & new in the same request:
     // This ensures that there's no need to handle possible conflicts between 'slaEmails' and
@@ -112,18 +112,11 @@ public class SlaRequestUtils {
     if (str == null || str.trim().isEmpty()) {
       return Collections.emptyList();
     }
-    String delimRegex = "\\s*,\\s*|\\s*;\\s*";
+    String delimRegex = Constants.ConfigurationKeys.ALERTER_PARAM_VALUE_DELIMITER;
     if (spacesAreDelims) {
       delimRegex = "\\s*,\\s*|\\s*;\\s*|\\s+";
     }
     return Arrays.asList(str.trim().split(delimRegex));
-  }
-
-  public static Set<String> getSetFromString(final String val) {
-    if (val == null || val.trim().length() == 0) {
-      return Collections.emptySet();
-    }
-    return new HashSet<>(Arrays.asList(val.split("\\s*,\\s*")));
   }
 
   private static SlaOption parseSlaSetting(final String set, final String flowName,
@@ -171,8 +164,6 @@ public class SlaRequestUtils {
     if (actions.isEmpty()) {
       throw new ServletException("Unable to create SLA as there is no action set");
     }
-    logger.info("Parsing sla as id:" + id + " type:" + type + " sla:"
-        + rule + " Duration:" + duration + " actions:" + actions);
     return new SlaOptionBuilder(type, flowName, dur).setJobName(id).setActions(actions)
         .setEmails(emails).setAlertersConfigs(alertersConfigs).createSlaOption();
   }
--- a/azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
@@ -93,6 +93,18 @@ public class SlaRequestUtils {
     return slaOptions;
   }
 
+  /**
+   * Transforms a String Map that looks like
+   * {
+   *  "key1", "value1,value2,value3"
+   *  "key2", "value4,value5,value6"
+   * }
+   * into a Map like this:
+   * {
+   *  "key1", ["value1", "value2", "value3"]
+   *  "key2", ["value4", "value5", "value6"]
+   * }
+   */
   private static Map<String, Map<String, List<String>>> getStringListMapFromStringMap(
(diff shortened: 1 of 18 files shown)


azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -85,8 +93,42 @@ public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, fina
...
+      result.put(alerter, processedConfigs);
+    }
+    return result;
+  }
+
+  private static List<String> getListFromString(final String str, final boolean spacesAreDelims) {
+    if (str == null || str.trim().isEmpty()) {
+      return Collections.emptyList();
+    }
+    String delimRegex = "\\s*,\\s*|\\s*;\\s*";
```
amielkelinkedin, line 115: Let's use ALERTER_PARAM_VALUE_DELIMITER here?

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -85,8 +93,42 @@ public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, fina
...
+    if (spacesAreDelims) {
+      delimRegex = "\\s*,\\s*|\\s*;\\s*|\\s+";
+    }
+    return Arrays.asList(str.trim().split(delimRegex));
+  }
+
+  public static Set<String> getSetFromString(final String val) {
+    if (val == null || val.trim().length() == 0) {
```
amielkelinkedin, line 123: Thoughts on isEmpty?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -183,6 +185,12 @@ public void init(final ServletConfig config) throws ServletException {
     this.downloadBufferSize =
         server.getServerProps().getInt(PROJECT_DOWNLOAD_BUFFER_SIZE_IN_BYTES, 8192);
     logger.info("downloadBufferSize: " + this.downloadBufferSize);
+
+    final Map<String, List<HTMLFormElement>> alerterPlugins = new HashMap<>();
+    server.getAlerterPlugins().forEach((name, alerter) -> alerterPlugins.put(name,
```
amielkelinkedin, line 190: A lambda and ternary operator! Nice!

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -85,8 +93,42 @@ public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, fina
...
+      return Collections.emptySet();
+    }
+    return new HashSet<>(Arrays.asList(val.split("\\s*,\\s*")));
```
amielkelinkedin, line 126: Can we abstract "\\s*,\\s*" ? Seems like magic to me...

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -85,8 +93,42 @@ public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, fina
     return slaOptions;
   }
 
-  private static SlaOption parseSlaSetting(final String set, final String flowName, final List<String> emails,
-      final Map<String, Map<String, String>> alertersConfigs) throws ServletException {
+  private static Map<String, Map<String, List<String>>> getStringListMapFromStringMap(
```
amielkelinkedin, line 96: Might want to consider adding a test for this in the future, as that might clarify what is happening a little :)

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
```diff
@@ -85,8 +93,35 @@ public static List<SlaOption> parseSlaOptions(final HttpServletRequest req, fina
...
+    return result;
+  }
+
+  private static List<String> getListFromString(final String str, final boolean spacesAreDelims) {
+    if (str == null || str.trim().isEmpty()) {
+      return Collections.emptyList();
+    }
+    String delimRegex = Constants.ConfigurationKeys.ALERTER_PARAM_VALUE_DELIMITER;
+    if (spacesAreDelims) {
+      delimRegex = "\\s*,\\s*|\\s*;\\s*|\\s+";
```
amielkelinkedin, line 117: Can this be delimRegex += |\\s+ ?

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: yes, done

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: Thanks for pointing this out, that method shouldn't be here. I temporarily copied it from ServerUtils.java file to merge `getSetFromString` and `getListFromString` since both do similar things but I decided to do it in a separate commit. I forgot to delete the method. Deleted now. I'll make sure to use isEmpty() when refactoring :)

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: Deleted the method, it shouldn't be here as explained above.

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: I thought that might make it harder to understand the regular expression therefore I kept it literal

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
ypadron-in: Added a comment in the meantime, hopefully it helps to clarify what the method does, thanks!

azkaban-common/src/main/java/azkaban/server/SlaRequestUtils.java
amielkelinkedin: Makes sense!

amielkelinkedin (approved):
Thank you so much for the change, Yeni!

PR #3247, opened 2023-03-02
subflow support fixes

ShellyWEI:
- set "true" case for isRootFlow property in FlowLifeCycleEvent;
- pass subFlow info to yarn tag

Screenshot 2023-03-01 at 19 18 43


tsangz2013 (approved):
lgtm!

PR #3248, opened 2023-03-02
Job level SLAs checks should only be set for/on the job the SLA rule matches.

ypadron-in:
Before all existing job level SLAs were set up for each job node in the DAG. In the logs we would see entries like this:
```
2023/03/01 22:38:01.172 -0800  INFO [TriggerManager] [JobRunner-jobZ-14213] [Azkaban] Adding sla trigger SlaOption{type=JOB_SUCCEED, flowName='myFlow', jobName='jobE', duration=PT1M, actions=[ALERT], emails=[abc@abc.com], alertersConfigs={}} to execution 14213, scheduled to trigger in 60 seconds

2023/03/01 22:41:51.873 -0800  INFO [TriggerManager] [JobRunner-jobE-14213] [Azkaban] Adding sla trigger SlaOption{type=JOB_SUCCEED, flowName='myFlow', jobName='jobE', duration=PT1M, actions=[ALERT], emails=[abc@abc.com], alertersConfigs={}} to execution 14213, scheduled to trigger in 60 seconds
```
Note that there is a job SLA rule defined for job `jobE` which is set up in `jobZ` even though that is not the target job.
 
Now job level SLAs checks are scheduled only for the job the SLA rule targets.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/sla/SlaOption.java
+++ b/azkaban-common/src/main/java/azkaban/sla/SlaOption.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * SLA option, which can be associated with a flow or job.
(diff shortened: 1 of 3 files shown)


inramana (dismissed):
LGTM!

inramana (approved):

PR #3254, opened 2023-03-06
Sync Schedule with Trigger when web sever is up

ShellyWEI:
Years ago, we separate schedule and trigger in favor to unify the use of trigger in the backend but aborted in the middle. It creates the sync issues between both. We have "TriggerBasedScheduleLoader" as the intermediate bridge to coordinate while trigger updates and ensure "schedule" updated to serve API request well and correctly.
This PR is to fix the missing sync on web server just restart and loading all the triggers into schedule format so schedule API can serve request immediately on restart, no need to wait for refresh to update them.

local build pass:
Screenshot 2023-03-01 at 19 18 43


tsangz2013:
code ltgm, please test in a test webserver

ShellyWEI:
> code ltgm, please test in a test webserver

It is tested, you can verify on faro4, I've kept the code deployed there. @tsangz2013

tsangz2013 (approved):
cool, thank you!

PR #3255, opened 2023-03-06
Adding ability to set Service Links in Pod Spec

amielkelinkedin:
Adding Ability to set Service Links In Pod Spec Template allowing for the modification of Kubernetes Behavior, especially ENV variablles.

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -693,7 +693,7 @@ public class Constants {
         KUBERNETES_POD_PREFIX + "service.account.token.automount";
 
     public static final String KUBERNETES_POD_SERVICE_LINKS =
-        KUBERNETES_POD_PREFIX + "service.account.token.automount";
+        KUBERNETES_POD_PREFIX + "service.account.links";
 
     // Kubernetes flow container related properties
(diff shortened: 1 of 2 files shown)


amielkelinkedin:
This builds locally, and runs on internal servers.

inramana (dismissed):
LGTM

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -692,6 +692,9 @@ public static class ContainerizedDispatchManagerProperties {
     public static final String KUBERNETES_POD_SERVICE_ACCOUNT_TOKEN_AUTOMOUNT =
         KUBERNETES_POD_PREFIX + "service.account.token.automount";
 
+    public static final String KUBERNETES_POD_SERVICE_LINKS =
+        KUBERNETES_POD_PREFIX + "service.account.links";
```
sshardool, line 696: Consider renaming to something like `inject.service.links` ?
This is not service account related and could be confusing.

ypadron-in (approved):
LGTM

sshardool (commented):
Looks good overall, with one comment. 
Also  - please add a unit test for this (can be in a follow up commit if this is urgent)

az-core/src/main/java/azkaban/Constants.java
amielkelinkedin: Will just change to service links. thanks!

sshardool (approved):

PR #3256, opened 2023-03-07
Include executions with EXECUTION_STOPPED status in the 'Recently Finished' page

ypadron-in:
Fixes a bug where executions with final status EXECUTION_STOPPED were not shown in the page of Recently Finished Executions.


ypadron-in:
> what was the trigger to this change? Did we get this from customers? Thanks @ypadron-in for fixing this.

Thanks for reviewing! I realized the bug while working on another issue.

PR #3257, opened 2023-03-09
Added a env variable security init container to verify source of uplo…

utk-12:
…ad user

Objective here is to be able to see who the upload user was, for the project / code in order to allow /deny access to certain resources based on their credibility. This can be done by custom logic in the security init container if present. 

Tested on cluster to verify the environment variable is assigned to init container.


djaiswal83 (approved):

PR #3259, opened 2023-03-13
fix the bug that returned the outdated schedule metadata for getSchedule call

ShellyWEI:
For the legacy reason schedule and trigger are self-maintained and schedule itself maintains a cache to get trigger data more quickly especially when no update on trigger. 
The better and cleaner fix is to remove the schedule layer but link API to trigger directly, in this way we don't have to maintain the status of schedule. However to quickly mitigate this issue, in this PR:
- fix the bug that store new schedule in cache;
- enable more debug level log to help increase debug efficiency.

local build passed.


tsangz2013 (approved):
nice catch!

PR #3261, opened 2023-03-22
Add customizable flow-level restart parameters & validation with Azkaban config

tsangz2013:
**Why we need this**
This is the initial user-interface/contract definition for flow/execution-level restart when one fails.

**What's done**
Defined 3 flow-parameter keys that is exposed to users: 
- `allow.execution.restart.on.status` that describe a set of terminated status when an execution can restart
- `"execution.restart.count"` == how many times can an execution restarts
- `"execution.restart.strategy"` == the strategy (e.g. "restart_from _root", "disable_finished_nodes")

Also Azkaban side controlled config:
- `"azkaban.execution.restartable.status"` the restartable-status allowlist for user to pick from
- `"azkaban.execution.restart.limit"` the limitation how many times an execution can restart

Plus the cross-validation at execution creation time:
- the `user-input status list` against `Azkaban allowlist`
- the `user-input restart count` against `Azkaban restart limit`

**Tests done**
- unit tests
- tested in staging cluster:
Screenshot 2023-03-22 at 5 47 38 PM
Screenshot 2023-03-22 at 5 46 35 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -16,17 +16,16 @@
 
 package azkaban.executor;
 
-import static azkaban.Constants.ConfigurationKeys.AZKABAN_EXECUTION_RESTARTABLE_STATUS;
 import static azkaban.Constants.ConfigurationKeys.AZKABAN_EXECUTION_RESTART_LIMIT;
 import static azkaban.Constants.EventReporterConstants.EXECUTION_RETRIED_BY_AZKABAN;
 import static azkaban.Constants.EventReporterConstants.ORIGINAL_FLOW_EXECUTION_ID_BEFORE_RETRY;
+import static azkaban.executor.Status.RESTARTABLE_TERMINATED_STATUSES;
 
 import azkaban.Constants.FlowParameters;
 import azkaban.executor.mail.DefaultMailCreator;
 import azkaban.sla.SlaOption;
 import azkaban.utils.Props;
 import azkaban.utils.TypedMapWrapper;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.GsonBuilder;
 import java.util.ArrayList;
@@ -49,9 +48,7 @@ public class ExecutionOptions {
   /* override dispatcher selection and use executor id specified */
   public static final String USE_EXECUTOR = "useExecutor";
   public static final int DEFAULT_FLOW_PRIORITY = 5;
-  public static final ImmutableList<String> DEFAULT_EXECUTION_RESTARTABLE_STATUS =
-      ImmutableList.of(Status.EXECUTION_STOPPED.name(), Status.FAILED.name());
-  public static final int DEFAULT_EXECUTION_RESTART_LIMIT = 2;
+  public static final int DEFAULT_FLOW_RESTART_LIMIT = 2;
 
   private static final String FLOW_PARAMETERS = "flowParameters";
   private static final String RUNTIME_PROPERTIES = "runtimeProperties";
@@ -188,9 +185,6 @@ public class ExecutionOptions {
       return;
     }
     if (flowParameters.containsKey(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS)) {
-      // allow list defined in azProps
-      final List<String> allowedStatuses = azProps.getStringList(
-          AZKABAN_EXECUTION_RESTARTABLE_STATUS, DEFAULT_EXECUTION_RESTARTABLE_STATUS);
 
       // user defined list
       final String[] statuses = flowParameters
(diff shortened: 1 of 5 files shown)


azkaban-common/src/test/java/azkaban/executor/ExecutionOptionsTest.java
```diff
@@ -0,0 +1,104 @@
+package azkaban.executor;
```
djaiswal83, line 1: Missing header

tsangz2013: nice catch

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -1010,8 +1010,10 @@ private void ajaxExecuteFlow(final HttpServletRequest req,
     exflow.setExecutionSource(Constants.EXECUTION_SOURCE_ADHOC);
 
     final ExecutionOptions options;
+    final Props azProps = getApplication().getServerProps();
     try {
       options = HttpRequestUtils.parseFlowOptions(req, flowId);
+      options.validate(azProps);
```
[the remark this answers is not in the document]
tsangz2013: good idea, for now it's only checking flow-parameters, then let's use `validateFlowParameter()`

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -166,6 +175,52 @@ public static ExecutionOptions createFromObject(final Object obj) {
...
+      for (String s : statuses) {
+        if (!allowedStatuses.contains(s)) {
+          errMsg.add(String.format("`%s` is not a valid restartable status, "
+              + "permitted status are %s", s, allowedStatuses));
+        }
+      }
+    }
+    if (flowParameters.containsKey(FlowParameters.FLOW_PARAM_RESTART_COUNT)){
+      // check restart count limit
+      final int azFlowRestartCountLimit = azProps.getInt(
```
utk-12, line 206: I think adding the prefix az to this variable name is redundant. Please consider removing this.

tsangz2013: I am good either way, removing

ypadron-in:
> 
    allow.execution.restart.on.status that describe a set of terminated status when an execution can restart
    "execution.restart.count" == how many times can an execution restarts
    "execution.restart.strategy" == the strategy (e.g. "restart_from _root", "disable_finished_nodes")

Job execution retries defines a property `retry.backoff` to specify a delay between each retry. Should we support it for flow execution retries too? 
Also, what to do you think about reusing the naming convention used in job retries for these new properties since these are similar features.
Existing properties for job retries:
```
retries -> The number of retries that will be automatically attempted for failed jobs
retry.backoff -> The millisec time between each retry attempt
``` 
One alternative for flow level retries following similar naming pattern could be:
```
flow.max.retries
flow.retry.status
flow.retry.strategy
flow.retry.backoff
```

ypadron-in:
If I recall correctly there is currently some support for flow level restarts where users can specify a property to restart their flow executions on EXECUTION_STOPPED, DISPATCHING and maybe other status. It feels like there is an overall between that feature and this one. Could you please comment of usage of both features? Are we planning to merge them?

ypadron-in:
Why do we need a config property (azkaban.execution.restartable.status) for the list of restartable status at flow level? Do we foresee that list changing often?
Thank you for setting a limit on max number of retries!

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -43,6 +49,9 @@ public class ExecutionOptions {
   /* override dispatcher selection and use executor id specified */
   public static final String USE_EXECUTOR = "useExecutor";
   public static final int DEFAULT_FLOW_PRIORITY = 5;
+  public static final ImmutableList<String> DEFAULT_EXECUTION_RESTARTABLE_STATUS =
+      ImmutableList.of(Status.EXECUTION_STOPPED.name(), Status.FAILED.name());
+  public static final int DEFAULT_EXECUTION_RESTART_LIMIT = 2;
```
ypadron-in, line 54: What's the difference between this list and the one in Status.java class:
```
public static final ImmutableSet<Status> RESTARTABLE_STATUSES =
      ImmutableSet.of(Status.READY, Status.DISPATCHING, Status.PREPARING, Status.EXECUTION_STOPPED);
```
Is it possible to consolidate/unify the user experience when it comes to flow restart-ability?

tsangz2013: actually I think there's some ambiguity, let me re-organize them.

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
```diff
@@ -166,6 +172,52 @@ public static ExecutionOptions createFromObject(final Object obj) {
...
+              + "permitted status are %s", s, RESTARTABLE_TERMINATED_STATUSES));
+        }
+      }
+    }
+    if (flowParameters.containsKey(FlowParameters.FLOW_PARAM_RESTART_COUNT)){
+      // check restart count limit
+      final int flowRestartCountLimit = azProps.getInt(
+          AZKABAN_EXECUTION_RESTART_LIMIT, DEFAULT_FLOW_RESTART_LIMIT);
+      final int flowRestartCount = Integer.parseInt(
+          flowParameters.getOrDefault(FlowParameters.FLOW_PARAM_RESTART_COUNT, "0"));
```
ypadron-in, line 206: I think parseInt can throw NumberFormatException. When/how are capturing that exception? What will the user see?

tsangz2013: It's in the screenshot, will see at flow submit time

utk-12: I think @ypadron-in meant if they pass a value which is not a number, like ,'x' or '1x' or something similar. 
parstInt will throw an exception then will it not ? The above will cover cases if a wrong number is passed.

utk-12 (changes requested):

azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
tsangz2013: oh I see. let me add this

utk-12: thanks for adding it in a try block.

utk-12 (approved):
thanks for incorporating all the suggestions !

PR #3264, opened 2023-04-03
Add ExecutionOption.executionMaxRetries to control flow retry

tsangz2013:
**Why we need this**
Previously Azkaban use only `isExecutionRetried` boolean flag to record and control flow/execution retry info, we now want to allow retrying for a given number of times.

**What's done**
1. `isExecutionRetried` only means whether an execution is retried
2. At flow submit time, convert user input flow-parameter `flow.max.retries` to `ExecutionOption.executionMaxRetries` member variable
3. use `executionMaxRetries` to count down how many times can this execution retry

**Tests done**
1. unit tests
2. deploy the change to test cluster, launched an execution with `flow.max.retries=2` and trigger EXECUTION_STOPPED by killing the k8s pod:

The old execution got into EXECUTION_STOPPED and an new execution is being restarted:
Screenshot 2023-04-03 at 9 39 09 PM

The old execution with "flow.max.retries=2" while the retried new execution has "flow.max.retries=1"
Screenshot 2023-04-03 at 9 30 02 PM
Screenshot 2023-04-03 at 9 30 18 PM

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -207,6 +207,7 @@ public class Constants {
     public static final String FLOW_PAUSE_DURATION = "flowPauseDuration";
     public static final String FLOW_PREPARATION_DURATION = "flowPreparationDuration";
     public static final String EXECUTION_RETRIED_BY_AZKABAN = "executionRetriedByAzkaban";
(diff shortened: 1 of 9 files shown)


tsangz2013:
abandon as this approach cannot work

PR #3265, opened 2023-04-04
Use flowParameters.executionMaxRetries to control flow retry

tsangz2013:
**Why we need this**
Previously Azkaban use only `isExecutionRetried` boolean flag to record and control flow/execution retry info, we now want to allow retrying for a number of times defined by the user.

**What's done**
1. `isExecutionRetried` only means whether an execution is retried
2. At flow submit time, if the `flow.retry.statuses` is set, then `flow.max.retries` fallback value is `1` if not set
3. use `flow.max.retries` to determine how many times can this execution retry at `restartFlow` method

**Tests done**
1. unit tests
2. deploy the change to test cluster, launched an execution with `flow.max.retries=2` and trigger EXECUTION_STOPPED by killing the k8s pod:

The old execution got into EXECUTION_STOPPED and an new execution is being restarted:
Screenshot 2023-04-03 at 9 39 09 PM

The old execution with "flow.max.retries=2" while the retried new execution has "flow.max.retries=1"
Screenshot 2023-04-03 at 9 30 02 PM
Screenshot 2023-04-03 at 9 30 18 PM

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -883,7 +883,7 @@ public class Constants {
     public static final String FLOW_PARAM_ALLOW_RESTART_ON_STATUS = "flow.retry.statuses";
 
     // Constant to define at most how many times can restart the flow
-    public static final String FLOW_PARAM_RESTART_COUNT = "flow.max.retries";
+    public static final String FLOW_PARAM_MAX_RETRIES = "flow.max.retries";
 
     // Constant to define the strategy to restart the execution, default to "restart_from_root",
     public static final String FLOW_PARAM_RESTART_STRATEGY = "flow.retry.strategy";
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -168,22 +168,18 @@ public class ExecutionControllerUtils {
   }
 
   /**
-   * This method tries to restart the flow for certain statuses otherwise simply return. There are
-   * three scenarios that this method is called: 1. a flow is cleaned up by ContainerCleanupManager;
-   * 2. a flow has dispatch failure; 3. a flow encounters pod failure Each flow execution can be
-   * retried once. If the original status is EXECUTION_STOPPED, then it will be retried only if
-   * allow.restart.on.execution.stopped is set to true
+   * This method tries to determine whether the flow should be restarted for certain statuses
+   * otherwise simply return.
+   * There are three scenarios that this method is called:
+   * 1. a flow is cleaned up by ContainerCleanupManager;
+   * 2. a flow has dispatch failure;
+   * 3. a flow encounters pod failure Each flow execution can be retried once.
+   * If the original status is EXECUTION_STOPPED or FAILED, then it will be retried only if
+   * "flow.retry.statuses" is defined with the status.
    *
    * @param flow
    * @param originalStatus
    */
-  public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
-    ExecutableFlow flowToRestart = getFlowToRestart(flow, originalStatus);
-    if (flowToRestart != null){
-      ExecutionControllerUtils.submitRestartFlow(flowToRestart);
-    }
-  }
-
   static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
       final Status originalStatus){
     final ExecutionOptions options = flow.getExecutionOptions();
@@ -241,6 +237,13 @@ public class ExecutionControllerUtils {
       return flow;
     }
     return null;
+  }
+
+  public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
+    ExecutableFlow flowToRestart = getFlowToRestart(flow, originalStatus);
+    if (flowToRestart != null){
+      ExecutionControllerUtils.submitRestartFlow(flowToRestart);
+    }
   }
 
   private static void submitRestartFlow(final ExecutableFlow flow) {
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -178,16 +178,40 @@ public class ExecutionControllerUtils {
    * @param originalStatus
    */
   public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
-    final ExecutionOptions options = flow.getExecutionOptions();
-    // flow can only be retried once
-    if (options == null || options.isExecutionRetried()) {
-      return;
+    ExecutableFlow flowToRestart = getFlowToRestart(flow, originalStatus);
+    if (flowToRestart != null){
+      ExecutionControllerUtils.submitRestartFlow(flowToRestart);
     }
+  }
+
+  static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
+      final Status originalStatus){
+    final ExecutionOptions options = flow.getExecutionOptions();
+    if (options == null) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
+    }
+    final Map<String, String> flowParams = options.getFlowParameters();
+    if (flowParams == null || flowParams.isEmpty()) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
+    }
+
+    // flow can only retry if not reach the limit
+    final int flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    if (flowMaxRetryLimit <= 0) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has exceed retry limit, count = "
+          + flowMaxRetryLimit);
+      return null;
+    }
+    // reduce the count
+    flowParams.put(FlowParameters.FLOW_PARAM_MAX_RETRIES, String.valueOf(flowMaxRetryLimit - 1));
+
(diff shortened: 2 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -178,16 +178,40 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
+    final int flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    if (flowMaxRetryLimit <= 0) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has exceed retry limit, count = "
+          + flowMaxRetryLimit);
+      return null;
+    }
+    // reduce the count
+    flowParams.put(FlowParameters.FLOW_PARAM_MAX_RETRIES, String.valueOf(flowMaxRetryLimit - 1));
+
```
tsangz2013, line 210: main change here

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -883,7 +883,7 @@ public static class FlowParameters {
     public static final String FLOW_PARAM_ALLOW_RESTART_ON_STATUS = "flow.retry.statuses";
 
     // Constant to define at most how many times can restart the flow
-    public static final String FLOW_PARAM_RESTART_COUNT = "flow.max.retries";
+    public static final String FLOW_PARAM_MAX_RETRIES = "flow.max.retries";
```
tsangz2013, line 886: this config is introduced for the flow-resumability project and is of no use before this PR.

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -214,24 +214,31 @@ public static void validatePreprocessFlowParameters(ExecutionOptions options, Pr
...
         errMsg.add(e.getMessage());
       }
     }
+    // doesn't contain FlowParameters.FLOW_PARAM_MAX_RETRIES
+    else if (!options.isExecutionRetried()
+        && flowParameters.containsKey(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS)) {
+      // if the MAX_RETRIES parameter is empty but set some retry_statuses, default to
+      // give 1 restart
+      flowParameters.put(FlowParameters.FLOW_PARAM_MAX_RETRIES, "1");
+    }
```
tsangz2013, line 241: if `flow.retry.statuses` is set while `flow.max.retries` not set, then it should fallback to `1` time

utk-12: thanks for adding this fallback mechanism for isExecutionRetried().

utk-12:
LGTM ! Thanks for incorporating the reordering of the functions so its easied to read.

utk-12 (approved):

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
    *
    * @param flow
    * @param originalStatus
    */
-  public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
+  static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
+      final Status originalStatus){
     final ExecutionOptions options = flow.getExecutionOptions();
-    // flow can only be retried once
-    if (options == null || options.isExecutionRetried()) {
```
ypadron-in, line 183: This `if` checks if a flow has already been retried for non terminal status scenarios which is different from checking number of retries for terminal status scenarios. We are introducing a check for number of retries for terminal status in this PR but deleting this check, which can prevent the "non terminal status" retries. Why are we deleting this check?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
+    // flow can only retry if not reach the limit
+    final int flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    if (flowMaxRetryLimit <= 0) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has exceed retry limit, count = "
+          + flowMaxRetryLimit);
+      return null;
+    }
+    // reduce the count
+    flowParams.put(FlowParameters.FLOW_PARAM_MAX_RETRIES, String.valueOf(flowMaxRetryLimit - 1));
```
ypadron-in, line 205: At this point we still don't know if there will be a retry or not (for example the method could return in line 219 without retrying). Why are we decreasing the counter already?

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
ypadron-in, line 241: It seems to me that Azkaban handles 2 types of retries: retries on non terminal status and retries on terminal status which is based on user configuration. If that's true then we should keep 2 separate counters. Why are we trying to use just 1 counter for both types of retries? Also, we should NOT modify user input.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
-   * retried once. If the original status is EXECUTION_STOPPED, then it will be retried only if
-   * allow.restart.on.execution.stopped is set to true
+   * This method tries to determine whether the flow should be restarted for certain statuses
+   * otherwise simply return.
+   * There are three scenarios that this method is called:
+   * 1. a flow is cleaned up by ContainerCleanupManager;
+   * 2. a flow has dispatch failure;
+   * 3. a flow encounters pod failure Each flow execution can be retried once.
+   * If the original status is EXECUTION_STOPPED or FAILED, then it will be retried only if
+   * "flow.retry.statuses" is defined with the status.
```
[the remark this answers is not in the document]
tsangz2013: thanks for raising this up, the comment needs to be updated.
For the case a flow failed (maybe due to user logic bug) and pod complete, we will still allow retry if set "flow.retry.status=FAILED"

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
-      return;
+    if (options == null) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
+    }
+    final Map<String, String> flowParams = options.getFlowParameters();
+    if (flowParams == null || flowParams.isEmpty()) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
     }
```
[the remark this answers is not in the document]
tsangz2013: good point, will remove this

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
-    if (options == null || options.isExecutionRetried()) {
-      return;
+    if (options == null) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
+    }
+    final Map<String, String> flowParams = options.getFlowParameters();
+    if (flowParams == null || flowParams.isEmpty()) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
```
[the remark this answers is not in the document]
tsangz2013: agree, will fix at once

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -168,26 +168,46 @@ public static void finalizeFlow(final EventHandler eventHandler,
...
+    final int flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    if (flowMaxRetryLimit <= 0) {
+      logger.warn("ExecutableFlow: " + flow.getExecutionId() + " has exceed retry limit, count = "
+          + flowMaxRetryLimit);
+      return null;
+    }
+    // reduce the count
+    flowParams.put(FlowParameters.FLOW_PARAM_MAX_RETRIES, String.valueOf(flowMaxRetryLimit - 1));
+
```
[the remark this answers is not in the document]
tsangz2013: We are actually modifying the flow-parameter values in the "preprocess" code, and I think adding a new attribute might be a little too complicated.
I agree we may need to keep the info how many times has been tried.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -215,12 +231,18 @@ public static void restartFlow(final ExecutableFlow flow, final Status originalS
...
-      ExecutionControllerUtils.submitRestartFlow(flow);
+      return flow;
+    }
+    return null;
+  }
+
+  public static void restartFlow(final ExecutableFlow flow, final Status originalStatus) {
+    ExecutableFlow flowToRestart = getFlowToRestart(flow, originalStatus);
+    if (flowToRestart != null){
+      ExecutionControllerUtils.submitRestartFlow(flowToRestart);
```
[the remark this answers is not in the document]
tsangz2013: the original flag `executionRetried` is set to true no matter which kind of originalStatus, in the "submitRestartFlow" method, which might not be ideal, we should discuss it.

PR #3266, opened 2023-04-05
Extend pod complete event handling to also check & retry execution in finalized status including FAILED

tsangz2013:
**Why we need this**
Previously when the pod-complete event comes, only if the execution status in Azkaban DB is Non-Finalized, then the `restartFlow()` logic got triggered.
Now to support the flow-retry on `FAILED`, we need to trigger the logic when execution status is Finalized (FAILED is a finalized status).

**What's done**
Since the logic to check for the permitted retry-statuses is already implemented in `restartFlow()` method #3261, now we just need to enter this logic when the execution status is finalized.

**Tests done**
Tested in the staging cluster:

Started the original execution 786255 with `flow.retry.statuses=FAILED,EXECUTION_STOPPED` and `flow.max.retries=1` and failed the flow by killing the hadoop application, a new execution 786256 being retried:
Screenshot 2023-04-04 at 4 33 41 PM

The webserver log for retrying 786255:
Screenshot 2023-04-04 at 4 34 40 PM

Failed the restarted execution 786256 and there's no further retry, since the `flow.max.retries` has reached `0`
Screenshot 2023-04-04 at 7 19 27 PM

The webserver log rejecting the retry of 786256:
Screenshot 2023-04-04 at 4 41 57 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
@@ -233,7 +233,7 @@ public class ExecutionControllerUtils {
 
     if (restartedStatuses.contains(originalStatus.name())) {
       logger.info("Submitted flow for restart: " + flow.getExecutionId()
-          + "from originalStatus: " + originalStatus);
+          + " from originalStatus: " + originalStatus);
       return flow;
     }
     return null;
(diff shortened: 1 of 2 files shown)


svnarumugam (approved):
It looks good to me. Thanks for adding a detailed PR description. :)

PR #3268, opened 2023-04-07
Use ExecutableFlow member variables to track and control flow-level retry

tsangz2013:
**Why we need this**
Part of the flow-retry feature, and it's improving the implementation: 
* Remove modifying the user-input `ExecutionOptions`, keep them as is
* clearer model to count & control how many times can an execution retry, respectively for non- / terminal originalStatus

**What's done**
* remove `isExecutionRetried` from `ExecutionOptions` because it doesn't distinguish type of retry, and retry count should be execution level info.
* use `systemRetriedTimes` to count retries on `dispatch/preparing/ready` status executions (by default should retry by the system)
* use `customRetriedTimes` to count retries on finalized status `failed/execution_stopped` defined by users

**Tests done**
* updated unit tests 
* tested in a cluster: 

Launched the original execution 786575:
Screenshot 2023-04-07 at 5 36 44 PM

Delete the k8s pod to trigger EXECUTION_STOPPED, and a new execution got retried:
Screenshot 2023-04-07 at 5 39 50 PM
Screenshot 2023-04-07 at 5 39 07 PM

The new execution has `customRetriedTimes=1`
Screenshot 2023-04-07 at 5 36 44 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -435,8 +435,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
       }
       this.slaOptionStr = slaBuilder.toString();
     }
-    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_RETRY_COUNT_PARAM);
-    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_RETRY_COUNT_PARAM);
+    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_RETRY_COUNT_PARAM, 0);
+    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_RETRY_COUNT_PARAM, 0);
 
     if (flowObj.containsKey(VERSIONSET_JSON_PARAM) && flowObj.containsKey(VERSIONSET_MD5HEX_PARAM) && flowObj.containsKey(VERSIONSET_ID_PARAM)) {
       // Checks if flow contains version set information
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -67,8 +67,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
   public static final String VERSIONSET_ID_PARAM = "versionSetId";
   private static final String PARAM_OVERRIDE = "param.override.";
   private static final String PROJECT_FILE_UPLOAD_USER = "uploadUser";
-  private static final String CUSTOM_RETRIED_TIMES_PARAM = "customRetriedTimes";
-  private static final String SYSTEM_RETRIED_TIMES_PARAM = "systemRetriedTimes";
+  private static final String USER_DEFINED_RETRY_COUNT_PARAM = "userDefinedRetryCount";
+  private static final String SYSTEM_DEFINED_RETRY_COUNT_PARAM = "systemDefinedRetryCount";
 
   private final HashSet<String> proxyUsers = new HashSet<>();
   private int executionId = -1;
@@ -94,10 +94,10 @@ public class ExecutableFlow extends ExecutableFlowBase {
   private String failedJobId = "unknown";
   private String modifiedBy = "unknown";
   private DispatchMethod dispatchMethod;
-  // how many times has flow level retry happened - retry defined by user's final status
-  private int customRetriedTimes = 0;
-  // retry due to stuck on "Dispatch/Preparing/Ready" status
-  private int systemRetriedTimes = 0;
+  // how many times flow level retry happened over user defined final status
+  private int userDefinedRetryCount = 0;
+  // how many times flow level retry happened due to stuck in "Dispatch/Preparing/Ready" status
+  private int systemDefinedRetryCount = 0;
 
   // For slaOption information
   private String slaOptionStr = "null";
@@ -315,20 +315,20 @@ public class ExecutableFlow extends ExecutableFlowBase {
     return slaOptionStr;
   }
 
-  public int getCustomRetriedTimes() {
-    return customRetriedTimes;
+  public int getUserDefinedRetryCount() {
+    return userDefinedRetryCount;
   }
 
-  public void setCustomRetriedTimes(int customRetriedTimes) {
-    this.customRetriedTimes = customRetriedTimes;
+  public void setUserDefinedRetryCount(int userDefinedRetryCount) {
+    this.userDefinedRetryCount = userDefinedRetryCount;
   }
 
-  public int getSystemRetriedTimes() {
-    return systemRetriedTimes;
+  public int getSystemDefinedRetryCount() {
+    return systemDefinedRetryCount;
   }
 
-  public void setSystemRetriedTimes(int systemRetriedTimes) {
-    this.systemRetriedTimes = systemRetriedTimes;
+  public void setSystemDefinedRetryCount(int systemDefinedRetryCount) {
+    this.systemDefinedRetryCount = systemDefinedRetryCount;
   }
 
   @Override
@@ -369,8 +369,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
     }
 
     flowObj.put(SLAOPTIONS_PARAM, slaOptions);
-    flowObj.put(CUSTOM_RETRIED_TIMES_PARAM, this.customRetriedTimes);
-    flowObj.put(SYSTEM_RETRIED_TIMES_PARAM, this.systemRetriedTimes);
+    flowObj.put(USER_DEFINED_RETRY_COUNT_PARAM, this.userDefinedRetryCount);
+    flowObj.put(SYSTEM_DEFINED_RETRY_COUNT_PARAM, this.systemDefinedRetryCount);
 
     flowObj.put(IS_LOCKED_PARAM, this.isLocked);
     flowObj.put(IS_OOM_Killed_PARAM, this.isOOMKilled);
@@ -435,8 +435,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
       }
       this.slaOptionStr = slaBuilder.toString();
     }
-    this.customRetriedTimes = flowObj.getInt(CUSTOM_RETRIED_TIMES_PARAM);
-    this.systemRetriedTimes = flowObj.getInt(SYSTEM_RETRIED_TIMES_PARAM);
+    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_RETRY_COUNT_PARAM);
+    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_RETRY_COUNT_PARAM);
 
     if (flowObj.containsKey(VERSIONSET_JSON_PARAM) && flowObj.containsKey(VERSIONSET_MD5HEX_PARAM) && flowObj.containsKey(VERSIONSET_ID_PARAM)) {
       // Checks if flow contains version set information
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -67,8 +67,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
   public static final String VERSIONSET_ID_PARAM = "versionSetId";
   private static final String PARAM_OVERRIDE = "param.override.";
   private static final String PROJECT_FILE_UPLOAD_USER = "uploadUser";
-  private static final String USER_DEFINED_RETRY_COUNT_PARAM = "userDefinedRetryCount";
-  private static final String SYSTEM_DEFINED_RETRY_COUNT_PARAM = "systemDefinedRetryCount";
+  private static final String USER_DEFINED_FLOW_RETRY_COUNT_PARAM = "userDefinedFlowRetryCount";
+  private static final String SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM = "systemDefinedFlowRetryCount";
 
   private final HashSet<String> proxyUsers = new HashSet<>();
   private int executionId = -1;
@@ -369,8 +369,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
     }
 
     flowObj.put(SLAOPTIONS_PARAM, slaOptions);
-    flowObj.put(USER_DEFINED_RETRY_COUNT_PARAM, this.userDefinedRetryCount);
-    flowObj.put(SYSTEM_DEFINED_RETRY_COUNT_PARAM, this.systemDefinedRetryCount);
+    flowObj.put(USER_DEFINED_FLOW_RETRY_COUNT_PARAM, this.userDefinedRetryCount);
+    flowObj.put(SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM, this.systemDefinedRetryCount);
 
     flowObj.put(IS_LOCKED_PARAM, this.isLocked);
     flowObj.put(IS_OOM_Killed_PARAM, this.isOOMKilled);
@@ -435,8 +435,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
       }
       this.slaOptionStr = slaBuilder.toString();
     }
-    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_RETRY_COUNT_PARAM, 0);
-    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_RETRY_COUNT_PARAM, 0);
+    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_FLOW_RETRY_COUNT_PARAM, 0);
+    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM, 0);
 
     if (flowObj.containsKey(VERSIONSET_JSON_PARAM) && flowObj.containsKey(VERSIONSET_MD5HEX_PARAM) && flowObj.containsKey(VERSIONSET_ID_PARAM)) {
       // Checks if flow contains version set information
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -39,6 +39,9 @@ import java.util.stream.Collectors;
 
 public class ExecutableFlow extends ExecutableFlowBase {
 
+  public static final int DEFAULT_FLOW_RETRY_LIMIT = 2;
+  public static final int DEFAULT_SYSTEM_FLOW_RETRY_LIMIT = 1;
+
   public static final String EXECUTIONID_PARAM = "executionId";
   public static final String EXECUTIONPATH_PARAM = "executionPath";
   public static final String EXECUTIONOPTIONS_PARAM = "executionOptions";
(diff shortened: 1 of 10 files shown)


azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -67,7 +67,11 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
     }
     options.setMailCreator(flow.getMailCreator());
     // Update the flow options so that the flow will be not retried again by Azkaban
-    options.setExecutionRetried(true);
+
+    // inherent the retry time counters
+    executableFlow.setCustomRetriedTimes(exFlow.getCustomRetriedTimes());
+    executableFlow.setSystemRetriedTimes(exFlow.getSystemRetriedTimes());
```
tsangz2013, line 73: Noted that the restarted execution's `*RetriedTimes` is updated here

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -184,38 +186,42 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
...
     // If the original execution status is restartable non-terminal status, it can be retried
     if (RESTARTABLE_NON_TERMINAL_STATUSES.contains(originalStatus)) {
-      logger.info("Submitted flow for restart: " + flow.getExecutionId());
+      if (flow.getSystemRetriedTimes() >= DEFAULT_SYSTEM_FLOW_RETRY_LIMIT){
+        logger.info("ExecutableFlow: " + flow.getExecutionId() + " has reached max retry limit "
+            + "for non-terminal status, ");
+        return null;
+      }
+      logger.info("Submitting flow for restart: " + flow.getExecutionId());
+      flow.setSystemRetriedTimes(flow.getSystemRetriedTimes() + 1);
```
tsangz2013, line 200: 1. use `SystemRetriedTimes` to control retry on `NON_TERMINAL_STATUSES`
2. update `SystemRetriedTimes++` if permitted

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -184,38 +186,42 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
...
+      logger.debug("ExecutableFlow: " + flow.getExecutionId() + " has ExecutionOptions == null");
+      return null;
+    }
+
+    // flow can only retry if not reach the limit
+    final int flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    if (flow.getCustomRetriedTimes() >= flowMaxRetryLimit) {
+      logger.info("ExecutableFlow: " + flow.getExecutionId() + " has reached max retry limit, "
+          + "retried=" + flow.getCustomRetriedTimes() + ", limit= " + flowMaxRetryLimit);
```
tsangz2013, line 224: use `CustomRetriedTimes` to control user-defined retries on `TERMINAL_STATUSES`

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -232,8 +238,9 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
     }
 
     if (restartedStatuses.contains(originalStatus.name())) {
-      logger.info("Submitted flow for restart: " + flow.getExecutionId()
-          + " from originalStatus: " + originalStatus);
+      logger.info("Submitting flow for restart: " + flow.getExecutionId()
+          + "from originalStatus: " + originalStatus);
+      flow.setCustomRetriedTimes(flow.getCustomRetriedTimes() + 1);
```
tsangz2013, line 243: `CustomRetriedTimes++` when permitted

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -64,6 +67,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
   public static final String VERSIONSET_ID_PARAM = "versionSetId";
   private static final String PARAM_OVERRIDE = "param.override.";
   private static final String PROJECT_FILE_UPLOAD_USER = "uploadUser";
+  private static final String CUSTOM_RETRIED_TIMES_PARAM = "customRetriedTimes";
```
svnarumugam, line 70: It was a little challenging for me to relate to the name `customRetriedTimes`. May I ask you to consider names like userDefinedRetryCount, systemDefinedRetryCount  something like that? we can take suggestions from another pair of eyes too.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -89,6 +94,10 @@ public class ExecutableFlow extends ExecutableFlowBase {
   private String failedJobId = "unknown";
   private String modifiedBy = "unknown";
   private DispatchMethod dispatchMethod;
+  // how many times has flow level retry happened - retry defined by user's final status
+  private int customRetriedTimes = 0;
```
svnarumugam, line 98: should we consider this variable as `AtomicInt` or it could be an overkill here

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -39,6 +39,9 @@
 
 public class ExecutableFlow extends ExecutableFlowBase {
 
+  public static final int DEFAULT_FLOW_RETRY_LIMIT = 2;
```
svnarumugam, line 42: Clarification: I was trying to see if we are recording these retry values in the event reporter. Are we recording this to be able to query and get numbers like "how many times we end up retrying due to system? how many times user specified retries?" etc

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
tsangz2013: I think it's fine as no racing here

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
tsangz2013: good suggestion, let me update the naming

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -408,6 +435,8 @@ public void fillExecutableFromMapObject(
       }
       this.slaOptionStr = slaBuilder.toString();
     }
+    this.userDefinedRetryCount = flowObj.getInt(USER_DEFINED_RETRY_COUNT_PARAM);
+    this.systemDefinedRetryCount = flowObj.getInt(SYSTEM_DEFINED_RETRY_COUNT_PARAM);
```
ypadron-in, line 439: It looks like the default value for integer is -1. Could that lead to allowing an extra retry? Should we set default value to zero?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -231,9 +229,23 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
       restartedStatuses.add(Status.EXECUTION_STOPPED.name());
     }
 
+    // flow can only retry if custom-retry-times not reach the limit
+    // and for the case where the retry-times not defined, default to 1
+    int flowMaxRetryLimit = 1;
+    if (flowParams.containsKey(FlowParameters.FLOW_PARAM_MAX_RETRIES)){
+      flowMaxRetryLimit = Integer.parseInt(
+        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
+    }
```
ypadron-in, line 238: Line 237 defines default value to zero when the property isn't defined. Why line 234 defines it to 1? If the user didn't define number of retries should we still retry once? Why?

ypadron-in, line 238: Is the if statement needed?
I think below statement is enough. WDYT?
```
final int flowMaxRetryLimit = Integer.parseInt(
        flowParams.getOrDefault(FlowParameters.FLOW_PARAM_MAX_RETRIES, "0"));
```

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -67,7 +67,11 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
     // Update the flow options so that the flow will be not retried again by Azkaban
-    options.setExecutionRetried(true);
+
+    // inherent the retry time counters
+    executableFlow.setUserDefinedRetryCount(exFlow.getUserDefinedRetryCount());
+    executableFlow.setSystemDefinedRetryCount(exFlow.getSystemDefinedRetryCount());
+
     // If a retried flow A gets retried again with a new execution id flow B, the original flow
     // execution id of flow B should be the same as flow A's original flow execution id.
     if (options.getOriginalFlowExecutionIdBeforeRetry() == null) {
```
ypadron-in, line 77: Wondering if we should track original execution id as an attribute of ExecutableFlow, just like counters. It would help keep related things together I think. What are the benefits of adding original exec id to ExecutionOptions?

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -233,8 +233,7 @@ public static void validatePreprocessFlowParameters(ExecutionOptions options, Pr
       }
     }
     // doesn't contain FlowParameters.FLOW_PARAM_MAX_RETRIES
-    else if (!options.isExecutionRetried()
-        && flowParameters.containsKey(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS)) {
+    else if (flowParameters.containsKey(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS)) {
```
ypadron-in, line 236: If the user didn't set `FLOW_PARAM_MAX_RETRIES` why should Azkaban set it on behalf of the user?

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
tsangz2013: good catch, will update

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
tsangz2013: When user set "flow.retry.statuses=FAILED" but forgot to set the "flow.max.retry", then should have 1 chance, right?

I think the question is:
When user set "flow.retry.statuses=FAILED" and set the "flow.max.retry=BAD_NOT_INT_THING", should they have 1 chance?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: Hi yes it's in the plan, should be in ExecutableFlow, and will have a follow up PR to address this.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -231,9 +229,23 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
...
+      logger.info("ExecutableFlow: " + flow.getExecutionId() + " has reached max retry limit, "
+          + "retried=" + flow.getUserDefinedRetryCount() + ", limit= " + flowMaxRetryLimit);
+      return null;
+    }
+
     if (restartedStatuses.contains(originalStatus.name())) {
-      logger.info("Submitted flow for restart: " + flow.getExecutionId()
+      logger.info("Submitting flow for restart: " + flow.getExecutionId()
           + " from originalStatus: " + originalStatus);
+      flow.setUserDefinedRetryCount(flow.getUserDefinedRetryCount() + 1);
```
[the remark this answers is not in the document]
tsangz2013: Good question, my answer is we are not sure how long it takes, could be 20ms or 1min.
If WS shutdown, then this retry is dropped, and unfortunately it's been like this for a while.

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
```diff
@@ -1740,8 +1740,6 @@ synchronized Map<String, String> getFlowMetadata(final FlowRunner flowRunner,
       metaData.put(EventReporterConstants.EXECUTION_ID, String.valueOf(rootFlow.getExecutionId()));
       metaData.put(EventReporterConstants.START_TIME, String.valueOf(flow.getStartTime()));
       metaData.put(EventReporterConstants.SUBMIT_TIME, String.valueOf(rootFlow.getSubmitTime()));
-      metaData.put(EventReporterConstants.EXECUTION_RETRIED_BY_AZKABAN,
-          String.valueOf(rootFlow.getExecutionOptions().isExecutionRetried()));
```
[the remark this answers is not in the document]
tsangz2013: the whole `isExecutionRetried ` field is removed.
The change will be capture in EventReporter, as part of the ExecutableFlow object

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
tsangz2013: I think it's a safe net, if ppl set `flow.retry.statuses` then they surely want to retry, but if they forget the `flow.max.retries` maybe should default to 1.

Also it's for backward compatibility, when ppl set `allow.restart.on.execution.stopped`, we default to have 1 retry, so the behavior should continue

ypadron-in: I see, thanks for clarifying! If maintaining backward compatibility then I think we should cover it inside the if of lines: 184-194. What do you think?

azkaban-exec-server/src/main/java/azkaban/execapp/FlowRunner.java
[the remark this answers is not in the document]
ypadron-in: Did quite follow sorry. Are you saying things will be added back in the next PR?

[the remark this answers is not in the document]
tsangz2013: The field `isExecutionRetried` is removed from `ExecutionOptions` thus removing here.

I will add the event handing for `ExecutableFlow` object in the next PR, where we will include the retry-count, the parent-execution (if retry)

azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
tsangz2013: Actually you are right, should not alter user input ExecutionOptions. Removing.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -64,6 +67,8 @@ public class ExecutableFlow extends ExecutableFlowBase {
   public static final String VERSIONSET_ID_PARAM = "versionSetId";
   private static final String PARAM_OVERRIDE = "param.override.";
   private static final String PROJECT_FILE_UPLOAD_USER = "uploadUser";
+  private static final String USER_DEFINED_RETRY_COUNT_PARAM = "userDefinedRetryCount";
+  private static final String SYSTEM_DEFINED_RETRY_COUNT_PARAM = "systemDefinedRetryCount";
```
ypadron-in, line 71: nit for subsequent PR: consider renaming to 
USER_DEFINED_**FLOW**_RETRY_COUNT_PARAM 
and
SYSTEM_DEFINED_**FLOW**_RETRY_COUNT_PARAM

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -231,9 +229,30 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
       restartedStatuses.add(Status.EXECUTION_STOPPED.name());
     }
 
+    // flow can only retry if custom-retry-times not reach the limit
+    // and for the case where the retry-times not defined, default to 1
+    int flowMaxRetryLimit = 0;
+    if (flowParams.containsKey(FlowParameters.FLOW_PARAM_MAX_RETRIES)){
+      flowMaxRetryLimit = Integer.parseInt(flowParams.get(FlowParameters.FLOW_PARAM_MAX_RETRIES));
+    }
+    else if (!restartedStatuses.isEmpty()){
```
ypadron-in, line 238: There may be small issue here. If there is no FLOW_PARAM_ALLOW_RESTART_ON_STATUS defined, would line  
221 return a set with one element which is the empty string? If yes, then I looks like we will set FLOW_PARAM_MAX_RETRIES to 1 when the user didn't configure flow retries which is probably not desired.

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
```diff
@@ -231,9 +229,30 @@ static ExecutableFlow getFlowToRestart(final ExecutableFlow flow,
...
+      logger.info(String.format("ExecutableFlow: %s has `%s=%s but `%s` not set, "
+              + "default to retry once",
+          FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS,
+          FlowParameters.FLOW_PARAM_MAX_RETRIES,
+          flow.getExecutionId(), restartedStatuses));
+      flowMaxRetryLimit = 1;
```
ypadron-in, line 244: Could you please check the order of the parameters passed to the format method? Shouldn't execution id be the first one?

azkaban-common/src/main/java/azkaban/executor/ExecutionControllerUtils.java
ypadron-in, line 238: Would line 221 return a set with one element which is an empty String when the user didn't configure FLOW_PARAM_ALLOW_RESTART_ON_STATUS?

tsangz2013: good point, let me add filter

ypadron-in:
To consider in a subsequent PR. 
Since we're modifying user input to maintain backward compatibility (see below code block), can we also set FLOW_PARAM_MAX_RETRIES to 1 if the user didn't do it and add a log to record what was done?
```
if (Boolean.parseBoolean(
        flowParameters.getOrDefault(
            FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_EXECUTION_STOPPED, "false"))
    ){
      flowParameters.put(
          FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS,
          Status.EXECUTION_STOPPED.name() + "," +
              flowParameters.getOrDefault(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_STATUS, "")
      );
      flowParameters.remove(FlowParameters.FLOW_PARAM_ALLOW_RESTART_ON_EXECUTION_STOPPED);
}
```

Also, as an alternative to automatically attempting 1 retry if the user didn't define the property `FLOW_PARAM_MAX_RETRIES` which can lead to confusion, can we make it a required property if `FLOW_PARAM_ALLOW_RESTART_ON_STATUS` is defined. For example: if `FLOW_PARAM_ALLOW_RESTART_ON_STATUS` is defined but not `FLOW_PARAM_MAX_RETRIES` we should return 404 - Bad request and a message like "FLOW_PARAM_MAX_RETRIES property is required when defining FLOW_PARAM_ALLOW_RESTART_ON_STATUS". WDYT?

ypadron-in (approved):
Please consider the suggestion above in a subsequent PR.

tsangz2013:
> Please consider the suggestion above in a subsequent PR.

Good advice! let me note down this task

PR #3269, opened 2023-04-11
Load ExecutionOption from the stored project config to the ExecutableFlow at creation, to honor input from DSL

tsangz2013:


azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -515,15 +512,10 @@ public void setVersionSet(final VersionSet versionSet) {
...
-   * loading approach. If the properties are not set, then they are first set.
-   * @return Returns the flattened flow props overridden by flow params.
+   * Extract the "param.override." properties from input Props and set to the
+   * flowParameters of executionOptions
    */
-  public void setFlowPropsAndParams(final ProjectLoader projectLoader) {
-    Props props = FlowLoaderUtils.isAzkabanFlowVersion20(this.azkabanFlowVersion) ?
-        FlowLoaderUtils.loadPropsFromYamlFile(projectLoader, this, null) :
-        projectLoader.fetchProjectProperty(projectId, version, Constants.PARAM_OVERRIDE_FILE);
-
```
tsangz2013, line 526: separate the Props loading from setting the ExecutionOptions

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -586,6 +587,10 @@ public void postScheduleEvent(final Project project, final azkaban.spi.EventType
     project.fireEventListeners(ProjectEvent.create(project, type, eventData));  // todo: add IP?
   }
 
+  public Props loadPropsForExecutableFlow(ExecutableFlow flow) throws ProjectManagerException {
+    return FlowLoaderUtils.loadPropsForExecutableFlow(this.projectLoader, flow);
+  }
+
```
tsangz2013, line 593: approach 1 to get the project Props

azkaban-common/src/main/java/azkaban/project/FlowLoaderUtils.java
```diff
@@ -388,6 +388,17 @@ public static Props loadPropsFromYamlFile(final ProjectLoader projectLoader,
...
+  public static Props loadPropsForExecutableFlow(
+      ProjectLoader projectLoader, ExecutableFlow executableFlow)
+      throws ProjectManagerException {
+    return FlowLoaderUtils.isAzkabanFlowVersion20(executableFlow.getAzkabanFlowVersion()) ?
+        FlowLoaderUtils.loadPropsFromYamlFile(projectLoader, executableFlow, null) :
+        projectLoader.fetchProjectProperty(
+            executableFlow.getProjectId(),
+            executableFlow.getVersion(),
+            Constants.PARAM_OVERRIDE_FILE);
+  }
```
tsangz2013, line 400: Extract the Props loading to a helper method

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
```diff
@@ -110,6 +112,20 @@ public static ExecutableFlow createExecutableFlow(final Project project, final F
...
+    try {
+      executorManagerAdapter.preloadExecutionOptions(exFlow);
+    } catch (ExecutorManagerException e) {
+      if(logger != null) {
+        logger.warn("Fail to preload ExecutableFlow, continue without loading props", e);
+      }
+    }
+    return exFlow;
+  }
+
```
tsangz2013, line 128: Use the new way to create ExecutableFlow so as the `ExecutionOptions` can be there

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -127,6 +127,16 @@ public ExecutableFlow getExecutableFlow(final int execId)
     return this.executorLoader.fetchExecutableFlow(execId);
   }
 
+  @Override
+  public void preloadExecutionOptions(ExecutableFlow exFlow) throws ExecutorManagerException {
+    try {
+      exFlow.setFlowParamsFromProps(this.projectManager.loadPropsForExecutableFlow(exFlow));
```
tsangz2013, line 133: 4. and in `AbstractExecutorManagerAdapter`, use `projectManager` to load the Props

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -771,6 +771,7 @@ private void ajaxFetchExecutableFlowInfo(final HttpServletRequest req,
     }
     ret.put("nodeStatus", nodeStatus);
     ret.put("disabled", options.getDisabledJobs());
+    ret.put("ExecutionOptions", options);
```
SilhDing, line 774: nit: `ExecutionOptions` -> `executionOptions` for consistency with above?

SilhDing (changes requested):

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -127,6 +127,16 @@ public ExecutableFlow getExecutableFlow(final int execId)
...
+    } catch (ProjectManagerException e) {
+      logger.info("failed to load and set the FlowParameters for ExecutableFlow", e);
```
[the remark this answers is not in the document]
tsangz2013: good call, maybe we don't need logging here

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
```diff
@@ -104,10 +106,27 @@ public static Flow getFlow(final Project project, final String flowName) {
...
+  public static ExecutableFlow createExecutableFlow(final Project project, final Flow flow,
+   ExecutorManagerAdapter executorManagerAdapter, Logger logger) {
+    ExecutableFlow exFlow = new ExecutableFlow(project, flow);
+    exFlow.addAllProxyUsers(project.getProxyUsers());
+
+    if (executorManagerAdapter == null) {
+      return exFlow;
+    }
+    try {
+      executorManagerAdapter.preloadExecutionOptions(exFlow);
```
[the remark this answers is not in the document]
tsangz2013: This way we are going back to the start.
Separate `fetching Props from project` (IO thing) from `setting Props into ExecutableFlow`(only in memory) is the key here.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -586,6 +587,10 @@ public void postScheduleEvent(final Project project, final azkaban.spi.EventType
     project.fireEventListeners(ProjectEvent.create(project, type, eventData));  // todo: add IP?
   }
 
+  public Props loadPropsForExecutableFlow(ExecutableFlow flow) throws ProjectManagerException {
```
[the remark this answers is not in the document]
SilhDing: Got it thanks!

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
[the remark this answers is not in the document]
SilhDing: ah okay I think I get the point now.

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
```diff
@@ -104,10 +106,27 @@ public static Flow getFlow(final Project project, final String flowName) {
...
+    } catch (Throwable e) {
+      if(logger != null) {
+        logger.warn("Fail to preload ExecutableFlow, continue without loading ExecutionOptions", e);
+      }
+      // having issue, fallback to without loading ExecutionOptions
+      exFlow = new ExecutableFlow(project, flow);
+      exFlow.addAllProxyUsers(project.getProxyUsers());
```
SilhDing, line 126: This is duplicate with line 113. I think we can just add the proxy users right before returning the `exFlow` at the end of this method.

SilhDing (changes requested):

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
tsangz2013: I think it's safer as we are not sure in the future the previous steps rely on these proxyUsers info

SilhDing (dismissed):

SilhDing (dismissed):

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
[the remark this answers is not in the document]
byjiang1996: How about taking out createExecutableFlow from util to a method of ExecutorManagerAdapter? 
I think it is a class method instead of util anymore b/c a well-defined ExecutorManagerAdapter class object is passed in

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
byjiang1996: Hmmm, i'm lost. Why not returning `return exFlow;` directly but constructing the exFlow again

amielkelinkedin: Perhaps the issue would be avoided entirely if we checked if the executorManagerAdapter was null at the top of the stack?

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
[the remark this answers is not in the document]
tsangz2013: good point!

azkaban-common/src/main/java/azkaban/flow/FlowUtils.java
tsangz2013: This line 125/126 is to avoid the case that `preloadExecutionOptions()` messed up the exFlow then throw the exception

byjiang1996 (approved):

ShellyWEI:
Can you also update the public user manual with the change? As ExecutionOptions are pined to very limited aspects: Notification/Concurrency/Failure Options/Runtime properties. However runtime properties itself is a concept that only should applies once for that unique execution. While we apply those flow level properties to every executions, we should give it another name. So both UI and DSL can be aligned.
Provide such example on https://azkaban.readthedocs.io/en/latest/getStarted.html

PR #3270, opened 2023-04-12
Reduce memory pressure on WebServer by preventing loading unnecessary FlowProps from MySQL to Memory

byjiang1996:
From our investigations, web server unnecessarily loads too many FlowProps objects from database to memory which causes too high memory footprints and thus causing web server out-of-memory easily under decent load. Each FlowProps object is a user defined hashmap which can be huge in size. As a result, we should prevent loading unnecessary FlowProps from MySQL to Memory if possible.

Warning: Possibly read one commit at a time is easier for review

1. Split ExecutorLoader.fetchUnfinishedFlow to fine-grained methods and onboard to rest of the codebase to reduce unnecessary memory footprint:
1.1. When render Azkaban Job/Flow logs for a given executionId in getFlowLogData() OR cancelling flow using cancelFlow(), no need to read all unfinished executions from MySQL and do filter in application layer but should only do filtering where execId == executionId in database layer
1.2. When counting the number of running flows using getRunningFlows() for flow concurrency level check and metrics reporting and etc, no need to read all columns (contains FlowProps) of ExecutableFlow object but only aggregated count or execId column is enough.
2. Replace ExecutorLoader.fetchQueuedFlows with selectQueuedFlows to reduce memory footprint
2.1. When counting the number of queued flows using getQueuedFlowSize(), no need to read all columns (contains FlowProps) of ExecutableFlow object but only aggregated count or execId column is enough.
3. Replace ExecutorLoader.fetchAgedQueuedFlows with selectAgedQueuedFlows to reduce memory footprint
3.1. When counting the number of agedQueued flows using getAgedQueuedFlowSize(), no need to read all columns (contains FlowProps) of ExecutableFlow object but only aggregated count or execId column is enough.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -38,12 +38,14 @@ import azkaban.utils.FileIOUtils.LogData;
 import azkaban.utils.Pair;
 import azkaban.utils.Props;
 import azkaban.utils.ServerUtils;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -423,29 +425,26 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
    */
   @Override
   public List<Integer> getRunningFlows(final int projectId, final String flowId) {
-    final List<Integer> executionIds = new ArrayList<>();
     try {
-      executionIds.addAll(ExecutorUtils.getRunningFlowsHelper(projectId, flowId,
-          this.executorLoader.fetchUnfinishedFlows().values()));
+      return this.executorLoader.selectUnfinishedFlows(projectId, flowId);
     } catch (final ExecutorManagerException e) {
       logger.error("Failed to get running flows for project " + projectId + ", flow "
           + flowId, e);
     }
-    return executionIds;
+    return new ArrayList<>();
   }
 
   /**
-   * Get all running (unfinished) flows from database. {@inheritDoc}
+   * Get all running (unfinished) flow ids from database. {@inheritDoc}
    */
   @Override
-  public List<ExecutableFlow> getRunningFlows() {
-    final ArrayList<ExecutableFlow> flows = new ArrayList<>();
+  public List<Integer> getRunningFlowIds() {
     try {
-      getFlowsHelper(flows, this.executorLoader.fetchUnfinishedFlows().values());
+      return this.executorLoader.selectUnfinishedFlows();
     } catch (final ExecutorManagerException e) {
-      logger.error("Failed to get running flows.", e);
+      logger.error("Failed to get running flow ids.", e);
     }
-    return flows;
+    return new ArrayList<>();
   }
 
   protected LogData getFlowLogData(final ExecutableFlow exFlow, final int offset, final int length,
@@ -569,11 +568,9 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
   public void cancelFlow(ExecutableFlow exFlow, String userId)
       throws ExecutorManagerException {
     synchronized (exFlow) {
-      final Map<Integer, Pair<ExecutionReference, ExecutableFlow>> unfinishedFlows = this.executorLoader
-          .fetchUnfinishedFlows();
-      if (unfinishedFlows.containsKey(exFlow.getExecutionId())) {
-        final Pair<ExecutionReference, ExecutableFlow> pair = unfinishedFlows
-            .get(exFlow.getExecutionId());
+      final Pair<ExecutionReference, ExecutableFlow> pair = this.executorLoader
+          .fetchUnfinishedFlow(exFlow.getExecutionId());
+      if (pair != null) {
         handleCancelFlow(pair.getFirst(), exFlow, userId);
       } else {
         final ExecutorManagerException eme = new ExecutorManagerException("Execution "
@@ -746,25 +743,6 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
   }
 
   /**
-   * Checks whether the given flow has an active (running, non-dispatched) execution from database.
-   * {@inheritDoc}
-   */
-  @Override
-  public boolean isFlowRunning(final int projectId, final String flowId) {
-    boolean isRunning = false;
-    try {
-      isRunning = isFlowRunningHelper(projectId, flowId,
-          this.executorLoader.fetchUnfinishedFlows().values());
-
-    } catch (final ExecutorManagerException e) {
-      logger.error(
-          "Failed to check if the flow is running for project " + projectId + ", flow " + flowId,
-          e);
-    }
-    return isRunning;
-  }
-
-  /**
    * Find all the Hadoop/Spark application ids present in the Azkaban job log. When iterating over
    * the set returned by this method the application ids are in the same order they appear in the
    * log.
@@ -812,26 +790,6 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
       final Collection<Pair<ExecutionReference, ExecutableFlow>> collection) {
     collection.stream().forEach(ref -> allIds.add(ref.getSecond().getExecutionId()));
     Collections.sort(allIds);
-  }
-
-  /* Search a running flow in a collection */
-  protected boolean isFlowRunningHelper(final int projectId, final String flowId,
-      final Collection<Pair<ExecutionReference, ExecutableFlow>> collection) {
-    for (final Pair<ExecutionReference, ExecutableFlow> ref : collection) {
-      if (ref.getSecond().getProjectId() == projectId
-          && ref.getSecond().getFlowId().equals(flowId)) {
-        return true;
-      }
-    }
-    return false;
-  }
-
-  /**
-   * Helper method to get all flows from collection.
-   */
-  protected void getFlowsHelper(final ArrayList<ExecutableFlow> flows,
-      final Collection<Pair<ExecutionReference, ExecutableFlow>> collection) {
-    collection.stream().forEach(ref -> flows.add(ref.getSecond()));
   }
 
   /* Helper method for getActiveFlowsWithExecutor */
--- a/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
+++ b/azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
@@ -143,9 +143,9 @@ public abstract class AbstractExecutorManagerAdapter extends EventHandler implem
     long startTime = System.currentTimeMillis();
     // TODO(anish-mal) FetchQueuedExecutableFlows does a lot of processing that is redundant, since
     // all we care about is the count. Write a new class that's more performant and can be used for
-    // metrics. this.executorLoader.fetchAgedQueuedFlows internally calls FetchQueuedExecutableFlows.
+    // metrics. this.executorLoader.selectAgedQueuedFlows internally calls SelectQueuedExecutableFlows.
     try {
-      size = this.executorLoader.fetchAgedQueuedFlows(Duration.ofMinutes(minimumAgeInMinutes))
+      size = this.executorLoader.selectAgedQueuedFlows(Duration.ofMinutes(minimumAgeInMinutes))
           .size();
       logger.info("Time taken to fetch size of queued flows is " + (System.currentTimeMillis() - startTime) / 1000);
     } catch (final ExecutorManagerException e) {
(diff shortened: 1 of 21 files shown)


azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -619,32 +600,8 @@ public List<Integer> handle(final ResultSet rs) throws SQLException {
...
-        "SELECT exec_id, enc_type, flow_data, status FROM execution_flows "
-            + "WHERE project_id=? AND flow_id=? AND status=? "
-            + "ORDER BY exec_id DESC LIMIT ?, ?";
-    // Fetch flows that are in preparing state for more than a certain duration.
-    private static final String FETCH_FLOWS_QUEUED_FOR_LONG_TIME =
-        "SELECT exec_id, enc_type, flow_data, status FROM execution_flows"
-            + " WHERE submit_time < ? AND status = "
-            + Status.PREPARING.getNumVal();
+    static String FETCH_EXECUTABLE_FLOW_BASE_QUERY =
+        "SELECT ef.exec_id, ef.enc_type, ef.flow_data, ef.status FROM execution_flows ef ";
```
tsangz2013, line 650: Can we manage to keep the corresponding new constant the same place / order so maybe github can help display it in a way that's easier to read the diff?

tsangz2013:
nit: maybe we can rename the new methods from `select...` to `fetch...IDs`

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -238,14 +238,14 @@ List<Pair<ExecutionReference, ExecutableFlow>> fetchQueuedFlows()
...
-   * This method is used to get flows fetched in Queue. Flows can be in queue in ready, dispatching
+   * This method is used to get flow ids fetched in Queue. Flows can be in queue in ready, dispatching
    * or preparing state while in queue. That is why it is expecting status in parameter.
    *
    * @param status
    * @return
    * @throws ExecutorManagerException
    */
-  List<Pair<ExecutionReference, ExecutableFlow>> fetchQueuedFlows(Status status)
+  List<Integer> selectQueuedFlows(Status status)
```
tsangz2013, line 248: nit: maybe instead of `select...` we can name the new methods to `fetch...IDs`

test/execution-test-data/basic_flow.flow
```diff
@@ -0,0 +1,43 @@
+!!azkaban.project.NodeBean
```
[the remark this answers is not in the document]
byjiang1996: Ah, somehow they are auto-generated. Let me remove. thanks for catching this

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
byjiang1996: I personally think so....
But in *DAO.java, we use the term "select" to select only execId/id field and term "fetch" to fetch the whole row. Wanna keep that convention.

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
byjiang1996: Agree. Let me do it

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -42,8 +42,12 @@ List<ExecutableFlow> fetchRecentlyFinishedFlows(Duration maxAge)
   Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchActiveFlows(DispatchMethod dispatchMethod)
       throws ExecutorManagerException;
 
+  Pair<ExecutionReference, ExecutableFlow> fetchUnfinishedFlow(final int executionId)
+      throws ExecutorManagerException;
   Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedFlows()
       throws ExecutorManagerException;
+  List<Integer> selectUnfinishedFlows(final int projectId, final String flowId) throws ExecutorManagerException;
+  List<Integer> selectUnfinishedFlows() throws ExecutorManagerException;
```
byjiang1996, line 50: We split fetchUnfinishedFlows into 4 different functions and let other classes to consume

azkaban-common/src/main/java/azkaban/executor/ExecutorLoader.java
```diff
@@ -264,7 +268,7 @@ List<ExecutableFlow> fetchFreshFlowsForStatus(final Status status,
       final ImmutableMap<Status, Pair<Duration, String>> validityMap)
       throws ExecutorManagerException;
 
-  List<ExecutableFlow> fetchAgedQueuedFlows(
+  List<Integer> selectAgedQueuedFlows(
```
byjiang1996, line 271: We replace fetchAgedQueuedFlows to selectAgedQueuedFlows and  let other classes to consume

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
byjiang1996: Fixed

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
```diff
@@ -124,6 +128,20 @@ Map<Integer, Pair<ExecutionReference, ExecutableFlow>> fetchUnfinishedFlows()
...
 
+  Pair<ExecutionReference, ExecutableFlow> fetchUnfinishedFlow(final int executionId)
+      throws ExecutorManagerException {
+    try {
+      Iterator<Pair<ExecutionReference, ExecutableFlow>> iterator = this.dbOperator.query(FetchActiveExecutableFlows.FETCH_UNFINISHED_EXECUTABLE_FLOW_BY_EXECID,
+          new FetchActiveExecutableFlows(), executionId).values().iterator();
+      if (iterator.hasNext()) {
+        return iterator.next();
+      }
+      return null;
```
amielkelinkedin, line 139: If we're returning null, can we make this into an optional?

azkaban-common/src/main/java/azkaban/executor/Status.java
```diff
@@ -57,6 +58,9 @@ public enum Status {
   public static final ImmutableSet<Status> RESTARTABLE_TERMINAL_STATUSES =
       ImmutableSet.of(Status.EXECUTION_STOPPED, Status.FAILED);
 
+  public static final ImmutableSet<Status> TERMINATING_STATUSES = ImmutableSet.of(Status.SUCCEEDED,
```
amielkelinkedin, line 61: Based on other names in the class, should this not be TERMINAL_STATUSES?

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -143,9 +143,9 @@ public long getAgedQueuedFlowSize() {
     long startTime = System.currentTimeMillis();
     // TODO(anish-mal) FetchQueuedExecutableFlows does a lot of processing that is redundant, since
```
ShellyWEI, line 144: We can remove this TODO too, since we're already improving the performance!

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
```diff
@@ -141,9 +143,9 @@ public long getAgedQueuedFlowSize() {
...
     // all we care about is the count. Write a new class that's more performant and can be used for
-    // metrics. this.executorLoader.fetchAgedQueuedFlows internally calls FetchQueuedExecutableFlows.
+    // metrics. this.executorLoader.selectAgedQueuedFlows internally calls SelectQueuedExecutableFlows.
     try {
-      size = this.executorLoader.fetchAgedQueuedFlows(Duration.ofMinutes(minimumAgeInMinutes))
+      size = this.executorLoader.selectAgedQueuedFlows(Duration.ofMinutes(minimumAgeInMinutes))
```
ShellyWEI, line 148: I also just wonder this is returning size only，why not count(exe_id) from DB. Since it is also just reading all exe_ids and appending them to a list, and we get list.size()

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -103,6 +108,36 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
   }
 
+  /**
+   * Select flows that are not in finished status, including both dispatched and non-dispatched
+   * flows.
+   *
+   * @return unfinished flow exec_ids
+   * @throws ExecutorManagerException the executor manager exception
+   */
+  List<Integer> selectUnfinishedFlows() throws ExecutorManagerException {
```
ShellyWEI, line 118: I am glad to see finally we're adopting smart queries!

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
```diff
@@ -103,6 +108,36 @@ public void uploadExecutableFlow(final ExecutableFlow flow)
...
+    try {
+      return this.dbOperator.query(SelectFromExecutionFlows.SELECT_EXECUTION_BASE_QUERY
+              + "WHERE status NOT IN ("
```
ShellyWEI, line 121: Let's also make sure that those selections have good index backup. execution_flows is a huge table, even if we're applying where clause but without index, it's still killing slow. 
I assume most of them had indexes already, but just to be safe to run a double check before merge, especially those time based columns.

azkaban-common/src/main/java/azkaban/jmx/JmxExecutionController.java
```diff
@@ -33,7 +33,7 @@ public JmxExecutionController(final ExecutionController controller) {
 
   @Override
   public int getNumRunningFlows() {
-    return this.controller.getRunningFlows().size();
+    return this.controller.getRunningFlowIds().size();
```
ShellyWEI, line 36: I searched this `getRunningFlowIds().size()` it appears 4 times in this single PR. Also we know getRunningFlows this is on our hotpath. 
So I think it's worth to have a separate method that we count from DB directly.

ShellyWEI (commented):
Very exciting PR! Great work @byjiang1996 
Left a few comments/questions.

azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java
byjiang1996: Yeah, I also thought about that. After looking into our codebase, returning null is already like a convention for us:
1. https://github.com/azkaban/azkaban/blob/bfd36e24835f292a38b2afac2451580f06637267/azkaban-common/src/main/java/azkaban/executor/FetchActiveFlowDao.java#L297
2. https://github.com/azkaban/azkaban/blob/dc7203d512cd03172b8f85a4012aae488c387f09/azkaban-common/src/main/java/azkaban/executor/ExecutionLogsDao.java#L209

azkaban-common/src/main/java/azkaban/executor/Status.java
byjiang1996: Makes sense. Let me rename

azkaban-common/src/main/java/azkaban/jmx/JmxExecutionController.java
byjiang1996: Agree but I think it is fine. 3 out of 4 such places occur in JMX* but we didn't enable JMX metrics server internally.

azkaban-common/src/main/java/azkaban/executor/Status.java
byjiang1996: DONE

azkaban-common/src/main/java/azkaban/executor/ExecutionFlowDao.java
byjiang1996: DONE. Added a submit_time index

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: DONE

azkaban-common/src/main/java/azkaban/executor/AbstractExecutorManagerAdapter.java
byjiang1996: It is a debatable question I think...
I leave such util there for future references in case if anyone wants to inspect into the AgedQueuedFlow's ExecIds. From my perspective, such exec_id list is small enough in size.
I also wonder how should we organize the code better. For each method in DAO, we need Fetch* (for fetching the row), Select* (for selecting the ids/keys) and Count* (only for counting purpose). Too many similar methods in one class will cause some maintenance overheads and cause some difficulty to write unit test: i.e. you probably need to mock all three of such combinations if you don't want to know which method is called internally. That's why I don't introduce Count* methods in this PR. All my assumption is: Select* is good enough in performance.

ShellyWEI: if we're selecting only running flows based on the limitations we put it is always flow amount. I agree that in all the scenarios (such as we need to know size and the exact exec_ids in the same code blocks) we should do select operation. Otherwise if I only need to know size, but I still create a list of exec_ids, (image the holdem3 outage where a n hourly flow has 200+ running flows, and it means each execution GC will have to collect multiple list with 200+ items verses only collecting an integer) in this case, the GC is "in general" faster to recycle an integer over an ArrayList.
That's why we should examine if such SIZE_USAGE_ONLY operations is frequent enough, it will make an impact.

ShellyWEI (approved):
Thanks for addressing my comments. 
lgtm overall

ShellyWEI (approved):

PR #3272, opened 2023-04-17
Add flow name only filter to VPA filter

byjiang1996:
Originally, we exclude some (projectName, \*) and (projectName, flowName) from VPA flow resource recommendation feature. It does not work for some edge cases where projectName is different but flowName is the same. As a result, we should further add (\*, flowName) support to exclude them from VPA feature.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
@@ -28,8 +28,12 @@ import org.slf4j.Logger;
  * proj3:flow1
  */
 public class VPAFlowCriteria {
-  // Flows are stored in a map where key is project name and value is set of flows.
-  private Map<String, Set<String>> flows = new HashMap<>(1);
+  // Projects are stored in a set: any flow under those projects will be
+  private Set<String> projects = new HashSet<>();
+  // Flows are stored in a set: will be filtered no matter which project it belongs to
+  private Set<String> flows = new HashSet<>();
+  // Flows are stored in a map where key is project name and value is set of flows: any (project, flow) pair will be filtered
+  private Map<String, Set<String>> projectFlows = new HashMap<>();
   private final String fileLocation;
   private final Logger logger;
 
@@ -42,7 +46,7 @@ public class VPAFlowCriteria {
     this.fileLocation = azkProps.getString(Constants.
         ContainerizedDispatchManagerProperties.KUBERNETES_VPA_FLOW_FILTER_FILE, null);
     this.logger = logger;
-    loadFlowFilter(new HashMap<>());
+    loadFlowFilter();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
```diff
@@ -42,33 +46,36 @@ public VPAFlowCriteria(final Props azkProps, final Logger logger) {
...
   }
 
   /**
    * Read the flow filter file and create a map of flow names. The read happens
    * on best effort basis. i.e, if a flow name is not correctly formatted, it
    * is ignored.
    */
-  private void loadFlowFilter(final Map<String, Set<String>> flowMap, final String fileLocation) {
+  synchronized private void loadFlowFilter(final String fileLocation) {
+    final Set<String> projectSet = new HashSet<>();
```
amielkelinkedin, line 76: It makes more sense to me to instantiate the objects closer to where they will be used.

azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
```diff
@@ -132,10 +145,6 @@ public boolean IsVPAEnabledForFlow(final ExecutableFlow flow) {
 
   @VisibleForTesting
   public boolean flowExists(final String projectName, final String flowName) {
-    if (!this.flows.containsKey(projectName)) {
-      return false;
-    }
-    final Set<String> flowNames = this.flows.get(projectName);
-    return flowNames == null || flowNames.contains(flowName);
+    return this.projects.contains(projectName) || this.flows.contains(flowName) || this.projectFlows.getOrDefault(projectName, new HashSet<>()).contains(flowName);
```
amielkelinkedin, line 148: his.projectFlows.getOrDefault(projectName.contains(flowName), false)?

amielkelinkedin:
OOC, any particular reason why we filter out flows when the flow name is the same?

azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
byjiang1996: Yeah, that's why I moved it from outside of loadFlowFilter to inside.

byjiang1996:
> OOC, any particular reason why we filter out flows when the flow name is the same?

In some edge cases, some customer creates a large amount of new project with different project names and launches flow executions from there. In the current VPA logic, for new flows, we grant them with max CPU and memory. If we do not exclude those bad actors, we will exhaust our nodepool capacity easily.

azkaban-common/src/main/java/azkaban/executor/container/VPAFlowCriteria.java
amielkelinkedin: You create the objects, fileLocation is null, now you have a bunch of objects you're not going to use. Why not create before the try?

byjiang1996: Ah, I see your point now. Agree. Let me refactor this piece.

byjiang1996: done

amielkelinkedin (approved):
Thanks for the change @byjiang1996!

PR #3274, opened 2023-04-19
Add timeout to dependency download at project upload stage

ShellyWEI:
This PR is mostly optimization for gcn-38523 where upload project zips was hanging during validationProject, but it is unclear to us which part (download dependency, or validate dependency) failed, or it's possible that upload thread suddenly gone was caused by maybe uncaught exception/throwable (for example, Downloading Dependencies if any exception, the completableFuture.completeExceptionally(Throwable), we throw throwable instead of Exception).
Basically two scenarios could cause a "hanging" upload:
- Downloading itself is hanging, such as waiting inputStream;
- There is a throwable, but our code does not have such logic to handle it, so this thread just dead without telling CRT it is failed;

To address the issue above, a few optimization were made:
- Add default timeout 5 minutes for each dependency download tasks. Allow config to override the default value;
- Optimize logs for download/validate dependency part to improve debug efficiency;
- Add catch native Throwable logic;

### DIFF
--- a/azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
+++ b/azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
@@ -142,14 +142,19 @@ public class DependencyTransferManager {
     }
   }
 
-  /* Cancel all the not-started tasks which are possibly waiting in the queue */
-  private static void cancelPendingTasks(CompletableFuture[] taskFutures, String projectName) {
-    logger.info("cancelling the pending tasks for project {} because one of the downloads failed.", projectName);
-    boolean cancelled = true;
+  /* Cancel all the not-started tasks which are possibly waiting in the queue,
+  * Cancel() itself will not throw any Exception to cause a break of the loop, but only completes those futures with
+  * CancellationException. If we want to query the status of the future, we need to use isCancelled() method.
+  * If we want to get the result of the future, we need to use get() method, which will throw CancellationException.
+  * Note that future completed/failed/cancelled will return false for cancel() method, as well as fail to cancel the task.
+  * It makes the returning value invaluable for us since it's possible that some tasks is already completed/failed.
+  *  */
+  @VisibleForTesting
+  void cancelPendingTasks(CompletableFuture[] taskFutures, String projectName) {
     for (CompletableFuture future : taskFutures) {
-      cancelled &= future.cancel(false);
+      future.cancel(false);
     }
-    logger.info("{} cancel the pending tasks for project {}.", cancelled ? "Successfully" : "Failed to", projectName);
+    logger.info("Cancelled the pending tasks for project {} because one of the downloads failed.", projectName);
   }
 
   private void downloadDependency(final DependencyFile f, final int retries, String projectName)
--- a/azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
+++ b/azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
@@ -28,12 +28,15 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.apache.commons.io.IOUtils;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import static azkaban.Constants.ConfigurationKeys.AZKABAN_DEPENDENCY_DOWNLOAD_THREADPOOL_SIZE;
+import static azkaban.Constants.ConfigurationKeys.*;
 import static azkaban.utils.ThinArchiveUtils.*;
 
 /**
@@ -47,11 +50,13 @@ public class DependencyTransferManager {
 
   public final int dependencyMaxDownloadTries;
 
+  public final int dependencyDownloadMaxTimeout;
+
   private final Storage storage;
 
   private final ExecutorService threadPool;
 
-  private static final Logger logger = Logger.getLogger(DependencyTransferManager.class);
+  private static final Logger logger = LoggerFactory.getLogger(DependencyTransferManager.class);
 
   @Inject
   public DependencyTransferManager(final Props props, final Storage storage) {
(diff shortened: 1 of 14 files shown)


utk-12:
thank you for making this change ! Could you please add screenshots or examples of how this was tested in the PR ?

ShellyWEI:
> thank you for making this change ! Could you please add screenshots or examples of how this was tested in the PR ?

You mean the timeout part?
It is coming from Java standard library. I believe the unit test is composed inside concurrency package. Otherwise it won't be an official method.

utk-12:
> > thank you for making this change ! Could you please add screenshots or examples of how this was tested in the PR ?
> 
> You mean the timeout part? It is coming from Java standard library. I believe the unit test is composed inside concurrency package. Otherwise it won't be an official method.

Not exactly. I meant, can we emulate an integration test where we emulate one of the dependency being unavailable, to see if the cancel works as expected.

ShellyWEI:
> > > thank you for making this change ! Could you please add screenshots or examples of how this was tested in the PR ?
> > 
> > 
> > You mean the timeout part? It is coming from Java standard library. I believe the unit test is composed inside concurrency package. Otherwise it won't be an official method.
> 
> Not exactly. I meant, can we emulate an integration test where we intentionally make one of the dependencies unavailable, to see if the cancel works as expected.

We don't have any integration test in this repo, but I do add a unit test. You can use that as reference

azkaban-common/src/test/java/azkaban/utils/DependencyTransferManagerTest.java
```diff
@@ -74,6 +78,7 @@ public void setup() throws Exception {
 
     Props sampleProps = new Props();
     sampleProps.put(Constants.ConfigurationKeys.AZKABAN_DEPENDENCY_MAX_DOWNLOAD_TRIES, DEPENDENCY_DOWNLOAD_MAX_TRIES);
+    sampleProps.put(AZKABAN_DEPENDENCY_DOWNLOAD_TIMEOUT_SECONDS, 10);
```
tsangz2013, line 81: suggest we lower to 2s, so as less test overhead

tsangz2013:
I see you added `projectName` at places for better logging, please add to description

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -93,45 +100,57 @@ public void downloadAllDependencies(final Set<DependencyFile> deps) {
...
 
   /* Cancel all the not-started tasks which are possibly waiting in the queue */
-  private static void cancelPendingTasks(CompletableFuture[] taskFutures) {
-    logger.error("cancelling the pending tasks because one of the downloads failed");
+  private static void cancelPendingTasks(CompletableFuture[] taskFutures, String projectName) {
+    logger.info("cancelling the pending tasks for project {} because one of the downloads failed.", projectName);
+    boolean cancelled = true;
     for (CompletableFuture future : taskFutures) {
-      future.cancel(false);
+      cancelled &= future.cancel(false);
```
byjiang1996, line 150: I'm wondering if we need to log cancelled in the log.
The canceled flag means: 
> The CompletableFuture.cancel method returns a boolean value that indicates whether the cancellation was successful or not. If the CompletableFuture was not previously completed, then it will be cancelled and the method will return true. However, if the CompletableFuture was already completed or cancelled, then the cancel operation will have no effect and the method will return false.

If the cancellation really fails, it will throw an exception

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
```diff
@@ -61,6 +66,7 @@ public DependencyTransferManager(final Props props, final Storage storage) {
         new ThreadFactoryBuilder().setNameFormat("azk-dependency-pool-%d").build());
     this.dependencyMaxDownloadTries =
         props.getInt(Constants.ConfigurationKeys.AZKABAN_DEPENDENCY_MAX_DOWNLOAD_TRIES, 2);
+    this.dependencyDownloadMaxTimeout = props.getInt(AZKABAN_DEPENDENCY_DOWNLOAD_TIMEOUT_SECONDS, 300);
```
byjiang1996: Do you have some metrics to show if 5min is enough or not?

azkaban-common/src/test/java/azkaban/utils/DependencyTransferManagerTest.java
ShellyWEI: I retried that, unfortunately the other tests in this class would randomly fail. We need some number that is good enough to cause less flakyness. I will try if 5s works

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
ShellyWEI: Not really, cancellation exception would always throw the difference is whether it is direct CancellationException or ExecutionException wrapping `CancellationException`.
Reference from Java doc:
>      * If not already completed, completes this CompletableFuture with
     * a {@link CancellationException}. Dependent CompletableFutures
     * that have not already completed will also complete
     * exceptionally, with a {@link CompletionException} caused by
     * this {@code CancellationException}.
I don't think it really matters if cancel successfully or not as those tasks are async, I just want to have a log to reveal any issues if we observe in the future that could caused by failure of cancel. But personally I don't think it matters to the upload itself. Just in case this might be useful in the future.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
ShellyWEI: unfortunately I don't know if that's enough, I am using your log "time takes to download" as reference, it's usually really fast. most of them finished with 0 or 1s. 
But again I am uncertain about that. So in case any issues we can quickly restart the webserver by changing config.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
byjiang1996: Yeah, it also confuses me. If exception would always be thrown, how can you run a for loop here? Wouldn't it break in the middle?

I agree having a log line is nice. I think probably we shouldn't log the "cancelled ? "Successfully" : "Failed to"" in the log line b/c it is not same definition of "Success" or "Fail" we think about when first see this log line. My definition of "Cancel Fail" is: cancellation failed completely but not "it has been completed or cancelled". Maybe that's my own thinking though.

It will confuse our developers who read that log line.

azkaban-common/src/test/java/azkaban/utils/DependencyTransferManagerTest.java
ShellyWEI: change to 3s.

azkaban-common/src/main/java/azkaban/utils/DependencyTransferManager.java
ShellyWEI: This is actually a good discussion. I add one more unit test to verify that.
It turns out cancel() itself won't throw any exception, but it will mark the future with cancellationException. So if we want to get result, it will throw that exception.
And thanks for the log description, I was wrong about the returning value too. Completed future will result in false, I used to think only cancel failed for other reasons will return false. Anyway I removed the boolean value.

byjiang1996 (approved):

PR #3275, opened 2023-04-19
Flow-retry - store, log & report the parent-child lineage

tsangz2013:
**Why we need this**
* As the flow-level-retry happens, we should let users know about the which execution is the "child" retried one, and which is the retried's "parent" execution. 
* In following PRs, we will also need to improve the webUI so as this information can be exposed to users in an efficient way.

**What's done**
1. add fields for an `ExecutableFlow`'s flow-level-retry `Root / (Direct) Parent / (Direct) Child` ExecutionIDs;
2. populate and persistent these fields for both the original and retried execution after successfully retrying;
3. log these lineage info + the retry-counts, also report these fields in the flow status event reporter.
4. expose all flow-retry info in ajax `fetchexecflow` endpoint

**Tests done**
1. unit tests
2. end to end tested in staging cluster:

Started execution 788402 with retry-enabled, and kill the pod to trigger EXECUTION_STOPPED status
Screenshot 2023-04-19 at 2 22 21 PM

Verify that the original execution 788402 has `flowRetryChildExecutionID` updated.
Screenshot 2023-04-19 at 2 22 49 PM

And the retried execution 788403 has `flowRetry{Root,Parent}ExecutionID` updated.
Screenshot 2023-04-19 at 2 22 57 PM

Also webserver log this:
Screenshot 2023-04-19 at 2 25 26 PM

---
Plus, when "Prepare execution" on one of these executions, these flow-retry properties all be cleaned
Screenshot 2023-04-19 at 2 27 55 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -100,9 +100,9 @@ public class ExecutableFlow extends ExecutableFlowBase {
   // how many times flow level retry happened due to stuck in "Dispatch/Preparing/Ready" status
   private int systemDefinedRetryCount = 0;
   // The IDs of the original ancestor root exec / the direct parent exec / the retried child exec
-  private int flowRetryRootExecutionID = 0;
-  private int flowRetryParentExecutionID = 0;
-  private int flowRetryChildExecutionID = 0;
+  private int flowRetryRootExecutionID = -1;
+  private int flowRetryParentExecutionID = -1;
+  private int flowRetryChildExecutionID = -1;
 
   // For slaOption information
   private String slaOptionStr = "null";
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -67,6 +67,11 @@ public class ExecutableFlow extends ExecutableFlowBase {
   private static final String USER_DEFINED_FLOW_RETRY_COUNT_PARAM = "userDefinedFlowRetryCount";
   private static final String SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM = "systemDefinedFlowRetryCount";
 
+  // Properties but non-define-able parameters
+  private static final String FLOW_RETRY_ROOT_EXECUTION_ID = "flowRetryRootExecutionID";
+  private static final String FLOW_RETRY_PARENT_EXECUTION_ID = "flowRetryParentExecutionID";
+  private static final String FLOW_RETRY_CHILD_EXECUTION_ID = "flowRetryChildExecutionID";
+
   private final HashSet<String> proxyUsers = new HashSet<>();
   private int executionId = -1;
   private int scheduleId = -1;
@@ -95,6 +100,10 @@ public class ExecutableFlow extends ExecutableFlowBase {
   private int userDefinedRetryCount = 0;
   // how many times flow level retry happened due to stuck in "Dispatch/Preparing/Ready" status
   private int systemDefinedRetryCount = 0;
+  // The IDs of the original ancestor root exec / the direct parent exec / the retried child exec
+  private int flowRetryRootExecutionID = 0;
+  private int flowRetryParentExecutionID = 0;
+  private int flowRetryChildExecutionID = 0;
 
   // For slaOption information
   private String slaOptionStr = "null";
@@ -328,6 +337,31 @@ public class ExecutableFlow extends ExecutableFlowBase {
     this.systemDefinedRetryCount = systemDefinedRetryCount;
   }
 
+  public int getFlowRetryRootExecutionID() {
+    return flowRetryRootExecutionID;
+  }
+
+  public void setFlowRetryRootExecutionID(int flowRetryRootExecutionID) {
+    this.flowRetryRootExecutionID = flowRetryRootExecutionID;
+  }
+
+  public int getFlowRetryParentExecutionID() {
+    return flowRetryParentExecutionID;
+  }
+
+  public void setFlowRetryParentExecutionID(int flowRetryParentExecutionID) {
+    this.flowRetryParentExecutionID = flowRetryParentExecutionID;
+  }
+
+  public int getFlowRetryChildExecutionID() {
+    return flowRetryChildExecutionID;
+  }
+
+  public void setFlowRetryChildExecutionID(int flowRetryChildExecutionID) {
+    this.flowRetryChildExecutionID = flowRetryChildExecutionID;
+  }
+
+
   @Override
   public Map<String, Object> toObject() {
     final HashMap<String, Object> flowObj = new HashMap<>();
@@ -381,6 +415,10 @@ public class ExecutableFlow extends ExecutableFlowBase {
       flowObj.put(VERSIONSET_MD5HEX_PARAM, this.versionSet.getVersionSetMd5Hex());
       flowObj.put(VERSIONSET_ID_PARAM, this.versionSet.getVersionSetId());
     }
+
+    flowObj.put(FLOW_RETRY_ROOT_EXECUTION_ID, flowRetryRootExecutionID);
+    flowObj.put(FLOW_RETRY_PARENT_EXECUTION_ID, flowRetryParentExecutionID);
+    flowObj.put(FLOW_RETRY_CHILD_EXECUTION_ID, flowRetryChildExecutionID);
 
     return flowObj;
   }
@@ -452,6 +490,10 @@ public class ExecutableFlow extends ExecutableFlowBase {
     // Dispatch Method default is POLL
     this.setDispatchMethod(DispatchMethod.fromNumVal(flowObj.getInt(FLOW_DISPATCH_METHOD,
         DispatchMethod.POLL.getNumVal())));
+
+    this.setFlowRetryRootExecutionID(flowObj.getInt(FLOW_RETRY_ROOT_EXECUTION_ID));
+    this.setFlowRetryParentExecutionID(flowObj.getInt(FLOW_RETRY_PARENT_EXECUTION_ID));
+    this.setFlowRetryChildExecutionID(flowObj.getInt(FLOW_RETRY_CHILD_EXECUTION_ID));
   }
 
   @Override
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
@@ -15,6 +15,12 @@
  */
 package azkaban.executor;
 
+import static azkaban.Constants.EventReporterConstants.FLOW_RETRY_CHILD_EXECUTION_ID;
+import static azkaban.Constants.EventReporterConstants.FLOW_RETRY_PARENT_EXECUTION_ID;
+import static azkaban.Constants.EventReporterConstants.FLOW_RETRY_ROOT_EXECUTION_ID;
+import static azkaban.Constants.EventReporterConstants.SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM;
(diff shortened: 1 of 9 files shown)


azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -37,29 +37,29 @@ public void onExecutionEvent(final ExecutableFlow flow, final String action) {
...
       return;
     }
-    final ExecutableFlow executableFlow =
+    final ExecutableFlow retryExFlow =
         this.executorManagerAdapter.createExecutableFlow(project, flow);
-    executableFlow.setSubmitUser(exFlow.getSubmitUser());
-    executableFlow.setExecutionSource(Constants.EXECUTION_SOURCE_ADHOC);
-    executableFlow.setUploadUser(project.getUploadUser());
+    retryExFlow.setSubmitUser(originalExFlow.getSubmitUser());
+    retryExFlow.setExecutionSource(Constants.EXECUTION_SOURCE_ADHOC);
```
ShellyWEI, line 59: I don't think the execution_source should be ad-hoc. It should be something like "auto_retry". Adhoc means user adhoc execution.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
     // Update the flow options so that the flow will be not retried again by Azkaban
 
     // inherent the retry time counters
-    executableFlow.setUserDefinedRetryCount(exFlow.getUserDefinedRetryCount());
-    executableFlow.setSystemDefinedRetryCount(exFlow.getSystemDefinedRetryCount());
+    retryExFlow.setUserDefinedRetryCount(originalExFlow.getUserDefinedRetryCount());
```
ShellyWEI, line 73: it is worthwhile to have a field indicating how many times it is retrying (e.g. "x"th retries). Plus, do we put a limit on number of times retry?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
+    retryExFlow.setSystemDefinedRetryCount(originalExFlow.getSystemDefinedRetryCount());
+
+    if (originalExFlow.getFlowRetryRootExecutionID() > 0) {
```
ShellyWEI, line 76: We should set those under if they're "allow.restart" flag is enabled. 
They could be null, as they're not enabled;
They could be 0, as retry not restarted yet.
I feel like current definition (0,-1,null) is a bit confusing to me,

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
+      options.setOriginalFlowExecutionIdBeforeRetry(originalExFlow.getExecutionId());
     }
-    executableFlow.setExecutionOptions(options);
+    retryExFlow.setExecutionOptions(options);
     // Submit new flow for execution
     try {
-      logger.info("Restarting flow " + project.getName() + "." + executableFlow.getFlowName());
-      this.executorManagerAdapter.submitExecutableFlow(executableFlow,
-          executableFlow.getSubmitUser());
+      logger.info("Restarting flow " + project.getName() + "." + retryExFlow.getFlowName());
```
ShellyWEI, line 91: log with number of retries, e.g. restarting flow *** with {[0~9]}th retry, the maxinum of auto retry is {}.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -995,6 +996,16 @@ private void ajaxFetchExecutableFlow(final HttpServletRequest req,
     ret.put("execid", exFlow.getExecutionId());
     ret.put("projectId", exFlow.getProjectId());
     ret.put("project", project.getName());
+    ret.put(EventReporterConstants.SYSTEM_DEFINED_FLOW_RETRY_COUNT_PARAM,
+        exFlow.getSystemDefinedRetryCount());
+    ret.put(EventReporterConstants.USER_DEFINED_FLOW_RETRY_COUNT_PARAM,
+        exFlow.getUserDefinedRetryCount());
+    ret.put(EventReporterConstants.FLOW_RETRY_ROOT_EXECUTION_ID,
```
ShellyWEI, line 1003: These also should only return if auto-retry feature is enabled.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
     // inherent the retry time counters
-    executableFlow.setUserDefinedRetryCount(exFlow.getUserDefinedRetryCount());
-    executableFlow.setSystemDefinedRetryCount(exFlow.getSystemDefinedRetryCount());
+    retryExFlow.setUserDefinedRetryCount(originalExFlow.getUserDefinedRetryCount());
+    retryExFlow.setSystemDefinedRetryCount(originalExFlow.getSystemDefinedRetryCount());
+
+    if (originalExFlow.getFlowRetryRootExecutionID() > 0) {
+      retryExFlow.setFlowRetryRootExecutionID(originalExFlow.getFlowRetryRootExecutionID());
+    } else {
+      retryExFlow.setFlowRetryRootExecutionID(originalExFlow.getExecutionId());
```
ShellyWEI, line 79: If this does not put under if flag, those null case (not enabled auto-retry) would be having rootExecId, which is quite confusing.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
+    }
+    retryExFlow.setFlowRetryParentExecutionID(originalExFlow.getExecutionId());
 
     // If a retried flow A gets retried again with a new execution id flow B, the original flow
     // execution id of flow B should be the same as flow A's original flow execution id.
     if (options.getOriginalFlowExecutionIdBeforeRetry() == null) {
```
ShellyWEI, line 85: I don't really get what this OriginalFlowExecutionBeforeRetry means. And can you remind me why this has to be put into executionOptions?

ShellyWEI (commented):
Maybe I am missing the previous PR's context. 
Left a few questions.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
ypadron-in: +1

tsangz2013: good point, let me add a new constant

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: yes we need this to count,
and the limit is in `ExecutionOption.FlowParameter` defined by key `"flow.max.retries"`.

This logic is implemented in #3268

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: good point! let me add

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: hmmm, this method `restartExecutableFlow` indicates it is an auto-retry, thus won't be a case "not enabled auto-retry"

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -70,22 +70,44 @@ private void restartExecutableFlow(final ExecutableFlow exFlow) {
...
-      options.setOriginalFlowExecutionIdBeforeRetry(exFlow.getExecutionId());
+      options.setOriginalFlowExecutionIdBeforeRetry(originalExFlow.getExecutionId());
```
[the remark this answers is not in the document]
tsangz2013: I don't know, maybe it's no, will check if can remove

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -452,6 +489,10 @@ public void fillExecutableFromMapObject(
     // Dispatch Method default is POLL
     this.setDispatchMethod(DispatchMethod.fromNumVal(flowObj.getInt(FLOW_DISPATCH_METHOD,
         DispatchMethod.POLL.getNumVal())));
+
+    this.setFlowRetryRootExecutionID(flowObj.getInt(FLOW_RETRY_ROOT_EXECUTION_ID));
+    this.setFlowRetryParentExecutionID(flowObj.getInt(FLOW_RETRY_PARENT_EXECUTION_ID));
+    this.setFlowRetryChildExecutionID(flowObj.getInt(FLOW_RETRY_CHILD_EXECUTION_ID));
```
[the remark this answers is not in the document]
tsangz2013: oh they are the execution IDs, default value should be -1 if not defined

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: we will only enter this method when auto-retry is defined;
plus this value is default to -1 at initialization, so no null concern

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ExecutorServlet.java
```diff
@@ -995,6 +997,21 @@ private void ajaxFetchExecutableFlow(final HttpServletRequest req,
...
+          exFlow.getSystemDefinedRetryCount());
+      ret.put(EventReporterConstants.USER_DEFINED_FLOW_RETRY_COUNT_PARAM,
+          exFlow.getUserDefinedRetryCount());
+      ret.put(EventReporterConstants.FLOW_RETRY_ROOT_EXECUTION_ID,
+          exFlow.getFlowRetryRootExecutionID());
+      ret.put(EventReporterConstants.FLOW_RETRY_PARENT_EXECUTION_ID,
+          exFlow.getFlowRetryParentExecutionID());
+      ret.put(EventReporterConstants.FLOW_RETRY_CHILD_EXECUTION_ID,
+          exFlow.getFlowRetryChildExecutionID());
+    }
```
ypadron-in, line 1014: Since these are multiple related fields, what do you think about grouping them together in a JSON object and perhaps also drop `flowRetry` from the properties names as it might be redundant (the api `fetchexecflow` retrieves information about the "flow"). When reporting events if wanted we could prefix the properties with "flowRetry". Here is an example api response with previous suggestions:
```
...
"retries": {
    "userDefinedCount": 1,
    "systemDefinedCount": 0,
    "rootExecutionID": 1234,
    "parentExecutionID: 1234,
    "childExecutionID: 0
}
...
```

ypadron-in (approved):
LGTM Thanks for addressing all the comments!

PR #3280, opened 2023-04-26
Implement an optional flow-retry-strategy "skip succeeded nodes" when submitting the retry-execution

tsangz2013:
**Why we need this**
Add to the normal default behavior that redo the whole flow all the nodes when retrying an execution, adding this new strategy to allow users opt in, so as when retrying, the retried execution will "skip" some nodes(jobs) already succeeded/skipped in the original execution.

This allows idempotent workflows to resume from previously execution faster and thus save time and compute resources.

**What's done**
* define the Constant values for the strategy.
* develop an method to walk-through (DFS) both the original execution & the new execution, disabling the newExecution's nodes if the corresponding nodes in original execution are successful.

**Tests done**
* Unit test, with normal flow structure and embedded flow structure.
* Tested in an staging webserver:

Started an execution 788534 with `"flow.retry strategy" : "disable succeeded nodes"` and `"flow.retry .statuses" : "FAILED"`
Screenshot 2023-04-20 at 5 02 11 PM

Manually failed the execution by killing its yarn-app after some jobs already succeeded
Screenshot 2023-04-20 at 5 06 06 PM

Screenshot 2023-04-20 at 5 06 12 PM

and it finally failed:
Screenshot 2023-04-20 at 5 07 35 PM

Then we can see an new execution 788535 is auto-retried:
Screenshot 2023-04-20 at 5 08 10 PM

And in the execution detail page, we can see the already succeeded job `cacheLibraries` is disabled
Screenshot 2023-04-20 at 5 09 17 PM

Also from the API we can see the status of `cacheLibraries` job is `DISABLED` and other jobs are `READY`
Screenshot 2023-04-20 at 5 08 22 PM


azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -53,13 +57,39 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
...
+        try {
+          disableSucceededSkippedJobsInRetryFlow(originalExFlow, retryExFlow);
+        } catch (ExecutorManagerException e){
+          // rebuild the execution in case data-structure already messed-up
+          retryExFlow = this.executorManagerAdapter.createExecutableFlow(project, flow);
+          logger.warn(String.format(
+              "fail to apply %s restart-strategy for the execution %s, fallback to default method",
+                  FLOW_RESTART_STRATEGY_DISABLE_SUCCEEDED_NODES,
+                  originalExFlow.getExecutionId()),
+              e);
```
ypadron-in, line 84: Wouldn't defaulting to start over strategy be unexpected behavior from user's perspective? They specifically set "start from fail nodes" but instead they notice that Azkaban rerun the entire dag for some reason not visible to them. Why not failing the retry on error and notify users that won't be one for XYZ reason?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -91,20 +121,74 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
...
         retryExFlow.getFlowRetryRootExecutionID(),
         retryExFlow.getSystemDefinedRetryCount(), retryExFlow.getUserDefinedRetryCount()));
   }
+
+
+  /**
+   * DFS walk through of the executableFlow, update the retry-flow's job status to disabled if
+   * the corresponding job status in origin-flow is considered success/skip/disable.
+   */
+  public static void disableSucceededSkippedJobsInRetryFlow(
```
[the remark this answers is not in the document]
tsangz2013: I checked this route, and I believe its a big way around:
1. DFS the original execution to find out the succeeded nodes
2. put into initiallyDisabledJobs
3. at new execution start time, apply initiallyDisabledJobs with DFS

Also, with this PR, users CAN see the nodes are being disabled by my algorithm for the new execution from UI, and if they want to restart it with the same setting, it's totally possible.
This is because when webUI display the nodes' status, it's from `ExecutableNode.status`, and when it submit execution via http call, the `disabled job` list are constructed by frontend, but not from the DB-stored `ExecutionOptions`

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -883,5 +883,11 @@ public static class FlowParameters {
 
     // Constant to define the strategy to restart the execution, default to "restart_from_root",
     public static final String FLOW_PARAM_RESTART_STRATEGY = "flow.retry.strategy";
+
+    // default restart strategy, restart the execution in a whole, fresh new
+    public static final String FLOW_RESTART_STRATEGY_DEFAULT = "default";
+
+    public static final String FLOW_RESTART_STRATEGY_DISABLE_SUCCEEDED_NODES =
```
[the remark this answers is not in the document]
tsangz2013: good call out, let me use Enum

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -883,11 +883,13 @@ public static class FlowParameters {
...
-    // default restart strategy, restart the execution in a whole, fresh new
-    public static final String FLOW_RESTART_STRATEGY_DEFAULT = "default";
+  public enum FlowRetryStrategy {
 
-    public static final String FLOW_RESTART_STRATEGY_DISABLE_SUCCEEDED_NODES =
-        "disable_succeeded_nodes";
+    // default restart strategy, restart the execution in a whole, fresh new
+    DEFAULT,
+    // restart the execution but disabling the corresponding succeeded nodes in previous execution
+    DISABLE_SUCCEEDED_NODES;
```
ypadron-in, line 893: What value do we want users to set for `flow.retry.strategy` property? 
1-DISABLE_SUCCEEDED_NODES
2-disable_succeeded_nodes
3-disableSucceededNodes - aligns with the naming pattern used by other value properties like finishCurrent, cancelImmediately, finishPossible

Also, for the DEFAULT retry strategy can we think of a more informative value? Thinking that value `default` may not explain much to users in terms of what will happen.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -76,12 +76,10 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
...
-          // rebuild the execution in case data-structure already messed-up
-          retryExFlow = this.executorManagerAdapter.createExecutableFlow(project, flow);
-          logger.warn(String.format(
-              "fail to apply %s restart-strategy for the execution %s, fallback to default method",
-                  FlowRetryStrategy.DISABLE_SUCCEEDED_NODES.name(),
-                  originalExFlow.getExecutionId()),
+          logger.error(String.format(
+              "Fail to restart execution %s due to error applying %s restart-strategy",
+                  originalExFlow.getExecutionId(), 
+                  FlowRetryStrategy.DISABLE_SUCCEEDED_NODES.name()),
```
ypadron-in, line 82: Should we notify users that Azkaban encountered an error while retrying the flow execution? If yes, it can be addressed in a separate PR but can we please add a TODO in the code for now?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -884,4 +884,12 @@ public static class FlowParameters {
     // Constant to define the strategy to restart the execution, default to "restart_from_root",
     public static final String FLOW_PARAM_RESTART_STRATEGY = "flow.retry.strategy";
   }
+
+  public enum FlowRetryStrategy {
```
ypadron-in, line 888: Should we validate these values in HttpRequestUtils class when a user submits a new execution? If addressed in a separate PR could you please add a TODO comment in the code?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -53,13 +56,39 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
       logger.error(e.getMessage());
       return;
     }
-    final ExecutableFlow retryExFlow =
+    ExecutableFlow retryExFlow =
         this.executorManagerAdapter.createExecutableFlow(project, flow);
+
+    final ExecutionOptions options = originalExFlow.getExecutionOptions();
+
+    final String restartStrategy = options.getFlowParameters().
```
ypadron-in, line 64: WDYT about converting the string value into a FlowRetryStrategy value and work with the enum type instead of a string type?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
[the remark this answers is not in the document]
ypadron-in: I see, thanks for thinking about this!
My concern is that the current code uses the original ExecutionOptions and overwrites attributes as needed and we're never modifying the list of disabled nodes so original configs will be set for retries. What nodes will ultimately be executed if the original execution had disabled nodes and the user specified retry strategy disable succeeded nodes?

az-core/src/main/java/azkaban/Constants.java
tsangz2013: good call, let me validate them in this PR

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: I don't have a good idea now, so let me add TODO and note in the design doc

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -53,13 +56,41 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
...
+
+    // default strategy - not applying anything on the new one, so as like a new execution
+    if (restartStrategy.isEmpty() || strategyEnum.equals(FlowRetryStrategy.DEFAULT)) {
+      logger.info(String.format("Use default strategy when restarting the execution %s",
+          originalExFlow.getExecutionId()));
+    } else {
+      // non-default strategies
+      if (strategyEnum.equals(FlowRetryStrategy.DISABLE_SUCCEEDED_NODES)){
+        try {
+          disableSucceededSkippedJobsInRetryFlow(originalExFlow, retryExFlow);
```
ypadron-in, line 78: I think the `restartExecutableFlow` method handles both user and system retries. In the case of system retries  (for status PREPARING, DISPATCHING) the flow execution hasn't started yet so traversing the DAG to disable some nodes doesn't make much sense as non have yet run. As an optimization, is it possible to detect the type of retry and avoid the DAG traversing if it's a system retry? It can be addressed in a separate PR but can we add a TODO in the code perhaps?

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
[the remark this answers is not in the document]
tsangz2013: According to the current code, below explains what's the ultimate state of nodes:
1. original execution being submitted with `DisabledJobs` specified
2. original execution data-structure being applied with these `DisabledJobs`, the nodes got marked statues `Disabled`
3. original execution got failed/exec-stopped, gets into retry
4. If using disabled-succeeded-nodes strategy: DFS walk through both executions, disable retry-execution's nodes whose corresponding node is `disabled` or `succeeded` in original execution.
5. retry-execution applied with these `DisabledJobs`, but because 4) is done, this == no-op.

To sum up: all that's originally disabled in original-execution, and those succeeded in original-execution, will be disabled in retry-execution.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
tsangz2013: good call out, I will add this as TODO so as system-retry on `PREPARING, DISPATCHING` can be improved

ypadron-in: thanks!

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -61,30 +61,30 @@ void restartExecutableFlow(final ExecutableFlow originalExFlow) {
...
                   originalExFlow.getExecutionId(),
-                  FlowRetryStrategy.DISABLE_SUCCEEDED_NODES.name()),
+                  FlowRetryStrategy.DISABLE_SUCCEEDED_NODES),
               e);
         }
-      }
+        break;
+      default:
+        logger.info(String.format("Use default strategy when restarting the execution %s",
+          originalExFlow.getExecutionId()));
```
ypadron-in, line 87: I think this can be removed as it's already logged in line 69-70 above.

ypadron-in (approved):
Thanks a lot for addressing all the comments!

tsangz2013:
Leave the discussion here for later reference:
This approach of disabling nodes directly set the status on the newly created retrying ExecutableFlow, skipped the step of adding the nodes to `ExecutionOptions.disabledJobs`.
* pro: not modifying the user original input `disabledJobs`;
* con: may results in gap between the original `disabledJobs` and the actually disabled nodes.

But overall this should be the case, as the node statuses are dynamic, plus it's users who opt-in the strategy and know this "disable succeeded nodes" will take effect.
CC @ypadron-in

PR #3282, opened 2023-04-26
Add more log lines to tell the latency of loading projects v.s. loadi…

byjiang1996:
…ng flow resource recommendations


djaiswal83 (approved):
Thanks for adding these log lines.

PR #3283, opened 2023-04-28
Add cleanup logic in schedule system when no Such project/flows in the system

ShellyWEI:
## Summary
We observe several missed schedule notification task failure due to no flows existed with schedule. It can be caused due to upload project failure before executing removing schedules and left schedule with no flows to execute stage.

This PR is to add the cleanup logic when detecting no such projects/flows associated with ExecuteFlowAction of schedules. 

## Details
- Introduce NoSuchProejctAndFlowsException;
- Introduce INVALID status of Trigger/Schedules, and remove triggers from DB if INVALID;
- Add sync with ScheduleManager from UI so UI can receive removed triggers timely;

## Testing
Testing done at Faro4, with successful removing triggers and UI can receive updated schedule info.


azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -176,6 +179,14 @@ public List<Trigger> getTriggers() {
     return new ArrayList<>(triggerIdMap.values());
   }
 
+  // get a list of removed triggers and clear the list
+  @Override
+  public List<Integer> getRemovedTriggerIds() {
+    List<Integer> removedTriggerIdsCopy = new ArrayList<>(removedTriggerIds);
+    removedTriggerIds.clear();
+    return removedTriggerIdsCopy;
+  }
```
tsangz2013, line 188: question: after clearing the list, if the later clean up logic got exception or broken, will those to-be-removed triggers got captured again in the future?

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
```diff
@@ -21,6 +21,7 @@
 import azkaban.executor.ExecutorManagerAdapter;
 import azkaban.flow.Flow;
 import azkaban.flow.FlowUtils;
+import azkaban.flow.NoSuchAzkabanResourceException;
```
tsangz2013, line 24: is this a void reference?

azkaban-common/src/test/java/azkaban/scheduler/MissedSchedulesManagerTest.java
```diff
@@ -19,6 +19,7 @@
 import azkaban.Constants;
 import azkaban.executor.ExecutionOptions;
 import azkaban.flow.Flow;
+import azkaban.flow.FlowUtils;
```
tsangz2013, line 22: here as well

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
ShellyWEI: No it would just be removed in memory, as the exception only possibly happen at DB layer. So for this time, we would stop looping this trigger but if webserver restart, this logic would happen again. "RemovedTriggerId" also is only valid in memory layer only, as it is used to clean scheduleUI cache.

azkaban-common/src/main/java/azkaban/trigger/builtin/ExecuteFlowAction.java
ShellyWEI: good catch let me clean it up

tsangz2013:
Can you also add to the description about in what cases / frequency will check and add the invalid triggers into the list, and also in what cases / frequency will fetch the list and do the cleanup?

ShellyWEI:
> 

It would every time when we detect a null flow/project case. Frequency would be 1 minute scanning frequency. This is accompanying with trigger scanner logic itself. Not a standalone cleanup logic.
We just find them -> remove them -> mark them removed so next time UI wants to fetch them tell them not to get those removed id. 
As simple as it is.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
tsangz2013: So is it true that, though removed from memory at this time, if they are not really correctly cleaned due to exception, they will still be captured and cleaned in the next round?

ShellyWEI: yes, the "next time" means when web server restarts. Those triggers will be re-detected and try to cleanup again.
As for "this time". During this session, we don't have reload logic from DB. We only can do that if user are adding/updating schedules. So they're gone in the code this time.

tsangz2013 (approved):
Thanks for explaining. lgtm!

PR #3285, opened 2023-05-02
fix flow trigger lost caused by failure project upload

ShellyWEI:
## Summary
Azkaban allows users to define flow triggers in HadoopDSL format, they're parsed and scheduled in the process of upload. The initial order is unscheduleTrigger -> handleUploadZip -> some postHandleWork such as rescheduleTrigger. However handleUploadZip step is a bit fragile and might fail due to intermittent issues from infra side, for example, artifatory down. It would lead to project upload in a broken stage where the previous triggers were removed by that broken upload. This PR is to change the order of unscheduleTrigger, moving it after uploadZip.

## Details
- adjust order of unschedule QuartzJob, so even if uploadFile fails, the previous trigger remains unimpacted.
- fix remove schedule logic by using getAllSchedule instead of getUpdatedSchedule.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
@@ -112,7 +112,10 @@ public class ScheduleManager implements TriggerAgent {
     return new ArrayList<>(this.scheduleIDMap.values());
   }
 
-  public List<Schedule> getALlSchedules() {
+  /**
+   * Retrieves a copy of the list of all schedules regardless the status, active or inactive ones.
+   */
+  public List<Schedule> getAllSchedules() {
     try {
       return this.loader.loadAllSchedules();
     } catch (ScheduleManagerException e) {
--- a/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
+++ b/azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
@@ -103,13 +103,22 @@ public class ScheduleManager implements TriggerAgent {
   }
 
   /**
-   * Retrieves a copy of the list of schedules.
+   * Retrieves a copy of the list of updated schedules.
    */
   public List<Schedule> getSchedules()
       throws ScheduleManagerException {
     final List<Schedule> updates = this.loader.loadUpdatedSchedules();
     refreshLocal(updates);
     return new ArrayList<>(this.scheduleIDMap.values());
+  }
+
+  public List<Schedule> getALlSchedules() {
+    try {
+      return this.loader.loadAllSchedules();
+    } catch (ScheduleManagerException e) {
+      logger.error("Failed to load all schedules", e);
+      return new ArrayList<>();
+    }
   }
 
   /**
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -2051,7 +2051,7 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
     final Set<String> flowNameList = project.getFlows().stream().map(f -> f.getId()).collect(
         Collectors.toSet());
 
-    for (final Schedule schedule : scheduleManager.getALlSchedules()) {
+    for (final Schedule schedule : scheduleManager.getAllSchedules()) {
       if (schedule.getProjectId() == project.getId() &&
           !flowNameList.contains(schedule.getFlowName())) {
         scheduleManager.removeSchedule(schedule);
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -1938,18 +1938,20 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
       out = new BufferedOutputStream(new FileOutputStream(archiveFile));
       IOUtils.copy(item.getInputStream(), out);
       out.close();
-      if (this.enableQuartz) {
-        //todo chengren311: should maintain atomicity,
-        // e.g, if uploadProject fails, associated schedule shouldn't be added.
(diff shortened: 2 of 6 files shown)


azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
```diff
@@ -112,6 +112,15 @@ public List<Schedule> getSchedules()
     return new ArrayList<>(this.scheduleIDMap.values());
   }
 
+  public List<Schedule> getALlSchedules() {
```
byjiang1996, line 115: ALi -> All

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1938,18 +1938,20 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
...
+      final List<Flow> existingFlows = project.getFlows();
 
+      // validate project zip's dependencies and persist the new project metadata into DB
       final Map<String, ValidationReport> reports = this.projectManager
           .uploadProject(project, archiveFile, lowercaseExtension, user, props, uploaderIPAddr);
+
+      // Post-processing after upload
+      // reschedule data triggers if quartz is enabled
       if (this.enableQuartz) {
+        this.scheduler.unschedule(existingFlows, projectName);
```
byjiang1996, line 1954: What will happen if project is uploaded successfully, then a flow trigger arrives before `unschedule` is called under these two cases?
1. Flow exists
2. Flow is renamed or deleted

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
byjiang1996: Also could you please elaborate more on the difference between getSchedules and getALlSchedules? I'm curious about why we use the latter in removeScheduleOfDeletedFlows but use the former in all other places.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ShellyWEI: what do you mean by flow trigger arrives?
I assume you're asking when flow trigger is working at the moment of upload?
The flowTrigger system is using Quartz library, it does not have the concept of "flow". What we do is we define the name of a QuartzJob using project.flowName. 
So unschedule means stop that running QuartzJob and delete it forever.
1. if flow exists;
If we schedule again, the flowTrigger itself has the logic to guarantee when a flow trigger is added/scheduled, it immediately checks availability and adding this job to the routine kafka listener thread.  

2. if flow renamed/deleted
the old flow triggers is stopped and removed anyways.

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
ShellyWEI: They actually don't have big difference technically speaking as for the current function.
getSchedules() more focusing on maintaining its own scheduleMap (by getting updated schedules and refresh local map) and returning that map's value. It's possible that some triggers are expired/invalid, it would not be included in the scheduleMap.
getAllSchedule() is more brutal, it's directly fetching all schedules regardless the status. So if we consider deleting non-existing flows, it might be bigger set to consider.

ShellyWEI: We observe several schedules with deleted flow in the system. That's my guess it might not include all the schedules. (Or could be not executing cleanup schedule logic in upload Project)

ShellyWEI: Anyway I thought it's good to include a bigger set of schedules.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
tsangz2013: I believe Binyao is asking what if before `unschedule` is called, the QuartzJob is triggered.

byjiang1996: > I believe Binyao is asking what if before unschedule is called, the QuartzJob is triggered.

Yes. that's my Q.

azkaban-common/src/main/java/azkaban/scheduler/ScheduleManager.java
byjiang1996: Got it. Thank you so much for your clarification!

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ShellyWEI: each QuartzJob are scheduled by Quartz library based on cron expression, but the details like ExecutionContext is controlled by us (refer to FlowTriggerService Class), we store all the information in JobDetails. FlowTriggerService would start/cancel trigger, which has our self-defined triggerInstance which contains lots of dali dependencies, each triggerInstance we store Project Object directly within it, and when triggerExecution happens, we use the stored project data to start execution, so if new project just updated in DB it will not change flow trigger's project data. 
 
If the flow has been renamed/deleted, but all dali dependencies are available so we're ready to executeFlow, the execution would try to start using old project/flows. But whether it's success to execute or not is not important and this should not affect uploading process. 
So even if flow was deleted/renamed, the schedule new trigger instance would succeed because JobKey contains flowName is different.

ShellyWEI: but I think if we want to prevent the execution on old flow, we might need to change project cache a bit: in order to always reflecting new project metadata right after persist project in DB.
then change the fetch project in triggerInstance to use ProjectId to get most updated record.
This might be a big change I wonder if it's worth the effort.

byjiang1996: That's nice to hear. Thanks! I think executing old flows is fine (good workaround already).

byjiang1996 (approved):

PR #3288, opened 2023-05-09
Add UploadLock in project level and have a feature flag enable.project.adhoc.upload to remove the lock

ShellyWEI:
### Summary
This project aims to protect grid security by attaching an Upload Lock to each project. The Upload Lock can be enabled by setting the config value azkaban.upload.privilege.user to the name of a privileged user who can upload projects. If this value is not set, no lock will be applied and the project will behave as usual. Once the privileged user has deployed a project, the Upload Lock will be automatically activated and prevent any ad hoc uploads from other users, unless the ad hoc upload feature flag is enabled. The ad hoc upload feature flag can only be accessed by azkaban admin. If it is enabled, it will override the Upload Lock and allow any user to upload projects, regardless of whether the privileged user has deployed or not.

### Details
    - Introduce Project Level Feature Flag as current azkaban does not have project level properties;
    - Add two ajax API to change project upload settings, changeUploadSettings will require azkaban admin to access
    - Auto add lock during privilege deployment if applies, disable adhoc upload when UploadLock is on
    - Add status on UI and allow changes directly from UI. (See pic)
    - Disable Upload Button once UploadLock is set. (See pic)
    - Hide UploadSetting panel if no privilege user is provided.

### Tests Done
Below testing are performed based on the upload privileged user is "crt", using CRTLock in the testing plan:
>1.AdhocUpload=true -> CRTLock=false
>         -> deploy by crt, the state doesn't change.
>         -> deploy manually, the state doesn't change.
>         -> restart web server, the state doesn't change.
>         -> disable adhocProject 
>                        ---> lastModifierUser is crt, CRTLock=true
>                        ---> lastModifierUser is not crt, the state doesn't change

> 2.AdhocUpload=false -> deploy by CRT, CRTLock = true
>         -> restart web server, the state doesn't change.
>         -> deploy manually, blocked/Forbidden
>         -> deploy on CRT, ok
>         -> enable adhocUpload, CRTLock=false

init state with CRT lock on
Screenshot 2023-05-08 at 2 52 08 PM
Disabled it by enable adhoc feature flag
Screenshot 2023-05-08 at 2 52 37 PM
Succeed
Screenshot 2023-05-08 at 2 52 52 PM
Hide upload when lock is on
Screenshot 2023-05-15 at 3 56 08 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/project/FeatureFlag.java
+++ b/azkaban-common/src/main/java/azkaban/project/FeatureFlag.java
@@ -1,0 +1,35 @@
+package azkaban.project;
+
+/**
+ * A Feature Flag class contains enum of all feature flags.
+ * */
+public enum FeatureFlag {
+
+  ENABLE_PROJECT_ADHOC_UPLOAD("enable.project.adhoc.upload");
+
+  private final String name;
+
+  FeatureFlag(String name) {
+    this.name = name;
+  }
+
+  public static FeatureFlag fromString(String key) {
+    for (FeatureFlag featureFlag : FeatureFlag.values()) {
+      if (featureFlag.getName().equals(key)) {
+        return featureFlag;
+      }
+    }
+    throw new IllegalArgumentException("FeatureFlag not found for key: " + key);
+  }
+
+  @Override
+  public String toString() {
+    return getName();
+  }
+
+  public String getName() {
+    return this.name;
+  }
+
+}
+
--- a/azkaban-common/src/main/java/azkaban/project/Project.java
+++ b/azkaban-common/src/main/java/azkaban/project/Project.java
@@ -27,7 +27,6 @@ import azkaban.user.User;
 import azkaban.utils.Pair;
 import com.google.common.collect.ImmutableMap;
 import java.util.concurrent.ConcurrentHashMap;
-import javax.annotation.Nonnull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -50,6 +49,9 @@ public class Project extends EventHandler {
       new LinkedHashMap<>();
   private final HashSet<String> proxyUsers = new HashSet<>();
   private boolean active = true;
+  private boolean isUploadLocked = false;
+
+  private Map<FeatureFlag, Object> featureFlags = new HashMap<>();
   private String description;
   private int version = -1;
   private long createTimestamp;
@@ -97,7 +99,8 @@ public class Project extends EventHandler {
     final int version = (Integer) projectObject.get("version");
     final Map<String, Object> metadata =
         (Map<String, Object>) projectObject.get("metadata");
-
+    final Map<String, Object> featureFlagsMap = (Map<String, Object>) projectObject.get("featureFlags");
+    final Boolean isUploadLocked = (Boolean) projectObject.get("isUploadLocked");
     final Project project = new Project(id, name);
     project.setVersion(version);
     project.setDescription(description);
@@ -106,6 +109,20 @@ public class Project extends EventHandler {
     project.setLastModifiedUser(lastModifiedUser);
     project.setActive(active);
     project.setUploadUser(uploadUser);
+    if (isUploadLocked != null) {
+      project.setUploadLock(isUploadLocked);
+    }
+
+    if (featureFlagsMap != null) {
+        featureFlagsMap.forEach((key, value) -> {
+          try {
+            project.addFeatureFlags(FeatureFlag.fromString(key), value);
+            logger.debug("load Feature flags key {} value {} for project {}", key, value, name);
+          } catch (final IllegalArgumentException e) {
+            logger.error("Error while parsing feature flags key {}/value {} for project {}", key, value, name, e);
+          }
+        });
+    }
 
     if (source != null) {
       project.setSource(source);
@@ -359,6 +376,9 @@ public class Project extends EventHandler {
     final ArrayList<String> proxyUserList = new ArrayList<>(this.proxyUsers);
     projectObject.put("proxyUsers", proxyUserList);
 
+    projectObject.put("featureFlags", this.featureFlags);
+    projectObject.put("isUploadLocked", this.isUploadLocked);
+
     return projectObject;
   }
 
@@ -494,6 +514,30 @@ public class Project extends EventHandler {
 
   public void setVersion(final int version) {
     this.version = version;
+  }
+
+  public void setUploadLock(boolean flag) {
+    this.isUploadLocked = flag;
+  }
+
+  public boolean isUploadLocked() {
+    return this.isUploadLocked;
+  }
+
+  public Map<FeatureFlag, Object> getFeatureFlags() {
+    return this.featureFlags;
+  }
+
+  public boolean isAdhocUploadEnabled() {
+    if(this.featureFlags.containsKey(FeatureFlag.ENABLE_PROJECT_ADHOC_UPLOAD)) {
+      return (boolean) this.featureFlags.get(FeatureFlag.ENABLE_PROJECT_ADHOC_UPLOAD);
+    } else {
+      return false;
+    }
+  }
+
+  public void addFeatureFlags(FeatureFlag featureFlag, Object value) {
+    this.featureFlags.put(featureFlag, value);
   }
 
   public AzkabanEventReporter getAzkabanEventReporter(){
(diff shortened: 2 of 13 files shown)


azkaban-common/src/main/java/azkaban/project/FeatureFlag.java
```diff
@@ -0,0 +1,35 @@
+package azkaban.project;
+
+/**
+ * A Feature Flag class contains enum of all feature flags, each feature flag has a name and a object value.
```
SilhDing, line 4: nit: a-> an

azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -97,7 +99,8 @@ public static Project projectFromObject(final Object object) {
     final int version = (Integer) projectObject.get("version");
     final Map<String, Object> metadata =
         (Map<String, Object>) projectObject.get("metadata");
-
+    final Map<String, Object> featureFlagsMap = (Map<String, Object>) projectObject.get("featureFlags");
+    final Boolean isCRTLocked = (Boolean) projectObject.get("isCRTLocked");
```
SilhDing, line 103: A class variable with the same name `isCRTLocked` is already defined at line 52; if these two are duplicate then we should remove one; otherwise it is better to rename one of them to avoid confusion

azkaban-common/src/main/java/azkaban/project/Project.java
```diff
@@ -496,6 +516,30 @@ public void setVersion(final int version) {
     this.version = version;
   }
 
+  public void isCRTLocked(boolean flag) {
```
SilhDing, line 519: Looks like this is a setter - can we rename it to sth like `setCRTLocked`?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -378,6 +388,14 @@ private void handleAJAXAction(final HttpServletRequest req,
         if (handleAjaxPermission(project, user, Type.ADMIN, ret)) {
           ajaxAddPermission(project, ret, req, user);
         }
+      } else if (API_CHANGE_UPLOAD_SETTING.equals(ajaxName)) {
+        if (handleAjaxPermission(project, user, Type.ADMIN, ret)) {
+          ajaxChangeUploadSetting(project, ret, req, user);
+        }
+      }else if (API_FETCH_UPLOAD_SETTING.equals(ajaxName)) {
```
SilhDing, line 395: nit: space between `}` and `else`

SilhDing (changes requested):

azkaban-common/src/main/java/azkaban/project/Project.java
ShellyWEI: I don't quite get why they're duplicate ones.
This variable is used to reconstruct "isCRTLocked" from DB, it could be null as this is newly introduced. and this variable is not class variable but only inside this method, it will be recycled after.

azkaban-common/src/main/java/azkaban/project/Project.java
ShellyWEI: It was setCRTLocked in the my first version, but I feel like set something locked == pass this value to true. Will setCRTLock(true/false) instead setCRTLocked makes more sense to you?

azkaban-common/src/main/java/azkaban/project/Project.java
SilhDing: ah okay my bad - this is a static method to create an object from JSON (from DB). So it is not related with the class variable

azkaban-common/src/main/java/azkaban/project/Project.java
SilhDing: Yup `setCRTLock` looks good to me

SilhDing (dismissed):

ypadron-in:
As an alternative solution, have we considered just adding a new project permission (ie: UPLOAD)? If we leverage the project permissions we don't need to implement new apis, new Java classes like FeatureFlag or hard-code LinkedIn specific things like crt user in the open source code. 
To meet the requirement of `it will be automatically enabled when a CRT deployment is completed` perhaps we can use a MySQL stored procedure.

azkaban-web-server/src/web/js/azkaban/view/project-modals.js
```diff
@@ -133,6 +135,51 @@ azkaban.ProjectDescriptionView = Backbone.View.extend({
...
+      if (newText !== '') {
+        $('#adhoc-upload').text(newText);
+      }
+      else {
+        $('#adhoc-upload').text('true or false');
+        $('#adhoc-upload').addClass('editable-placeholder');
+      }
+      $('#adhoc-upload').show();
+    };
+    $.get(requestURL, requestData, successHandler, "json");
```
ypadron-in, line 180: This operation is modifying the project so it should probably be a POST request.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1900,6 +1955,12 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
     if (project == null) {
       return;
     }
+    // fail the upload if the project is CRT locked and the user is not the CRT user
+    if (project.isCRTLocked() && !user.getUserId().equals(CRT_USER)) {
+      registerError(ret, "Installation Failed. Project '" + projectName + " has CRT LOCK on. "
```
ypadron-in, line 1960: Authorization to upload operation is handled in `validateUploadAndGetProject()`. This seems to conflict with the results of that method as it's removing permission to upload to Azkaban admins. Is that the intention? If yes, why?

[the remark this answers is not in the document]
ShellyWEI: it moved into configurable value to avoid CRT word.

ShellyWEI: This intention is to reduce human effect on CICD enabled projects. Azkaban admin would have the adhoc feature flag enabled. Once it is enabled, normal validation would apply on that upload.

azkaban-common/src/main/java/azkaban/project/FeatureFlag.java
```diff
@@ -0,0 +1,35 @@
+package azkaban.project;
+
+/**
+ * A Feature Flag class contains enum of all feature flags.
+ * */
+public enum FeatureFlag {
```
tsangz2013, line 6: The `FeatureFlag` here is only for `project`, right? 
Is there anywhere else can generalize this? like in azkaban-common?

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ProjectPageHeaderTest.java
```diff
@@ -19,7 +19,7 @@ public class ProjectPageHeaderTest {
   @Test
   public void testUploadButtonIsPresent() {
     final VelocityContext context = VelocityContextTestUtil.getInstance();
-    context.put("hideUploadProject", false);
+    context.put("projectUploadLock", false);
```
tsangz2013, line 22: For these 2 changes on the test case, I suggest we keep the one tests, while copying and create new, if the old `hideUploadProject` key is still in use.

azkaban-common/src/main/java/azkaban/project/FeatureFlag.java
ShellyWEI: I was intended to use featureFlag for projects, but feel free to extend it. The initial thought is flow has props concept but we don't have project level props, and I know some other users has mentioned this project level props to us. So I thought it is worthwhile to have a standalone class for that.

azkaban-web-server/src/test/java/azkaban/webapp/servlet/ProjectPageHeaderTest.java
ShellyWEI: no, the old will not work because I overwrite it. This hideUploadProject value depends on projectUploadLock

azkaban-web-server/src/web/js/azkaban/view/project-modals.js
ShellyWEI: Yes technically it should be POST. but our backend ajax handling does not differentiate get or post. I've changed to post already.

tsangz2013 (approved):
Thanks @ShellyWEI for updating the PR title & description, and explaining the design.
lgtm

PR #3289, opened 2023-05-10
Add support to delete output after azkaban.jobtype.examples.java.Word…

byjiang1996:
…Count

### DIFF
--- a/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/examples/java/WordCount.java
+++ b/az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/examples/java/WordCount.java
@@ -46,14 +46,17 @@ public class WordCount extends AbstractHadoopJob {
 
   private final String inputPath;
   private final String outputPath;
-  private boolean forceOutputOverrite;
+  private final boolean forceOutputOverwrite;
+  private final boolean outputDelete;
 
   public WordCount(String name, Props props) {
     super(name, props);
     this.inputPath = props.getString("input.path");
     this.outputPath = props.getString("output.path");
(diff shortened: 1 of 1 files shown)


ShellyWEI (approved):

svnarumugam (approved):
Thank you!:)

az-hadoop-jobtype-plugin/src/main/java/azkaban/jobtype/examples/java/WordCount.java
```diff
@@ -46,14 +46,17 @@ public class WordCount extends AbstractHadoopJob {
...
-  private boolean forceOutputOverrite;
+  private final boolean forceOutputOverwrite;
+  private final boolean outputDelete;
 
   public WordCount(String name, Props props) {
     super(name, props);
     this.inputPath = props.getString("input.path");
     this.outputPath = props.getString("output.path");
-    this.forceOutputOverrite =
+    this.forceOutputOverwrite =
```
svnarumugam, line 56: It looks like we can set this value `force.output.overwrite` to true and then get the concurrent instances passing! :)

byjiang1996: It won't work as expected. Thinking about this timeline in order when two WordCount instances run:
1. Instance1 started
2. Instance2 started
3. Instance1 overwrite the output by deleting the HDFS files
4. Instance2 overwrite the output by deleting the HDFS files
5. Both Instance1 and instance2 tries to writing data to same output.path at same timestamp. HDFS will report exception saying: output.path is written by another process and corrupted

PR #3294, opened 2023-05-23
split missed schedule metrics with back execution info

ShellyWEI:
- Reveal the missed schedule count with back execution feature enabled or not.
- Reveal the count of back execution enabled triggers.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
@@ -52,6 +52,7 @@ public class TriggerManager extends EventHandler implements TriggerManagerAdapte
   private static final Logger logger = Logger.getLogger(TriggerManager.class);
   private static final Map<Integer, Trigger> triggerIdMap = new ConcurrentHashMap<>();
   private static final Set<Integer> removedTriggerIds = new HashSet<>();
+  private static final Set<Integer> backExecuteEnabledTriggerIds = new HashSet<>();
   private final TriggerScannerThread runnerThread;
   private final MetricsManager metricsManager;
   private final Meter heartbeatMeter;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -52,6 +52,7 @@ public class TriggerManager extends EventHandler implements TriggerManagerAdapte
   private static final Logger logger = Logger.getLogger(TriggerManager.class);
   private static final Map<Integer, Trigger> triggerIdMap = new ConcurrentHashMap<>();
   private static final Set<Integer> removedTriggerIds = new HashSet<>();
+  private static final Set<Integer> backExecuteEnabledTriggerIds = new HashSet<>();
```
byjiang1996, line 55: I'm surprised all the Sets in this class are not "ConcurrentHashSet". The data structure is not thread-safe.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -148,6 +156,11 @@ public void updateTrigger(final Trigger t) throws TriggerManagerException {
     this.runnerThread.deleteTrigger(triggerIdMap.get(t.getTriggerId()));
     this.runnerThread.addTrigger(t);
     triggerIdMap.put(t.getTriggerId(), t);
+    if (t.isBackExecuteOnceOnMiss()) {
+      backExecuteEnabledTriggerIds.add(t.getTriggerId());
+    } else {
+      backExecuteEnabledTriggerIds.remove(t.getTriggerId());
```
byjiang1996, line 162: Do you need to run
"if backExecuteEnabledTriggerIds.containsKey(t.getTriggerId())" before removing it?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -164,6 +177,7 @@ public void removeTrigger(final Trigger t) throws TriggerManagerException {
     this.runnerThread.deleteTrigger(t);
     triggerIdMap.remove(t.getTriggerId());
     removedTriggerIds.add(t.getTriggerId());
+    backExecuteEnabledTriggerIds.remove(t.getTriggerId());
```
byjiang1996, line 180: same above.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
ShellyWEI: This is a boolean method, if the value is not present, it simply return false but I don't care the return value.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
byjiang1996: Oh, i see it now. Thanks for your explanation!

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
byjiang1996: Interesting. Thanks for letting me know.

byjiang1996 (approved):

PR #3296, opened 2023-05-23
Emit webServer metrics around cancle flow in general / unreachable / ungraceful kill / failure

tsangz2013:
**Why we need this**
* To allow platform admin, partners and users know better about the overall webServer `cancel flow` experience
* So as can react faster if any issue happens, avoid large-scale killing failure resulting in resources not released and later executions blocked

**What's done** 
Define meters, each accordingly mark:
1. trigger whenever a cancel flow action
2. trigger when fail to reach the execution
3. trigger when fail to gracefully call executor cancel endpoint, force killing the pod
4. trigger when fail the whole thing

**Tests done**
Deployed to a test cluster and verified the killing action will be reflected in the in the metrics dashboard.


byjiang1996 (approved):

tsangz2013:
nit: the branch name is a mistake, can ignore

PR #3298, opened 2023-05-30
pass production flow marker as job props

ShellyWEI:
### Summary
This PR is mainly to pass flow production marker so the each job that extending azkaban's abstract job class would receive that props passed in.


azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
```diff
@@ -100,6 +103,7 @@ public JdbcProjectImpl(final Props props, final DatabaseOperator databaseOperato
     }
     int maxFlowFileSizeInMB = props.getInt(MAX_FLOW_FILE_SIZE_KEY, MAX_FLOW_FILE_SIZE_IN_MB_DEFAULT);
     maxFlowFileSizeInBytes = maxFlowFileSizeInMB * 1024 * 1024;
+    uploadProjectLockFeatureEnabled = props.get(AZKABAN_UPLOAD_PRIVILEGE_USER) != null;
```
byjiang1996, line 106: How about changing
"props.get(AZKABAN_UPLOAD_PRIVILEGE_USER) != null;"
to 
props.get(AZKABAN_UPLOAD_PRIVILEGE_USER, false);

byjiang1996 (dismissed):

azkaban-common/src/main/java/azkaban/project/JdbcProjectImpl.java
ShellyWEI: Ah, I see where the confusion is, AZKABAN_UPLOAD_PRIVILEGE_USER is not a boolean config as it might look like it is. It has the real user define in our config, and I am using it as a button to check if the feature is turned on or off.
When we want to roll out this feature, we can pass the value to job props otherwise we don't.

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
```diff
@@ -16,6 +16,7 @@
 
 package azkaban.project;
 
+import static azkaban.Constants.*;
```
amielkelinkedin, line 19: Why import * here and then import specifics?

az-core/src/main/java/azkaban/Constants.java
```diff
@@ -572,6 +572,8 @@ public static class FlowProperties {
     public static final String AZKABAN_FLOW_SUBMIT_USER = "azkaban.flow.submituser";
     public static final String AZKABAN_FLOW_EXEC_ID = "azkaban.flow.execid";
     public static final String AZKABAN_FLOW_PROJECT_VERSION = "azkaban.flow.projectversion";
+
+    public static final String AZKABAN_FLOW_PRODUCTION_MARKER = "azkaban.flow.production.marker";
```
amielkelinkedin, line 576: In my opinion this is not reflective of what we are doing here. While "production" has some internal meaning, that meaning will surely drift over time. Additionally, this is an OS repo, and production has no external meaning.

azkaban-common/src/main/java/azkaban/executor/ExecutableFlow.java
```diff
@@ -86,7 +86,11 @@ public class ExecutableFlow extends ExecutableFlowBase {
   private String executionPath;
   private ExecutionOptions executionOptions;
   private double azkabanFlowVersion;
+  // flow lock, once locked, the flow won't be able to execute/be modified/be scheduled...
   private boolean isLocked;
+  // If the flow is a part of production project, i.e. the project is enabled with UploadLock,
+  // then the generated executable flow will mark as production flow.
+  private boolean isProductionFlow;
```
amielkelinkedin, line 93: perhaps a better name for this is something along the lines of isProjectLockable?

byjiang1996 (approved):

azkaban-common/src/main/java/azkaban/project/ProjectManager.java
ShellyWEI: thanks, removed.

az-core/src/main/java/azkaban/Constants.java
ShellyWEI: Talked offline. I kind of agree all the points here.
Made additional comment on property name to address the correct usage in the future.
Sorry this has done in a scrappy way in a short time constraint, but I feel we're better to fill the gap when it makes sense to each jobtype who is responsible to manage production access.

PR #3302, opened 2023-06-08
Add the config control to toggle the ability of disabling upload

ShellyWEI:
This allows we control and rollout restrictions inside platform gradually and independently to control the effect to users. azkaban.disable.adhoc.upload is set, we can only disallow the adhoc uploads but it would not affect the lock status being updated at the project level.

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -156,7 +156,7 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
   private boolean lockdownCreateProjects = false;
   private boolean lockdownUploadProjects = false;
   private boolean enableQuartz = false;
-  private boolean disableAdhocUploadClusterFlag = false;
+  private boolean disableAdhocUploadWhenProjectUploadLocked = false;
   private String uploadPrivilegeUser;
   private Map<String, List<HTMLFormElement>> alerterPlugins;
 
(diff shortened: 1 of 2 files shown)


tsangz2013:
This flag impacts the whole webserver that all the users are subjected to it, right?

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1965,8 +1967,11 @@ private void ajaxHandleUpload(final HttpServletRequest req, final HttpServletRes
     if (project == null) {
       return;
     }
+
     // fail the upload if the project is UPLOAD locked and the user is not the upload privilege user
-    if (uploadPrivilegeUser != null && project.isUploadLocked() && !user.getUserId().equals(uploadPrivilegeUser)) {
+    // and the cluster-wide disableAdhocUpload flag is set
+    if (uploadPrivilegeUser != null && disableAdhocUploadClusterFlag
```
svnarumugam, line 1973: [Clarification] It appears to me that `lockdownUploadProjects` flag can be used in place of this. Could you help me understand why we need this flag in this place?

ShellyWEI: We don't want to misuse the flags while they appear to have different meanings.
"lockdownUploadProjects" is disallow **ALL** the uploads across the cluster unless it is from admin users and users with special permissions.
"disableAdhocUploadClusterFlag" is disallow the "**upload locked**" projects with the ability to manually upload even if Admin.
I can not apply "lockdownUploadProjects" which would put the restriction on all projects while my intention is to only target "LOCKED" projects.

svnarumugam: Thanks for the explanation @ShellyWEI. Clearer now!

PR #3303, opened 2023-06-13
CleanupManager: only cleaning yarn-apps of executions that are canceled by user or is EXECUTION_STOPPED

tsangz2013:
**Why we need this**
To fix an design fault that kills the yarn-apps of jobs defined to be "Finish Current Running" or "Finish All Possible".

**What's done**
Removed "FAILED/FAILED_FINISHING/FAILED_SUCCEEDED" from to-be-cleaned statuses for yarn-app-kill task.

**Tests done**
* unit test
* tested in staging cluster

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/ContainerCleanupManager.java
@@ -175,18 +175,16 @@ public class ContainerCleanupManager {
             new Pair<>(Duration.ofMinutes(runningFlowValidity), START_TIME))
         .build();
 
+    // The task will find executions of these statuses and kill all their yarn applications,
+    // ONLY under these set of statuses should their yarn-app being killed.
+    // Counter case: if execution is FAILED, yarn-apps of those jobs defined with "fail-finishing"
+    // should continue running until finish, thus cannot clean them up.
     this.recentTerminatedStatusMap = new Builder<Status, Pair<Duration, String>>()
         .put(Status.EXECUTION_STOPPED,
             new Pair<>(Duration.ofMinutes(recentTerminationValidity), UPDATE_TIME))
         .put(Status.KILLED,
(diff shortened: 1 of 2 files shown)


byjiang1996:
> To fix an design fault that kills the yarn-apps of jobs defined to be "Finish Current Running" or "Finish All Possible".

What happens if we don't have this PR? I think FAILED jobs are still okay to be yarn-cleaned?

tsangz2013:
> > To fix an design fault that kills the yarn-apps of jobs defined to be "Finish Current Running" or "Finish All Possible".
> 
> What happens if we don't have this PR? I think FAILED jobs are still okay to be yarn-cleaned?

This assumption was exactly the mistake we made. 
The idea is that, if there's an executionX, it failed because its job1 failed, but job2, job3... are config to be "finish all running", and we kill all the yarn apps of X, then job2 and job3 will also failed, breaking the idea of "finish all running".

byjiang1996:
> This assumption was exactly the mistake we made.

Got it. It makes sense to me now. One more Q: If execution is FAILED, will job2, job3... are possibly Running still due to "finish all running" or they must have been FAILED before the execution is marked as FAILED?

tsangz2013:
> > This assumption was exactly the mistake we made.
> 
> Got it. It makes sense to me now. Thanks for your elaboration!
> 
> One more Q: If execution is FAILED, are job2, job3... possibly RUNNING still due to "finish all running" config or they must have been FAILED before the execution is marked as FAILED?
> 
> Curious why ".put(Status.FAILED, new Pair<>(Duration.ofMinutes(recentTerminationValidity), UPDATE_TIME))" line is removed.

I think in some cases, one job failed will mark the execution "FAILED", and the execution's other RUNNING jobs defined with "Finish Current Running" or "Finish All Possible" will keep running until finish (no matter they succeed or fail).

byjiang1996 (approved):
LGTM. Thanks for fixing this!

PR #3306, opened 2023-06-22
Fixing bug where proxy user from flow parameter was not included

utk-12:
Adding a bug fix for including proxy users from flow param, in addition to users from the DAG.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -1172,17 +1172,22 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
             Boolean.parseBoolean(flowParam.get(FlowParameters.PROXY_USER_PREFETCH_ALL));
     /*
      As an optimization for not walking through the DAG and making DB queries for large DAGs we
-     try to check if the proxy users list from the project page is less then a certain threshold.
+     try to check if the proxy users list from the project page is less than a certain threshold.
      If it is less than this threshold, we simply add all of them. This configuration will be
      custom to a given environment. An option to fetch all of them via the project page is
      also provided with a flow parameter.
      */
 
-    if (prefetchAllProxyUserCredentials || flow.getProxyUsers().size() <= this.proxyUserPrefetchThreshold ) {
+    if (prefetchAllProxyUserCredentials || flow.getProxyUsers().size() <= this.proxyUserPrefetchThreshold) {
       logger.info("Fetched proxy users from permissions page");
       proxyUsersMap.addAll(flow.getProxyUsers());
       } else{
       Instant proxyUserFetchStartTime = Instant.now();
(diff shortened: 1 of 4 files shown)


utk-12:
Fixed bug for specific instance where there exist large list of proxy users, are not included in any .flow files or .job files but only in the flow parameters/ run time parameters. 
Tested on clusters with above conditions.

azkaban-web-server/build.gradle
```diff
@@ -26,7 +26,7 @@ node {
     npmVersion = '5.6.0'
 
     // Base URL for fetching node distributions (change if you have a mirror).
-    distBaseUrl = 'https://nodejs.org/dist'
+    distBaseUrl = 'https://direct.nodejs.org/dist/'
```
sshardool, line 29: why is this change needed ?

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1172,17 +1173,22 @@ private void createPod(final int executionId)
...
-    if (prefetchAllProxyUserCredentials || flow.getProxyUsers().size() <= this.proxyUserPrefetchThreshold ) {
+    if (prefetchAllProxyUserCredentials || flow.getProxyUsers().size() <= this.proxyUserPrefetchThreshold) {
       logger.info("Fetched proxy users from permissions page");
       proxyUsersMap.addAll(flow.getProxyUsers());
       } else{
       Instant proxyUserFetchStartTime = Instant.now();
+      // First add the proxy user from the top level flow parameter. Adding this here as individual job
+      // may override this for any given reason.
+      if (flowParam != null && flowParam.containsKey(USER_TO_PROXY)) {
+        proxyUsersMap.add(flowParam.get(USER_TO_PROXY));
```
sshardool, line 1190: What's the difference b/w this line and line 1184 ?
I.e `proxyUsersMap.addAll(flow.getProxyUsers());` -> where are these proxy-users fetched from if not the flow-params ?

I assume they are  from the .flow or .job files. However, in that case does not require a DAG walk ? (the comment above this code seems a bit confusing if this is the case)

utk-12: No. they are from the project permissions page. which is the where the comprehensive list lies, to do an authorization check in the beginning.

utk-12: for the .job .flow files we get them from here, ContainerImplUtils.getProxyUsersForFlow(), if we opt to do the DAG walk.

azkaban-web-server/build.gradle
utk-12: Get a Proxy error during building of azkaban-webserver.

utk-12: https://github.com/srs/gradle-node-plugin/issues/292
Faced this issue.

sshardool (approved):
Thanks for addressing this!

PR #3307, opened 2023-06-26
Fix schedule expiration bug

ShellyWEI:
### Bug Summary
When a cron expression has expiration date itself, current system would not handle such edge case properly and everytime the triggers loop over, the trigger that satisfied with trigger condition would try to fire execution everytime. Thus we see the bug presented as the schedule executions flow in minutely frequent level when the schedule should past.

### Fix Details
Even at latest quartz 2.5.0, the getFinalFireTime's implementation is still missing: https://www.javadoc.io/static/org.quartz-scheduler/quartz/2.5.0-rc1/org/quartz/CronExpression.html#getFinalFireTime(). To identify the last execution, we simply check whether the getNextCheckTime updated is in the past, usually getNextCheckTime should update to a future time after now, however, if it is returning a past value, we could treat this trigger as expired state, thus the cleanup/removal logic would be triggered properly.

### Test
scheduled a flow at 14:45 at the time 14:44, expecting only one execution would be triggered.
Screenshot 2023-06-26 at 3 40 51 PM
And until 15:00, the actual matches the expectation.
Screenshot 2023-06-26 at 3 42 07 PM

### DIFF
--- a/azkaban-common/src/main/java/azkaban/trigger/Trigger.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/Trigger.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.log4j.Logger;
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 
 
 public class Trigger {
@@ -190,8 +191,17 @@ public class Trigger {
   }
 
   public void updateNextCheckTime() {
+    // expireCondition's nextCheckTime is the pre-set value, for most cases it's DEFAULT_SCHEDULE_END_EPOCH_TIME,
+    // It is not truly reflecting the real expiration time, so we need to update trigger manually.
     this.nextCheckTime = Math.min(this.triggerCondition.getNextCheckTime(),
         this.expireCondition.getNextCheckTime());
+    // manually update trigger's expiration status based on whether the nextCheckTime is before now.
+    // As nextCheckTime is supposed to get the next valid time, so if it's before now, it means the trigger is expired.
+    final DateTimeZone timezone = DateTimeZone.getDefault();
+    final DateTime date = new DateTime(this.nextCheckTime).withZone(timezone);
+    if (date.isBeforeNow()) {
+      this.status = TriggerStatus.EXPIRED;
+    }
   }
 
   public long getNextCheckTime() {
--- a/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
+++ b/azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
@@ -392,6 +392,7 @@ public class TriggerManager extends EventHandler implements TriggerManagerAdapte
         } catch (final Throwable th) {
           //skip this trigger, moving on to the next one
           TriggerManager.logger.error("Failed to process trigger with id : " + t, th);
+          th.printStackTrace();
         } finally {
           t.unlock();
         }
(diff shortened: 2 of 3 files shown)


azkaban-common/src/main/java/azkaban/trigger/Trigger.java
```diff
@@ -190,8 +191,17 @@ public static Trigger fromJson(final Object obj) throws Exception {
...
   public void updateNextCheckTime() {
+    // expireCondition's nextCheckTime is the pre-set value, for most cases it's DEFAULT_SCHEDULE_END_EPOCH_TIME,
+    // It is not truly reflecting the real expiration time, so we need to update trigger manually.
     this.nextCheckTime = Math.min(this.triggerCondition.getNextCheckTime(),
         this.expireCondition.getNextCheckTime());
+    // manually update trigger's expiration status based on whether the nextCheckTime is before now.
+    // As nextCheckTime is supposed to get the next valid time, so if it's before now, it means the trigger is expired.
+    final DateTimeZone timezone = DateTimeZone.getDefault();
+    final DateTime date = new DateTime(this.nextCheckTime).withZone(timezone);
+    if (date.isBeforeNow()) {
```
dudcat, line 202: Do we need to worry about DST here? If next execution is at 01:30, and current time is 01:59 PDT, we will mark it as expired, and then one minute later we "fall back" to 01:00 PST (not PDT anymore) and so we should be executing it in 30 minutes.. I think?

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
```diff
@@ -392,6 +392,7 @@ private void checkAllTriggers() throws TriggerManagerException {
         } catch (final Throwable th) {
           //skip this trigger, moving on to the next one
           TriggerManager.logger.error("Failed to process trigger with id : " + t, th);
+          th.printStackTrace();
```
dudcat, line 395: This might be redundant if the logger already adds the stacktrace, but more importantly this is printing to stderr and not the log file configured for the logger.

ShellyWEI: Not really, this log lib does not give us the stacktrace. an example from current web server:

> 2023/06/20 10:58:25.396 +0000 ERROR [TriggerManager] [TriggerRunnerManager-Trigger-Scanner-Thread] [Azkaban] Failed to process trigger with id : Trigger Id: 832072, Description: Trigger from SimpleTimeTrigger with trigger condition of BasicTimeChecker_1.eval() and expire condition of EndTimeChecker_1.eval(), Execute flow selfidETL from project selfid_debias_lasso_runs
> java.lang.ArrayIndexOutOfBoundsException

It's hard to investigate where this Index issue coming out.

azkaban-common/src/main/java/azkaban/trigger/Trigger.java
ShellyWEI: it's fine. The nextCheckTime we calculated, will only based on PST, it will not give us the 01:30PDT.

azkaban-common/src/main/java/azkaban/trigger/TriggerManager.java
dudcat: log4j should do the right thing, it says it will include the stacktrace:
https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#error-java.lang.CharSequence-java.lang.Throwable-
But if we are not using log4j here or it's configured differently, then just pass th.getStackTrace() concatenated with the message to the logger. th.printStackTrace() won't pipe the stacktrace to the right place, it's going to STDERR...

ShellyWEI: make sense, updated

ShellyWEI: I actually fix it by fillInStacktrace, we're doing the rethrowing exception a lot, might lost the stacktrace somewhere. As I've tested it I think the configs are fine and successfully print the stacktrace.

dhilsand (approved):

PR #3308, opened 2023-06-27
Updating db

amielkelinkedin:


svnarumugam (approved):
Thanks for clarifying questions about this change on the slack thread.
Adding it [here](https://linkedin-randd.slack.com/archives/GM6H48Q68/p1687986466171449) if anyone is curious!

PR #3311, opened 2023-07-12
CleanupManager: emit metrics that count the number of stale flows of different statuses

tsangz2013:
**Why we need this**
By adding these metrics, we got to know the number of stale flows of each concerning statuses, and with this info, we can set up alerts to react to incidents quicker.

**Tests done**
Tested in our staging cluster's webserver, the metrics correctly got published to the metrics system and dashboard can view the outcome.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/metrics/ContainerizationMetricsImpl.java
+++ b/azkaban-common/src/main/java/azkaban/metrics/ContainerizationMetricsImpl.java
@@ -16,11 +16,14 @@
 
 package azkaban.metrics;
 
+import azkaban.executor.Status;
 import azkaban.utils.Props;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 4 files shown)


azkaban-common/src/main/java/azkaban/metrics/ContainerizationMetricsImpl.java
```diff
@@ -185,6 +203,13 @@ public void sendCleanupStaleFlowHeartBeat() {
     cleanupStaleFlowHeartBeat.mark();
   }
 
+  @Override
+  public void markCleanupStaleStatusFlowNumber(String status, long n){
```
byjiang1996, line 207: do you think this function name is a little bit confusing? It says mark but its impl is to "add".

tsangz2013: good catch, let me update

byjiang1996 (approved):
lgmt. ty!

PR #3317, opened 2023-07-18
allow disabling Ambassador on web server and call pod's service directly

SilhDing:
This PR allows us to disable the reverse proxy for the executor resources on k8s pods and the web server uses the flow pod's *Service* endpoint to interact with the executor pods.

If the executor pod is deployed with a Service, it usually can be directly reached from another pod in the same cluster. Thus, if the Azkaban web server is containerized and running in the same k8s cluster, the reverse proxy (e.g. Ambassador) is not needed anymore and the web server can talk to the executor pod with the URL: `http://[serviceName].[namespace]:[servicePod]`. Note that it still holds even the web server pod and executor pod are under different namespaces (tho they have to be in the same k8s cluster).

## How to Enable
To fully enable this feature, two fields in azkaban.properties might need to be changed:

```
# Set to false to disable the reverse proxy
azkaban.executor.reverse.proxy.enabled=false
# Set to false if the Service of the pod only support HTTP
azkaban.executor.client.tls.enabled=false
```

## Test Done
- Tested on a cluster where the web server is **NOT** containerized to make sure the code changes will not break anything;
- Tested on a cluster where the web server **IS** containerized
	- Still enable the reverse proxy to make sure nothing is broken by the code changes;
	- Disable the reverse proxy, and trigger flows to verify:
		- Flow status can be correctly showed on UI;
		- Logs of running jobs can be fetched on UI;
		- Running flows can be killed on UI.

Also confirmed the web server is indeed using the Service endpoint when talking to the executor pods:

```
2023/07/17 22:19:13.363 +0000  INFO [ExecutorApiClient] [215351125@qtp-146799499-0] [Azkaban] URI is: http://fc-svc-z50azweb50-19.grid-integration-testing:54343/container
```
BUG=[BDP-18537]

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
+++ b/azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
@@ -1360,7 +1360,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
     final String namespace = azkProps
         .getString(ContainerizedDispatchManagerProperties.KUBERNETES_NAMESPACE);
 
-    String serviceName = getServiceNameHelper(serviceNamePrefix, clusterName, executionId);
+    String serviceName = getK8sEntityName(serviceNamePrefix, clusterName, executionId);
     Integer servicePort = azkProps.getInt(Constants.ContainerizedDispatchManagerProperties.KUBERNETES_SERVICE_PORT,
         DEFAULT_KUBERNETES_SERVICE_PORT);
     return new Pair<>(String.join(".", serviceName, namespace), servicePort);
@@ -1592,7 +1592,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
    * @return
    */
   private String getVPAName(final int recommendationId) {
-    return String.join("-", this.vpaPrefix, this.clusterName, String.valueOf(recommendationId));
+    return getK8sEntityName(this.vpaPrefix, this.clusterName, recommendationId);
   }
 
   /**
@@ -1603,7 +1603,7 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
    * @return
    */
   private String getServiceName(final int executionId) {
-    return getServiceNameHelper(this.servicePrefix, this.clusterName, executionId);
+    return getK8sEntityName(this.servicePrefix, this.clusterName, executionId);
   }
 
   /**
@@ -1614,35 +1614,19 @@ public class KubernetesContainerizedImpl extends EventHandler implements Contain
    * @return
    */
   private String getPodName(final int executionId) {
-    return getPodNameHelper(this.podPrefix, this.clusterName, executionId);
+    return getK8sEntityName(this.podPrefix, this.clusterName, executionId);
   }
 
   /**
-   * A helper method to generate a k8s service name of the given execution ID.
-   * @param serviceNamePrefix prefix of the k8s service
+   * A helper method to get name of a k8s entity (e.g. Pod or Service) associated with a resource on Azkaban (e.g.
+   * Execution or VPA).
+   * @param prefix name prefix
    * @param clusterName current cluster's name
-   * @param executionId ID of the execution on the pod
-   * @return the k8s service name of the given execution ID
+   * @param resourceId ID of the Azkaban resource
+   * @return the k8s entity name of the given Azkaban resource ID
    */
-  public static String getServiceNameHelper(
-      final String serviceNamePrefix,
-      final String clusterName,
-      final int executionId) {
-    return String.join("-", serviceNamePrefix, clusterName, String.valueOf(executionId));
-  }
-
-  /**
-   * A helper method to generate a k8s pod name of the given flow execution ID.
-   * @param podNamePrefix prefix of the k8s pod of the flow execution
-   * @param clusterName current cluster's name
-   * @param executionId ID of the execution on the pod
-   * @return the k8s pod name of the given flow execution ID
-   */
-  public static String getPodNameHelper(
-      final String podNamePrefix,
-      final String clusterName,
-      final int executionId) {
-    return String.join("-", podNamePrefix, clusterName, String.valueOf(executionId));
+  public static String getK8sEntityName(final String prefix, final String clusterName, final int resourceId) {
+    return String.join("-", prefix, clusterName, String.valueOf(resourceId));
   }
 
   private final Set<State> getImageVersionState(Map<String, String> flowParams) {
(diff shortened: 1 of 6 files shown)


azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1591,7 +1603,7 @@ private String getVPAName(final int recommendationId) {
    * @return
    */
   private String getServiceName(final int executionId) {
```
byjiang1996, line 1605: Shall we create a static function to getServiceName and let getFlowServiceEndpoint to consume that static function.

Given we have two similar impls about getting the service name (one static; one non-static).

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1605,6 +1617,20 @@ private String getPodName(final int executionId) {
...
   }
 
+  /**
+   * A helper method to generate a k8s service name of the given execution ID.
+   * @param serviceNamePrefix prefix of the k8s service
+   * @param clusterName current cluster's name
+   * @param executionId ID of the execution on the pod
+   * @return the k8s service name of the given execution ID
+   */
+  public static String getServiceNameHelper(
```
ShellyWEI, line 1627: If you want to generialize the name generation helper, why not include getPodName too, just serviceNameHelper for getServiceName isn't necessary to add additional layer on unless it is more generic.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
SilhDing: The non-static one is calling the static one to get the service name - we only keep the actual implementation in the static method.

azkaban-web-server/build.gradle
```diff
@@ -15,7 +15,8 @@ buildscript {
 // manage NodeJS distributions, use them from there.
 // ***npm install*** installs all dependencies in package.json. It will only run when changes are made to package.json
 plugins {
-    id "com.moowork.node" version "1.2.0"
+//    id "com.moowork.node" version "1.2.0"
+    id("com.github.node-gradle.node") version "2.2.0"
```
ShellyWEI, line 19: why do we need bump the node version here? It is for web server, not revelant to k8s pod side if I understand correctly

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
```diff
@@ -1354,6 +1341,31 @@ public static String getLabelSelector(final Props azkProps) {
...
+   * Given a flow ID, return the callable endpoint of the flow pod's service where the flow is executed on. The endpoint
+   * is accessible by another pod which resides in the same k8s cluster.
+   * Naming rule of the endpoint: "[serviceName].[namespace]:[servicePort]"
+   * The returned endpoint is presented by a {@link Pair} consisting of the host name and port.
+   *
+   * @param azkProps the Azkaban props
+   * @param executionId the execution ID
+   * @return a {@link Pair} consisting of the endpoint host and port
+   */
+  public static Pair<String, Integer> getFlowServiceEndpoint(final Props azkProps, int executionId) {
```
ShellyWEI, line 1354: nit, final keywords for executionid. (I think everything that not going to be changed would be final here.

ShellyWEI (commented):
a few minor comments, good craftmanship 👍

azkaban-web-server/build.gradle
SilhDing: I cannot build this module on my side without it - always get this error on node setup:

```
Execution failed for task ':azkaban-web-server:nodeSetup'.
> Could not resolve all files for configuration ':azkaban-web-server:detachedConfiguration1'.
   > Could not resolve org.nodejs:node:8.10.0.
     Required by:
         project :azkaban-web-server
      > Could not resolve org.nodejs:node:8.10.0.
         > Could not get resource 'https://nodejs.org/dist/v8.10.0/ivy.xml'.
            > Could not GET 'https://nodejs.org/dist/v8.10.0/ivy.xml'. Received status code 403 from server: Forbidden
```

Are you able to build it locally? If so then I believe the issue is only on me and I can revert the changes

svnarumugam (commented):
I did a first pass of the code. Looks good to me. Excellent PR description Yihang! :) Thank you for working on this.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
SilhDing: that's a good point - did not include getPodName since it is used by others, but let me add it for consistency.

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
SilhDing: thanks for catching it!

azkaban-web-server/build.gradle
ShellyWEI: I can build it locally but I guess it is only because I have that cache copy, the dependency haven't been built for a long time. 
I might encourage to resolve this issue in another PR. As it is reflecting for web server (not directly impact/necessary in your change ) and I do see the entire package changed too, so we don't really know the real impact for that. I would be conservative for any dependency change.

SilhDing: sure, I have reverted this change. Also made my local build work too without this change (disabling the download locally).

azkaban-common/src/main/java/azkaban/executor/container/KubernetesContainerizedImpl.java
ShellyWEI: I actually think it can be simpfied into one method, as they both are just String.join using "-", just a NamePrefix change.
the helper function would be just String.join("-", prefix, clusterName, execId). 
So I feel it would make the helper function into a real reusable helper. : )
Sorry I didn't make it super clear at the beginning.
You can disagree it is just a matter of format, people have different opinion of what a clean code looks like.

SilhDing: yup that makes sense. Updated the PR and now we only have a static helper method `getK8sEntityName` to return the pod, service and VPA name.

ShellyWEI (approved):
thanks for addressing my comments! lgtm

byjiang1996 (approved):

SilhDing:
Tested again on Holdem6 and GHD.

PR #3318, opened 2023-07-19
flow-retry: emit count metrics

tsangz2013:
**Why we need this**
To make flow-retry a fully functional feature, adding observability

**What's dione** 
Emit metrics when restarting the executin, separately for each retry-strategy

**Tests done** 
Tested in staging cluster:
* Started a flow with retry option and kill its yarn-app to fail it 
* The flow got retry and in metric dashboard saw the counted data point

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
+++ b/azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
@@ -24,9 +24,9 @@ public class OnContainerizedExecutionEventListener implements OnExecutionEventLi
   private final ExecutorLoader executorLoader;
   private final ExecutorManagerAdapter executorManagerAdapter;
   private final ProjectManager projectManager;
-  // number of missed schedules with no back execution enabled
+  // number flow retries using "RetryAsNew" strategy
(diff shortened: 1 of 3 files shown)


byjiang1996 (dismissed):
Discussed with Zhi in offline. It may be worthwhile to emit metrics to flow life cycle event as well so it can break down the retryAsNew and disableSucceeded count to project/flow level and can be easily aggregated by certain time window.

LGTM.

azkaban-common/src/main/java/azkaban/executor/OnContainerizedExecutionEventListener.java
```diff
@@ -7,8 +7,10 @@
 import azkaban.DispatchMethod;
 import azkaban.flow.Flow;
 import azkaban.flow.FlowUtils;
+import azkaban.metrics.MetricsManager;
 import azkaban.project.Project;
 import azkaban.project.ProjectManager;
+import com.codahale.metrics.Counter;
```
SilhDing, line 13: I am wondering what this dependency is; if we use this Counter, how can we get the metrics value then?
If we want to integrate with InGraphs then we might need to use the internal libraries for this so the value will be reflected on InGraphs.

SilhDing (approved):

PR #3319, opened 2023-07-19
Disable job property edit permission when project lock is set

ShellyWEI:
### Summary
This PR is the third step to address the security which restricted the permission to edit job properties when project lock is on. This limitation is aiming to protect production projects from any adhoc modifications unless the projects have to be gone through a review process and reuploaded by upload privileged user who intented to be set by a deployment automation user. This restriction would also be put under a feature flag azkaban.disable.job.props.override.on.locked. 

### Details
- Disable "Job Properties" Edit via UI and via API when project lock is set;
- Fix the lock status when adhocUpload restriction is not set, we should still properly change the project lock status;
- UI improvements:
  1. hide edit button for job properties when lock is on;
  2. show adhocUpload editable only when this feature flag is turned on; Normally we only show the status of project lock in UI;

### Test Done
- test feature flag not set, job property edit function normal;
- test feature flag azkaban.disable.job.props.override.on.locked=true, and project prod identifier is set, no edit button accessible.
- test feature flag azkaban.disable.job.props.override.on.locked=true, project prod identifier not set, job property edit function normal

### DIFF
--- a/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
+++ b/azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
@@ -442,7 +442,7 @@ public class ProjectManagerServlet extends LoginAbstractAzkabanServlet {
         }
       } else if (API_SET_JOB_OVERRIDE_PROPERTY.equals(ajaxName)) {
         if (uploadPrivilegeUser != null && disableJobPropsOverrideWhenProjectUploadLocked && project.isUploadLocked()) {
-          ret.put(ERROR_PARAM, "Project " + projectName + "is locked for edit job property while upload privilege user "
+          ret.put(ERROR_PARAM, "Project " + projectName + "is locked for editing job property while upload privilege user "
               + "is set to " + uploadPrivilegeUser + ". If you really need to edit job property, "
               + "please contact oncall to remove this lock.");
           resp.setStatus(HttpServletResponse.SC_FORBIDDEN);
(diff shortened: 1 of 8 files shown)


azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -435,6 +441,13 @@ private void handleAJAXAction(final HttpServletRequest req,
           ajaxFetchJobInfo(project, ret, req);
         }
       } else if (API_SET_JOB_OVERRIDE_PROPERTY.equals(ajaxName)) {
+        if (uploadPrivilegeUser != null && disableJobPropsOverrideWhenProjectUploadLocked && project.isUploadLocked()) {
+          ret.put(ERROR_PARAM, "Project " + projectName + "is locked for edit job property while upload privilege user "
```
SilhDing, line 445: nit: editing

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
```diff
@@ -1391,9 +1404,18 @@ private void handleProjectLogsPage(final HttpServletRequest req,
...
+        // show if a project has prod identifier
         page.add("projectUploadLock", uploadPrivilegeUser != null && project.isUploadLocked());
         page.add("adhocUpload", project.isAdhocUploadEnabled());
         page.add("showUploadLockPanel", uploadPrivilegeUser != null);
+        // only show adhocUpload changeable button when this feature is enabled
+        page.add("showAdhocUploadFeature",
+            uploadPrivilegeUser != null && disableAdhocUploadWhenProjectUploadLocked);
+        // hide upload project button when project prod identifier is set
+        page.add("hideUploadProjectButton",
+            uploadPrivilegeUser != null && disableAdhocUploadWhenProjectUploadLocked && project.isUploadLocked());
```
SilhDing, line 1418: I see these lines are repeated several times in the same file - do you think we can have a common function to populate the common fields for `page`?

SilhDing (changes requested):

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ShellyWEI: these pages has different properties, but I do extract what it makes sense to project sidebars. Project sidebar is shared component to those different web pages.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
ShellyWEI: updated.

azkaban-web-server/src/main/java/azkaban/webapp/servlet/ProjectManagerServlet.java
SilhDing: Thanks for making the changes.

SilhDing:
LGTM - thanks for working on this!

SilhDing (approved):

PR #3320, opened 2023-08-03
Azkaban security enhancement by cert management

ShellyWEI:
### Summary
This PR is to implement a security update in Azkaban to enhance data protection and access control. Grestin certificates will now utilize a new principal name format for those projects that not uploaded by upload privilege user. This change affects all clusters and is designed to safeguard production data integrity. Existing flows remain unaffected, while new flows will require updated ACLs to accommodate the certificate principal name change. Detailed change would come in init containers following this change.

### Details
- Introduce SecurityTag in each azkaban resource (project.flow) to identify if a flow should subject to new security cert model;
- Only projects are not uploaded by upload-privileged-user and this is NEW_FLOW, the adhoc cert fetch ENV would be passed into init container;
- Otherwise, FETCH_ADHOC_CERT should set to false, means everything remains unchanged.

### Test Done
1. no flag was turned on, expect to see FETCH_ADHOC_CERT = true flag turned on:
2. new project upload && prod_identifier is off, expect to see FETCH_ADHOC_CERT = true yes
3. new project re-upload with different flows && prod_identifier is off, expect to see FETCH_ADHOC_CERT = true yes
4. new project flow && prod_identifier is off, expect to see FETCH_ADHOC_CERT = true  yes new project flow reupload again && prod_identifier is off, expect to see FETCH_ADHOC_CERT = true yes
5. new project/new project flow && crt lock is on, expect to see FETCH_ADHOC_CERT = false yes
6. existing flows == legacy flow, expect to see FETCH_ADHOC_CERT = false yes

### DIFF
--- a/az-core/src/main/java/azkaban/Constants.java
+++ b/az-core/src/main/java/azkaban/Constants.java
@@ -463,6 +463,9 @@ public class Constants {
 
     public static final String SECURITY_USER_GROUP = "azkaban.security.user.group";
 
+    // A config to control whether to enable new security model to dynamically change ACLs
+    public static final String ENABLE_SECURITY_CERT_MANAGEMENT = "azkaban.enable.security.cert.management";
+
     public static final String CSR_KEYSTORE_LOCATION = "azkaban.csr.keystore.location";
 
     // dir to keep dependency plugins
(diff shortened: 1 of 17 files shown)


byjiang1996 (approved):
LGTM. Both PR description and code are nicely written!

PR #3321, opened 2023-08-07
Provide “job.run.on.java11=true” Job Config to launch Job JVM with Ja…

byjiang1996:
…va11 runtime

1. java11_command will point to the JAVA binary specified by "java11.command" config in commonprivate.properties
2. job.run.on.java11 can be either set to default per JobType in <jobtype>/plugin.properties or set in ad-hoc by users per Az Job in Job Configs.

### DIFF
--- a/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
+++ b/azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
@@ -38,6 +38,8 @@ public class JavaProcessJob extends ProcessJob {
   public static final String MAIN_ARGS = "main.args";
   public static final String JVM_PARAMS = "jvm.args";
   public static final String GLOBAL_JVM_PARAMS = "global.jvm.args";
+  public static final String RUN_ON_JAVA11_PARAMS = "job.run.on.java11";
+  public static final String JAVA11_BINARY_PATH = "java11.binary.path";
   public static final String DEPENDENCY_CLS_RAMP_PROP_PREFIX = "azkaban.ramp.jar:";
   public static final String DEPENDENCY_REG_RAMP_PROP_PREFIX = "azkaban.ramp.reg:";
   public static final String DEPENDENCY_CFG_RAMP_PROP_PREFIX = "azkaban.ramp.cfg:";
@@ -48,6 +50,7 @@ public class JavaProcessJob extends ProcessJob {
   public static final String NATIVE_LIBRARY_PATH_PREFIX = "native.library.path.";
 
   public static String JAVA_COMMAND = "java";
+  public static String DEFAULT_JAVA11_COMMAND = "java11";
 
   public JavaProcessJob(final String jobid, final Props sysProps, final Props jobProps,
       final Logger logger) {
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
```diff
@@ -48,6 +50,7 @@ public class JavaProcessJob extends ProcessJob {
   public static final String NATIVE_LIBRARY_PATH_PREFIX = "native.library.path.";
 
   public static String JAVA_COMMAND = "java";
+  public static String DEFAULT_JAVA11_COMMAND = "java11";
```
SilhDing, line 53: Just curious: in what case the job is gonna use a non-default Java11 cmd (when java11 is enabled)? And what else the Java11 cmd could be?

SilhDing (approved):

azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
ShellyWEI: I also want to check this default java11, will we create ENV that reflect java11 into a workable java11-msft in pod? 
And do we allow user to custom pass a path_to_java and allow them to install this custom version in the pod?

byjiang1996: Good point. I just changed the getJobProps().getString(JAVA11_COMMAND_PARAMS) to getSysProps().getString(JAVA11_COMMAND_PARAMS). In that way, user cannot modify that config value by themselves. Instead, it will always be the value coming from our commonprivate.properties

utk-12: hi ! Does DEFAULT_JAVA11_COMMAND indicate the binary path ? 
This does assume we will map the java11 sdk where ever it lies to 'java11' right ? 

Also,  do you think JAVA11_COMMAND_PARAMS can be renamed to JAVA11_BINARY_PATH

and. RUN_ON_JAVA11_PARAMS to JAVA11_JVM_PARAMS . it seems a little clearer to me from a naming convention perspective.

utk-12 (changes requested):

azkaban-common/src/main/java/azkaban/jobExecutor/JavaProcessJob.java
byjiang1996: > Also, do you think JAVA11_COMMAND_PARAMS can be renamed to JAVA11_BINARY_PATH & RUN_ON_JAVA11_PARAMS to JAVA11_JVM_PARAMS

Agreed. I also think the naming is confusing...

> This does assume we will map the java11 sdk where ever it lies to 'java11' right ?

Correct. But I think using JAVA11_COMMAND_PARAMS is better to specify the Java11 binary path. not like python and python3, i didn't see java8 and java11 are named java8 and java11 as the binary name. Typically both are named java. Using binary path is probably a better way to distinguish them.

SilhDing, line 53: probably rename to DEFAULT_JAVA11_BINARY _PATH?

byjiang1996: > I also want to check this default java11, will we create ENV that reflect java11 into a workable java11-msft in pod?

I think we can use JAVA11_BINARY_PATH to specify the java11 path. It is more appropriate i think.

byjiang1996: i also thought of this... But i think the default "java11" "java" are commands (shortcuts) but not the actual binary paths.

utk-12 (approved):
LGTM, unless @SilhDing has more comments.

SilhDing (approved):

PR #3322, opened 2023-08-08
ExecutionOptions: fields initialized to null, and only return default value at get()

tsangz2013:
**Why we need this**
Because there's a bug reported that, when user schedule or execute a flow from UI, they didn't set the `failureAction` field, and in their project-zip they set `failureAction=finish_all_possible` but in the executions the default value `finish_currently_running` is used.

Then I found this bug that, even not set from UI, the created ExecutionOptions object still has `non-null failureAction value default to finish_currently_running`, thus when merging with the value from project-zip, it overwrites.

**What is done**
In this PR, the default value only exposed to other classes at `get...()` methods, thus merging / parsing will keeps the fields null if not set.

**Tests done**
* Unit tests
* // TBD(add screenshots) Tested in staging cluster, not-set failureAction field from UI won't overwrite the value from project-zip

### DIFF
--- a/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
+++ b/azkaban-common/src/main/java/azkaban/executor/ExecutionOptions.java
@@ -74,15 +74,17 @@ public class ExecutionOptions {
 
   private Integer pipelineLevel = null;
   private Integer pipelineExecId = null;
-  private Integer queueLevel = 0;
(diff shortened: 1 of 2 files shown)


tsangz2013:
abort: after investigation finding out the `failureAction` field from UI will not overwrite with the value from project-zip until `failureActionOverride` enabled.

PR #3323, opened 2023-08-09
Filter out flowParameters starting with azkaban.{flow|job} for non-Azkaban-admin

tsangz2013:
**Why we need this**
These parameters starting with `azkaban.{flow|job}` should be generated by the system to honestly track the status of the execution / job, but not config-able by users.

**Tests done**
Simple change, unit tests is good enough

### DIFF
--- a/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
+++ b/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
@@ -42,8 +42,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java
```diff
@@ -276,6 +278,11 @@ public static void filterAdminOnlyFlowParams(final UserManager userManager,
       params.remove(FlowParameters.FLOW_PARAM_DISABLE_POD_CLEANUP);
       // Passing test version will be allowed for Azkaban ADMIN role only
       params.remove(FlowParameters.FLOW_PARAM_ALLOW_IMAGE_TEST_VERSION);
+
```
tsangz2013, line 281: This method is used whenever an execution is to be submitted (no matter adhoc or scheduled), see:
https://github.com/azkaban/azkaban/blob/284e8bf518239bf391a3f298fd3c881b070e2035/azkaban-common/src/main/java/azkaban/server/HttpRequestUtils.java#L263

utk-12 (approved):
lgtm

PR #3344, opened 2025-06-03
fixed hadoop, node and native dependencies

kindlerprince:
# Fixed Dependcies issues in Azkaban build

## Overview

This PR updates the build configuration Hadoop, native build and node dependencies.

### 1. changed build.grade 

- Added a step-by-step process to download and import the full certificate chain for `linkedin.jfrog.io`.
- The process includes:
  - Downloading the full certificate chain using `openssl`.
  - Splitting the certificates into separate files.
  - Importing each certificate into the Java truststore using `keytool`.

### 3. Updated Node.js Version

- Updated the Node.js version in `build.gradle` to a newer version (e.g., `12.22.12`) to avoid issues with older versions no longer being available.

## Testing

- Verified that the certificate import process works correctly.
- Confirmed that Gradle can now download dependencies from LinkedIn JFrog Artifactory without SSL errors.
- Tested the build process to ensure it completes successfully.

## Notes

- If you are using Docker, make sure the certificate import commands run in the same container where the build happens.
- If you still encounter SSL errors, check which Java is being used at build time and ensure the truststore is updated for that Java.


vaishnavisangal (approved):

palvaibhav (approved):
LGTM

